PR #26, opened 2016-03-17
update Swagger instructions

hainingzhang:


docs/configure_swagger.md
```diff
@@ -44,39 +34,41 @@ First, you should checkout the Harbor project from github.
...
+* Start the new Harbor build
 ```docker
    docker-compose up
 ```
-* Because session ID is required for using Harbor APIs. **You should log in first from UI by using a browser.**
-* Open another tab in the current browser to keep one session.
-* Input the URL deployed by Harbor project. **You should change the IP address with your deployed Harbor project node.**
+* Because a session ID is usually required by Harbor API, **you should log in first from a browser.**
+* Open another tab in the same browser so that the session is shared between tabs.
+* Enter the URL of the Swagger page in Harbor as below. The <HARBOR_SERVER> should be replaced by the IP address or the hostname of the Harbor server.
```
wknet123: This line missed a word at ' The ... should be'.

PR #42, opened 2016-03-25
update logger for autoenticator.go, passed on docker-compose build

wy65701436:
update the logger for this go file, and delete bee import as the build failure.

### DIFF
--- a/auth/authenticator.go
+++ b/auth/authenticator.go
@@ -17,12 +17,11 @@ package auth
 
 import (
 	"fmt"
-	"log"
 	"os"
 
-	"github.com/vmware/harbor/models"
+	log "github.com/vmware/harbor/utils/log"
 
-	"github.com/astaxie/beego"
+	"github.com/vmware/harbor/models"
 )
 
 // Authenticator provides interface to authenticate user credentials.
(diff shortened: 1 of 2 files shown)


reasonerjt:
LGTM

ywk253100:
LGTM

auth/authenticator.go
```diff
@@ -17,12 +17,11 @@ package auth
 
 import (
 	"fmt"
-	"log"
 	"os"
 
-	"github.com/vmware/harbor/models"
+	log "github.com/vmware/harbor/utils/log"
```
reasonerjt: "github.com/vmware/harbor/utils/log" should be Good, no need to rename

auth/ldap/ldap.go
```diff
@@ -83,7 +83,7 @@ func (l *Auth) Authenticate(m models.AuthModel) (*models.User, error) {
 		return nil, err
 	}
 	if len(result.Entries()) != 1 {
-		log.Printf("Found more than one entry.")
+		log.Infof("Found more than one entry.")
```
reasonerjt: I think this is a warning.

PR #43, opened 2016-03-25
update sql script for role refactor in db

reasonerjt:


Deploy/db/registry.sql
```diff
@@ -65,31 +72,22 @@ create table project (
 );
 
 insert into project values 
-(null, 1, 'library', NOW(), 0, 1);
+(null, 1, 'library', NOW(), NOW(), 0, 1);
```
wknet123: The field 'Project ID' is a not null field, it should not insert a null value as default.

PR #60, opened 2016-03-29
fix potential index out of range error

reasonerjt:
This should probably fix https://github.com/vmware/harbor/issues/58 and https://github.com/vmware/harbor/issues/49


wknet123:
LGTM

PR #66, opened 2016-03-30
Enable user to turn off self registration, and enable admin user to add user.

wknet123:
added self-registration feature.


controllers/base.go
```diff
@@ -96,12 +98,29 @@ func (b *BaseController) Prepare() {
...
 	if authMode == "" {
 		authMode = "db_auth"
 	}
 	b.Data["AuthMode"] = authMode
+
+	var selfRegistration = strings.ToLower(os.Getenv("SELF_REGISTRATION"))
+
+	switch selfRegistration {
+	case "on":
+		if sessionUserID != nil && sessionUserID.(int) != models.SYSADMIN {
```
reasonerjt: This is not needed as a bool variable's default value is false.

controllers/base.go
```diff
@@ -96,12 +98,29 @@ func (b *BaseController) Prepare() {
...
+			enableAddUserByAdmin = false
+		} else {
+			enableAddUserByAdmin = true
+		}
+	case "":
```
reasonerjt: this is not needed

controllers/register.go
```diff
@@ -33,16 +35,37 @@ type RegisterController struct {
...
 }
 
 // SignUp insert data into DB based on data in form.
 func (rc *CommonController) SignUp() {
+
+	if enableAddUserByAdmin {
+		sessionUserID := rc.GetSession("userId")
+		if sessionUserID == nil || sessionUserID.(int) != models.SYSADMIN {
+			log.Error("Self registration can only be used by admin user.\n")
+			rc.Redirect("/signIn", http.StatusFound)
```
reasonerjt: this is for ajax call and won't handle 302

controllers/base.go
```diff
@@ -96,12 +98,29 @@ func (b *BaseController) Prepare() {
...
 		b.Data["Username"] = b.GetSession("username")
+		b.Data["UserId"] = sessionUserID.(int)
 	}
 	authMode := os.Getenv("AUTH_MODE")
 	if authMode == "" {
 		authMode = "db_auth"
 	}
 	b.Data["AuthMode"] = authMode
+
+	var selfRegistration = strings.ToLower(os.Getenv("SELF_REGISTRATION"))
```
reasonerjt: this has to be a global var so each package can read it.

wknet123:
@xiahaoshawn please review my new updates, thanks.

controllers/base.go
```diff
@@ -96,12 +98,27 @@ func (b *BaseController) Prepare() {
...
 	}
 	b.Data["AuthMode"] = authMode
+
+	strSelfRegistration := strings.ToLower(os.Getenv("SELF_REGISTRATION"))
+
+	if strSelfRegistration == "on" {
+		selfRegistration = true
+	}
+
+	if sessionUserID != nil && sessionUserID.(int) != adminUserID {
```
reasonerjt: b.Data["EnableAddUserByAdmin"] = !selfRegistration && isadmin

xiahaoshawn:
which page should stay on after add user?

wknet123:
Admin role user should always can add user when logged in whether the self-registration flag is on or off.

controllers/register.go
```diff
@@ -34,22 +34,36 @@ type RegisterController struct {
...
-		pageTitleKey = "page_title_add_user"
+// AddUserController handles request for adding user with an admin role user
+type AddUserController struct {
+	BaseController
+}
+
+// Get renders the Sign In page, it only works if the auth mode is set to db_auth
+func (ac *AddUserController) Get() {
+
+	if !ac.BaseController.IsAdminLoginedUser {
```
reasonerjt: ac.IsAdminxxx should be fine

controllers/register.go
```diff
@@ -61,8 +75,8 @@ func (rc *CommonController) SignUp() {
 		rc.CustomAbort(http.StatusForbidden, "")
```
reasonerjt: is it rc or ac???

controllers/register.go
```diff
@@ -61,8 +75,8 @@ func (rc *CommonController) SignUp() {
...
 	}
 
-	if enableAddUserByAdmin && !isAdminLoginedUser {
-		log.Error("Self registration can only be used by admin user.\n")
+	if !(rc.BaseController.SelfRegistration || rc.BaseController.IsAdminLoginedUser) {
```
reasonerjt: I think it should be ac right?

controllers/base.go
```diff
@@ -40,6 +40,8 @@ func (c *CommonController) Render() error {
 type BaseController struct {
 	beego.Controller
 	i18n.Locale
+	SelfRegistration   bool
+	IsAdminLoginedUser bool
```
reasonerjt: I think this var is too long, IsAdmin should convey the meaning...

controllers/register.go
```diff
@@ -34,22 +34,36 @@ type RegisterController struct {
 // Get renders the Sign In page, it only works if the auth mode is set to db_auth
 func (rc *RegisterController) Get() {
 
-	if enableAddUserByAdmin && !isAdminLoginedUser {
-		log.Error("Self registration can only be used by admin user.\n")
+	if !rc.BaseController.SelfRegistration {
```
reasonerjt: don't need the BaseController

controllers/register.go
```diff
@@ -32,16 +33,53 @@ type RegisterController struct {
 
 // Get renders the Sign In page, it only works if the auth mode is set to db_auth
 func (rc *RegisterController) Get() {
+
+	if !rc.BaseController.SelfRegistration {
+		log.Error("Registration can only be used by admin user when self-registrion is off.\n")
```
reasonerjt: Error message is wrong

controllers/register.go
```diff
@@ -32,16 +33,53 @@ type RegisterController struct {
...
 
 // SignUp insert data into DB based on data in form.
 func (rc *CommonController) SignUp() {
+
+	authMode := os.Getenv("AUTH_MODE")
+	if !(authMode == "" || authMode == "db_auth") {
+		rc.CustomAbort(http.StatusForbidden, "")
+	}
+
+	if !(rc.BaseController.SelfRegistration || rc.BaseController.IsAdminLoginedUser) {
```
reasonerjt: BaseController is not needed?

controllers/base.go
```diff
@@ -96,12 +100,33 @@ func (b *BaseController) Prepare() {
...
 	}
 	b.Data["AuthMode"] = authMode
+
+	selfRegistration := strings.ToLower(os.Getenv("SELF_REGISTRATION"))
+
+	if selfRegistration == "on" {
+		b.SelfRegistration = true
+	}
+
+	if sessionUserID != nil {
```
reasonerjt: you should be two chunks of if sessionUserID != nil  into one

controllers/base.go
```diff
@@ -96,12 +100,33 @@ func (b *BaseController) Prepare() {
...
+		var err error
+		b.IsAdminLoginedUser, err = dao.IsAdminRole(sessionUserID)
```
reasonerjt: cast it to int?

controllers/register.go
```diff
@@ -32,16 +33,53 @@ type RegisterController struct {
...
+}
+
+// Get renders the Sign In page, it only works if the auth mode is set to db_auth
+func (ac *AddUserController) Get() {
+
+	if !ac.BaseController.IsAdminLoginedUser {
+		ac.Redirect("/signIn", http.StatusFound)
+	}
+
+	authMode := os.Getenv("AUTH_MODE")
```
reasonerjt: Move authMode to basic Controller so there won't be dup code?

controllers/register.go
```diff
@@ -32,16 +33,53 @@ type RegisterController struct {
...
 func (rc *CommonController) SignUp() {
+
+	authMode := os.Getenv("AUTH_MODE")
+	if !(authMode == "" || authMode == "db_auth") {
+		rc.CustomAbort(http.StatusForbidden, "")
+	}
+
+	if !(rc.BaseController.SelfRegistration || rc.BaseController.IsAdminLoginedUser) {
+		log.Error("Registration can only be used by admin role user when self-registration is off.\n")
+		rc.CustomAbort(http.StatusForbidden, "")
```
reasonerjt: why is CommonController's instance called rc?

controllers/base.go
```diff
@@ -38,6 +40,9 @@ func (c *CommonController) Render() error {
 type BaseController struct {
 	beego.Controller
 	i18n.Locale
+	SelfRegistration bool
```
ywk253100: Can we change "SelfRegistration" and "AuthMode" to global variables? We will read them from env only once in this way.

controllers/register.go
```diff
@@ -32,35 +32,70 @@ type RegisterController struct {
 
 // Get renders the Sign In page, it only works if the auth mode is set to db_auth
 func (rc *RegisterController) Get() {
-	authMode := os.Getenv("AUTH_MODE")
-	if authMode == "" || authMode == "db_auth" {
+
+	if !rc.SelfRegistration {
+		log.Error("Registration is disabled when self-registration is off.\n")
```
ywk253100: log.Error is not proper here, maybe Info or Debug is more suitable.

controllers/register.go
```diff
@@ -32,35 +32,70 @@ type RegisterController struct {
...
+		rc.Redirect("/signIn", http.StatusFound)
```
ywk253100: We should response a 403 and redirect a 403 page, or users will confused.

controllers/register.go
```diff
@@ -32,35 +32,70 @@ type RegisterController struct {
...
+	}
+}
+
+// AddUserController handles request for adding user with an admin role user
+type AddUserController struct {
+	BaseController
+}
+
+// Get renders the Sign In page, it only works if the auth mode is set to db_auth
+func (ac *AddUserController) Get() {
```
ywk253100: AddUserController is not needed. RegisterController is enough.

PR #77, opened 2016-04-01
change register to api/users

wy65701436:
Passed on following testing:

1, Signup with self_registration on and off
2, Admin Add user with self_registration on and off
3, Post API call.
4, Post API call with basic auth.

Todo:
No update the API document, will do it once the code review is done.


reasonerjt:
Thanks @wy65701436 
But I think the API should accept json request

wy65701436:
@reasonerjt sure, will update it to accept json data. Thanks for the review, any other comments?

api/user.go
```diff
@@ -117,6 +135,41 @@ func (ua *UserAPI) Put() { //currently only for toggle admin, so no request body
...
+		return
+	}
+
+	userID, err := dao.Register(user)
+	if err != nil {
+		log.Errorf("Error occurred in Register: %v", err)
+		ua.RenderError(http.StatusInternalServerError, "Internal error.")
+		return
+	}
+	if userID == 0 {
```
reasonerjt: Is this necessary

api/user.go
```diff
@@ -31,9 +32,26 @@ type UserAPI struct {
 	userID        int
 }
 
+type userReq struct {
+	UserName string `json:"username"`
```
reasonerjt: Is there a reason not to reuse models.User?

wy65701436: I just follow the project.go under /api. It's ok to use the model.User, but I am not sure add the extension, like `json:"password"`, will impact the DB schema.

api/user.go
wy65701436: Because the dao.Register will return 0 when encounter any error. And I will remove the valid request step as it's duplicate with validate(user models.User) under dao/register.go

reasonerjt: My point is, I remember err == nil && userID != 0 will always be true so the chunk if userID==0 is not necessary?
Is there any case err == nil && userID == 0 is true?

wy65701436: It's an misunderstanding. Already updated the code, please review the latest at your free time. Appreciate it.

api/user.go
```diff
@@ -27,13 +29,34 @@ import (
...
+		authMode = "db_auth"
+	}
+	ua.AuthMode = authMode
+
+	selfRegistration := strings.ToLower(os.Getenv("SELF_REGISTRATION"))
+	if selfRegistration == "on" {
+		ua.SelfRegistration = true
+	}
+
+	if ua.Ctx.Input.IsPost() {
```
wknet123: Why do you check userId in session here? 
I think it should depends on whether self-registration flag is on or off.

api/user.go
```diff
@@ -27,13 +29,34 @@ import (
...
+		sessionUserID := ua.GetSession("userId")
+		if sessionUserID == nil {
+			return
+		}
+	}
+
 	ua.currentUserID = ua.ValidateUser()
```
wknet123: I think this should be differentiated by logged in and not logged in status as for adding sign up process.

static/resources/js/register.js
```diff
@@ -39,9 +39,10 @@ jQuery(function(){
             var isAdmin = $("#isAdmin").val();
             
 			$.ajax({
```
wknet123: Try to user AjaxUtil as for consistency. It's a wrapper of primitive ajax call, you can refer to other JS files for the usage.

api/user.go
```diff
@@ -27,13 +29,34 @@ import (
...
-	userID        int
+	currentUserID    int
+	userID           int
+	SelfRegistration bool
+	IsAdmin          bool
+	AuthMode         string
 }
 
 // Prepare validates the URL and parms
 func (ua *UserAPI) Prepare() {
```
wknet123: This API should meet both for log in as for checking user status and not log in as for signing up when self-registration is on.

api/user.go
```diff
@@ -27,13 +29,34 @@ import (
...
+	}
+	ua.AuthMode = authMode
+
+	selfRegistration := strings.ToLower(os.Getenv("SELF_REGISTRATION"))
+	if selfRegistration == "on" {
+		ua.SelfRegistration = true
+	}
+
+	if ua.Ctx.Input.IsPost() {
+		sessionUserID := ua.GetSession("userId")
```
reasonerjt: Please also consider the case when user uses Basic-Authentication to call POST /api/users, so here you also need to check the http header.

ui/router.go
```diff
@@ -32,7 +32,6 @@ func initRouters() {
 	beego.Router("/login", &controllers.CommonController{}, "post:Login")
 	beego.Router("/logout", &controllers.CommonController{}, "get:Logout")
 	beego.Router("/language", &controllers.CommonController{}, "get:SwitchLanguage")
-	//	beego.Router("/signUp", &controllers.CommonController{}, "post:SignUp")
```
reasonerjt: I think it's OK to remove the SignUp method from the code?

wy65701436: I removed it from the code, what's your suggestion?

api/user.go
```diff
@@ -151,12 +156,18 @@ func (ua *UserAPI) Post() {
 	user := models.User{}
 	ua.DecodeJSONReq(&user)
 
+	if ua.IsBasicAuth {
+		user.Username = ua.UserNameInBasicAuth
```
reasonerjt: Sorry but this seems incorrect, username and password in basic auth header is for identifying the caller of the API not the user to be added.

reasonerjt:
LGTM thanks！

PR #97, opened 2016-04-12
Deploy harbor on kubernetes

perhapszzy:
This pr allows harbor to run on kubernetes. However, it needs a cert and a dns name to host the registry.


reasonerjt:
1)Please fix typo in the name of directories
2)Is it possible to reuse the files in Deploy/ directory?
3)Please write a document to elaborate the design and how to deploy it to k8s.

README.md
```diff
@@ -44,13 +44,66 @@ The host must be connected to the Internet.
...
+        ```
+        add the line at the end of file:
+        ```
+        DOCKER_OPTS="$DOCKER_OPTS --insecure-registry={HOST}"
+        ```
+        and restart docker service
+        ```
+        sudo service docker restart
+        ```
+     2. The registry config file need to know the IP (or DNS name) of the registry, but on kubernetes, you won't know the IP before the service is created. There are several workarounds to solve this problem for now:
```
perhapszzy: I'll add reference point about where it should be changed after checking in the config file.

perhapszzy:
fixed all comments, please take another look. Thanks.

reasonerjt:
Hi, sorry for late response, please resolve the conflict and write a separate .md file for deploying on k8s, add more instruction on some configurations users may want to customize, such as db password, then update Readme.md to reference it.

perhapszzy:
Should addressed all the comments. Please take another look. Thanks.

PR #98, opened 2016-04-12
fix swagger.yaml key duplicated

tinytub:
yaml key "/projects/{project_id}/members/" duplicated.
move get and post block in same "/projects/{project_id}/members/" key


wknet123:
LGTM

PR #102, opened 2016-04-14
expose 443 port on nginx

reasonerjt:


acaranta:
right sorry, did not see that PR ... :/

(However, I wonder if it is really necessary to export port 443 if https is not configured...)

I'll close this PR nevertheless ;)

PR #113, opened 2016-04-15
API for deleting repository and tag 

ywk253100:


reasonerjt:
1)Please initialize the registry client via handler instance, instead of creating Newxxxx functions
2)Please consider making Standard handler abstract, such that we can use different strategy to prepare the request (basic auth, secret ......).

utils/registry/auth/handler.go
```diff
@@ -38,13 +44,35 @@ type Handler interface {
...
 
 // Credential ...
-type Credential struct {
-	// Username ...
-	Username string
-	// Password ...
-	Password string
-	// SecretKey ...
-	SecretKey string
+type Credential interface {
```
reasonerjt: I think just one func such as "PrepareRequest" would be ok

utils/registry/auth/handler.go
```diff
@@ -103,8 +131,10 @@ func (t *standardTokenHandler) AuthorizeRequest(req *http.Request, params map[st
 		return err
 	}
 
-	// TODO support secretKey
-	r.SetBasicAuth(t.credential.Username, t.credential.Password)
+	if t.credential.Type() == basicAuth {
```
reasonerjt: use a func such as PrepareReq and there won't need such if, the code would scale better?

PR #123, opened 2016-04-18
Update for index page title, image detail create time localization.

wknet123:
Update for index page title, image detail create time localization.


ywk253100:
LGTM

PR #137, opened 2016-04-20
add customizing certificate feature into prepare file

saga92:


saga92:
@reasonerjt bugs have been fixed.

PR #140, opened 2016-04-21
add ssl support for send email

xionglun:
Some mail service providers use ssl connection to send/receive emails.
Had test with tencent's enterprise mail success: **smtp.exmail.qq.com:465**

### DIFF
--- a/Deploy/harbor.cfg
+++ b/Deploy/harbor.cfg
@@ -14,6 +14,7 @@ email_server_port = 25
 email_username = sample_admin@mydomain.com
 email_password = abc
 email_from = admin <sample_admin@mydomain.com>
+email_ssl = false
 
 ##The password of Harbor admin, change this before any production use.
 harbor_admin_password= Harbor12345
--- a/Deploy/prepare
+++ b/Deploy/prepare
@@ -29,6 +29,7 @@ email_server_port = rcp.get("configuration", "email_server_port")
 email_username = rcp.get("configuration", "email_username")
 email_password = rcp.get("configuration", "email_password")
 email_from = rcp.get("configuration", "email_from")
+email_ssl = rcp.get("configuration", "email_ssl")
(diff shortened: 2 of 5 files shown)


reasonerjt:
@xionglun What error does this PR try to fix?

xionglun:
@reasonerjt 
image

Currently, code didn't support **SSL** connection!
Mail providers such as [QQ Exmail](https://exmail.qq.com/) don't support non-ssl smtp email send.

This PR try to add **SSL support** for emails.

reasonerjt:
https://golang.org/src/net/smtp/smtp.go#L300

godoc says smtp pkg supports TLS by default, we'll do more testing before make decision on this PR.
Could you let us know what error did you see and you have to make the change to fix it?

xionglun:
Golang's [implementation](https://github.com/golang/go/blob/master/src/net/smtp/smtp.go#L146)  does not use TLS by default, it will upgrade an existing insecure connection to a secure connection if server support.

And you can read **SendMail**'s [comment](https://github.com/golang/go/blob/master/src/net/smtp/smtp.go#L280)
> // SendMail connects to the server at addr, switches to TLS if
> // possible, authenticates with the optional mechanism a if possible,
> // and then sends an email from address from, to addresses to, with
> // message msg.

I think the problem is that: 
Some mail providers don't support insecure connection at all.
So you can't start off as plaintext, then upgrade the connection to an SSL/TLS encrypted one.



Additional Doc:
[SSL vs TLS vs STARTTLS](https://www.fastmail.com/help/technical/ssltlsstarttls.html)

reasonerjt:
@xionglun Thanks I see your point.  Seems smtp package should provide more api...
But as long as this fixes your problem, could you please update the default value of email_ssl to false and add more explanation of this attribute in the .md file?

xionglun:
@reasonerjt  I have squashed my commit, and set email_ssl to false by default.

PR #141, opened 2016-04-21
update changepassword to /api/users/:id/password

wy65701436:
Changes an account password for Guests; old password required, and Guests can only change their own account.

PUT /api/users/:id/password 	

Passed on UI and API test.


reasonerjt:
LGTM, please don't forget to update document accordingly.

PR #160, opened 2016-04-26
Fixed issue with undefined values in admin tab

senk:
Fixes the issue with the "undefined" fields in admin tab:

before:
screenshot 2016-04-26 08 46 53

after:
screenshot 2016-04-26 08 47 27


reasonerjt:
Thanks a lot @senk

PR #162, opened 2016-04-26
Updated admin tab layout to match the width with other tabs

senk:
Updated admin tab layout to match the width with other tabs
now there is no style change when switching between tabs


wknet123:
Thanks, LGTM.

PR #193, opened 2016-05-06
Set admin as all projects admin

ywk253100:


reasonerjt:
Let's not added admin as a member of a project, instead, when a user tries to modify a project check if the user is sys admin, if he is, then the action is allowed.

Issue #38, opened 2016-03-24
missing description.


xiaods:
<img width="447" alt="2016-03-24 3 40 54" src="https://cloud.githubusercontent.com/assets/37678/14011135/0799f608-f1d7-11e5-95b4-8743b774abad.png">

bobbyjunzhang:
@xiaods Thanks for the suggestion! We are asking our organization admin to add the description for us.

bobbyjunzhang:
The description has been added. Project Harbor does not have an independent website yet and such will be added to the description when available.

PR #195, opened 2016-05-09
Sysadmin has all privileges for projects

ywk253100:


ywk253100:
The pr resolves #38

api/member.go
```diff
@@ -241,3 +241,27 @@ func (pma *ProjectMemberAPI) Delete() {
 		return
 	}
 }
+
+//sysadmin has all privileges to all projects
+func listRoles(userID int, projectID int64) ([]models.Role, error) {
+	roles := make([]models.Role, 1)
+	isSysAdmin, err := dao.IsAdminRole(userID)
+	if err != nil {
+		return roles, err
```
wy65701436: It's better to understand if use "return nil, err"

ywk253100: @wy65701436 I think returning an empty array is more reasonable, which can avoid nil pointer exception if user use it directly.

PR #197, opened 2016-05-09
email field data type size changed to 128 characters from 30 characters

nagarjung:
This patch changes the email field to 128 characters from 30 characters.


nagarjung:
This patch refers to the issue number 173.

https://github.com/vmware/harbor/issues/173

reasonerjt:
@nagarjung thanks for being interested, however there's still some check on backend and UI code, so we will fix this issue on different level in one PR.

PR #203, opened 2016-05-09
change password for admin user when in LDAP mode

xiahaoshawn:


controllers/base.go
```diff
@@ -115,17 +116,24 @@ func (b *BaseController) Prepare() {
...
+		}
+		
 		var err error
 		b.IsAdmin, err = dao.IsAdminRole(sessionUserID.(int))
 		if err != nil {
 			log.Errorf("Error occurred in IsAdminRole:%v", err)
 			b.CustomAbort(http.StatusInternalServerError, "Internal error.")
 		}
+	}else {
+		b.IsLdapAdminUser = false
```
wknet123: The variable default value is false no need to set.

views/segment/header-content.tpl
```diff
@@ -57,6 +57,12 @@
 						<li><a id="aChangePassword" href="/changePassword" target="_blank"><span class="glyphicon glyphicon-pencil"></span>&nbsp;&nbsp;{{i18n .Lang "change_password"}}</a></li>
 						<li role="separator" class="divider"></li>
 						{{ end }}
+						  {{ if eq .AuthMode "ldap_auth" }}
+							  {{ if eq .IsLdapAdminUser true }}
+						<li><a id="aChangePassword" href="/changePassword" target="_blank"><span class="glyphicon glyphicon-pencil"></span>&nbsp;&nbsp;{{i18n .Lang "change_password"}}</a></li>
```
wknet123: Shall we make a combination in if condition?

wknet123:
LGTM

PR #235, opened 2016-05-17
Update document to support mirror registry and multiple storage backend

int32bit:
Although those can easily be found in the official page, it's necessary for those who want to learn about it more directly with some demo examples.


reasonerjt:
@int32bit Thanks for the update!  Could you resolve the conflict?

int32bit:
@reasonerjt  Done, thx~

PR #236, opened 2016-05-17
Create rsyslog.conf

yhua123:
Using static 'rsyslog.conf' instead 'sed' command in Dockerfile, this is the modified config file.


reasonerjt:
please merge the to change to one PR.

PR #238, opened 2016-05-17
Update for using static 'rsyslog.conf' instead command 'sed' in  Dockerfile of  the log container

yhua123:
1 add modified rsyslog.conf file in ./Deploy/log/
2 remove 'sed' command in ./Deploy/log/Dockerfile and add "ADD" command for copy modified rsyslog.conf


ywk253100:
LGTM

PR #241, opened 2016-05-17
Fix: sed compatible with mac os env

xiaods:
sed -i need explicit extention on mac, but linux is optinoal.
so for developer need a compatible request on macos


wknet123:
LGTM, thanks.

PR #256, opened 2016-05-20
add orm singleton

saga92:


reasonerjt:
GetOrmer is not thread safe?

wy65701436:
Seems that beego already implement it in the source code, bootstrap will be called once in the new orm().

BTW, can refer to sync.Once to implement thread safe singleton.

saga92:
rewrite the singleton code in a thread safe way, because NewOrm() will new **struct orm** every time. Just make it a absolute singleton. @wy65701436 Refer to beego docs, it says go built-in connection pool is used in beego orm module. @reasonerjt

PR #257, opened 2016-05-20
add api for listing recent logs and changing user profile

wemeya:


reasonerjt:
I think PUT /api/users/{id} is sufficient for updating user information, so the "profile" is not needed in url.

wemeya:
@reasonerjt  PUT /api/users/{id} was now assigned  to set a user as sysadmin.

reasonerjt:
@wemeya,  is it easy to update the code so the API could handle both cases?

I think ideally the PUT /api/users/{id} should handle all the cases to update a user

wemeya:
@reasonerjt  I just saw the code, that's exactly ok, I‘ll handle this.

api/log.go
```diff
@@ -0,0 +1,66 @@
...
+	if len(startTime) <= 0 {
+		log.Errorf("Get parameters error--startTime: %s", startTime)
+		l.CustomAbort(http.StatusBadRequest, "bad request of startTime")
+	}
+	if len(endTime) <= 0 {
+		log.Errorf("Get parameters error--endTime: %s", endTime)
+		l.CustomAbort(http.StatusBadRequest, "bad request of endTime")
+	}
+	var logList []models.AccessLog
+	logList, err = dao.GetRecentLogs(lines, startTime, endTime)
```
ywk253100: The API should provide one user's logs, not all users'.

api/log.go
```diff
@@ -0,0 +1,66 @@
...
+	userID int
+}
+
+//Prepare validates the URL and the user
+func (l *LogAPI) Prepare() {
+	l.userID = l.ValidateUser()
+}
+
+//Get returns the recent logs according to parameters
+func (l *LogAPI) Get() {
```
ywk253100: If I want to get the logs of past 24 hours, the API can handle this?

api/project.go
```diff
@@ -290,5 +289,11 @@ func validateProjectReq(req projectReq) error {
 	if len(pn) > projectNameMaxLen {
 		return fmt.Errorf("Project name is too long")
 	}
```
ywk253100: Not needed any more, as you check the length by function isIllegalLength.

api/log.go
wemeya: @ywk253100 I've changed it to only list logs relevant to the user.

api/log.go
wemeya: @ywk253100  Yes. For example, you can send request like: 
/api/logs?start_time=2016-05-12 00:00:00&end_time=2016-05-13 00:00:00

api/project.go
wemeya: Have changed!

api/project.go
```diff
@@ -287,8 +287,11 @@ func validateProjectReq(req projectReq) error {
 	if len(pn) == 0 {
 		return fmt.Errorf("Project name can not be empty")
 	}
-	if len(pn) > projectNameMaxLen {
-		return fmt.Errorf("Project name is too long")
+	if isIllegalLength(req.ProjectName, projectNameMinLen, projectNameMaxLen) {
```
ywk253100: Please add another restriction: the name of project must be in lower case.

dao/accesslog.go
```diff
@@ -115,3 +115,38 @@ func AccessLog(username, projectName, repoName, repoTag, action string) error {
...
+
+//GetRecentLogs returns recent logs according to parameters
+func GetRecentLogs(userID, linesNum int, startTime, endTime string) ([]models.AccessLog, error) {
+	var recentLogList []models.AccessLog
+	queryParam := make([]interface{}, 1)
+
+	sql := "select user_id, project_id, repo_name, repo_tag, GUID, operation, op_time from access_log where user_id = ?"
+	queryParam = append(queryParam, userID)
+	if startTime != "" || len(startTime) > 0 {
+		sql += " and op_time >= ?"
```
ywk253100: The startTime!="" is equivalent to len(startTime)>0.

dao/accesslog.go
```diff
@@ -115,3 +115,38 @@ func AccessLog(username, projectName, repoName, repoTag, action string) error {
...
+		queryParam = append(queryParam, startTime)
+	}
+
+	if endTime != "" || len(endTime) > 0 {
+		sql += " and op_time <= ?"
+		queryParam = append(queryParam, endTime)
+	}
+
+	sql += " order by op_time desc"
+	if linesNum > 0 {
```
ywk253100: The logic should be in controller.

dao/accesslog.go
```diff
@@ -115,3 +115,38 @@ func AccessLog(username, projectName, repoName, repoTag, action string) error {
 	}
 	return err
 }
+
+//GetRecentLogs returns recent logs according to parameters
+func GetRecentLogs(userID, linesNum int, startTime, endTime string) ([]models.AccessLog, error) {
```
ywk253100: The user of logs returned by this function is not always the login user.

api/log.go
```diff
@@ -0,0 +1,87 @@
...
+	var err error
+	startTime := l.GetString("start_time")
+	if len(startTime) == 0 {
+		startTime = ""
+	} else {
+		i, err := strconv.ParseInt(startTime, 10, 64)
+		if err != nil {
+			l.CustomAbort(http.StatusInternalServerError, "Internal error")
```
ywk253100: Log the error.

dao/accesslog.go
```diff
@@ -115,3 +115,33 @@ func AccessLog(username, projectName, repoName, repoTag, action string) error {
...
+	var recentLogList []models.AccessLog
+	queryParam := make([]interface{}, 1)
+
+	sql := "select log_id, access_log.user_id, project_id, repo_name, repo_tag, GUID, operation, op_time, username from access_log left join  user on access_log.user_id=user.user_id where project_id in (select distinct project_id from access_log where user_id = ?)"
```
ywk253100: You should get all projects which the user is member of, and filter the logs with the project ID.

wemeya: @ywk253100  yes, I know your meaning, but is there anthing wrong with the sql?

api/log.go
wemeya: ok, I will add the log

dao/accesslog.go
wemeya: ok, I will change that.

api/project.go
wemeya: I have added.

dao/accesslog.go
wemeya: good advice!

dao/accesslog.go
wemeya: yes, I have modified.

dao/accesslog.go
ywk253100: If the login user A is a member of project P, but there is no record which projectID=P and useID=A, how could you get the logs of project P?

api/user.go
```diff
@@ -133,14 +135,40 @@ func (ua *UserAPI) Get() {
...
-	dao.ToggleUserAdminRole(userQuery)
+	emailExist, err := dao.UserExists(user, "email")
+	if err != nil {
+		log.Errorf("Error occurred in change user profile: %v", err)
+		ua.RenderError(http.StatusInternalServerError, "Internal error.")
+		return
+	}
+	if emailExist {
+		log.Warning("email has already been used!")
+		ua.CustomAbort(http.StatusForbidden, "email has already been used!")
```
ywk253100: Status code: 409

api/user.go
```diff
@@ -157,7 +185,22 @@ func (ua *UserAPI) Post() {
...
+	}
+	userExist, err := dao.UserExists(user, "username")
+	if err != nil {
+		log.Errorf("Error occurred in Register: %v", err)
+		ua.RenderError(http.StatusInternalServerError, "Internal error.")
+		return
+	}
+	if userExist {
+		log.Warning("username has already been used!")
+		ua.CustomAbort(http.StatusForbidden, "username has already been used!")
```
ywk253100: 409

ywk253100:
LGTM

api/project.go
```diff
@@ -197,10 +199,9 @@ func (p *ProjectAPI) List() {
 	p.ServeJSON()
 }
 
-// Put ...
-func (p *ProjectAPI) Put() {
+// ToggleProjectPublic handles request POST /api/projects/:id/toggle_project_public
```
reasonerjt: this url does not match the actual url of the api

api/user.go
```diff
@@ -228,3 +279,78 @@ func (ua *UserAPI) ChangePassword() {
 		ua.CustomAbort(http.StatusInternalServerError, "Internal error.")
 	}
 }
+
+// ToggleUserAdminRole handles Post api/users/{}/toggleadmin
+func (ua *UserAPI) ToggleUserAdminRole() {
```
reasonerjt: 1)api should match the actual uri of the api
2)As for the implementation, it should be consistent with the project/:id/publicity, i.e. read the value of from the request.

api/user.go
```diff
@@ -228,3 +279,78 @@ func (ua *UserAPI) ChangePassword() {
...
+	if !ua.IsAdmin {
+		log.Warningf("current user, id: %d does not have admin role, can not update other user's role", ua.currentUserID)
+		ua.RenderError(http.StatusForbidden, "User does not have admin role")
+		return
+	}
+	userQuery := models.User{UserID: ua.userID}
+	if err := dao.ToggleUserAdminRole(userQuery); err != nil {
```
reasonerjt: If there's the only reference of toggleuser admin role, it would be better to update to dao.TollgleUserAdminRole(${userid},${adminornot})

api/user.go
```diff
@@ -228,3 +279,78 @@ func (ua *UserAPI) ChangePassword() {
...
+		log.Errorf("Error occurred in ToggleUserAdminRole: %v", err)
+		ua.CustomAbort(http.StatusInternalServerError, "Internal error.")
+	}
+}
+
+func validate(user models.User) error {
```
reasonerjt: why are there validate and commonvalidate?

auth/ldap/ldap.go
```diff
@@ -111,6 +111,9 @@ func (l *Auth) Authenticate(m models.AuthModel) (*models.User, error) {
 		u.Realname = m.Principal
 		u.Password = "12345678AbC"
 		u.Comment = "registered from LDAP."
+		if u.Email == "" {
+			u.Email = u.Username + "@vmware.com"
```
reasonerjt: let's not use vmware.com, use placeholder.com instead.

ui/router.go
```diff
@@ -55,14 +55,16 @@ func initRouters() {
 	beego.Router("/api/projects/:pid/members/?:mid", &api.ProjectMemberAPI{})
 	beego.Router("/api/projects/", &api.ProjectAPI{}, "get:List")
 	beego.Router("/api/projects/?:id", &api.ProjectAPI{})
+	beego.Router("/api/projects/:id/publicity", &api.ProjectAPI{}, "post:ToggleProjectPublic")
```
reasonerjt: publicity and sysadmin api should be PUT rather than POST? as it's changing the attribute of a resource.

PR #261, opened 2016-05-23
store log in fs, provide api for getting log of a job

reasonerjt:
@ywk253100 please review as I made update to utils/log.go


ywk253100:
The callDepth in line 65 of utils/log/logger.go should be 3.

PR #317, opened 2016-06-06
Sync image

ywk253100:


api/base.go
```diff
@@ -51,6 +53,30 @@ func (b *BaseAPI) DecodeJSONReq(v interface{}) {
...
+		message := ""
+		for _, e := range validator.Errors {
+			message += fmt.Sprintf("%s %s \n", e.Field, e.Message)
+		}
+		b.CustomAbort(http.StatusBadRequest, message)
+	}
+}
+
+// DecodeJSONReqAndValidate does both decoding and validation
+func (b *BaseAPI) DecodeJSONReqAndValidate(v interface{}) {
```
reasonerjt: We probably need DecodJSON which does not run validate?

PR #321, opened 2016-06-07
remove .swp files

reasonerjt:


reasonerjt:
@yhua123  try not to use git add --all

PR #324, opened 2016-06-08
Use alpine based for images log, ui and nginx

victoriazhengwf:
Local docker-compose build and start succeeded, log can be generated. The total size is shrunk around 586M as I have described in wechat.

### DIFF
--- a/Deploy/docker-compose.yml
+++ b/Deploy/docker-compose.yml
@@ -5,7 +5,8 @@ services:
     volumes:
       - /var/log/harbor/:/var/log/docker/
     ports:
-      - 1514:514
+      - 1514:514/udp
+      - 2514:10514
   registry:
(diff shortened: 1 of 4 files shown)


reasonerjt:
@victoriazhengwf  this PR will be held for a few days.

Deploy/docker-compose.yml
```diff
@@ -5,7 +5,8 @@ services:
     volumes:
       - /var/log/harbor/:/var/log/docker/
     ports:
-      - 1514:514
+      - 1514:514/udp
+      - 2514:10514
```
reasonerjt: is it necessary to  expose the tcp port if it is not used at all?

reasonerjt:
We've shifted to photon OS.

PR #341, opened 2016-06-15
bug fix: when no rows found, return nil not error

ywk253100:


reasonerjt:
why?

ywk253100:
Adding logic in API codes to handle the ErrNoRows is also OK, but I think handle it in DAO is more reasonable.

PR #346, opened 2016-06-15
add code to list author of top repos

wemeya:


dao/accesslog.go
```diff
@@ -159,5 +159,21 @@ func GetTopRepos(countNum int) ([]models.TopRepo, error) {
 	if err != nil {
 		return nil, err
 	}
+	var repoString string
+	for _, v := range lists {
```
reasonerjt: Why use plural as it's only one list?

dao/accesslog.go
```diff
@@ -159,5 +159,21 @@ func GetTopRepos(countNum int) ([]models.TopRepo, error) {
...
+		repoString += v.RepoName + ","
+	}
+	repoString = repoString[0 : len(repoString)-1]
```
reasonerjt: This will throw runtime error if the list is empty.

reasonerjt:
I may have missed some discussion but there's a question, different tags of a repository may be pushed by different users, so why add "creator" as attribute of a repository?

wemeya:
@reasonerjt added according to the new UI

PR #361, opened 2016-06-20
bug fix: fix concurrent use of map

ywk253100:


ywk253100:
commit this pr with #374

PR #441, opened 2016-06-29
User guide

ywk253100:


docs/user_guide.md
```diff
@@ -119,4 +145,23 @@ Push the image:
...
+
+Repositories deletion runs in two steps.  
+First, delete repositories in Harbor's UI. This is soft deletion. You can delete the entire repository or just a tag of it.  
+
+![browse project](img/new_delete_repository.png)
+
+**Note: If both tag A and tag B reference the same image, after deleting tag A, B will also disappear.**  
+
+Second, delete the real data using registry's garbage colliection(GC). 
+
```
hainingzhang: 1. add some explanation of this command, and if harbor should be stopped or assume no push operation is allowed. 

2. state on which machine this command should be run

hainingzhang: add clean up command

PR #474, opened 2016-07-05
support job retrying

ywk253100:


reasonerjt:
should 
return models.JobRetrying, nil

PR #477, opened 2016-07-05
add auth check as registry does not distinguish 401 and 403

ywk253100:


reasonerjt:
Sorry, too much dup code...

PR #479, opened 2016-07-05
return count of jobs which are failed when listing policies

ywk253100:


reasonerjt:
probably need to count "retrying" as well?

PR #524, opened 2016-07-12
migration guide

hainingzhang:


README.md
```diff
@@ -9,8 +9,10 @@
 Project Harbor is an enterprise-class registry server, which extends the open source Docker Registry server by adding the functionality usually required by an enterprise, such as security, control, and management. Harbor is primarily designed to be a private registry - providing the needed security and control that enterprises require.  It also helps minimize bandwidth usage, which is helpful to both improve productivity (local network access) as well as performance (for those with poor internet connectivity).
 
 ### Features
-* **Role Based Access Control**: Users and Docker repositories are organized via "projects", a user can have different permission for images under a project.
+* **Role based access control**: Users and Docker repositories are organized via "projects", a user can have different permission for images under a project.
+* **Image synchronization**: Images can be synchronized(replicated) between multiple registry instances. Great for load balancing and distributed data centers.
```
reasonerjt: Maybe we should call it "image **replication**" so it's consistent with the UI?

README.md
```diff
@@ -9,8 +9,10 @@
...
 * **Graphical user portal**: User can easily browse, search Docker repositories, manage projects/namespaces.
+* **Garbage collection of images**: Recycle storage space by garbage collection for deleted images.
```
reasonerjt: This is not a part of harbor but part of registry and it's not easy to use currently...

Issue #522, opened 2016-07-12
Sync error: tag the same image with lots of tags, can't see all the tags in destination 


wemeya:
I tagged the hello-world from 1 to 100, and pushed them to the source host.  It displayed correctly in the UI of the source host. Also, I can see a lot of sync records. However, when I turned to the UI of the destination host, it only showed two tags of the image. See below:
src:
![image](https://cloud.githubusercontent.com/assets/12540577/16767090/5a10126a-4870-11e6-93d5-060c4e434ffa.png)
![image](https://cloud.githubusercontent.com/assets/12540577/16767100/6bc9d4dc-4870-11e6-8b4e-b463b123b474.png)
dst:
![image](https://cloud.githubusercontent.com/assets/12540577/16767124/7a0d3656-4870-11e6-83e0-74ca74754de8.png)

reasonerjt:
@ywk253100 I guess this is because the tags have the identical digests?  Could you revisit the logic so even the tags have same digest they should be synced to destination?

ywk253100:
I will fix it.

wemeya:
fixed after check, closed.

Issue #521, opened 2016-07-12
Project publicity changed when sync to the destination


wemeya:
When I sync a public project to the destination, it changed to private. See below:
src: ![image](https://cloud.githubusercontent.com/assets/12540577/16766770/88990b02-486e-11e6-99ee-9e8f704c80e7.png)
dst: ![image](https://cloud.githubusercontent.com/assets/12540577/16766777/98668050-486e-11e6-885b-20b2c821eff2.png)

reasonerjt:
@ywk253100 I think this is the design decision?  Maybe we want to mention that in doc?

ywk253100:
Maybe we can keep the publicity of destination project same with source, I will fix it.

reasonerjt:
ok thanks!

wemeya:
fixed after check, closed.

PR #531, opened 2016-07-13
fix #520 #521 #522

### LINKED ISSUE #520: new policy issue in the  project
when new policy in the project, we checked new destination checkbox to create a new destination. we find that we can create the same endpoint by different user and destination name,  but it should't be allowed.
@wknet123

ywk253100:


job/replication/transfer.go
```diff
@@ -495,13 +505,12 @@ func (m *ManifestPusher) enter() (string, error) {
 	} else {
 		m.logger.Infof("manifest of %s:%s exists on source registry %s, continue manifest pushing", name, tag, m.srcURL)
 
-		_, manifestExist, err := m.dstClient.ManifestExist(m.digest)
+		_, manifestExist, err := m.dstClient.ManifestExist(tag)
```
reasonerjt: Need to compare the digest of given tag to determine if it really dups

job/replication/transfer.go
```diff
@@ -192,7 +192,18 @@ enter:
...
+			c.logger.Errorf("an error occurred while getting project %s on %s: %v", c.project, c.srcURL, err)
+			return "", err
+		}
+
+		isPublic := false
+		if project != nil {
+			isPublic = project.Public == 1
+		}
+
+		err = c.createProject(isPublic)
```
reasonerjt: maybe c.createProject(project.Public) is sufficient as project can never be nil otherwise an error will return?

PR #553, opened 2016-07-15
Replicate repository deletion by calling API of Harbor to log the operation

ywk253100:


reasonerjt:
LGTM, noted this is temporary solution, as soon as docker/distribution supports deleting manifest via tag not digest, we'll revisit and make update.

PR #564, opened 2016-07-18
fix #561

### LINKED ISSUE #561: Error occured while refreshing the ui of destination in a going on sync process
The error seems like: Failed to get [object][object] 
I didn't capture the screen when it occured, it can be reproduced when syncing several repos to a destination and refreshing the ui of detination.

ywk253100:


api/repository.go
```diff
@@ -227,12 +227,18 @@ func (ra *RepositoryAPI) GetTags() {
...
+		regErr, ok := err.(*registry_error.Error)
+		if !ok {
+			log.Errorf("error occurred while listing tags of %s: %v", repoName, err)
+			ra.CustomAbort(http.StatusInternalServerError, "internal error")
+		}
+		// TODO remove the logic if the bug of registry is fixed
+		// It's a workaround for a bug of registry: when listing tags of
+		// a repository which is being pushed, a "NAME_UNKNOWN" error will
+		// been returned, while the catalog API can list this repository.
+		if regErr.StatusCode != http.StatusNotFound {
```
reasonerjt: Is it possible to differentiate if the repo does truly not exist or is the repo partially pushed?

ywk253100: There is no way to differentiate them, they both return the same error.

PR #618, opened 2016-08-01
add security check on job service

ywk253100:


api/jobs/replication.go
```diff
@@ -46,8 +46,25 @@ type ReplicationReq struct {
...
+	}
+
+	if cookie.Value != config.UISecret() {
+		rj.CustomAbort(http.StatusForbidden, "")
+	}
+}
+
 // Post creates replication jobs according to the policy.
 func (rj *ReplicationJob) Post() {
+	rj.authenticate()
```
reasonerjt: I think you can do it in Prepare()?

api/utils.go
```diff
@@ -207,6 +223,16 @@ func postReplicationAction(policyID int64, acton string) error {
 	return fmt.Errorf("%d %s", resp.StatusCode, string(b))
 }
 
+func authenticateRequest(req *http.Request) {
```
reasonerjt: I think you can reuse code from job service?  it should be identical?

PR #640, opened 2016-08-04
add pkg for support coveralls

yhua123:
add coveralls support package


reasonerjt:
Seems this works! Thanks!

PR #692, opened 2016-08-18
update ldap configuration

wemeya:
an example of we discussed, some still need to prove.


auth/ldap/ldap.go
```diff
@@ -65,16 +65,65 @@ func (l *Auth) Authenticate(m models.AuthModel) (*models.User, error) {
...
+	}else {
+		scope = openldap.LDAP_SCOPE_SUBTREE
+	}
+	attributes := []string{"cn", "mail", "email", "uid"}
+	result, err := ldap.SearchAll(ldapBaseDn, scope, filter, attributes)
+	if err != nil {
+		return nil, err
+	}
+	if len(result.Entries()) != 1 {
+		log.Warningf("Found more than one entry.")
```
reasonerjt: Could you dump more debug information such as base dn and filters?

auth/ldap/ldap.go
```diff
@@ -65,16 +65,65 @@ func (l *Auth) Authenticate(m models.AuthModel) (*models.User, error) {
...
+			u.Email = val
+		//some use email
+		case "email":
+			u.Email = val
+		//AD use cn
+		case "cn":
+			u.Username = val
+		}
+	}	
+	log.Debug("Let me see bindDN (result.Base): ", result.Base())
```
reasonerjt: What's this?

auth/ldap/ldap.go
```diff
@@ -65,16 +65,65 @@ func (l *Auth) Authenticate(m models.AuthModel) (*models.User, error) {
...
 
-	err = ldap.Bind(baseDn, m.Password)
+	err = ldap.Bind(bindDN, m.Password)
```
reasonerjt: maybe you should first check the password and if it passed, populate the attributes?

Deploy/harbor.cfg
```diff
@@ -2,7 +2,8 @@
 
 #The IP address or hostname to access admin UI and registry service.
 #DO NOT use localhost or 127.0.0.1, because Harbor needs to be accessed by external clients.
-hostname = reg.mydomain.com
+
+hostname = 10.117.4.245
```
reasonerjt: Use default values in harbor.cfg!

Deploy/prepare
```diff
@@ -99,6 +102,9 @@ render(os.path.join(templates_dir, "ui", "env"),
         harbor_admin_password=harbor_admin_password,
         ldap_url=ldap_url,
         ldap_basedn=ldap_basedn,
+        ldap_filter=ldap_filter,
+        ldap_uid=ldap_uid,
+        ldap_scope=ldap_scope,
```
reasonerjt: you should validate the scope? or mention any invalid values is subtree scope?

auth/ldap/ldap.go
```diff
@@ -65,16 +65,65 @@ func (l *Auth) Authenticate(m models.AuthModel) (*models.User, error) {
...
 	if err != nil {
 		return nil, err
 	}
 	defer ldap.Close()
 
-	scope := openldap.LDAP_SCOPE_SUBTREE // LDAP_SCOPE_BASE, LDAP_SCOPE_ONELEVEL, LDAP_SCOPE_SUBTREE
+/*	scope := openldap.LDAP_SCOPE_SUBTREE // LDAP_SCOPE_BASE, LDAP_SCOPE_ONELEVEL, LDAP_SCOPE_SUBTREE
```
reasonerjt: delete the code if not needed.

auth/ldap/ldap.go
```diff
@@ -65,16 +65,65 @@ func (l *Auth) Authenticate(m models.AuthModel) (*models.User, error) {
...
+		return nil, err
+	}
+	if len(result.Entries()) != 1 {
+		log.Warningf("Found more than one entry.")
+		return nil, nil
+	}
+	en := result.Entries()[0]
+	bindDN := en.Dn()
+	u := models.User{}
+	for _, attr := range en.Attributes() {
```
reasonerjt: So is it possible the entry has no uid nor cn? we probably should check that and fail before inserting to DB?  or shall we use the uid attribute user set in the .cfg as username?

reasonerjt:
please squash the commits.

wemeya:
@reasonerjt squashed.

reasonerjt:
Sorry, this has to goto dev branch.... I didn't check it before

wemeya:
ok, I will open a new PR.

reasonerjt:
And I don't quite like the idea about searchdn, since you cannot ask user to comment something.
I'd suggest you create pr without searchdn support.

wemeya:
So I initially comment the searchdn in harbor.cfg and if needed, ask the user themselves to uncomment it. Is that you mean?

PR #695, opened 2016-08-19
The user can be added again after deletion

ywk253100:


reasonerjt:
1) test it
2) do we need to extend the length of the column in DB?

ywk253100:
@reasonerjt Updated.

PR #705, opened 2016-08-22
support /v1/search API and configurable cache interval

Dieken:
(1) `docker search` uses /v1/search, this patch enables this function.
(2) the expiration interval of repository list cache is hardcoded to
    720s, this patch makes it configurable and 8 hours in configuration
    file by default.
(3) upgrade docker-registry to v2.5 due to a bug in garbage-collect,
    see https://github.com/docker/distribution/issues/1640
(4) relax minimal length of project name to one character, kindly allow
    users to shoot their feet.
(5) relax manifest pattern for Docker 1.7.1, it sends media type
    application/vnd.docker.distribution.manifest.v1+prettyjws.

### DIFF
--- a/Deploy/config/nginx/nginx.https.conf
+++ b/Deploy/config/nginx/nginx.https.conf
@@ -42,6 +42,19 @@ http {
     # required to avoid HTTP 411: see Issue #1486 (https://github.com/docker/docker/issues/1486)
     chunked_transfer_encoding on;
 
+    location = /v1/search {
+      proxy_pass http://ui/v1/search;
+      proxy_set_header Host $http_host;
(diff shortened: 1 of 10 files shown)


Dieken:
The default `cache::interval` in config file is much longer than the default value in code,  because I feel it's fine due to the event notification mechanism,  any docker push will trigger cache refresh.

Dieken:
Boring license form....

I seriously claim here that you can use my patch in any form in any time and I drop the right to sue you for this patch.

reasonerjt:
@Dieken Thanks for the PR but I don't think we can have it on master.

PR #720, opened 2016-08-25
add repo into DB.

wy65701436:
1, Add one new table to store repo that gets for the web hook of docker registry.
2, Run at the each UI image boot.
3, Admin can call it via RestAPI.

ToDo: Add dao_test.
ToDo: Add new PR for migration.


dao/repository.go
```diff
@@ -0,0 +1,86 @@
...
+	"fmt"
+
+	"github.com/astaxie/beego/orm"
+	"github.com/vmware/harbor/models"
+)
+
+// AddRepository adds a repo to the database.
+func AddRepository(repo models.RepoRecord) error {
+	o := GetOrmer()
+	sql := "insert into repository (owner_id, project_id, name, description, pull_count, star_count, creation_time, update_time) " +
```
reasonerjt: can we just use owner_id and project_id here so we can use orm to ad the record?

api/utils.go
```diff
@@ -207,6 +210,116 @@ func postReplicationAction(policyID int64, acton string) error {
 	return fmt.Errorf("%d %s", resp.StatusCode, string(b))
 }
 
+// SyncRegistry syncs the repositories of registry with database.
+func SyncRegistry() error {
+
+	rc, err := initRegistryClient()
```
reasonerjt: could you add more logs here, such as "start syncing repositories from registry to DB"and "adding repository xxxx, deleting repositories xxxx"

Just for debug and when the bootstrap of UI is slow we can see the progress.

ui/router.go
```diff
@@ -65,6 +65,7 @@ func initRouters() {
 	beego.Router("/api/projects/:id([0-9]+)/logs/filter", &api.ProjectAPI{}, "post:FilterAccessLog")
 	beego.Router("/api/users/?:id", &api.UserAPI{})
 	beego.Router("/api/users/:id([0-9]+)/password", &api.UserAPI{}, "put:ChangePassword")
+	beego.Router("/api/registry/sync", &api.RegistryAPI{}, "put:Sync")
```
reasonerjt: Could you update the url to /api/internal/syncregistry  and let the action be "POST"?

such that we may apply some global policy to /api/internal/***

reasonerjt:
Sorry but the migration should be to 0.4.0 not 0.3.0

And is it possible to move the commits of migration to another PR?

PR #729, opened 2016-08-29
project api test

zgdxiaoxiao:


yhua123:
**1**
in previous .travis yaml golint and go vet not include api path, need re test this part.

- go list ./... | grep -v -E 'vendor|tests|**api**' | xargs -L1 fgt golint
- go list ./... | grep -v -E 'vendor|tests|**api**' | xargs -L1 go vet

**2**
squash commit

**3**
don't change code under apilib
 tests/apitests/apilib/search_project.go 
these code generate from swagger.yaml

**4**
request and authRequest better using variable-length parameters, ref: http://www.ituring.com.cn/article/118545

....

PR #730, opened 2016-08-29
update ldap configuration

wemeya:
add additional configuration for ldap auth.


reasonerjt:
Is it possible to test it? :)

wemeya:
@reasonerjt 
possible, I am thinking about the test solution.

toughIQ:
@wemeya 
I tried your code against my company AD and login to Harbor worked.
But fetching the correct email and full name from AD failed. I got 

__Email: *myusername*@placeholder.com__  
__Full Name: *myusername*__ 

Maybe add params to set the fields for mail an full name (eg. cn, displayName, sAMAccountName,...)

wemeya:
@toughIQ 
I know what you mean. Since the attribute to distinguish a user can be uid or cn or mail or email, which is uncertain, now we only fetch the attribute you set and set it as the username. We will discuss about this and improve. If you have a good idea, welcome!

PR #746, opened 2016-09-01
update api test and swagger.yaml

zgdxiaoxiao:
1.Fix bug of project and user type  in swagger and test struct.
2.Add member get API test.
3.handle some conflict in statistics_test, log_test.go and harborapi_test.go


yhua123:
1 better add the projectreq struct to swagger.yaml
2 why change the user.go for role, better using the default struct created by swagger.yaml
thanks

PR #754, opened 2016-09-03
add support for reading privateKey path from env

feilengcui008:
when deploy harbor standalone but not by docker-compose, we may store the private_key.pem file in other places, but it seems that there is no env option to set that.


reasonerjt:
Hi, thanks for the contribution, please submit PR to dev branch.

feilengcui008:
@reasonerjt 
ok, moved PR to dev branch

PR #755, opened 2016-09-03
add support for reading privateKey path from env

feilengcui008:
when deploy harbor standalone but not by docker-compose, we may store the private_key.pem file in other places, but it seems that there is no env option to set that.


reasonerjt:
Thanks for the PR, but we may not merge it into master as we only support using docker-compose to deploy Harbor for now.

PR #769, opened 2016-09-06
Update for lazy loading of repo manifest.

wknet123:
Update for lazy loading of repo manifest.

### DIFF
--- a/static/resources/js/components/repository/popup-details.directive.js
+++ b/static/resources/js/components/repository/popup-details.directive.js
@@ -20,31 +20,15 @@
     .module('harbor.repository')
     .directive('popupDetails', popupDetails);
   
-  PopupDetailsController.$inject = ['ListManifestService', '$filter', 'dateLFilter'];
+  PopupDetailsController.$inject = [];
   
-  function PopupDetailsController(ListManifestService, $filter, dateLFilter) {
-    var vm = this;
-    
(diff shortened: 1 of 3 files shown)


static/resources/js/components/repository/popup-details.directive.js
```diff
@@ -20,31 +20,15 @@
...
-    
-    function retrieve() {
-      ListManifestService(vm.repoName, vm.tag)
-        .success(getManifestSuccess)
-        .error(getManifestFailed);
-    }
-    
-    function getManifestSuccess(data, status) {
-	  console.log('Successful get manifest:' + data);
-	  vm.manifest = angular.fromJson(data.manifest.history[0].v1Compatibility);
```
quickwind: Is this correct? I tried it but it failed, I changed this to simply:

vm.manifest = data.responseJSON;

And then later on refer to its fields like vm.manifest.Id, vm.manifest.OS, and vm.manifest["Docker Version"] etc.

quickwind:
Another comment: when the UI is left for a while and the session expires, clicking the details button doesn't show any error message or redirect to login page, just no effect, it appears strange to user, don't know what happens, unless open debugging window to see the error in console...

wknet123:
I made a workaround with #773 to handle the error you mentioned above.

PR #776, opened 2016-09-08
updateREADME

hainingzhang:


tianweidut:
two commits are same, can use `git rebase`

PR #781, opened 2016-09-09
Expire time of cached token

ywk253100:
1. record the local time when token is issued to avoid the mismatch of time for client and registry 2.recover the env variable


reasonerjt:
It's too tricky here, is it necessary to cache the token?  Let's discuss it tomorrow.

PR #788, opened 2016-09-12
Support define the log level in harbor.cfg

xiaoping378:
provide a way to define the log level of all container.


reasonerjt:
It does  not seem correct to try to configure log level via on configuration parm as different component may have different log level settings.

PR #791, opened 2016-09-12
bug fix: delete multiple tags should not return error even though they reference the same image

ywk253100:


api/repository.go
```diff
@@ -165,11 +165,13 @@ func (ra *RepositoryAPI) Delete() {
 	for _, t := range tags {
 		if err := rc.DeleteTag(t); err != nil {
 			if regErr, ok := err.(*registry_error.Error); ok {
-				ra.CustomAbort(regErr.StatusCode, regErr.Detail)
+				if regErr.StatusCode != http.StatusNotFound {
+					ra.CustomAbort(regErr.StatusCode, regErr.Detail)
+				}
+			} else {
+				log.Errorf("error occurred while deleting tags of %s: %v", repoName, err)
```
reasonerjt: also add "t" to the message?

PR #803, opened 2016-09-14
bug fix: handle error when checking the existence of repos whose project does not exist in Harbor

ywk253100:


reasonerjt:
Thanks

PR #804, opened 2016-09-14
fix #801, add permission check before filtering access log of a project.

### LINKED ISSUE #801: Access log retrieval API is bypassing the role-base control.

reasonerjt:


ywk253100:
LGTM

PR #808, opened 2016-09-18
Add sqlite support

ywk253100:


reasonerjt:
is it possible to use one sql script to both mysql and sqllite?

ywk253100:
The keywords for auto increment in MySQL and SQLite are different. So it could be difficult to share the same sql script.

reasonerjt:
Closing it for now, just to avoid merge my mistake.

reasonerjt:
Too busy this week, let's go through this after Oct. 10 :)

PR #829, opened 2016-09-22
Update swagger.yml

wemeya:


docs/swagger.yaml
```diff
@@ -611,6 +611,12 @@ paths:
           format: int
           required: true
           description: Registered user ID
+        - name: has_admin_role
```
yhua123: better change name to admin_role_flag

Issue #839, opened 2016-09-26
UI repo search doesn't work well


wemeya:
I pushed a new image, then refresh the UI and search, but it can't find. Also, it can't find the older images. After I refreshed a few more times, it works well.

![image](https://cloud.githubusercontent.com/assets/12540577/18828872/62ce4f98-840c-11e6-8667-d33b3c33253c.png)

![image](https://cloud.githubusercontent.com/assets/12540577/18828890/7331d27e-840c-11e6-8e8e-860449f795f7.png)

reasonerjt:
Seems it's because adding data and refresh cache are in two go routines so sometimes refresh cache doesnt' get the latest data, thanks for the finding, I'll fix it.

reasonerjt:
Fixed in #844

PR #844, opened 2016-09-28
filter access should work when user use email to docker login

reasonerjt:


reasonerjt:
This PR also contains the change to fix the push&search issue #839

PR #845, opened 2016-09-28
add search test case after push an image

wemeya:


reasonerjt:
There's a panic in the execution of TC, if it fails let it fail gracefully.

PR #847, opened 2016-09-28
Resize the defaultPageSize in api/base.go

KaleoCheng:
The defaultPageSize now could case the lack of initial replication if the project's repository is more than 10,So I resize it from 10 to 100.

### DIFF
--- a/api/base.go
+++ b/api/base.go
@@ -32,7 +32,7 @@ import (
 )
 
 const (
-	defaultPageSize int64 = 10
+	defaultPageSize int64 = 100
 	maxPageSize     int64 = 100
 )
 


reasonerjt:
Sorry I don't think this is very necessary for code in master branch, you can update it in your deployment, but probably not to everyone?

KaleoCheng:
Well,  and then how about add some tips in the documents? It  could make someone like me confused.

reasonerjt:
yep, or maybe we'll consider to provide such configuration in cfg file, I'll keep this PR open until we provide such information in doc or conf.

reasonerjt:
@KaleoCheng Sorry, could you elaborate in what scenario are you seeing this issue?  Are you feeling inconvenient using UI? or writing your own script?

KaleoCheng:
Yeah, You could use the following steps to reproduce it:

1. create a new project;
2. push image to this project until it has more than 10 repositories;
3. make a replication for this project to another harbor;

the replication will be end when the another harbor gets 10 repositories.

reasonerjt:
Ah I see this is a bug, let me open an issue, you can use this code change as a work around.

KaleoCheng:
Fine,  then I'll close it for now.   Issue with #856

PR #860, opened 2016-10-08
minor fix log print

ScorpioCPH:
Do not print successful log if error occur.


ScorpioCPH:
sorry, should not submit to master branch, close.

PR #868, opened 2016-10-12
Update for loading projects with pagination of UI.

wknet123:
Update for loading projects with pagination of UI.

### DIFF
--- a/static/resources/js/components/details/retrieve-projects.directive.js
+++ b/static/resources/js/components/details/retrieve-projects.directive.js
@@ -35,9 +35,10 @@
     var DEFAULT_PAGE = 1;
     var DEFAULT_PAGE_SIZE = 15;
       
-    vm.page = Number(getParameterByName('page', $location.absUrl()) || DEFAULT_PAGE);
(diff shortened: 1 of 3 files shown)


reasonerjt:
@ywk253100 please review, is the pattern looking correct?

static/resources/js/components/details/retrieve-projects.directive.js
```diff
@@ -62,33 +66,49 @@
...
-        for(var i in vm.projects) {
-          var project = vm.projects[i];
-          if(project['project_id'] == getParameterByName('project_id', $location.absUrl())) {
-            vm.selectedProject = project;
-            break;
-          }
-        } 
+      
+      var link = response.headers("Link") || '';
+      var pattern = /^<\/api\/projects\?is_public=(\d+)&page\=(\d+)&page_size=(\d+)\&project_name=>; rel=\"(\w+)\"$/;
```
reasonerjt: I think you should only need to get the url and trigger a request, you don't need to care what the url of "next page" is

PR #891, opened 2016-10-17
add project role for non-sysadmin user

wemeya:


ywk253100:
You can use function listRoles in api/utils.go.

reasonerjt:
If this is an API I'd also wish to suggest adding a testcase

reasonerjt:
If you look at the code, you don't need the IsAdmin code path...

reasonerjt:
Also please add a test case for this particular API.

wemeya:
I don't think call listRole is a better way to do this, though it judged whether a user is admin. In the List method, I need to judge whether the user is admin in advance, if yes, I just set the "projectList[i].Role = models.PROJECTADMIN", if not, I just need to query db to set the value. So I don't need the judgement again. I have talked about this with WenKai, you can have a whole look at the List method.

reasonerjt:
@ywk253100 @wemeya let's accelerate this, it blocks some UI refinement work.

PR #899, opened 2016-10-19
store secretkey in data volume and remove it from harbor.cfg

reasonerjt:


ywk253100:
secrectkey should be secretkey?  Line 37 in prepare.

PR #932, opened 2016-10-25
Fixed redirection when deleted user.

wknet123:


src/ui/static/resources/js/session/session.current-user.js
```diff
@@ -32,16 +32,27 @@
...
+        currentUser.set(response.data);
+        if(location.pathname === '/') {   
           $window.location.href = '/dashboard';
         }
       }   
     }
     
     function getCurrentUserFailed(e){
       console.log('No session of current user.');
+      LogOutService()
```
reasonerjt: Is the error message correct>

PR #971, opened 2016-10-28
User can not be deleted in LDAP authentication mode.

ywk253100:


src/ui/api/user.go
```diff
@@ -241,6 +241,12 @@ func (ua *UserAPI) Delete() {
 		return
 	}
 
+	// TODO read from conifg
+	authMode := os.Getenv("AUTH_MODE")
+	if authMode == "ldap_auth" {
+		ua.CustomAbort(http.StatusMethodNotAllowed, "user can not be deleted in LDAP authentication mode")
```
reasonerjt: Let's just use 403 here, 405 should container a "Allow" header in the response.

wknet123:
Part of UI has been fixed with #974

reasonerjt:
@ywk253100, sorry, I see you removed that chunk?

ywk253100:
@reasonerjt Updated.

Issue #1012, opened 2016-11-04
Harbor and Notary


rikatz:
I've been testing some of Docker Notary, with Vmware Harbor.

As those are separate and different components that can work independently, when I try to integrate the Notary Authentication with Registry I get an unauthorized error.

The following is my 'Notary' config. Note that I've copied the root.crt certificate from the 'registry' to the Notary Server config directory, naming it as token.crt:

```
{
        "server": {
                "http_addr": ":4443",
                "tls_key_file": "./192.168.1.34.key",
                "tls_cert_file": "./192.168.1.34.crt"
        },
        "trust_service": {
          "type": "remote",
          "hostname": "notarysigner",
          "port": "7899",
          "tls_ca_file": "./ca.crt",
          "key_algorithm": "ecdsa",
          "tls_client_cert": "./192.168.1.34.crt",
          "tls_client_key": "./192.168.1.34.key"
        },
        "auth": {
          "type": "token",
          "options": {
               "realm": "https://192.168.1.34/service/token",
               "service": "token-service",
               "issuer": "registry-token-issuer",
               "rootcertbundle": "/etc/docker/notary-server/token.crt"
          }
        },
        "logging": {
                "level": "debug"
        },
        "storage": {
                "backend": "mysql",
                "db_url": "server@tcp(mysql:3306)/notaryserver?parseTime=True"
        }
}

```

As this is not a Notary bug, I'm sending you the notary config so you can try to reproduce this with Harbor.

I think notary sends a token request/validation to Harbor, and Harbor does not understand the request :)

I'm logged on Harbor (docker login my_registry) succesfully, but when I try to push a signed image to Harbor with this integrated I get an error.

If I disable the Notary Authentication, I'm able to sign and push an image. The thing is, I need to verify if a user is authorized to sign an image on that project :)

Thanks!

rikatz:
An update:

Checking for the differences between signed and unsigned images:

**With Token Login enabled:**
```
192.168.1.35 - - [04/Nov/2016:01:59:21 +0000] "GET /v2/ HTTP/1.1" 401 87 "-" "docker/1.12.3 go/go1.6.3 git-commit/6b644ec kernel/3.10.0-327.36.3.el7.x86_64 os/linux arch/amd64 UpstreamClient(Docker-Client/1.12.3 \x5C(linux\x5C))"
[...]

PR #1019, opened 2016-11-05
Improves namespace validation for Docker Notary integration

rikatz:
This commit solves the issue related on #1012 

Basically, instead of verifying if there is a '/' character on the a.Name variable, it's applied a split function on this string with the '/' character, then verifying the size of array.

Being bigger than 1, it verifies if the array is bigger than 3 (like privaterepo.com/rpkatz/debian) and getting only the 2nd item from array (rpkatz), otherwise (as being smaller than 3, and bigger than 1, like rpkatz/debian) it gets the first item of the array.

### DIFF
--- a/src/ui/service/token/authutils.go
+++ b/src/ui/service/token/authutils.go
@@ -105,8 +105,15 @@ func FilterAccess(username string, a *token.ResourceActions) {
 	//clear action list to assign to new acess element after perm check.
 	a.Actions = []string{}
 	if a.Type == "repository" {
-		if strings.Contains(a.Name, "/") { //Only check the permission when the requested image has a namespace, i.e. project
-			projectName := a.Name[0:strings.LastIndex(a.Name, "/")]
+		repoSplit := strings.Split(a.Name, "/")
+		repoLength := len(repoSplit)
+		if repoLength > 0 { //Only check the permission when the requested image has a namespace, i.e. project
+			var projectName string
+			if repoLength > 2 { //If the repo contains more than 1 separation (as privateregistry.local/library/alpine) consider the second item from array (library)
+				projectName = repoSplit[1]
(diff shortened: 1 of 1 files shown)


reasonerjt:
@rikatz thanks for the PR, this code change can be used to workaround the notary issue you are seeing but I don't think it universally correct as it disregard the hostname.

rikatz:
@reasonerjt I've opened an issue on Docker Notary Server (https://github.com/docker/notary/issues/1023) to clarify this.

I think this might be some kind of inconsistence, but anyway the usage of hostname on the scope is specified here: https://docs.docker.com/registry/spec/auth/scope/

I can't figure why this happens, and if Harbor should handle this on a more gentle way (like parsing its own 'hostname' configuration used on harbor bootstrap - harbor.cfg).

Anyway, let's wait for the answer on the related issue.

Thanks.

rikatz:
@reasonerjt So guys from notary said that this behaviour (sending the whole context) is right, as notary can sign anything (including tagged repositories), and this is a part of the image.

So I think it would be better to read the entire 'GUN' sended by notary, and check if the first field matches the configured hostname of Harbor, and if (and only if) this matches, it removes the first field and proceeds to the validation of the repo permission.

I'm busy now, but will try to fix this with the described approach, and update the PR. Do you think this is a fine approach?

Thanks!

reasonerjt:
@rikatz Thanks for the update!  This PR looks good to me. Could you 
1) Let us know the issue you opened to notary
2) submit this PR to dev branch so we can do more test on that before merging to master.

rikatz:
@reasonerjt The issue is https://github.com/vmware/harbor/issues/1012

I'm going to change the PR here to DEV branch.

reasonerjt:
Thanks!

Issue #900, opened 2016-10-19
utils/mail.go: make Identity configurable via [mail] options or use a saner default


marc0s:
Given this setup:

```
Docker version 1.12.2, build bb80604
docker-compose version 1.8.1, build 878cff1
```

```
# config/ui/app.conf
[mail]
host = my.smtp.example.org
port = 587
username = validuser@example.org
password = validpassword
from = validuser@example.org
ssl = false
```

And being `my.smtp.example.org` a working SMTP server which uses Dovecot for SASL authentication, sending of emails fails because `SASL PLAIN` is sending `Mail Config` as the `Identity` parameter and Dovecot disallows the space (`0x20`) as a valid character for it.

By reading the corresponding [RFC](https://tools.ietf.org/html/rfc4616#section-2) I'm not sure it's a harbor fault or a Dovecot one (I'd bet for Dovecot) but, anyway, it seems reasonable to use a saner default without "special" characters or better to expose such parameter as a config item.

marc0s:
As expected, this is a bug in the Dovecot server version I'm using.

Nevertheless, I find it weird to use such hardcoded `Identity`.  At least in Dovecot, it's only ignored if set to the empty string `''` (otherwise is understood as the username to use for authentication).  Even in the golang docstring in `net/smtp/auth.go` is stated that it's usually left empty:

```
    48	// The returned Auth uses the given username and password to authenticate
    49	// on TLS connections to host and act as identity. Usually identity will be
    50	// left blank to act as username.
```

wknet123:
Merged #1023.

PR #1023, opened 2016-11-07
Make mail identity configurable

fillet54:
The identity field parameter passed to smtp.SendMail is now
configurable instead of being hardcoded to 'Mail Config'.

This fixes issue #900


wknet123:
LGTM, thanks.

PR #1032, opened 2016-11-07
update configure https guide

reasonerjt:


docs/configure_https.md
```diff
@@ -26,85 +28,61 @@ Otherwise, if you use IP address to connect your registry host, CN can be anythi
...
-##Installation of Harbor
-Next, edit the file make/harbor.cfg , update the hostname and the protocol:
+Next, edit the file make/harbor.cfg , update the hostname and the protocol, and update the attributes ```ssl_cert``` and ```ssl_cert_key```:
+
 ```
   #set hostname
   hostname = reg.yourdomain.com
   #set ui_url_protocol
   ui_url_protocol = https
+  ......
```
yhua123: better add the line:
#The path of cert and key files for nginx, they are applied only the protocol is set to https

PR #1036, opened 2016-11-08
update Makefile, clear make/common/config and update .gitignore

yhua123:


.gitignore
```diff
@@ -1,11 +1,9 @@
...
-make/common/config/ui/env
-make/common/config/ui/app.conf
-make/common/config/db/env
-make/common/config/jobservice/env
-make/common/config/nginx/nginx.conf
-make/common/config/nginx/cert/*
+make/common/config/registry/*
+make/common/config/ui/*
+make/common/config/db/*
+make/common/config/jobservice/*
```
reasonerjt: is this correct? we don't have app.conf in template folder?

PR #1039, opened 2016-11-08
modify prepare for genrate all common/config files from template

yhua123:


make/prepare
```diff
@@ -262,8 +277,8 @@ if customize_crt == 'on':
     if openssl_is_installed(shell_stat):
         private_key_pem = os.path.join(config_dir, "ui", "private_key.pem")
         root_crt = os.path.join(config_dir, "registry", "root.crt")
-        crt_conf_files = [ private_key_pem, root_crt ]
-        rmdir(crt_conf_files)
+        delfile(private_key_pem)
```
reasonerjt: if customize_crt is "off" there should be default private key and root crt.

make/prepare
```diff
@@ -262,11 +277,17 @@ if customize_crt == 'on':
...
```
reasonerjt: since the key and cert files are moved to template directory, you should remove them from config directory either way.

yhua123: you mean this?
 make/common/config/ui/private_key.pem → make/common/templates/ui/private_key.pem 
make/common/config/registry/root.crt → make/common/templates/registry/root.crt 
already in last commit
https://github.com/vmware/harbor/pull/1039/files

PR #1046, opened 2016-11-09
Fixes #1002

### LINKED ISSUE #1002: certificate and key files should be put into config directory
/path/to is too ad-hoc.

Let's put it into some where in config directory such that it can be backup during upgrading or trouble shooting.

ywk253100:


tools/ova/script/config.sh
```diff
@@ -58,11 +58,10 @@ then
 	echo "ssl_cert and ssl_cert_key are set, using HTTPS protocal"
 	protocal=https
 	sed -i -r s%"#?ui_url_protocol = .*"%"ui_url_protocol = $protocal"% $cfg
-	mkdir -p /path/to
-	echo $ssl_cert > /path/to/server.crt
-	format /path/to/server.crt
-	echo $ssl_cert_key > /path/to/server.key
-	format /path/to/server.key
+	echo $ssl_cert > /data/server.crt
```
reasonerjt: I see there's a typo, protocal should be protocol.

reasonerjt:
LGTM ,thanks.

PR #1058, opened 2016-11-10
Fixes #995, fixes #1047

### LINKED ISSUE #995: Harbor appliance is always setting/preferring "IP address" as hostname in harbor.cfg
Harbor Build: 0.4.5
Steps:
1. Follow https://github.com/vmware/harbor/blob/master/docs/installation_guide_ova.md  - with DHCP
2. Boot the Harbor appliance
3. Identifier of the machine "hostname" in harbor.cfg is always set to IP address. Harbor must prefer Hostname as primary network identifier

[logs.tar.gz](https://github.com/vmware/harbor/files/565743/logs.tar.gz)
[config.tar.gz](https://github.com/vmware/harbor/files/565742/config.tar.gz)
Otherwise the https certs will have to be generated using IP address as SAN. Ideally Harbor should trust both FQDN and IP address (IP can be specified as SAN) along with FQDN in subject CN 

https://github.com/vmware/harbor/blob/master/docs/installation_guide_ova.md to use SSL Cert: & SSL Cert: Key

### LINKED ISSUE #1047: It's not allowed to switch auth mode
We need to add that into the description of attribute when importing the .ova

ywk253100:


tools/ova/script/config.sh
```diff
@@ -37,13 +36,18 @@ function format {
 }
 
 #Modify hostname
-ip=$(ip addr show eth0|grep "inet "|tr -s ' '|cut -d ' ' -f 3|cut -d '/' -f 1)
-if [ -n "$ip" ]
+hostname=$(hostname)
```
reasonerjt: you may want to use "hostname --fqdn"?

And I recall sometimes this command will generate error, if that happens we may want to fallback to using IP to setup, but explicitly log it.

ywk253100:
@reasonerjt Updated.

PR #1071, opened 2016-11-11
modify openssl ca to openssl x509 for certificate format

yhua123:


reasonerjt:
Since you use -CAcreateserial section, is it still necessary to create demoCA and create the serial file?

docs/configure_https.md
```diff
@@ -27,26 +27,20 @@ Otherwise, if you use IP address to connect your registry host, CN can be anythi
 ```
 3) Generate the certificate of your registry host:
 
-On Ubuntu, the config file of openssl locates at **/etc/ssl/openssl.cnf**. Refer to openssl document for more information. The default CA directory of openssl is called demoCA. Let's create necessary directories and files:
+On Ubuntu, the config file of openssl locates at **/etc/ssl/openssl.cnf**. Refer to openssl document for more information.
```
reasonerjt: If we are not asking user to view or update the openssl.cnf, we can probably remove this chunk.

PR #1102, opened 2016-11-16
debug travis

reasonerjt:
@ywk253100 this seems may workaround the issue in travis, please consider apply this fix if the problem remains.


ywk253100:
Travis comes back, so will close it.

PR #1118, opened 2016-11-18
OVA script update

ywk253100:
1.copy ca to download directory if nedded 
2.add IP to san 
3.ignore harbor admin password and database password in subsequent boot


reasonerjt:
LGTM

PR #1133, opened 2016-11-21
LVM guide

ywk253100:


reasonerjt:
1)Could you investigate if it's feasible to use lvextend to make the process and commands easier.
2)Please revise the doc to make it more task-centric, to fit into a detailed user scenario of Harbor.

thanks.

PR #1186, opened 2016-11-29
Fix save comment issue.

wknet123:
Fix issue about saving comment while sign up.


reasonerjt:
We're not allowing change to master, please target this to dev.

PR #1193, opened 2016-11-30
remove timeout for replicating large image

reasonerjt:
To fix the frequent timeout issue during recent testing.
Will revisit the transport initialization code to tune the different timeouts.


ywk253100:
LGTM

PR #1197, opened 2016-12-01
remove rsyslogd.pid before start to allow container restarts

dejwsz:
This change mitigate problems with container restarts (stop, start) or automatic restart after host machine restart. Rsyslogd strictly checks existence of its pid file and won't start if such one exists.


reasonerjt:
@dejwsz Thanks!

PR #1198, opened 2016-12-02
Fixed encoding semicolon in query URL.

wknet123:


reasonerjt:
let's not do simple replacement.

PR #1205, opened 2016-12-02
Remove the workaround for DNS issue and escape special characters used in sed

ywk253100:


reasonerjt:
@ywk253100 I have fixed the build issue, but "restart build" won't pick up the latest code on master, so could you reopen the PR and we can merge it if the build is green...

PR #1207, opened 2016-12-04
Move self-registrition option to authentication group in installation guide

ywk253100:


reasonerjt:
@ywk253100 see comment in #1205

PR #1229, opened 2016-12-08
Change LDAP library to go-ldap and configure a Default Timeout for LDAP

rikatz:
This PR is to update the go library used to connect to LDAP.

Also, it defines a default Dial Timeout to LDAP, so if there is some connectivity problem, it's still possible to pull Images from public repositories.


rikatz:
Opened a new PR, more descritive, with the right commits.

Issue #1228, opened 2016-12-08
LDAP Library used is deprecated


rikatz:
According to the OpenLDAP library used on Harbor project (https://github.com/mqu/openldap), it's deprecated and the maintainer is not going to continue it's development.

This brings some other non desirable behaviour, as not supporting StartTLS, non configurable dial timeout (this brings a lot of problems when you try to pull an image from an anonymous repository) and limits a future work to support LDAP Group (and a lot of other problems).

Fortunatelly, there is another library (that is much more updated) here: https://github.com/go-ldap/ldap

So, this issue is to register an attempt to migrate from the old library to a new one. I'm going to open a PR submitting changes to this, and then some other commits to solve the "TODO" marked on the code.

Any suggestion? :smile:

yhua123:
@rikatz 
Thank you for your recommend~ 
We already start working on this. Will update LDAP library soon.
#1136

rikatz:
Solved in the PR, closing the issue.

PR #1231, opened 2016-12-08
Change OpenLDAP library

rikatz:
As described on #1228 the LDAP library used on Harbor is deprecated.

Changing this to the new go-ldap library can help on future issues and features, as group authentication, TLS comunication.

Also, this library implements the Timeout handling, so when the LDAP is offline, users still can pull images from public repositories.


reasonerjt:
@rikatz thanks  for the PR, actually we are thinking about the same thing, allow us some time for review and legal clearance.

rikatz:
@reasonerjt @yhua123 I'm going to open a new PR, pointing to my DEV Branch. I think this is right. Sorry for the mess guys

PR #1261, opened 2016-12-21
Added escape for angle brackets in email_from

guiguan:
This solved the "[ERROR] [password.go:83]: Send email failed: 555 5.5.2 Syntax error. b192sm9510409pga.30 - gsmtp" error.


wknet123:
I think these changes should be applied in codes not for configuration.

weaktyper:
Any progress on this issue?

weaktyper:
I would suspect the problem is on this line 
`https://github.com/vmware/harbor/blob/99e1b5cd85ad9aa679a095d351ea4b5d000dc20e/src/common/utils/mail.go#L77`
I don't know `golang` so I could be wrong. 😖

Issue #1145, opened 2016-11-22
[issue] No "tdnf" Will Report Error when Running "install.sh"


yhua123:
...

yhua123:
@lwh

reasonerjt:
Let's add more details to this issue to describe the problem.

lwh:
If you automate installing harbor, (i.e. in chef or puppet) when you run install.sh the tput command fails because the shell isn't interactive.  If it could detect and disable the tput calls for non-interactive shells it would help.

yhua123:
thanks a lot ~ I will work on this issue next week~

yhua123:
@lwh
please try this, thanks~
we check `tput` exist or not.

`#/bin/bash

#docker version: 1.11.2 
#docker-compose version: 1.7.1 
#Harbor version: 0.4.5+ 
set +e
set -o noglob

#
# Set Colors
#

if type "tput" > /dev/null; then
	bold=$(tput bold)
	underline=$(tput sgr 0 1)
	reset=$(tput sgr0)
	red=$(tput setaf 1)
	green=$(tput setaf 76)
	white=$(tput setaf 7)
	tan=$(tput setaf 202)
	blue=$(tput setaf 25)
else
  	bold= 
  	underline=
  	reset=
  	red=
  	green=
  	white=
  	tan=
  	blue=
fi

#
# Headers and Logging
#

underline() { printf "${underline}${bold}%s${reset}\n" "$@"
}
h1() { printf "\n${underline}${bold}${blue}%s${reset}\n" "$@"
}
h2() { printf "\n${underline}${bold}${white}%s${reset}\n" "$@"
}
debug() { printf "${white}%s${reset}\n" "$@"
}
info() { printf "${white}➜ %s${reset}\n" "$@"
}
success() { printf "${green}✔ %s${reset}\n" "$@"
}
error() { printf "${red}✖ %s${reset}\n" "$@"
}
warn() { printf "${tan}➜ %s${reset}\n" "$@"
}
bold() { printf "${bold}%s${reset}\n" "$@"
}
note() { printf "\n${underline}${bold}${blue}Note:${reset} ${blue}%s${reset}\n" "$@"
}

set -e

usage=$'Checking environment for harbor build and install. Include golang, docker and docker-compose.'

while [ $# -gt 0 ]; do
        case $1 in
            --help)
            note "$usage"
            exit 0;;
            *)
            note "$usage"
            exit 1;;
        esac
        shift || true
done

[...]

PR #1263, opened 2016-12-22
modify shell 

wemeya:


yhua123:
Waiting more response from #1145, close this first.

PR #1264, opened 2016-12-22
update beego to 1.7.0

wemeya:


yhua123:
you don't need submit any PR on vmware/dev for verify new beego lib.
please test on your or my private repo (github.com/yhua123/harbor)
thanks

PR #1285, opened 2017-01-03
add registry python api and cli tool

feilengcui008:
this pr add a little python api class for registry and a little tool for repo, tag, manifest operations.
more details, please see contrib/registryapi/README.md

### DIFF
--- a/contrib/registryapi/registry.py
+++ b/contrib/registryapi/registry.py
@@ -1,0 +1,167 @@
+#!/usr/bin/env python
+# -*- coding:utf-8 -*-
+# bug-report: feilengcui008@gmail.com
+
+""" api for docker registry """
+
+import urllib2
+import urllib
+import json
+import base64
+
+
+class RegistryException(Exception):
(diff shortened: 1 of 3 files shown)


contrib/registryapi/registry.py
```diff
@@ -0,0 +1,167 @@
...
+        except Exception as e:
+            return False
+        return True
+
+    def getManifestWithConf(self, repository, reference="latest", v1=False):
+        """ get manifest for tag or digest """
+        manifest = self.getManifest(repository, reference, v1)
+        if manifest is None:
+            raise RegistryException("manifest for %s %s not exist" % (repository, reference))
+        config_digest = manifest["config"]["digest"]
```
ywk253100: If v1 is true, the config part of manifest does not need any more, right?

feilengcui008: yes, getManifestWithConf should only support schema v2, I'll add a patch

Issue #1290, opened 2017-01-05
[New Request] Customize the Log level


gunboe:
Would be good customize the Log Level in `harbor.cfg` for all services (mainly registry, ui, jobservice). Like this:
`log_level = [ fatal | error | warn | info | debug | trace ]  # Default: debug`

gunboe:
Created a PR #1291

ywk253100:
Thank you for your contribution. We are doing some modifications to provide a more friendly user experience to customize some configurations of Harbor, the log level is included. So your PR may not be merged. Thanks anyway.

ywk253100:
Duplicate with #2354 .

PR #1291, opened 2017-01-05
Adjustments for cusomize the Log Level

gunboe:
I've decide to try make that request #1290 . So I hope you inspect my suggestions to allow changes in the Log level. Thanks.


reasonerjt:
I'm not sure if we want to merge it right away, user may set different log level for different components.

gunboe:
You're rigth (@reasonerjt)... but that's a global setting, and users can change common config files showed in installation and configuration guide. Or We can make 3 or 5 new params in harbor.cfg for each service...

reasonerjt:
close as won't fix.

PR #1298, opened 2017-01-09
Update testcase

sigsbee:
Update testcase


sigsbee:
will rewrite test case

PR #1299, opened 2017-01-09
add ldap ping feature

yhua123:


reasonerjt:
Would be better if we can also verify other options in this api?

yhua123:
Thanks for your suggestion.
For my design, this feature take minimum options, focus on verify LDAP service on server side.
Other parameters configured by customer, plan implement in new feature-search.
Reason: 1) goldap.search need these parameters
                  2) reuse code for verify

reasonerjt:
@yhua123, I think that makes sense, but as per our experience, users often make mistake setting up the scope and other parms, IMO the "test connection" should verify as much as possible to help user avoid such cases.

PR #1300, opened 2017-01-09
Happy new year

tronicum:
you might do a 
s/Copyright (c) 2016 VMware, Inc. All Rights Reserved./opyright (c) 2017 VMware, Inc. All Rights Reserved./


ywk253100:
@tronicum The PR is used for?

tronicum:
Set copyright to include 2017

reasonerjt:
@tronicum Thanks for the PR
However, I assume our legal team will provide new copy right headers for source files, so let me close it for now...

tronicum:
i would advice tom remove the year from the file header something like "since 201x". 

the field i changed is the single file that is user facing in the UI.

tronicum:
going to 2017 anytime soon?

reasonerjt:
Not very soon, probably end of March

PR #1349, opened 2017-02-10
Update Makefile and swagger

ywk253100:


src/jobservice/replication/transfer.go
```diff
@@ -460,6 +462,14 @@ func (m *ManifestPusher) enter() (string, error) {
 func newRepositoryClient(endpoint string, insecure bool, credential auth.Credential, repository, scopeType, scopeName string,
 	scopeActions ...string) (*registry.Repository, error) {
 
+	domain, err := config.DomainName()
```
reasonerjt: Didn't we call it hostname in ui process?

src/ui/api/repository.go
```diff
@@ -442,6 +443,15 @@ func newRepositoryClient(endpoint string, insecure bool, username, password, rep
 	scopeActions ...string) (*registry.Repository, error) {
 
 	credential := auth.NewBasicAuthCredential(username, password)
+
+	domain, err := config.DomainName()
+	if err != nil {
+		return nil, err
+	}
+	if err := os.Setenv("DOMAIN_NAME", domain); err != nil {
```
reasonerjt: Why set env? using env as global var?

src/jobservice/replication/transfer.go
ywk253100: I think "hostname" is inaccurate as the value is the external URL which Harbor can be reached, the hostname is not the correct value if Harbor is deployed behind a proxy. So I change it to domain name instead.

src/common/utils/registry/auth/tokenauthorizer.go
```diff
@@ -133,19 +133,24 @@ func (t *tokenAuthorizer) updateCachedToken(token string, expiresIn int) {
...
-	credential Credential
+	client               *http.Client
+	credential           Credential
+	tokenServiceEndpoint string
 }
 
 // NewStandardTokenAuthorizer returns a standard token authorizer. The authorizer will request a token
 // from token server and add it to the origin request
-func NewStandardTokenAuthorizer(credential Credential, insecure bool, scopeType, scopeName string, scopeActions ...string) Authorizer {
+// If tokenServiceURL is set, the token request will be sent to it instead of the server get from authorizer
```
reasonerjt: in the parm it's tokenServiceEndpoint

src/jobservice/replication/transfer.go
```diff
@@ -457,10 +459,20 @@ func (m *ManifestPusher) enter() (string, error) {
...
+func newRepositoryClient(endpoint string, insecure bool, credential auth.Credential,
+	tokenServiceEndpoint, repository, scopeType, scopeName string,
 	scopeActions ...string) (*registry.Repository, error) {
 
-	authorizer := auth.NewStandardTokenAuthorizer(credential, insecure, scopeType, scopeName, scopeActions...)
+	domain, err := config.ExtEndpoint()
+	if err != nil {
+		return nil, err
+	}
+	if err := os.Setenv("DOMAIN_NAME", domain); err != nil {
```
reasonerjt: Why still Setenv?

PR #1361, opened 2017-02-15
Implement user sign-in component with clarity and Angular2

steven-zou:
Implement the sign-in component


steven-zou:
Discussed with @wknet123 , everything is ok. Merged into clarity branch!

PR #1387, opened 2017-02-16
Add global message component to UI.

wknet123:


steven-zou:
Should support different message type later. At present, I think it's ok

PR #1434, opened 2017-02-23
Support changing all configurations through API

ywk253100:


reasonerjt:
why coverage falls? :)

ywk253100:
The logic of a function is simplified.

reasonerjt:
Ah I see, fine.

PR #1435, opened 2017-02-23
Using different secret to mark himself when communicates with other components

ywk253100:


reasonerjt:
Does it cover all scenarios where "secret" is involved?

ywk253100:
All covered.

reasonerjt:
Thanks!

PR #1462, opened 2017-02-26
Token service refactory

reasonerjt:
Due to the implementation in Docker client, seems it use the token for registry to access notary, I had use same token creator for both notary and registry, I will check the code of docker and discuss with them later.

### DIFF
--- a/src/ui/service/token/creator.go
+++ b/src/ui/service/token/creator.go
@@ -1,0 +1,238 @@
+/*
+   Copyright (c) 2016 VMware, Inc. All Rights Reserved.
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package token
+
+import (
+	"fmt"
+	"github.com/docker/distribution/registry/auth/token"
+	"github.com/vmware/harbor/src/common/dao"
+	"github.com/vmware/harbor/src/common/utils/log"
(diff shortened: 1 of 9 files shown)


src/ui/service/token/creator.go
```diff
@@ -0,0 +1,237 @@
...
+	endpoint string
+}
+
+func (e endpointParser) parse(s string) (*image, error) {
+	repo := strings.SplitN(s, "/", 2)
+	if len(repo) < 2 {
+		return nil, fmt.Errorf("Unable to parse image from string: %s", s)
+	}
+	//Workaround, need to use endpoint Parser to handle both cases.
+	if strings.ContainsRune(repo[0], '.') {
```
ywk253100: It seems that '.' is a valid character in repo name.

src/ui/service/token/creator.go
```diff
@@ -0,0 +1,237 @@
...
+	if len(scopeParm) > 0 {
+		scopes = strings.Split(r.URL.Query()["scope"][0], " ")
+	}
+	log.Debugf("scopes: %v", scopes)
+	for _, v := range g.validators {
+		user, err = v.validate(r)
+		if user != nil {
+			break
+		}
+		if err != nil {
```
ywk253100: Why put the error handling logic after the judgement statement?

src/ui/service/token/creator.go
```diff
@@ -0,0 +1,237 @@
...
+		if len(scopes) == 0 {
+			return nil, &unauthorizedError{}
+		}
+		user = &userInfo{}
+	}
+	access := GetResourceActions(scopes)
+	for _, a := range access {
+		f, ok := g.filterMap[a.Type]
+		if !ok {
+			log.Warningf("No filter found for access type: %s, skip.", a.Type)
```
ywk253100: Does there need a continue?

reasonerjt: yes,thanks

src/ui/service/token/creator.go
reasonerjt: This is the first chunk, the parser makes best effort to guess if it starts with a host name.  I think docker has the same logic?

reasonerjt: I'll fix it so that if endpoints mismatch it falls back to basic parser.

PR #1477, opened 2017-02-27
fix the 'v2' URL conflict issue, and remove the workaround

reasonerjt:
User will have to set the env var like this:

DOCKER_CONTENT_TRUST_SERVER=https://10.117.4.142/notary


src/ui/service/token/creator.go
```diff
@@ -63,9 +63,7 @@ func InitCreators() {
 		filterMap: map[string]accessFilter{
 			"repository": &repositoryFilter{
 				//Workaround, had to use same service for both notary and registry
```
ywk253100: The comment can be removed?

reasonerjt: Thanks!

PR #1490, opened 2017-03-01
Read some configs from env when start up

### LINKED ISSUE #1440: Provide a proper way for user to change configurations which are not exposed on UI

ywk253100:
Fix #1440 .

### DIFF
--- a/src/jobservice/config/config.go
+++ b/src/jobservice/config/config.go
@@ -25,6 +25,7 @@ import (
 
 const (
 	defaultKeyPath string = "/etc/jobservice/key"
+	defaultLogDir  string = "/var/log/jobs"
 )
 
 var (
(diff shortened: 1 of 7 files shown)


src/jobservice/config/config.go
```diff
@@ -120,11 +121,11 @@ func LocalRegURL() (string, error) {
...
 func LogDir() (string, error) {
-	cfg, err := mg.Get()
-	if err != nil {
-		return "", err
+	dir := os.Getenv("LOG_DIR")
+	if len(dir) == 0 {
+		dir = defaultLogDir
 	}
-	return cfg[comcfg.JobLogDir].(string), nil
+	return dir, nil
```
reasonerjt: does it need to return the second err? seems it's always nil.

PR #1512, opened 2017-03-05
Support email ping

ywk253100:


src/ui/api/email.go
```diff
@@ -0,0 +1,117 @@
...
+	m := map[string]string{}
+	e.DecodeJSONReq(&m)
+
+	settings, err := config.Email()
+	if err != nil {
+		e.CustomAbort(http.StatusInternalServerError,
+			http.StatusText(http.StatusInternalServerError))
+	}
+
+	host, ok := m[comcfg.EmailHost]
```
reasonerjt: is it possible to reuse the code in admin server to validate the email setting?

src/common/utils/email/mail.go
```diff
@@ -117,24 +118,77 @@ func sendMailWithTLS(m Mail, auth smtp.Auth, content []byte) error {
...
-		panic(err)
+		return
 	}
+	defer client.Close()
 
-	var useTLS = false
-	if config["ssl"] != "" && strings.ToLower(config["ssl"]) == "true" {
-		useTLS = true
+	//swith to SSL/TLS
+	if !tls && starttls {
```
reasonerjt: seems the parm "starttls" is useless, you can alway check and fall back if it doesn't has the "STARTTLS" Extension?

src/common/utils/email/mail.go
```diff
@@ -117,24 +118,77 @@ func sendMailWithTLS(m Mail, auth smtp.Auth, content []byte) error {
...
-	config, err := beego.AppConfig.GetSection("mail")
+// Ping tests the connection and authentication with email server
+// If tls is true, a secure connection is established, or the
+// connection is insecure, and if starttls is true, Ping trys to
+// upgrate the insecure connection to a secure one if email server
+// supports it.
+// Ping doesn't verify the server's certificate and hostname when
+// needed if the parameter insecure is ture
+func Ping(addr, identity, username, password string,
+	timeout int, tls, starttls, insecure bool) (err error) {
```
reasonerjt: Maybe trivial.
I'm not sure if it's my personal taste....
but I would prefer the form:
```
 func xxx() error{
    err := xxxxx
    return err
}
```

src/ui/api/email.go
```diff
@@ -0,0 +1,117 @@
...
+
+	if !isSysAdmin {
+		e.CustomAbort(http.StatusForbidden, http.StatusText(http.StatusForbidden))
+	}
+}
+
+// Ping tests connection and authentication with email server
+func (e *EmailAPI) Ping() {
+	m := map[string]string{}
+	e.DecodeJSONReq(&m)
```
reasonerjt: So the request may look like: 
```
{
"host":"xxx", 
"port":xxx,
....}
```
Is this consistent with the ldap ping api?
I thought the request for ldap ping was like:
{"setting":{"host":"xxx" ....}}
could you double check?

reasonerjt:
Would be good if you can add the swagger change in this PR.

src/ui/api/email.go
ywk253100: It is consistent with ldap ping api.

src/ui/api/email.go
ywk253100: The validation is done in UI API now.

PR #1536, opened 2017-03-08
Refactor get repo API

ywk253100:


src/ui/api/repository_test.go
```diff
@@ -20,7 +20,7 @@ func TestGetRepos(t *testing.T) {
 	fmt.Println("Testing Repos Get API")
 	//-------------------case 1 : response code = 200------------------------//
 	fmt.Println("case 1 : response code = 200")
-	httpStatusCode, err = apiTest.GetRepos(*admin, projectID)
```
reasonerjt: Let's also verify the response body both when detail is "true" or "0".

ywk253100: Will open another PR to cover this.

PR #1541, opened 2017-03-08
to support nortary

wy65701436:
Will create an new PR for install.sh script to support Nortary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@ COMPILETAG=compile_normal
 REGISTRYSERVER=
 REGISTRYPROJECTNAME=vmware
 DEVFLAG=true
-NORTARYFLAG=false
+NOTARYFLAG=false
 
 #clarity parameters
 CLARITYIMAGE=danieljt/harbor-clarity-base[:tag]
@@ -232,7 +232,7 @@ compile:check_environment $(COMPILETAG)
 
 prepare: 
 	@echo "preparing..."
-	@if [ "$(NORTARYFLAG)" = "true" ] ; then \
+	@if [ "$(NOTARYFLAG)" = "true" ] ; then \
 		$(MAKEPATH)/$(PREPARECMD) --conf $(CONFIGPATH)/$(CONFIGFILE) --with-notary; \
 	else \
 		$(MAKEPATH)/$(PREPARECMD) --conf $(CONFIGPATH)/$(CONFIGFILE) ; \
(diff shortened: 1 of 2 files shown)


reasonerjt:
Pls add the change of install script to the PR and squash the commits.

make/install.sh
```diff
@@ -49,14 +49,20 @@ note() { printf "\n${underline}${bold}${blue}Note:${reset} ${blue}%s${reset}\n"
 set -e
 set +o noglob
 
-usage=$'Please set hostname and other necessary attributes in harbor.cfg first. DO NOT use localhost or 127.0.0.1 for hostname, because Harbor needs to be accessed by external clients.'
+usage=$'Please set hostname and other necessary attributes in harbor.cfg first. DO NOT use localhost or 127.0.0.1 for hostname, because Harbor needs to be accessed by external clients.
+Please set --with-notary if needs enable Notary in Harbor, and set ui_url_protocol in harbor.cfg to https bacause notary must run under https.'
```
reasonerjt: set ui_url_protocol is a bit misleading, because when it's set to https user should also need to configure the cert and key.

make/install.sh
```diff
@@ -74,6 +80,20 @@ then
 	exit 1
 fi
 
+# The ui_url_protocol in harbor.cfg has not been modified to https in notary mode
```
reasonerjt: this is covered by prepare.

Makefile
```diff
@@ -313,6 +313,16 @@ package_offline: compile build modify_composefile
 		nginx:1.11.5 registry:2.5.1 photon:1.0 ; \
 	fi
 	
+	@$(TARCMD) -zcvf harbor-offline-installer-$(VERSIONTAG).tgz \
+	          --exclude=$(HARBORPKG)/common/db --exclude=$(HARBORPKG)/common/config\
+			  --exclude=$(HARBORPKG)/common/log --exclude=$(HARBORPKG)/ubuntu \
```
ywk253100: We don't have directory ubuntu anymore. Besides, maybe the include is more convenient and strict？

reasonerjt: +1 let's use include.

wy65701436: sure, will update to include.

make/install.sh
wy65701436: will remove this from install.sh to avoid duplicate checking.

reasonerjt:
LGTM @ywk253100 please double check the tar commands if it's ok, update the review to "approve"

PR #1544, opened 2017-03-08
notary containers should use syslog as log driver

reasonerjt:


reasonerjt:
I think verifying this would be low priority, but it should be easy, just check /var/log/harbor for the log file of containers.

PR #1554, opened 2017-03-09
bug fix in search API: do not return duplicate projects when user does not login

ywk253100:


src/common/dao/project.go
```diff
@@ -171,7 +171,7 @@ func SearchProjects(userID int) ([]models.Project, error) {
 
 	sql :=
 		`select distinct p.project_id, p.name, p.public, 
-			p.owner_id, p.creation_time, p.update_time, pm.role role
```
reasonerjt: no impact for other UI/API?

src/ui/api/search.go
```diff
@@ -77,26 +77,36 @@ func (s *SearchAPI) Get() {
...
 		}
-		if match {
-			if userID != dao.NonExistUserID {
-				if isSysAdmin {
-					p.Role = models.PROJECTADMIN
-				}
-				if p.Role == models.PROJECTADMIN {
-					p.Togglable = true
-				}
+		if !match {
```
reasonerjt: that variable "match" is not needed

src/common/dao/project.go
reasonerjt: Will it impact other API?

src/ui/api/search.go
reasonerjt: Seems the variable "match" is not needed.

src/ui/api/search.go
```diff
@@ -77,26 +77,36 @@ func (s *SearchAPI) Get() {
...
+			continue
+		}
+
+		if userID != dao.NonExistUserID {
+			if isSysAdmin {
+				p.Role = models.PROJECTADMIN
 			}
 
-			repos, err := dao.GetRepositoryByProjectName(p.Name)
+			roles, err := dao.GetUserProjectRoles(userID, p.ProjectID)
```
reasonerjt: if isSysAdmin == True this is not needed?

src/common/dao/project.go
ywk253100: No

reasonerjt:
Please add more test case to test the search API.

PR #1582, opened 2017-03-13
Updates for verified tags deletion.

wknet123:
Remove mock data.
Block deletion of verified tags.


reasonerjt:
I think we'd better provide the notary command to remove the signature.

PR #1584, opened 2017-03-14
fix  #1573 admin without oldpassword change user's password

### LINKED ISSUE #1573: the administrator without having to provide the user's original password can reset the user password exist a bug
When modifying user.go in  /harbor/src/ui/api to realize the function that the administrator without having to provide the user's original password can reset the user password . After debugging found there is a bug in harbor/src/common/dao user. go , we need to change the func ChangeUserPassword
image

bianhua2015:


reasonerjt:
Closing due to inactivity.

PR #1676, opened 2017-03-20
DELETE /api/repository returns 412 if the tag is signed in notary

reasonerjt:


reasonerjt:
Manually verified e2e scenarios.

PR #1710, opened 2017-03-22
Refactor API URL.

ywk253100:


docs/swagger.yaml
```diff
@@ -757,19 +758,42 @@ paths:
               $ref: '#/definitions/DetailedTag'
         500:
           description: Unexpected internal errors.
-  /repositories/manifests:
+    delete:
```
ywk253100: There is another API DELETE /api/repositories/${name}/tags/${tag}, so DELETE /api/repositories/${name}/tags/ may be make more sense?

docs/swagger.yaml
```diff
@@ -438,6 +438,64 @@ paths:
           description: Project ID does not exist.
         500:
           description: Unexpected internal errors.
+  /projects/{project_id}/repositories:
```
ywk253100: I'll update this API.

PR #1717, opened 2017-03-22
restrict access to notary db

reasonerjt:
Had to build a new package to update a configuration in my.cnf.


reasonerjt:
The new image is rebuilt to overwrite the "skip-resolve-hostname" option of mysql

PR #1803, opened 2017-03-26
Fix markdown in docs to be github compatible.

dlinsley:
The markdown in ``user_guide.md``, ``user_guide_ova.md``, and ``configure_httpd.md`` was not compatible with Github markdown parser.

This pull request just fixes the markdown so it is parsed correctly by the Github parser.


dlinsley:
Already resolved with other updates.

PR #1849, opened 2017-03-29
Fix sending email bug

ywk253100:


src/ui/controllers/base.go
```diff
@@ -87,72 +87,80 @@ func (cc *CommonController) UserExists() {
...
+	}
 
-		message := new(bytes.Buffer)
+	queryUser := models.User{Email: email}
+	exist, err := dao.UserExists(queryUser, "email")
+	if err != nil {
+		log.Errorf("Error occurred in UserExists: %v", err)
+		cc.CustomAbort(http.StatusInternalServerError, "Internal error.")
+	}
+	if !exist {
```
reasonerjt: add a log

src/ui/controllers/base.go
```diff
@@ -87,72 +87,80 @@ func (cc *CommonController) UserExists() {
...
-		if err != nil {
-			log.Errorf("Can not load app.conf: %v", err)
-			cc.CustomAbort(http.StatusInternalServerError, "internal_error")
-		}
+	harborURL, err := config.ExtEndpoint()
+	if err != nil {
+		log.Errorf("failed to get domain name: %v", err)
+		cc.CustomAbort(http.StatusInternalServerError, http.StatusText(http.StatusInternalServerError))
+	}
+	if harborURL == "" {
```
reasonerjt: I don't think this make any sense now, let's remove this.

src/common/utils/email/mail.go
```diff
@@ -106,7 +53,11 @@ func sendMailWithTLS(m Mail, auth smtp.Auth, content []byte) error {
 		return err
 	}
 
-	_, err = w.Write(content)
+	template := "From: %s\r\nTo: %s\r\nSubject: %s\r\nMIME-version: 1.0;\r\nContent-Type: text/html; charset=\"UTF-8\"\r\n\n%s\r\n"
```
reasonerjt: why is \r\n needed?

and btw, I think you can use `` to quote a string so "\n" is not needed.

src/common/utils/email/mail.go
```diff
@@ -147,19 +109,17 @@ func Ping(addr, identity, username, password string,
...
 	log.Debugf("creating SMTP client for %s ...", host)
 	client, err := smtp.NewClient(conn, host)
 	if err != nil {
-		return
+		return nil, err
 	}
-	defer client.Close()
 
 	//try to swith to SSL/TLS
 	if !tls {
```
reasonerjt: else?

src/ui/controllers/base.go
```diff
@@ -87,72 +89,81 @@ func (cc *CommonController) UserExists() {
...
+	if err != nil {
+		log.Errorf("Send email failed: %v", err)
+		cc.CustomAbort(http.StatusInternalServerError, "send_email_failed")
 	}
 
+	user := models.User{ResetUUID: uuid, Email: email}
+	if err = dao.UpdateUserResetUUID(user); err != nil {
+		log.Errorf("failed to update user reset UUID: %v", err)
+		cc.CustomAbort(http.StatusInternalServerError, http.StatusText(http.StatusInternalServerError))
+	}
```
reasonerjt: Shall we update the reset id first and send email if success?

src/common/utils/email/mail.go
ywk253100: CRLF is needed to terminate the line and new line in `` only contains LF in linux.

src/ui/controllers/base.go
ywk253100: OK.

src/common/utils/email/mail.go
reasonerjt: ok thanks.

PR #1896, opened 2017-03-31
Do not return 403 for HEAD

reasonerjt:


src/ui/api/project.go
```diff
@@ -140,14 +140,10 @@ func (p *ProjectAPI) Head() {
 		return
 	}
 
-	userID := p.ValidateUser()
+	_ = p.ValidateUser()
```
ywk253100: Should we check user first? And seems "project.Public == 1" in the if statement is not needed anymore.

PR #1909, opened 2017-04-05
ignore all the files under static

wy65701436:
fix issue https://github.com/vmware/harbor/issues/1822


wy65701436:
Pending this PR as it needs to discuss.

PR #1929, opened 2017-04-05
Modify the subject of resetting password email

ywk253100:


src/ui/controllers/base.go
```diff
@@ -158,7 +158,7 @@ func (cc *CommonController) SendEmail() {
 		60, settings.SSL,
 		false, settings.From,
 		[]string{email},
-		cc.Tr("reset_email_subject"),
+		"Reset Password",
```
reasonerjt: Let it be "Reset Harbor user password"

PR #1952, opened 2017-04-07
add a missing format string argument

maplain:


src/common/utils/notary/helper.go
```diff
@@ -84,7 +84,7 @@ func GetTargets(notaryEndpoint string, username string, fqRepo string) ([]Target
 	rootJSON := path.Join(notaryCachePath, "tuf", fqRepo, "metadata/root.json")
 	rmErr := os.Remove(rootJSON)
 	if rmErr != nil {
-		log.Warningf("Failed to clear cached root.json: %s, error: %v, when repo is removed from notary the signature status maybe incorrect")
+		log.Warningf("Failed to clear cached root.json, error: %v, when repo is removed from notary the signature status maybe incorrect", rmErr)
```
reasonerjt: Thanks for the finding!  However the first parm should better be the var "rootJSON" as it contains the repo's information.

maplain: 👌

reasonerjt:
Thanks @maplain !

PR #1955, opened 2017-04-07
fix issue 1932

wy65701436:


reasonerjt:
Could you squash the change seems this PR contains some change from another one.

PR #1975, opened 2017-04-08
Modify ping targat and put target api

ywk253100:


ywk253100:
There is a scenario that the endpoint and username are got from request body but the password is got from database by the target ID. The two URIs seem not apply to this.

PR #1980, opened 2017-04-10
add swagger validator

wy65701436:


tests/swaggervalidate.sh
```diff
@@ -0,0 +1,33 @@
...
+    STATUS=$(curl --insecure -s -o /dev/null -w '%{http_code}' $VALIDATOR)
+    if [ $STATUS -eq 200 ]; then
+		break
+    fi
+    TIMEOUT=$(($TIMEOUT - 1))
+    sleep 2
+done
+
+if [ $TIMEOUT -eq 0 ]; then
+    echo "Swagger online checke cannot reach, would not fail travis here."
```
reasonerjt: typo?

reasonerjt: typo?

tests/swaggervalidate.sh
```diff
@@ -0,0 +1,33 @@
...
+    exit 0
+fi
+
+curl -X GET $VALIDATOR | grep "{}"  > /dev/null
```
reasonerjt: why not write the json response in previous curl command and check the content of the file here?  Otherwise this curl may still timeout...

tests/swaggervalidate.sh
```diff
@@ -0,0 +1,33 @@
...
+if [ $? -eq 0 ]; then 
+	echo "Swagger yaml check success."
+else
+	echo "Swagger yaml check fail."
+	echo $(curl -X GET $VALIDATOR)
```
reasonerjt: Again, read the result of previous curl should be OK?

tests/swaggervalidate.sh
```diff
@@ -0,0 +1,33 @@
...
+
+SWAGGERFILE=https://raw.githubusercontent.com/vmware/harbor/$1/docs/swagger.yaml
+VALIDATOR=http://online.swagger.io/validator/debug?url=$SWAGGERFILE
+
+echo $SWAGGERFILE
+
+TIMEOUT=10
+while [ $TIMEOUT -gt 0 ]; do
+    STATUS=$(curl --insecure -s -o /dev/null -w '%{http_code}' $VALIDATOR)
+    if [ $STATUS -eq 200 ]; then
```
reasonerjt: curl will return non-zero if the connection is time-out.
if the status code should be 200 only, no need to check its value.
Or there maybe other non-200 status code? that indicate other errors?

PR #1982, opened 2017-04-11
update test case and user guide

sigsbee:
update test case and user guide


sigsbee:
close for separate  pr

PR #2000, opened 2017-04-12
Override page title and search placeholder when in VIC mode

steven-zou:
Override page title and search placeholder when in VIC mode

### DIFF
--- a/src/ui_ng/src/index.html
+++ b/src/ui_ng/src/index.html
@@ -3,7 +3,7 @@
 
 <head>
     <meta charset="utf-8">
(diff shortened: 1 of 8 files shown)


src/ui_ng/src/index.html
```diff
@@ -3,7 +3,7 @@
 
 <head>
     <meta charset="utf-8">
-    <title>Harbor</title>
```
reasonerjt: VMware?

wy65701436:
@ywk253100  Can you help to check the error on API test?

PR #2006, opened 2017-04-13
Swagger check

wy65701436:


reasonerjt:
Could you squash the commits?

.travis.yml
```diff
@@ -101,6 +101,7 @@ script:
   
   - docker ps
   - ./tests/notarytest.sh
+  - ./tests/swaggerchecker.sh $TRAVIS_BRANCH
```
reasonerjt: probably should use the commit id, other wise it's checking the old file on the branch in the case of PR.

PR #2010, opened 2017-04-13
replace go header

wy65701436:


wy65701436:
this request is only for go source files, will PR for ts and sh.

PR #2017, opened 2017-04-14
update ignore file

wy65701436:


ywk253100:
Could you delete the useless language files under directory src/ui/static/i18n/lang/? Git prompts that they are deleted every time running make install.

Issue #2034, opened 2017-04-17
LDAP problem with Harbor 1.1.0


roldancer:
Hi All, We had Harbor 0.5.0 working properly with Active Directory (LDAP) but after migrating to Harbor 1.1.0 LDAP integration is not working I get the following error.

Apr 17 17:34:27 172.19.0.1 ui[20184]: 2017-04-17T15:34:27Z [DEBUG] [authenticator.go:59]: Current AUTH_MODE is ldap_auth
Apr 17 17:34:27 172.19.0.1 ui[20184]: 2017-04-17T15:34:27Z [DEBUG] [ldap.go:145]: one or more ldapFilter: FILTER_DELETED_FOR_SECURITY_REASONS
Apr 17 17:34:27 172.19.0.1 ui[20184]: 2017-04-17T15:34:27Z [DEBUG] [ldap.go:280]: username:,email:cXXXXXn@produban.com
Apr 17 17:34:27 172.19.0.1 ui[20184]: 2017-04-17T15:34:27Z [ERROR] [ldap.go:97]: Can't import user , error: duplicate_mailbox
**Apr 17 17:34:27 172.19.0.1 ui[20184]: 2017-04-17T15:34:27Z [ERROR] [base.go:64]: Error occurred in UserLogin: can't import user , error: duplicate_mailbox**

Any idea ?

ericallen:
I ran into the same problem.  I was using sAMAccountName for the ldap uid.  I changed it to cn and it was able to find the user name.  It seems to be in harbor/src/common/utils/ldap/ldap.go lines 204-218 that LDAP attributes are parsed but only uid, cn, mail, and email.  sAMAccountName isn't in the list so it never parses the attribute.  At login, the user is successfully authenticated but it is not able to get the sAMAccountName attribute for the username so the attempt is rejected.  

I would think in adding a case for sAMAccountName would solve the issue.  Something like 

for _, ldapEntry := range result.Entries {
		var u models.LdapUser
		for _, attr := range ldapEntry.Attributes {
			val := attr.Values[0]
			switch attr.Name {
			case ldapConfs.LdapUID:
				u.Username = val
			case "uid":
				u.Realname = val
			case "cn":
				u.Realname = val
			case "mail":
				u.Email = val
			case "email":
				u.Email = val
                       case "samaccountname":
                               u.Email = val
			}
		}
		ldapUsers = append(ldapUsers, u)
	}

ywk253100:
[...]

PR #2037, opened 2017-04-17
Adding in sAMAccountName attribute as possible username issue #2034

ericallen:
When using sAMAccountName as the ldap uid when connecting to Active Directory, harbor is unable to  set the username.  sAMAccountName is not in the list of available attributes to pull the Username from.  #2034


ericallen:
Another possible solution is to not limit the attributes that ldap returns with by changing 

`var attributes = []string{"uid", "cn", "mail", "email"}`  to `var attributes=[]string{}` so that all attributes are returned and allowed to be used to set the username.

reasonerjt:
@ericallen I'm thinking we would just set whatever attribute the user used to authenticate as username, we'll fix it in 1.1.0 branch.

PR #2116, opened 2017-04-24
Return error when got more than one users in GetUser method

ywk253100:


reasonerjt:
@ywk253100 Thanks, but please go through the reference to ensure the callers of this func all expect no more than one user.

PR #2124, opened 2017-04-24
Enable UI UT test framework (karma and jasmine) for UI lib

steven-zou:
Enable UI UT test framework (karma and jasmine) for UI lib


reasonerjt:
shouldn't it be triggered in travis.yml?

steven-zou:
yes, it can be triggered in travis. But I suggest to enable it in travel a little later after  real components are implemented.

PR #2158, opened 2017-04-27
Implement security context for secret

ywk253100:


src/ui/security/context.go
```diff
@@ -17,16 +17,15 @@ package security
 // Context abstracts the operations related with authN and authZ
 type Context interface {
 	// IsAuthenticated returns whether the context has been authenticated or not
-	IsAuthenticated() bool
+	IsAuthenticated() (bool, error)
```
reasonerjt: Please remove the error from the return values of the functions.

reasonerjt:
As discussed, let's move the security context to common package, so it can be reused by other components.

PR #2193, opened 2017-05-02
Implement security context interface

ywk253100:


src/common/security/db/context.go
```diff
@@ -13,3 +13,76 @@
...
 
 package db
+
+import (
+	"github.com/vmware/harbor/src/common/models"
+	"github.com/vmware/harbor/src/ui/pms"
+)
+
+// SecurityContext implements security.Context interface based on database
+type SecurityContext struct {
```
reasonerjt: since pms is a member, it doesn't have to be based on DB?

src/common/security/db/context.go
```diff
@@ -13,3 +13,76 @@
...
+	user *models.User
+	pms  pms.PMS
```
reasonerjt: It maybe OK us pms as a pointer?  As in DB mode, pms can be a system wide variable?

src/ui/pms/service.go
```diff
@@ -12,9 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package project
+package pms
 
 // PMS is the project mamagement service which abstracts
 // the operations related to projects
 type PMS interface {
+	IsPublic(projectIDOrName interface{}) bool
```
reasonerjt: PMS should just return data, it doesn't need to check permission, permission should be checked/interpreted by context.

PR #2197, opened 2017-05-03
Delete column user_id from table access_log

ywk253100:


src/common/models/accesslog.go
```diff
@@ -21,17 +21,16 @@ import (
 // AccessLog holds information about logs which are used to record the actions that user take to the resourses.
 type AccessLog struct {
 	LogID          int       `orm:"pk;auto;column(log_id)" json:"log_id"`
-	UserID         int       `orm:"column(user_id)"  json:"user_id"`
+	Username       string    `orm:"column(username)"  json:"username"`
```
reasonerjt: swagger?

ywk253100: Updated.

src/ui/api/project.go
```diff
@@ -118,6 +118,29 @@ func (p *ProjectAPI) Post() {
...
+
+	user, err := dao.GetUser(models.User{
+		UserID: p.userID,
+	})
+	if err != nil {
+		log.Errorf("failed to get user by ID %d: %v", p.userID, err)
+		p.CustomAbort(http.StatusInternalServerError, http.StatusText(http.StatusInternalServerError))
+	}
+
+	accessLog := models.AccessLog{
```
wy65701436: Do we need an rest api for access log? Then it can call the post to create log here.

src/ui/api/repository.go
```diff
@@ -267,7 +267,19 @@ func (ra *RepositoryAPI) Delete() {
 		go TriggerReplicationByRepository(repoName, []string{t}, models.RepOpDelete)
 
 		go func(tag string) {
```
ywk253100: Use a new goroutine to record the log of project creation.

PR #2239, opened 2017-05-05
Support UI UT

wy65701436:


tests/run-clarity-ut.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -e
+
+cp -r /harbor_ui/lib/* /harbor_ui
+npm run test > lib/npm-ut-test-results
+
+cat /harbor_ui/lib/npm-ut-test-results
```
reasonerjt: seems you cat twice?
And why note fail the process if UI ut fails?  Want to wait for stablization?

PR #2254, opened 2017-05-08
Create different security contexts and project managers according to the request

ywk253100:


src/ui/filter/security.go
```diff
@@ -15,17 +15,25 @@
 package filter
 
 import (
-	"net/http"
 	"strings"
 
 	"github.com/astaxie/beego/context"
-	"github.com/vmware/harbor/src/common/security"
+	"github.com/vmware/harbor/src/common"
```
reasonerjt: Can we call beego/context as beegoContext to avoid confusion?

src/ui/filter/security.go
```diff
@@ -40,20 +48,88 @@ func SecurityFilter(ctx *context.Context) {
...
+	var err error
+
+	// session
+	username := ctx.Input.Session("username")
+	if username != nil {
+		user, err = dao.GetUser(models.User{
+			Username: username.(string),
+		})
+		if err != nil {
+			log.Errorf("failed to get user %s: %v", username.(string), err)
```
reasonerjt: Explicitly indicate the username is from session.

src/ui/filter/security.go
```diff
@@ -40,20 +48,88 @@ func SecurityFilter(ctx *context.Context) {
...
+		}
+
+		if user != nil {
+			log.Info("get credential from session")
```
reasonerjt: Let's say "got user information from session"

src/ui/filter/security.go
```diff
@@ -40,20 +48,88 @@ func SecurityFilter(ctx *context.Context) {
...
+		}
+	}
+
+	// basic auth
+	if user == nil {
+		username, password, ok := ctx.Request.BasicAuth()
```
reasonerjt: I think basic auth has higher priority than session.

src/ui/filter/security.go
```diff
@@ -40,20 +48,88 @@ func SecurityFilter(ctx *context.Context) {
...
 	}
 
-	ctx.Input.SetData(HarborSecurityContext, securityCtx)
+	var user *models.User
+	var err error
+
+	// session
+	username := ctx.Input.Session("username")
+	if username != nil {
+		user, err = dao.GetUser(models.User{
```
reasonerjt: You can't get user from DB when it's in integrated mode.

src/ui/filter/security.go
```diff
@@ -40,20 +48,88 @@ func SecurityFilter(ctx *context.Context) {
...
+			// with vic
+			user, err = auth.Login(models.AuthModel{
+				Principal: username,
+				Password:  password,
+			})
+			if err != nil {
+				log.Errorf("failed to authenticate %s: %v", username, err)
+			}
+			if user != nil {
+				log.Info("get credential from basic auth")
```
reasonerjt: got user information via basic auth

src/ui/filter/security.go
```diff
@@ -40,20 +48,88 @@ func SecurityFilter(ctx *context.Context) {
...
+			}
```
reasonerjt: Also, maybe we should set user id/username, and whatever necessary information in session and do not try to read datasource to get user profile (L80) when there's a valid session?

src/ui/filter/security.go
```diff
@@ -40,20 +48,88 @@ func SecurityFilter(ctx *context.Context) {
...
+		}
+	}
+
+	if user == nil {
+		log.Info("get no credential")
```
reasonerjt: User information is nil.

src/ui/filter/security.go
```diff
@@ -40,20 +48,88 @@ func SecurityFilter(ctx *context.Context) {
...
+	log.Info("creating a rbac security context...")
+	ctx.Input.SetData(HarborSecurityContext,
+		rbac.NewSecurityContext(user, pm))
+
+	return
 }
 
-func createSecurityContext(req *http.Request) (security.Context, error) {
-	return nil, nil
+func getProjectManager(token ...string) projectmanager.ProjectManager {
```
reasonerjt: I think you should use request as input parm, just in case we put token in session?

PR #2290, opened 2017-05-11
Refactor auth of token service

ywk253100:


src/common/security/rbac/context.go
```diff
@@ -59,6 +59,11 @@ func (s *SecurityContext) IsSysAdmin() bool {
 
 // HasReadPerm returns whether the user has read permission to the project
 func (s *SecurityContext) HasReadPerm(projectIDOrName interface{}) bool {
+	// project does not exist
+	if !s.pm.Exist(projectIDOrName) {
+		return false
+	}
+
```
reasonerjt: what would pm.IsPublic() return if the project does not exist? wouldn't it be false already?

src/common/api/base.go
```diff
@@ -136,6 +136,7 @@ func (b *BaseAPI) ValidateUser() int {
 
 // GetUserIDForRequest tries to get user ID from basic auth header and session.
 // It returns the user ID, whether need further verification(when the id is from session) and if the action is successful
+// TODO remove
```
reasonerjt: thanks!

src/ui/service/token/creator.go
```diff
@@ -150,45 +146,29 @@ type repositoryFilter struct {
...
-		if err != nil {
-			log.Errorf("Error occurred in GetPermission: %v", err)
-			//just leave empty permission
-			return nil
-		}
-		if dao.IsProjectPublic(project) {
-			permission += "R"
-		}
+	if ctx.HasAllPerm(project) {
+		permission += "RWM"
```
reasonerjt: permission =, instead of permission +=

src/ui/api/base.go
```diff
@@ -38,14 +38,14 @@ type BaseController struct {
 // context
 func (b *BaseController) Prepare() {
 	ok := false
-	ctx := b.Ctx.Input.GetData(filter.HarborSecurityContext)
+	ctx := b.Ctx.Request.Context().Value(filter.HarborSecurityContext)
```
reasonerjt: need a function "getSecurityContext(req)"?

src/ui/api/base.go
```diff
@@ -38,14 +38,14 @@ type BaseController struct {
...
 	b.SecurityCtx, ok = ctx.(security.Context)
 	if !ok {
 		log.Error("failed to get security context")
 		b.CustomAbort(http.StatusInternalServerError, "")
 	}
 
-	pm := b.Ctx.Input.GetData(filter.HarborProjectManager)
+	pm := b.Ctx.Request.Context().Value(filter.HarborProjectManager)
```
reasonerjt: Same as above.

PR #2299, opened 2017-05-12
Update for restriction check.

wknet123:


src/ui_ng/lib/src/endpoint/endpoint.component.ts
```diff
@@ -143,15 +143,21 @@ export class EndpointComponent implements OnInit {
   }
 
   editTarget(target: Endpoint) {
-    if (target) {
+    if(target) {
       let editable = true;
+      if (!target.id) {
```
steven-zou: let id:number = target.id?target.id:0;
if(id ===0 ){
  return;
}

src/ui_ng/lib/src/create-edit-endpoint/create-edit-endpoint.component.ts
```diff
@@ -267,20 +269,23 @@ export class CreateEditEndpointComponent implements AfterViewChecked {
...
-          let origin: string = comparison[key];
-          if(((this.actionType === ActionType.EDIT && this.editable && !current)  || current) && 
-            current !== origin) {
-            this.hasChanged = true;
-            break;
-          } else {
-            this.hasChanged = false;
-            this.inlineAlert.close();
+      let self: CreateEditEndpointComponent | any = this;
+      if(self) {
```
steven-zou: if(self&&self.targetForm){
  ...
}

PR #2305, opened 2017-05-14
修改初始化脚本表字段缺失问题

leoterry-ulrica:
修改初始化脚本表字段缺失问题


reasonerjt:
We are doing refactory and the columns are removed intentionally.

PR #2308, opened 2017-05-15
Add the discription of notary in User Guide

m-masataka-ns:
In user_guide.md, Content trust paragraph does't  has enough  information of pushing images.
In this commit, I added the description of notary's ''Passphrase``.


reasonerjt:
@m-masataka thanks for the PR.

I'm a bit reluctant to add that into our user guide, as that's implementation details of docker.

PR #2316, opened 2017-05-16
Refactor project API

ywk253100:


src/common/const.go
```diff
@@ -24,10 +24,9 @@ const (
 	LDAPScopeOnelevel   = "2"
 	LDAPScopeSubtree    = "3"
 
-	RoleSystemAdmin  = 1
-	RoleProjectAdmin = 2
-	RoleDeveloper    = 3
-	RoleGuest        = 4
+	RoleProjectAdmin = 1
```
reasonerjt: They are for internally communication, and has nothing to do with the role_code in DB right?

src/common/dao/project.go
```diff
@@ -159,7 +160,7 @@ func ToggleProjectPublicity(projectID int64, publicity int) error {
 // which satisfies the following conditions:
 // 1. the project is not deleted
 // 2. the prject is public or the user is a member of the project
-func SearchProjects(userID int) ([]models.Project, error) {
+func SearchProjects(userID int) ([]*models.Project, error) {
```
reasonerjt: why return pointers here? Does "QueryRows" in orm supports both pointer to struct and struct?

src/common/models/project.go
```diff
@@ -39,7 +40,7 @@ type Project struct {
 
 // ProjectSorter holds an array of projects
 type ProjectSorter struct {
-	Projects []Project
+	Projects []*Project
```
reasonerjt: why pointers?

src/ui/projectmanager/db/pm.go
```diff
@@ -106,29 +109,107 @@ func (p *ProjectManager) GetRoles(username string, projectIDOrName interface{})
...
+		pro := p.Get(projectIDOrName)
+		if pro == nil {
+			return fmt.Errorf(fmt.Sprintf("project %v not found", projectIDOrName))
+		}
+		id = pro.ProjectID
+	}
+	return dao.ToggleProjectPublicity(id, project.Public)
+}
+
+// GetAll ...
```
reasonerjt: need to explain each parm or point to the interface file so user can see the doc.

src/ui/projectmanager/db/pm.go
```diff
@@ -106,29 +109,107 @@ func (p *ProjectManager) GetRoles(username string, projectIDOrName interface{})
...
-		return []models.Project{}
+		log.Errorf("failed to get total of projects: %v", err)
+		return []*models.Project{}, 0
 	}
-	projects, err := dao.GetUserRelevantProjects(user.UserID, "")
+
+	return filter(owner, name, public, member, role, page, size), total
+}
+
+func filter(owner, name, public, member string,
```
reasonerjt: See comments of "GetAll" in "src/ui/projectmanager/pm.go"

src/ui/projectmanager/pm.go
```diff
@@ -26,7 +26,20 @@ type ProjectManager interface {
...
+	Update(projectIDOrName interface{}, project *models.Project) error
+	// GetAll returns a project list and the total count according to
+	// the query conditions:
+	// owner: username of owner
+	// name: name of project
+	// public: public or not, can be "true", "false" or ""
+	// member: username of the member
+	// role: the role of member specified by member parameter
+	// page, size: pagination parameters
+	GetAll(owner, name, public, member string, role int, page,
```
reasonerjt: This long list of parms is a bit error prone as a public func
1, public should be bool
2,And why we need the role parm?  in this PR, I see all "role" are zero.  And since member and role are tied so probably put them in one struct.
3. page and size are tied so probably put them into one struct.  And I don't think the project service support pagination
4, as for the second return value, does it has to be returned in one call?
5, I see in the implementation you use "filter" for different functions, it's acceptable as it's internal, but why we need a "GetAll" public func?

src/common/dao/project.go
```diff
@@ -180,99 +181,102 @@ func SearchProjects(userID int) ([]models.Project, error) {
...
+	err := GetOrmer().Raw(sql, params).QueryRow(&total)
 	return total, err
 }
 
-// GetUserRelevantProjects returns the user relevant projects
-// args[0]: public, args[1]: limit, args[2]: offset
-func GetUserRelevantProjects(userID int, projectName string, args ...int64) ([]models.Project, error) {
-	return getProjects(userID, projectName, args...)
+// GetProjects returns a project list according to the query conditions
+func GetProjects(owner, name, public, member string,
```
reasonerjt: why is public a string? 
And the long list of parm is a bit error prone see comments in "src/ui/projectmanager/pm.go"

src/common/api/base.go
```diff
@@ -39,6 +39,17 @@ type BaseAPI struct {
 	beego.Controller
 }
 
+// GetStringFromPath gets the param from path and returns it as string
+func (b *BaseAPI) GetStringFromPath(key string) string {
```
reasonerjt: How is this used?  I don't see any reference in this PR?

src/common/api/base.go
```diff
@@ -39,6 +39,17 @@ type BaseAPI struct {
...
+	return b.Ctx.Input.Param(key)
+}
+
+// GetInt64FromPath gets the param from path and returns it as int64
+func (b *BaseAPI) GetInt64FromPath(key string) (int64, error) {
```
reasonerjt: How is this used? I don't see any reference in this PR?

src/ui/projectmanager/pm.go
```diff
@@ -26,7 +26,20 @@ type ProjectManager interface {
 	Exist(projectIDOrName interface{}) bool
 	GetRoles(username string, projectIDOrName interface{}) []int
 	// get all public project
-	GetPublic() []models.Project
+	GetPublic() []*models.Project
```
reasonerjt: Need to return error for all functions that depend on other services.

src/common/dao/project.go
ywk253100: "public" can be "true", "false" and "". If it is nil, the query will not filter the publicity.

src/common/dao/project.go
```diff
@@ -180,99 +181,102 @@ func SearchProjects(userID int) ([]models.Project, error) {
...
+	role int, page, size int64) ([]*models.Project, error) {
```
wy65701436: A little confuse about these two parameters -- page and size.  The swagger file gives the name are page and page_size, and also gives that the default value. But don't see any default value in the source code.

src/common/dao/project.go
wy65701436: update the swagger accordingly.

src/common/dao/project.go
```diff
@@ -180,99 +181,102 @@ func SearchProjects(userID int) ([]models.Project, error) {
...
-	switch len(args) {
-	case 1:
-		sql += ` and p.public = ?`
-		queryParam = append(queryParam, args[0])
-		sql += ` order by p.name `
-	case 2:
-		sql += ` order by p.name `
-		sql = paginateForRawSQL(sql, args[0], args[1])
-	case 3:
+	if len(public) != 0 {
```
wy65701436: what about if I set public as "somevalue"? As it may come from Rest, the check if not enough I think.

src/common/dao/project.go
ywk253100: It's a parameter of a function not API, API does not change, so no swagger needs to be updated.

src/common/dao/project.go
ywk253100: They are set in API handler.

src/common/dao/project.go
ywk253100: The validation is done in API handler.

src/ui/api/project.go
```diff
@@ -120,24 +132,15 @@ func (p *ProjectAPI) Post() {
...
+			models.AccessLog{
+				Username:  p.SecurityCtx.GetUsername(),
+				ProjectID: projectID,
+				RepoName:  pro.ProjectName + "/",
+				RepoTag:   "N/A",
+				Operation: "create",
+				OpTime:    time.Now(),
+			}); err != nil {
 			log.Errorf("failed to add access log: %v", err)
 			p.CustomAbort(http.StatusInternalServerError, http.StatusText(http.StatusInternalServerError))
```
reasonerjt: I find here is not consistent with other parts that create access logs, probably should not return 500. just log the error.

src/ui/projectmanager/pm.go
```diff
@@ -18,15 +18,43 @@ import (
 	"github.com/vmware/harbor/src/common/models"
 )
 
+// QueryParam can be used to set query parameters when listing projects
+type QueryParam struct {
```
reasonerjt: Can we open another issue to re-use these parms in dao?

src/ui/projectmanager/pm.go
```diff
@@ -18,15 +18,43 @@ import (
...
+	Name       string      // the name of project
+	Owner      string      // the username of project owner
+	Public     string      // the project is public or not, can be "ture","false" and ""
+	Member     *Member     // the member of project
```
reasonerjt: could you explain how this is used?  what if it's none?

src/ui/projectmanager/pm.go
```diff
@@ -18,15 +18,43 @@ import (
 	"github.com/vmware/harbor/src/common/models"
 )
 
+// QueryParam can be used to set query parameters when listing projects
+type QueryParam struct {
+	Name       string      // the name of project
+	Owner      string      // the username of project owner
+	Public     string      // the project is public or not, can be "ture","false" and ""
```
reasonerjt: I'm not comfortable with this, essentially we need a union type here.  If you don't want to implement a interface with limited "isxxx" function, maybe a pointer to a bool? so it has three condition, and can't be wrong?  Also explains what will happen is it's nil.

src/jobservice/replication/delete.go
```diff
@@ -90,7 +90,7 @@ func (d *Deleter) enter() (string, error) {
 
 	// delete repository
 	if len(d.tags) == 0 {
-		u := url + d.repository + "/tags"
+		u := url + d.repository
```
reasonerjt: Huh? is it part of the refactory? Or is it for fixing other bugs?

src/ui/projectmanager/pm.go
```diff
@@ -18,15 +18,43 @@ import (
...
 	// get projects which the user is a member of
-	GetByMember(username string) []models.Project
+	GetByMember(username string) ([]*models.Project, error)
+	Create(*models.Project) (int64, error)
+	Delete(projectIDOrName interface{}) error
+	Update(projectIDOrName interface{}, project *models.Project) error
+	// GetAll returns a project list according to the query parameters
+	GetAll(query *QueryParam) ([]*models.Project, error)
+	// GetTotal returns the total count according to the query parameters
+	GetTotal(query *QueryParam) (int64, error)
```
reasonerjt: Please explicily explain the reason to introduce this "GetTotal" func, per my understanding it is for paging right?

src/jobservice/replication/delete.go
ywk253100: It's the refactor. The API is changed in repository API refactor.

src/ui/projectmanager/pm.go
ywk253100: Yes, it is for paging.

PR #2332, opened 2017-05-17
merge 1.1.0 to master

reasonerjt:


ywk253100:
We have merged release-1.1.0 into master by #2303 , why here is another one?

reasonerjt:
Maybe because I messed up when rebasing release-1.1.0 branch, however, the concern is if we don't merge this commit, it will make it harder to merge future changes.  I wish to propose we do this merge to master, in case there's issue in code in master branch we can fix it later.

PR #2341, opened 2017-05-19
Refactor member API

ywk253100:


src/ui/api/member.go
```diff
@@ -41,47 +39,82 @@ type memberReq struct {
...
-	pma.currentUserID = pma.ValidateUser()
-	mid := pma.Ctx.Input.Param(":mid")
-	if mid == "current" {
-		pma.memberID = pma.currentUserID
-	} else if len(mid) == 0 {
-		pma.memberID = 0
-	} else if len(mid) > 0 {
-		memberID, err := strconv.Atoi(mid)
+
+	if len(pma.GetStringFromPath(":mid")) != 0 {
```
reasonerjt: Sorry but I really don't recall, what was the ` mid == "current"` or `len(mid) == 0` for?
And are we sure they're no longer needed?

src/ui/api/member.go
```diff
@@ -41,47 +39,82 @@ type memberReq struct {
...
+			if err != nil {
+				text += err.Error()
+			} else {
+				text += fmt.Sprintf("%d", mid)
+			}
+			pma.HandleBadRequest(text)
+			return
+		}
+
+		member, err := dao.GetUser(models.User{
```
reasonerjt: It should be my fault, the member id seems to map to user id?  This doesn't look quite right, could you open an issue for us to follow up?

ywk253100: OK.

src/ui/api/member.go
ywk253100: I have confirmed with kun and made a global search, they are no longer needed.

src/ui/api/member.go
ywk253100: There is no ID in table project_member, it uses project ID and user ID as a combined primary key. So does we need to do some modification?

reasonerjt: not now, please open an issue to follow up.

PR #2348, opened 2017-05-19
refactor job service

reasonerjt:
As long as Job interface is implemented, an arbitrary kind of job should be handled by statemachine.

For now, we will use different workpools to handle different kinds of jobs.


src/jobservice/job/statemachine.go
```diff
@@ -182,125 +165,100 @@ func (sm *SM) Init() {
...
-		}
-		pwd, err = uti.ReversibleDecrypt(pwd, key)
-		if err != nil {
-			return fmt.Errorf("failed to decrypt password: %v", err)
+func (sm *SM) kickOff() error {
+	if repJob, ok := sm.CurrentJob.(*RepJob); ok {
+		if repJob.parm.Enabled == 0 {
+			log.Debugf("The policy of job:%v is disabled, will cancel the job", repJob)
+			err := repJob.UpdateStatus(models.JobCanceled)
+			log.Warningf("Failed to update status of job: %v to 'canceled', error: %v", repJob, err)
```
ywk253100: if err != nil ?

ywk253100: And return an error?

src/jobservice/job/workerpool.go
```diff
@@ -15,29 +15,35 @@
...
+	poolType   Type
 	workerChan chan *Worker
 	workerList []*Worker
 }
 
-// WorkerPool is a set of workers each worker is associate to a statemachine for handling jobs.
-// it consists of a channel for free workers and a list to all workers
-var WorkerPool *workerPool
+// WorkerPools is a map contains workerpools for different types of jobs.
+var WorkerPools map[Type]*workerPool
```
ywk253100: Why we need two kinds of worker pool? Should not the workers be same as they all handle jobs? The handlers of jobs can be different.

src/jobservice/job/jobs.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+// LogPath returns the absolute path of the particular replication job.
+func (rj *RepJob) LogPath() string {
+	return GetJobLogPath(config.LogDir(), rj.id)
+}
+
+// UpdateStatus ...
+func (rj *RepJob) UpdateStatus(status string) error {
+	return dao.UpdateScanJobStatus(rj.id, status)
```
ywk253100: Should be UpdateRepJobStatus I think.

reasonerjt: correct...thanks

src/jobservice/job/workerpool.go
reasonerjt: They are two worker pool with same kind, just each resource pool for one kind of jobs.

src/jobservice/job/statemachine.go
reasonerjt: Will change, thanks

PR #2355, opened 2017-05-23
Return signatures in GET tags API

ywk253100:


src/ui/api/repository.go
```diff
@@ -319,41 +321,115 @@ func (ra *RepositoryAPI) GetTags() {
...
+	list := []*tag{}
+
+	for _, t := range ts {
+		digest, manifest, config, err := getV2Manifest(client, t)
+		if err != nil {
+			return nil, err
+		}
+
+		// manifest can be used to calculate the image size
+		_ = manifest
```
reasonerjt: move the comment to line 382, and ignore the manifest in line 382 
"digest, _, config, err := xxxxxxxx"

PR #2357, opened 2017-05-23
Fix bug

ywk253100:
1. Modify the repoisotory API call in jobservice as the API changed.
2. Init secret store map
3. Init worker pools map, or the jobservice can not start up


src/jobservice/job/workerpool.go
```diff
@@ -124,6 +124,7 @@ func InitWorkerPools() error {
 	if err != nil {
 		return err
 	}
+	WorkerPools = make(map[Type]*workerPool)
```
reasonerjt: What???  And the CI passed...

PR #2367, opened 2017-05-24
Refactor search API

ywk253100:


src/ui/api/search.go
```diff
@@ -76,17 +69,19 @@ func (s *SearchAPI) Get() {
 			continue
 		}
 
-		if userID != dao.NonExistUserID {
-			roles, err := dao.GetUserProjectRoles(userID, p.ProjectID)
+		if isAuthenticated {
+			roles, err := s.ProjectMgr.GetRoles(username, p.ProjectID)
```
wy65701436: Why doesn't call get role before the loop?

src/ui/api/search.go
```diff
@@ -39,33 +40,25 @@ type searchResult struct {
...
 	var projects []*models.Project
+	var err error
 
-	if isSysAdmin {
-		projects, err = dao.GetProjects(nil)
-		if err != nil {
-			log.Errorf("failed to get all projects: %v", err)
-			s.CustomAbort(http.StatusInternalServerError, "internal error")
-		}
+	if !isAuthenticated {
```
wy65701436: if !isAuthenticated {
	projects, err = s.ProjectMgr.GetPublic()
} 
if isSysAdmin {
	projects, err = s.ProjectMgr.GetAll(nil)
} else {
	projects, err = s.ProjectMgr.GetHasReadPerm(username)
}

It could be better to understand.

src/ui/api/search.go
```diff
@@ -76,17 +69,19 @@ func (s *SearchAPI) Get() {
 			continue
 		}
 
-		if userID != dao.NonExistUserID {
-			roles, err := dao.GetUserProjectRoles(userID, p.ProjectID)
+		if isAuthenticated {
```
wy65701436: extract an method, isProjectTogglable() ?

ywk253100: The codes not only populate the public property of project, but also set the role which user has to the project.

PR #2376, opened 2017-05-25
Refactor email API

ywk253100:


src/ui/api/email.go
```diff
@@ -32,20 +30,20 @@ const (
...
 }
 
 // Prepare ...
 func (e *EmailAPI) Prepare() {
-	userID := e.ValidateUser()
-	isSysAdmin, err := dao.IsAdminRole(userID)
-	if err != nil {
-		log.Errorf("failed to check the role of user: %v", err)
-		e.CustomAbort(http.StatusInternalServerError, http.StatusText(http.StatusInternalServerError))
+	e.BaseController.Prepare()
```
wy65701436: Is there any better way to have a general rule to check the authentication?  I see the duplicate code block in several APIs.

ywk253100: Yes, we can add a filter to check the authentication. But it's complicate as not all API calls need authentication, we need a filter to do the authentication according to the url, method, etc. Maybe we can refactor this part after completing the current work.

wy65701436: Can you please add an new issue for it? I can merge code firstly.

ywk253100: OK. Thanks.

ywk253100: Open issue #2426 to track this.

PR #2378, opened 2017-05-25
Refactor sync registry API

ywk253100:


src/ui/api/utils.go
```diff
@@ -377,7 +379,7 @@ func diffRepos(reposInRegistry []string, reposInDB []string) ([]string, []string
 	for i < len(reposInRegistry) {
 		repoInR = reposInRegistry[i]
 		i++
-		exist, err := projectExists(repoInR)
+		exist, err := projectExists(pm, repoInR)
```
wy65701436: pm.Exist(utils.ParseRepository(repoInR))

Is it better?

ywk253100: The function utils.ParseRepository(repo) has two return values, so we can not put it as the parameter of pm.Exist().

PR #2422, opened 2017-06-05
fix #2382

### LINKED ISSUE #2382: The retrying jobs can not be canceled although the policy is disabled already

reasonerjt:


reasonerjt:
btw the decreased coverage was due to some previous code change in api

PR #2455, opened 2017-06-07
Add system info service to shareable components.

wknet123:


src/ui_ng/lib/src/tag/tag.component.ts
```diff
@@ -128,6 +121,14 @@ export class TagComponent implements OnInit {
           if(this.tags && this.tags.length === 0) {
             this.refreshRepo.emit(true);
           }
+          toPromise<SystemInfo>(this.systemInfoService.getSystemInfo())
```
steven-zou: Move this function call to ngOnInit()

PR #2483, opened 2017-06-09
Fix timestamp value in query params

wknet123:


src/ui_ng/src/app/log/audit-log.service.ts
```diff
@@ -36,8 +36,12 @@ export class AuditLogService {
...
     let params: URLSearchParams = new URLSearchParams(queryParam.keywords);
-    params.set('begin_timestamp', <string>queryParam.begin_timestamp);    
-    params.set('end_timestamp', <string>queryParam.end_timestamp);
+    if(queryParam.begin_timestamp) {
+      params.set('begin_timestamp', <string>queryParam.begin_timestamp);    
+    }
+    if(queryParam.end_timestamp) {
+      params.set('end_timestamp', <string>queryParam.end_timestamp);
+    }
     params.set('username', queryParam.username);
```
steven-zou: If empty value, do not set into params

PR #2506, opened 2017-06-13
handlers for image scan, store results overview in DB

reasonerjt:


make/docker-compose.clair.yml
```diff
@@ -35,6 +38,9 @@ services:
     restart: always
     depends_on:
       - postgres
+#    environment:
```
wy65701436: remove the proxy here

make/common/templates/clair/config.yaml
```diff
@@ -16,7 +16,7 @@ clair:
     # Deadline before an API request will respond with a 503
     timeout: 300s
   updater:
-    interval: 0h
+    interval: 2h
```
wy65701436: will we let user to config these?

reasonerjt: IMO we won't

make/docker-compose.clair.yml
reasonerjt: oops, thanks!

PR #2525, opened 2017-06-15
GET tags api includes scan overview

reasonerjt:


src/ui/api/repository.go
```diff
@@ -636,3 +639,27 @@ func getSignatures(repository, username string) (map[string]*notary.Target, erro
...
+	}
+	if data == nil {
+		return nil
+	}
+	job, err := dao.GetScanJob(data.JobID)
+	if err != nil {
+		log.Errorf("Failed to get scan job for id:%d, error: %v", data.JobID, err)
+		return nil
+	}
+	data.Status = job.Status
```
ywk253100: Need to check whether the job is nil

PR #2558, opened 2017-06-18
Add security context based on admiral

ywk253100:


src/common/security/admiral/context.go
```diff
@@ -0,0 +1,174 @@
...
+	"github.com/vmware/harbor/src/common/security/authcontext"
+	"github.com/vmware/harbor/src/common/utils/log"
+	"github.com/vmware/harbor/src/ui/projectmanager"
+)
+
+// SecurityContext implements security.Context interface based on
+// auth context and project manager
+type SecurityContext struct {
+	ctx *authcontext.AuthContext
+	pm  projectmanager.ProjectManager
```
reasonerjt: should be a Admiral PMS project manager.

src/common/security/admiral/context.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// NewSecurityContext ...
+func NewSecurityContext(ctx *authcontext.AuthContext, pm projectmanager.ProjectManager) *SecurityContext {
```
reasonerjt: See previous comment.

src/ui/filter/security.go
```diff
@@ -54,93 +78,175 @@ func SecurityFilter(ctx *beegoctx.Context) {
 	}
 
 	if !strings.HasPrefix(req.URL.RequestURI(), "/api/") &&
-		!strings.HasPrefix(req.URL.RequestURI(), "/service/") {
+		!strings.HasPrefix(req.URL.RequestURI(), "/service/token") {
```
reasonerjt: By default let the request pass through the modifiers.
Let's use a blacklist instead of a white list.

PR #2578, opened 2017-06-20
Add get tag api to swagger.yaml

ywk253100:


docs/swagger.yaml
```diff
@@ -769,6 +769,30 @@ paths:
         403:
           description: Forbidden.
   /repositories/{repo_name}/tags/{tag}:
+    get:
+      summary: Get the tag of the repository.
+      description: |
+        This endpoint aims to retrieve the tag of the repository.
```
reasonerjt: please add more description about details of tag. For example when Harbor is deployed with notary ....

PR #2621, opened 2017-06-25
[WIP] Harbor HA

m-masataka:
I implement Harbor HA with MySQL Cluster, and want to discuss this architecture with you.
In my usecase, I must build up Harbor with HA and set mode to  Active-Active. 
Here is the architecture that I have thought.


reasonerjt:
@m-masataka thanks for the PR, HA is definitely on our roadmap, but currently we are focusing on the delivery for next release, so let's keep this PR open and discuss the scenario/solutions.  We may have to do it intermittently.

m-masataka:
@reasonerjt
Thank's for your reply. 
I got it. Let's discuss about HA.
Please let me know if there is already a place that discuss about HA.

reasonerjt:
There is a very old issue https://github.com/vmware/harbor/issues/327
, but the scope and roadmap has changed a lot since then, I think you can open a new one to elaborate your scenario and proposal to fix them.

PR #2629, opened 2017-06-26
Update clair postgresql to use photon os

yixingjia:
fix  https://github.com/vmware/harbor/issues/2628

### DIFF
--- a/make/common/postgresql/Dockerfile
+++ b/make/common/postgresql/Dockerfile
@@ -6,6 +6,7 @@ RUN touch /etc/localtime.bak \
     && tdnf install -y sed shadow gzip postgresql\
     && groupadd -r postgres --gid=999 \
     && useradd -r -g postgres --uid=999 postgres \
+    && mkdir -p /docker-entrypoint-initdb.d \
     && mkdir -p /run/postgresql \
     && chown -R postgres:postgres /run/postgresql \
     && chmod 2777 /run/postgresql \
--- a/make/common/postgresql/Dockerfile
+++ b/make/common/postgresql/Dockerfile
@@ -1,0 +1,25 @@
+FROM library/photon:1.0
+
+ENV PGDATA /var/lib/postgresql/data
+
+RUN touch /etc/localtime.bak \
+    && tdnf install -y sed shadow gzip postgresql\
+    && groupadd -r postgres --gid=999 \
+    && useradd -r -g postgres --uid=999 postgres \
+    && mkdir -p /run/postgresql \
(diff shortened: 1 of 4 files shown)


make/common/postgresql/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+    && useradd -r -g postgres --uid=999 postgres \
+    && mkdir -p /run/postgresql \
+    && chown -R postgres:postgres /run/postgresql \
+    && chmod 2777 /run/postgresql \
+    && mkdir -p "$PGDATA" && chown -R postgres:postgres "$PGDATA" && chmod 777 "$PGDATA" \
+    && sed -i "s|#listen_addresses = 'localhost'.*|listen_addresses = '*'|g" /usr/share/postgresql/postgresql.conf.sample \
+    && sed -i "s|#unix_socket_directories = '/tmp'.*|unix_socket_directories = '/run/postgresql'|g" /usr/share/postgresql/postgresql.conf.sample \
+    && touch /usr/share/locale/locale.alias \
+    && locale-gen.sh en_US.UTF-8
+
```
reasonerjt: I recall there's a xxxx-initdb.d directory and when the sql file is put there it will be executed automatically?
Could you investigate that, it's fine if it's covered in future work item.

make/docker-compose.clair.yml
```diff
@@ -17,7 +17,7 @@ services:
         aliases:
           - postgres
     container_name: clair-db
-    image: postgres:9.6.3
+    image: vmware/harbor-postgresql:9.6.3
```
reasonerjt: Is it postgresql 9.6.3?
Also, shall we call it vmware/postgresql:$version-photon to follow the convention?

make/common/postgresql/Dockerfile
yixingjia: that part will be covered by make/common/postgresql/docker-entrypoint.sh 

we have all the code there , but for clair specify , we can investigate later.

make/docker-compose.clair.yml
yixingjia: This is for postgresql9.6.3.

I think we should use postgreql version instead of photon version.

as we may have different PostgreSQL version with same photon os version in the future.

I can move the magic number 9.6.3 to a central place like the Makefile.

and change it on the run , just like __version__ in the docker-compose.tpl,  
What do you think?

make/common/postgresql/Dockerfile
yixingjia: add the missing directory and test it locally, please recheck.

PR #2651, opened 2017-06-27
configurations API on UI should accept not only string as JSON attrib…

reasonerjt:
…ute type

### DIFF
--- a/src/ui/api/config.go
+++ b/src/ui/api/config.go
@@ -17,7 +17,7 @@ package api
 import (
 	"fmt"
 	"net/http"
-	"strconv"
+	"reflect"
 
 	"github.com/vmware/harbor/src/common"
 	"github.com/vmware/harbor/src/common/dao"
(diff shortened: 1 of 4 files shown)


src/ui/api/config.go
```diff
@@ -199,111 +199,122 @@ func (c *ConfigAPI) Reset() {
...
-		if !ok {
-			continue
-		}
-
-		n, err := strconv.Atoi(v)
-		if err != nil || n < 0 {
-			return isSysErr, fmt.Errorf("invalid %s: %s", k, v)
-		}
-
+	for k, n := range numMap {
```
ywk253100: Check whether the numeric value > 0

src/ui/api/config.go
```diff
@@ -199,111 +199,122 @@ func (c *ConfigAPI) Reset() {
...
 			common.ProjectCreationRestriction,
 			common.ProCrtRestrAdmOnly,
 			common.ProCrtRestrEveryone)
 	}
 
-	return isSysErr, nil
+	return false, nil
 }
 
 //convert map[string]string to map[string]interface{}
```
ywk253100: This method is not used anymore, maybe we can delete it.

PR #2658, opened 2017-06-28
Provide a method to get token from token service

ywk253100:


src/common/utils/registry/auth/tokenauthorizer.go
```diff
@@ -166,66 +160,21 @@ func NewStandardTokenAuthorizer(credential Credential, insecure bool,
 	return authorizer
 }
 
-func (s *standardTokenAuthorizer) generateToken(realm, service string, scopes []string) (token string, expiresIn int, issuedAt *time.Time, err error) {
+func (s *standardTokenAuthorizer) generateToken(realm, service string, scopes []string) (string, int, *time.Time, error) {
```
reasonerjt: Irrelevant to this refactory but I feel it'd be better to return an instance of Time instead of a pointer?  Is there any reason for that?

I'm ok it's be fixed later.

PR #2670, opened 2017-06-29
Implement scan all policy configuration

steven-zou:
Implement scan all policy configuration


wy65701436:
Please fix the test error.

PR #2685, opened 2017-07-03
Update codes of auth context

ywk253100:


src/common/security/authcontext/authcontext.go
```diff
@@ -158,52 +152,51 @@ func Login(url, username, password string) (string, *AuthContext, error) {
...
+		return nil, err
 	}
-	return resp.StatusCode, resp.Header, data, nil
-}
 
-func buildCtxURL(url, principalID string) string {
-	url = strings.TrimRight(url, "/") + "/sso/auth-context"
-	if len(principalID) > 0 {
-		url += "/" + principalID
+	if resp.StatusCode != http.StatusOK {
```
reasonerjt: Need to double check for requests the expected response code is 200, otherwise let's put it as a parm.

src/ui/projectmanager/pms/pm.go
```diff
@@ -304,16 +304,12 @@ func (p *ProjectManager) GetPublic() ([]*models.Project, error) {
 // GetByMember ...
 func (p *ProjectManager) GetByMember(username string) ([]*models.Project, error) {
 	projects := []*models.Project{}
-	ctx, err := authcontext.GetAuthCtxOfUser(p.endpoint, p.getToken(), username)
-	if err != nil {
-		return projects, err
-	}
-
-	names, err := ctx.GetMyProjects()
+	ctx, err := authcontext.GetAuthCtxOfSpecificUser(p.client, p.endpoint, p.getToken(), username)
```
reasonerjt: We probably can read the projects from security context directly instead of depending on PM.

src/ui/projectmanager/pms/pm.go
```diff
@@ -304,16 +304,12 @@ func (p *ProjectManager) GetPublic() ([]*models.Project, error) {
 // GetByMember ...
 func (p *ProjectManager) GetByMember(username string) ([]*models.Project, error) {
```
reasonerjt: Probably we should move "GetMyProjects" to security context so in admiral case there's no dependency on PM.

src/common/security/authcontext/authcontext.go
reasonerjt: Need to make sure 200 is always expected in correct scenario.

src/common/security/authcontext/authcontext.go
```diff
@@ -90,66 +93,57 @@ func (a *AuthContext) HasAllPerm(project string) bool {
...
 	}
-	return projects, nil
+	return projects
 }
 
-// GetByToken ...
-func GetByToken(url, token string) (*AuthContext, error) {
-	return get(url, token)
+// GetAuthCtxOfCurrentUser returns the auth context of the current user
+func GetAuthCtxOfCurrentUser(client *http.Client, url, token string) (*AuthContext, error) {
```
reasonerjt: Better rename:
GetAuthCtxOfCurrentUser -> GetAuthCtx
GetAuthCtxOfSpecificUser -> GetAuthCtxOfUser

src/common/security/authcontext/authcontext.go
ywk253100: I'll update this after testing.

src/ui/projectmanager/pms/pm.go
ywk253100: Open another issue #2692 to track this.

PR #2693, opened 2017-07-04
Clair notification handler

reasonerjt:


make/common/templates/clair/config.yaml
```diff
@@ -16,8 +16,10 @@ clair:
     # Deadline before an API request will respond with a 503
     timeout: 300s
   updater:
-    interval: 2h
+    interval: 1h
```
yixingjia: We properly need to change this number to a larger one.

reasonerjt: Thanks, this is for testing, so Clair will generate notifications more frequently, I'll open another issue to put interval to 12h in next sprint.

src/ui/service/notifications/clair/handler.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// returns true to indicate it should reshedule the "rescan" action.
+func (t *timer) needReschedule() bool {
+	t.Lock()
+	defer t.Unlock()
+	if time.Now().Before(t.next) {
+		return false
+	}
+	t.next = time.Now().Add(rescanInterval)
```
ywk253100: I don't know what the frequency that the notifications will be received is. If it's high, maybe we can put the lock operation after the if statement.

src/ui/service/notifications/clair/handler.go
```diff
@@ -0,0 +1,108 @@
...
+	var ne models.ClairNotificationEnvelope
+	if err := json.Unmarshal(h.Ctx.Input.CopyBody(1<<32), &ne); err != nil {
+		log.Errorf("Failed to decode the request: %v", err)
+		return
+	}
+	log.Debugf("Received notification from Clair, name: %s", ne.Notification.Name)
+	notification, err := clairClient.GetNotification(ne.Notification.Name)
+	if err != nil {
+		log.Errorf("Failed to get notification details from Clair, name: %s, err: %v", ne.Notification.Name, err)
+	}
```
ywk253100: return?

src/ui/service/notifications/clair/handler.go
```diff
@@ -0,0 +1,108 @@
...
+	if new := notification.New; new != nil {
+		if vuln := new.Vulnerability; vuln != nil {
+			log.Debugf("new vulnerability namespace: %s", vuln.NamespaceName)
+			ns[vuln.NamespaceName] = true
+		}
+	}
+	for k, v := range ns {
+		if v {
+			if err := dao.SetClairVulnTimestamp(k, time.Now()); err != nil {
+				log.Debugf("Updated the timestamp for namespaces: %s", k)
```
ywk253100: err == nil

src/ui/service/notifications/clair/handler.go
reasonerjt: right...

src/ui/service/notifications/clair/handler.go
reasonerjt: right...

src/ui/service/notifications/clair/handler.go
reasonerjt: But Add may not be atomic then Before may accepts an incomplete data as parm?

PR #2697, opened 2017-07-04
Prepare rsyslog docker based on photon

yixingjia:


make/photon/log/Dockerfile
```diff
@@ -1,12 +1,5 @@
-FROM library/photon:1.0
+FROM vmware/rsyslog-photon:8.15.0
```
reasonerjt: I think we'd better put the Dockerfile of this image somewhere in the repo, just in case we need to upgrade or modify the image.

yixingjia: good catch. please check the updated PR.

PR #2700, opened 2017-07-04
Update Clair to 2.0.1

yixingjia:
fix the Debian mapping issue.


yixingjia:
fixes #2701

reasonerjt:
Shall we also update the makefile to package the 2.0.1 img?

PR #2705, opened 2017-07-04
Implement policy scheduler

steven-zou:


reasonerjt:
@steven-zou we'll spend a bit more time on this PR, probably done by this Friday.

src/common/scheduler/policy/alternate_policy.go
```diff
@@ -0,0 +1,185 @@
+package policy
+
+import "github.com/vmware/harbor/src/common/scheduler/task"
```
reasonerjt: Let's be consistent to use import (...)

src/common/scheduler/policy/alternate_policy.go
```diff
@@ -0,0 +1,185 @@
...
+import "errors"
+import "time"
+
+//AlternatePolicyConfiguration store the related configurations for alternate policy.
+type AlternatePolicyConfiguration struct {
```
reasonerjt: Do we need this?  Given policy can be very lightweight and highly flexible, how about implement a simple "Daily" policy?

src/common/scheduler/policy/evaluation_result.go
```diff
@@ -0,0 +1,15 @@
+package policy
+
+//EvaluationResult is defined to carry the policy evaluated result.
+//
+//Filed 'Result' is optional.
+//Filed 'Error' is optional
+//
+type EvaluationResult struct {
```
reasonerjt: Let's not do this, Scheduler didn't even handle the Error in result.
Maybe the "Evaluate" can return a channel and an error to capture the error when generating the "evalchan" and other error can be sent via another channel,  *if needed*

src/common/scheduler/policy/policy.go
```diff
@@ -0,0 +1,37 @@
...
+
+	//Done will setup a channel for other components to check whether or not
+	//the policy is completed. Possibly designed for the none loop policy.
+	Done() chan bool
+
+	//Evaluate the policy based on its definition and return the result via
+	//result channel. Policy is enabled after it is evaluated.
+	//Make sure Evaluate is idempotent, that means one policy can be only enabled
+	//only once even if Evaluate is called more than one times.
+	Evaluate() chan EvaluationResult
```
reasonerjt: may return and chan and an error?

src/common/scheduler/scheduler.go
```diff
@@ -0,0 +1,258 @@
...
+	statTaskFail         = "Task Fail"
+)
+
+//StatItem is defined for the stat metrics.
+type StatItem struct {
+	//Metrics catalog
+	Type string
+
+	//The stat value
+	Value uint32
```
reasonerjt: This is always 1, if no need, shall we remove this field?

src/common/scheduler/scheduler_store.go
```diff
@@ -0,0 +1,140 @@
...
+type ConcurrentStore struct {
+	//Read-write mutex to synchronize the data map.
+	mutex *sync.RWMutex
+
+	//Map used to keep the policy list.
+	data map[string]*Watcher
+}
+
+//NewConcurrentStore is used to create a new store and return the pointer reference.
+func NewConcurrentStore(initialSize uint32) *ConcurrentStore {
```
reasonerjt: is the initsize necessary?

src/common/scheduler/scheduler_store.go
```diff
@@ -0,0 +1,140 @@
...
+	GetAll() []*Watcher
+
+	//Clear store.
+	Clear()
+}
+
+//ConcurrentStore implements Store interface and supports concurrent operations.
+type ConcurrentStore struct {
+	//Read-write mutex to synchronize the data map.
+	mutex *sync.RWMutex
```
reasonerjt: Use a anonymous " *sync.RWMutex" may be more idiomatic, so you only need 
```
cs.Lock()....
```
instead of cs.mutex.Lock()

src/common/scheduler/task/scan_all_task.go
```diff
@@ -0,0 +1,21 @@
...
+//ScanAllTask is task of scanning all tags.
+type ScanAllTask struct{}
+
+//NewScanAllTask is constructor of creating ScanAllTask.
+func NewScanAllTask() *ScanAllTask {
+	return &ScanAllTask{}
+}
+
+//TaskName returns the name of the task.
+func (sat *ScanAllTask) TaskName() string {
```
reasonerjt: TaskName is a bit repetitive, let's call it "Name()"

src/common/scheduler/scheduler.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+}
+
+//Start the scheduler damon.
+func (sch *Scheduler) Start() {
+	if !sch.stopped {
+		return
+	}
+	go func() {
+		defer func() {
```
reasonerjt: Is it needed? since all the panic should have been caught in the layer of Watcher.

src/common/scheduler/scheduler.go
```diff
@@ -0,0 +1,258 @@
...
+		return errors.New("Policy should be assigned a name")
+	}
+
+	tasks := scheduledPolicy.Tasks()
+	if tasks == nil || len(tasks) == 0 {
+		return errors.New("Policy must attach task(s)")
+	}
+
+	if sch.policies.Exists(scheduledPolicy.Name()) {
+		return errors.New("Duplicated policy")
```
reasonerjt: Let's put the name in the error message.

src/common/scheduler/scheduler.go
```diff
@@ -0,0 +1,258 @@
...
+		unscheduleQueue: usq,
+		statChan:        stChan,
+		terminateChan:   tc,
+		stats: &StatSummary{
+			PolicyCount:    0,
+			Tasks:          0,
+			CompletedTasks: 0,
+			TasksWithError: 0,
+		},
+		stopped: true,
```
reasonerjt: In watcher,  isRunning is used , but in schedule it uses isStopped, maybe we should be more consistent here?

src/common/scheduler/watcher.go
```diff
@@ -0,0 +1,128 @@
...
+		for {
+			select {
+			case <-evalChan:
+				{
+					//Start to run the attached tasks.
+					for _, t := range pl.Tasks() {
+						go func(tk task.Task) {
+							defer func() {
+								if r := recover(); r != nil {
+									log.Errorf("Runtime error in task execution:%s\n", r)
```
reasonerjt: generate a statItem?

src/common/scheduler/watcher.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+//Start the running.
+func (wc *Watcher) Start() {
+	if wc.isRunning {
+		return
+	}
+
+	if wc.p == nil {
+		return
```
reasonerjt: In this case, wc's  status is still "running", which it shouldn't be.

src/common/scheduler/watcher.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	go func(pl policy.Policy) {
```
reasonerjt: policy and wc is 1-1 mapping, so probably don't need this parm?

src/common/scheduler/watcher.go
```diff
@@ -0,0 +1,128 @@
...
+				return
+			}
+		}
+	}(wc.p)
+
+	wc.isRunning = true
+}
+
+//Stop the running.
+func (wc *Watcher) Stop() {
```
reasonerjt: if this can only be called by scheduler, make it a package-private func

src/common/scheduler/watcher.go
```diff
@@ -0,0 +1,128 @@
...
+						}
+					}
+				}
+			case <-done:
+				{
+					//Policy is automatically completed.
+					wc.isRunning = false
+
+					//Report policy change stats.
+					wc.doneChan <- wc.p.Name()
```
reasonerjt: is it possible the doneChan is nil?

src/common/scheduler/policy/alternate_policy.go
steven-zou: Sure!

src/common/scheduler/policy/alternate_policy.go
steven-zou: The reason why implement it like this is to provide flexible compatibilities for future possible more complicated policy. This policy can support 'hourly','daily','weekly' even 'yearly'. It can also supports **ly in a specified scope. As the logic is very similar, doing this can avoid redevelopment. So I insist suggesting to keep this policy as it is.

src/common/scheduler/policy/evaluation_result.go
steven-zou: Only the current daily policy implementation does not handle the errors and results. Other policies might need this channel to pass the evaluation result to the tasks and send async errors to the upper module. We need to consider these scenarios.

I have considered your suggestion. Return separate chan for result and error.  I just don't want to make the return signature too complicated. If you think separate ones are better, I can refine it.

src/common/scheduler/scheduler.go
steven-zou: Current situation is 1, not for other future possible scenarios. Let's keep it as it is.

src/common/scheduler/scheduler_store.go
steven-zou: Not really, set a proper initial size can avoid reallocation action which costs efforts.

src/common/scheduler/scheduler_store.go
steven-zou: Good

src/common/scheduler/task/scan_all_task.go
steven-zou: Ok, call Name()

src/common/scheduler/scheduler.go
steven-zou: Tried, seems panic in the sub goroutine might not be caught by upper component. Besides double confirmation, it can also catch the panic errors from itself.

src/common/scheduler/scheduler.go
steven-zou: Sure!

src/common/scheduler/scheduler.go
steven-zou: Ok, to keep consistent, rename it to IsRunning

src/common/scheduler/watcher.go
steven-zou: Good

src/common/scheduler/watcher.go
steven-zou: No, this is the start method, the wc should not be running, it's ready to run. Then we check if it bound policy is valid.

src/common/scheduler/watcher.go
steven-zou: It's fine, just different Closure parameter binding method, this one might be more safer. No influence, let's just keep it as it is.

src/common/scheduler/watcher.go
steven-zou: Might be called by the upper component. In our scenario, it will not be called once the scheduler is started. Just keep it as it is.

src/common/scheduler/watcher.go
steven-zou: If it is created by scheduler, then must not be null. If created by other, it can not be guaranteed. To keep safer, let's add guard code.

src/common/scheduler/policy/evaluation_result.go
reasonerjt: But currently in watcher.go around line 60, it's not handling the Result, are you saying it should handle it in future?

The concern is, for example if someone implements his own Policy and return some error in the result send it in channel, it will be useless because the framework doesn't handle it...  So this type is misleading.

src/common/scheduler/policy/alternate_policy.go
```diff
@@ -0,0 +1,185 @@
...
+	//The interval of executing attached tasks.
+	Duration time.Duration
+
+	//The execution time point of each turn
+	//It's a number to indicate the seconds offset to the 00:00 of UTC time.
+	OffsetTime int64
+
+	//Time should be later than start time.
+	//If set <=0 value, no limitation.
+	StartTimestamp int64
```
reasonerjt: use time.Time?

src/common/scheduler/policy/alternate_policy.go
```diff
@@ -0,0 +1,185 @@
...
+
+	//Time should be earlier than end time.
+	//If set <=0 value, no limitation.
+	EndTimestamp int64
```
reasonerjt: use time.Time?

src/common/scheduler/scheduler.go
reasonerjt: But why itself can panic?  Only the methods in task and policy can introduce new panic as they can be extended.

src/common/scheduler/watcher.go
reasonerjt: yep,ok

src/common/scheduler/watcher.go
reasonerjt: ok

PR #2706, opened 2017-07-05
[BAT]add manage project member

sigsbee:


sigsbee:
close for there is another pr on master

PR #2721, opened 2017-07-07
Clair photon

wy65701436:


reasonerjt:
The change looks good, but have we verified everything works as expected? (including Clair APIs, updaters, notifiers...)

reasonerjt:
btw please squash the commits.

PR #2725, opened 2017-07-07
Return real status code returned by admiral APIs

ywk253100:


src/common/security/admiral/authcontext/authcontext.go
```diff
@@ -202,7 +203,10 @@ func send(client *http.Client, req *http.Request) (*AuthContext, error) {
 	}
 
 	if resp.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("unexpected status code: %d %s", resp.StatusCode, string(data))
+		return nil, &errwrapper.Error{
```
reasonerjt: let's rename it to HTTPError to be more explicit.

PR #2730, opened 2017-07-09
restrict scan all API, it should be called once every a period of time

reasonerjt:


src/ui/api/repository.go
```diff
@@ -745,6 +745,12 @@ func (ra *RepositoryAPI) ScanAll() {
 		ra.HandleForbidden(ra.SecurityCtx.GetUsername())
 		return
 	}
+
+	if !utils.ScanAllMarker().Mark() {
+		log.Warningf("There is a scan all scheduled at: %v, the request will not be processed.", utils.ScanAllMarker().Next())
+		ra.RenderError(http.StatusPreconditionFailed, "Unable handle frequent scan all requests")
+	}
```
ywk253100: Need return

src/ui/service/notifications/clair/handler.go
```diff
@@ -93,7 +94,7 @@ func (h *Handler) Handle() {
 			}
 		}
 	}
-	if rescanTimer.needReschedule() {
+	if utils.ScanOverviewMarker().Mark() {
```
ywk253100: The timer isn't needed anymore.

PR #2741, opened 2017-07-11
read last update from clair db, return in system info

reasonerjt:


src/common/dao/clair/base.go
```diff
@@ -0,0 +1,69 @@
...
+		s, ok := list[0].(string)
+		if !ok { // shouldn't be here.
+			return 0, fmt.Errorf("The value: %v, is non-string", list[0])
+		}
+		res, err := strconv.ParseInt(s, 0, 64)
+		if err != nil { //shouldn't be here.
+			return 0, err
+		}
+		return res, nil
+	}
```
ywk253100: Need to check whether the num > 1

PR #2777, opened 2017-07-13
provide default timestamp for all distros in system info api

reasonerjt:


src/common/utils/clair/client.go
```diff
@@ -47,6 +47,22 @@ func NewClient(endpoint string, logger *log.Logger) *Client {
...
+	resp, err := c.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+	b, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != http.StatusOK {
```
ywk253100: Should use expectedStatus passed as the parameter instead of http.StatusOK?

PR #2778, opened 2017-07-13
Support backend server pagination in repo list

steven-zou:
Support backend server pagination in repo list


reasonerjt:
Please make sure the commits are squashed before merging.

Issue #2564, opened 2017-06-19
project can not created when click ok button immediately


sigsbee:
v1.1.2-294-g6ee631d
project can not created when click ok button immediately

steven-zou:
Fixed by @pengpengshui , code merged into master. Close issue.

Issue #2758, opened 2017-07-12
make project public or private will see a bad request


sigsbee:
v1.1.2-423-g0b28203
![ppppp](https://user-images.githubusercontent.com/23101283/28107862-0cbb64c6-671c-11e7-9cd7-af82ccffb9f2.PNG)

steven-zou:
Fixed by @pengpengshui , code merged into master. Close issue.

Issue #2552, opened 2017-06-16
new replication rule window button not properly grayed out


hainingzhang:
when creating a new replication rule, 
test connection button should be grayed out when there is no endpoint selected or input.

When testing connection, "test connection" and "cancel" button should be grayed out.

Or "Cancel" can be enabled, but should really delete the connection, not just hiding the dialog box. 

Same to the create endpoint dialog.

steven-zou:
Fixed by @pengpengshui , code merged into master. Close issue.

PR #2779, opened 2017-07-14
Fix the issue #2564,#2552 and #2758

pengpengshui:
when verify project name,button turn gray 
 connect button grayed out when no select value an not checked


reasonerjt:
Please make sure the commits are squashed before merging.

src/ui_ng/src/app/project/project.component.ts
```diff
@@ -184,7 +183,7 @@ export class ProjectComponent implements OnInit, OnDestroy {
         response => {
           this.messageHandlerService.showSuccess('PROJECT.TOGGLED_SUCCESS');
           this.statisticHandler.refresh();
-          this.getProjects("", this.currentFilteredType - 1);
+          this.getProjects("");
```
steven-zou: Please consider the filter type here.

PR #2786, opened 2017-07-14
Clair offline data update guide

yixingjia:


docs/clair_offlinedata_guide.md
```diff
@@ -0,0 +1,64 @@
...
+    ```
+    e. The update finished indicate that Clair has finished an vulnerability update round. You need to check that logs above it to make sure all the endpoints are update correctly.
+
+### Data dump
+
+A. Login into the PostgreSQL container of Clair.
+* $>docker exec -it clair-db bash
+
+B. Dump the Clair vulnerability database by the follow command
+* $> pg_dump -U postgres -a -t feature -t keyvalue -t namespace -t schema_migrations -t vulnerability -t vulnerability_fixedin_feature > vulnerability.sqll
```
reasonerjt: would be possible to run `docker exec clair-db pg_dump ....` such that the sql can be dumped to a file on host?

docs/clair_offlinedata_guide.md
```diff
@@ -0,0 +1,64 @@
...
+* $> pg_dump -U postgres -c -s > clear.sql
+
+C. Collect the offline data
+* Exit the container and back to the host where the container running. (assume you are using aufs as storage driver)
+* cd /var/lib/docker/aufs/mnt
+* find . -name clear.sql
```
reasonerjt: see previous comment, if the .sql can be dumped to host this is not needed.

docs/clair_offlinedata_guide.md
```diff
@@ -0,0 +1,64 @@
...
+* find . -name vulnerability.sql
+* Copy the above two file to the host where Harbor is running.
+
+### Back Up Clair DB
+A. Before update the offline data, user are strongly suggested to backup their Clair db.
+* pg_dump -U postgres -c > all.sql
+
+### Update Clair DB
+A. Copy the vulnerability.sql and clear.sql to the clair-db container which you want to update.
```
reasonerjt: See the first comment, it should be possible to use a file on the host as input of pg_dump in container.

Issue #2795, opened 2017-07-17
unable to create project.


wy65701436:
the save button is not clickable in the create project panel.

steven-zou:
Fixed by @pengpengshui , code merged into master.  #2797

PR #2797, opened 2017-07-17
fix #2662,#2795

### LINKED ISSUE #2662: Left nav loses highlighting when in Replication -> Replication Rule
when navigating to admin replication, the highlight is lost when selecting the second tab (Replication Rule) see images. (using https://10.112.122.204/harbor/replications/rules)

screen shot 2017-06-28 at 10 45 44 am

screen shot 2017-06-28 at 10 45 53 am

pengpengshui:


src/ui_ng/src/app/project/create-project/create-project.component.ts
```diff
@@ -111,7 +111,7 @@ export class CreateProjectComponent implements AfterViewChecked, OnInit, OnDestr
   }
 
   onSubmit() {
-    this.btnIsOk=false;
+    this.isSubmitValid=false;
```
steven-zou: if(this.onGoing){
  return;
}

this.onGoing = true;

steven-zou: .subscribe(
  ....

  response=> this.onGoing = false;

  error=> this.onGoing = false;
)

PR #2798, opened 2017-07-17
The password to access clair db can be configured in harbor.cfg

reasonerjt:
The password to access clair db can be configured in harbor.cfg, auto-scan if clair-db is not ready

### DIFF
--- a/src/ui/service/notifications/registry/handler.go
+++ b/src/ui/service/notifications/registry/handler.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"github.com/vmware/harbor/src/common/dao"
+	clairdao "github.com/vmware/harbor/src/common/dao/clair"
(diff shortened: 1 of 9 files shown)


src/ui/service/notifications/registry/handler.go
```diff
@@ -105,8 +106,14 @@ func (n *NotificationHandler) Post() {
 			}()
 
 			go api.TriggerReplicationByRepository(pro.ProjectID, repository, []string{tag}, models.RepOpTransfer)
+
 			if autoScanEnabled(project) {
```
wy65701436: why not pass pro(GlobalProjectManager) to autoScanEnabled?

reasonerjt: This will be tracked by another issue.

PR #2799, opened 2017-07-17
Fix issue of detecting configuration changes

steven-zou:


src/common/notifier/config_watcher.go
```diff
@@ -0,0 +1,47 @@
...
+				DailyTime: 0,
+			}
+
+			if t, yes := policyCfg.Parm["daily_time"]; yes {
+				if reflect.TypeOf(t).Kind() == reflect.Int {
+					policyNotification.DailyTime = (int64)(t.(int))
+				}
+			}
+
+			return Publish(ScanAllPolicyTopic, policyNotification)
```
ywk253100: Maybe we need to check whether the config is equal to the old value

src/common/scheduler/scheduler.go
```diff
@@ -263,3 +268,12 @@ func (sch *Scheduler) IsRunning() bool {
 func (sch *Scheduler) HasScheduled(policyName string) bool {
 	return sch.policies.Exists(policyName)
 }
+
+//GetPolicy is used to get related policy reference by its name.
+func (sch *Scheduler) GetPolicy(policyName string) policy.Policy {
+	if sch.policies.Exists(policyName) {
```
reasonerjt: Is it thread-safe?

src/common/notifier/scan_policy_notitification_handler.go
```diff
@@ -53,20 +57,61 @@ func (s *ScanPolicyNotificationHandler) Handle(value interface{}) error {
 	hasScheduled := scheduler.DefaultScheduler.HasScheduled(alternatePolicy)
 	if notification.Type == PolicyTypeDaily {
```
reasonerjt: What if just unschedule and reschedule without having to do the compare?

src/common/notifier/config_watcher.go
```diff
@@ -0,0 +1,47 @@
...
+
+//WatchConfigChanges is used to watch the configuration changes.
+func WatchConfigChanges(cfg map[string]interface{}) error {
+	if cfg == nil {
+		return errors.New("Empty configurations")
+	}
+
+	//Currently only watch the scan all policy change.
+	if v, ok := cfg[ScanAllPolicyTopic]; ok {
+		if reflect.TypeOf(v).Kind() == reflect.Map {
```
reasonerjt: I believe you can simply marshal v to a []byte and unmarshal it to policyCfg?

reasonerjt:
Let's discuss it...

src/common/notifier/scan_policy_notitification_handler.go
```diff
@@ -53,20 +57,61 @@ func (s *ScanPolicyNotificationHandler) Handle(value interface{}) error {
...
+		if pl := scheduler.DefaultScheduler.GetPolicy(alternatePolicy); pl != nil {
+			if reflect.TypeOf(pl).Kind() == reflect.Ptr &&
+				reflect.TypeOf(pl).String() == "*policy.AlternatePolicy" {
+				spl := pl.(*policy.AlternatePolicy)
+				plConfig := spl.GetConfig()
+				if plConfig != nil {
+					if plConfig.OffsetTime != notification.DailyTime {
+						//Parameter changed.
+						//Unschedule policy.
+						if err := scheduler.DefaultScheduler.UnSchedule(alternatePolicy); err != nil {
```
reasonerjt: Unschedule should be synchrounous

src/common/scheduler/policy/alternate_policy.go
```diff
@@ -152,3 +155,25 @@ func (alp *AlternatePolicy) Evaluate() (<-chan bool, error) {
...
+	if p == nil {
+		return false
+	}
+
+	if reflect.TypeOf(p).Kind() != reflect.Ptr ||
+		!strings.Contains(reflect.TypeOf(p).String(), "AlternatePolicy") {
+		return false
+	}
+
+	pl := p.(*AlternatePolicy)
```
reasonerjt: pl, ok := p.(*AlternatePolicy); !ok.....

src/common/scheduler/scheduler.go
```diff
@@ -254,7 +236,17 @@ func (sch *Scheduler) UnSchedule(policyName string) error {
...
+	//Find the watcher.
+	watcher := sch.policies.Remove(policyName)
+	if watcher != nil && watcher.IsRunning() {
+		watcher.Stop()
+
+		//Update stats and log info.
+		log.Infof("Policy %s is unscheduled", policyName)
+		sch.statChan <- &StatItem{statUnSchedulePolicy, 1, nil}
+	} else {
+		return errors.New("Inconsistent policy status")
```
reasonerjt: probably log a warning saying the watcher of policy:%s is not running, something maybe wrong, and return a nil?
because the same policy can be rescheduled.

src/common/utils/utils.go
```diff
@@ -129,45 +130,25 @@ func ParseTimeStamp(timestamp string) (*time.Time, error) {
...
-	structValue := reflect.ValueOf(object).Elem()
-
-	structFieldValue := structValue.FieldByName(field)
-	if !structFieldValue.IsValid() {
-		return fmt.Errorf("No such field: %s in obj", field)
-	}
-
-	if !structFieldValue.CanSet() {
-		return fmt.Errorf("Cannot set value for field %s", field)
+	if reflect.TypeOf(values).Kind() != reflect.Map {
```
reasonerjt: Maybe unnecessary but let's change it in future...

PR #2802, opened 2017-07-18
fix #2775 about create-rule modal connect test issue

### LINKED ISSUE #2775: Ping the existing endpoint doesn't work in integration mode
The request GET /hbr-api/targets/ping is sent, it should be GET /hbr-api/targets/{endpoint_id}/ping, the endpoint_id is missing.

pengpengshui:


src/ui_ng/lib/src/create-edit-rule/create-edit-rule.component.ts
```diff
@@ -144,7 +144,9 @@ export class CreateEditRuleComponent implements AfterViewChecked {
   }
   get connectAbled():boolean{
     return !this.createEditRule.endpointId &&  !this.isCreateEndpoint;
-
+  }
+  get hasConnectData():boolean{
```
steven-zou: The name of method is not matched your expression.

has, but you use "!" that means not existing

PR #2822, opened 2017-07-20
update swagger doc

reasonerjt:


docs/swagger.yaml
```diff
@@ -2022,58 +2145,64 @@ definitions:
   TopRepo:
```
wy65701436: please fix all warnings raised by the online editor.  Definition was declared but never used in this document.

PR #2827, opened 2017-07-20
Fix replicate issue

ywk253100:


src/common/security/admiral/context.go
```diff
@@ -64,6 +64,11 @@ func (s *SecurityContext) IsSysAdmin() bool {
 	return s.ctx.IsSysAdmin()
 }
 
+// IsSolutionUser ...
+func (s *SecurityContext) IsSolutionUser() bool {
```
reasonerjt: Let's make it "GetSolutionUser" or "SolutionUser" and return a customized type. so we can handle request from different components differently.

ywk253100: We can distinguish them by GetUsername.

PR #2830, opened 2017-07-20
provide api to show log of scan job

reasonerjt:


src/ui/api/scan_job.go
```diff
@@ -0,0 +1,68 @@
...
+		sj.CustomAbort(http.StatusBadRequest, "ID is invalid")
+	}
+	sj.jobID = id
+
+	data, err := dao.GetScanJob(id)
+	if err != nil {
+		log.Errorf("Failed to load job data for job: %d, error: %v", id, err)
+		sj.CustomAbort(http.StatusInternalServerError, "Failed to get Job data")
+	}
+	projectName := strings.SplitN(data.Repository, "/", 2)[0]
```
ywk253100: Use ParseRepository in common/utils package

reasonerjt: Will address this with a follow up PR.

PR #2855, opened 2017-07-24
build registry

wy65701436:


make/photon/registry/Dockerfile_Golang
```diff
@@ -0,0 +1,10 @@
+FROM golang:1.7.3
```
reasonerjt: This is not a golang image, so let's change the name to Dockerfile.binary

make/photon/registry/builder
```diff
@@ -0,0 +1,62 @@
...
+
+echo "Build registry binary success, then to build photon image..."
+cd $cur
+echo $PHOTONIMAGE
+docker build -f Dockerfile -t $PHOTONIMAGE .
+
+rm -rf $TEMP
+rm -rf binary
+
+echo 'Push image to server.'
```
reasonerjt: I think it pushed image to docker hub?

make/photon/registry/builder
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+
+set +e
+
+echo "Usage: #./builder [registry version] [photon tag] [registry username] [registry password]"
```
reasonerjt: I think it's docker hub?

make/photon/registry/builder
```diff
@@ -0,0 +1,62 @@
...
+cd `dirname $0`
+cur=$PWD
+
+# the temp folder to store distribution source code...
+TEMP=`mktemp -d /$TMPDIR/distribution.XXXXXX`
+git clone -b $VERSION https://github.com/docker/distribution.git $TEMP
+
+echo 'build the registry binary bases on the golang:1.7.3...'
+cp Dockerfile_Golang $TEMP
+docker build -f $TEMP/Dockerfile_Golang -t registry-golang $TEMP
```
reasonerjt: should it be a docker run or docker build?

wy65701436: it's should be the docker build.

make/photon/registry/builder
wy65701436: yes, it will push the image to docker hub. What's your suggestion?

reasonerjt: I'm fine with that, so let's clarify in the message.

make/photon/registry/builder
reasonerjt: This is a little bit, I was thinking we should provide a builder image, and build whatever source via "docker run" put the binary in volume?

reasonerjt:
An overall comment, this PR may not have to be merged in 1.2 timeframe as it does not block anything.

PR #2856, opened 2017-07-24
Provide API for scanning images under a project

reasonerjt:


src/ui/utils/utils.go
```diff
@@ -42,33 +42,52 @@ func ScanAllImages() error {
...
+}
+
+// ScanImagesByProjectName scans all images under a projet, the whole process will move on if failed to submit any job of a single image.
+func ScanImagesByProjectName(project string) error {
+	regURL, err := config.RegistryURL()
+	if err != nil {
+		log.Errorf("Failed to load registry url")
+		return err
+	}
+	repos, err := dao.GetRepositoryByProjectName(project)
```
ywk253100: The method can not be called under integration mode, getting repositories by project ID instead.

src/ui/api/repository.go
```diff
@@ -741,23 +741,35 @@ func (ra *RepositoryAPI) ScanAll() {
...
-		ra.HandleForbidden(ra.SecurityCtx.GetUsername())
-		return
-	}
-
-	if !utils.ScanAllMarker().Check() {
-		log.Warningf("There is a scan all scheduled at: %v, the request will not be processed.", utils.ScanAllMarker().Next())
-		ra.RenderError(http.StatusPreconditionFailed, "Unable handle frequent scan all requests")
-		return
-	}
+	projectName := ra.GetString("project_name")
```
ywk253100: I think project_id is better than project_name.

src/ui/utils/utils.go
reasonerjt: ok

PR #2906, opened 2017-07-28
Apply security filter to certain patterns and enable basic auth for deleting repo

ywk253100:
1. only apply security filter to /api/ and /service/ 
2. support basic auth for deleting repository and tag in integration mode

### DIFF
--- a/src/ui/filter/security.go
+++ b/src/ui/filter/security.go
@@ -50,11 +50,6 @@ const (
 
 var (
 	reqCtxModifiers []ReqCtxModifier
-	// the pattern needs security filter
-	reqPatterns = []string{
-		"/api/",
-		"/service/",
-	}
 	// basic auth request context modifier only takes effect on the patterns
 	// in the slice
 	basicAuthReqPatterns = []*pathMethod{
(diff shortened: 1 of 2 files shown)


reasonerjt:
Is it for fixing some bug?  

The pattern and the actual "API" func is still a little disconnected, ideally we can invent a "API" struct which contains the pattern, the middle ware and the actual logic of the API.

And is there reason to support basic auth in integration mode?  I assume if there's any basic auth required for Harbor API the client should access Admiral instead of Harbor?

ywk253100:
2) is for fixing bug. Basic auth is needed for requesting token, creating projects and deleting repository and tag. The requesting token is for docker client, creating projects and deleting repo is for replication.

src/ui/filter/security.go
```diff
@@ -78,6 +113,11 @@ func SecurityFilter(ctx *beegoctx.Context) {
 		return
 	}
 
+	if !filterReq(req) {
```
reasonerjt: As discussed, we don't need this filter for now as the ReqCtxModifier has been doing the filter already, we'll consider optimize when we see a perf issue.

Issue #2762, opened 2017-07-12
Data Race issue in the policy scheduler.


wy65701436:
https://travis-ci.org/vmware/harbor/builds/252716791?utm_source=github_status&utm_medium=notification

For the go data race document:
https://golang.org/doc/articles/race_detector.html

steven-zou:
Code merged, close this issue.

PR #2945, opened 2017-08-02
Fix data race issues of go sources

steven-zou:
Fix issue #2762 
Fix data race issues of go sources


reasonerjt:
Let's simplify by adding lock to func, given some func should be accessed by only one thread.  think "synchrozied" in java.

PR #2960, opened 2017-08-03
Fix performance issue and connection leakage

reasonerjt:
1)Use the layer digest as the name of the layer when calling Clair API
2)Add header "Connection":"close"


reasonerjt:
closing the PR as seen some issue of dup layer name.  May need to reconsider the way to generate the unique and re-usable layer name.

PR #2969, opened 2017-08-04
fix #2893 about add tooltips title

### LINKED ISSUE #2893: [UI] The scan button should shows at conspicuous place.
image

pengpengshui:


src/ui_ng/lib/src/vulnerability-scanning/result-tip.component.ts
```diff
@@ -35,6 +35,22 @@ export class ResultTipComponent implements OnInit {
         }
     };
 
+    get scanLevel(){
+        let level;
```
steven-zou: should explicitly declare the type of the variable and give its initial value

steven-zou: Use VulnerabilitySeverity

Issue #2936, opened 2017-08-01
Harbor proxy cause download speed degrade to half compare with directly connection


yixingjia:
user curl to download a layer from harbor it takes about 4 minutes.

100 93.7M  100 93.7M    0     0   397k      0  0:04:01  0:04:01 --:--:--  441k

Setup a registry and download the same layer use curl it only take 1min48 seconds.

100 93.7M  100 93.7M    0     0   885k      0  0:01:48  0:01:48 --:--:-- 2705k

if seems the proxy has largely degrade the registry download speed.

reasonerjt:
It's true the proxy will impact the speed, I'll investigate.

reasonerjt:
There are two factors to impact the speed:
1)nginx: it's buffer is turned off, we'll turn it on and tune the buffer number/size to improve
2)interceptors: we use httputils.reverseproxy which has bad reputation in terms of performance.

Enabling buffer on nginx is the low hanging fruit,  but let's keep it open for further perf improvement.

reasonerjt:
Closing due to inactivity, we're grooming issues around perf for a while and this issue didn't re-appear in our radar.

PR #2973, opened 2017-08-04
Enable buffer on nginx

reasonerjt:
to accelerate the speed to download image.

to mitigate the issue in #2936


reasonerjt:
@wy65701436 I haven't done a strict testing, but by issuing a curl to download a 100MB layer:
```
curl -H "authorization: Bearer ${token}" http://xxxxxx/v2/llibrary/chrome/blobs/sha256:d05a36469a0f1da76c1afb951ba6ae9ac2ce71f04a2479980f7b114af6406c24 -o /dev/null
```
 approximately the speed increased from below 100MB/s to 200+MB/s

PR #2988, opened 2017-08-08
update migrator

wy65701436:


tools/migration/run.sh
```diff
@@ -72,8 +72,8 @@ case $key in
 up|upgrade)
     VERSION="$2"
     if [[ -z $VERSION ]]; then
-        VERSION="head"
-        echo "Version is not specified. Default version is head."
+        echo "Exit, please specify version and run again."
+        exit 0
```
andrewtchin: If the script has an error, exit code should be nonzero

PR #2998, opened 2017-08-08
Fix harborclient module logging

cpeel:
Per python logging best practices, modules shouldn't set the global logging level and should log with a named logger matching the module. This allows the application to control the logging.


cpeel:
Sorry it took so long to get the CLA taken care of.

I have a few more of these types of changes to harbor-py but will open them in series since they build on top of each other.

cpeel:
Ping?

PR #2999, opened 2017-08-08
Updated tests to be more resilient to failures

srm912:
This patch is aimed at making the *dao* tests more resistant to failures. The motivation behind this patch was to make the tests more resilient to code panics or exceptions which does not allow the post test cleanup to complete.
This would also make sure that the cleanup tasks are fired irrespective of the test run output, thus eliminating the need for manual cleanups.

### DIFF
--- a/src/common/dao/dao_test.go
+++ b/src/common/dao/dao_test.go
@@ -15,6 +15,7 @@
 package dao
 
 import (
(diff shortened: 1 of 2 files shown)


srm912:
Abandoned because of CLA instructions, will open a new one.

PR #3001, opened 2017-08-08
Updated tests to be more resilient to failures

srm09:
This patch is aimed at making the dao tests more resistant to failures. The motivation behind this patch was to make the tests more resilient to code panics or exceptions which does not allow the post test cleanup to complete.
This would also make sure that the cleanup tasks are fired irrespective of the test run output, thus eliminating the need for manual cleanups.

### DIFF
--- a/src/common/dao/dao_test.go
+++ b/src/common/dao/dao_test.go
@@ -15,6 +15,7 @@
 package dao
 
 import (
+	"fmt"
 	"os"
 	"strconv"
 	"testing"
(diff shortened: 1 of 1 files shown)


srm912:
This PR is a WIP, I just wanted to get some initial thoughts whether this approach makes sense.

src/common/dao/dao_test.go
```diff
@@ -871,27 +834,27 @@ func TestAddProjectMember(t *testing.T) {
...
 	}
 	roles, err := GetUserProjectRoles(1, currentProject.ProjectID)
 	if err != nil {
-		t.Errorf("Error occurred in GetUserProjectRoles: %v", err)
+		t.Fatalf("Error occurred in GetUserProjectRoles: %v", err)
 	}
-	if roles[0].Name != "guest" {
-		t.Errorf("The user with ID 1 is not guest role after update, the acutal role: %s", roles[0].Name)
+	if assert.Len(t, roles, 1) {
+		assert.True(t, roles[0].Name != "guest",
```
ywk253100: Should be roles[0].Name == "guest". I think using assert.Equal() here is better.

ywk253100:
Closing due to inactivity.

Issue #2313, opened 2017-05-16
The pattern of updating project API


ywk253100:
Maybe we can change PUT /api/projects/:id([0-9]+)/publicity to PUT /api/projects/:id([0-9]+)?

srm912:
Does this mean that the publicity be toggled? So basically, public -> private and vice versa, right?

reasonerjt:
@srm912 thanks, please go ahead, and this issue will likely be deferred to post-1.2

ywk253100:
Closing due to moving public property into metadata and will provide API to update metadata.

PR #3002, opened 2017-08-08
Updated project publicity API

srm09:
This patch changes the toggle project publicity API to remove the
existing publicity being passed from the UI to the DAO layer.
This makes the API replicate the toggling functionality, by
switching it from public -> private and vice versa without the need
to know the existing publicity.

References #2313

### DIFF
--- a/docs/swagger.yaml
+++ b/docs/swagger.yaml
@@ -193,9 +193,8 @@ paths:
           description: 'Project contains policies, can not be deleted.'
         '500':
           description: Internal errors.
-  '/projects/{project_id}/publicity':
     put:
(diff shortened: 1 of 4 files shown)


ywk253100:
@srm912 Sorry for the late reply, maybe we didn't communicate this clearly in the issue #2313, we want to put the parameter "publicity" in the body rather than in the path just like the standard REST API does. So the request should be like this:
```
PUT /api/projects/{project_id}
request body:
{
        "publicity": true,
} 
```

ywk253100:
Closing due to inactivity.

PR #3037, opened 2017-08-11
[BAT] fix issues

wy65701436:


src/ui/proxy/interceptors.go
```diff
@@ -236,19 +316,35 @@ func matchNotaryDigest(img imageInfo) (bool, error) {
...
+				}
+			}
 		}
 	}
 	log.Debugf("image: %#v, not found in notary", img)
 	return false, nil
 }
 
+func isDigest(ref string) bool {
+	return strings.HasPrefix(ref, "sha256")
```
reasonerjt: Would be better to has prefix "sha256:" and the length is a constant

src/ui/proxy/interceptors.go
```diff
@@ -224,7 +304,7 @@ func (fu funnelHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	imgRaw := req.Context().Value(imageInfoCtxKey)
 	if imgRaw != nil {
 		log.Debugf("Return the original response as no the interceptor takes action.")
-		copyResp(rec, rw)
+		fu.next.ServeHTTP(rw, req)
```
reasonerjt: looking at the logic seems we can remove the funnel handler?

PR #3049, opened 2017-08-15
ui_ng fix

qinyuanbin:
I commit this pr for two problem I melt in page: project -> detail -> repositories
    
1. the cell of the subgrid are not aligned
2. the 'author' and 'version' cell of the subgrid is not big enough when my author name is an email, and version is something looks like '17.06.0-ce', it also cause a word wrap issue of the grid title when it is in chinese languge.

ps: hope you can understand what I wrote, I tried but poor in english :)

### DIFF
--- a/src/ui_ng/lib/src/tag/tag.component.css.ts
+++ b/src/ui_ng/lib/src/tag/tag.component.css.ts
@@ -32,6 +32,14 @@ export const TAG_STYLE = `
   background-color: #eee;
 }
 
+:host >>> .datagrid .datagrid-cell {
+  min-width: 96px;
+}
+
+:host >>> .datagrid .datagrid-body .datagrid-row:first-child .datagrid-cell {
+  min-width: auto;
+}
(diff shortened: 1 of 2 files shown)


steven-zou:
@qinershi，
  Thanks a lot for contributing code to Harbor. Just one question,  have you checked and verified the UI changes you made? Seems the widths of the columns are changed to be bigger a lot.

qinershi:
@steven-zou, 
Thank you for replying me.
You are right, it's too big, I have restore the width what it is before. 
And one more question, as the 'author' is come for username of the user, can harbor limits the length of the username when creating users? In this case the 'author' will not be too long for the cell.

steven-zou:
@qinyuanbin 

  Seems this PR is already outdated, I'll close it. If you still want to resize the column width, you can submit a new one. 
  BTW, we already have resized the column width,  please check the latest code before do additional changes. Thanks

PR #3101, opened 2017-08-22
Convert 500 error returned by Admiral to duplicate project error when creating duplicate project

ywk253100:


reasonerjt:
Let's add UT for DB project manager.
And at least cover the Admiral project manager error conversion logic.

Issue #3100, opened 2017-08-22
down load ca text display error


sigsbee:
v1.2.0-rc1-37-gf4a87f8

use self-signed https cert, download ca link has changed position,should not display here
![downloadcert](https://user-images.githubusercontent.com/23101283/29550466-7335af4c-8740-11e7-971c-aefe67ddc17b.PNG)

steven-zou:
Link removed

Issue #3097, opened 2017-08-21
ca cert download link should not use hardcode url link


steven-zou:
<a #certDownloadLink href="/api/systeminfo/getcert" target="_blank">{{'CONFIG.ROOT_CERT_LINK' | translate}}</a>

In VIC, the api endpoint is different and will not work anymore.

change it to a variable property reading endpoint from module configuration

steven-zou:
Related with  https://jira-hzn.eng.vmware.com/browse/VBV-1559

steven-zou:
Fixed. Close this issue.

PR #3102, opened 2017-08-22
Fix projetcId not ready and ca download link issues

steven-zou:
#3100 
#3097


src/ui_ng/lib/src/list-replication-rule/list-replication-rule.component.ts
```diff
@@ -64,64 +76,80 @@ export class ListReplicationRuleComponent implements OnInit {
...
   ngOnInit(): void {
-    this.projectScope = (!this.projectId);
-    this.retrieveRules();
+    //Global scope
+    if (!this.projectScope) {
+      this.retrieveRules();
+    }
+  }
+
+  ngOnChanges(changes: SimpleChanges): void {
```
pengpengshui: no need so much jugment ,"projectId only have undefined and 1 two values change

PR #3103, opened 2017-08-22
use $REGISTRY instead of $4

wdhif:
Just a quick fix for the pushimage.sh file.

The $REGISTRY variable was created but not used in the `docker login` instruction.


wy65701436:
@wdhif, please sign the agreement as above mentioned.

Issue #3061, opened 2017-08-16
[Feature Request] Harbor registry images hardening


wdhif:
Some companies need to be compliant with security standards as PCI-DSS. To do so, all the docker container running must be compliant with the [CIS Docker](https://www.cisecurity.org/benchmark/docker/).

To be compliant with the CIS Docker, the Harbor registry image should implement those basic changes :
- Read only, only use volumes
- Non root user for containers
- Add HEALTHCHECK to the images

I've run some tests on a development environment and I can contribute on this subject.

Here is the list of the current MR : 
- [ ] https://github.com/vmware/harbor/pull/3123
- [ ] https://github.com/vmware/harbor/pull/3110
- [ ] https://github.com/vmware/harbor/pull/3200
- [ ] https://github.com/vmware/harbor/pull/3263

reasonerjt:
By `images` do you mean the images of Harbor itself?

wdhif:
Yes, the images for Harbor itself and the docker-compose.yml for some options (like the `read_only: true` flag).

I work for a company that would like to use Harbor but also make it compliant with PCI-DSS standards. It might be a valuable addition to Harbor itself.

reasonerjt:
@wdhif
That would be helpful, but it would be helpful if we can take small steps for the refinement.

Appreciate your help!

wdhif:
OK glad to hear that, I'll start checking what can be done and I will open a PR if I get any results. I'll leave the issue open to reference it in a PR if necessary.

Also, are the Dockerfiles for the Harbor images in open source?

wdhif:
Hi @reasonerjt

Just to give you a little update. I currently have the read_only flag set for all images in the 3 docker-compose files, with the necessary volumes.

You can check the modifications here https://github.com/wdhif/harbor

I also had some question regarding the non-root user and the healthchecks, those modifications must be made in the Dockerfiles of each images.
I've found the Dockerfiles for all Harbor and Clair images but not for Notary, also, I don't really understand the file structure for the Dockerfiles, like 'common', 'dev', 'photon'. Do you have any information about this ?

Also feel free to comment about the work done so far.

Thanks, regards

wy65701436:
Hi wdhif, 

It's an historical issue, eventually all Harbor images will be based on photon. 
Please just refer to the Dockerfiles in the 'photon', and you do not need to care about the files in the 'dev', it just CI used temporary folder.

-Yan

wdhif:
[...]

PR #3110, opened 2017-08-23
read_only and volumes added for #3061

wdhif:
This add read_only flag for the images of Harbor, as described in #3061

- read_only and volumes (in Dockerfiles) for harbor
- read_only and volumes (in Compose file) for notary
- read_only and volumes (in Dockerfiles) for clair

The images will need to be rebuild and push with the new volumes, or the compose will fail.

### DIFF
--- a/make/common/db/Dockerfile
+++ b/make/common/db/Dockerfile
@@ -1,5 +1,8 @@
 FROM mysql:5.6.35
 
+VOLUME ["/tmp"]
+VOLUME ["/var/run/mysqld"]
+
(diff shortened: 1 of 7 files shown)


reasonerjt:
@wdhif thanks for the contribution, but per discussion we want to disable the "read_only" by default, and we will refine the docker files such that the containers will work when user enable this flag in docker-compose template.

reasonerjt:
closing as the requirement is addressed in other PRs.

PR #3112, opened 2017-08-24
Try to fix some wired permission error

yixingjia:
Move the temp buffer folder for upload/download to /tmp directory.


reasonerjt:
LGTM.
Let's try to apply this to some problematic env before merging.

PR #3119, opened 2017-08-25
block pull scan failed image

wy65701436:


reasonerjt:
@wy65701436 please handle the build failure.

PR #3123, opened 2017-08-25
healthcheck added for #3061

wdhif:
Healthcheck added as described in #3061 for :

- vmware/registry
- vmware/nginx-photon
- vmware/harbor-adminserver
- vmware/harbor-ui
- vmware/harbor-jobservice

It might be better to wait for harbor-db (since it's a mysql image it doesn't have curl) and harbor-log use rsyslog, which is UDP so I cant't use curl.

### DIFF
--- a/make/photon/registry/Dockerfile
+++ b/make/photon/registry/Dockerfile
@@ -10,6 +10,10 @@ COPY binary/registry /usr/bin
 RUN chmod u+x /usr/bin/registry
 
 VOLUME ["/var/lib/registry"]
+
+ENV PORT 5000
 EXPOSE 5000
+HEALTHCHECK CMD curl http://localhost:$PORT/v2 || exit 1
+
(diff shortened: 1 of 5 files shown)


wdhif:
We maybe want to wait for https://github.com/vmware/harbor/issues/2492

make/photon/registry/Dockerfile
```diff
@@ -10,6 +10,10 @@ COPY binary/registry /usr/bin
 RUN chmod u+x /usr/bin/registry
 
 VOLUME ["/var/lib/registry"]
+
+ENV PORT 5000
 EXPOSE 5000
+HEALTHCHECK CMD curl http://localhost:$PORT/v2 || exit 1
```
reasonerjt: I think you don' t need the env "PORT", just use 5000 instead.

reasonerjt:
Closing this PR as the requirement has been addressed in other PRs.

PR #3125, opened 2017-08-28
update error format

wy65701436:


src/ui/proxy/interceptors.go
```diff
@@ -341,13 +340,17 @@ func copyResp(rec *httptest.ResponseRecorder, rw http.ResponseWriter) {
...
-		Message: msg,
-		Code:    statusCode,
-		Details: msg,
+func marshalError(msg string) string {
+	var tmpErrs struct {
+		Errors []JSONError `json:"errors,omitempty"`
 	}
-	str, err := json.Marshal(je)
+	tmpErrs.Errors = append(tmpErrs.Errors, JSONError{
+		Code:    "UNKNOWN",
```
reasonerjt: Would be better set some meaningful, even though it can't be handled by vanilla docker currently?

reasonerjt: How'bout PROJECT_POLICY_VIOLATION ?

PR #3146, opened 2017-08-31
fix issue 2362

wy65701436:


tests/validatecontainers.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+set -e
+
+if docker ps --filter "status=restarting" | grep 'vmware'; then
```
reasonerjt: may better consider the failed ones.

PR #3149, opened 2017-09-04
support release build

wy65701436:


.drone.yml
```diff
@@ -51,20 +51,31 @@ pipeline:
...
+  notify-slack-on-pass:
+    image: plugins/slack
+    webhook: ${SLACK_URL}
+    username: drone
+    template: >
+      build https://ci.vcna.io/vmware/harbor/{{ build.number }} finished with a {{ build.status }} status, Please find logs at https://storage.googleapis.com/harbor-ci-logs/integration_logs_{{ build.number }}_{{ build.commit }}.zip
+    when:
+      repo: vmware/harbor
+      branch: [ master, release-*, refs/tags/* ]
+      status: success
```
reasonerjt: Let's simply use one target "notify slack" to notify both success and failure statuses.

PR #3200, opened 2017-09-12
replace ADD with COPY for photon/log

wdhif:
COPY should be used instead of ADD for the log Dockerfile.

https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy


reasonerjt:
Thanks @wdhif, due to conflict, I'll close this PR and make change in the code.

wy65701436:
@reasonerjt , should we close it?

PR #3255, opened 2017-09-20
Readjust package structure of project manager

ywk253100:


ywk253100:
OK, I'll change them in the later PR.

PR #3263, opened 2017-09-20
Non root user

wdhif:
This add the user-non-root for the following images :
- adminserver
- ui
- jobservice
- nodeclarity

As mentioned in https://github.com/vmware/harbor/issues/3061


reasonerjt:
This has to be pending because this change will requirement coordination when Harbor's integrated with other bits of vmware software.

make/dev/nodeclarity/Dockerfile
```diff
@@ -8,7 +8,14 @@ COPY make/dev/nodeclarity/entrypoint.sh /
 
 WORKDIR /harbor_resources
 
+RUN useradd -m -d /home/node -s /bin/bash node \
```
reasonerjt: This is not needed as this docker image is for building the ui code only, not for running.

reasonerjt:
@wdhif  we are in the middle of making update to all Dockerfiles, to avoid conflict, it would be helpful to make update to one Dockerfile in one PR, make sure things work and it will be merged.

Issue #3266, opened 2017-09-21
Move remote cert verification checkbox from system configuration to target level


ywk253100:

steven-zou:
@pengpengshui 

  Please coordinate with @ywk253100 to refactor the UI part. Thanks

PR #3339, opened 2017-09-28
fix issue about move remote cert verification checkbox  to target level  #3266

pengpengshui:


src/ui_ng/lib/src/create-edit-endpoint/create-edit-endpoint.component.html.ts
```diff
@@ -39,6 +39,10 @@ export const CREATE_EDIT_ENDPOINT_TEMPLATE: string = `
           <input type="password" class="col-md-8" id="destination_password" [disabled]="testOngoing" [readonly]="!editable" [(ngModel)]="target.password" size="20" name="password" #password="ngModel" (focus)="clearPassword($event)">
         </div>
         <div class="form-group">
+          <label for="destination_insecure" class="col-md-4 form-group-label-override">{{'CONFIG.VERIFY_REMOTE_CERT' | translate }}</label>
+          <clr-checkbox #insecure  class="col-md-8" name="insecure" id="destination_insecure" (clrCheckedChange)="insecureChange(insecure.checked)" [(ngModel)]="target.insecure"></clr-checkbox>
```
steven-zou: You have two-directions bound the variable 'target.insecure', why do you add event monitor "insecureChange"? I think these are some functions.

src/ui_ng/lib/src/create-edit-endpoint/create-edit-endpoint.component.ts
```diff
@@ -64,6 +64,7 @@ export class CreateEditEndpointComponent implements AfterViewChecked, OnDestroy
 
     endpointHasChanged: boolean;
     targetNameHasChanged: boolean;
+    checkboxHasChanged: boolean;
```
steven-zou: Non-meaningful variable, should be insecure etc.

src/ui_ng/lib/src/create-edit-endpoint/create-edit-endpoint.component.ts
```diff
@@ -91,7 +92,8 @@ export class CreateEditEndpointComponent implements AfterViewChecked, OnDestroy
                 (this.target.endpoint && this.target.endpoint.trim() !== "") ||
                 (this.target.name && this.target.name.trim() !== "") ||
                 (this.target.username && this.target.username.trim() !== "") ||
-                (this.target.password && this.target.password.trim() !== ""));
+                (this.target.password && this.target.password.trim() !== "")) ||
+                (this.target.insecure !== this.initVal.insecure);
```
steven-zou: Here is for create new, that means no old values, all are new values. No need to check if == with the original value, right?

src/ui_ng/lib/src/create-edit-endpoint/create-edit-endpoint.component.ts
```diff
@@ -289,6 +295,10 @@ export class CreateEditEndpointComponent implements AfterViewChecked, OnDestroy
             payload.password = this.target.password;
             delete payload.name;
         }
+        if (this.checkboxHasChanged) {
```
steven-zou: What does this logic mean?

PR #3352, opened 2017-10-05
Ignore generated files

everett-toews:


steven-zou:
Build failed, let's re-run the build and wait for the result.

PR #3354, opened 2017-10-06
Quote mysql params in alembic.tpl

tvon:
Previously, passwords with special characters (namely `%`) would result in an error.  Quoting the parameters should allow for these special characters.


tvon:
I'm kinda curious how coverage decreased by adding quotes around variables and removing some trailing whitespace.

tvon:
It was suggested (in Slack) that the params would need to be `uroquote()`'d in python to work like this, which means this PR won't actually do anything.

As such, I'm going to close this.

PR #3381, opened 2017-10-16
new: move remote cert verification checkbox #3266

pengpengshui:


src/ui_ng/lib/src/create-edit-endpoint/create-edit-endpoint.component.ts
```diff
@@ -289,6 +293,10 @@ export class CreateEditEndpointComponent implements AfterViewChecked, OnDestroy
             payload.password = this.target.password;
             delete payload.name;
         }
+        if (this.checkboxHasChanged) {
```
steven-zou: Password might be override

src/ui_ng/lib/src/create-edit-endpoint/create-edit-endpoint.component.spec.ts
```diff
@@ -21,6 +21,7 @@ describe('CreateEditEndpointComponent (inline template)', () => {
     "name": "target_01",
     "username": "admin",
     "password": "",
+    "insecure": false,
```
steven-zou: rename property 'insecure'

PR #3385, opened 2017-10-16
Simple code and fix typos.

xuri:
Simple code: omit nil check, `len()` for nil slices is defined as zero.


reasonerjt:
@xuri you need to fix the build break.

reasonerjt:
Could you squash the commits and refine the commit message, you can copy the description of the PR to the squashed commit in this case.

reasonerjt:
@xuri could you squash the commit?

PR #3409, opened 2017-10-19
Apply project level policies to standalone Harbor

ywk253100:
The following features are only enabled in integration mode, this commit moves
these to standalone Harbor:
 - Content trust policy: only signed images can be pulled
 - Vulnerability policy: only images whose severity is below the threshold can be pulled
 - Automatic scan policy: automatic scan pushed images


reasonerjt:
LGTM.
And Kudos to the commit message!

PR #3411, opened 2017-10-19
enable server end pagination for project list

steven-zou:
This commit is to support the server end pagination in the project list view. It is already merged into the master branch before. As we're planning to release a new version 1.2.x, use cherry-pick to add it in 1.2.0 branch.


reasonerjt:
As for the build break, could you update the tag of admiral in the script on 1.2.0 branch?

please refer to: https://github.com/vmware/harbor/blob/master/tests/admiral.sh

PR #3423, opened 2017-10-23
Update OSS in rsyslog images

yixingjia:
1> change to new photon base images
2> update OSS to latest

### DIFF
--- a/make/common/rsyslog/Dockerfile
+++ b/make/common/rsyslog/Dockerfile
@@ -1,4 +1,4 @@
-FROM vmware/photon:1.0-20170928
+FROM vmware/photon:1.0
 
 #base image for rsyslog base on photon
 
--- a/make/common/rsyslog/Dockerfile
+++ b/make/common/rsyslog/Dockerfile
@@ -1,9 +1,11 @@
-FROM library/photon:1.0
+FROM vmware/photon:1.0-20170928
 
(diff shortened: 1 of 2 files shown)


make/common/rsyslog/Dockerfile
```diff
@@ -1,9 +1,11 @@
-FROM library/photon:1.0
+FROM vmware/photon:1.0-20170928
```
reasonerjt: The bug of vmware/photon has been fixed, so you can use 1.0 now.

make/photon/log/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM vmware/rsyslog-photon:8.15.0
+FROM vmware/rsyslog-photon:8.15.0-6
```
reasonerjt: Shall we use stage build?
https://docs.docker.com/engine/userguide/eng-image/multistage-build/#name-your-build-stages

make/common/rsyslog/Dockerfile
yixingjia: Nice.

make/photon/log/Dockerfile
yixingjia: For this special case we only need to copy some env files instead of build/compile something.

Not sure what benefit will we get from the stage build?

PR #3436, opened 2017-10-23
Add project metadata API

ywk253100:
Project metadata API can be used to integrated with project management
service which can not provide all metadatas needed by Harbor.


src/ui/api/metadata.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	if project == nil {
+		m.HandleNotFound(fmt.Sprintf("project %d not found", id))
+		return
+	}
+
+	m.project = project
+
+	switch m.Ctx.Request.Method {
```
steven-zou: Why not use if

if !(m.Ctx.Request.Method == http.MethodGet && project.IsPublic){
  //do auth
}

if m.Ctx.Request.Method == other {
  //not allowed
}

src/ui/api/metadata.go
```diff
@@ -0,0 +1,245 @@
...
+		return
+	}
+
+	if len(ms) != 1 {
+		m.HandleBadRequest("invalid request: has no valid key/value pairs or has more than one valid key/value pairs")
+		return
+	}
+
+	name := ""
+	for name = range metas {
```
steven-zou: keys := reflect.ValueOf(a).MapKeys()
    strkeys := make([]string, len(keys))
    for i := 0; i < len(keys); i++ {
        strkeys[i] = keys[i].String()
    }
    fmt.Print(strings.Join(strkeys, ","))

PR #3441, opened 2017-10-23
python API: Work with LDAP-connected harbor

cpeel:
The code currently assumes that every instance of HarborClient will call login() before being destroyed. But login() is only needed for non-LDAP-connected harbor instances, and even then not for all APIs.

This updates the code to work with an LDAP-connected harbor instance and updates the README.md file and HarborClient class doc to clarify that.


clouderati:
Cleaning up old PRs and issues. If you're still interested in contributing please feel free to open a new PR. Thanks! :smile:

PR #3457, opened 2017-10-25
Rebuild Harbor DB docker image on top of Maria DB

reasonerjt:
This change reworked the vmware/harbor-db image to build it on top of
vmware/mariadb-photon.  Also made minor change in the entrypoint script
of mariadb image to execute upgrade script during bootstrap.


reasonerjt:
Drone CI has passed!
Travis CI seems to have issue to access docker hub, will wait a full day see if it will be fixed.

PR #3472, opened 2017-10-27
define filter interfaces for replication service including:interface Filter,Converter and FilterChain and other supporting models

steven-zou:


ywk253100:
Please open the pull request against branch replication_enhancemnet.

PR #3476, opened 2017-10-30
Update UI image oss package to latest

yixingjia:
Update base image to photon
Remove vim from the image

### DIFF
--- a/make/photon/ui/Dockerfile
+++ b/make/photon/ui/Dockerfile
@@ -1,7 +1,9 @@
-FROM library/photon:1.0
+FROM vmware/photon:1.0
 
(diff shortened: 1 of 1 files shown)


make/photon/ui/Dockerfile
```diff
@@ -1,7 +1,9 @@
-FROM library/photon:1.0
-
-RUN mkdir /harbor/
+FROM vmware/photon:1.0
 
+RUN tdnf distro-sync -y \
```
wy65701436: tdnf distro-sync || echo ?

@reasonerjt

reasonerjt: it's fine.

PR #3482, opened 2017-10-30
Update Clair base images

yixingjia:
1>update clair base iamges to vmware/photon:1.0
2>update oss packages to latest

### DIFF
--- a/make/photon/clair/Dockerfile
+++ b/make/photon/clair/Dockerfile
@@ -1,6 +1,9 @@
-FROM library/photon:1.0
+FROM vmware/photon:1.0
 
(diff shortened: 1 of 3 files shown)


make/photon/clair/Dockerfile
```diff
@@ -1,6 +1,9 @@
-FROM library/photon:1.0
+FROM vmware/photon:1.0
 
-RUN tdnf install -y git bzr rpm xz \
+RUN tdnf distro-sync -y \
```
wy65701436: tdnf distro-sync || echo？

@reasonerjt

PR #3484, opened 2017-10-30
Enable https support in CI testing

wy65701436:
This change is to enable https support in CI, add harbor.ca into Chrome trust list, then can use Chrome headless to access harbor with https.


reasonerjt:
Please refine the commit message and fix the tab .vs. space issue

PR #3512, opened 2017-11-01
modify the remote cert checkbox value

pengpengshui:


steven-zou:
Please make sure the all the builds are passed

PR #3530, opened 2017-11-02
Deprivilege harobr-log, harbor-db, registry image

reasonerjt:


make/common/mariadb/Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM vmware/photon:1.0
 
 RUN tdnf distro-sync -y || echo \
     && tdnf install -y sed shadow procps-ng gawk gzip sudo net-tools \
-    && groupadd -r -g 999 mysql && useradd --no-log-init -r -g 999 -u 999 mysql \
+    && groupadd -r -g 10000 mysql && useradd --no-log-init -r -g 10000 -u 10000 mysql \
```
yixingjia: do we need to run some chown 10000:10000 -R /var/lib/mysql  in the docker-entrypoint.sh file for the db?

reasonerjt: entrypoint.sh of mariadb will call `chown`

https://github.com/vmware/harbor/blob/master/make/common/mariadb/docker-entrypoint.sh#L70

make/docker-compose.tpl
```diff
@@ -6,16 +6,18 @@ services:
...
     ports:
-      - 127.0.0.1:1514:514
+      - 127.0.0.1:1514:10514
     networks:
       - harbor
   registry:
     image: vmware/registry:2.6.2-photon
     container_name: registry
     restart: always
+    read_only: true
```
yixingjia: with this restrict will it be difficult for later issue investigation?

reasonerjt: Thanks make sense I'll remove this flag from compose template.

However let's make sure the images are ready for `read_only`

PR #3534, opened 2017-11-02
Update replication policy API to support trigger and filter

ywk253100:


docs/swagger.yaml
```diff
@@ -2428,6 +2437,18 @@ definitions:
...
+      trigger:
+        type: string
+        description: The trigger for schedule job.
+      filters:
+        type: array
+        description: >-
+          The replication policy filter array.
+        items:
+          $ref: '#/definitions/RepFilter'
+      replicate_deletion:
```
steven-zou: Where is the option of replicating the existing images immediately

PR #3535, opened 2017-11-02
db migration from 1.2.0 to 1.3.0

ninjadq:
1. add table `project_meta`
2. remove `public` column from `project`
3. add `insecure` column to `replication_target` table
4. add migration file


wy65701436:
Do not forget to push the 1.3 image.

PR #3536, opened 2017-11-02
Fail earlier when found database schema dismatch

ywk253100:


src/common/dao/base.go
```diff
@@ -71,6 +71,16 @@ func InitDatabase(database *models.Database) error {
 	if err := db.Register(); err != nil {
 		return err
 	}
+
+	version, err := GetSchemaVersion()
+	if err != nil {
+		return err
+	}
+	if version.Version != SchemaVersion {
+		return fmt.Errorf("unexpected database schema version, expected %s, got %s, please upgrade the database first",
```
reasonerjt: Please refine the error message as a mismatch maybe due to DB schema version is higher than code.

PR #3541, opened 2017-11-02
Implement filter chain

ywk253100:


src/replication/source/filter_chain.go
```diff
@@ -19,3 +19,37 @@ type FilterChain interface {
...
 	DoFilter(filterItems []models.FilterItem) []models.FilterItem
 }
+
+// DefaultFilterChain provides a default implement for interface FilterChain
+type DefaultFilterChain struct {
+	filters []Filter
+}
+
+// NewDefaultFilterChain returns an instance of DefaultFilterChain
+func NewDefaultFilterChain(filters []Filter) *DefaultFilterChain {
```
steven-zou: Suggest to use separate file to implement the interface

src/replication/source/filter_chain.go
```diff
@@ -19,3 +19,37 @@ type FilterChain interface {
...
+}
+
+// Filters returns the filter list
+func (d *DefaultFilterChain) Filters() []Filter {
+	return d.filters
+}
+
+// DoFilter does the filter works for filterItems
+func (d *DefaultFilterChain) DoFilter(filterItems []models.FilterItem) []models.FilterItem {
+	for _, filter := range d.filters {
```
steven-zou: Should add more guard and fast-fail code here

PR #3546, opened 2017-11-03
Run clair with limited user

yixingjia:
1>creat user clair
2>run clair with user clair

### DIFF
--- a/make/photon/clair/Dockerfile
+++ b/make/photon/clair/Dockerfile
@@ -2,15 +2,18 @@ FROM vmware/photon:1.0
 
 RUN tdnf distro-sync -y \
     && tdnf erase vim -y \
-    && tdnf install -y git bzr rpm xz \
+    && tdnf install -y git shadow sudo bzr rpm xz python-xml \
     && tdnf clean all \
-	&& mkdir /clair2.0.1/ 
-
+    && mkdir /clair2.0.1/ \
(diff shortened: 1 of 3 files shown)


make/photon/clair/Dockerfile
```diff
@@ -2,15 +2,18 @@ FROM vmware/photon:1.0
...
-EXPOSE 6060 6061	
-
-RUN chmod u+x /clair2.0.1/clair
-
-ENTRYPOINT ["/clair2.0.1/clair"]
+EXPOSE 6060 6061
+RUN chown -R 10000:10000 /clair2.0.1 \
+    && chmod u+x /clair2.0.1/clair \
+    && chmod u+x /docker-entrypoint.sh
+HEALTHCHECK --interval=30s --timeout=10s --retries=3 CMD curl -sS 127.0.0.1:6061/health || exit 1
```
reasonerjt: is there a reason we don't use the default attributes for HEALTHCHECK?
```
The options that can appear before CMD are:

--interval=DURATION (default: 30s)
--timeout=DURATION (default: 30s)
--start-period=DURATION (default: 0s)
--retries=N (default: 3)
```

yixingjia: NO special reason just consider the default value sometimes may change by different version of docker

For the 500 error currently it will not be able to handler for this patch.

PR #3554, opened 2017-11-06
Remove the Dockerfile of rsyslog image

reasonerjt:


make/photon/log/Dockerfile
```diff
@@ -1,8 +1,14 @@
-FROM vmware/rsyslog-photon:8.15.0
+FROM vmware/photon:1.0
+
+RUN tdnf distro-sync -y || echo \
```
yixingjia: It will be better not ignore the error when run tdnf distryo-sync, otherwise we may have a problematic image.

reasonerjt: This is a workaround when the vmware/photon randomly fails in `tdnf distro-sync` I don't wanna be blocked by issues on photon, will remove it from different Dockerfiles altogether in another PR.

PR #3613, opened 2017-11-14
Add Prometheus monitoring endpoint.

vasartori:
At this point we don't expose any custom metric. 
We need discuss a little bit more what we want monitoring.
Issue: #3195

### DIFF
--- a/src/adminserver/main.go
+++ b/src/adminserver/main.go
@@ -22,6 +22,7 @@ import (
 	syscfg "github.com/vmware/harbor/src/adminserver/systemcfg"
 	sysinfo "github.com/vmware/harbor/src/adminserver/systeminfo"
 	"github.com/vmware/harbor/src/common/utils/log"
+	"github.com/vmware/harbor/src/common/utils/prometheus"
 )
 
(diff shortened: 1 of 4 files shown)


likunbyl:
Good idea, I want to monitor the replication status of two harbor server, thus I can get alerts when replication go wrong, and keep replication ok in case I need to do a switchover between two harbor server in emergency.

clouderati:
Thanks for the contributing @vasartori – much appreciated. Also, apologies for dropping the ball on this and leaving you hanging so long. ☹️  I completely agree that we need a thorough story around metrics.

Does it make sense to open a discussion around what this would look like? For example, if we were to rebase / merge this PR, what would the complete Prometheus monitoring story look like?

Let me know what your thoughts are. Happy to work with you to get something like this merged.

clouderati:
Closing due to inactivity. Will re-open when we have the cycles to tackle this – hopefully in the near future.

mdraijer:
@likunbyl Replication monitoring is now built into https://github.com/c4po/harbor_exporter/

PR #3639, opened 2017-11-20
publish npm package for VIC after ci

ninjadq:
add script to build and publish npm package for VIC after CI


tests/integration.sh
```diff
@@ -132,6 +133,12 @@ if [ $nightly_run == true ]; then
     echo "Sendout Nightly Run Email success."
 fi
 
+## ---------------------------------- build & publish npm package for VIC ----------------------------------
+if [ $publish_npm == true ] && [ $rc -eq 0 ]; then
```
wy65701436: this should be happen at Push not PR?

PR #3640, opened 2017-11-20
Add database driver for Harbor configurations

yixingjia:
1>Add a new database driver for configurations
2> change the current default driver from json
to database


reasonerjt:
Let's hold this until we have a 1.3 branch.

make/common/db/registry.sql
```diff
@@ -221,9 +221,11 @@ UNIQUE(namespace)
 );
 
 create table properties (
+ id int NOT NULL AUTO_INCREMENT,
  k varchar(64) NOT NULL,
  v varchar(128) NOT NULL,
- primary key (k)
+ PRIMARY KEY(id),
+ UNIQUE (k)
```
ywk253100: Please record the schema changes in tools/migration/changelog.md so that  the others can finish the upgrade script work.

src/adminserver/systemcfg/systemcfg.go
```diff
@@ -215,17 +218,55 @@ func Init() (err error) {
...
+		if err != nil {
+			return err
+		}
+		//migration check: if no data in the db , then will try to load from path
+		m, err := CfgStore.Read()
+		if err != nil {
+			return err
+		}
+		if m == nil || len(m) == 0 {
+			if _, err := os.Stat(path); !os.IsNotExist(err) {
```
ywk253100: Seems the logic is incorrect, we should make sure there is no error before we can continue.

src/adminserver/systemcfg/systemcfg.go
```diff
@@ -215,17 +218,55 @@ func Init() (err error) {
...
+				jsondriver, err := json.NewCfgStore(path)
+				if err != nil {
```
ywk253100: The same as above.

src/adminserver/systemcfg/systemcfg.go
```diff
@@ -215,17 +218,55 @@ func Init() (err error) {
...
+					log.Errorf("Failed to read old configuration from %s", path)
+				}
+				err = CfgStore.Write(jsonconfig)
+				if err != nil {
+					log.Error("Failed to update old configuration to dattabase")
+				}
+			}
+		}
+	} else {
+		CfgStore, err = json.NewCfgStore(path)
```
ywk253100: Check the error.

src/adminserver/systemcfg/systemcfg.go
```diff
@@ -215,17 +218,55 @@ func Init() (err error) {
...
+			if _, err := os.Stat(path); !os.IsNotExist(err) {
+				jsondriver, err := json.NewCfgStore(path)
+				if err != nil {
+					log.Errorf("Failed to migrate configuration from %s", path)
+				}
+				jsonconfig, err := jsondriver.Read()
+				if err != nil {
+					log.Errorf("Failed to read old configuration from %s", path)
+				}
+				err = CfgStore.Write(jsonconfig)
```
ywk253100: If the configuration read from the database is null and the JSON config file does not exist either, we should write the configuration loaded from env to the database.

src/common/const.go
```diff
@@ -73,6 +73,8 @@ const (
 	UAAEndpoint                = "uaa_endpoint"
 	UAAClientID                = "uaa_client_id"
 	UAAClientSecret            = "uaa_client_secret"
-
+	VerifyRemoteCert           = "verify_remote_cert"
```
ywk253100: The configuration item is removed as the certificate verification is moved to the target level.

make/common/db/registry.sql
yixingjia: Done

src/adminserver/systemcfg/systemcfg.go
yixingjia: done

PR #3666, opened 2017-11-22
Implement the methods of WatchList

ywk253100:


src/replication/core/controller.go
```diff
@@ -118,15 +118,34 @@ func (ctl *Controller) UpdatePolicy(updatedPolicy models.ReplicationPolicy) erro
...
+		}
 	}
 
-	if err = ctl.policyManager.UpdatePolicy(updatedPolicy); err != nil {
-		return err
+	if reset {
+		if err = ctl.triggerManager.UnsetTrigger(id, *originPolicy.Trigger); err != nil {
+			return err
+		}
+		if err = ctl.policyManager.UpdatePolicy(updatedPolicy); err != nil {
```
steven-zou: Only need to call one update policy function and after that calling unset and setup to update the trigger

PR #3683, opened 2017-11-23
Ldap enhancement

stonezdj:
Fix includes:
1. Add LDAP remote certificate validation
2.Search and import ldap user when add project members
3.Trim space in ldap username, fix issue 3553

### DIFF
--- a/make/common/templates/adminserver/env
+++ b/make/common/templates/adminserver/env
@@ -11,6 +11,7 @@ LDAP_FILTER=$ldap_filter
 LDAP_UID=$ldap_uid
 LDAP_SCOPE=$ldap_scope
 LDAP_TIMEOUT=$ldap_timeout
+LDAP_VERIFY_CERT=true
 DATABASE_TYPE=mysql
 MYSQL_HOST=$db_host
 MYSQL_PORT=$db_port
--- a/src/common/models/ldap.go
+++ b/src/common/models/ldap.go
@@ -24,6 +24,7 @@ type LdapConf struct {
 	LdapUID               string `json:"ldap_uid"`
 	LdapScope             int    `json:"ldap_scope"`
 	LdapConnectionTimeout int    `json:"ldap_connection_timeout"`
+	LdapVerifyCert        bool   `json:"ldap_verify_cert"`
 }
 
 // LdapUser ...
--- a/src/common/utils/ldap/ldap.go
+++ b/src/common/utils/ldap/ldap.go
@@ -60,6 +60,7 @@ func GetSystemLdapConf() (models.LdapConf, error) {
 	ldapConfs.LdapUID = ldap.UID
 	ldapConfs.LdapScope = ldap.Scope
 	ldapConfs.LdapConnectionTimeout = ldap.Timeout
+	ldapConfs.LdapVerifyCert = ldap.VerifyCert
 
 	//	ldapConfs = config.LDAP().URL
 	//	ldapConfs.LdapSearchDn = config.LDAP().SearchDn
@@ -205,7 +206,8 @@ func SearchUser(ldapConfs models.LdapConf) ([]models.LdapUser, error) {
 	for _, ldapEntry := range result.Entries {
 		var u models.LdapUser
 		for _, attr := range ldapEntry.Attributes {
-			val := attr.Values[0]
+			//OpenLDAP sometimes contains leading space in username
+			val := strings.TrimSpace(attr.Values[0])
 			log.Debugf("Current ldap entry attr name: %s\n", attr.Name)
 			switch strings.ToLower(attr.Name) {
 			case strings.ToLower(ldapConfs.LdapUID):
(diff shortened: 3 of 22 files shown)


reasonerjt:
Could you squash the commits and refine the commit message?

make/common/templates/adminserver/env
```diff
@@ -11,6 +11,7 @@ LDAP_FILTER=$ldap_filter
 LDAP_UID=$ldap_uid
 LDAP_SCOPE=$ldap_scope
 LDAP_TIMEOUT=$ldap_timeout
+LDAP_VERIFY_CERT=true
```
ywk253100: Should read the config from harbor.cfg.

src/common/models/ldap.go
```diff
@@ -24,6 +24,7 @@ type LdapConf struct {
 	LdapUID               string `json:"ldap_uid"`
 	LdapScope             int    `json:"ldap_scope"`
 	LdapConnectionTimeout int    `json:"ldap_connection_timeout"`
+	LdapVerifyCert        bool   `json:"ldap_verify_cert"`
```
ywk253100: Duplicate of LDAP struct defined in common/models/config.go

make/common/templates/adminserver/env
stonezdj: This value is not intend to config in harbor.cfg

src/ui/api/member.go
```diff
@@ -158,12 +161,26 @@ func (pma *ProjectMemberAPI) Post() {
...
-	username := req.Username
+	username := strings.TrimSpace(req.Username)
 	userID := checkUserExists(username)
 	if userID <= 0 {
-		log.Warningf("User does not exist, user name: %s", username)
-		pma.RenderError(http.StatusNotFound, "User does not exist")
-		return
+		//check current authorization mode
+		authMode, err := config.AuthMode()
+		if err != nil || authMode != "ldap_auth" {
```
ywk253100: Return 500 and log the error if exists.

src/ui/api/member.go
```diff
@@ -158,12 +161,26 @@ func (pma *ProjectMemberAPI) Post() {
...
+			log.Warningf("User does not exist, user name: %s", username)
+			pma.RenderError(http.StatusNotFound, "User does not exist")
+			return
+		}
+
+		//search and import user
+		newUserID, err := ldapUtils.SearchAndImportUser(username)
+		if err == nil || newUserID > 0 {
```
ywk253100: if err != nil{
  // log error and return 500
}

//else
if newUserID == 0{
  // not found return 404
}

// otherwise continue

src/ui/config/config.go
```diff
@@ -183,6 +182,11 @@ func LDAP() (*models.LDAP, error) {
 	ldap.Filter = cfg[common.LDAPFilter].(string)
 	ldap.Scope = int(cfg[common.LDAPScope].(float64))
 	ldap.Timeout = int(cfg[common.LDAPTimeout].(float64))
+	if cfg[common.LDAPVerifyCert] != nil {
+		ldap.VerifyCert = cfg[common.LDAPVerifyCert].(bool)
+	} else {
+		ldap.VerifyCert = false
```
ywk253100: Set it as true for security reason.

src/common/utils/ldap/ldap.go
```diff
@@ -401,3 +405,38 @@ func searchLDAP(ldapConfs models.LdapConf, ldap *goldap.Conn) (*goldap.SearchRes
...
+
+	ldapUsers, err := SearchUser(ldapConfs)
+	if err != nil {
+		log.Errorf("Can not search ldap, error %v, filter: %s", err, ldapConfs.LdapFilter)
+		return 0, err
+	}
+
+	if len(ldapUsers) > 0 {
+		log.Debugf("Importing user %s to local database", ldapUsers[0].Username)
+		if userID, err = ImportUser(ldapUsers[0]); err != nil {
```
ywk253100: Will more than one users be returned? How do we handle this?

src/ui/config/config.go
stonezdj: OK

src/common/models/ldap.go
stonezdj: Will do it in refactor ldap code

src/ui/api/member.go
stonezdj: Updated

src/ui/api/member.go
stonezdj: Updated

make/common/templates/adminserver/env
ywk253100: Why isn't it configured in harbor.cfg as other properties are?

stonezdj: If it can be re-configured in harbor admin console, it doesn't have to put in harbor.cfg

src/common/models/ldap.go
stonezdj: the LDAP struct hold the configure values returned to UI, and the LdapConfig hold the value in internal configure and http request, their json format is different.

Issue #3691, opened 2017-11-27
Add unit test for repository update API after merging replication_enhancement branch as the new test framework is in that branch


ywk253100:

ywk253100:
Test Put method of Repository API.

ywk253100:
Done by #3950.

PR #3692, opened 2017-11-27
Add update repository API to support description of repository

ywk253100:


ywk253100:
Open #3691 to track UT.

src/ui/api/repository.go
```diff
@@ -605,6 +605,43 @@ func (ra *RepositoryAPI) GetTopRepos() {
 	ra.ServeJSON()
 }
 
+// Put updates description info for the repository
+func (ra *RepositoryAPI) Put() {
+	name := ra.GetString(":splat")
+	repository, err := dao.GetRepositoryByName(name)
```
ninjadq: Is RepoName enough? I think also need ProjectID to locate the repository.

ywk253100: The first part of the repository name is the project name, so the repository name is unique.

PR #3699, opened 2017-11-28
Refactor ldap code

stonezdj:


src/common/utils/ldapv2/ldap.go
```diff
@@ -0,0 +1,447 @@
+package ldapv2
```
reasonerjt: we don't need a ldapv2 package, there should be only one ldap pkg.

reasonerjt:
@stonezdj I assume we don't need this?

PR #3715, opened 2017-11-29
Add LDAP e2e test

stonezdj:
Add LDAP e2e test, add filter to LDAP user, and log in with the user mike in group harbor_user


tests/ldap_test.ldif
```diff
@@ -12,3 +12,37 @@ gidNumber: 1001
 userPassword: 123456
 mail: test@example.com
 gecos: test
+
```
wy65701436: could you add some comments? Otherwise, it's hard for others to modify without background.

stonezdj: Updated

PR #3736, opened 2017-12-04
HA installation script

yixingjia:
Add --ha options when install Harbor.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -230,13 +230,14 @@ PACKAGE_OFFLINE_PARA=-zcvf harbor-offline-installer-$(GITTAGVERSION).tgz \
 				  $(HARBORPKG)/prepare $(HARBORPKG)/NOTICE \
 				  $(HARBORPKG)/upgrade $(HARBORPKG)/harbor_1_1_0_template \
 				  $(HARBORPKG)/LICENSE $(HARBORPKG)/install.sh \
-				  $(HARBORPKG)/harbor.cfg $(HARBORPKG)/$(DOCKERCOMPOSEFILENAME)
+				  $(HARBORPKG)/harbor.cfg $(HARBORPKG)/$(DOCKERCOMPOSEFILENAME) \
+				  $(HARBORPKG)/ha
 PACKAGE_ONLINE_PARA=-zcvf harbor-online-installer-$(GITTAGVERSION).tgz \
 		          $(HARBORPKG)/common/templates $(HARBORPKG)/prepare \
 				  $(HARBORPKG)/LICENSE $(HARBORPKG)/NOTICE \
 				  $(HARBORPKG)/upgrade $(HARBORPKG)/harbor_1_1_0_template \
 				  $(HARBORPKG)/install.sh $(HARBORPKG)/$(DOCKERCOMPOSEFILENAME) \
-				  $(HARBORPKG)/harbor.cfg
+				  $(HARBORPKG)/harbor.cfg $(HARBORPKG)/ha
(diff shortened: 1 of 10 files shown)


make/prepare
```diff
@@ -337,7 +400,23 @@ if customize_crt == 'on' and openssl_installed():
     empty_subj = "/C=/ST=/L=/O=/CN=/"
     private_key_pem = os.path.join(config_dir, "ui", "private_key.pem")
     root_crt = os.path.join(config_dir, "registry", "root.crt")
-    create_root_cert(empty_subj, key_path=private_key_pem, cert_path=root_crt)
+    if args.ha_mode:
+        shared_private_key_pem = os.path.join(base_dir, "ha", "private_key.pem")
```
reasonerjt: a little confused with the naming of the variables, sometime it's "common_xxx" sometime it's "share_xxxx".

Also don't quite understand the purpose to copy files back and forth between `config/` and `ha/` directories.

Let's discuss it.

make/ha/sample/ha_storage.cfg
```diff
@@ -0,0 +1,50 @@
+# Uncomment the storge provide for your HA environment.
```
reasonerjt: Do you plan to use a script to copy the chunk to config_ha.yml?

Or do you plan to just mention in the doc and let user manually update?

make/ha/docker-compose.tpl
```diff
@@ -0,0 +1,119 @@
+version: '2'
```
reasonerjt: Is there a reason we need this `ha/docker-compose.tpl`?

yixingjia: Since the ha docker-compose will be different with standalone. we don't need some containers like mysql.

make/ha/sample/active_standby/keepalived_active_standby.conf
```diff
@@ -0,0 +1,36 @@
+global_defs {
+  router_id haborlb
+}
+vrrp_script check_harbor {
+  script "/usr/local/bin/check_harbor.sh"
```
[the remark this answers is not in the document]
yixingjia: no, user need to put this file to some path, and we would suggested to put it at the /usr/local/bin  and we will specify this in the document.

make/ha/sample/ha_storage.cfg
yixingjia: When install with HA, the config_ha.yml will be placed to ~/common/config/registry/config.yml. It will be done by script.

This is because we have different config.yml for HA mode and standalone mode

Makefile
```diff
@@ -230,13 +230,14 @@ PACKAGE_OFFLINE_PARA=-zcvf harbor-offline-installer-$(GITTAGVERSION).tgz \
 				  $(HARBORPKG)/prepare $(HARBORPKG)/NOTICE \
 				  $(HARBORPKG)/upgrade $(HARBORPKG)/harbor_1_1_0_template \
 				  $(HARBORPKG)/LICENSE $(HARBORPKG)/install.sh \
-				  $(HARBORPKG)/harbor.cfg $(HARBORPKG)/$(DOCKERCOMPOSEFILENAME)
+				  $(HARBORPKG)/harbor.cfg $(HARBORPKG)/$(DOCKERCOMPOSEFILENAME) \
+				  $(HARBORPKG)/ha
 PACKAGE_ONLINE_PARA=-zcvf harbor-online-installer-$(GITTAGVERSION).tgz \
```
wy65701436: did you try to package online and offline installer?

yixingjia: We will only have 1 offline package and 1 online installer, both HA mode and standalone mode will use the same package.

I have try to test the offline package locally, I will check the online package.

PR #3768, opened 2017-12-09
Unify images tags and build process

wy65701436:


Makefile
```diff
@@ -75,26 +71,28 @@ UIPATH=$(BUILDPATH)/src/ui
...
+BUILDBIN=false
+MIGRATORFLAG=false
 
-#clair parameters
-CLAIRVERSION=v2.0.1-photon
-CLAIRFLAG=false
-CLAIRDBVERSION=9.6.5-photon
+#versions
+REGISTRYVERSION=v2.6.2
+NGINXVERSION=1.11.13
```
reasonerjt: The version of nginx, mariadb, postgresql can be misleading because they are pulled from photon's repo.
How about remove their own version and tag the images as `harbor-${VERSIONTAG}`

make/photon/registry/Dockerfile
```diff
@@ -13,7 +13,6 @@ COPY entrypoint.sh /
 RUN chmod u+x /entrypoint.sh
 
 RUN mkdir -p /etc/registry
-COPY config.yml /etc/registry/config.yml
```
reasonerjt: Why remove?

make/photon/notary/builder_public
```diff
@@ -41,19 +27,18 @@ echo 'build the notary binary bases on the golang:1.7.3...'
 cp binary.Dockerfile $TEMP 
 cd $TEMP
 docker build -f binary.Dockerfile -t notary-golang $TEMP
+cp $TEMP/migrations binary
```
reasonerjt: migrations is a directory right? so it needs `cp -r`?

make/photon/db/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM vmware/mariadb-photon:10.2.10
+FROM vmware/mariadb-photon:__version__
```
reasonerjt: Can we make sure the ` vmware/mariadb-photon:__version__` should never exist?

make/photon/notary/builder_public
```diff
@@ -41,19 +27,18 @@ echo 'build the notary binary bases on the golang:1.7.3...'
...
 
 echo 'copy the notary binary to local...'
 ID=$(docker create notary-golang)
 echo $ID
 cd $cur
-docker cp $ID:/go/src/github.com/docker/notary/notary-server binary
-docker cp $ID:/go/src/github.com/docker/notary/notary-signer binary
+docker cp $ID:/go/src/github.com/theupdateframework/notary/notary-server binary
```
reasonerjt: You are using `go install` in the binary.Dockerfile so the binary should be put into `/root/go/bin/` right?

make/photon/clair/Dockerfile
```diff
@@ -8,7 +8,7 @@ RUN tdnf distro-sync -y \
     && mkdir /clair2.0.1/ \
     && groupadd -r -g 10000 clair \
     && useradd --no-log-init -m -r -g 10000 -u 10000 clair
-COPY clair /clair2.0.1/
+COPY /binary/clair /clair2.0.1/
```
reasonerjt: It should be `./binary`?

.travis.yml
```diff
@@ -82,6 +82,10 @@ script:
   - sudo make run_clarity_ut CLARITYIMAGE=vmware/harbor-clarity-ui-builder:1.2.7
   - cat ./src/ui_ng/lib/npm-ut-test-results
   - sudo ./tests/testprepare.sh 
+  - sudo make -f make/photon/Makefile -e MARIADBVERSION=10.2.10 -e VERSIONTAG=dev
+  - sudo make -f make/photon/Makefile _build_registry -e REGISTRYVERSION=v2.6.2 -e VERSIONTAG=dev
+  - sudo sed -i 's/__reg_version__/v2.6.2-dev/g' ./make/docker-compose.test.yml
```
reasonerjt: seems the tag of registry image is not updated in `docker-compose.test.yml` in this PR

Makefile
wy65701436: agree. we can just use ${VERSIONTAG}.

make/photon/notary/builder_public
wy65701436: sure

.travis.yml
wy65701436: fixed.

make/photon/db/Dockerfile
wy65701436: no better solution so far, just leave it to replace the dockerfile.

make/photon/notary/builder_public
wy65701436: as notary has build issue, just leave it as it won't block our build process. I will open another issue later.

make/photon/registry/Dockerfile
wy65701436: no use anymore as the config is generated by prepare.

PR #3783, opened 2017-12-13
Replication enhancement

ywk253100:


src/replication/policy/manager.go
```diff
@@ -7,18 +21,28 @@ import (
...
+	RemovePolicy(int64) error
+}
+
+//DefaultManager provides replication policy CURD capabilities.
+type DefaultManager struct{}
 
-//NewManager is the constructor of Manager.
-func NewManager() *Manager {
-	return &Manager{}
+//NewDefaultManager is the constructor of DefaultManager.
```
steven-zou: Suggest to change the name , it's confusion

src/replication/replicator/replicator.go
```diff
@@ -0,0 +1,43 @@
...
+	Replicate(*api.ReplicationReq) error
+}
+
+// DefaultReplicator provides a default implement for Replicator
+type DefaultReplicator struct {
+	client jobserviceclient.Client
+}
+
+// NewDefaultReplicator returns an instance of DefaultReplicator
+func NewDefaultReplicator(endpoint string, client ...client.Client) *DefaultReplicator {
```
steven-zou: Same 'default' problem

PR #3790, opened 2017-12-14
Read image author from label 'maintainer' if author is null

ywk253100:


reasonerjt:
Checked the code, I think we can split a func and add UT?

PR #3794, opened 2017-12-14
Retry when test ldap server load data fail

stonezdj:
Retry when test ldap server load data fail

### DIFF
--- a/tests/ldapprepare.sh
+++ b/tests/ldapprepare.sh
@@ -14,3 +14,15 @@ sleep 5
 docker cp ldap_test.ldif ldap_server:/
 docker exec ldap_server ldapadd -x -D "cn=admin,dc=example,dc=com" -w admin -f /ldap_test.ldif -ZZ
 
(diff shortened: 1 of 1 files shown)


tests/ldapprepare.sh
```diff
@@ -14,3 +14,8 @@ sleep 5
 docker cp ldap_test.ldif ldap_server:/
 docker exec ldap_server ldapadd -x -D "cn=admin,dc=example,dc=com" -w admin -f /ldap_test.ldif -ZZ
 
+# failed and retry
+if [ ! $? -eq 0 ]; then
+    sleep 5
+    docker exec ldap_server ldapadd -x -D "cn=admin,dc=example,dc=com" -w admin -f /ldap_test.ldif -ZZ
```
reasonerjt: would be better to put line 15 and line 20 in one `for` loop, so after several retry this script exit with a non-zero return code and the CI/test fails.

PR #3804, opened 2017-12-15
Print stack trace when recover from panic and print warning message rather than returning an error when updating 0 records

ywk253100:


src/jobservice/job/statemachine.go
```diff
@@ -80,7 +81,7 @@ func (sm *SM) Start(s string) {
 	defer func() {
 		if r := recover(); r != nil {
 			sm.Logger.Errorf("Panic: %v, entering error state", r)
-			log.Warningf("Panic when handling job: %v, panic: %v, entering error state", sm.CurrentJob, r)
+			log.Warningf("Panic when handling job: %v, panic: %v \n %s \n, entering error state", sm.CurrentJob, r, debug.Stack())
```
reasonerjt: https://www.godoc.org/runtime/debug#Stack
Will this actually print the stack trace where the panic happens?

PR #3840, opened 2017-12-20
Allow project developer to edit repository description

ywk253100:


ywk253100:
UT works are tracked in #3691.

PR #3847, opened 2017-12-21
[WIP] Search user from UAA while adding member to project.

reasonerjt:


src/common/utils/uaa/client.go
```diff
@@ -92,6 +118,45 @@ func (dc *defaultClient) GetUserInfo(token string) (*UserInfo, error) {
 	return info, nil
 }
 
+func (dc *defaultClient) SearchUser(username string) ([]*SearchUserEntry, error) {
+	token, err := dc.twoLegCfg.Token(dc.prepareCtx())
+	if err != nil {
+		return nil, err
+	}
+	url := dc.endpoint + "/uaa/Users"
```
jessehu: In Pivotal Cloud Foundry, the UAA end point url is https://uaa.mydomain.local, and the Users API is https://uaa.mydomain.local/Users. There is no prefix '/uaa'.

reasonerjt: ok will remove it in this PR.  Thanks！

Issue #3663, opened 2017-11-22
LDAP info update?


yixingjia:
1> Whether we should allow import user if the user don't have email

2> Should we allow update LDAP user info when user login?

stonezdj:
User can not update its information if user is imported from LDAP
Sync user information when user is login

stonezdj:
Blocked by ldap refactor PR, move to sprint 23

PR #3870, opened 2017-12-26
Sync user email in ldap #3663

stonezdj:
Sync user email between harbor DB and LDAP #3663

### DIFF
--- a/src/common/dao/user.go
+++ b/src/common/dao/user.go
@@ -258,9 +258,12 @@ func DeleteUser(userID int) error {
 }
 
 // ChangeUserProfile ...
-func ChangeUserProfile(user models.User) error {
+func ChangeUserProfile(user models.User, cols ...string) error {
 	o := GetOrmer()
-	if _, err := o.Update(&user, "Email", "Realname", "Comment"); err != nil {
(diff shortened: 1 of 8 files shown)


src/common/dao/user.go
```diff
@@ -258,9 +258,12 @@ func DeleteUser(userID int) error {
 }
 
 // ChangeUserProfile ...
-func ChangeUserProfile(user models.User) error {
+func ChangeUserProfile(user models.User, cols ...string) error {
```
reasonerjt: Could you update the comment to explain the usage of `cols`?

src/ui/auth/ldap/ldap.go
```diff
@@ -26,7 +27,9 @@ import (
 )
 
 // Auth implements AuthenticateHelper interface to authenticate against LDAP
-type Auth struct{}
+type Auth struct {
+	auth.DefaultAuthenticateHelper
```
reasonerjt: I understand the intention but a little concerned about this approach, let's discuss.

src/common/dao/user.go
stonezdj: OK, add comment to it

PR #3903, opened 2018-01-03
Enable configuring the CA Certificate for UAA

reasonerjt:
Enable configuring the path of root cert of UAA in harbor.cfg.  It only
takes effects if the verify_cert is set to "true" If the file does not
exist, the configuration is skipped.
The intention for this commit is to support integration with nested UAA
in PAS or PKS, we don't expect user to manually configure this value,
though he can do it if he wants.

### DIFF
--- a/make/prepare
+++ b/make/prepare
@@ -250,6 +250,7 @@ uaa_endpoint = rcp.get("configuration", "uaa_endpoint")
 uaa_clientid = rcp.get("configuration", "uaa_clientid")
 uaa_clientsecret = rcp.get("configuration", "uaa_clientsecret")
 uaa_verify_cert = rcp.get("configuration", "uaa_verify_cert")
+uaa_ca_cert = rcp.get("configuration", "uaa_ca_cert")
 
 secret_key = get_secret_key(secretkey_path)
 log_rotate_count = rcp.get("configuration", "log_rotate_count")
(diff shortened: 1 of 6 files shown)


src/ui/auth/uaa/uaa.go
```diff
@@ -38,6 +39,7 @@ func CreateClient() (uaa.Client, error) {
 		ClientSecret:  UAASettings.ClientSecret,
 		Endpoint:      UAASettings.Endpoint,
 		SkipTLSVerify: !UAASettings.VerifyCert,
+		CARootPath:    os.Getenv("UAA_CA_ROOT"),
```
jessehu: Root CA is used commonly, so better to name it UAA_ROOT_CA?

make/prepare
```diff
@@ -387,6 +389,16 @@ shutil.copyfile(os.path.join(templates_dir, "jobservice", "app.conf"), jobservic
...
 
+if auth_mode == "uaa_auth":
+    if os.path.isfile(uaa_ca_cert):
+        if not os.path.isdir(ui_cert_dir):
+            os.makedirs(ui_cert_dir, mode=0o600)
+        ui_uaa_ca = os.path.join(ui_cert_dir, "uaa_ca.pem")
+        print("Copying UAA CA cert to %s" % ui_uaa_ca)
+        shutil.copyfile(uaa_ca_cert, ui_uaa_ca)
+    else:
+        print("Can not find UAA CA cert: %s, skip" % uaa_ca_cert)
```
jessehu: should this be an error msg or warning msg ?

reasonerjt: warning, given the limitation of current approach of reading harbor.cfg

src/ui/auth/uaa/uaa.go
reasonerjt: I think it's fine as it's an attribute in uaa.ClientConfig

PR #3918, opened 2018-01-04
Add registry storage config in harbor.cfg

jessehu:
Refer to https://docs.docker.com/registry/configuration/#storage
for all available configuration.

### DIFF
--- a/make/harbor.cfg
+++ b/make/harbor.cfg
@@ -147,7 +147,7 @@ uaa_ca_cert = /path/to/ca.pem
 
 ### Docker Registry setting ###
 #registry_storage_provider can be: filesystem, s3, gcs, azure, etc.
-registry_storage_provider = filesystem
-#registry_storage_provider_cfg is a comma separated "key: value" pairs, e.g. "key1: value, key2: value2".
+registry_storage_provider_name = filesystem
+#registry_storage_provider_config is a comma separated "key: value" pairs, e.g. "key1: value, key2: value2".
 #Refer to https://docs.docker.com/registry/configuration/#storage for all available configuration.
-registry_storage_provider_cfg =
+registry_storage_provider_config =
--- a/make/harbor.cfg
+++ b/make/harbor.cfg
@@ -133,14 +133,21 @@ clair_db_username = postgres
 
 #Clair default database
 clair_db = postgres
-
-
 ################### end of HA section #####################
+
 #************************END INITIAL PROPERTIES************************
+
 #The following attributes only need to be set when auth mode is uaa_auth
 uaa_endpoint = uaa.mydomain.org
 uaa_clientid = id
 uaa_clientsecret = secret
 uaa_verify_cert = true
 uaa_ca_cert = /path/to/ca.pem
-#############
+
+
+### Docker Registry setting ###
+#registry_storage_provider can be: filesystem, s3, gcs, azure, etc.
+registry_storage_provider = filesystem
+#registry_storage_provider_cfg is a comma separated "key: value" pairs, e.g. "key1: value, key2: value2".
+#Refer to https://docs.docker.com/registry/configuration/#storage for all available configuration.
+registry_storage_provider_cfg =
--- a/make/prepare
+++ b/make/prepare
@@ -23,13 +23,14 @@ def validate(conf, args):
     if args.ha_mode:
         db_host = rcp.get("configuration", "db_host")
         if db_host == "mysql":
-            raise Exception("Error: In HA mode, db_host in harbor.cfg needs to point to an external DB address")
-        registry_config_path = os.path.join(templates_dir,"registry","config_ha.yml")
-        if check_storage_config(registry_config_path):
-            raise Exception("Error: In HA model shared storage configuration is required registry, refer HA installation guide for detail.")
+            raise Exception("Error: In HA mode, db_host in harbor.cfg needs to point to an external DB address.")
+        registry_storage_provider = rcp.get("configuration",
+                                            "registry_storage_provider").strip()
+        if registry_storage_provider == 'filesystem':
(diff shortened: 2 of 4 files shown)


jessehu:
cc @yixingjia @steven-zou

reasonerjt:
Do we still need registry_ha.yaml?

make/harbor.cfg
```diff
@@ -133,14 +133,21 @@ clair_db_username = postgres
...
 uaa_ca_cert = /path/to/ca.pem
-#############
+
+
+### Docker Registry setting ###
+#registry_storage_provider can be: filesystem, s3, gcs, azure, etc.
+registry_storage_provider = filesystem
+#registry_storage_provider_cfg is a comma separated "key: value" pairs, e.g. "key1: value, key2: value2".
+#Refer to https://docs.docker.com/registry/configuration/#storage for all available configuration.
+registry_storage_provider_cfg = rootdirectory: /storage
```
reasonerjt: This is tricky, because the `/storage` is a path inside the container, which we mounted to `/data/registry`.  If we don't want to handle it we need to hard code it.

I think the form `k1: v1, k2: v2` is OK but for `filesystem` storage provider, let's just hard code it to `/storage`

jessehu: If the user selects 'filesystem' as the registry storage in Harbor Tile, then we do not provide  text box for the user to specify rootdirectory. So it's hard-coded in Harbor Tile.

make/prepare
```diff
@@ -23,13 +23,14 @@ def validate(conf, args):
...
         if db_host == "mysql":
-            raise Exception("Error: In HA mode, db_host in harbor.cfg needs to point to an external DB address")
-        registry_config_path = os.path.join(templates_dir,"registry","config_ha.yml")
-        if check_storage_config(registry_config_path):
-            raise Exception("Error: In HA model shared storage configuration is required registry, refer HA installation guide for detail.")
+            raise Exception("Error: In HA mode, db_host in harbor.cfg needs to point to an external DB address.")
+        registry_storage_provider = rcp.get("configuration",
+                                            "registry_storage_provider").strip()
+        if registry_storage_provider == 'filesystem':
+            raise Exception("Error: In HA mode, shared storage configuration in harbor.cfg is required. Refer to HA installation guide for details.")
```
yixingjia: User can still use file system. as Ceph, NFS.  they are also filesystem.

make/harbor.cfg
```diff
@@ -133,14 +133,21 @@ clair_db_username = postgres
...
 uaa_ca_cert = /path/to/ca.pem
-#############
+
+
+### Docker Registry setting ###
+#registry_storage_provider can be: filesystem, s3, gcs, azure, etc.
+registry_storage_provider = filesystem
+#registry_storage_provider_cfg is a comma separated "key: value" pairs, e.g. "key1: value, key2: value2".
+#Refer to https://docs.docker.com/registry/configuration/#storage for all available configuration.
+registry_storage_provider_cfg =
```
yixingjia: Just my two cents: 
what about we provide a storage configuration file , the content of the file will be same as https://docs.docker.com/registry/configuration/#storage  says.  so that we don't need to explain the format. 
eg:
registry_storge_conf = storage.cfg

jessehu: got your point. then we need to parse or read yaml formatted file storage.cfg and replace its content with the placeholder in config.yml.
It's doable. And if more config sections are needed (besides the storage section), then more separate cfg snippet file are needed. BTW, we also need to explain the yaml format and note the indent in the snippet storage.cfg.

make/prepare
jessehu: 'filesystem' is a type supported by Registry per https://docs.docker.com/registry/configuration/#storage. For NFS, it's already mounted on the OS, so we don't need to distinguish it.

jessehu: Get your point now. Since we do not know whether it's local file system or NFS when the storage provider is 'filesystem', then no need to check it here? @yixingjia is this error tip in the HA installation guide too ? I guess the old code  prints this error message to the users to guide them configure the correct registry storage provider.

yixingjia: The old message is just to make sure User understanding what they are doing when setup HA. Since by default we use local fileysystem.  But the we don't know the backend support if they use local storage then there will be issue in HA setup.

make/harbor.cfg
jessehu: After discussion with Yixing, we will ask the user a question to confirm shared storage is configured for Registry:
  Is the Harbor Docker Registry configured to use shared storage (e.g. NFS, S3, GCS, etc.)? [yes/no]:

make/prepare
jessehu: After discussion with Yixing, we will ask the user a question to confirm shared storage is configured for Registry:
Is the Harbor Docker Registry configured to use shared storage (e.g. NFS, S3, GCS, etc.)? [yes/no]:

make/prepare
```diff
@@ -354,16 +357,16 @@ render(os.path.join(templates_dir, "ui", "env"),
...
-        storage_provider_cfg = 'rootdirectory: /storage' + ',' + storage_provider_cfg
+storage_provider_name = rcp.get("configuration", "registry_storage_provider_name").strip()
+storage_provider_config = rcp.get("configuration", "registry_storage_provider_config").strip()
+if storage_provider_name == "filesystem":
+    if not storage_provider_config:
+        storage_provider_config = "rootdirectory: /storage"
+    elif "rootdirectory:" not in storage_provider_config:
+        storage_provider_config = "rootdirectory: /storage" + "," + storage_provider_config
 # generate storage configuration section in yaml format
 storage_provider_info = ('\n' + ' ' * 4).join(
```
yixingjia: It should be 8 space right?

jessehu: It's 4 spaces since #3908 Fix indentation in registry yaml files was merged.

PR #3931, opened 2018-01-04
Modify filter's position and Modify member module switch member's app…

pengpengshui:
Modify filter's position in project, user, replication modules

Modify switch members' apperence in project member module


pengpengshui:
useless

PR #3995, opened 2018-01-10
Provide API to rename admin

reasonerjt:
This is to provide a workaround for very corner case that in user's
authentication backend (LDAP, UAA) has a user called "admin" and because
Harbor's super user is hard coded to "admin" it's not possible to login
the "admin" with credentials in LDAP or UAA.

To minimize the impact, we'll provide an internal API for user to update
the super user's username from "admin" to "admin@harbor.local", this API
can be called by "admin" only, and is not reversible.

### DIFF
--- a/src/ui/api/internal.go
+++ b/src/ui/api/internal.go
@@ -15,6 +15,10 @@
 package api
 
 import (
+	"github.com/vmware/harbor/src/common"
+	"github.com/vmware/harbor/src/common/dao"
+	"github.com/vmware/harbor/src/common/models"
+	"github.com/vmware/harbor/src/common/utils/log"
 	"net/http"
 )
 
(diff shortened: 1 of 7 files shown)


reasonerjt:
@ywk253100 Shall we merge this PR to unblock the UI work.  I can deal with the coverage issue later.

src/ui/api/internal.go
```diff
@@ -43,3 +47,27 @@ func (ia *InternalAPI) SyncRegistry() {
...
+	if req.Name != "" {
+		newName = req.Name
+	} else {
+		newName = common.NewHarborAdminName
+	}
+	if err := dao.ChangeUserProfile(models.User{
+		UserID:   1,
+		Username: newName,
+	}, "username"); err != nil {
+		log.Errorf("Failed to change admin's username, error: %v", err)
```
ywk253100: Return 500?

PR #4016, opened 2018-01-12
Enable user to edit UAA settings on UI

reasonerjt:


src/ui/api/config.go
```diff
@@ -338,11 +338,7 @@ func convertForGet(cfg map[string]interface{}) (map[string]*value, error) {
 		return nil, err
 	}
 	//All configuration of UAA will be un-editable for PKS 1.0 (1.4)
```
jessehu: CI failed and the UAA un-editable comment should be updated as well.

reasonerjt: @jessehu Thanks the misleading comment will be removed, the CI failure maybe due to slow render, and the code failed to get the element, we'll see

PR #4020, opened 2018-01-15
Move some tests into nightly from BAT

wy65701436:


reasonerjt:
Let's hold this until we are confident about the nightly.

PR #4036, opened 2018-01-16
Propagate registry storage driver name to adminserver and return it in /api/systeminfo

ywk253100:


src/common/const.go
```diff
@@ -29,58 +29,59 @@ const (
 	RoleDeveloper    = 2
 	RoleGuest        = 3
 
-	ExtEndpoint                = "ext_endpoint"
-	AUTHMode                   = "auth_mode"
-	DatabaseType               = "database_type"
-	MySQLHost                  = "mysql_host"
-	MySQLPort                  = "mysql_port"
```
jessehu: what's the change in these lines? Spaces to Tab?

src/ui/api/systeminfo.go
```diff
@@ -86,18 +86,19 @@ var (
...
+	AdmiralEndpoint             string                           `json:"admiral_endpoint"`
+	AuthMode                    string                           `json:"auth_mode"`
+	RegistryURL                 string                           `json:"registry_url"`
+	ProjectCreationRestrict     string                           `json:"project_creation_restriction"`
+	SelfRegistration            bool                             `json:"self_registration"`
+	HasCARoot                   bool                             `json:"has_ca_root"`
+	HarborVersion               string                           `json:"harbor_version"`
+	NextScanAll                 int64                            `json:"next_scan_all"`
+	ClairVulnStatus             *models.ClairVulnerabilityStatus `json:"clair_vulnerability_status,omitempty"`
+	RegistryStorageProviderName string                           `json:"registry_storage_provider_name"`
```
jessehu: Do we need to consider migration case? Because in previous version, there is no registry_storage_provider_name in harbor.cfg.

src/common/const.go
ywk253100: The IDE does some format works to keep the alignment of the variables.

src/ui/api/systeminfo.go
ywk253100: No migration works for this release.

PR #4037, opened 2018-01-16
add admin rename ui

ninjadq:
Add UI Element for rename admin


src/ui_ng/src/app/account/account-settings/account-settings-modal.component.html
```diff
@@ -2,11 +2,29 @@
     <h3 class="modal-title">{{'PROFILE.TITLE' | translate}}</h3>
     <inline-alert class="modal-title" (confirmEvt)="confirmCancel($event)"></inline-alert>
     <div class="modal-body" style="overflow-y: hidden;">
```
steven-zou: Reuse inline-alert

PR #4039, opened 2018-01-16
Add update ldapScope for migrate

stonezdj:
Add update ldapScope for migrate

### DIFF
--- a/src/adminserver/systemcfg/systemcfg.go
+++ b/src/adminserver/systemcfg/systemcfg.go
@@ -35,6 +35,7 @@ import (
 const (
 	defaultJSONCfgStorePath string = "/etc/adminserver/config/config.json"
 	defaultKeyPath          string = "/etc/adminserver/key"
+	ldapScopeKey            string = "ldap_scope"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


reasonerjt:
Please check the CI failure.

src/adminserver/systemcfg/systemcfg.go
```diff
@@ -274,6 +275,11 @@ func initCfgStore() (err error) {
 					log.Errorf("Failed to read old configuration from %s", path)
 					return err
 				}
+				// Update LDAP Scope for migration
+				// only used when migrating harbor release before v1.3
+				// after v1.3 there is always a db configuration before migrate.
+				validLdapScope(jsonconfig, true)
```
jessehu: One suggestion on upgrading these kind of value changes in database: can it be added in https://github.com/vmware/harbor/blob/master/tools/migration/migration_harbor/versions/1_4_0.py ? Otherwise we need to ensure the Harbor version <= 1.3.0 in validLdapScope() in Harbor 1.4.1+ code ?

stonezdj: Harbor release > 1.3.0: The ldapScope value is stored in the database after 1.3.0.  and don't need to update after migration
Harbor release <1.3.0 : The ldapScope is stored in /data/config/config.json before 1.3, the ldapScope need to be changed from 1,2,3 to 0,1,2. we can not get the previous harbor version from the database because it is already updated. this part of logic only run for Harbor version < 1.3.0.

jessehu: Had a offline discussion with Daojun. This code will be triggered only if the LDAP info exists in  /data/config/config.json and this config.json only exists in Harbor <=1.3.0. So when upgrading from 1.4.0 to 1.4.x/1.5.x, this code will not run.
Thanks Daojun a lot for the clarification!

PR #4058, opened 2018-01-17
OVA migration guide

stonezdj:
OVA migration guide


reasonerjt:
In Disk expansion guide the screenshot is from old build.

PR #4098, opened 2018-01-22
Add a space after ':' when generating registry storage yaml config

jessehu:
yaml requires 1 or more spaces between the key and value. This patch
is useful in case the user inputs 'key:value' instead of 'key: value'.

### DIFF
--- a/make/prepare
+++ b/make/prepare
@@ -27,7 +27,7 @@ def validate(conf, args):
         registry_storage_provider_name = rcp.get("configuration",
                                                  "registry_storage_provider_name").strip()
         if registry_storage_provider_name == "filesystem" and not args.yes:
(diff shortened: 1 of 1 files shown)


make/prepare
```diff
@@ -264,6 +264,8 @@ else:
 
 storage_provider_name = rcp.get("configuration", "registry_storage_provider_name").strip()
 storage_provider_config = rcp.get("configuration", "registry_storage_provider_config").strip()
+# yaml requires 1 or more spaces between the key and value
+storage_provider_config = storage_provider_config.replace(":", ": ")
```
yixingjia: What if you already specify a space after the ":" then we will have two spaces, will this still work?

jessehu: yes, the comment says "yaml requires 1 or more spaces between the key and value"

PR #4124, opened 2018-01-24
Add place holder for injecting UAA host

reasonerjt:
As this is for tile deployment only, so add a shortcut for tile/bosh
script to add entry in /etc/hosts inside the container.
Due to effort consideration I don't think we want to render
docker-compose in `prepare` script.

### DIFF
--- a/make/docker-compose.tpl
+++ b/make/docker-compose.tpl
@@ -80,6 +80,8 @@ services:
       - /data/secretkey:/etc/ui/key:z
       - /data/ca_download/:/etc/ui/ca/:z
       - /data/psc/:/etc/ui/token/:z
+    extra_hosts:
+      - "PKS_UAA_HOST:127.0.0.1"
     networks:
       - harbor
     depends_on:


make/docker-compose.tpl
```diff
@@ -80,6 +80,8 @@ services:
       - /data/secretkey:/etc/ui/key:z
       - /data/ca_download/:/etc/ui/ca/:z
       - /data/psc/:/etc/ui/token/:z
+    extra_hosts:
+      - "PKS_UAA_HOST:PKS_UAA_IP_ADDR"
```
jessehu: better to name it 'UAA_HOSTNAME' and 'UAA_IP', because it might be PKS or PAS or other in future.

reasonerjt: but for PAS there is a DNS in the network?  So I believe this placeholder is for PKS only?

jessehu: just in case adding other UAA provider in future

PR #4168, opened 2018-01-29
Update HA document.

yixingjia:
Format the docs

### DIFF
--- a/docs/high_availability_installation_guide.md
+++ b/docs/high_availability_installation_guide.md
@@ -1,30 +1,23 @@
 # Harbor High Availability Guide 
 
-## Abstract 
-
-This guide describes how to install and configure Harbor for high availability. It supplements the Installation Guides and assumes that you are familiar with the material in those guides. 
-
-**Important!** 
-
-This guide was last updated as of the Harbor 1.4 release. It will not apply to releases before 1.4. We advise that you read this at your own discretion when planning on your Harbor high availability implementation.  
-
-This guide is intended as advice only. It doesn't support Notary(Content trust functionality) for now. 
-
 ## Contents 
 
 **Notices** 
 
 Notices take these forms: 
 
-**Note!**: A comment with additional information that explains a part of the text. 
+**Note**: A comment with additional information that explains a part of the text. 
 
-**Important!**: Something you must be aware of before proceeding. 
+**Important**: Something you must be aware of before proceeding. 
 
-**Tip!**: An extra but helpful piece of practical advice. 
+**Tip**: An extra but helpful piece of practical advice. 
 
-**Caution!**: Helpful information that prevents the user from making mistakes. 
+## Abstract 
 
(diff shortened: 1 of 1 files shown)


docs/high_availability_installation_guide.md
```diff
@@ -164,36 +159,102 @@ run the follow command to apply the change.
...
+##### 2> Extract the harbor-offline-installer-v.x.x.x.tgz You will get a folder ```harbor``` in the current directory 
+##### 3> cd to ```harbor``` directory 
+##### 4> Configure hostname
+In ```harbor.cfg```
+
+```
+hostname = reg.mydomain.com 
+
+```
+Change reg.mydomain.com to your FQDN or VIP(eg. 192.168.111.220)
```
wy65701436: 192.168.1.220?

docs/high_availability_installation_guide.md
```diff
@@ -164,36 +159,102 @@ run the follow command to apply the change.
...
 
-**Important!**
+**Important**
 
 If you set ```filesystem``` for the ```registry_storage_provider_name``` you must make sure the registry directory ```/data/registry``` mount to a shared storage like NFS,Ceph, etc. You need to create the /data/registry directory first and change it's owner to 10000:10000 as registry will run as userID 10000 and groupID 10000. 
 
- - 13> Import Harbor database schema to your external MariaDB 
+##### 9> Import Harbor database schema to your external MariaDB 
 ```
     Copy the ~/harbor/ha/registry.sql file to a machine that has MariaDB client installed
```
wy65701436: Could we move this step up to the part of MariaDB? It's so confuse to put this step here because it's not a setting of Harbor node.

Let user to get the registry.sql from source and then import it with Maria Client.

docs/high_availability_installation_guide.md
```diff
@@ -1,30 +1,21 @@
 # Harbor High Availability Guide
```
wy65701436: What's this document for? If it's only an simple HA set up, I think it's not necessary to describe what are the stateful and stateless services in Harbor? This is not enough to let reader knows the technical details but introduce more questions.

yixingjia: It's help user to better under Harbor and can be used as a reference for user to setup HA. This document also give an example on how to setup HA.

docs/high_availability_installation_guide.md
```diff
@@ -229,41 +290,53 @@ You need to change 192.168.1.220 to your VIP address before issue the follow com
...
- ```
+Move the harbor_ha.tar file to your home directory on harbor_node2
+
+##### 2> Untar the file
+```
+#> tar -xvf harbor_ha.tar
+```
+You will get a “harbor” folder in your home directory. 
+
+##### 3> Create certificate folder
```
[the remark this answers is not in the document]
yixingjia: Yes, just update the comment to mark it as optional

Issue #4164, opened 2018-01-29
Issue when changing or deleting member role of user himself


ywk253100:
Login as user02 and try to change the role of himself:
![member_01](https://user-images.githubusercontent.com/5835782/35498956-8b6b489e-050b-11e8-97a4-42a89672d307.png)

Got error:
![member_02](https://user-images.githubusercontent.com/5835782/35498959-8d7510ca-050b-11e8-924e-870820b34ea1.png)

The `REMOVE` should be disabled when trying to delete himself:
![member_delete](https://user-images.githubusercontent.com/5835782/35499279-a481f836-050c-11e8-9d79-14745d31514e.png)

wy65701436:
Encounter the same issue.

pengpengshui:
had already change to when delete himself ,the remove button can't be clicked

PR #4174, opened 2018-01-29
Modify users about delete myself display issue #4164

pengpengshui:


src/ui_ng/src/app/user/user.component.ts
```diff
@@ -223,16 +227,12 @@ export class UserComponent implements OnInit, OnDestroy {
     let userArr: string[] = [];
     this.batchDelectionInfos = [];
     if (users && users.length) {
```
steven-zou: add double confirm

PR #4230, opened 2018-02-06
Refactor project member

stonezdj:
Refactor project member so that it can contain group member.
Changes include:
1)  Update table project_member, add id column to project_member, change user_id to entity_id, and add entity_type ('u' for user member, 'g' for group member), primary key changed.
2)  Change project member CRUD API,  add parameter entityType so that it can contain group members.

### DIFF
--- a/src/common/dao/projectmember.go
+++ b/src/common/dao/projectmember.go
@@ -15,25 +15,39 @@
 package dao
 
 import (
+	"github.com/vmware/harbor/src/common"
 	"github.com/vmware/harbor/src/common/models"
 )
 
 // AddProjectMember inserts a record to table project_member
-func AddProjectMember(projectID int64, userID int, role int) error {
+func AddProjectMember(projectID int64, userID int, role int, entityType string) (int, error) {
 	o := GetOrmer()
+	if !(entityType == common.UserMember || entityType == common.GroupMember) {
+		entityType = common.UserMember
+	}
+	sql := `insert into project_member (project_id, entity_id , role, entity_type) values (?, ?, ?, ?)`
+	_, err := o.Raw(sql, projectID, userID, role, entityType).Exec()
+	if err != nil {
+		return 0, err
+	}
+	var pmid int
+	querySQL := `select id from project_member where project_id = ? and entity_id = ? and entity_type = ? limit 1`
 
-	sql := "insert into project_member (project_id, user_id , role) values (?, ?, ?)"
-
-	_, err := o.Raw(sql, projectID, userID, role).Exec()
-
-	return err
+	err = o.Raw(querySQL, projectID, userID, entityType).QueryRow(&pmid)
+	if err != nil {
+		return 0, err
+	}
+	return pmid, err
 }
 
 // UpdateProjectMember updates the record in table project_member
-func UpdateProjectMember(projectID int64, userID int, role int) error {
+func UpdateProjectMember(projectID int64, userID int, role int, entityType string) error {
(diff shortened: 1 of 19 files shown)


reasonerjt:
Please add more infor in the commit message to elaborate what and why.

src/common/dao/projectmember.go
```diff
@@ -15,66 +15,80 @@
...
+	err = o.Raw(querySQL, projectID, userID, entityType).QueryRow(&pmid)
+	if err != nil {
+		return 0, err
+	}
+	return pmid, err
 }
 
 // UpdateProjectMember updates the record in table project_member
-func UpdateProjectMember(projectID int64, userID int, role int) error {
+func UpdateProjectMember(projectID int64, userID int, role int, entityType string) error {
```
reasonerjt: rename the parameter "userID"  and I don't think we need entityType in the param list.

src/common/dao/projectmember.go
```diff
@@ -15,66 +15,80 @@
...
+	sql := `update project_member set role = ? where project_id = ? and entity_id = ?`
 
 	_, err := o.Raw(sql, role, projectID, userID).Exec()
 
 	return err
 }
 
 // DeleteProjectMember delete the record from table project_member
-func DeleteProjectMember(projectID int64, userID int) error {
+func DeleteProjectMember(projectID int64, userID int, entityType string) error {
```
reasonerjt: I don't think we need entityType in param list. Since `entityID` is unique.

src/common/dao/role.go
```diff
@@ -22,7 +22,7 @@ import (
 )
 
 // GetUserProjectRoles returns roles that the user has according to the project.
-func GetUserProjectRoles(userID int, projectID int64) ([]models.Role, error) {
+func GetUserProjectRoles(userID int, projectID int64, entityType string) ([]models.Role, error) {
```
reasonerjt: If it's only for user, we don't need `entityType` in the param list.

src/common/models/member.go
```diff
@@ -16,6 +16,17 @@ package models
...
+	ID         int    `orm:"pk;column(id)" json:"id"`
+	ProjectID  int64  `orm:"column(project_id)" json:"project_id"`
+	Entityname string `orm:"column(entity_name)" json:"entity_name"`
+	Rolename   string `json:"role_name"`
+	Role       int    `json:"role_id"`
+	EntityID   int    `orm:"column(entity_id)" json:"entity_id"`
+	EntityType string `orm:"column(entity_type)" json:"entity_type"`
+}
+
+// UserMember ...
```
reasonerjt: need more comments...

src/common/dao/projectmember.go
```diff
@@ -15,66 +15,80 @@
...
+	}
+	var pmid int
+	querySQL := `select id from project_member where project_id = ? and entity_id = ? and entity_type = ? limit 1`
 
-	sql := "insert into project_member (project_id, user_id , role) values (?, ?, ?)"
-
-	_, err := o.Raw(sql, projectID, userID, role).Exec()
-
-	return err
+	err = o.Raw(querySQL, projectID, userID, entityType).QueryRow(&pmid)
```
reasonerjt: This may not be needed, `o.Raw(sql, projectID, userID, role).Exec()` will return the `id` of the new record.

reasonerjt:
Per discussion quite a few items that I have comments are temporary and there will be further change to clean them up.  so I'll merge this PR so the work won't be blocked.

PR #4234, opened 2018-02-06
Upgrade docker to latest and check for docker login fail

wy65701436:


.drone.yml
```diff
@@ -11,7 +11,7 @@ pipeline:
     recursive: false
 
   test-and-issue-build:
-    image: vmware/harbor-e2e-engine:1.38
+    image: vmware/harbor-e2e-engine:1.39
```
reasonerjt: is this upgrade for using latest docker binary?

PR #4301, opened 2018-02-27
Modify test step to avoid an autotest bug

sigsbee:


tests/robot-cases/Group11-Nightly/Nightly.robot
```diff
@@ -153,6 +153,9 @@ Test Case - Project Level Policy Public
     Public Should Be Selected 
     Back To Projects
     # Project${d}  default should be private
+    # Here logout and login to try avoid a bug only in autotest
+    Logout Harbor
+    Sign In Harbor  ${HARBOR_URL}  ${HARBOR_ADMIN}  ${HARBOR_PASSWORD} 
     Project Should Be Public  project${d}
```
wy65701436: We need to confirm whether the project is listed the original page, otherwise the case will fail by unable to locate element.

PR #4307, opened 2018-02-28
Return zero instead of error if the volume directory is not found at …

reasonerjt:
…admin server

### DIFF
--- a/src/adminserver/systeminfo/imagestorage/filesystem/driver.go
+++ b/src/adminserver/systeminfo/imagestorage/filesystem/driver.go
@@ -15,9 +15,11 @@
 package filesystem
 
 import (
+	"os"
 	"syscall"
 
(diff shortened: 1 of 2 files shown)


src/adminserver/systeminfo/imagestorage/filesystem/driver.go
```diff
@@ -43,6 +44,11 @@ func (d *driver) Name() string {
 // Cap returns the capacity of the filesystem storage
 func (d *driver) Cap() (*storage.Capacity, error) {
 	var stat syscall.Statfs_t
+	if _, err := os.Stat(d.path); os.IsNotExist(err) {
+		// Return zero value if the path does not exist.
```
jessehu: better to print a warning in case the user specifies a wrong image store dir

reasonerjt: yep sure! Thanks.

PR #4342, opened 2018-03-05
Add monitor API for Harbor components

yixingjia:
1>AdminServer
2>Jobservice
3>UI

### DIFF
--- a/src/adminserver/handlers/router.go
+++ b/src/adminserver/handlers/router.go
@@ -27,5 +27,6 @@ func newRouter() http.Handler {
 	r.HandleFunc("/api/configurations", api.UpdateCfgs).Methods("PUT")
 	r.HandleFunc("/api/configurations/reset", api.ResetCfgs).Methods("POST")
 	r.HandleFunc("/api/systeminfo/capacity", api.Capacity).Methods("GET")
+	r.HandleFunc("/api/ping", api.Ping).Methods("GET")
 	return r
 }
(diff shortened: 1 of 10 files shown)


reasonerjt:
@yixingjia please rebase to resolve the conflict.

src/adminserver/handlers/router.go
```diff
@@ -27,5 +27,6 @@ func newRouter() http.Handler {
 	r.HandleFunc("/api/configurations", api.UpdateCfgs).Methods("PUT")
 	r.HandleFunc("/api/configurations/reset", api.ResetCfgs).Methods("POST")
 	r.HandleFunc("/api/systeminfo/capacity", api.Capacity).Methods("GET")
+	r.HandleFunc("/api/v1/configurations/ping", api.Ping).Methods("GET")
```
reasonerjt: let's not add version to a specific API.
And the `ping` shouldn't be under `configuration`

src/ui/router.go
```diff
@@ -60,6 +60,7 @@ func initRouters() {
 	}
 
 	// API
+	beego.Router("/api/v1/ui/ping", &api.SystemInfoAPI{}, "get:Ping")
```
reasonerjt: Similarly, how about just set the URL `/api/ping`

just4est:
Hello. I get 404 on /api/v1/configurations/ping . 
How can i get heath status of jobservice ?

PR #4349, opened 2018-03-06
Relead Harbor db and Clair db when restart Harbor

yixingjia:
Reload Harbor DB,
Reload Clair DB.

### DIFF
--- a/src/adminserver/systemcfg/systemcfg.go
+++ b/src/adminserver/systemcfg/systemcfg.go
@@ -152,6 +152,13 @@ var (
 	repeatLoadEnvs = map[string]interface{}{
 		common.ExtEndpoint:   "EXT_ENDPOINT",
 		common.MySQLPassword: "MYSQL_PWD",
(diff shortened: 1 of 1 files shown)


src/adminserver/systemcfg/systemcfg.go
```diff
@@ -152,6 +152,12 @@ var (
 	repeatLoadEnvs = map[string]interface{}{
 		common.ExtEndpoint:   "EXT_ENDPOINT",
 		common.MySQLPassword: "MYSQL_PWD",
+		common.MySQLUsername: "MYSQL_USR",
+		common.MySQLDatabase: "MYSQL_DATABASE",
+		common.MySQLPort: &parser{
```
ywk253100: Also needs MySQLHost?

yixingjia: Yes, this is the same issue as Clair. since we allow user to use third part DB.

PR #4372, opened 2018-03-07
Implement label management API

ywk253100:


make/photon/db/registry.sql
```diff
@@ -254,6 +254,19 @@ create table properties (
  UNIQUE (k)
  );
 
+create table label (
```
reasonerjt: How about add a unique constraint?

docs/swagger.yaml
```diff
@@ -1639,6 +1639,164 @@ paths:
...
+      responses:
+        '201':
+          description: Create successfully.
+        '400':
+          description: Invalid parameters.
+        '401':
+          description: User need to log in first.
+        '409':
+          description: >-
+            Label with the same name already exists.
```
reasonerjt: same name and same scope?

src/common/dao/label.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/astaxie/beego/orm"
+	"github.com/vmware/harbor/src/common/models"
+)
+
+// AddLabel creates a label
+func AddLabel(label *models.Label) (int64, error) {
+	now := time.Now()
+	label.CreationTime = now
+	label.UpdateTime = now
+	return GetOrmer().Insert(label)
```
reasonerjt: Could you Insert and Query again if the id is zero and error is `nil`?  For compatibility.

make/photon/db/registry_sqlite.sql
```diff
@@ -240,6 +240,18 @@ create table properties (
  UNIQUE(k)
  );
 
+create table label (
```
reasonerjt: how about `resource_label` or `harbor_label`?

src/ui/api/label.go
```diff
@@ -0,0 +1,276 @@
...
+
+// List labels according to the query strings
+func (l *LabelAPI) List() {
+	query := &models.LabelQuery{
+		Name:  l.GetString("name"),
+		Level: common.LabelLevelUser,
+	}
+
+	scope := l.GetString("scope")
+	if len(scope) == 0 {
```
reasonerjt: not needed?

src/ui/api/label.go
```diff
@@ -0,0 +1,276 @@
...
+			l.HandleBadRequest("project_id is required")
+			return
+		}
+		projectID, err := strconv.ParseInt(projectIDStr, 10, 64)
+		if err != nil || projectID <= 0 {
+			l.HandleBadRequest(fmt.Sprintf("invalid project_id: %s", projectIDStr))
+			return
+		}
+
+		exist, err := l.ProjectMgr.Exists(projectID)
```
reasonerjt: The list will be empty if the project does not exist?

make/photon/db/registry.sql
```diff
@@ -254,6 +254,19 @@ create table properties (
...
+ id int NOT NULL AUTO_INCREMENT,
+ name varchar(128) NOT NULL,
+ description text,
+ color varchar(16),
+ level char(1) NOT NULL,
```
reasonerjt: better also explain the values of column `scope` and `level` in the `.sql` file?

PR #4373, opened 2018-03-07
Update Harbor helm chart to deploy Harbor 1.4.0 release

jessehu:
cc @reasonerjt @paulczar for review

Verified following features:
- login UI
- basic function
- docker login
- docker push
- replicate image to another harbor
- scan image via clair

Notary support is not added yet.

### DIFF
--- a/contrib/helm/harbor/Chart.yaml
+++ b/contrib/helm/harbor/Chart.yaml
@@ -1,7 +1,7 @@
 name: harbor
-version: 0.0.1
-appVersion: 1.3.0
-description: An Enterprise-class Docker Registry Harbor by VMware
+version: 0.2.0
+appVersion: 1.4.0
+description: An Enterprise-class Docker Registry by VMware
 keywords:
 - vmware
 - docker
@@ -10,8 +10,10 @@ keywords:
 home: https://github.com/vmware/harbor
 icon: https://github.com/vmware/harbor/blob/master/docs/img/harbor_logo.png
 sources:
-- https://github.com/vmware/harbor
+- https://github.com/vmware/harbor/tree/master/contrib/helm/harbor
 maintainers:
+- name: Jesse Hu
+  email: huh@vmware.com
 - name: paulczar
   email: username.taken@gmail.com
 engine: gotpl
--- a/contrib/helm/harbor/Chart.yaml
+++ b/contrib/helm/harbor/Chart.yaml
@@ -1,5 +1,5 @@
 name: harbor
-version: 0.2.0
+version: 0.1.0
 appVersion: 1.4.0
 description: An Enterprise-class Docker Registry by VMware
 keywords:
--- a/contrib/helm/harbor/README.md
+++ b/contrib/helm/harbor/README.md
@@ -94,9 +94,9 @@ The following tables lists the configurable parameters of the Harbor chart and t
 | -----------------------    | ---------------------------------- | ----------------------- |
 | **Harbor** |
 | `harborImageTag`     | The tag for Harbor docker images | `v1.4.0` |
-| `externalDomain`       | Harbor will run on (https://*externalDomain*/). Make sure this FQDN resolves to the K8s Ingress Controller IP. | `harbor.my.domain` |
-| `insecureRegistry`     | Set to true if setting Harbor Registry as insecure-registries for docker | `false` |
-| `tlsCrt`               | TLS certificate to use for Harbor's https endpoint | auto-generated |
+| `externalDomain`       | Harbor will run on (https://`externalDomain`/). Recommend using K8s Ingress Controller FQDN as `externalDomain`, or make sure this FQDN resolves to the K8s Ingress Controller IP. | `harbor.my.domain` |
+| `insecureRegistry`     | If set to true, you don't need to set tlsCrt/tlsKey/caCrt, but must add Harbor FQDN as insecure-registries for your docker client. | `false` |
+| `tlsCrt`               | TLS certificate to use for Harbor's https endpoint. Its CN must match `externalDomain`. | auto-generated |
 | `tlsKey`               | TLS key to use for Harbor's https endpoint | auto-generated |
 | `caCrt`                | CA Cert for self signed TLS cert | auto-generated |
 | `persistence.enabled` | enable persistent data storage | `false` |
@@ -145,7 +145,7 @@ The following tables lists the configurable parameters of the Harbor chart and t
 | **Registry** |
 | `registry.image.repository` | Repository for registry image | `vmware/registry-photon` |
 | `registry.image.tag` | Tag for registry image | `v2.6.2-v1.4.0` |
-| `registry.image.pullPolicy` | Pull Policy for admregistryinserver image | `IfNotPresent` |
+| `registry.image.pullPolicy` | Pull Policy for registry image | `IfNotPresent` |
 | `registry.rootCrt` | registry root cert | see values.yaml |
 | `registry.httpSecret` | registry secret | `not-a-secure-secret` |
 | `registry.resources` | [resources](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/) to allocate for container   | undefined |
--- a/contrib/helm/harbor/README.md
+++ b/contrib/helm/harbor/README.md
@@ -35,8 +35,8 @@ helm init --canary-image
 Download Harbor helm chart code.
 
 ```bash
-git clone https://github.com/vmware/harbor/tree/master/contrib/helm/harbor helm-harbor
-cd helm-harbor
+git clone https://github.com/vmware/harbor
+cd harbor/contrib/helm/harbor
 ```
 
 ### Insecure Registry Mode
--- a/contrib/helm/harbor/README.md
+++ b/contrib/helm/harbor/README.md
@@ -4,24 +4,75 @@
 
 ## Introduction
 
-This is an experimental monolithic chart that installs and configures VMWare Harbor and its dependencies. The initial implementation of this includes all of the components required to run Harbor. As upstream harbor becomes more cloud native we will be able to break apart the monolith and utitlize helm dependencies.
+This chart installs and configures Harbor.
 
 ## Prerequisites
 
-- Kubernetes 1.7+ with Beta APIs enabled
+- Kubernetes cluster 1.8+ with Beta APIs enabled
+- Kubernetes Ingress Controller is enabled
+- kubectl CLI 1.8+
 - PV provisioner support in the underlying infrastructure
+
+## Setup a Kubernetes cluster
+
+You can use any tools to setup a K8s cluster.
+In this guide, we use [minikube](https://github.com/kubernetes/minikube) to setup a K8s cluster as the dev/test env.
+
+```bash
+# Start minikube
+minikube start --vm-driver=none
+# Enable Ingress Controller
+minikube addons enable ingress
+```
 
 ## Installing the Chart
 
-To install the chart with the release name `my-release`:
-
+First install [Helm CLI](https://github.com/kubernetes/helm#install), then initialize Helm.
 ```bash
-$ git clone https://github.com/vmware/harbor.git
-$ cd harbor/contrib/helm/harbor
-$ helm install --name my-release incubator/harbor
+helm init --canary-image
 ```
 
-The command deploys Harbor on the Kubernetes cluster in the default configuration. The [configuration](#configuration) section lists the parameters that can be configured during installation.
+Download Harbor helm chart code.
+
+```bash
+git clone https://github.com/vmware/harbor/tree/master/contrib/helm/harbor helm-harbor
+cd helm-harbor
+```
+
+### Insecure Registry Mode
+
+If setting Harbor Registry as insecure-registries for docker,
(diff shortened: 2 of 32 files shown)


contrib/helm/harbor/README.md
```diff
@@ -4,24 +4,75 @@
...
 - PV provisioner support in the underlying infrastructure
 
+## Setup a Kubernetes cluster
+
+You can use any tools to setup a K8s cluster.
+In this guide, we use [minikube](https://github.com/kubernetes/minikube) to setup a K8s cluster as the dev/test env.
+
+```bash
+# Start minikube
+minikube start --vm-driver=none
```
paulczar: I don't think we should recommend people use `--vm-driver=none` for minikube.

contrib/helm/harbor/README.md
```diff
@@ -4,24 +4,75 @@
...
+
+Then add `"insecure-registries": ["harbor.my.domain"]` in the docker daemon config file and restart docker service.
+
+### Secure Registry Mode
+
+By default this chart will generate a root CA and SSL certificate for your Harbor.
+You can also use your own CA signed certificate:
+
+open values.yaml, set the value of 'externalDomain' to your Harbor FQDN, and
+set value of 'tlsCrt', 'tlsKey', 'caCrt'. The common name of the certificate must match your Harbor FQDN.
```
paulczar: We should let them know that the defaults of these should work if using minikube with the default minikube ip of `192.168.99.100` (if we leave the externalDomain as the original default below)

contrib/helm/harbor/README.md
```diff
@@ -4,24 +4,75 @@
...
+
+Install the Harbor helm chart with a release name `my-release`:
 
 ```bash
-$ git clone https://github.com/vmware/harbor.git
-$ cd harbor/contrib/helm/harbor
-$ helm install --name my-release incubator/harbor
+helm install . --debug --name my-release --set externalDomain=harbor.my.domain
```
paulczar: `--set externalDomain` isn't needed if they're updating values.yaml as above.

paulczar: I don't like removing the `$` from in front of the commands.  It's used to tell people that's the bit they need to run, and allows us to show the output of the commands.   It also discourages people from copy/pasting blocks of commands directly to a terminal which is a bad practice.

contrib/helm/harbor/templates/ingress/secret.yaml
```diff
@@ -1,11 +1,15 @@
+{{ if not .Values.insecureRegistry }}
+{{ $ca := genCA "harbor-ca" 365 }}
```
paulczar: we should only do this if `.Values.tlsCrt`, `.Values.tlsKey`, `.Values.tlsCrt` is not set.

contrib/helm/harbor/values.yaml
```diff
@@ -28,89 +28,41 @@
...
-  7ULKyvaK+RPYKb4K2igMFHZD3KhaayShEG3rw/Y53hdmU+3I82tQ+txGmQoqicxg
-  BwAODdvixH5gP4idh7G1Q6tDvgJGGl2HvcE8fzbVIO3qDKefPlif20eX2gUc/Ut5
-  gyiyJutOQKVjEUb5bmUaeRyTXo8Vf2TIhIRfdXHg2ueWj2lDWbtVxQbn/m7aSqON
-  9YN5xfXY36tpVp40RV1J36FUskkhgc/DZcgEMYdAr2XrjDS1A0TnEaDatQUgYgpd
-  J0oP9V+2FMfDFvIhX5tNrEuIIFMyO+HR0wxV7huTUeus4knyXBZur3If+g==
-  -----END CERTIFICATE-----
+# Set to true if setting Harbor Registry as insecure-registries for docker
+insecureRegistry: false
+# Need to specify your own FQDN for Harbor
+externalDomain: harbor.my.domain
```
paulczar: I like setting this to `harbor.192.168.99.100.xip.io` which should allow it to work without change for a default minikube install.

contrib/helm/harbor/values.yaml
```diff
@@ -28,89 +28,41 @@
 persistence:
   enabled: false
 
-externalDomain: harbor.192.168.99.100.xip.io
-## tls_crt, tls_key, ca_crt should match the domain above
+# The tag for Harbor docker images
+harborImageTag: &harbor_image_tag v1.4.0
```
paulczar: I don't like having a common tag like this, especially if it can't be used for _all_  ( see `tag: v2.6.2-v1.4.0` ).  Also eventually we should look to break out each component into its own helm chart which would then require each service to have its own tag anyways.

contrib/helm/harbor/values.yaml
```diff
@@ -28,89 +28,41 @@
...
+insecureRegistry: false
+# Need to specify your own FQDN for Harbor
+externalDomain: harbor.my.domain
+# The common name of tlsCrt, tlsKey, caCrt must match the externalDomain above
+tlsCrt:
+tlsKey:
+caCrt:
+
+# The secret key used for encryption. Must be a string of 16 chars.
+secretKey: not-a-secure-key
```
paulczar: I would rather see each component have its own secret key and document that it should be the same across all of the components.  Eventually when we break them out into their own helm charts we would need this anyway.

contrib/helm/harbor/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 name: harbor
-version: 0.0.1
-appVersion: 1.3.0
-description: An Enterprise-class Docker Registry Harbor by VMware
+version: 0.2.0
```
paulczar: we shouldn't jump a minor version here.  it should either go to `0.1.0` or `1.0.0`.

contrib/helm/harbor/README.md
jessehu: My minikube doesn't have the IP 192.168.99.100, but a public IP.

contrib/helm/harbor/README.md
jessehu: Using '--vm-driver=none' to keep it simple to set up.

contrib/helm/harbor/README.md
jessehu: Yes, I think helm chart users should know how to use '--set' and values.yaml.

jessehu: I'd prefer not adding '$' for a command.

contrib/helm/harbor/templates/ingress/secret.yaml
jessehu: The code below uses `.Values.tlsCrt` if it's set by users, and uses the generated values if `.Values.tlsCrt` is not set.

  tls.crt: {{ .Values.tlsCrt | default $cert.Cert | b64enc | quote }}
  tls.key: {{ .Values.tlsKey | default $cert.Key | b64enc | quote }}
  ca.crt: {{ .Values.caCrt | default $ca.Cert | b64enc | quote }}

contrib/helm/harbor/values.yaml
jessehu: My minikube install doesn't have IP 192.168.99.100. And we should not assume minikube is the default env.  We strongly suggest the end users to specific the harbor FQDN via externalDomain.

contrib/helm/harbor/values.yaml
jessehu: There is not a nice approach to support generating tag 'v2.6.2-v1.4.0'. Maybe we can specific a base tag 'v2.6.2' and concat('v2.6.2', '-', .Values.harborImageTag).
Will keep current design for now.

contrib/helm/harbor/values.yaml
jessehu: this secretKey must be the same for all harbor components. It's different from the secret.
I want to prevent the potential errors when users update the secretKey for every components,

contrib/helm/harbor/Chart.yaml
jessehu: Will use 0.1.0.

contrib/helm/harbor/README.md
```diff
@@ -89,36 +143,32 @@ The following tables lists the configurable parameters of the Percona chart and
 | `mysql.resources` | [resources](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/) to allocate for container   | undefined |
 | `mysql.volumes` | used to create PVCs if persistence is enabled (see instructions in values.yaml) | see values.yaml |
 | **Registry** |
-| `registry.image.repository` | Repository for registry image | `vmware/harbor-registry` |
-| `registry.image.tag` | Tag for registry image | `v1.3.0` |
-| `registry.image.pullPolicy` | Pull Policy for registry image | `IfNotPresent` |
+| `registry.image.repository` | Repository for registry image | `vmware/registry-photon` |
+| `registry.image.tag` | Tag for registry image | `v2.6.2-v1.4.0` |
+| `registry.image.pullPolicy` | Pull Policy for admregistryinserver image | `IfNotPresent` |
```
reasonerjt: typo `admregistryinserver`

contrib/helm/harbor/README.md
```diff
@@ -30,26 +81,29 @@ The command deploys Harbor on the Kubernetes cluster in the default configuratio
...
 | Parameter                  | Description                        | Default                 |
 | -----------------------    | ---------------------------------- | ----------------------- |
 | **Harbor** |
-| `externalDomain`               | domain harbor will run on (https://*harbor.url*/) |`harbor.192.168.99.100.xip.io`   |
-| `tls_crt`               | TLS certificate to use for Harbor's https endpoint | see values.yaml |
-| `tls_key`               | TLS key to use for Harbor's https endpoint | see values.yaml |
-| `ca_crt`               | CA Cert for self signed TLS cert | see values.yaml |
+| `harborImageTag`     | The tag for Harbor docker images | `v1.4.0` |
+| `externalDomain`       | Harbor will run on (https://*externalDomain*/). Make sure this FQDN resolves to the K8s Ingress Controller IP. | `harbor.my.domain` |
+| `insecureRegistry`     | Set to true if setting Harbor Registry as insecure-registries for docker | `false` |
```
reasonerjt: Let's explicitly say if `insecureRegistry` is set to true, the tls cert of ingress controller will not be set by helm.

contrib/helm/harbor/templates/adminserver/adminserver-cm.yaml
```diff
@@ -45,3 +51,12 @@ data:
...
   CLAIR_DB_HOST: "{{ template "harbor.fullname" . }}-clair-pg"
+  CLAIR_DB_PORT: "5432"
+  CLAIR_DB: "postgres"
+  CLAIR_DB_USERNAME: "postgres"
+  CLAIR_DB_PASSWORD: "{{ .Values.clair.postgresPassword }}"
+  UAA_ENDPOINT: ""
+  UAA_CLIENTID: ""
+  UAA_CLIENTSECRET: ""
+  UAA_VERIFY_CERT: "True"
+  REGISTRY_STORAGE_PROVIDER_NAME: "filesystem"
```
reasonerjt: reference the setting in values.yaml?

reasonerjt:
An overall comment is that some components such as `adminserver` `registry` may not need to be stateful set (considering we use `PV` to store the data), but we can revisit and change that in another change.

contrib/helm/harbor/README.md
paulczar: I disagree that this is simple, we should expect anyone using minikube to use it with its default behavior, if people want to use `--vm-driver=none` they're probably an expert user that will know how to deal with the changes.

contrib/helm/harbor/README.md
paulczar: I assume that's because you're using `--vm-driver=none`.

contrib/helm/harbor/README.md
paulczar: sure, but we shouldn't tell them to do both.  If they want to learn helm they should read the helm instructions.

contrib/helm/harbor/templates/ingress/secret.yaml
paulczar: I understand, but we shouldn't generate certs if they're not needed, unless the logic to not do so is too complicated to do in a helm template.

contrib/helm/harbor/values.yaml
paulczar: understood.  I'm okay with this approach for now, even though I think long term each should have it set seperately to allow for breaking each application out into its own independent helm chart.

contrib/helm/harbor/README.md
paulczar: without putting some sort of marker to show the command prompt like `$`, how does the user know what is input vs output ?  Currently our instructions are fairly simple and only show the output, but we may at some stage want to show the full output of a command ( example the `helm install` output of the NOTES.txt ).

jessehu: I understand your point.

contrib/helm/harbor/README.md
jessehu: I'm not an expert of minikube, when reading the minikube doc (https://github.com/kubernetes/minikube#linux-continuous-integration-without-vm-support), it tells me to run 'minikube start --vm-driver=none' on Linux, which is very easy and straightforward for me to start a k8s cluster.

PR #4377, opened 2018-03-08
migrate maria to postgresql

wy65701436:


paulczar:
Is there a design goal to switch from MySQL based to Postgres? This would be a fairly major breaking change that would affect anyone currently running Harbor when they go to upgrade.  We would need to ensure a smooth upgrade process including data migration scripts ( or at least documentation ) from one to the other.

With the potential disruption to existing users, I think we'd need a really good reason to do this as a straight cutover like this, may make more sense to support both for a period of time to give people a chance to move their data.

yixingjia:
@paulczar Thanks a lot for you comment and suggestions.
We plan to migrate to Postgresql due to several reasons.

1> To limit the maintenance effort for our customer.
Currently Harbor has both Mariadb, postgresql, ( For old version of Harbor there are Mysql as well)

2> We have plan to Support HA deployment of Harbor. Setting up different DB clusters are expensive. 

This Patch is only for database driver change.
In the coming patches, we will surely provide data migration scripts for user to transfer from MariaDB to Postgresql. 
We will also document the detail/solid steps for this part. So please don't worry about the upgrade.

Please let us know if you have further more questions/concerns.  

As always thanks a lot for choose Harbor.

paulczar:
:+1:

src/common/dao/base.go
```diff
@@ -87,14 +87,13 @@ func InitDatabase(database *models.Database) error {
...
 func getDatabase(database *models.Database) (db Database, err error) {
 	switch database.Type {
-	case "", "mysql":
-		db = NewMySQL(database.MySQL.Host,
-			strconv.Itoa(database.MySQL.Port),
-			database.MySQL.Username,
-			database.MySQL.Password,
-			database.MySQL.Database)
-	case "sqlite":
-		db = NewSQLite(database.SQLite.File)
```
yixingjia: Is this "sqlite" section removed by mistake?

yixingjia: Ok, seem we will not support sqlite

src/common/dao/pgsql.go
```diff
@@ -24,7 +24,7 @@ import (
 
 type pgsql struct {
 	host     string
-	port     int
+	port     string
```
yixingjia: any reason we change it from int to string?

src/common/dao/testutils.go
```diff
@@ -60,23 +68,7 @@ func PrepareTestForMySQL() {
...
-		log.Fatalf("environment variable SQLITE_FILE is not set")
-	}
-
-	database := &models.Database{
-		Type: "sqlite",
-		SQLite: &models.SQLite{
-			File: file,
-		},
-	}
+	log.Infof("POSTGRES_HOST: %s, POSTGRES_USR: %s, POSTGRES_PORT: %d, POSTGRES_PWD: %s\n", dbHost, dbUser, dbPort, dbPassword)
```
yixingjia: Let's try to avoid log password in any place.

src/common/dao/base.go
wy65701436: yes, will not support sqlite and maria.

PR #4388, opened 2018-03-09
Add new filter field for user api

aladjadj:
add capability to retrieve current admins on harbor api


aladjadj:
I added positive coverage.

PR #4389, opened 2018-03-09
update helm chart to use external postgresql dependency

paulczar:


contrib/helm/harbor/values.yaml
```diff
@@ -280,3 +275,13 @@ clair:
 #
 notary:
   enabled: false
+
+## Settings for postgresql dependency.
+## see https://github.com/kubernetes/charts/tree/master/stable/postgresql
+## for further configurables.
+postgresql:
+  postgresUser: clair
+  postgresPassword: not-a-secure-password
```
jessehu: Thanks Paul for this patch!  Could we link clair.postgresPassword to postgresql.postgresPassword in values.yaml because they must be the same ?

paulczar: I specifically didn't do that because there's a good chance harbor is changing its database to postgres too and we probably want clair and harbor to have different database/username/password to postgres.

I can document them better in the values.yaml file to make it more clear that they need to match.

jessehu: Didn't catch your point on "we probably want clair and harbor to have different database/username/password to postgres." 
Code is always better than doc when the code doesn't take big extra effort.
The clair database username/password/schema must match the username/password/schema specified for postgresql chart, then we'd better reference the former to latter. Is there other reason you don't prefer using 'password: &clair_db_password xxx' in values.yaml ?
```
clair:
  enabled: true
  image:
    repository: vmware/clair-photon
    tag: v2.0.1-v1.4.0
    pullPolicy: IfNotPresent
  postgresPassword: not-a-secure-password
  postgresUser: clair
postgresDatabase: clair
```

jessehu: We'd better not expect the end users to read the doc in clair section when he modifies password in postgresql section.

paulczar: have you got docs on using 'password: &clair_db_password xxx' ?  I haven't seen it in other charts and want to make sure I understand it fully before I use it.

paulczar: I may also update it to be better at providing options for external database vs the depdendency,  something like the way wordpress does it ( https://github.com/kubernetes/charts/blob/master/stable/wordpress/values.yaml ) combined with your pointer trick above (I may chat with helm community about that first and make sure its a viable path for helm charts in general since  i don't see it in existing charts I've looked at )

paulczar: ahhh YAML anchors.  I learn something today :)

paulczar: Spoke to a few of the core helm chart maintainers, and they suggest we should avoid YAML anchors as they're only processed once when values.yaml are read and are not re-evaluated when using `--set` and `--values` to override.  I'll update this to look more like how the stable/wordpress chart handles external DB.

jessehu:
Thanks a lot Paul. For “not re-evaluated when using --set and --values to override“, the yaml anchor in values.yaml works when I used "--set externalDomain=xxx". What's the specific case they meant?



-Jesse
sent from my miPhone
在 Paul Czarkowski <notifications@github.com>，2018年3月13日 下午10:41写道：

@paulczar commented on this pull request.

________________________________

In contrib/helm/harbor/values.yaml<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_vmware_harbor_pull_4389-23discussion-5Fr174155689&d=DwMCaQ&c=uilaK90D4TOVoH58JNXRgQ&r=5_CTlBx8_0Z_zmXyLKLlQg&m=lpVs4A9pKMNkYrcIW-0isbvuogSOhpCXsBLM7zpYb3s&s=YWTKBFUuMuURth5LKopR7D2V4mLzEqdO1Uejz5SPI-E&e=>:

> @@ -280,3 +275,13 @@ clair:
 #
 notary:
   enabled: false
+
+## Settings for postgresql dependency.
+## see https://github.com/kubernetes/charts/tree/master/stable/postgresql
+## for further configurables.
+postgresql:
+  postgresUser: clair
+  postgresPassword: not-a-secure-password


Spoke to a few of the core helm chart maintainers, and they suggest we should avoid YAML anchors as they're only processed once when values.yaml are read and are not re-evaluated when using --set and --values to override. I'll update this to look more like how the stable/wordpress chart handles external DB.

paulczar:
the specific use case is if the user specified `--set harborImageTag=v1.4.1` the values like `adminserver.image.tag` would remain as `v1.4.0`.

jessehu:
I see. I would argue it's a helm bug. Maybe we can use harborImageTag='* harbor_image_tag v1.4.1', or uses values. yml instead.
That's why so many feature requests on support  templating in values.yml



-Jesse
sent from my miPhone
在 Paul Czarkowski <notifications@github.com>，2018年3月13日 下午11:20写道：

the specific use case is if the user specified --set harborImageTag=v1.4.1 the values like adminserver.image.tag would remain as v1.4.0.

jessehu:
Thanks @paulczar for helping on the issue and deep dive. I'm approving this PR and let's address the password alignment and yaml anchor issue once there is a best solution.

jessehu:
Hi @paulczar , there is bug fix in our CI 2 days ago, so the CI for this old PR can not pass.
Could you please create a new PR with the same code ? Then CI can pass.
BTW suggest change the commit subject to "Let helm chart create clair database via postgresql chart".

jessehu:
Hi @paulczar, Could you please create a new PR with the same code in this PR ? So we can get the CI pass.

reasonerjt:
The code change in this PR is for Helm Chart only, low risk, I'll merge the PR.

jessehu:
Thanks @reasonerjt to merge it manually. @paulczar we can make improvement in future patch.

PR #4392, opened 2018-03-11
Rest API for project group member

stonezdj:
1.  Add REST API for creating user group (currently only support LDAP group)
2. Add REST API for assign role to the user group


docs/swagger.yaml
```diff
@@ -626,7 +626,161 @@ paths:
         '404':
           description: Project ID does not exist.
         '500':
-          description: Unexpected internal errors.
+          description: Unexpected internal errors.©
```
reasonerjt: There's an extra character.

docs/swagger.yaml
```diff
@@ -626,7 +626,161 @@ paths:
...
+     parameters:
+      - name: project_id
+        in: path
+        type: integer
+        format: int64
+        required: true
+        description: Relevant project ID.
+     tags:
+      - Products
+     responses:
```
reasonerjt: There should be the definition of the response body.

docs/swagger.yaml
```diff
@@ -626,7 +626,161 @@ paths:
...
+      '200':
+        description: Project role members deleted successfully.
```
reasonerjt: description is incorrect.

docs/swagger.yaml
```diff
@@ -626,7 +626,161 @@ paths:
...
+        required: true
+        description: Relevant project ID.
+      - name: projectmember
+        in: body
+        description: Project Member    
+        schema:
+         $ref: '#/definitions/ProjectMember'
+     responses:
+        '200':
+          description: Project member deleted successfully.
```
reasonerjt: description is incorrect.

docs/swagger.yaml
```diff
@@ -626,7 +626,161 @@ paths:
...
+          description: User need to log in first.
+        '403':
+          description: User in session does not have permission to the project.
+        '404':
+          description: Project or projet member does not exist.
+        '500':
+          description: Unexpected internal errors.  
+    put:
+     summary: Update project member
+     description: Create project member relationship
```
reasonerjt: I think it should be used for updating role?

docs/swagger.yaml
```diff
@@ -626,7 +626,161 @@ paths:
...
+      - Products
+     responses:
+      '200':
+        description: Project role members deleted successfully.
+      '400':
+        description: Illegal format of provided ID value.
+      '401':
+        description: User need to log in first.
+      '403':
+        description: User in session does not have permission to the project.
```
reasonerjt: For the 403, it would be better to specify what permission is needed.

docs/swagger.yaml
```diff
@@ -3241,3 +3538,44 @@ definitions:
...
+  ProjectMember:
+    type: object
+    properties:
+      entity_id: 
+        type: integer
+        description: The ID of the member
+      entity_type: 
+        type: string
+        description: The type of the entity, u for user, g for group member
+      role_id:
```
reasonerjt: Why call it `role_id` instead of `role` ?  Is it to avoid conflict with reserved word?

docs/swagger.yaml
```diff
@@ -2122,6 +2276,149 @@ paths:
...
+          description: User group deleted successfully.
+        '400':
+          description: Illegal format of provided ID value.
+        '401':
+          description: User need to log in first.
+        '403':
+          description: User in session does not have permission to the user group.
+        '500':
+          description: Unexpected internal errors.        
+  /usergroups/{group_id}:
```
reasonerjt: If we treat group as a resource, it would be weird if we don't return group members in get.

src/ui/config/config.go
```diff
@@ -205,6 +206,35 @@ func LDAPConf() (*models.LdapConf, error) {
...
 
+// LDAPGroupConf returns the setting of ldap group search
+func LDAPGroupConf() (*models.LdapGroupConf, error) {
+
+	cfg, err := mg.Get()
+	if err != nil {
+		return nil, err
+	}
+
+	ldapGroupConf := &models.LdapGroupConf{LdapGroupSearchScope: 2}
```
reasonerjt: Use a const instead of `2`?

docs/swagger.yaml
```diff
@@ -2122,6 +2276,149 @@ paths:
...
+      '401':
+        description: User need to log in first.
+      '403':
+        description: User in session does not have permission to the user group.
+      '404':
+        description: user group does not exist.
+      '500':
+        description: Unexpected internal errors.  
+   put:
+    summary: Update group information
```
reasonerjt: Shall we allow updating group name when the group is onboarded from LDAP?

docs/swagger.yaml
```diff
@@ -3241,3 +3538,44 @@ definitions:
       update_time:
         type: string
         description: The update time of label.
+  ProjectMember:
+    type: object
+    properties:
+      entity_id: 
+        type: integer
+        description: The ID of the member
```
reasonerjt: This is the ID of the user or group, right?

docs/swagger.yaml
```diff
@@ -3241,3 +3538,44 @@ definitions:
...
+      entity_type: 
+        type: string
+        description: The type of the entity, u for user, g for group member
+      role_id: 
+        type: integer
+        description: The role id 1 for projectAdmin, 2 for developer, 3 for guest
+      ldap_user_name:
+        type: string
+        description: the LDAP username need to assign role
+      ldap_group_dn:
```
reasonerjt: How'bout follow the "search and onboard" approach?

src/common/dao/group/usergroup.go
```diff
@@ -0,0 +1,96 @@
...
+		return nil, err
+	}
+	return &userGroup, nil
+}
+
+// DeleteUserGroup ...
+func DeleteUserGroup(id int) error {
+	userGroup := models.UserGroup{ID: id}
+	o := dao.GetOrmer()
+	_, err := o.Delete(&userGroup)
```
reasonerjt: so there won't be foreign key issue in members table?

src/common/dao/group/usergroup.go
```diff
@@ -0,0 +1,96 @@
...
+	return err
+}
+
+// UpdateUserGroup ...
+func UpdateUserGroup(id int, groupName string) error {
```
reasonerjt: Will we allow updating LDAP group's group name?

src/common/dao/project/projectmember.go
```diff
@@ -0,0 +1,193 @@
...
+	var member models.Member
+	member.EntityID = memberReq.EntityID
+	member.EntityType = memberReq.EntityType
+	member.Role = memberReq.Role
+	member.ProjectID = memberReq.ProjectID
+
+	memberReq.LdapUserName = strings.TrimSpace(memberReq.LdapUserName)
+	memberReq.LdapGroupDN = strings.TrimSpace(memberReq.LdapGroupDN)
+	if len(memberReq.LdapUserName) > 0 {
+		userID, err := auth.SearchAndOnboardUser(memberReq.LdapUserName)
```
reasonerjt: This is incorrect, I see why there is a project pkg, but IMO the search and onboard should be done at API level, not dao pkg.

src/common/dao/project/projectmember.go
```diff
@@ -0,0 +1,193 @@
...
+
+	err = o.Raw(querySQL, member.ProjectID, member.EntityID, member.EntityType).QueryRow(&pmid)
+	if err != nil {
+		return 0, err
+	}
+	return pmid, err
+}
+
+// UpdateProjectMember updates the record in table project_member, only role can be changed
+func UpdateProjectMember(pmID int, role int) error {
```
reasonerjt: How about call it update project member role.

src/common/dao/project/projectmember.go
```diff
@@ -0,0 +1,193 @@
...
+	if role <= 0 {
```
reasonerjt: should also check the role <= 3

PR #4396, opened 2018-03-12
Implement adding/removing labels to/from repositories and images API

ywk253100:


docs/swagger.yaml
```diff
@@ -995,6 +995,82 @@ paths:
...
+          required: true
+          description: The name of repository.
+      tags:
+        - Products
+      responses:
+        '200':
+          description: Successfully.
+        '401':
+          description: Unauthorized.
+        '403':
```
reasonerjt: It would be helpful to specify what permission is needed under `403`

make/photon/db/registry.sql
```diff
@@ -272,6 +272,20 @@ create table harbor_label (
  CONSTRAINT unique_name_and_scope UNIQUE (name,scope)
  );
 
+create table harbor_resource_label (
+ id int NOT NULL AUTO_INCREMENT,
+ label_id int NOT NULL,
+ resource_id varchar(256) NOT NULL,
```
reasonerjt: Could you specify what the `resource_id` would look like for different `resource_type`?

docs/swagger.yaml
```diff
@@ -1075,6 +1151,97 @@ paths:
               $ref: '#/definitions/DetailedTag'
         '500':
           description: Unexpected internal errors.
+  '/repositories/{repo_name}/tags/{tag}/labels':
+    get:
```
reasonerjt: schema of the response?

docs/swagger.yaml
```diff
@@ -1075,6 +1151,97 @@ paths:
...
+          in: path
+          type: string
+          required: true
+          description: The name of repository.
+        - name: tag
+          in: path
+          type: string
+          required: true
+          description: The tag of the image.
+        - name: label_id
```
reasonerjt: This is the `resource_label_id` right? not the `label_id`.

src/ui/router.go
```diff
@@ -70,7 +70,11 @@ func initRouters() {
 	beego.Router("/api/repositories", &api.RepositoryAPI{}, "get:Get")
 	beego.Router("/api/repositories/scanAll", &api.RepositoryAPI{}, "post:ScanAll")
 	beego.Router("/api/repositories/*", &api.RepositoryAPI{}, "delete:Delete;put:Put")
+	beego.Router("/api/repositories/*/labels", &api.RepositoryLabelAPI{}, "get:GetOfRepository;post:AddToRepository")
+	beego.Router("/api/repositories/*/labels/:id([0-9]+", &api.RepositoryLabelAPI{}, "delete:RemoveFromRepository")
```
reasonerjt: there seems an extra `(`?
same with line 77

docs/swagger.yaml
ywk253100: It's `label_id`.

src/ui/api/repository_label.go
```diff
@@ -0,0 +1,248 @@
...
+		r.HandleInternalServerError(fmt.Sprintf("failed to add label %d to resource %s %s: %v",
+			rl.LabelID, rl.ResourceType, rl.ResourceID, err))
+		return
+	}
+
+	// return the ID of label and return status code 200 rather than 201 as the label is not created
+	r.Redirect(http.StatusOK, strconv.FormatInt(rl.LabelID, 10))
+}
+
+func (r *RepositoryLabelAPI) removeLabel(rl *models.ResourceLabel) {
```
reasonerjt: The resource_label should be removed if the label is deleted.

docs/swagger.yaml
reasonerjt: I don't fully agree but given this is how github works we can leave this implement as-is.

src/ui/api/repository_label.go
ywk253100: Open #4410 to track this, will submit another PR to fix it.

PR #4415, opened 2018-03-15
Add DAO for user group and project member

stonezdj:
Add DAO for user group and project member

### DIFF
--- a/src/common/dao/group/usergroup.go
+++ b/src/common/dao/group/usergroup.go
@@ -1,0 +1,94 @@
+// Copyright (c) 2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package group
+
+import (
+	"github.com/vmware/harbor/src/common/dao"
+	"github.com/vmware/harbor/src/common/models"
+	"github.com/vmware/harbor/src/common/utils/log"
+)
+
+// AddUserGroup - Add User Group
+func AddUserGroup(userGroup models.UserGroup) (int, error) {
+	o := dao.GetOrmer()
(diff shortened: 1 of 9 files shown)


src/common/dao/group/usergroup.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/vmware/harbor/src/common/dao"
+	"github.com/vmware/harbor/src/common/models"
+	"github.com/vmware/harbor/src/common/utils/log"
+)
+
+// AddUserGroup - Add User Group
+func AddUserGroup(userGroup models.UserGroup) (int, error) {
+	o := dao.GetOrmer()
+	query := models.UserGroup{GroupType: userGroup.GroupType, LdapGroupDN: userGroup.LdapGroupDN}
+	result, err := QueryUserGroup(query)
```
ywk253100: Maybe we should move the query logic to the API layer as we need to return 409?

stonezdj: Updated

src/common/dao/project/projectmember.go
```diff
@@ -0,0 +1,121 @@
...
+		return 0, fmt.Errorf("Invalid project_id, member: %+v", member)
+	}
+
+	sql := "insert into project_member (project_id, entity_id , role, entity_type) values (?, ?, ?, ?)"
+	_, err := o.Raw(sql, member.ProjectID, member.EntityID, member.Role, member.EntityType).Exec()
+	if err != nil {
+		return 0, err
+	}
+	var pmid int
+	querySQL := "select id from project_member where project_id = ? and entity_id = ? and entity_type = ? limit 1"
```
ywk253100: We can use sql.Result returned by Exec() to get the ID. Or use the beego ORM directly.

src/common/dao/project/projectmember.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return 0, err
+	}
+	return pmid, err
+}
+
+// UpdateProjectMemberRole updates the record in table project_member, only role can be changed
+func UpdateProjectMemberRole(pmID int, role int) error {
+	if role <= 0 || role >= 3 {
+		return fmt.Errorf("Failed to update project member role < 0, role:%v", role)
```
ywk253100: The error message should also contain role >= 3?

src/common/dao/project/projectmember.go
stonezdj: Updated

src/common/dao/project/projectmember.go
stonezdj: Updated

PR #4419, opened 2018-03-15
Implement filter repository and tags by label API

ywk253100:


src/common/dao/repository_test.go
```diff
@@ -32,61 +32,68 @@ var (
...
+	n, err := GetTotalOfRepositories()
+	require.Nil(t, err)
+	assert.Equal(t, total+1, n)
+}
 
-	exist := false
-	for _, repo := range repositories {
-		if repo.Name == name {
-			exist = true
+func TestGetRepositories(t *testing.T) {
```
reasonerjt: labelID is not covered.

src/common/dao/repository.go
```diff
@@ -124,46 +103,79 @@ func GetTopRepos(projectIDs []int64, n int) ([]*models.RepoRecord, error) {
...
+
+	if len(q.ProjectName) > 0 {
+		sql += `and p.name = ? `
+		params = append(params, q.ProjectName)
 	}
-	_, err := qs.All(&repositories)
 
-	return repositories, err
+	if len(q.LabelIDs) > 0 {
+		sql += fmt.Sprintf(`and rl.label_id in ( %s ) `,
```
reasonerjt: does it mean the repo will be returned if it has one of the labels in the query parm?

src/common/models/repo.go
```diff
@@ -37,3 +37,12 @@ type RepoRecord struct {
...
 	return RepoTable
 }
+
+// RepositoryQuery : query parameters for repository
+type RepositoryQuery struct {
+	Name        string
+	ProjectIDs  []int64
+	ProjectName string
+	LabelIDs    []int64
+	Pagination
```
reasonerjt: The usage of `Pagination` in `RepositoryQuery` and `Label` is not consistent with other models, such as `project` `user`

src/ui/api/repository.go
```diff
@@ -377,14 +382,19 @@ func (ra *RepositoryAPI) GetTag() {
 	}
 
 	result := assembleTags(client, repository, []string{tag},
-		ra.SecurityCtx.GetUsername())
+		ra.SecurityCtx.GetUsername(), nil)
```
reasonerjt: Why the labels is `nil`?

src/ui/api/statistic.go
```diff
@@ -64,17 +64,22 @@ func (s *StatisticAPI) Get() {
 	}
 
 	statistic[PubPC] = (int64)(len(pubProjs))
```
reasonerjt: This line should also be removed?

ywk253100: It's the count of public projects, it's needed.

PR #4423, opened 2018-03-16
Provide migration scripts for harbor.cfg

reasonerjt:
Default target version is 1.5.0
This is mainly for VIC-appliance upgrade, and should be considered
experimental for oss due to limited test.
Tested with 1.2 and 1.3 harbor.cfg from VIC appliance.


make/harbor.cfg
```diff
@@ -1,5 +1,7 @@
 ## Configuration file of Harbor
 
+#This attribute is for migrator to detect the version of the .cfg file, DO NOT MODIFY!
+_version = 1.5.0
```
wy65701436: let's use the pass-by value instead of adding an new key in cfg.

reasonerjt:
The change is in /tools/migration, no risk to impact CI or integration, I'll merge it now.

PR #4433, opened 2018-03-19
add read only mode to stop docker push

wy65701436:


src/ui/proxy/proxy.go
```diff
@@ -41,7 +41,7 @@ func Init(urls ...string) error {
 		return err
 	}
 	Proxy = httputil.NewSingleHostReverseProxy(targetURL)
-	handlers = handlerChain{head: urlHandler{next: listReposHandler{next: contentTrustHandler{next: vulnerableHandler{next: Proxy}}}}}
+	handlers = handlerChain{head: readonlyHandler{next: urlHandler{next: listReposHandler{next: contentTrustHandler{next: vulnerableHandler{next: Proxy}}}}}}
```
reasonerjt: Doesn't need to be addressed in this PR, but we should consider refactor it  to like 
`handlerChain.append(handlers...)`

PR #4466, opened 2018-03-22
Enhance registry docker file to make it work with NFS server

jessehu:


make/photon/registry/entrypoint.sh
```diff
@@ -7,10 +7,16 @@ if [ -d /etc/registry ]; then
 fi
 if [ -d /var/lib/registry ]; then
     chown 10000:10000 -R /var/lib/registry
-fi  
+fi
 if [ -d /storage ]; then
-    chown 10000:10000 -R /storage
-fi  
+    if ! ls -ld /storage | grep 'harbor harbor' ; then
```
reasonerjt: How about using `stat -c "%u"` ?

jessehu: sure, we can use 'stat -c "%u:%g" /data/registry'

PR #4472, opened 2018-03-23
Add Redis docker file

yixingjia:
Add the redis docker files


reasonerjt:
I think we need to make sure the uid and gid is 10000

yixingjia:
@reasonerjt  no need since it will run as redis user.

PR #4474, opened 2018-03-23
Update SQL script to add uuid to job tables.

reasonerjt:
We have to add the uuid/id mapping as new job service will only store uuid.

Further work is in feature branch for now, commit this change to
accelerate migration work.


ninjadq:
LGTM

PR #4504, opened 2018-03-27
Integrate with status hook of jobservice

reasonerjt:


reasonerjt:
Only error in travis-ci is due to ut failure in jobservice pkg, merging this PR.

PR #4511, opened 2018-03-28
add all in one migrator scripts

wy65701436:


tools/migration/migrator.py
```diff
@@ -0,0 +1,178 @@
...
+
+    db_migrator = DBMigrator(commandline_input.target_version)
+    cfg_migrator = CfgMigrator(commandline_input.target_version)
+
+    try:
+        # test
+        if commandline_input.action == "test":
+            if commandline_input.is_migrate_db:
+                if not db_migrator.validate():
+                    print ("Fail to validate: please make sure your cfg path is correct.")
```
reasonerjt: The error message is incorrect.

tools/migration/migrator.py
```diff
@@ -0,0 +1,178 @@
...
+        last_argv = sys.argv[argv_len:][0]
+        if not self.is_action(last_argv):
+            print ("Fail to parse input: the last parameter should in test:up:restore:backup")
+            sys.exit(RC_GEN) 
+
+        if argv_len == 1:
+            return (True, True, '', last_argv)  
+        
+        parser = argparse.ArgumentParser(description='migrator of harbor') 
+        parser.add_argument('--db', action="store_true", dest='is_migrate_db', required=False, default=False, help='The falg to upgrede db.')
```
reasonerjt: upgrede -> upgrade.

tools/migration/migrator.py
```diff
@@ -0,0 +1,178 @@
...
+import argparse
+
+RC_VALIDATE = 101
+RC_UP = 102
+RC_DOWN = 103
+RC_BACKUP = 104
+RC_RESTORE = 105
+RC_UNNKNOW_TYPE = 106
+RC_GEN = 110
+
```
reasonerjt: Let's move the logic of confirm from shell of DB migrator to this .py file.

tools/migration/migrator.py
```diff
@@ -0,0 +1,178 @@
...
+            return False
+        try:
+            copyfile(self.backup_path+"/harbor.cfg", self.cfg_path)
+            print ("Success to restore harbor.cfg.")
+            return True
+        except Exception, e:
+            print ("Restore error: %s" % str(e))
+            return False
+
+    def up(self):
```
reasonerjt: check if self.cfg_path exists, if not, print some message and skip.

PR #4524, opened 2018-03-29
Bypass docker client retry on push in readonly mode

wy65701436:
The output is,

The push refers to a repository [myregistry/library/busybox]
123abc123abc: Preparing 
denied: Upload/Delete is prohibited in read only mode.

### DIFF
--- a/src/ui/filter/readonly.go
+++ b/src/ui/filter/readonly.go
@@ -42,6 +42,7 @@ func filter(req *http.Request, resp http.ResponseWriter) {
 	}
 	if matchRepoTagDelete(req) {
 		resp.WriteHeader(http.StatusServiceUnavailable)
+		resp.Write([]byte("The system is in read only mode. Any modification is not prohibited."))
 	}
 }
 
--- a/src/ui/proxy/interceptors.go
+++ b/src/ui/proxy/interceptors.go
@@ -35,7 +35,7 @@ const (
 var rec *httptest.ResponseRecorder
 
 // NotaryEndpoint , exported for testing.
-var NotaryEndpoint =""
+var NotaryEndpoint = ""
 
 // MatchPullManifest checks if the request looks like a request to pull manifest.  If it is returns the image and tag/sha256 digest as 2nd and 3rd return values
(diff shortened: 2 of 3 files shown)


src/ui/filter/readonly.go
```diff
@@ -42,6 +42,7 @@ func filter(req *http.Request, resp http.ResponseWriter) {
 	}
 	if matchRepoTagDelete(req) {
 		resp.WriteHeader(http.StatusServiceUnavailable)
```
jessehu: Should it be 403?

src/ui/proxy/interceptors.go
```diff
@@ -161,7 +161,7 @@ type readonlyHandler struct {
 func (rh readonlyHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	if config.ReadOnly() {
 		if req.Method == http.MethodDelete || req.Method == http.MethodPost || req.Method == http.MethodPatch {
-			http.Error(rw, "Upload/Delete is prohibited in read only mode.", http.StatusServiceUnavailable)
+			http.Error(rw, marshalError("DENIED", "Upload/Delete is prohibited in read only mode."), http.StatusForbidden)
```
jessehu: Can we extract this msg string "Upload/Delete is prohibited in read only mode." into a const ? Because it appears in two files in this commit.
And how about using  "The system is in read only mode. Any modification is not prohibited."

src/ui/filter/readonly.go
wy65701436: actually, 403 is a workaround here. Docker defined several DoNoRetry errors, 403 is one of them, and no 503 in.

src/ui/proxy/interceptors.go
wy65701436: The message could be refine.

src/ui/filter/readonly.go
reasonerjt: yep, conventionally when the system set to a mode that some features are disabled, the error should be `503`

PR #4535, opened 2018-03-29
Return zero when input version is same as target

reasonerjt:


tools/migration/cfg/migrate.py
```diff
@@ -24,9 +25,12 @@ def main():
     input_version =  utils.get_conf_version(args.input_path)
     curr_dir = os.path.dirname(__file__)
     chain = []
+    if input_version == target_version:
+        print ("version of input cfg is identical to target:", input_version)
```
jessehu: suggestion: "The target version and source version in harbor.cfg  is the same: %s. No need to upgrade.", input_version

tools/migration/cfg/migrate.py
```diff
@@ -24,9 +25,12 @@ def main():
     input_version =  utils.get_conf_version(args.input_path)
```
jessehu: 1) input_version could be 'unsupported', but seems we don't handle this case here.
2) Better to ensure target_version > input_version to do migration.
3) The logic in get_conf_version() is a little cool. Can we check the greatest tag of a harbor container image (e.g. harbor-log) to get current installed version?

reasonerjt: for 1) and 2) currently it's all handled in `search`, which tries to find a handler chain to migrate the .cfg file from source to target.  So both cases will fail as expected, if you are suggesting refining error messages, those are cosmetic change can be made in future.

for 3) For integration with VIC appliance, this script has to be executed in a container to which the harbor.cfg is single source of truth.

jessehu: Got it. Can we handle "input_version == target_version" in search too ?

jessehu: I think a message is needed to indicate why upgrade is not needed, at least print out the source version and target version.

tools/migration/cfg/migrate.py
reasonerjt: Will remove this since we'll handle the `==` case in `search`

tools/migration/cfg/migrate.py
reasonerjt: So then let's keep it as is for now.

PR #4545, opened 2018-03-30
Add contributers list for Harbor Chart

jessehu:


paulczar:
LGTM, although we should start making sure we bump the chart version on any changes,  even a change to README etc should get a bump since the readme is shipped with a chart.

jessehu:
Thanks @paulczar. However I don't agree on bumping version on every commit. Bumping version usually happens when making a new release. Did you see other helm charts bump version on every commit ?

reasonerjt:
Low risk, I'll merge it.

PR #4550, opened 2018-03-30
Switch job service from old version to new version

steven-zou:
Enable new job service framework


steven-zou:
Close it, will raise new PR after rebase

PR #4561, opened 2018-04-02
Enable output parameter in harbor migrator

wy65701436:


tools/migration/docker-entrypoint.sh
```diff
@@ -3,4 +3,4 @@ set -e
 
 python ./migrator.py "$@"
 
-exec "$@"
\ No newline at end of file
+exit $?
```
reasonerjt: This is necessary.

tools/migration/migrator.py
```diff
@@ -37,10 +37,12 @@ def validate(self):
 
 class CfgMigrator():
 
-    def __init__(self, target):
+    def __init__(self, target, out_put):
```
reasonerjt: rename to `output`

PR #4571, opened 2018-04-03
Add initia configuration for  CLAIR_DB.

rootdeep:
When the harbor is deloyed in k8s cluster, the adminserver is not started  due to the lack of clair_db infomation in adminserver.cm.yaml. 
Please refer  to issue 4543 for related error description.

### DIFF
--- a/make/kubernetes/templates/adminserver.cm.yaml
+++ b/make/kubernetes/templates/adminserver.cm.yaml
@@ -20,6 +20,11 @@ data:
   MYSQL_USR: root
   MYSQL_PWD: "{{db_password}}"
   MYSQL_DATABASE: registry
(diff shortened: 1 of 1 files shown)


reasonerjt:
@rootdeep  thanks for this PR however the manual deployment kind for k8s will be deprecated and we'll focus on helm chart for deploying Harbor onto k8s.

rootdeep:
@reasonerjt   The problem exists.

reasonerjt:
Per discussion the doc will be supporting 1.2.x version of Harbor only, so no need further change to support later version of images, we'll only recommend deploying the Chart if user wants to use later versions of Harbor on top of K8S.
Closing this PR.

PR #4587, opened 2018-04-04
Fix issue that insecureRegistry mode does not work

jessehu:
Close issue https://github.com/vmware/harbor/issues/4499


reasonerjt:
Per discussion, let's hold this change until we clarify the expected usage of this parm.

PR #4595, opened 2018-04-08
Apply auth checking to all the incoming requests

steven-zou:
Get secret from 'Authorization' header of the request

### DIFF
--- a/src/jobservice/api/authenticator.go
+++ b/src/jobservice/api/authenticator.go
@@ -1,0 +1,62 @@
+// Copyright 2018 The Harbor Authors. All rights reserved.
+
+package api
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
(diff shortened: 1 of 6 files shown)


src/jobservice/api/authenticator.go
```diff
@@ -0,0 +1,62 @@
...
+//SecretAuthenticator implements interface 'Authenticator' based on simple secret.
+type SecretAuthenticator struct{}
+
+//DoAuth implements same method in interface 'Authenticator'.
+func (sa *SecretAuthenticator) DoAuth(req *http.Request) error {
+	if req == nil {
+		return errors.New("nil request")
+	}
+
+	h := strings.TrimSpace(req.Header.Get(authHeader))
```
reasonerjt: should probably re-use common/secret/

Issue #4496, opened 2018-03-27
k8s updates break Harbor Helm deployments


MnrGreg:
The k8s security update 1.7.14, 1.8.9, and 1.9.4 make configMaps readonly by default: [CHANGELOG-1.7.md](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.7.md#other-notable-changes-1)

> Changes secret, configMap, downwardAPI and projected volumes to mount read-only, instead of allowing applications to write data and then reverting it automatically. Until version 1.11, setting the feature gate ReadOnlyAPIDataVolumes=false will preserve the old behavior. (#58720, @joelsmith)

This breaks the Registry and Clair deployments which try to chown the configMap mounted config files on startup, resulting in a crash loop. Logs from Clair:
`chown: changing ownership of '/config/config.yaml': Read-only file system`

Clair Entrypoint = docker-entrypoint.sh:
```
#!/bin/bash
set -e
chown -R 10000:10000 /config
sudo -E -H -u \#10000 sh -c "/dumb-init -- /clair2.0.1/clair -config /config/config.yaml"
set +e
```

Kubernetes version 1.9.5
Helm version 2.8.1

jessehu:
Thanks @MnrGreg for reporting.
/cc @reasonerjt to take a log on this chown issue.

jessehu:
@MnrGreg We are discussing how to solve this problem. Please stay tuned.

MnrGreg:
Thanks @jessehu. I'm using the --feature-flags in the interim.

reasonerjt:
I think the design change makes sense that as for all `config volume` we don't have to run `chown`.

I'll fix the Dockerfile

reasonerjt:
Unfortunately, we'll need more time to reproduce this issue on k8s and provide a fix.

It's possible this fix will be pushed out of 1.5.0 to avoid introducing regression when integrating with other products.

jessehu:
@messagell You can run `kube-controller-manager --feature-gates ReadOnlyAPIDataVolumes=false` as a workaround, then deploy harbor chart again.

philosifer:
Having just hit this issue trying to move to a new K8S 1.10.3 cluster I though i'd note that this issue also affect 1.10 in the same way and that from 1.11 onwards it cannot be disabled with the feature gate

ywk253100:
@philosifer Are you using the chart in master branch? The chart in master branch should have fixed this issue.

[...]

PR #4604, opened 2018-04-09
Do not run `chown` to config files

reasonerjt:
As described in #4496, a recent security enhancement in kubernetes mount
configMap volume as readonly.  This commit make necessary update to the
entry points scripts and path of the config files to follow this
convention, such that the `chown` will not be called against them.

### DIFF
--- a/make/docker-compose.clair.tpl
+++ b/make/docker-compose.clair.tpl
@@ -41,7 +41,7 @@ services:
     depends_on:
       - postgres
     volumes:
-      - ./common/config/clair:/config:z
+      - ./common/config/clair/config.yaml:/etc/clair/config.yaml:z
     logging:
       driver: "syslog"
       options:  
--- a/make/photon/notary/server-start.sh
+++ b/make/photon/notary/server-start.sh
@@ -1,3 +1,3 @@
 #!/bin/sh
-chown 10000:10000 -R /config
-sudo -E -u \#10000 sh -c "/usr/bin/env /migrations/migrate.sh && /bin/notary-server -config=/config/server-config.json -logf=logfmt"
+chown 10000:10000 -R /etc/notary/private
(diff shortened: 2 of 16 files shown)


make/docker-compose.clair.tpl
```diff
@@ -41,7 +41,7 @@ services:
     depends_on:
       - postgres
     volumes:
-      - ./common/config/clair:/config:z
+      - ./common/config/clair/config.yaml:/etc/clair/config.yaml:z
```
yixingjia: Just want to make sure there is no other files under ./common/config/clair  since we only map the config.yaml

make/photon/notary/server-start.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
-chown 10000:10000 -R /config
-sudo -E -u \#10000 sh -c "/usr/bin/env /migrations/migrate.sh && /bin/notary-server -config=/config/server-config.json -logf=logfmt"
+chown 10000:10000 -R /etc/notary/private
```
yixingjia: Will we save some thing to /etc/notary/private ?   Or this is the default directory for notary?

make/photon/notary/signer-start.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
-chown 10000:10000 -R /config
-sudo -E -u \#10000 sh -c "/usr/bin/env && /migrations/migrate.sh && /bin/notary-signer -config=/config/signer-config.json -logf=logfmt"
+chown 10000:10000 -R /etc/notary/private
```
yixingjia: Same question as above.

reasonerjt:
Per discussion with @jessehu seems the `private` sector is unnecessary, meanwhile there will still be issues as K8S may generate certs/keys as secret and mount into container. 
We may consider avoid running `chown` in containers, @jessehu let's reproduce this issue on latest k8s env and find a better solution to make sure it fixes #4496 , I'll close this PR.

PR #4610, opened 2018-04-10
Add bread crumb under project

pengpengshui:


lweitzman:
not exactly sure what this is about but the breadcrumb in projects is wrong. it should be
"< Projects < Repositories < <image repo name>" 
currently it uses "Registries" instead of "Repositories"
also, if a UX issue, when code is pushed, put issue into verify pipeline and add "pending-ux-review" label


screen shot 2018-04-10 at 10 26 24 am

PR #4624, opened 2018-04-11
Add dbname paramter in dburl

yixingjia:
add dbname to dburl

### DIFF
--- a/make/common/templates/clair/config.yaml
+++ b/make/common/templates/clair/config.yaml
@@ -2,7 +2,7 @@ clair:
   database:
     type: pgsql
     options:
-      source: postgresql://$username:$password@$host:$port?sslmode=disable
+      source: postgresql://$username:$password@$host:$port/$dbname?sslmode=disable
 
       # Number of elements kept in the cache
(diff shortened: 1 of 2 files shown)


make/prepare
```diff
@@ -565,8 +565,10 @@ if args.clair_mode:
             password = clair_db_password,
             username = clair_db_username,
             host = clair_db_host,
-            port = clair_db_port)
-    # config http proxy for Clair
+            port = clair_db_port,
+            dbname = clair_db)
+
+# config http proxy for Clair
```
jessehu: please indent this comment line

make/common/templates/clair/config.yaml
```diff
@@ -2,7 +2,7 @@ clair:
   database:
     type: pgsql
     options:
-      source: postgresql://$username:$password@$host:$port?sslmode=disable
+      source: postgresql://$username:$password@$host:$port/$dbname?sslmode=disable
```
jessehu: Does this change impact on the upgrade script?
and is the default dbname in harbor.cfg is set in another patch?

yixingjia: The default name is already there before this patch, We just not use it.

PR #4637, opened 2018-04-12
Bug fix

ywk253100:
When deleting tags belonged to the same repository concurrently, the repository may be deleted by one API call and the other calls get 404. This commit logs a warning message and doesn't return 404.

### DIFF
--- a/src/ui/api/repository.go
+++ b/src/ui/api/repository.go
@@ -324,7 +324,7 @@ func (ra *RepositoryAPI) Delete() {
 			return
 		}
 		if repository == nil {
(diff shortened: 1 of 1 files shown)


src/ui/api/repository.go
```diff
@@ -324,7 +324,7 @@ func (ra *RepositoryAPI) Delete() {
 			return
 		}
 		if repository == nil {
-			ra.HandleNotFound(fmt.Sprintf("repository %s not found", repoName))
+			log.Warningf("repository %s not found", repoName)
```
reasonerjt: Please add more details, such as `after deleting tag, the repo is not found` so we differentiate the scenario between tag deletion and repo deletion.

PR #4638, opened 2018-04-12
Use redis URL address to replace host:port when connecting to redis server

steven-zou:
replace tcp host:port with
'redis://arbitrary_usrname:password@ipaddress:port/database_index'


make/harbor.cfg
```diff
@@ -141,7 +141,8 @@ db_user = root
 ##### End of Harbor DB configuration#######
 
 #The redis server address. Only needed in HA installation.
```
ywk253100: The line should be removed as we need Redis in standalone Harbor either.

PR #4689, opened 2018-04-17
Migrate harbor DB to postgresql

wy65701436:
1, move harbor db from maria to postgresql
2, move notary db from maria to postgresql
3, merge harbor_db, clair_db and notary_db into one pgsql instance.


make/photon/db/postgresql/registry.sql
```diff
@@ -0,0 +1,348 @@
...
+CREATE FUNCTION update_update_time_at_column() RETURNS trigger
+    LANGUAGE plpgsql
+    AS $$
+  BEGIN
+    NEW.update_time = NOW();
+    RETURN NEW;
+  END;
+$$;
+
+CREATE TRIGGER project_member_update_time_at_modtime BEFORE UPDATE ON project_member FOR EACH ROW EXECUTE PROCEDURE update_update_time_at_column();
```
stonezdj: Please do not create trigger in database, because it may cause performance issue and not easy to migrate in future

make/common/templates/notary/server-config.postgres.json
```diff
@@ -0,0 +1,28 @@
...
+		"port": "7899",
+		"tls_ca_file": "./notary-signer-ca.crt",
+		"key_algorithm": "ecdsa"
+	},
+	"logging": {
+		"level": "debug"
+	},
+	"storage": {
+		"backend": "postgres",
+		"db_url": "postgres://server:password@postgresql:5432/notaryserver?sslmode=disable"
```
yixingjia: Just want to make sure this "db_url" value change to real value in deployment

make/photon/db/postgresql/initial-notaryserver.sql
```diff
@@ -0,0 +1,4 @@
+CREATE DATABASE notaryserver;
+CREATE USER server;
+alter user server with encrypted password 'password';
```
yixingjia: So we are hard code notary password to "password" . It it possible we share the same password with harbor db?  which is also random generated.

make/photon/db/postgresql/registry.sql
wy65701436: Is there any other way to update the time column? I followed the document of postgresql.

make/common/templates/notary/server-config.postgres.json
wy65701436: we do not provide any configuration for notary, this will be used for internal only.  What do you mean by real value?

make/photon/db/postgresql/initial-notaryserver.sql
wy65701436: as notary is using no password way to connect DB, add the password here is because that pgsql raises one exception on "no password to connect db".  I guess that pqsql could not accept one db  both hold connect with and without password.

But, I agree that we could provide better way to generate password random. it could be fixed in another PR.

make/common/templates/notary/server-config.postgres.json
reasonerjt: I think he means in the case when DB is external.

PR #4701, opened 2018-04-18
Provide guideline about how to customize the look&feel of Harbor UI

steven-zou:
support customizing product name
support customizing logo
support customizing login background image
support customizing about text with different languages


steven-zou:
wrong, close

PR #4707, opened 2018-04-18
[WIP] Do not call `chown` to config files

reasonerjt:
This commit fixes a recently discovered issue on Kubernetes #4496
It make necessary to avoid calling `chown` to config files during the
bootstrap of the containers.

### DIFF
--- a/make/photon/jobservice/start.sh
+++ b/make/photon/jobservice/start.sh
@@ -1,7 +1,4 @@
 #!/bin/sh
-if [ -d /etc/jobservice/ ]; then
-    chown -R 10000:10000 /etc/jobservice/ 
-fi
 if [ -d /var/log/jobs ]; then
     chown -R 10000:10000 /var/log/jobs/
 fi
--- a/make/prepare
+++ b/make/prepare
@@ -138,16 +138,16 @@ def _get_secret(folder, filename, length=16):
             print("loaded secret from file: %s" % key_file)
         return key
     if not os.path.isdir(folder):
-        os.makedirs(folder, mode=0o600)
+        os.makedirs(folder)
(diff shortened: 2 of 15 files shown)


make/prepare
```diff
@@ -534,6 +537,10 @@ if args.notary_mode:
         shutil.copy2(os.path.join(notary_temp_dir, "notary-signer.key"), notary_config_dir)
         shutil.copy2(os.path.join(notary_temp_dir, "notary-signer-ca.crt"), notary_config_dir)
     shutil.copy2(os.path.join(registry_config_dir, "root.crt"), notary_config_dir)
+    mark_file(os.path.join(notary_config_dir, "notary-signer.crt"))
```
jessehu: maybe consider setting the owner and group on the dir, instead of each files? Not necessary to make it in this patch

src/common/utils/notary/helper.go
```diff
@@ -36,7 +36,7 @@ import (
 )
 
 var (
-	notaryCachePath = "/etc/ui/notary-cache"
+	notaryCachePath = "/tmp/notary-cache"
```
jessehu: Do we need to pre-create this dir ? Or the notary code will create it ?

make/photon/jobservice/start.sh
```diff
@@ -1,7 +1,4 @@
 #!/bin/sh
-if [ -d /etc/jobservice/ ]; then
-    chown -R 10000:10000 /etc/jobservice/ 
-fi
 if [ -d /var/log/jobs ]; then
     chown -R 10000:10000 /var/log/jobs/
```
jessehu: shall we put this 'chown' into prepare? and set the dir mode for "/data/job_logs" to 700?
/data/job_logs:/var/log/jobs

Issue #4688, opened 2018-04-17
User who does not have permission can push image to any project.


reasonerjt:
Due to a change in response body, there's a condition `0==0` that will allow user to push image to any project.
A ticket for Admiral has been opened.
But we should update the code to check project more strictly.

ywk253100:
Fixed.

PR #4716, opened 2018-04-18
Fix bug #4688

ywk253100:
Fix bug: the user can push images although he have no permisson by checking empty value before assign permissions.


reasonerjt:
Please rebase

PR #4722, opened 2018-04-19
 Update security context for assign role to project group member

stonezdj:
The project list will contain all public projects, user is a member of this project, or user is in the group which is a member of this projects.
Change the behaviour of user roles, if the user is not a member of this project, then return the user's groups role of current project

### DIFF
--- a/src/common/dao/project.go
+++ b/src/common/dao/project.go
@@ -135,42 +135,85 @@ func GetTotalOfProjects(query *models.ProjectQueryParam) (int64, error) {
 
 // GetProjects returns a project list according to the query conditions
 func GetProjects(query *models.ProjectQueryParam) ([]*models.Project, error) {
-	sql, params := projectQueryConditions(query)
-	if query == nil {
-		sql += ` order by p.name`
-	}
+	sqlStr, queryParam := projectQueryConditions(query)
+	sqlStr = `select distinct p.project_id, p.name, p.owner_id, 
+		p.creation_time, p.update_time ` + sqlStr + ` order by p.name`
+	sqlStr, queryParam = CreatePagination(query, sqlStr, queryParam)
 
+	log.Debugf("sql:=%+v, param= %+v", sqlStr, queryParam)
+	var projects []*models.Project
+	_, err := GetOrmer().Raw(sqlStr, queryParam).QueryRows(&projects)
+
+	return projects, err
+
+}
+
+// GetGroupProjects - Get user's all projects, including user is the user member of this project
+// and the user is in the group which is a group member of this project.
+func GetGroupProjects(groupDNCondition string, query *models.ProjectQueryParam) ([]*models.Project, error) {
+	sql, params := projectQueryConditions(query)
 	sql = `select distinct p.project_id, p.name, p.owner_id, 
 				p.creation_time, p.update_time ` + sql
+	if len(groupDNCondition) > 0 {
+		sql = fmt.Sprintf(
+			`%s union select distinct p.project_id, p.name, p.owner_id, p.creation_time, p.update_time  
+		     from project p 
+		     left join project_member pm on p.project_id = pm.project_id
+		     left join user_group ug on ug.id = pm.entity_id and pm.entity_type = 'g' and ug.group_type = 1
+			 where ug.ldap_group_dn in ( %s ) order by name`,
+			sql, groupDNCondition)
+	}
+	sqlStr, queryParams := CreatePagination(query, sql, params)
+	log.Debugf("query sql:%v", sql)
 	var projects []*models.Project
-	_, err := GetOrmer().Raw(sql, params).QueryRows(&projects)
+	_, err := GetOrmer().Raw(sqlStr, queryParams).QueryRows(&projects)
 	return projects, err
+}
+
+// GetTotalGroupProjects - Get the total count of projects, including  user is the member of this project and the
+// user is in the group, which is the group member of this project.
+func GetTotalGroupProjects(groupDNCondition string, query *models.ProjectQueryParam) (int, error) {
+	var sql string
+	sqlCondition, params := projectQueryConditions(query)
+	if len(groupDNCondition) == 0 {
+		sql = `select count(1) ` + sqlCondition
+	} else {
+		sql = fmt.Sprintf(
+			`select count(1) 
+			   from ( select  p.project_id %s  union select  p.project_id  
+			   from project p 
+			   left join project_member pm on p.project_id = pm.project_id
+			   left join user_group ug on ug.id = pm.entity_id and pm.entity_type = 'g' and ug.group_type = 1
+			   where ug.ldap_group_dn in ( %s )) t`,
+			sqlCondition, groupDNCondition)
+	}
+	log.Debugf("query sql:%v", sql)
+	var count int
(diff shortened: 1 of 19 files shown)


src/common/security/context.go
```diff
@@ -34,6 +34,8 @@ type Context interface {
 	HasWritePerm(projectIDOrName interface{}) bool
 	// HasAllPerm returns whether the user has all permissions to the project
 	HasAllPerm(projectIDOrName interface{}) bool
+	//Get current user's all project
 	GetMyProjects() ([]*models.Project, error)
+	//Get user's role in provided project
 	GetProjectRoles(projectIDOrName interface{}) []int
```
reasonerjt: So after the group roles are introduced this func will still return the `user role` in contrast to `combined role`?  Could you remind me why we made such decision?

src/common/security/local/context.go
```diff
@@ -89,20 +90,24 @@ func (s *SecurityContext) HasReadPerm(projectIDOrName interface{}) bool {
 
 	roles := s.GetProjectRoles(projectIDOrName)
 
-	return len(roles) > 0
+	if len(roles) > 0 {
+		return true
+	}
+
+	//Get Group Roles
+	groupRoles := s.GetRolesByGroup(projectIDOrName)
```
ywk253100: The result of GetProjectRoles() has already contained the group roles. No need to call GetRolesByGroup() again.
The same for HasWritePerm() and HasAllPerm()

src/common/dao/project.go
```diff
@@ -241,12 +284,31 @@ func DeleteProject(id int64) error {
...
+	o := GetOrmer()
+	sql := fmt.Sprintf(
+		`select pm.role from project_member pm 
+		left join user_group ug on pm.entity_type = 'g' and pm.entity_id = ug.id 
+		where ug.ldap_group_dn in ( %s ) and pm.project_id = ? `,
+		groupDNCondition)
+	log.Debugf("sql:%v", sql)
+	if _, err := o.Raw(sql, projectID).QueryRows(&roles); err != nil {
+		log.Warningf("Error in GetRolesByLDAPGroup, error: %v", err)
+		return nil
```
ywk253100: Why don't we return the err but just log a warning message?

src/common/security/local/context.go
```diff
@@ -191,8 +203,35 @@ func (s *SecurityContext) GetProjectRoles(projectIDOrName interface{}) []int {
...
+	user := s.user
+	project, err := s.pm.Get(projectIDOrName)
+	//No user, group or project info
+	if err != nil || project == nil || user == nil || len(user.GroupList) == 0 {
+		return roles
+	}
+	//Get role by LDAP group
+	groupDNConditions := group.GetGroupDNQueryCondition(user.GroupList)
+	roles = dao.GetRolesByLDAPGroup(project.ProjectID, groupDNConditions)
+	if err != nil {
```
ywk253100: The error checking is not necessary as dao.GetRolesByLDAPGroup() doesn't return any error.

src/common/security/context.go
ywk253100: No. It should return all roles of the user.

src/common/security/local/context.go
```diff
@@ -191,8 +203,35 @@ func (s *SecurityContext) GetProjectRoles(projectIDOrName interface{}) []int {
...
+		return nil
+	}
 	return roles
 }
+
+// GetMyProjects ...
+func (s *SecurityContext) GetMyProjects() ([]*models.Project, error) {
```
ywk253100: Seems the function doesn't return the projects which the user is a group member of.

src/ui/promgr/pmsdriver/local/local.go
```diff
@@ -131,14 +131,32 @@ func (d *driver) Update(projectIDOrName interface{},
...
-	}
-
-	projects, err := dao.GetProjects(query)
-	if err != nil {
-		return nil, err
+	var total int64
+	var projects []*models.Project
+	var err error
+
+	if query != nil && query.User != nil && len(query.User.GroupList) > 0 {
```
ywk253100: How can we handle the request that wants all project a user is a member of(no matter user member or group member)?

src/common/security/local/context.go
stonezdj: Fixed it

src/common/dao/project.go
stonezdj: Return error now

src/common/security/local/context.go
stonezdj: Return error now

src/common/security/local/context.go
stonezdj: Updated

src/ui/api/project.go
```diff
@@ -348,7 +348,13 @@ func (p *ProjectAPI) List() {
 				}
 				projects = append(projects, pros...)
 
+				user, err := dao.GetUser(models.User{Username: p.SecurityCtx.GetUsername()})
+				if err != nil {
+					p.HandleInternalServerError(fmt.Sprintf("failed to get user: %v", err))
+					return
+				}
 				mps, err := p.ProjectMgr.List(&models.ProjectQueryParam{
+					User: user,
```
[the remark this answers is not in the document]
stonezdj: Updated, call projectManager.List in SecurityContext.GetMyProjects()

src/common/security/local/context.go
reasonerjt: @stonezdj So do we still need  `GetRolesByGroup`?

stonezdj: This method is used in by GetProjectRoles(), still need it

src/common/security/context.go
```diff
@@ -24,6 +24,8 @@ type Context interface {
 	IsAuthenticated() bool
 	// GetUsername returns the username of user related to the context
 	GetUsername() string
+	//GetUserGroupList used to retrieve user group information
+	GetUserGroupList() []*models.UserGroup
```
ywk253100: Why we need this method?

src/ui/api/project.go
```diff
@@ -347,10 +347,10 @@ func (p *ProjectAPI) List() {
 					return
 				}
 				projects = append(projects, pros...)
-
 				mps, err := p.ProjectMgr.List(&models.ProjectQueryParam{
 					Member: &models.MemberQuery{
-						Name: p.SecurityCtx.GetUsername(),
+						Name:      p.SecurityCtx.GetUsername(),
+						GroupList: p.SecurityCtx.GetUserGroupList(),
```
ywk253100: If we call SecurityCtx.GetMyProjects() directly, we can hide the getting user group list logic inside GetMyProjects and do not need the GetUserGroupList anymore?

src/common/security/local/context.go
ywk253100: No need to call GetRolesByGroup() here.

src/ui/api/project.go
stonezdj: change to SecurityContext.GetMyProjects()

src/common/security/local/context.go
stonezdj: Removed it, issue maybe caused by code merge

src/common/security/context.go
stonezdj: Remove it

PR #4731, opened 2018-04-19
Do not call `chown` to config files

reasonerjt:
This commit fixes a recently discovered issue on Kubernetes #4496
It make necessary to avoid calling `chown` to config files during the
bootstrap of the containers.

### DIFF
--- a/make/photon/registry/entrypoint.sh
+++ b/make/photon/registry/entrypoint.sh
@@ -2,12 +2,12 @@
 
 set -e
 
-if [ -d /etc/registry ]; then
-    chown 10000:10000 -R /etc/registry
-fi
+# The directory /var/lib/registry is within the container, and used to store image in CI testing.
+# So for now we need to chown to it to avoid failure in CI.
 if [ -d /var/lib/registry ]; then
     chown 10000:10000 -R /var/lib/registry
 fi
+
 if [ -d /storage ]; then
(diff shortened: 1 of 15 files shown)


make/prepare
```diff
@@ -173,6 +182,10 @@ def delfile(src):
             itemsrc=os.path.join(src,item)
             delfile(itemsrc)
 
+if not os.path.exists(JOB_LOG_DIR):
+    os.makedirs(JOB_LOG_DIR)
+mark_file(JOB_LOG_DIR, mode=0o655)
```
jessehu: Should the mode be 755?  655 is seldomly seen.

make/photon/registry/entrypoint.sh
```diff
@@ -2,12 +2,11 @@
 
 set -e
 
-if [ -d /etc/registry ]; then
-    chown 10000:10000 -R /etc/registry
-fi
 if [ -d /var/lib/registry ]; then
+    # For CI testng.
```
jessehu: please add detail description for this chown, and 'testng' is a typo.

make/prepare
reasonerjt: You are right...

make/photon/registry/entrypoint.sh
reasonerjt: yep, will do.

cristifalcas:
Can we have new images with those changes?

PR #4736, opened 2018-04-20
Return the logger.Interface instead of the pointer of logger to avoid nil checking issue

steven-zou:
An interface value is nil only if the inner value and type are both unset, (nil, nil). In particular, a nil interface will always hold a nil type. If we store a nil pointer of type *int inside an interface value, the inner type will be *int regardless of the value of the pointer: (*int, nil). Such an interface value will therefore be non-nil even when the pointer inside is nil.

### DIFF
--- a/src/jobservice/job/impl/logger/job_logger.go
+++ b/src/jobservice/job/impl/logger/job_logger.go
@@ -4,6 +4,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/steven-zou/harbor/src/jobservice/logger"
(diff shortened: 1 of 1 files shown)


src/jobservice/job/impl/logger/job_logger.go
```diff
@@ -4,6 +4,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/steven-zou/harbor/src/jobservice/logger"
```
reasonerjt: package name is incorrect.

PR #4769, opened 2018-04-23
[Chart] Fix registry storage config issue

draeron:
- when persistence is disabled
- when using s3 storage

### DIFF
--- a/contrib/helm/harbor/templates/registry/registry-cm.yaml
+++ b/contrib/helm/harbor/templates/registry/registry-cm.yaml
@@ -14,13 +14,12 @@ data:
     storage:
       {{- if .Values.registry.objectStorage }}
 {{ toYaml .Values.registry.objectStorage | indent 6 }}
-      {{- end }}
-      cache:
-        layerinfo: inmemory
-      {{- if not .Values.registry.objectStorage }}
+      {{- else }}
       filesystem:
         rootdirectory: /var/lib/registry
       {{- end }}
+      cache:
+        layerinfo: inmemory
       maintenance:
         uploadpurging:
           enabled: false
--- a/contrib/helm/harbor/templates/registry/registry-cm.yaml
+++ b/contrib/helm/harbor/templates/registry/registry-cm.yaml
@@ -8,7 +8,7 @@ data:
   config.yml: |+
     version: 0.1
     log:
-      level: info
+      level: {{ .Values.registry.logLevel }}
       fields:
         service: registry
     storage:
@@ -17,8 +17,10 @@ data:
       {{- end }}
       cache:
         layerinfo: inmemory
+      {{- if not .Values.registry.objectStorage }}
       filesystem:
         rootdirectory: /var/lib/registry
+      {{- end }}
       maintenance:
         uploadpurging:
           enabled: false
--- a/contrib/helm/harbor/templates/registry/registry-ss.yaml
+++ b/contrib/helm/harbor/templates/registry/registry-ss.yaml
@@ -35,8 +35,10 @@ spec:
         - containerPort: 5000
         - containerPort: 5001
         volumeMounts:
+{{- if not .Values.registry.objectStorage }}
         - name: registry-data
           mountPath: /var/lib/registry
+{{- end }}
         - name: registry-root-certificate
           mountPath: /etc/registry/root.crt
           subPath: root.crt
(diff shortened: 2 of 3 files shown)


contrib/helm/harbor/templates/registry/registry-cm.yaml
```diff
@@ -8,7 +8,7 @@ data:
   config.yml: |+
     version: 0.1
     log:
-      level: info
+      level: {{ .Values.registry.logLevel }}
```
jessehu: There is no default value for .Values.registry.logLevel in values.yaml. Need to add it.

contrib/helm/harbor/templates/registry/registry-cm.yaml
```diff
@@ -17,8 +17,10 @@ data:
       {{- end }}
       cache:
         layerinfo: inmemory
+      {{- if not .Values.registry.objectStorage }}
       filesystem:
         rootdirectory: /var/lib/registry
+      {{- end }}
```
jessehu: we can move this up in the 1st {{- if .Values.registry.objectStorage }} by adding a 'else' there.

contrib/helm/harbor/templates/registry/registry-ss.yaml
```diff
@@ -44,11 +44,9 @@ spec:
           mountPath: /etc/registry/config.yml
           subPath: config.yml
       volumes:
-{{- if not .Values.registry.objectStorage }}
-{{- if not .Values.persistence.enabled }}
+{{- if or (not .Values.persistence.enabled) .Values.registry.objectStorage }}
```
jessehu: this could be simply `{{- if not .Values.registry.objectStorage }}`. Since `.Values.persistence.enabled` is not used other places related to registry storage. Sorry for the original confusing code.

jessehu: Also need to add `{{- if not .Values.registry.objectStorage }}` around line 38-39

jessehu: Thanks very much for this PR.

jessehu: BTW: suggest change the commit tile to "[Chart] Fix registry storage config issue"

contrib/helm/harbor/templates/registry/registry-cm.yaml
draeron: added in db67636fdd7debb1a88cf7b707f553165ef1d702

contrib/helm/harbor/templates/registry/registry-ss.yaml
draeron: i've renamed the PR

=)

contrib/helm/harbor/templates/registry/registry-ss.yaml
```diff
@@ -35,16 +35,18 @@ spec:
...
+{{- end }}
         - name: registry-root-certificate
           mountPath: /etc/registry/root.crt
           subPath: root.crt
         - name: registry-config
           mountPath: /etc/registry/config.yml
           subPath: config.yml
       volumes:
-{{- if or (not .Values.persistence.enabled) .Values.registry.objectStorage }}
+{{- if .Values.registry.objectStorage }}
```
jessehu: Should it be `if not .Values.registry.objectStorage` ?

contrib/helm/harbor/templates/registry/registry-ss.yaml
```diff
@@ -35,16 +35,18 @@ spec:
         - containerPort: 5000
         - containerPort: 5001
         volumeMounts:
+{{- if .Values.registry.objectStorage }}
```
jessehu: Should it be `if not .Values.registry.objectStorage` ? Because when .Values.registry.objectStorage is specified, no need to create the data dir.

draeron: you're right! the condition should be inverted and the _emptyDir_ should be removed altogether since if using an object storage, there won't be `registry-data` mount.

So the volume mount would look like this : 
```yaml
        volumeMounts:
{{- if not .Values.registry.objectStorage }}
        - name: registry-data
          mountPath: /var/lib/registry
{{- end }}
```

and volume claim :
```yaml
  {{- if and (not .Values.registry.objectStorage) .Values.persistence.enabled }}
  volumeClaimTemplates:
  - metadata:
      name: "registry-data"
# yadda yadda
  {{- end -}}
```

Issue #4712, opened 2018-04-18
Intermittent restart of job service 


reasonerjt:
This maybe an env issue, but we need an issue to track this.

On my dev env, I noticed the jobservice container has been restarting every a few hours, relevant log:
```
Apr 17 19:41:50 172.17.0.1 jobservice[10984]: [mysql] 2018/04/18 02:41:50 packets.go:130: write tcp 172.17.0.8:54794->172.17.0.5:3306: write: broken pipe
Apr 17 19:41:50 172.17.0.1 jobservice[10984]: [mysql] 2018/04/18 02:41:50 packets.go:130: write tcp 172.17.0.8:54794->172.17.0.5:3306: write: broken pipe
Apr 17 19:43:55 172.17.0.1 jobservice[10984]: 2018-04-18T02:43:55Z [INFO] Job incoming: IMAGE_SCAN:7ad28e356d4adc1629d8b226
Apr 17 19:45:35 172.17.0.1 jobservice[10984]: 2018-04-18T02:45:35Z [INFO] Job incoming: IMAGE_SCAN:c2b7e89e55b9bdeac3dbaa1b
Apr 17 19:45:35 172.17.0.1 jobservice[10984]: 2018-04-18T02:45:35Z [INFO] Job incoming: IMAGE_SCAN:273e32af90e33e2ea2a98cbf
Apr 17 19:45:35 172.17.0.1 jobservice[10984]: 2018-04-18T02:45:35Z [INFO] Job incoming: IMAGE_SCAN:de541c0386aa6dbb25a68574
Apr 17 19:45:35 172.17.0.1 jobservice[10984]: 2018-04-18T02:45:35Z [INFO] Job incoming: IMAGE_SCAN:7c8381388ae3749146044d62
Apr 17 19:45:35 172.17.0.1 jobservice[10984]: 2018-04-18T02:45:35Z [INFO] Job incoming: IMAGE_SCAN:7b99ea10888a6ee15db155f7
Apr 17 19:45:35 172.17.0.1 jobservice[10984]: 2018-04-18T02:45:35Z [INFO] Job incoming: IMAGE_SCAN:941e14163b2ef538f8e9daff
Apr 17 19:45:35 172.17.0.1 jobservice[10984]: 2018-04-18T02:45:35Z [INFO] Job incoming: IMAGE_SCAN:b02d9b50febd897bb59ee20f
[...]

PR #4773, opened 2018-04-24
Fix the issue of intermittent restarting of job service

steven-zou:
github issue: #4712
ping redis server before pool starting
Let worker pool to restart the message server if message server exits with error (controlled by max retries)

### DIFF
--- a/src/jobservice/pool/message_server.go
+++ b/src/jobservice/pool/message_server.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"reflect"
 	"time"
 
@@ -16,6 +17,10 @@ import (
 	"github.com/garyburd/redigo/redis"
 	"github.com/vmware/harbor/src/jobservice/models"
 	"github.com/vmware/harbor/src/jobservice/utils"
+)
+
+const (
+	msgServerRetryTimes = 5
 )
 
 //MessageServer implements the sub/pub mechanism via redis to do async message exchanging.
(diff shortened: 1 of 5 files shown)


src/jobservice/pool/message_server.go
```diff
@@ -18,40 +19,75 @@ import (
...
 type MessageServer struct {
-	context   context.Context
-	redisPool *redis.Pool
-	namespace string
-	callbacks map[string]reflect.Value //no need to sync
+	context      context.Context
+	redisPool    *redis.Pool
+	namespace    string
+	callbacks    map[string]reflect.Value //no need to sync
+	retryFlag    atomic.Value
```
reasonerjt: Need more context to understand why this has to be atomic value.

src/jobservice/pool/message_server.go
```diff
@@ -155,6 +200,22 @@ func (ms *MessageServer) Subscribe(event string, callback interface{}) error {
 	return nil
 }
 
+func (ms *MessageServer) shouldRetry() bool {
+	flag := ms.retryFlag.Load().(bool)
+	count := ms.retryCounter.Load().(int)
+	yesOrNo := flag && count < retryTimes
+	if yesOrNo {
+		count++
```
reasonerjt: If the func `Start()` is not thread-safe, `shouldRetry()` is not, so then `count++` is problematic.

src/jobservice/pool/message_server.go
steven-zou: Just to avoid -race problems. It assumes the start might be called by different goroutines. Actually it's only called once

src/jobservice/pool/message_server.go
steven-zou: Right. Should use safe increase method

src/jobservice/pool/redis_pool.go
```diff
@@ -130,8 +135,18 @@ func (gcwp *GoCraftWorkPool) Start() {
 			return
 		}
 
+		startTimes := 0
+	START_MSG_SERVER:
```
reasonerjt: Can we use a for loop？

src/jobservice/pool/redis_pool.go
```diff
@@ -593,6 +610,37 @@ func (rpc *RedisPoolContext) logJob(job *work.Job, next work.NextMiddlewareFunc)
...
+	}
+
+	count := 1
+	timer := time.NewTimer(time.Duration(count) * time.Second)
+	defer timer.Stop()
+
+	var err error
+	for {
+		select {
+		case <-timer.C:
```
reasonerjt: this may be trivial, but there's only one channel to be listened, by do we need to use `select`?
And why not use time.Sleep?

src/jobservice/pool/redis_pool.go
steven-zou: Just preference, label and goto is more simple

Issue #4617, opened 2018-04-10
Adding labels to an image needs a couple updates


lweitzman:
When adding a label to an image, 
- the menu should put all the selected labels grouped at the top. (while keeping the system and project scoped labels together)
- when filtering the existing labels, the checkmarks for current labels are lost
- removing the filter the checkmarks have disappeared (re-entering the menu they still are shown)
- filtering in this menu uses the "contains" relationship which is good, but the search on the datagrid uses "starts with" which should change to be consistent with this menu

![Screen Shot 2018-04-10 at 10.49.58 AM.png](https://images.zenhubusercontent.com/56afa7c46caf56c9d38c1f12/8e9ea39b-f1aa-4d40-8645-f2262e11fd1f)

PR #4784, opened 2018-04-26
Add sort fun when filter labels in tag page  #4617

pengpengshui:
Add sort function when filter labels in tag page

Modify tag detail page layout and add scan button

Add page freash code to verify name exist when create new label


reasonerjt:
Could you split the PR such that each commit focus on one thing?

PR #4815, opened 2018-04-27
Fix URL hardcode on replication service

ninjadq:
Url hardcoded in replication service. This commit is to fix it.


src/ui_ng/lib/src/service/replication.service.ts
```diff
@@ -149,11 +149,9 @@ export class ReplicationDefaultService extends ReplicationService {
...
     ) {
         super();
-        this._ruleBaseUrl = config.replicationRuleEndpoint ?
-            config.replicationRuleEndpoint : '/api/policies/replication';
-        this._jobBaseUrl = config.replicationJobEndpoint ?
-            config.replicationJobEndpoint : '/api/jobs/replication';
-        this._replicateUrl = '/api/replications';
+        this._ruleBaseUrl = config.replicationRuleEndpoint ? config.replicationRuleEndpoint : '/api/policies/replication';
+        this._jobBaseUrl = config.replicationJobEndpoint ? config.replicationJobEndpoint : '/api/jobs/replication';
+        this._replicateUrl = config.repositoryBaseEndpoint ? config.repositoryBaseEndpoint : '/api/jobs/replication'
```
steven-zou: Is it the right replication url?

Issue #4840, opened 2018-05-02
Upgrade failed in VIC if replication policies exist


ywk253100:

reasonerjt:
This is a blocker for vic appliance upgrade.

After the fix is applied, user needs to update the trigger from manual to immediate, such that when an image is pushed the new image will be replicated.

A workaround is to disable all replication policies before upgrade/migration.

ywk253100:
Fixed.

PR #4848, opened 2018-05-02
Fix bug #4840

ywk253100:
Fix the bug: upgrade failed in VIC if replication policies exist


tools/migration/db/migration_harbor/versions/1_4_0.py
```diff
@@ -52,17 +53,26 @@ def upgrade():
     # Divided policies into unabled and enabled group
     unenabled_policies = session.query(ReplicationPolicy).filter(ReplicationPolicy.enabled == 0)
```
wy65701436: disabled?

Issue #4771, opened 2018-04-24
Harbor API "automatically_scan_images_on_push" not work


jeffweilee:
Harbor create project API:
POST /project
{
  "project_name": "string",
  "public": 0,
  "enable_content_trust": true,
  "prevent_vulnerable_images_from_running": true,
  "prevent_vulnerable_images_from_running_severity": "High",
  "automatically_scan_images_on_push": true
}

Response 201 Created but the following parameter in API body did not take affect (In project/Configuration, the corresponding settings are not checked or value changed)

 "automatically_scan_images_on_push":true not work
 "prevent_vulnerable_images_from_running_severity":"High" not work
 "prevent_vulnerable_images_from_running":true not work

We really need "automatically_scan_images_on_push" from Harbor project creation API because we are creating lots of projects day by day and leverage Clair for vulnerability scan.

1) docker 17.03.2-ce, docker-compose 1.16.1
2) HTTPS enabled
3) Log I found related to create project from API is proxy.log:
Apr 24 10:27:38 172.18.0.1 proxy[17712]: 10.233.128.93 - "POST /api/projects HTTP/1.1" 201 0 "-" "Mozilla/5.0 (Windows NT 6.3; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/63.0.3239.84 Safari/537.36" 0.008 0.008 .

ywk253100:
Post the data as the following:
```
POST /project
{
  "project_name": "string",
  "metadata": {
    "public": 0,
    "enable_content_trust": true,
    "prevent_vulnerable_images_from_running": true,
    "prevent_vulnerable_images_from_running_severity": "string",
    "automatically_scan_images_on_push": true
  }
}
```
There are some changes in API and the doc isn't updated properly. We'll update it later.

jeffweilee:
Hi @ywk253100, does Harbor 1.4.0 release support the POST data format as you mentioned above?  I tried but got 400 bad request: invalid json request ...

jeffweilee:
I found the following format works on Harbor v.1.4.0

POST api/projects
{
  "project_name": "string",
  "metadata": {
    "public": "1",
    "enable_content_trust": "true",
    "prevent_vul":"true",
    "severity": "high",
    "auto_scan":"true"
  }
}

ywk253100:
No `"` needed for the `boolean` values.

PR #4862, opened 2018-05-03
Update API doc

ywk253100:
Update the parameters used in request and the data of response of projects API in swagger


ywk253100:
Fix #4771.

PR #4877, opened 2018-05-07
Add package mariadb-server-galera on image photon/mariadb

vasartori:
This commit add the binaries needed by mariadb cluster.


reasonerjt:
Sorry we are moving from MariaDB to postgresql

vasartori:
OK! No problem!

PR #4885, opened 2018-05-09
Put user info into session

stonezdj:
Fix the following issues.
1) GroupList is not found in SecurityContext user info
2) Retrieve multiple memberof information from LDAP.
3) If user is in two groups with guest, administrator role separately, display the max privilege role.

Manual Test includes

	1. Create three groups in LDAP. 
cn=harbor_guest,ou=groups,dc=example,dc=com
cn=harbor_developer,ou=groups,dc=example,dc=com
cn=harbor_admin,ou=groups,dc=example,dc=com

Create three users in ldap , and assign member to these groups 

guest_user -- harbor_guest
dev_user   -- harbor_developer
admin_user -- harbor_guest, harbor_admin

	2. Configure harbor with ldap_auth. Configure to LDAP authentication with following configuration:
{

  "ldap_group_base_dn": "ou=groups,dc=example,dc=com",
  "ldap_group_search_filter": "objectclass=groupOfNames",
  "ldap_group_search_scope": 2,
  "ldap_group_attribute_name":"cn"
}

	3. Create a project named sample, it is private project, 
	Assign role to group with rest api POST  /api/projects/<id>/member
	
	harbor_guest, guest
	harbor_developer, developer
	harbor_admin , administrator

{
  "role_id": 1,
  "member_group": {
    "ldap_group_dn": "cn=harbor_admin,ou=groups,dc=example,dc=com"
  }
}

{
  "role_id": 2,
  "member_group": {
    "ldap_group_dn": "cn=harbor_developer,ou=groups,dc=example,dc=com"
  }
}

{
  "role_id": 3,
  "member_group": {
    "ldap_group_dn": "cn=harbor_guest,ou=groups,dc=example,dc=com"
  }
}

  4. Login in harbor with guest_user

The user can pull images from sample

but can not push images 

  5. Login Ldap user dev_user, the user can see the sample project and have developer role

  6. Login ldap user admin_user
The user can push and pull images from sample

(If a user is in multiple groups, it will select the highest privilege in these groups)

 7. Login with ldap user mike
The sample project is invisible to mike.

	8. Add user member (admin_user) in project member with role guest, 

check the admin_user  can see the sample project, and has guest role.
(if a user has a role in project member, then the group role should not take effect)


src/common/dao/project.go
```diff
@@ -293,7 +293,7 @@ func GetRolesByLDAPGroup(projectID int64, groupDNCondition string) ([]int, error
 	}
 	o := GetOrmer()
 	sql := fmt.Sprintf(
-		`select pm.role from project_member pm 
+		`select min(pm.role) from project_member pm
```
reasonerjt: Will there be sql injection issue?
And why there can be multiple ldap_group_dn?

src/ui/filter/security.go
```diff
@@ -215,32 +215,34 @@ func (b *basicAuthReqCtxModifier) Modify(ctx *beegoctx.Context) bool {
...
-
-	log.Debug("got user information from session")
-	user := &models.User{
-		Username: username.(string),
-	}
-	isSysAdmin := ctx.Input.Session("isSysAdmin")
-	if isSysAdmin != nil && isSysAdmin.(bool) {
-		user.HasAdminRole = true
+	var user models.User
+	userInterface := ctx.Input.Session("user")
```
reasonerjt: How about we only store userInterface in session?

src/common/dao/project.go
```diff
@@ -302,5 +302,8 @@ func GetRolesByLDAPGroup(projectID int64, groupDNCondition string) ([]int, error
 		log.Warningf("Error in GetRolesByLDAPGroup, error: %v", err)
 		return nil, err
 	}
+	if len(roles) == 1 && roles[0] == 0 { //Avoid return 0 as role
```
reasonerjt: Add more comments to elaborate the reason please.

waqadahmad:
Pit user info into session¿

src/common/dao/project.go
stonezdj: a user can be memberof many groups

src/common/dao/project.go
stonezdj: OK

src/ui/filter/security.go
```diff
@@ -215,32 +215,28 @@ func (b *basicAuthReqCtxModifier) Modify(ctx *beegoctx.Context) bool {
...
 	log.Debug("got user information from session")
-	user := &models.User{
-		Username: username.(string),
-	}
-	isSysAdmin := ctx.Input.Session("isSysAdmin")
-	if isSysAdmin != nil && isSysAdmin.(bool) {
-		user.HasAdminRole = true
-	}
-
+	user = userInterface.(models.User)
```
reasonerjt: Maybe a type assertion here would be safer, otherwise there will be a panic.

And could you add a test case to cover the scenario when the data in session is not a user model?

Issue #4895, opened 2018-05-09
UAA user see a lot of forbidden error after promoted to admin.


reasonerjt:
Customer found that after login with UAA user, and promote this user to system admin in Harbor.  Login with UAA user again he see a lot of forbidden errors on the UI.

**root cause**
The authenticator of UAA didn't populate all data into the user model in subsequent login of UAA user.  And this is not covered in the TC.

reasonerjt:
Fixed in #4900 and cherry picked to 1.4 and 1.5 branch respectively.

nlowe:
@reasonerjt I see a `1.4.1` point release that contains this. Will `1.5` get a point release as well?

reasonerjt:
@nlowe 
Yes, but we'll need a couple of more weeks to include other fixes in `1.5.1`

Are you using UAA?

nlowe:
No. is that the only fix in `1.4.1`? I'm planning on upgrading to `1.5.0` here in 15ish minutes. If there's nothing else in that patch release I'll just go ahead with the upgrade.

reasonerjt:
@nlowe 
Yes the only fix in 1.4.1 is the UAA bug fix.

There will be a few UI fixes going into 1.5.1 in a addition to the UAA fix.  I don't think you have to skip 1.5.0

eymriseFrance:
Thanks hope I can make it...

PR #4900, opened 2018-05-09
Sync user info after UAA user is logged in

reasonerjt:
This is a fix to #4895.
Currently the user data will not be populated correctly after a user in
UAA is logged in.  This problem is fixed in this commit.

### DIFF
--- a/src/ui/auth/uaa/uaa.go
+++ b/src/ui/auth/uaa/uaa.go
@@ -68,6 +68,12 @@ func (u *Auth) OnBoardUser(user *models.User) error {
 	if len(user.Password) == 0 {
 		user.Password = "1234567ab"
 	}
+	fillEmailRealName(user)
+	user.Comment = "From UAA"
+	return dao.OnBoardUser(user)
+}
+
+func fillEmailRealName(user *models.User) {
 	if len(user.Realname) == 0 {
 		user.Realname = user.Username
 	}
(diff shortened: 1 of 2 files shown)


src/ui/auth/uaa/uaa.go
```diff
@@ -87,13 +91,11 @@ func (u *Auth) PostAuthenticate(user *models.User) error {
 	}
 	if dbUser == nil {
 		return u.OnBoardUser(user)
+	} else {
+		user.UserID = dbUser.UserID
+		user.HasAdminRole = dbUser.HasAdminRole
```
jessehu: Does this line fix the bug ? if dbUser is not nil, how about read the User info from database (similiar to "return dao.OnBoardUser(user)") ? There is no need to call `dao.ChangeUserProfile(*user, "Email", "Realname")` in line 97, unless we want to update UAA user's email and Realname when the UAA user logs in Harbor ?

reasonerjt: Thanks for the comment.
Correct, the intention here is to try to make sure the user record in DB is update to date with user info in UAA.

jessehu: Per discussison with @reasonerjt, UserID and HasAdminRole fields are defined by Harbor core, not by the different auth providers (e.g. uaa, ldap, ...), it's better to assign these 2 fields in a method of a common parent class (or say the caller of PostAuthenticate). So as to avoid missing these 2 assignment lines in PostAuthenticate() of new auth provider.

jessehu: Will track this enhancement in new ticket and address it in future. Now merge this commit to prepare with the patch release.

PR #4910, opened 2018-05-10
Improve the error handling capabilities of job service

steven-zou:
Return none zero code when job service exit with an error
Add retry when getting configs from admin server failed

### DIFF
--- a/src/jobservice/job/impl/context.go
+++ b/src/jobservice/job/impl/context.go
@@ -6,7 +6,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"reflect"
+	"time"
 
 	"github.com/vmware/harbor/src/adminserver/client"
 	"github.com/vmware/harbor/src/common"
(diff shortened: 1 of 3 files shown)


src/jobservice/job/impl/context.go
```diff
@@ -51,9 +57,24 @@ func NewContext(sysCtx context.Context, adminClient client.Client) *Context {
...
+	)
+
+	for counter == 0 || err != nil {
+		counter++
+		configs, err = c.adminClient.GetCfgs()
+		if err != nil {
+			logger.Errorf("Job context initialization error: %s\n", err.Error())
+			if counter < maxRetryTimes {
+				backoff := (int)(math.Pow(2, (float64)(counter))) + 2*counter + 5
+				time.Sleep(time.Duration(backoff) * time.Second)
```
reasonerjt: Could you add a log saying "retry in xxx seconds"?

steven-zou: added

PR #4962, opened 2018-05-14
Add label filter to filter chain

ywk253100:
Label filters are used to filter the resources according to the labels, adding the filters to the chain to enable the function.

### DIFF
--- a/src/replication/source/label_filter.go
+++ b/src/replication/source/label_filter.go
@@ -1,0 +1,88 @@
+// Copyright (c) 2018 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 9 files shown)


src/replication/source/label_filter.go
```diff
@@ -0,0 +1,97 @@
...
+
+// DoFilter filter the resources according to the label
+func (l *LabelFilter) DoFilter(items []models.FilterItem) []models.FilterItem {
+	candidates := []string{}
+	for _, item := range items {
+		candidates = append(candidates, item.Value)
+	}
+	log.Debugf("label filter candidates: %v", candidates)
+	result := []models.FilterItem{}
+	for _, item := range items {
```
steven-zou: why create new struct type for 'hasLabel'?

Issue #4664, opened 2018-04-13
repositories listing spinner is never removed


doctori:
Hi, 
the more I add project/repositories inside this wonderful project that is Harbor,
the slower it gets (which is fine, it's mostly based on the fact that the registry storage is based on S3)
after a certain point, the spinner that hide the empty list of repostories is never removed and the repositories list is never displayed like this : 
![image](https://user-images.githubusercontent.com/3294449/38728143-544a58e4-3f0f-11e8-8248-b347f95669bf.png)
I'm using harbor those images : 
```
vmware/nginx-photon:v1.4.0             
vmware/harbor-ui:v1.4.0                
vmware/clair-photon:v2.0.1-v1.4.0      
vmware/harbor-adminserver:v1.4.0       
vmware/registry-photon:v2.6.2-v1.4.0   
vmware/postgresql-photon:v1.4.0        
vmware/harbor-db:v1.4.0                
vmware/harbor-log:v1.4.0               
vmware/harbor-jobservice:v1.4.0        
```
with this docker version : 
```
root@pebitdocreg01:/var/log/harbor# docker info
Containers: 16
 Running: 9
 Paused: 0
 Stopped: 7
Images: 204
Server Version: 17.05.0-ce
Storage Driver: aufs
 Root Dir: /var/lib/docker/aufs
 Backing Filesystem: extfs
 Dirs: 403
 Dirperm1 Supported: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host macvlan null overlay
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 9048e5e50717ea4497b757314bad98ea3763c145
runc version: 9c2d8d184e5da67c95d601382adf14862e4f2228
init version: 949e6fa
Kernel Version: 3.16.0-5-amd64
Operating System: Debian GNU/Linux 8 (jessie)
OSType: linux
Architecture: x86_64
CPUs: 2
Total Memory: 7.804GiB
Name: xxxxxxxxxxxxxx
ID: LSED:OOLL:RKVR:XBY7:6E47:7LZI:GPJ3:2A75:Q4XG:4VZU:J3ND:ND3A
Docker Root Dir: /var/lib/docker
Debug Mode (client): false
Debug Mode (server): false
Http Proxy: http://isp-xxxxxxxx:3128/
No Proxy: localhost,127.0.0.1
[...]

Issue #2754, opened 2017-07-12
Request tags of repo in parallel when listing repositories


ywk253100:

nuzz:
I've been trying to trackdown slowness with getting tags from a repo. I'm running a harbor installation on a kubernetes cluster using s3 for registry storage. I have a few repos with a few tags (under 50). As an example it takes about 15 seconds to turn  a list of 32 tags. I have a few repos with hundreds of tags and the requests for tag info on the bigger repos wend up timing out after 1 minute.

example
curl -X GET https://myregistry/api/repositories/project/repo/tags

PR #4976, opened 2018-05-15
Request repository and tag list in parallel

ywk253100:
When using S3 or others as the registry storage, the request for repository and tag list may be very slow. This commit fixes this by parallel requesting the resources.

### DIFF
--- a/src/ui/api/repository.go
+++ b/src/ui/api/repository.go
@@ -154,39 +154,53 @@ func getRepositories(query *models.RepositoryQuery) ([]*repoResp, error) {
 		return nil, err
 	}
 
-	return assembleRepos(repositories)
+	return assembleReposInParallel(repositories), nil
 }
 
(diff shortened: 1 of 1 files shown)


ywk253100:
Fix #4664 and #2754

src/ui/api/repository.go
```diff
@@ -154,39 +154,53 @@ func getRepositories(query *models.RepositoryQuery) ([]*repoResp, error) {
 		return nil, err
 	}
 
-	return assembleRepos(repositories)
+	return assembleReposInParallel(repositories), nil
 }
 
-func assembleRepos(repositories []*models.RepoRecord) ([]*repoResp, error) {
-	result := []*repoResp{}
+func assembleReposInParallel(repositories []*models.RepoRecord) []*repoResp {
```
reasonerjt: I don't think we should eat errors here.

PR #4985, opened 2018-05-17
Make Helm Chart support Harbor 1.5.0

reasonerjt:
Update the helm chart of Harbor to use 1.5 image.  Specifically
`1.5.0-chart-patch` to include the fix to #4496 (for integration reason
this fix will not be merge to release branch).
Verified against kubenetes v1.10.0


contrib/helm/harbor/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 name: harbor
-version: 0.1.2
-appVersion: 1.4.0
+version: 0.1.3
```
jessehu: How about bump the version to 0.2.0 ?

contrib/helm/harbor/templates/adminserver/adminserver-cm.yaml
```diff
@@ -54,6 +54,7 @@ data:
   CLAIR_DB_PORT: "5432"
   CLAIR_DB: "{{ .Values.clair.postgresDatabase }}"
   CLAIR_DB_USERNAME: "{{ .Values.clair.postgresUser }}"
+  CLAIR_URL: "http://clair:6060"
```
jessehu: It should be `http://{{ template "harbor.fullname" . }}-clair:6060`

contrib/helm/harbor/templates/clair/clair-dpl.yaml
```diff
@@ -20,7 +20,7 @@ spec:
     spec:
       containers:
       - name: clair
-        image: {{ .Values.clair.image.repository }}:{{ .Values.clair.image.tag }}
+        image: {{ .Values.clair.image.repository }}:{{ .Values.clair.image.tag }}
```
jessehu: pls remove the trailing spaces

contrib/helm/harbor/templates/clair/clair-dpl.yaml
```diff
@@ -20,7 +20,7 @@ spec:
     spec:
       containers:
       - name: clair
-        image: {{ .Values.clair.image.repository }}:{{ .Values.clair.image.tag }}
+        image: {{ .Values.clair.image.repository }}:{{ .Values.clair.image.tag }} 
         imagePullPolicy: {{ .Values.clair.image.pullPolicy }}
         args: ["-insecure-tls", "-config", "/config/config.yaml"]
```
jessehu: should it be /etc/clair/config.yaml ?

contrib/helm/harbor/templates/ui/ui-dpl.yaml
```diff
@@ -38,8 +38,6 @@ spec:
             value: "1"
           - name: ADMINSERVER_URL
             value: "http://{{ template "harbor.fullname" . }}-adminserver"
-          - name: CLAIR_DB_HOST
```
jessehu: is this not needed in 1.5.0?

contrib/helm/harbor/values.yaml
```diff
@@ -26,10 +26,10 @@
 #       cpu: 100m
 
 persistence:
-  enabled: false
+  enabled: true
```
jessehu: If this enabled by default, users need to config all the storageClass sections. How about keep it as false by default?

contrib/helm/harbor/values.yaml
```diff
@@ -301,17 +302,30 @@ postgresql:
...
+## see https://github.com/kubernetes/charts/tree/master/stable/redis
+## for further configurables.
+redis:
+# Update needed in the cm that defines redis_url if usePassword is set to true.
+  usePassword: false
+  cluster:
+    enabled: false
+  master:
+    persistence:
+# temp workaround the permission issue.
```
jessehu: suggestion add a TODO here, and explain the details for permission issue (e.g. a bug url)

contrib/helm/harbor/templates/ui/ui-dpl.yaml
reasonerjt: Will call admin server to get the value.

contrib/helm/harbor/templates/adminserver/adminserver-cm.yaml
reasonerjt: It's not:
https://github.com/vmware/harbor/blob/master/contrib/helm/harbor/templates/clair/clair-svc.yaml#L8

We'll fix this in further refactory to make things more consistent.

contrib/helm/harbor/values.yaml
reasonerjt: I had to set it to true to pass the test when using local storage:
https://github.com/vmware/harbor/blob/master/contrib/helm/harbor/templates/registry/registry-ss.yaml#L55

contrib/helm/harbor/templates/clair/clair-dpl.yaml
reasonerjt: yes

contrib/helm/harbor/values.yaml
jessehu: Do you mean if set to false the chart deployment will fail ?

contrib/helm/harbor/Chart.yaml
jessehu: How about bump the version to 0.2.0 ?  It should be at least a minor version bump.

contrib/helm/harbor/values.yaml
reasonerjt: Correct.

contrib/helm/harbor/Chart.yaml
reasonerjt: ok

reasonerjt:
Updated the version of chart and the commit message according to review comment.

Issue #4364, opened 2018-03-07
When adding a new member, filter out existing members of project


reasonerjt:
@lweitzman commented on [Tue Mar 06 2018](https://github.com/vmware/harbor-planning/issues/31)

When adding a new member, filter out members that are already included in the project from the popup menu to avoid an error.

![Screen Shot 2018-03-06 at 10.16.11 AM.png](https://images.zenhubusercontent.com/56afa7c46caf56c9d38c1f12/7ec62f93-78c2-4bae-88e9-fb01797e2e1d)

lweitzman:
when adding a new member that already exists, if you 
paste the name in, you get an error 
then delete the input and start typing the name 
click on the drop down of the name
there is no error and it says the member was added successfully

![screen shot 2018-05-01 at 10 39 04 am](https://user-images.githubusercontent.com/626032/39477102-3cb51090-4d2c-11e8-9df3-a297f8dea49c.png)
![screen shot 2018-05-01 at 10 39 16 am](https://user-images.githubusercontent.com/626032/39477110-419bb0dc-4d2c-11e8-8381-8dd925037f57.png)
![screen shot 2018-05-01 at 10 39 50 am](https://user-images.githubusercontent.com/626032/39477115-455e0daa-4d2c-11e8-837b-d5594f254a0c.png)

PR #4986, opened 2018-05-17
Filter out existimg members when add a new member #4364

pengpengshui:
Filter out existimg members when add a new member #4364


lweitzman:
when typing in a name of a user already a member of a project, if you partially type in the name and select it from the dropdown you can add the same user over and over again. if you fully type in the name, you get the error  "Username already exists"

Issue #4847, opened 2018-05-02
Build db migrator in the make process


wy65701436:

wy65701436:
Done by PR: https://github.com/vmware/harbor/pull/5029

PR #5029, opened 2018-05-22
Add support for mysql to pgsql to DB migrator

wy65701436:
This commit is to implement db migration from mysql to pgsql, this is specific step for user to upgrade harbor across v1.5.0, as we have move harbor DB to pgsql from 1.5.0. It covers harbor db and notary db migration.

The upgrade path from v1.2.0 to v1.6.0;
v1.2.0(mysql) -> v1.5.0(mysql) -> v1.5.0(pgsql) -> v1.6.0(pgsql)

[Tested]
1, From 1.2.0 to current version(with and without notary)
2, From 1.3.0 to current version(with and without notary)
3, From 1.4.0 to current version(with and without notary)

[Usage]
1, upgrade harbor DB: docker run -it -e DB_USR=root -e DB_PWD=${pwd} -v /data/database/:/var/lib/mysql vmware/harbor-migrator:dev --db up
2, upgrade notary DB: docker run -it  -v /data/notary-db/:/var/lib/mysql -v /data/database:/var/lib/postgresql/data vmware/harbor-migrator:dev --db up

You need to run above two steps to migrate all DBs. If no require to migrate notary, just run step 1 only.

[Limitation]
For notary migration, there is a necessary precondition that you should not launch harbor before doing it, otherwise the migrator will give a waring message and stop the process.

[Where to get migrator]
From the local offline installer, this PR also fix issue #4847


tools/migration/db/run.sh
```diff
@@ -1,125 +1,505 @@
...
+            ## dump mysql
+            mysqldump --compatible=postgresql --default-character-set=utf8 --databases registry > /harbor-migration/db/registry.mysql     
+            stop_mysql $DB_USR $DB_PWD
+
+            ## migrate 1.5.0-mysql to 1.5.0-pqsql.
+            python /harbor-migration/db/pgsql_migrator.py /harbor-migration/db/registry.mysql /harbor-migration/db/registry.pgsql
+
+            launch_pgsql $PGSQL_USR
+            psql -U $PGSQL_USR -f /harbor-migration/db/schema/registry_from_$cur_version.pgsql
+            psql -U $PGSQL_USR -f /harbor-migration/db/schema/notaryserver.pgsql
```
reasonerjt: Why need to call notary pgsql in `up_harbor`?

tools/migration/db/run.sh
```diff
@@ -1,125 +1,505 @@
...
+        fi
+    fi
+
+    # $cur_version <='1.5.0', $target_version >'1.5.0', it needs to upgrade to 1.5.0.mysql => 1.5.0.pgsql => target_version.pgsql.
+    if [ $v1_com != 1 ] && [ $v2_com = 1 ]; then
+        if [ $ISMYSQL != true ]; then
+            echo "Please make sure to mount the correct the data volumn."
+            return 1
+        else
+            alembic_up mysql '1.5.0' false
```
reasonerjt: the third parm should be removed.

tools/migration/Dockerfile
```diff
@@ -1,14 +1,43 @@
...
+    && chown -R mysql:mysql /var/lib/mysql /var/run/mysqld \
+    && chmod 777 /var/run/mysqld /docker-entrypoint-initdb.d /docker-entrypoint-updatedb.d \
     && tdnf install -y mariadb-devel python2 python2-devel python-pip gcc \
     linux-api-headers glibc-devel binutils zlib-devel openssl-devel \
     && pip install mysqlclient alembic \
-    && tdnf clean all \
-    && mkdir -p /harbor-migration
+    && mkdir -p /harbor-migration \
+    && touch /etc/localtime.bak \
+    && tdnf install -y sed shadow gzip postgresql >> /dev/null\
```
reasonerjt: We should merge the `tdnf install`

tools/migration/Dockerfile
```diff
@@ -1,14 +1,43 @@
...
+    && groupadd -r postgres --gid=999 \
+    && useradd -r -g postgres --uid=999 postgres \
+    && mkdir -p /docker-entrypoint-initdb.d \
```
reasonerjt: This is not needed.

tools/migration/db/run.sh
```diff
@@ -1,125 +1,505 @@
...
+    return 1
+}
+
+# This function is only for notary <= 1.5.0 mysql to pgsql.
+function up_notary {
+
+    set +e
+    chown -R 10000:10000 /var/lib/mysql
+    mysqld &
+    for i in {60..0}; do
```
reasonerjt: Could we update `launch_mysql` to make it support different scenarios and pass an empty string as password?

tools/migration/db/pgsql_migrator.py
```diff
@@ -0,0 +1,174 @@
...
+                insert_lines.append(line)
+                continue
+            else:    
+                insert_lines.append(line)
+    
+    write_database(pgsql_dump, "notarysigner")
+    write_insert(pgsql_dump, insert_lines)
+    write_sequence(pgsql_dump, "private_keys", "id")
+
+def drop_line(line):
```
reasonerjt: it's not needed if all we wanna catch is `INSERT INTO` this method is not needed.

tools/migration/db/pgsql_migrator.py
```diff
@@ -0,0 +1,174 @@
...
+        if drop_line(line):
+            continue
+
+        # catch insert
+        if line.startswith("INSERT INTO"):
+            if line.find("\\") != -1:
+                line = line.replace('\\', '')
+                insert_lines.append(line)
+                continue
+            elif line.startswith('INSERT INTO "change_category"') | line.startswith('INSERT INTO "changefeed"') | line.startswith('INSERT INTO "schema_migrations"'):
```
reasonerjt: We should not skip the `INSERT INTO` 'change_category' and 'changefeed'

tools/migration/db/pgsql_migrator.py
```diff
@@ -0,0 +1,174 @@
...
+
+if __name__ == "__main__":
+    if sys.argv[1].find("registry") != -1:
+        do_parse_registry_db(sys.argv[1], sys.argv[2])
+    elif sys.argv[1].find("notaryserver") != -1:
+        do_parse_notary_server_db(sys.argv[1], sys.argv[2])
+    elif sys.argv[1].find("notarysigner") != -1:
+        do_parse_notary_signer_db(sys.argv[1], sys.argv[2])
+    else:
+        print ("Unsupport mysql dump file.")
```
reasonerjt: Better print the sys.argv[1] into error message to provide more details.

tools/migration/db/pgsql_migrator.py
reasonerjt: This is not needed as we need only `INSERT INTO`

reasonerjt:
Two overall comments:
1)  There should not be multiple pgsql script for different schema version, there should be only one for 1.5
2)  The run.sh should be refactored that there's one script to specifically handle mysql-> pgsql

eymriseFrance:
Merci

tools/migration/db/util/alembic.sh
```diff
@@ -0,0 +1,41 @@
...
+    export PYTHONPATH=$PYTHONPATH:/harbor-migration/db/alembic
+    
+    if [ $db_type = "mysql" ]; then
+        source /harbor-migration/db/alembic/alembic.tpl > /harbor-migration/db/alembic/alembic.ini     
+        echo "Performing upgrade $target_version..."
+        alembic -c /harbor-migration/db/alembic/alembic.ini current
+        alembic -c /harbor-migration/db/alembic/alembic.ini upgrade $target_version
+        rc="$?"
+        alembic -c /harbor-migration/db/alembic/alembic.ini current
+    elif [ $db_type = "mysql" ]; then
```
reasonerjt: This is still incorrect?

tools/migration/db/util/alembic.sh
```diff
@@ -0,0 +1,41 @@
...
+    local target_version="$2"
+
+    export PYTHONPATH=$PYTHONPATH:/harbor-migration/db/alembic
+    
+    if [ $db_type = "mysql" ]; then
+        source /harbor-migration/db/alembic/alembic.tpl > /harbor-migration/db/alembic/alembic.ini     
+        echo "Performing upgrade $target_version..."
+        alembic -c /harbor-migration/db/alembic/alembic.ini current
+        alembic -c /harbor-migration/db/alembic/alembic.ini upgrade $target_version
+        rc="$?"
```
reasonerjt: how is `rc` used?

tools/migration/db/util/alembic.sh
```diff
@@ -0,0 +1,41 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+source $PWD/db/util/pgsql.sh
+source $PWD/db/util/mysql.sh
```
reasonerjt: Are these `source` needed?

tools/migration/db/run.sh
```diff
@@ -1,125 +1,247 @@
...
+        exit 0
+    fi
+
+    init
+
+    local key="$1"
+
+    case $key in
+    up|upgrade)
+        upgrade
```
reasonerjt: Where's the `$1`?

tools/migration/db/util/mysql_pgsql_1_5_0.sh
```diff
@@ -0,0 +1,81 @@
...
+#
+
+source $PWD/db/util/pgsql.sh
+source $PWD/db/util/alembic.sh
+
+DBCNF="-hlocalhost -u${DB_USR}"
+
+function mysql_2_pgsql_1_5_0 {
+    
+    alembic_up mysql '1.5.0'
```
reasonerjt: what if this fails?

tools/migration/db/util/mysql_pgsql_1_5_0.sh
```diff
@@ -0,0 +1,81 @@
...
+function up_notary {
+
+    set +e
+    if [[ $(mysql $DBCNF -N -s -e "select count(*) from information_schema.tables \
+                where table_schema='notaryserver' and table_name='tuf_files'") -eq 0 ]]; then
+        echo "no content trust data needs to be updated."
+        return 0
+    else
+
+        ## it's not a clean notary db, so cannot execute the create tables step.
```
reasonerjt: Doesn't Harbor registry DB has the same issue?

tools/migration/db/util/alembic.sh
wy65701436: will fix the typo.

tools/migration/db/util/alembic.sh
wy65701436: could be removed.

tools/migration/db/util/mysql_pgsql_1_5_0.sh
wy65701436: the shell will be exited with non-zero code as "set -e"

tools/migration/db/util/mysql_pgsql_1_5_0.sh
wy65701436: no, the init() handles the repeats run.

tools/migration/db/run.sh
wy65701436: at line 225.

tools/migration/db/run.sh
```diff
@@ -1,125 +1,247 @@
...
+cur_version=""
+PGSQL_USR="postgres" 
+
+function init {
+    if [ "$(ls -A /var/lib/mysql)" ]; then
+        # As after the first success run, the data will be migrated to pgsql,
+        # the PG_VERSION should be in /var/lib/mysql if user repeats the UP command.
+        if [ -e '/var/lib/mysql/PG_VERSION' ]; then
+            ISPGSQL=true
+        elif [ "$(ls -A /var/lib/mysql/mysql)" ]; then
```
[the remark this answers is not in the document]
wy65701436: -d is to check if a directory exists. 
ls -A is to check if a directory is empty.  

I believe later is better.

tools/migration/db/util/mysql_pgsql_1_5_0.sh
wy65701436: will add set -e to each of these shell scripts.

tools/migration/db/util/alembic.sh
wy65701436: could be removed.

tools/migration/db/util/mysql_pgsql_data_converter.py
```diff
@@ -0,0 +1,157 @@
...
+                                    'INSERT INTO "access_log" ("log_id", "username", "project_id", "repo_name", "repo_tag", "guid", "operation", "op_time")')
+                insert_lines.append(line)
+                continue
+            # pgsql doesn't support 0 as a time data, change it to the minimum value.
+            elif line.find("0000-00-00 00:00:00") != -1:
+                line = line.replace("0000-00-00 00:00:00", "0001-01-01 00:00:00")
+                insert_lines.append(line)
+                continue
+            # it's only for con_str, mysql dump the con_str to {\"kind\":\"Manual\",\"schedule_param\":null}.
+            elif line.find('INSERT INTO "replication_policy"') != -1:
```
reasonerjt: Please also consider the table `img_scan_overview`

Issue #840, opened 2016-09-27
User API doesn't support changing password without providing old password by admin user


wknet123:
User API doesn't support changing password without providing old password by admin user

reasonerjt:
Due to security compliance, let's defer this feature

Issue #1992, opened 2017-04-11
new password should not the same with the old one


sigsbee:
when user change password, the new password should not the same with the old.

steven-zou:
Should be enhanced from backend api side

PR #5030, opened 2018-05-22
Update change password API

ywk253100:
Modify the changing password API to support that admin user can change the password of normal users without old password


ywk253100:
Fix #840 and #1992.

Issue #5001, opened 2018-05-19
Replication breaks when target registry is configured to use relative urls


Capitrium:
When replicating images, the target URL for uploading the layer blobs is built from the `Location` header in the response to a `POST` request to the target registry:

https://github.com/vmware/harbor/blob/d3930ae17c0a044cf3d149c6a7aaef803bfd226a/src/common/utils/registry/repository.go#L348-L364

This breaks replication when the target registry is configured to use relative URLs:
```yaml
http:
  relativeurls: true
```

The `Location` header has no scheme/host in this case, so the URL in the `PUT` requests is malformed, as shown in the replication job logs:
```
2018-05-18T22:40:22Z [INFO] transferring blob sha256:yyy of ws_server/ws_server:test to http://harbor-1.harbor ...
2018-05-18T22:40:22Z [ERROR] [transfer.go:443]: an error occurred while pushing blob sha256:yyy of ws_server/ws_server:test to http://harbor-1.harbor : Put /v2/ws_server/ws_server/blobs/uploads/abcd?_state=xxx&digest=sha256:yyy: unsupported protocol scheme ""
```

The `Location` header is apparently needed for the `_state` parameter in the `PUT` request URL. If the target registry uses relative URLs, should the scheme/host in the `PUT` requests be populated from the repository's endpoint (`r.Endpoint.String()`)?

Our environment:
- k8s 1.9.3
- harbor 1.4.0
- docker 17.12.1-ce

ywk253100:
Fixed.

PR #5034, opened 2018-05-24
Fix replication issue when the remote registry enables relativeurls

ywk253100:
The location header returned by the remote registry contains no scheme and host parts if "relativeurls" is enabled,
this commit fix it by adding them at the beginning of location.


ywk253100:
Fixes #5001.

PR #5051, opened 2018-05-29
Add dropdown for label color panel and modify color disappear on Mac 

pengpengshui:
Add dropdown for color panel

modify color disappear on Mac


ninjadq:
most of the code is about visual change, can you provide an environment

PR #5056, opened 2018-05-30
Add function about admin reset user's password

pengpengshui:
Add function about admin reset user's password


ninjadq:
Can you provide a environment for test?

PR #5062, opened 2018-05-31
Setting custom database name is implemented.

odg0318:
I added that custom database name can be set in harbor.cfg.


odg0318:
If this code were merged, db migrator should be changed available to set database name.

wy65701436:
Thanks for the PR, but it's not workable for current code base as the database is hard coded in the init sql. Actually, it's not a configurable item now.

https://github.com/vmware/harbor/blob/master/make/photon/db/registry.sql#L3

odg0318:
@wy65701436
I knew my code is not enough to change database name after making pr.
I read codes more carefully and then there is no consideration for external database.
In my office, database must be separated from application.
What do you think of using external db?

wy65701436:
your PR is to make the data base name configurable, is it your request? Your office doesn't accept "registry" as the Harbor data base name? 

For how to configure external DB, you can refer to the HA guide(https://github.com/vmware/harbor/blob/master/docs/high_availability_installation_guide.md), just go through the part of DB configuration. 

Basically, you need to modify the db_host in .cfg and set up the DB schema of harbor on your env.

odg0318:
@wy65701436 When I installed version 1.3.0, I added database name to harbor.cfg. And also I dumped registry.sql used in harbor-db image and manually imported it to external database. Harbor worked well.

Now upgrading 1.5.0 is required to use enhanced replication feature. But db migrator supports only mounting database raw files to access exsting database. For security, it is denied to connect to database machine to copy them.

In my opinion, to change this, a lot of codes should be changed but using extenal database is very normal in every system.

wy65701436:
Again, this request is about to make data base name configurable. I am talking about the code change, it's not workable for the current code base, because that you have to do further manual steps to modify the "registry". So I don't think this PR is acceptable for now, and this PR cannot resolve your security problem.

For now, Harbor only provides the migrator to update the raw files. If it's denied on your office, I think you need to do it manually in the external DB machine. You can open one issue to discuss the security issue on DB migrator.

odg0318:
Okay, I see. I will create new issue to discuss. I will close. Thanks

PR #5110, opened 2018-06-08
Hidden pwd button if auth_mode is not db_auth

pengpengshui:
if auth_node is not db_auth, should hidden reset password button in user component .


reasonerjt:
Please fix the  typo in commit message and add more details, such as what button should be hidden and why.

PR #5112, opened 2018-06-08
Add clair data migration in DB migrator

wy65701436:
this commit to fix issue https://github.com/vmware/harbor/issues/5097, add the clair data migration.

[Usage]:
docker run -it -e DB_USR=root -e DB_PWD=${pwd} -v /data/database/:/var/lib/postgresql/data -v /data/clair-db/:/clair-db vmware/harbor-migrator:dev --db up

[Limitation]
For clair migration, there is a necessary precondition that you should not launch harbor before doing it, otherwise the migrator will give a waring message and stop the process.

### DIFF
--- a/tools/migration/db/util/mysql_pgsql_1_5_0.sh
+++ b/tools/migration/db/util/mysql_pgsql_1_5_0.sh
@@ -55,7 +55,7 @@ function up_notary {
                 WARNING: Notary migration will only allow anyone haven't migrated notary or 
                          launched harbor yet. 
                          If you want to migrate notary data, please delete all the notaryserver 
-                         and notarysigner DB tables in pgsql manually fistly.
+                         and notarysigner DB tables in pgsql manually firstly.
                 *******************************************************************************
 EOF
             exit 0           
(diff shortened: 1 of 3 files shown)


tools/migration/db/util/mysql_pgsql_1_5_0.sh
```diff
@@ -79,4 +79,38 @@ EOF
...
+
+    set +e
+    if [[ $(psql -U $1 -d postgres -t -c "select count(*) from vulnerability;") -eq 0 ]]; then
+        echo "no vulnerability data needs to be updated."
+        return 0
+    else
+    
+        ## it's not a clean clair db, so cannot execute the import step.
+        ## fail at here to call user to clean DB, then to run clair db migration.
+        if [[ $(psql -U $1 -d postgres -t -c "select count(*) from pg_tables where schemaname='public';") -ne 0 ]]; then
```
reasonerjt: This should be checked when pgsql is running with new directory as the data directory?

tools/migration/db/util/mysql_pgsql_1_5_0.sh
```diff
@@ -79,4 +79,38 @@ EOF
...
+            cat >&2 <<-EOF
+                *******************************************************************************
+                WARNING: Clair migration will only allow anyone haven't migrated clair or 
+                        launched harbor yet. 
+                        If you want to migrate clair data, please delete all the clair DB tables 
+                        in pgsql manually fistly.
```
reasonerjt: fix typo `fistly`

PR #5121, opened 2018-06-12
If group already exist return 409 when adding a LDAP user group

stonezdj:
Not to create the LDAP user group again or update the user group name.


reasonerjt:
How about adding a test case?

Issue #4861, opened 2018-05-03
Image replication based on labels.


reasonerjt:
- API/schema change
- UI change
- **Open question** Shall we support replication images in different projects if the images are labeled with system level label?
- What if the label is updated/removed?

PR #5157, opened 2018-06-15
Add replication rule label filter function #4861

pengpengshui:


reasonerjt:
Merge to feature branch.

PR #5163, opened 2018-06-19
Fix bug for Harbor UI not updating status of Registry

hasonhai:
Because of newline, notification endpoint configuration is ignored from the config.yml that is mounted into the registry pod. This causes Harbor UI not able to update what happened on the registry and break the role base access control provide by Harbor.


ywk253100:
@hasonhai Thank you for your contribution. But everything works well in my environment even with the new line in the config file. Could you provide more detail information about the fix?

ywk253100:
Closing due to inactivity.

PR #5164, opened 2018-06-20
Handle Invalid syntax and not found error

stonezdj:
Given an invalid DN, return 400 error
If the group DN is not found, return 404 error

### DIFF
--- a/src/common/utils/ldap/ldap.go
+++ b/src/common/utils/ldap/ldap.go
@@ -16,6 +16,7 @@ package ldap
 
 import (
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"net/url"
 	"strconv"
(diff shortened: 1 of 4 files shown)


src/common/utils/ldap/ldap.go
```diff
@@ -336,22 +362,25 @@ func (session *Session) Close() {
...
 func (session *Session) SearchGroupByDN(groupDN string) ([]models.LdapGroup, error) {
-	ldapGroupConfig, err := config.LDAPGroupConf()
-	log.Debugf("Ldap group config: %+v", ldapGroupConfig)
-	if err != nil {
-		return nil, err
+	if _, err := goldap.ParseDN(groupDN); err != nil {
+		return nil, ErrDNSyntax
+	}
+	groupList, err := session.searchGroup(groupDN, session.ldapGroupConfig.LdapGroupFilter, "", session.ldapGroupConfig.LdapGroupNameAttribute)
+	if serverError, ok := err.(*goldap.Error); ok {
```
ywk253100: Will panic if err is nil

stonezdj: This type assertion will not panic when there are two return values.

PR #5177, opened 2018-06-22
Update ingress rule for Harbor chart

ywk253100:
Rewrite /v2/* to UI instead of registry to apply the filters to the requests


ywk253100:
Issue #5189 opened.

PR #5180, opened 2018-06-23
Helm: Options to disable ingress objects

jkroepke:
If you are on other kubernetes distros like Openshift, you shouldn't use Ingress.


reasonerjt:
@jkroepke could you please rebase to resove the conflict?

jkroepke:
Sure.

PR #5190, opened 2018-06-26
Fix issue that harbor tile can not save customized settings 

stonezdj:
If customized settings are matched by skip_reload_pattern in harbor.cfg, then these values are kept in the config store will be unchange after reload_config. This feature is used by harbor tile to reconfigure tile.

### DIFF
--- a/make/prepare
+++ b/make/prepare
@@ -275,6 +275,10 @@ if rcp.has_option("configuration", "redis_url"):
 else:
     redis_url = ""
 
+if rcp.has_option("configuration", "skip_reload_env_pattern"):
+    skip_reload_env_pattern = rcp.get("configuration", "skip_reload_env_pattern")
+else:
+    skip_reload_env_pattern = "$^"
 storage_provider_name = rcp.get("configuration", "registry_storage_provider_name").strip()
 storage_provider_config = rcp.get("configuration", "registry_storage_provider_config").strip()
 # yaml requires 1 or more spaces between the key and value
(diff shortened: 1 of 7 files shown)


src/adminserver/systemcfg/systemcfg.go
```diff
@@ -363,7 +372,18 @@ func LoadFromEnv(cfgs map[string]interface{}, all bool) error {
 		}
 	}
 
+	reloadCfg := os.Getenv("RESET")
+	skipReg := os.Getenv("SKIP_RELOAD_ENV_REG")
+	skipReg = strings.TrimSpace(skipReg)
+	if len(skipReg) == 0 {
+		skipReg = "$^" // doesn't match any string by default
+	}
+	skipMatcher := regexp.MustCompile(skipReg)
```
reasonerjt: This maybe too strict that if user input a wrong pattern it will panic, how about if it fails to compile the pattern match will not work?

make/prepare
```diff
@@ -275,6 +275,10 @@ if rcp.has_option("configuration", "redis_url"):
 else:
     redis_url = ""
 
+if rcp.has_option("configuration", "skip_reload_env_reg"):
+    skip_reload_env_reg = rcp.get("configuration", "skip_reload_env_reg")
```
reasonerjt: Is calling the attribute `reg` a convention?  I thought normally it should be called as "pattern"?

src/adminserver/systemcfg/systemcfg.go
```diff
@@ -249,13 +251,20 @@ func Init() (err error) {
 	loadAll := false
 	cfgs := map[string]interface{}{}
 
-	if os.Getenv("RESET") == "true" {
-		log.Info("RESET is set, will load all configurations from environment variables")
+	reloadKey := os.Getenv("RELOAD_KEY")
```
reasonerjt: Maybe add some comment to explain why we need to introduce this RELOAD_KEY?

Additionally we didn't test the new code...

src/adminserver/systemcfg/systemcfg.go
stonezdj: Updated

make/prepare
stonezdj: Updated

src/adminserver/systemcfg/systemcfg.go
stonezdj: Updated

PR #5197, opened 2018-06-28
Initialise Harbor DB schema in Harbor UI/adminserver container

### LINKED ISSUE #5040: Move the .sql script from DB container to harbor container.
Currently the `.sql` script of `harbor-db` is built into DB docker image, this becomes an issue when user wants to configure Harbor against a remote DB, common in a HA deployment.

The initial sql script should be moved out, so when Harbor instance starts it will access the DB and initialize it.

reasonerjt:
This commit fixes #5040, the harbor-db image will only contain empty
databases, and harbor ui container will use migrate tool to run initial
SQL scripts to do initialization.  This is helpful for the case to
configure Harbor against external DB or DBaaS like RDS for HA deployment
However, this change will results some confusion as there are two tables
to track schema versions have been using alembic for migration, for this
release we'll try to use alembic to mock a `migration` table during
upgrade so the migrator will be bypassed, in future we'll consider to
consolidate to the golang based migrator.
Another issue is that the UI and adminserver containers will access DB
after start up in different congurations, can't ensure the sequence, so
both of them will try to update the schema when started up.


make/photon/adminserver/Dockerfile
```diff
@@ -7,6 +7,8 @@ RUN tdnf erase vim -y \
     && groupadd -r -g 10000 harbor && useradd --no-log-init -r -g 10000 -u 10000 harbor \
     && mkdir /harbor/
 COPY ./make/dev/adminserver/harbor_adminserver ./make/photon/adminserver/start.sh /harbor/
+COPY ./make/migrations /harbor/migrations
```
wy65701436: please add more comment for this.

src/common/dao/pgsql.go
```diff
@@ -82,3 +90,31 @@ func (p *pgsql) Register(alias ...string) error {
...
+	m, err := migrate.New(srcURL, dbURL)
+	if err != nil {
+		return err
+	}
+	log.Infof("Upgrading schema for pgsql ...")
+	err = m.Up()
+	if err == migrate.ErrNoChange {
+		log.Infof("No change in schema, skip.")
+	} else if err != nil {
+		log.Errorf("Failed to upgrade schema, error: %q", err)
```
wy65701436: needs to handle close() in the up() error.

PR #5203, opened 2018-06-28
Update migrator 1 6 0

ninjadq:
1. Add new alembic_pg folder for postgres
2. Add migration file for 1.6.0
3. Update version to 1.6.0
4. update migrator dockerfile


tools/migration/db/alembic_pg/alembic.tpl
```diff
@@ -0,0 +1,75 @@
+echo "
```
wy65701436: can we use the alembic/pg and alembic/mysql to separate these two DBs?

tools/migration/db/util/alembic.sh
```diff
@@ -19,17 +19,31 @@ set -e
...
+
+        if [[ -z "${PG_HOST}" ]]; then
+            export DB_HOST=localhost
+        fi
+
+        if [[ -z "${PG_PORT}" ]]; then
+            export DB_PORT=5432
+        fi
+        export PYTHONPATH=/harbor-migration/db/alembic_pg
         echo "TODO: add support for pgsql."
```
wy65701436: remove it?

reasonerjt:
Please fix ut.

Issue #5205, opened 2018-06-28
Replication failed because of insufficient scope


YEXINGZHE54:
harbor replication job failed, because of insufficient scope, reported in registry log

1. log content:
```
time="2018-06-28T04:39:16.00077525Z" level=warning msg="error authorizing context: insufficient scope" go.version=go1.7.3 http.request.host="XXX" http.request.id=0c4126b6-7202-411f-9d43-0ae6b4775c34 http.request.method=POST http.request.remoteaddr=XXX http.request.uri="/v2/custom/test/blobs/uploads/" http.request.useragent=harbor-registry-client instance.id=96b32b3a-e43e-49e3-987a-b457f5b3a3a2 service=registry vars.name="custom/test" version=v2.6.2
127.0.0.1 - - [28/Jun/2018:04:39:15 +0000] "POST /v2/custom/test/blobs/uploads/ HTTP/1.1" 401 224 "" "harbor-registry-client"
```

2. I change the code and change the scope of auth token, and then replication is ok

reasonerjt:
@YEXINGZHE54 I'm a little confused about the error and why the pull scope is needed.
Could you provide more details and context?

ywk253100:
Closing as inactive.

PR #5206, opened 2018-06-28
fix: add pull scope for push image

YEXINGZHE54:
fix issue: #5205


cd1989:
The problem already got fixed in https://github.com/goharbor/harbor/pull/5781, this PR can be closed now.

PR #5213, opened 2018-06-29
Label filter UI for replication

steven-zou:
Let user select the label from both system and project labels and use as a filter for the replication


ninjadq:
LGTM

lweitzman:
to be complete, this issue should include a screen grab of the new UI

PR #5230, opened 2018-07-02
Define the related handlers to handler the requests of chart server/repo

steven-zou:
add chartserver directory to put chart server code
add controller to coordinate the flow
add base/repo/manipulation handlers to handle requests
add operator to parse more details of chart version
call dep ensure


steven-zou:
Invalid, close it

PR #5257, opened 2018-07-05
Add registry controller

wy65701436:
it's a http server that holds GC api to start GC on the registry storage. This server shares the
same storage with registry, use UI_SECRET to auth, and it could
only be called with job service.


reasonerjt:
As we discussed, please split the PR into two branches/commits and add some tests.

PR #5265, opened 2018-07-07
Add the registry controller httpserver

wy65701436:
docker regsitry. This version has the API to call regsitry GC with jobservice
secret. Seprates it into a standalone container as do not want to invoke two
processes in one container.

It needs to mount the registry storage into this container in order to do GC,
and needs to copy the registry binary into it.


src/registryctl/main.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+// RegistryCtl for registry controller
+type RegistryCtl struct {
+	ServerConf config.Configuration
+	Handler    http.Handler
+}
+
+// Start the registry controller
+func (s *RegistryCtl) Start() error {
```
ywk253100: Seems the returned error is not needed as the function will hang?

src/registryctl/client/client.go
```diff
@@ -0,0 +1,100 @@
...
+
+// Client defines methods that an Regsitry client should implement
+type Client interface {
+	// Health tests the connection with registry server
+	Health() error
+	// StartGC enable the gc of registry server
+	StartGC() (*api.GCResult, error)
+}
+
+type client struct {
```
ywk253100: Do we need to consider that the server serves with HTTPS?

src/registryctl/client/client.go
```diff
@@ -0,0 +1,100 @@
...
+		client.client = common_http.NewClient(nil, authorizer)
+	}
+	return client
+}
+
+// Health ...
+func (c *client) Health() error {
+	addr := strings.Split(c.baseURL, "://")[1]
+	if !strings.Contains(addr, ":") {
+		addr = addr + ":80"
```
ywk253100: How about the server that serves with HTTPS?

src/registryctl/client/client.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+	data, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("Failed to start GC: %d", resp.StatusCode)
```
ywk253100: Log the response body data in case for debugging?

src/registryctl/auth/secret.go
```diff
@@ -0,0 +1,61 @@
...
+package auth
+
+import (
+	"errors"
+	"net/http"
+	"strings"
+)
+
+//HarborSecret is the prefix of the value of Authorization header.
+const HarborSecret = "Harbor-Secret "
```
ywk253100: How about using the const defined in common/secret?

src/registryctl/api/registry.go
```diff
@@ -0,0 +1,57 @@
...
+// StartGC ...
+func StartGC(w http.ResponseWriter, r *http.Request) {
+	cmd := exec.Command("/bin/bash", "-c", "registry garbage-collect "+regConf)
+	var outBuf, errBuf bytes.Buffer
+	cmd.Stdout = &outBuf
+	cmd.Stderr = &errBuf
+
+	start := time.Now()
+	if err := cmd.Run(); err != nil {
+		log.Errorf("Fail to execute GC: %s", errBuf.String())
```
ywk253100: The log only prints the error message contained in stderr, but how about the err itself?

src/registryctl/api/registry.go
```diff
@@ -0,0 +1,57 @@
...
+		handleInternalServerError(w)
```
ywk253100: Need return here and all other HTTP error handlers. BTW, I think we should reuse the implements in common package for the HTTP error handlers.

src/registryctl/api/health.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+// Health ...
+func Health(w http.ResponseWriter, r *http.Request) {
+	if r.Method == "GET" {
+		if err := writeJSON(w, "healthy"); err != nil {
+			log.Errorf("Failed to write response: %v", err)
+			return
+		}
+	} else {
```
ywk253100: Already handle this in the router?

src/common/registryctl/client.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	log.Infof("initializing client for reigstry %s ...", registryCtlURL)
+	cfg := &client.Config{
+		Secret: os.Getenv("JOBSERVICE_SECRET"),
+	}
+	RegistryCtlClient = client.NewClient(registryCtlURL, cfg)
+}
+
+func initKeyProvider() {
```
ywk253100: Seems this function is useless.

make/common/templates/ui/env
```diff
@@ -6,3 +6,4 @@ GODEBUG=netdns=cgo
 ADMINSERVER_URL=$adminserver_url
 UAA_CA_ROOT=/etc/ui/certificates/uaa_ca.pem
 _REDIS_URL=$redis_url
+REGISTRY_CONTROLLER_URL=$registry_controller_url
```
ywk253100: Does UI need to talk to the controller directly? If it does, put it into the env of adminserver and UI gets the URL from adminserver.

Issue #5078, opened 2018-06-01
Strange behaviour when deleting an repo after GC


wy65701436:
[Steps]:
1, Push one docker image into harbor.
2, Delete the repository.
3, Run GC with cmd: /usr/bin/registry garbage-collect /etc/registry/config.yml
4, Re-push the deleted images into harbor.

After that, I get the repo size with 0. And, I tried to delete it, UI shows as success and 200 from API, but actually the repo cannot be deleted.

[Root Cause]
After investigation, the root cause should be the GC bug of docker registry v2.6.2(cache invalidation). See issues:
https://github.com/docker/distribution/issues/2270
https://github.com/docker/distribution/issues/2094
https://github.com/docker/distribution/issues/2434
From the issue response, they could provide fix on 2.7, but no mention the schedule.

[Workaround]
Re-start registry and re-push the images.

[Issue on harbor]
For harbor, we eat the 404 error from docker registry, that's why it can get 200 from API when to delete the repository and cause the strange behaviour. https://github.com/vmware/harbor/blob/master/src/ui/api/repository.go#L295

[Log]
Jun  1 08:38:00 172.18.0.1 ui[1463]: 2018-06-01T15:38:00Z [ERROR] [repository.go:538]: failed to get v2 manifest of library1/tomcat:latest: 404 {"errors":[{"code":"MANIFEST_UNKNOWN","message":"manifest unknown","detail":{"Name":"library1/tomcat","Revision":"sha256:8238fb3b0f8e15219399069a4aa78e88cbbce28834de5f1e55749067b82d14dc"}}]}

clouderati:
@reasonerjt – I know we talked about this but can't recall if it was completely fixed in #5273. Thoughts?

wy65701436:
from v1.6.1, harbor has upgraded to use redis as the cache of registry, and the GC job will delete all the keys from cache, which could workaround this issue. 

Close it as fix.

lmdkfs:
i try to upgdate  to harbor v1.6.1. how can i view this gc job, i try to visite `api/system/gc`, the response is 404

lmdkfs:
i also encountered this problem, so i upgrade harbor to 1.6.1, but this problem still  exists,  how can to use gc, thanks @wy65701436

louyiping:
Also meet this problem with 1.6.0 version. Any idea?

wy65701436:
[...]

PR #5273, opened 2018-07-09
Update docker registry cache from inmemory to redis.

wy65701436:
It gives Harbor the capability to controll the cache of docker
registry, and the workaround for cache invalidation bug caused
by garbage collection, that is clean cache in GC job.

For more details, see Harbor issue #5078.

### DIFF
--- a/make/prepare
+++ b/make/prepare
@@ -415,7 +415,8 @@ render(os.path.join(templates_dir, "ui", "env"),
         adminserver_url = adminserver_url
         )
 
-registry_config_file = "config_ha.yml" if args.ha_mode else "config.yml"
+registry_config_file_ha = "config_ha.yml"
+registry_config_file = "config.yml"
 if storage_provider_name == "filesystem":
     if not storage_provider_config:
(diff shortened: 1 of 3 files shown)


make/prepare
```diff
@@ -426,6 +427,13 @@ storage_provider_conf_list = [storage_provider_name + ':']
 for c in storage_provider_config.split(","):
     storage_provider_conf_list.append(c.strip())
 storage_provider_info = ('\n' + ' ' * 4).join(storage_provider_conf_list)
+render(os.path.join(templates_dir, "registry", registry_config_file_ha),
+    registry_conf,
+    storage_provider_info=storage_provider_info,
+    public_url=public_url,
+    ui_url=ui_url,
+    redis_url=redis_url)
```
reasonerjt: What if redis_url contains username and password?

reasonerjt:
I'll approve but please open another issue to handle the case when the redis requires a password.
It is important when user wants to integrate with external redis instances.

wy65701436:
https://github.com/vmware/harbor/issues/5275

PR #5294, opened 2018-07-12
add sync registry env

timchenxiaoyu:
add sync env to determine whether sync registry to db
fix https://github.com/vmware/harbor/issues/5271


timchenxiaoyu:
thanks，I have committd !

PR #5304, opened 2018-07-13
Add API routes of chart repo services in harbor API router

steven-zou:
**Add API routes of chart repo services in harbor API router**

- add chart repo API controller
- update GetChartVersion to extract repo name from the request
- update the corresponding UI cases
- update the chart controller initialization process

### DIFF
--- a/src/ui/api/chart_repository.go
+++ b/src/ui/api/chart_repository.go
@@ -1,0 +1,260 @@
+package api
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+
+	"github.com/vmware/harbor/src/chartserver"
+	hlog "github.com/vmware/harbor/src/common/utils/log"
+)
+
+const (
+	backendChartServerAddr = "BACKEND_CHART_SERVER"
+	namespaceParam         = ":repo"
+
+	accessLevelPublic = iota
+	accessLevelRead
+	accessLevelWrite
+	accessLevelAll
+	accessLevelSystem
+)
+
+//chartController is a singleton instance
+var chartController = initializeChartController()
+
+//ChartRepositoryAPI provides related API handlers for the chart repository APIs
+type ChartRepositoryAPI struct {
+	//The base controller to provide common utilities
+	BaseController
+
+	//Keep the namespace if existing
+	namespace string
+}
+
+//Prepare something for the following actions
+func (cra *ChartRepositoryAPI) Prepare() {
+	//Call super prepare method
+	cra.BaseController.Prepare()
+
+	//Try to extract namespace for parameter of path
+	//It may not exist
+	cra.namespace = strings.TrimSpace(cra.GetStringFromPath(namespaceParam))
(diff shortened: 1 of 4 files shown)


src/ui/router.go
```diff
@@ -122,6 +122,21 @@ func initRouters() {
 
 	beego.Router("/registryproxy/*", &controllers.RegistryProxy{}, "*:Handle")
 
+	//APIs for chart repository
+	chartRepositoryAPIType := &api.ChartRepositoryAPI{}
+	beego.Router("/api/chartserver/health", chartRepositoryAPIType, "get:GetHealthStatus")
+	beego.Router("/api/:repo/charts", chartRepositoryAPIType, "get:ListCharts")
```
reasonerjt: Suggest we add a prefix to all the chart related APIs, for example `/api/chart_registry/:repo/charts`

src/ui/api/chart_repository.go
```diff
@@ -0,0 +1,240 @@
...
+		if !cra.requireNamespace(cra.namespace) {
+			return
+		}
+	}
+}
+
+//GetHealthStatus handles GET /api/chartserver/health
+func (cra *ChartRepositoryAPI) GetHealthStatus() {
+	//Check access
+	cra.requireAccess(cra.namespace, accessLevelSystem)
```
reasonerjt: should not move on if it returns false.

src/ui/router.go
```diff
@@ -122,6 +122,21 @@ func initRouters() {
...
+	beego.Router("/api/:repo/charts/:name", chartRepositoryAPIType, "get:ListChartVersions")
+	beego.Router("/api/:repo/charts/:name/:version", chartRepositoryAPIType, "get:GetChartVersion")
+	beego.Router("/api/:repo/charts/:name/:version", chartRepositoryAPIType, "delete:DeleteChartVersion")
+	beego.Router("/api/:repo/charts", chartRepositoryAPIType, "post:UploadChartVersion")
+	beego.Router("/api/:repo/prov", chartRepositoryAPIType, "post:UploadChartProvFile")
+
+	//Repository services
+	beego.Router("/:repo/index.yaml", chartRepositoryAPIType, "get:GetIndexByRepo")
```
reasonerjt: Same concern here, shall we add a prefix to isolate all requests for the charts?
I understand there maybe limitation as the request is from `helm` command. 
Maybe we can ask user to set `context` to `helm` command?
Let's open a follow-up issue for this concern.

src/ui/router.go
steven-zou: Maybe we can enable context

steven-zou: Will find a proper way and submit code in separate PR

src/ui/api/chart_repository.go
steven-zou: Will be fixed

src/ui/router.go
steven-zou: Will find a proper way and submit code in separate PR

PR #5311, opened 2018-07-13
Add SafeCast function and set default value for some sytem configure

stonezdj:
Add SafeCastString, SafeCastInt, SafeCastFloat64, SafeCastBool function to check
the type matched and avoid panic in runtime

Add default value to configure settings to avoid cannot save configure
issue


reasonerjt:
Seems only need to make sure there's no "missed key" when the very first time the cfg store is initialized when the adminserver is started.

src/adminserver/systemcfg/systemcfg.go
```diff
@@ -248,6 +249,7 @@ func Init() (err error) {
 	if err != nil {
 		return err
 	}
+	utils.AddMissedKey(cfgs)
```
reasonerjt: Is this necessary?

PR #5313, opened 2018-07-16
Add admin job api, it supports Harbor admin to trigger job

wy65701436:
either manual or schedule. The job will be populated to job
service to execute. The api includes 1. POST /api/admin/jobs
2, GET /api/admin/jobs/?name 3, GET /api/admin/jobs/:id/log.


src/common/dao/admin_job.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// GetAdminJob ...
+func GetAdminJob(id int64) (*models.AdminJob, error) {
+	o := GetOrmer()
+	aj := models.AdminJob{ID: id}
+	err := o.Read(&aj)
+	if err == orm.ErrNoRows {
+		return nil, nil
+	}
```
ywk253100: else?

src/ui/api/admin_job.go
```diff
@@ -0,0 +1,178 @@
...
+		return
+	}
+
+}
+
+//Post ...
+func (aj *AdminJobAPI) Post() {
+
+	ajr := models.AdminJobReq{}
+	aj.DecodeJSONReq(&ajr)
```
ywk253100: Call DecodeJSONReqAndValidate(), it will decode and validate the request data.

src/ui/api/admin_job.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// submitJob submits a job to job service per request
+func (aj *AdminJobAPI) submitJob(ajr models.AdminJobReq) {
+	id, err := dao.AddAdminJob(&common_models.AdminJob{
+		Name: ajr.Name,
+		Kind: ajr.Kind,
+	})
+	if err != nil {
+		aj.HandleInternalServerError(fmt.Sprintf("%v", err))
```
ywk253100: Should we update the job status in the database if got error?

ywk253100: Should consider the scenario that users submit the GC job more than once, and how can we determine the status of the GC process?

wy65701436:
Close it as API change, and use https://github.com/vmware/harbor/pull/5344 to track.

PR #5319, opened 2018-07-16
Remove lost+found folder in database volume on initial start

BowlingX:
Postgres was complaning about a `lost+found` folder in the data folder. This happens because the volume is directly mounted into the directory `/var/lib/postgresql/data`. This pull request will remove the folder if it exists before the container starts.

Another solution could be to create a subfolder and set the `PGDATA` env to this subfolder, but this will break current deployments.

### DIFF
--- a/contrib/helm/harbor/templates/database/database-ss.yaml
+++ b/contrib/helm/harbor/templates/database/database-ss.yaml
@@ -19,6 +19,13 @@ spec:
 {{ include "harbor.labels" . | indent 8 }}
         component: database
     spec:
+      initContainers:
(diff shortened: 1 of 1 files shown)


timgclark:
@BowlingX I'm definitely running into this same issue, and I was going to submit the subPath ("subfolder") fix when I found this PR. Has there actually been a release with a helm chart deploying postgresql? Is backward compatibility with subPath a concern at this point?

BowlingX:
I just though people might use this chart already, then it would not break anything.

ywk253100:
@BowlingX Thanks for your contribution.

PR #5335, opened 2018-07-18
Support the configuration of registry storage in helm chart

ywk253100:
This commit adds the registry storage configuration to the values.yaml. By this way, users can cusomize the storage they use.

### DIFF
--- a/contrib/helm/harbor/values.yaml
+++ b/contrib/helm/harbor/values.yaml
@@ -189,22 +189,68 @@ registry:
     pullPolicy: IfNotPresent
   httpSecret: not-a-secure-secret
   logLevel: info
-# comment out one of the below to use your cloud's object storage.
-#  objectStorage:
-#    gcs:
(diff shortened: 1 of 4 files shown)


contrib/helm/harbor/values.yaml
```diff
@@ -189,22 +189,68 @@ registry:
...
-#      bucket: ""
-#      encrypt: "true"
-#    azure:
-#      accountname: ""
-#      accountkey: ""
-#      container: ""
+  storage:
+    # specify the type of storage: "filesystem", "azure", "gcs", "s3", "swift", 
+    # "oss" and fill the information needed in the corresponding section
+    type: filesystem
```
reasonerjt: Personally I prefer a more self- explanatory approach, i.e. get rid of `type` and leave only one chunk of storage setting.  
Given the current approach seems also a common practice, I'll approve but let's make sure we keep this consistent

PR #5347, opened 2018-07-19
Remove vmware/photon image from offline package

reasonerjt:
We used to package vmware/photon:1.0 in the offline package for product
integration.  Currently this is no longer needed by downstream product.
This commit removes it.


reasonerjt:
@wy65701436 confirmed there's no impact to vic-product.

PR #5383, opened 2018-07-24
Notification should not hide content according to ux

zhoumeina:
Issue description:
Notification will hide some content when it popup.

According to ux requirement, we should make it not extend to the whole widh of UI, and should not hide the content.

Fix:
Modify the position from 'absolute' to 'relative', so that it will not hide the content.


ninjadq:
LGTM

PR #5386, opened 2018-07-24
Use current time as regsitration time for LDAP and UAA users

stonezdj:
Previous user registration time always displays "1/1/1". change to first login time.


reasonerjt:
This should be resolved at DB level I believe?
Such as
https://github.com/vmware/harbor/blob/master/make/migrations/postgresql/0001_initial_schema.up.sql#L80

reasonerjt:
Per discussion, closing this pr.

Issue #5349, opened 2018-07-19
documentation on user_guide.md needs some editing


lweitzman:
the user_guide.md needs the following edits
https://github.com/vmware/harbor/blob/master/docs/user_guide.md

There are two views to show repositories, listview, and cardview, you can switch between them by click correspond icon.
should read “…listview and card view. You can switch between them by clicking the corresponding icon.”

choose one role to batch switch checked members’s 
should read “… checked members’ “

The function is project-oriented, and once the system administrator set a rule to one project,
should read “…system administrator has set a rule in one project,”

If the project does not exist on the remote registry, a new project will be created automatically, but if it already exists and the user configured in policy has no write privilege to it, the process will fail. The member information will not be replicated.
break into 2 sentences. “…created automatically. If it already exists…”

There may be a bit of delay during replication according to the situation of the network...will keep trying until the network issue resolved.
should read “… during replication based on the situation of the network… until the network issue is  resolved"

The replication feature is incompatible between Harbor instance before version 0.3.5(included) and after version 0.3.5.
should read “… between Harbor versions before version 0.3.5 (included) and those versions after 0.3.5”

Screenshots used under 
"Administrator options" -> "Managing authentication” 
"Administrator options" -> "Managing project creation” 
"Administrator options" -> "Managing self-registration” 
are using the old left navigation

ywk253100:
Done

PR #5390, opened 2018-07-25
Update the user guide

ywk253100:
This commit updates the user guide according to the comments from Louis


ywk253100:
Fix #5349.

PR #5430, opened 2018-07-30
Fix pgsql creation column bug

wy65701436:
Root cause: Use default 'now'::timestamp will not generate timestamp for each transaction,
PG will convert now to a timestamp as soon as the constant is parsed. To fix it, update it
to defult CURRENT_TIMESTAMP, thie setting is the same as default now(), which returns the
start time of current transaction because ther are fuction calls, hey will give the desired
behavior of defaulting to the time of row insertion.

Reference: https://www.postgresql.org/docs/9.6/static/functions-datetime.html#FUNCTIONS-DATETIME-CURRENT
PG version: 9.6.9

### DIFF
--- a/make/migrations/postgresql/0001_initial_schema.up.sql
+++ b/make/migrations/postgresql/0001_initial_schema.up.sql
@@ -39,11 +39,22 @@ create table harbor_user (
  reset_uuid varchar(40) DEFAULT NULL,
  salt varchar(40) DEFAULT NULL,
  sysadmin_flag boolean DEFAULT false NOT NULL,
- creation_time timestamp(0),
- update_time timestamp(0),
+ creation_time timestamp default CURRENT_TIMESTAMP,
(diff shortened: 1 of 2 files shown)


make/migrations/postgresql/0001_initial_schema.up.sql
```diff
@@ -39,8 +39,8 @@ create table harbor_user (
  reset_uuid varchar(40) DEFAULT NULL,
  salt varchar(40) DEFAULT NULL,
  sysadmin_flag boolean DEFAULT false NOT NULL,
- creation_time timestamp(0),
- update_time timestamp(0),
+ creation_time timestamp(0) default CURRENT_TIMESTAMP,
```
reasonerjt: why is it `timestamp(0)` in contrast to `timestamp`?

reasonerjt:
Need triggers for update_time of `harbor_user` and `project`

Issue #5143, opened 2018-06-13
500 internal error when 'project' is not define during docker pull


aladjadj:
Harbor return a HTTP status 500 Internal Server error when we pull an image without 'project' like library
example : myregistry.domain.com/issue_test:latest

Normaly the default project is 'library', the same comportment to tag 'latest'

docker: 17.12.0-ce
docker-compose: 1.21.2
harbor: 1.4

Logs:
user@harbor:/var/log/harbor$ sudo tail -f *.log | grep 'issue_test'
Jun 13 15:41:02 172.20.0.1 ui[30781]: 2018-06-13T13:41:02Z [DEBUG] [token.go:35]: URL for token request: /service/token?account=email%40company.com&scope=repository%3Aissue_test%3Apull&service=harbor-registry
Jun 13 15:41:02 172.20.0.1 ui[30781]: 2018-06-13T13:41:02Z [DEBUG] [creator.go:200]: scopes: [repository:issue_test:pull]
Jun 13 15:41:02 172.20.0.1 ui[30781]: 2018-06-13T13:41:02Z [DEBUG] [authutils.go:47]: scopes: [repository:issue_test:pull]
Jun 13 15:41:02 172.20.0.1 ui[30781]: 2018-06-13T13:41:02Z [DEBUG] [authutils.go:94]: user: email@company.com, access: &{repository  issue_test [pull]}
Jun 13 15:41:02 172.20.0.1 ui[30781]: 2018-06-13T13:41:02Z [ERROR] [token.go:48]: Unexpected error when creating the token, error: Unable to parse image from string: issue_test
Jun 13 15:41:02 172.20.0.1 proxy[30781]: 10.3.50.17 - "GET /service/token?account=email%40company.com&scope=repository%3Aissue_test%3Apull&service=harbor-registry HTTP/1.1" 500 0 "-" "docker/18.03.0-ce go/go1.9.4 git-commit/0520e24 kernel/4.13.0-43-generic os/linux arch/amd64 UpstreamClient(Docker-Client/18.03.0-ce \x5C(linux\x5C))" 0.017 0.017 .

reasonerjt:
An improvement is to return `401` rather than `500` in this case.  But I don't think we'll add `library` for your command, because it's added by docker client.

aladjadj:
I'm not sure than 'library' added by docker client because in another docker registry we have image on top of  path ( without library ), and logs proved inverse.

reasonerjt:
@aladjadj 
[...]

PR #5448, opened 2018-07-31
add default namespace library if not specified

aladjadj:
PR refer to issue #5143

### DIFF
--- a/src/ui/service/token/creator.go
+++ b/src/ui/service/token/creator.go
@@ -99,9 +99,6 @@ type endpointParser struct {
 
 func (e endpointParser) parse(s string) (*image, error) {
 	repo := strings.SplitN(s, "/", 2)
-	if len(repo) < 2 {
-		return nil, fmt.Errorf("Unable to parse image from string: %s", s)
-	}
 	if repo[0] != e.endpoint {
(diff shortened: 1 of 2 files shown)


reasonerjt:
This change seems only add a namespace `library` in permission checking?
So if a user pushes to an image like `$registry_host/$repo:$tag`  will require the permission of library, but it won't actually be pushed to the `/library/$repo` in the storage.

This seems a workaround to me.

aladjadj:
yes you right,
the solution maybe to fix the error 500 into harbor when pull image without "project"
and redirect root path to "library" in router ( nginx ) to match with Harbor concept 

what do you think ?

jessehu:
This feature is great!

reasonerjt:
Sorry I don't think it's the right thing to do, because in the registry it will still be stored in the storage of 
/$repo not /library/$repo.  

The code change in permission checking will introduce confusing inconsistency.  You'll at least also have to deal with the case when calling `GET v2/_catalog` and add the `library` to the images that does not have  namespace.  And it may become tricky when triggering Clair to scan the image(it may not, we need to check the code).

The benefit of this change is just a shortcut but the inconsistency it introduced may cause problems in the future.

PR #5450, opened 2018-08-01
Update job status when failed to submit it

ywk253100:
Set the status of replication job to "error" when the submitting to jobservice fails


src/replication/replicator/replicator.go
```diff
@@ -119,6 +119,9 @@ func (d *DefaultReplicator) Replicate(replication *Replication) error {
 
 			uuid, err := d.client.SubmitJob(job)
 			if err != nil {
+				if er := dao.UpdateRepJobStatus(id, common_models.JobError); er != nil {
+					log.Errorf("failed to update the status of job %d: %s", id, er)
```
wy65701436: failed to update the status of job => failed to update the status of the replication job

PR #5456, opened 2018-08-01
Fix access level can not sort

zhoumeina:
Issue description:
In project list page. User click sort according access level, it does not work.

Fix:

Remodify the compare method.


src/ui_ng/src/app/shared/shared.utils.ts
```diff
@@ -143,15 +143,34 @@ export class CustomComparator<T> implements Comparator<T> {
     compare(a: { [key: string]: any | any[] }, b: { [key: string]: any | any[] }) {
         let comp = 0;
         if (a && b) {
-            let fieldA = a[this.fieldName];
-            let fieldB = b[this.fieldName];
+            let fieldA, fieldB;
+            let propKeys = Object.keys(a);
+            for (let key of propKeys) {
```
ninjadq: why not iterate `Objects.keys(a)` directly.

PR #5520, opened 2018-08-03
Fix LDAP user group related issues

stonezdj:
User groups should not have same DN
Should not import an LDAP group if there is a user group with same DN

### DIFF
--- a/src/ui/api/projectmember.go
+++ b/src/ui/api/projectmember.go
@@ -15,9 +15,11 @@
 package api
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"strconv"
+	"strings"
 
 	"github.com/vmware/harbor/src/common"
 	"github.com/vmware/harbor/src/common/dao/project"
@@ -34,6 +36,12 @@ type ProjectMemberAPI struct {
 	entityType string
 	project    *models.Project
 }
+
+// ErrDuplicateProjectMember ...
+var ErrDuplicateProjectMember = errors.New("The project member specified already exist")
+
+// ErrInvalidRole ...
+var ErrInvalidRole = errors.New("Failed to update project member, role is not in 1,2,3")
 
 // Prepare validates the URL and parms
 func (pma *ProjectMemberAPI) Prepare() {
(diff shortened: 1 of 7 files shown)


src/ui/api/projectmember.go
```diff
@@ -121,11 +135,28 @@ func (pma *ProjectMemberAPI) Post() {
 	projectID := pma.project.ProjectID
 	var request models.MemberReq
 	pma.DecodeJSONReq(&request)
-	pmid, err := AddOrUpdateProjectMember(projectID, request)
+	request.MemberGroup.LdapGroupDN = strings.TrimSpace(request.MemberGroup.LdapGroupDN)
+	//Check valid DN specified
+	if len(request.MemberGroup.LdapGroupDN) > 0 {
+		if _, err := goldap.ParseDN(request.MemberGroup.LdapGroupDN); err != nil {
+			pma.HandleBadRequest("Invalid LDAP DN specified.")
```
reasonerjt: can we log the error?

src/ui/api/projectmember.go
```diff
@@ -179,18 +210,27 @@ func AddOrUpdateProjectMember(projectID int64, request models.MemberReq) (int, e
...
+		//Check duplicate LDAP DN in usergroup, if usergroup exist, return error
+		userGroupList, err := group.QueryUserGroup(models.UserGroup{LdapGroupDN: request.MemberGroup.LdapGroupDN})
+		if err != nil {
+			return 0, err
+		}
+		if len(userGroupList) > 0 {
+			return 0, ErrDuplicateLDAPGroup
+		}
+		//If groupname provided, use the provided groupname to name this group
 		groupID, err := auth.SearchAndOnBoardGroup(request.MemberGroup.LdapGroupDN, request.MemberGroup.GroupName)
```
reasonerjt: I think we should fail in `SearchAndOnBoard` or let SearchAndOnBoard in ldap package ensure there's no duplicate entry?

src/ui/api/projectmember.go
stonezdj: pma.HandleBadRequest will log the error

src/ui/api/projectmember.go
stonezdj: OK

PR #5522, opened 2018-08-03
Fix go gas issue for chart

wy65701436:
Add the error handling to fix go gas issue "Error unhandled"

### DIFF
--- a/src/chartserver/utils.go
+++ b/src/chartserver/utils.go
@@ -18,8 +18,10 @@ const (
 func WriteError(w http.ResponseWriter, code int, err error) {
 	errorObj := make(map[string]string)
 	errorObj["error"] = err.Error()
-	errorContent, _ := json.Marshal(errorObj)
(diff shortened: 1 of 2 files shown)


src/chartserver/utils.go
```diff
@@ -18,8 +18,11 @@ const (
 func WriteError(w http.ResponseWriter, code int, err error) {
 	errorObj := make(map[string]string)
 	errorObj["error"] = err.Error()
-	errorContent, _ := json.Marshal(errorObj)
-
+	errorContent, errorMarshal := json.Marshal(errorObj)
+	if errorMarshal != nil {
+		w.WriteHeader(code)
+		w.Write([]byte("Error occured while marshal error content:%v" + errorMarshal.Error()))
```
steven-zou: how about

if errorMarshal != nil {
  errorContent=err.Error()
}

PR #5524, opened 2018-08-03
Enhance the project deletable checking logic to include helm charts checking

steven-zou:
**Enhance the project deletable checking logic to include helm charts checking**

- add utility handler in chart controller to support getting charts by ns
- add ut case for utility handler
- update deletable checking logic in project controller

### DIFF
--- a/src/ui/api/project.go
+++ b/src/ui/api/project.go
@@ -217,7 +217,7 @@ func (p *ProjectAPI) Delete() {
 		return
 	}
 
-	result, err := deletable(p.project.ProjectID)
+	result, err := p.deletable(p.project.ProjectID)
 	if err != nil {
 		p.HandleInternalServerError(fmt.Sprintf(
 			"failed to check the deletable of project %d: %v", p.project.ProjectID, err))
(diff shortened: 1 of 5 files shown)


src/ui/api/project.go
```diff
@@ -296,6 +296,19 @@ func deletable(projectID int64) (*deletableResp, error) {
 		}, nil
 	}
 
+	//Check helm charts number
```
wy65701436: needs to check with chart museum.

src/ui/api/project.go
```diff
@@ -296,6 +296,21 @@ func deletable(projectID int64) (*deletableResp, error) {
...
+	if config.WithChartMuseum() {
+		charts, err := chartController.GetUtilityHandler().GetChartsByNs(p.project.Name)
+		if err != nil {
+			return nil, err
+		}
+
+		if len(charts) > 0 {
+			return &deletableResp{
+				Deletable: false,
+				Message:   "the project contains helm charts, can not be deleled",
```
wy65701436: deleled -> deleted

PR #5545, opened 2018-08-07
Upgrade docker and docker-compose version in requirement

wy65701436:


README.md
```diff
@@ -26,7 +26,7 @@ Project Harbor is an an open source trusted cloud native registry project that s
 
 **System requirements:**
 
-**On a Linux host:** docker 1.10.0+ and docker-compose 1.6.0+ .
+**On a Linux host:** docker 17.12.1+ and docker-compose 1.10.0+ .
```
reasonerjt: 17.03+  should be fine

LoLei:
When will this new Docker version requirement be in effect in a release? The next one already?
Because I installed 1.5.2 with Docker version 1.10.3 still.

PR #5550, opened 2018-08-07
Unify redis configuration for harbor components

wy65701436:
* Unify redis configuration for harbor components

### DIFF
--- a/make/common/templates/ui/env
+++ b/make/common/templates/ui/env
@@ -5,6 +5,6 @@ JOBSERVICE_SECRET=$jobservice_secret
 GODEBUG=netdns=cgo
 ADMINSERVER_URL=$adminserver_url
 UAA_CA_ROOT=/etc/ui/certificates/uaa_ca.pem
-_REDIS_URL=$redis_url
+_REDIS_URL=$redis_host:$redis_port,100,$redis_password
 SYNC_REGISTRY=false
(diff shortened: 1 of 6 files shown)


make/common/templates/ui/env
```diff
@@ -5,6 +5,6 @@ JOBSERVICE_SECRET=$jobservice_secret
 GODEBUG=netdns=cgo
 ADMINSERVER_URL=$adminserver_url
 UAA_CA_ROOT=/etc/ui/certificates/uaa_ca.pem
-_REDIS_URL=$redis_url
+_REDIS_URL=$redis_host:$redis_port,100,$redis_password
```
steven-zou: The db_index is set to default to 0?

make/prepare
```diff
@@ -80,6 +80,10 @@ def validate(conf, args):
         if storage_provider_config == "":
             raise Exception("Error: no provider configurations are provided for provider %s" % storage_provider_name)
 
+    redis_db_index = rcp.get("configuration", "redis_db_index").strip()
```
steven-zou: we should add Redis host and port checking to make sure the redis config is conrrect because redis is a must-have component for js so far.

make/common/templates/ui/env
wy65701436: cannot set index value for beego, 0 is the hard coded.

PR #5552, opened 2018-08-08
provide manual installation documentation for sdk/harbor-py

Cplo:
The version of harbor-py installed from PiPY is old, We can provide documentation installed from source code.


reasonerjt:
@Cplo sorry, but we are considering to move harbor-py out of the repo.

PR #5564, opened 2018-08-09
Unify redis configuration for harbor components

wy65701436:
this commit is to separate the redis_url into the host, port, pwd, and index for different components, and make it possible to set external Redis server.

### DIFF
--- a/make/prepare
+++ b/make/prepare
@@ -21,7 +21,8 @@ if sys.version_info[:3][0] == 3:
 
 DATA_VOL = "/data"
 
-def validate(conf, args): 
+def validate(conf, args):
+
     if args.ha_mode:
         db_host = rcp.get("configuration", "db_host")
         if db_host == "mysql":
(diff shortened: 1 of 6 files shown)


make/prepare
```diff
@@ -80,6 +78,18 @@ def validate(conf, args):
         if storage_provider_config == "":
             raise Exception("Error: no provider configurations are provided for provider %s" % storage_provider_name)
 
+    redis_host = rcp.get("configuration", "redis_host")
+    if redis_host is None or len(redis_host) < 1:
+        raise Exception("Error: redis_host in harbor.cfg needs to point to a Redis cluster.")
```
steven-zou: "... to an endpoint of Redis server or cluster"

make/prepare
```diff
@@ -282,10 +292,22 @@ secret_key = get_secret_key(secretkey_path)
...
+redis_db_index_js = db_indexs[1]
+redis_db_index_chart = db_indexs[2]
+
+#redis://[arbitrary_username:password@]ipaddress:port/database_index
+redis_url_js = ''
+if len(redis_password) > 0:
+    redis_url_js = "redis://anonymous:%s@%s:%s/%s" % (redis_password, redis_host, redis_port, redis_db_index_js)
 else:
-    redis_url = ""
+    redis_url_js = "redis://%s:%s/%s" % (redis_host, redis_port, redis_db_index_js)
```
steven-zou: How about:
```
redis_url_js = "%s:%s/%s" % (redis_host, redis_port, redis_db_index_js)
if len(redis_password) > 0:
  redis_url_js = "anonymous:%s@%s" % (redis_password, redis_url_js)

redis_url_js = "redis://%s" % redis_url_js
```

make/prepare
```diff
@@ -670,22 +700,9 @@ if args.chart_mode:
 
     # process redis info
     cache_store = ""
```
steven-zou: cache_store is missed

PR #5581, opened 2018-08-10
Update README.md

clouderati:
Updating readme to point to `gh/goharbor/harbor` location, and removing references to VMware's CLA.


clouderati:
@steven-zou can you PTAL? Thanks!

Issue #5410, opened 2018-07-27
Support to configure the storage of chartmuseum in Harbor chart


ywk253100:

ywk253100:
Done.

PR #5582, opened 2018-08-12
Allow to define own env vars to the chartmuseum

jkroepke:
This is may required, if an admin wants to use S3 or Google Cloud Storage.

I got this idea from here: https://github.com/helm/charts/tree/master/stable/chartmuseum

### DIFF
--- a/contrib/helm/harbor/templates/chartmuseum/chartmuseum-cm.yaml
+++ b/contrib/helm/harbor/templates/chartmuseum/chartmuseum-cm.yaml
@@ -12,7 +12,6 @@ data:
   CACHE_REDIS_DB: "{{ template "harbor.redis.databaseIndex" }}"
   BASIC_AUTH_USER: "chart_controller"
   DEPTH: "1"
-  STORAGE: "local"
   STORAGE_LOCAL_ROOTDIR: "/chart_storage"
   DEBUG: "false"
   LOG_JSON: "true"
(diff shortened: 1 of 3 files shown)


jkroepke:
I would like to see it in 1.6.0, because we want to use our existing harbor registry for our helm chart but s3 is a requirement.

ywk253100:
@jkroepke Thank you for your contribution.

But we have the storage configuration in the image registry already https://github.com/goharbor/harbor/blob/master/contrib/helm/harbor/values.yaml#L141, we want to reuse this part for chart museum either.

Here is an issue #5410 to track this, but this isn't in the 1.6 scope.

jkroepke:
So you want to use the variables from the Docker Registry for the Chartmuseum, too?

Or you want to use the variable layout?

An other case whould be somethink like this:

### Amazon

```
chartmuseum:
  storage: amazon
  options:
    bucket:
    endpoint:
    prefix:
    region:
    sse:
```

Result into:

AMAZON_BUCKET
AMAZON_ENDPOINT
AMAZON_PREFIX
AMAZON_REGION

### Google

```
chartmuseum:
  storage: google
  options:
    bucket:
    prefix:
```

Result into:

GOOGLE_BUCKET
GOOGLE_PREFIX

### Filesystem

```
chartmuseum:
  storage: filesystem
  options:
    storageClass:
    size:
```

Result into a volumeClaimTemplates inside the StatefulSet

jkroepke:
IHMO you should prioritize object storage like in the same way you do it with filesystem. 

You are a member of the CNCF now IMHO you should prefer cloud native.

ywk253100:
We want to separate the storage part from registry in values.yaml, it would be something like this:
```
registry:
  image: xxx
  ...
chartmuseum:
  image: xxx
  ...
storage:
  type: filesystem
  filesystem:
    ...
  s3:
    ...
  gcs:
    ...
```

We have no enough resource to do this for now. You can contribute this part if you like, we'll appreciate it very much.

ywk253100:
Closing the pull request but the contribution is welcomed.

Issue #5546, opened 2018-08-07
add helm repo part to the user guide document


steven-zou:
As we support helm repo server, a guideline about the helm repo should be added to the user guide document

reasonerjt:
PR pending for review, move to SP 41

steven-zou:
Work is already done, close it.

PR #5590, opened 2018-08-13
Append chart repository server related guideline into the user guide document

steven-zou:
**Append chart repository server related guideline into the user guide document**


steven-zou:
#5546

PR #5613, opened 2018-08-14
Fix issue ldap_group_admin_dn doesn't take effect

stonezdj:


src/ui/api/user.go
```diff
@@ -116,6 +116,7 @@ func (ua *UserAPI) Get() {
 			ua.CustomAbort(http.StatusInternalServerError, "Internal error.")
 		}
 		u.Password = ""
+		u.HasAdminRole = ua.SecurityCtx.IsSysAdmin()
```
reasonerjt: This is incorrect, it does not cover the case when the user id of the api is not current user's

reasonerjt:
And you should sign off using your vmware Email

src/ui/api/user.go
stonezdj: fixed it

PR #5627, opened 2018-08-15
API:api/projects?name=xxx is fuzzy queries,should be an exact query

wutongjie23hao:
API:api/projects?name=xxx is fuzzy queries,should be an exact query if name is provided.

### DIFF
--- a/src/ui/api/project.go
+++ b/src/ui/api/project.go
@@ -371,11 +371,22 @@ func (p *ProjectAPI) List() {
 			}
 		}
 		//Query projects by user group
+		dstProject, err := p.ProjectMgr.Get(query.Name)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


reasonerjt:
Sorry but it's the current design.

e-nikolov:
What if there was an extra parameter to specify that an exact match is required? Would that be acceptable?

wutongjie23hao:
The featureof this design: int field of json query string is exact search, but string field of json query string is fuzzy search.

reasonerjt:
@e-nikolov yes, another parm to specify it's exact match would be acceptable.

As for the project ID you can get it via the `Location` header of the response.

clouderati:
Hi @e-nikolov – please let us know if you're willing and able to contribute this change. Thanks!

PR #5644, opened 2018-08-16
Update registry image to enable inject root cert (1.6 cherry pick)

reasonerjt:
In some user's environment, there's local object storage hosted with
self-signed certificate.
Because registry process runs in a photon container, it has to trust
the certificate in the photon level such that the registry can access
the storage service.

This commit updates the registry image to append custom cert to the root
bundle when the container is started.  And make the customer cert
configurable in `harbor.cfg`

### DIFF
--- a/make/photon/registry/entrypoint.sh
+++ b/make/photon/registry/entrypoint.sh
@@ -17,6 +17,21 @@ if [ -d /storage ]; then
     fi
 fi
 
+if [ ! -f /etc/pki/tls/certs/ca-bundle.crt.original ]; then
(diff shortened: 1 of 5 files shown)


make/photon/registry/entrypoint.sh
```diff
@@ -17,6 +17,17 @@ if [ -d /storage ]; then
     fi
 fi
 
+if [ ! -f /etc/pki/tls/certs/ca-bundle.crt.original ]; then
```
wy65701436: it's Photon OS only, should add checking.

PR #5663, opened 2018-08-19
Cherry pick ui bug fix

ninjadq:
Fix Group related issues
Edit the wording of version
Fix chart upload related issues
Fix incorrect UI views when logging in with a user in the group assigned with the proper role
Diable delete button when user not admin
Remove Upload Button on Chart Version List
Fix i18n translate issue

### DIFF
--- a/src/ui_ng/lib/src/helm-chart/helm-chart.component.ts
+++ b/src/ui_ng/lib/src/helm-chart/helm-chart.component.ts
@@ -17,6 +17,7 @@ import { ErrorHandler } from "../error-handler/error-handler";
 import { toPromise, DEFAULT_PAGE_SIZE } from "../utils";
 import { HelmChartService } from "../service/helm-chart.service";
 import { DefaultHelmIcon} from "../shared/shared.const";
+import { Roles } from './../shared/shared.const';
 
 @Component({
   selector: "hbr-helm-chart",
@@ -30,7 +31,7 @@ export class HelmChartComponent implements OnInit {
   @Input() projectName = "unknown";
   @Input() urlPrefix: string;
   @Input() hasSignedIn: boolean;
-  @Input() projectRoleID = 0;
+  @Input() projectRoleID = Roles.OTHER;
   @Input() hasProjectAdminRole: boolean;
   @Output() chartClickEvt = new EventEmitter<any>();
   @Output() chartDownloadEve = new EventEmitter<string>();
@@ -74,8 +75,7 @@ export class HelmChartComponent implements OnInit {
   }
 
   public get developerRoleOrAbove(): boolean {
-    // 1: admin, 2: developer, 3: guest
-    return this.projectRoleID === 2 || this.hasProjectAdminRole;
+    return this.projectRoleID === Roles.DEVELOPER || this.hasProjectAdminRole;
   }
 
   ngOnInit(): void {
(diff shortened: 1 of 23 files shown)


src/ui_ng/lib/src/helm-chart/helm-chart.component.ts
```diff
@@ -106,10 +117,22 @@ export class HelmChartComponent implements OnInit {
...
+    this.uploadForm.reset();
+  }
+
   onChartUpload() {
+    this.resetUploadForm();
     this.isUploadModalOpen = true;
   }
 
+  cancelUpload() {
+    this.resetUploadForm();
```
steven-zou: Do we really need to reset uploading forms two times?

src/ui_ng/lib/src/helm-chart/versions/helm-chart-version.component.html
```diff
@@ -104,13 +103,16 @@
...
                     </div>
                 </div>
                 <div class="card-footer">
                     <clr-dropdown [clrCloseMenuOnItemClick]="false">
-                        <button type="button" class="btn btn-link" (click)="versionDownload($event, item)">{{'HELM_CHART.DOWNLOAD' | translate}}</button>
-                        <button type="button" class="btn btn-link" (click)="deleteVersionCard($event, item)">{{'BUTTON.DELETE' | translate}}</button>
+                        <button type="button" class="btn btn-link"
+                        (click)="versionDownload($event, item)">{{'HELM_CHART.DOWNLOAD' | translate}}</button>
+                        <button type="button" class="btn btn-link"
+                        [disabled]="selectedRows.length<=0 || !hasProjectAdminRole"
```
steven-zou: Suggest extracting these logics to a property.

src/ui_ng/src/app/project/list-charts/list-charts.component.ts
```diff
@@ -35,6 +36,8 @@ export class ListChartsComponent implements OnInit {
       let project = <Project>(resolverData["projectResolver"]);
       this.projectName = project.name;
       this.hasProjectAdminRole = project.has_project_admin_role;
+      this.project_member_role_id = project.current_user_role_id;
+      console.log(this.project_member_role_id);
```
steven-zou: Suggest removing log logic or add debug switch

src/ui_ng/lib/src/helm-chart/helm-chart.component.ts
```diff
@@ -68,6 +74,10 @@ export class HelmChartComponent implements OnInit {
     return this.systemInfo ? this.systemInfo.registry_url : "";
   }
 
+  public get developerRoleOrAbove(): boolean {
+    return this.projectRoleID === Roles.DEVELOPER || this.hasProjectAdminRole;
```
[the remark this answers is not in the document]
ninjadq: Yes, it already have considerate the system admin situation. The logic of assigning project admin role is written by you 😓

src/ui_ng/lib/src/helm-chart/helm-chart.component.ts
ninjadq: No, because only once can assure the whole system works fine.
Yes, because we don't know how the user operation our system and we only can do is  to assure that every operation will work.

PR #5693, opened 2018-08-22
Remove the Harbor chart directory

ywk253100:
As we moving the Harbor chart into a separate repository, this commit removes the chart directory from goharbor/harbor repository.


reasonerjt:
Could you update the `README.md` to reference to that repo?

Issue #5694, opened 2018-08-22
Volume capacity incorrect when run in MacOS


cd1989:
**Version**
- OS: MacOS 10.13.4
- docker: 18.03.0-ce, build 0520e24
- docker-compose: version 1.20.1, build 5d8c71b

![image](https://user-images.githubusercontent.com/16573984/44454923-5a180f00-a62f-11e8-8e2b-db329d606c86.png)

```
$ df -h
Filesystem      Size   Used  Avail Capacity iused               ifree %iused  Mounted on
/dev/disk1s1   234Gi   81Gi  147Gi    36% 1460577 9223372036853315230    0%   /
...
```

As show above, when run Harbor in MacOS, the volume capacity is incorrect, actually it's 1024 times the actual value.

For the reason, we can refer to: https://github.com/docker/for-mac/issues/2136

PR #5695, opened 2018-08-22
Fix volume capacity when run in MacOS

cd1989:
When run Harbor in MacOS, volume capacity is incorrect due to the use of `statfs`.

Refer to: https://github.com/docker/for-mac/issues/2136

This PR fix: https://github.com/goharbor/harbor/issues/5694


clouderati:
lgtm – @reasonerjt please review

cd1989:
@ywk253100 Sure.

PR #5709, opened 2018-08-23
Add github flavered table in markdown

ninjadq:
Add table style in markdown and
Fix deleted all version may cause 404


ninjadq:
image

PR #5731, opened 2018-08-27
Fix logger test case, add SetSkipLine func.

Colstuwjx:
RT, fix bad test case.

### DIFF
--- a/src/common/utils/log/logger_test.go
+++ b/src/common/utils/log/logger_test.go
@@ -25,6 +25,11 @@ var (
 	message = "message"
 )
 
+// contains reports whether the string is contained in the log.
+func contains(t *testing.T, str string, lvl string, line, msg string) bool {
+	return strings.Contains(str, lvl) && strings.Contains(str, line) && strings.Contains(str, msg)
+}
+
 func TestSetx(t *testing.T) {
 	logger := New(nil, nil, WarningLevel)
 	logger.SetOutput(os.Stdout)
@@ -52,7 +57,7 @@ func TestDebug(t *testing.T) {
 	Debug(message)
 
 	str := buf.String()
-	if len(str) != 0 {
+	if str != "" {
 		t.Errorf("unexpected message: %s != %s", str, "")
 	}
 }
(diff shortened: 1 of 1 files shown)


Colstuwjx:
Ping @reasonerjt

ywk253100:
@Colstuwjx Can you explain why this PR is needed?

Colstuwjx:
I'm learning from harbor sourcecode, and I found there is a logical mistake in test case, so I opened PR to fix it, that' it.

ywk253100:
What logic mistake? Which line?

Colstuwjx:
Like
 
```
if strings.HasSuffix(str, "[ERROR] message") {
    ..
}
```

due to the default logger set skipLine `false`, the message would be

```
2018-09-04T10:50:27+08:00 [ERROR] [logger_test.go:124]: message
```

Of cause it wouldn't be true for `if strings.HasSuffix(str, "[ERROR] message")` this condition
But I don't think it tested the logger functions.

So I raised this PR.
Thanks.

src/common/utils/log/logger.go
```diff
@@ -93,6 +93,14 @@ func (l *Logger) SetLevel(lvl Level) {
 	l.lvl = lvl
 }
 
+//SetSkipLine sets the skipLine of Logger l
+func (l *Logger) SetSkipLine(skipLine bool) {
```
ywk253100: The `skipLine` is an internal variable, we shouldn't expose it. 
Even though we call the method to set `skipLine` as `true`, the `Debug()`, `Info()` and `Warning()` don't print the information about line either.

ywk253100:
@Colstuwjx Got it, thanks.

src/common/utils/log/logger.go
ywk253100: For 1, the `skipLine` is in the same scope with the test codes, you can just set it to `true` by calling `logger.skipLine = true` directly. 
And I think maybe we shouldn't set it to `true`, we should test it for the real usage.
For 2, I don't think it's the issue that Harbor should fix. The log collector shouldn't require that all the applications are using the same log format.

Colstuwjx: Fixed, thanks.

ywk253100:
@Colstuwjx Thanks a lot.

Issue #5481, opened 2018-08-02
The repository API should return the list with a fixed sorting order


steven-zou:
Currently, the order is random

```
[{id: 2, name: "migrating/registry", project_id: 2, description: "", pull_count: 0, star_count: 0,…},…]
0
:
{id: 2, name: "migrating/registry", project_id: 2, description: "", pull_count: 0, star_count: 0,…}
1
:
{id: 1, name: "migrating/redis", project_id: 2, description: "", pull_count: 0, star_count: 0,…}
```

-----

```
[{id: 1, name: "migrating/redis", project_id: 2, description: "", pull_count: 0, star_count: 0,…},…]
0
:
{id: 1, name: "migrating/redis", project_id: 2, description: "", pull_count: 0, star_count: 0,…}
1
:
{id: 2, name: "migrating/registry", project_id: 2, description: "", pull_count: 0, star_count: 0,…}
```

ywk253100:
Fixed in #5748. Thanks @cd1989.

PR #5748, opened 2018-08-28
Support repo list sorting

cd1989:
Support repo list sorting.

Add parameter `sort` to API `/api/repositories`. Support sorting based on:

- name
- creation_time
- update_time

Default to name ascending ordering.

Related to issue: https://github.com/goharbor/harbor/issues/5481

### DIFF
--- a/src/common/dao/repository.go
+++ b/src/common/dao/repository.go
@@ -22,16 +22,16 @@ import (
 	"github.com/goharbor/harbor/src/common/models"
 )
 
-var orderMap = map[string]string {
-	"name": "name asc",
-	"+name": "name asc",
-	"-name": "name desc",
-	"creation_time": "creation_time asc",
+var orderMap = map[string]string{
+	"name":           "name asc",
+	"+name":          "name asc",
+	"-name":          "name desc",
+	"creation_time":  "creation_time asc",
 	"+creation_time": "creation_time asc",
 	"-creation_time": "creation_time desc",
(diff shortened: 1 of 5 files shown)


src/common/dao/repository.go
```diff
@@ -114,12 +126,16 @@ func GetTotalOfRepositories(query ...*models.RepositoryQuery) (int64, error) {
 }
 
 // GetRepositories ...
-func GetRepositories(query ...*models.RepositoryQuery) ([]*models.RepoRecord, error) {
+func GetRepositories(sorting string, query ...*models.RepositoryQuery) ([]*models.RepoRecord, error) {
```
ywk253100: How about putting the `sorting` into the `query` as a field just like the `Pagination` in https://github.com/goharbor/harbor/blob/master/src/common/models/repo.go#L42? Maybe `sorting` can be a struct so that other APIs can reuse it.

ywk253100:
@cd1989 The root cause of #5481 is that after getting the repository list from database requests for repository detail information are sent in parallel. 
We can merge this PR and then fix that issue later.

cd1989:
The issue already got fixed in this pr, you want to revert that ?

src/common/dao/repository.go
cd1989: OK

cd1989:
@ywk253100 Got it, I have already updated the commit accordingly.

PR #5752, opened 2018-08-29
Execute dep ensure to update lib vendors

steven-zou:
Execute `dep ensure` to update lib vendors


steven-zou:
Close PR to ignore the vendor changes

PR #5761, opened 2018-08-30
Extend configuration length to 1024 in DB

wy65701436:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,8 @@ before_script:
   - echo $REGISTRY_URL
   - cd tests && sudo ./ldapprepare.sh && cd ..
   - sudo ./tests/admiral.sh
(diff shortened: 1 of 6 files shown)


make/migrations/postgresql/0001_initial_schema.up.sql
```diff
@@ -285,7 +285,7 @@ UNIQUE(namespace)
 create table properties (
  id SERIAL NOT NULL,
  k varchar(64) NOT NULL,
- v varchar(128) NOT NULL,
+ v varchar(256) NOT NULL,
```
reasonerjt: put the `ALTER TABLE` statement in `0002_xxxx.up.sql` so the migrator in adminserver can handle the schema upgrade.

PR #5823, opened 2018-09-05
Test TCP connection before upgrading database schema

ywk253100:
This commit moves the database schema upgrading after database initialization. The init will test TCP connection.

### DIFF
--- a/src/common/utils/utils.go
+++ b/src/common/utils/utils.go
@@ -88,15 +88,17 @@ func TestTCPConn(addr string, timeout, interval int) error {
 	cancel := make(chan int)
 
 	go func() {
+		n := 1
 		for {
 			select {
 			case <-cancel:
 				break
 			default:
-				conn, err := net.DialTimeout("tcp", addr, time.Duration(timeout)*time.Second)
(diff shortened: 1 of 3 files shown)


src/common/utils/utils.go
```diff
@@ -88,15 +88,17 @@ func TestTCPConn(addr string, timeout, interval int) error {
...
 			select {
 			case <-cancel:
 				break
 			default:
-				conn, err := net.DialTimeout("tcp", addr, time.Duration(timeout)*time.Second)
+				conn, err := net.DialTimeout("tcp", addr, time.Duration(n)*time.Second)
 				if err != nil {
 					log.Errorf("failed to connect to tcp://%s, retry after %d seconds :%v",
 						addr, interval, err)
+					n = n * 2
```
reasonerjt: Do you intend to update the interval or the timeout?

reasonerjt:
Please also resolve the conflict.

src/common/utils/utils.go
ywk253100: It's used to update the timeout for every try, not the total timeout or interval.

ywk253100:
@reasonerjt Updated.

PR #5827, opened 2018-09-05
Apply consistent format for comments

cd1989:
Add a whitespace after `//`.

`//Comment`  --> `// Comment` 

```
$ find src -name "*.go" | grep -v vendor | xargs sed -i -r 's/(^|\s)\/\/(\S)/\1\/\/ \2/g'
```


steven-zou:
@cd1989 

About the comment spec, is there any tools we can integrate into CI/CD pipeline to make sure the incoming pr can follow the same spec?

cd1989:
@steven-zou Sorry, I don't know there are such tools

PR #5832, opened 2018-09-05
Add comment format check to makefile and travis

cd1989:

Related to: https://github.com/goharbor/harbor/pull/5827

Add comment format check to makefile and travis CI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -376,13 +376,23 @@ gosec:
 		$(GOPATH)/bin/gosec -fmt=json -out=harbor_gas_output.json -quiet ./... | true ; \
 	fi
 
-go_check: misspell golint govet gofmt
+go_check: misspell golint govet gofmt commentfmt
 
 gofmt:
 	@echo checking gofmt...
 	@res=$$(gofmt -d -e -s $$(find . -type d \( -path ./src/vendor -o -path ./tests \) -prune -o -name '*.go' -print)); \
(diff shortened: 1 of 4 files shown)


reasonerjt:
@cd1989  thanks for the PR but is there a strong reason we need this check?  Why is it required since it's not part of `golint`?

cd1989:
@reasonerjt It should be a good practice to follow consistent comment spec, and `// Comment` is commonly accepted one rule, so I make a pr to correct this previously. @steven-zou thinks we'd better integrate one check into travis CI to ensure incoming PRs also follow this. I totally agree this, so I make this check.

wy65701436:
@cd1989 is there a way to fix it if my commit fails at check comments phase? Like gofmt error could be fixed by "-w".

cd1989:
@wy65701436 You can use following command to fix it.

```
find src -name "*.go" | grep -v vendor | xargs sed -i -r 's/(^|\s)\/\/(\S)/\1\/\/ \2/g'
```

wy65701436:
@cd1989 the regular expression fails to check the string like "// here is my comments //something bala bala"

cd1989:
@wy65701436 

```
$ echo "// here is my comments //something bala bala" | egrep '(^|\s)\/\/(\S)'
// here is my comments //something bala bala
```
It matches.

As for cases with more than one spaces, I ignore it since it's not easy to deal with cases like:

image

PR #5840, opened 2018-09-06
Fix `TestTCPConn` break issue.

Colstuwjx:
Currently, the `TestTCPConn` is a helper to test connection issue between harbor and database, and it may cause an uncontrolled goroutine issue while do `TestTCPConn` without break out the whole loop. The break inside dial goroutine would ONLY break out the select-case statement, rather than the whole for loop.

See the [golang specification](https://golang.org/ref/spec#Break_statements), `A "break" statement terminates execution of the innermost "for", "switch" or "select" statement.`

BTW, the test case wasn't broken, due to the go test is the main routine, while it returned, the other goroutine would also exit.

Thanks.

### DIFF
--- a/src/common/utils/utils.go
+++ b/src/common/utils/utils.go
@@ -89,10 +89,12 @@ func TestTCPConn(addr string, timeout, interval int) error {
 
 	go func() {
 		n := 1
+
+	loop:
 		for {
 			select {
 			case <-cancel:
-				break
+				break loop
 			default:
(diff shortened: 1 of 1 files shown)


Colstuwjx:
Ping @ywk253100

ywk253100:
@Colstuwjx Will review later, thanks.

ywk253100:
@Colstuwjx Thanks.

PR #5852, opened 2018-09-10
Let stale bot use the label "staled"

reasonerjt:
"wontfix" is used to mark the issues that were decided will not be fixed
after triage, so let's use a different label to mark stale issues.


steven-zou:
Seems the passive verb. 'staled' more reasonable?

reasonerjt:
@steven-zou 
Sure, updated.

PR #5890, opened 2018-09-14
Make the replication dialog a little bit wider

pureshine:
Signed-off-by: chengfangyuan <1309173081@qq.com>
I adjust the layout about 'Trigger Mode' to fit on one line


pureshine:
This is the final result

Issue #5495, opened 2018-08-02
In groups view. If no group , the delete button is enabled. 


zhoumeina:
The delete button is enabled by default if no group in groups view. Should make it disabled.

pureshine:
fiexed by pr: #5892

PR #5892, opened 2018-09-14
Disable delete button when there is no group

pureshine:
This is the error of #5495.
If no group , the delete button is enabled.
Signed-off-by: ChengFangyuan <1309173081@qq.com>


pureshine:
This is the end result

PR #5893, opened 2018-09-14
Rename harbor-ui to harbor-core

ninjadq:
1. Update the nginx.conf
2. Update Makefile
3. Update docker-compose
4. Update image name
5. Rename folder ui to core
6. Change the harbor-ui's package name to core
7. Remove unused static file on harbor-core
8. Remove unused code for harbor-portal


reasonerjt:
I recall the "secret" will be stored in DB.  Could you confirm if the renaming from "ui-secret" to "core-secret" will break migration?

PR #5896, opened 2018-09-15
Normalize (make lowercase) ldap_group_dn during onboarding

### LINKED ISSUE #5895: "Add a group from LDAP to project member" not normalize ldap_group_dn

Adding an LDAP group to a project through `Project > Members > + Group > Add a group from LDAP to project member` does not work if the LDAP Group DN (returned from the LDAP server) has uppercase letters. A user belonging to the LDAP group doesn't have the appropriate role.

- Prerequisite: An LDAP group whose DN on the LDAP server contains uppercase letters, e.g. `CN=test-group,OU=groups,DC=example,DC=com`
- Go to `Project > Members > + Group > Add a group from LDAP to project member`
- Enter the group DN
- Click Save
- Login with a user belonging to the above LDAP group
- Actual: The user does not have the corresponding role/access to the project
- Expected: The user should have access to the project with the role assigned to the group

**Versions:**
- harbor version: 1.6.0
- docker engine version: 18.06
- docker-compose version: 1.22.0

**Additional context:**
* **Harbor config files:** You can get them by packaging `harbor.cfg` and files in the same directory, including subdirectory.
* **Log files:** You can get them by package the `/var/log/harbor/` .

erks:
Fixes #5895 

This is because a user's group DNs are always lowercased, and comparisons are done in a case-sensitive fashion. 

https://github.com/goharbor/harbor/blob/9dca49ba6e4978a6ed22e471e105533322b68e5d/src/ui/auth/ldap/ldap.go#L95-L97

Following the practice in the same file to lowercase the LDAP group DN before persisting into the DB:

https://github.com/goharbor/harbor/blob/dcb518bcdda4fcede33823417273c10c98eee73f/src/common/dao/group/usergroup.go#L37

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ src/portal/typings/
 **/*npm-debug.log.*
 **/*yarn-error.log.*
 .idea/
+*.iml
 .DS_Store
 .project
 .vscode/
--- a/src/common/dao/group/usergroup.go
+++ b/src/common/dao/group/usergroup.go
@@ -116,6 +116,8 @@ func UpdateUserGroupName(id int, groupName string) error {
 // This is used for ldap and uaa authentication, such the usergroup can have an ID in Harbor.
 // the keyAttribute and combinedKeyAttribute are key columns used to check duplicate usergroup in harbor
 func OnBoardUserGroup(g *models.UserGroup, keyAttribute string, combinedKeyAttributes ...string) error {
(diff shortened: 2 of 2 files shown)


clouderati:
Thanks for the contribution @erks!

erks:
@clouderati is it possible that this gets cherry-picked to 1.6.x?

Issue #5472, opened 2018-08-02
[ui]In replication list, endpoint name can not sort. 


zhoumeina:
docker-compose version 1.17.1, build unknown

Docker version 18.05.0-ce, build f150324

zhoumeina:
Project also can not sort

PR #5897, opened 2018-09-17
Remove the sort of project and endpoint name

pureshine:
fiexed by pr: #5472 
Because endpoint name and project can not be sorted in replication list, sorting is removed.
Signed-off-by: Cheng Fangyuan <1309173081@qq.com>


pureshine:
This is the end result

PR #5923, opened 2018-09-19
[1.6.0] Fix statuses condition when trigger replication

cd1989:

When trigger a replication policy, Harbor will check whether the policy has related jobs that are pending or running. But there is some mistake in the check.

```
count, err := dao.GetTotalCountOfRepJobs(&models.RepJobQuery{
	PolicyID: replication.PolicyID,
	Statuses: []string{models.RepOpTransfer, models.RepOpDelete},
})
```


clouderati:
lgtm – @reasonerjt ptal

ywk253100:
@cd1989 Could cherry pick the #6025 to branch 1.6 to fix the travis failure?

PR #5924, opened 2018-09-19
Fix statuses condition when trigger replication

cd1989:

When trigger a replication policy, Harbor will check whether the policy has related jobs that are pending or running. But there is some mistake in the check.

```
count, err := dao.GetTotalCountOfRepJobs(&models.RepJobQuery{
	PolicyID: replication.PolicyID,
	Statuses: []string{models.RepOpTransfer, models.RepOpDelete},
})
```


clouderati:
lgtm – @reasonerjt ptal

reasonerjt:
lgtm
@ywk253100 ?

wy65701436:
@cd1989 please syne the latest code to make the travis pass.

cd1989:
@wy65701436 travis passed

ywk253100:
The wrong change is introduced in PR #4572, it lost the `Operations` property and set wrong value for `Statuses ` when querying replication jobs. So the correct query condition should be:
```
count, err := dao.GetTotalCountOfRepJobs(&models.RepJobQuery{
		PolicyID: replication.PolicyID,
		Statuses: []string{models.JobPending, models.JobRunning},
               Operations: []string{models.RepOpTransfer, models.RepOpDelete},
	})
```

cd1989:
Operations added

Issue #5729, opened 2018-08-27
use jib-maven-plugin push image to harbor success, also can pull image from harbor, but can not find image in harbor's web ui


dragontree101:
If you are reporting a problem, please make sure the following information are provided:
1)Version of docker engine and docker-compose.
docker version is 18.06.1-ce 
docker-compose version is 1.22.0
harbor verison is 1.6.0-rc2

2)Config files of harbor, you can get them by packaging "harbor.cfg" and files in the same directory, including subdirectory.
```
## Configuration file of Harbor

#This attribute is for migrator to detect the version of the .cfg file, DO NOT MODIFY!
_version = 1.6.0
#The IP address or hostname to access admin UI and registry service.
#DO NOT use localhost or 127.0.0.1, because Harbor needs to be accessed by external clients.
hostname = hb.cnhnkj.cn

#The protocol for accessing the UI and token/notification service, by default it is http.
#It can be set to https if ssl is enabled on nginx.
ui_url_protocol = http

#Maximum number of job workers in job service
max_job_workers = 10

#Determine whether or not to generate certificate for the registry's token.
#If the value is on, the prepare script creates new root cert and private key
#for generating token to access the registry. If the value is off the default key/cert will be used.
#This flag also controls the creation of the notary signer's cert.
customize_crt = on

#The path of cert and key files for nginx, they are applied only the protocol is set to https
ssl_cert = /data/cert/server.crt
ssl_cert_key = /data/cert/server.key

#The path of secretkey storage
secretkey_path = /data

#Admiral's url, comment this attribute, or set its value to NA when Harbor is standalone
admiral_url = NA

#Log files are rotated log_rotate_count times before being removed. If count is 0, old versions are removed rather than rotated.
log_rotate_count = 50
#Log files are rotated only if they grow bigger than log_rotate_size bytes. If size is followed by k, the size is assumed to be in kilobytes.
#If the M is used, the size is in megabytes, and if G is used, the size is in gigabytes. So size 100, size 100k, size 100M and size 100G
#are all valid.
log_rotate_size = 200M

#Config http proxy for Clair, e.g. http://my.proxy.com:3128
#Clair doesn't need to connect to harbor ui container via http proxy.
http_proxy =
https_proxy =
no_proxy = 127.0.0.1,localhost,ui,registry

#NOTES: The properties between BEGIN INITIAL PROPERTIES and END INITIAL PROPERTIES
#only take effect in the first boot, the subsequent changes of these properties
[...]

PR #5929, opened 2018-09-20
To allow goharbor UI to show the images pushed by other docker clients such as Skopeo

saikirandusari:
To resolve https://github.com/goharbor/harbor/issues/5729

To allow go harbor to show the images pushed by other docker clients such as Skopeo

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,31 +2,44 @@ sudo: true
 
 language: go
 
-go: 
-  - 1.9.2
+go:
+- 1.9.2
 
 go_import_path: github.com/goharbor/harbor
 
-services: 
-  - docker
+services:
+- docker
 
 dist: trusty
 
+matrix:
+  include:
+  - go: "1.9.2"
+    env:
+    - UTTEST=true
+  - go: "1.9.2"
+    env:
+    - APITEST_DB=true
+  - go: "1.9.2"
+    env:
+    - APITEST_LDAP=true
+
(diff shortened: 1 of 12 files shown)


clouderati:
I'm not opposed to the `UserAgent` change, though I do have a few thoughts:

* Spokeo should probably use a slightly more descriptive `UserAgent` string so that future Go-based clients can differentiate. Obviously that's out of our control. 😄 
* We've not tested this _at all_, so we need to update our documentation to make that clear.
* @saikirandusari, just want to confirm: have you actually tested Spokeo + Harbor functionality after making this change?

@reasonerjt any other concerns, technical or otherwise, that we should discuss?

clouderati:
@goharbor/harbor – ptal and provide feedback. Thanks!

reasonerjt:
IMO this change should not be merged to `master` as it will make notification handler unable to differentiate a `pull` from a rest call by regular Go HTTP Client.
As I suggested in another issue or thread, a better approach should be update the notification handler it can identify a `pull` or `push` command by checking a custom HTTP header of the request, rather than adding more and more user agents to the white list.

clouderati:
Thanks @reasonerjt. @saikirandusari any questions about @reasonerjt's approach?

clouderati:
Closing this for now. I still think it's a valuable feature to have and contributions to make this happen as per @reasonerjt's welcomed.

Issue #5114, opened 2018-06-08
After selecting a color when creating a label, the color menu should close.


lweitzman:
The color label doesn't close after making a selection.

zhoumeina:
I tried to set [clrCloseMenuOnItemClick]="true"  still not working, currently, clarity version is 0.10.27

PR #5930, opened 2018-09-20
After selecting color, the color button is closed

pureshine:
fixed: #5114 
Remove the trigger button and make our own button style
Signed-off-by: Cheng Fangyuan<1309173081@qq.com>


src/portal/lib/src/create-edit-label/create-edit-label.component.scss
```diff
@@ -61,5 +61,22 @@ form {
       border: 1px solid #A1A1A1 !important; 
       line-height: 22px;
     }
+
+    .icon-style
```
zhoumeina: why need add so many styles?

PR #5932, opened 2018-09-20
Refactor backend api for authrization

ninjadq:
1. Change backend api
2. Change frontend api
3. Change the proxy config file


nlowe:
Unless there is a separate API external clients should be using to get a session token, this is a breaking change. Shouldn't this cause a major version bump?

PR #5945, opened 2018-09-25
Return the total count of charts under the project in project API

steven-zou:
**Return the total count of charts under the project in project API**

- add new interface method to get the total count of charts under namespaces by calling get index
- add new field 'chart_count' in the project model
- append chart count to the project model in project API

### DIFF
--- a/src/chartserver/handler_utility.go
+++ b/src/chartserver/handler_utility.go
@@ -15,6 +15,21 @@ const (
 	maxDeletionThreads = 10
 )
 
+// GetCountOfCharts calculates and returns the total count of charts under the specified namespaces.
+// See @ServiceHandler.GetCountOfCharts
+func (c *Controller) GetCountOfCharts(namespaces []string) (uint64, error) {
+	if namespaces == nil || len(namespaces) == 0 {
+		return 0, nil // Directly return 0 instead of non-nil error
+	}
+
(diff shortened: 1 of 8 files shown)


src/chartserver/handler_utility.go
```diff
@@ -15,6 +15,21 @@ const (
 	maxDeletionThreads = 10
 )
 
+// GetCountOfCharts calculates and returns the total count of charts under the specified namespaces.
+// See @ServiceHandler.GetCountOfCharts
+func (c *Controller) GetCountOfCharts(namespaces []string) (uint64, error) {
```
reasonerjt: What's the point to return the sum of chart counts under multiple namespaces?

src/chartserver/handler_utility_test.go
```diff
@@ -4,6 +4,33 @@ import (
 	"testing"
 )
 
+// Test the function GetCountOfCharts
+func TestGetCountOfCharts(t *testing.T) {
```
reasonerjt: Better add a test case to include only non-existing namespace in chartmuseum.

src/chartserver/handler_utility.go
steven-zou: For the overall statistics like repo count in the project main page.

PR #5950, opened 2018-09-26
Make the replication dialog a little bit wider

### LINKED ISSUE #4621: New Replication Rule dialog should be wider to accommodate trigger mode (scheduled + weekly)
the New replication rule dialog should be a bit wider so the scheduled + weekly trigger mode fits on one line. also, all drop down menus for under trigger mode show scroll bar and should not (on Safari).

pureshine:
fixed #4621 
I adjust the layout about 'Trigger Mode' to fit on one line
Signed-off-by: Cheng Fangyuan<fangyuanc@vmware.com>

### DIFF
--- a/src/portal/lib/src/create-edit-rule/create-edit-rule.component.html
+++ b/src/portal/lib/src/create-edit-rule/create-edit-rule.component.html
@@ -88,8 +88,10 @@
               </select>
             </div>
           </div>
-            <label *ngIf="noEndpointInfo.length != 0" class="colorRed alertLabel">{{noEndpointInfo | translate}}</label>
-            <span class="alertLabel goLink" *ngIf="noEndpointInfo.length != 0" (click)="goRegistry()">{{'REPLICATION.ENDPOINTS' | translate}}</span>
+            <div>
+              <label *ngIf="noEndpointInfo.length != 0" class="colorRed alertLabel">{{noEndpointInfo | translate}}</label>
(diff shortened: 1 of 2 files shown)


src/portal/lib/src/create-edit-rule/create-edit-rule.component.html
```diff
@@ -81,7 +81,7 @@ <h3 class="modal-title">{{headerTitle | translate}}</h3>
         <!--Targets-->
         <div class="form-group form-group-override">
           <label class="form-group-label-override  required">{{'DESTINATION.ENDPOINT' | translate}}</label>
-          <div formArrayName="targets">
+          <div formArrayName="targets" class="form-select">
```
zhoumeina: CLASS should use "-".

src/portal/lib/src/create-edit-rule/create-edit-rule.component.html
```diff
@@ -105,7 +105,7 @@ <h3 class="modal-title">{{headerTitle | translate}}</h3>
               </select>
             </div>
             <!--on push-->
-            <div formGroupName="schedule_param">
+            <div formGroupName="schedule_param" class="schedule_style">
```
zhoumeina: class should use "-" instead of "_"

pureshine: Ok, I have changed it.

PR #5967, opened 2018-09-27
Remove unused Bootstrap from portal dependencies

### LINKED ISSUE #5961: [xss report] XSS report for Security about bootstrap 4.0.0-alpha.
There is XSS risk in github security report about bootstrap 4.0.0-alpha. This is caused by Clarity dependency on bootstrap.

gnomeontherun:
The Harbor project doesn't actually use Bootstrap, so this is an unnecessary dependency to have in your package.json file. Clarity doesn't require it unless you are building a custom theme, and Harbor doesn't do that.

jQuery was implicitly depended upon by some piece of the application, so this explicitly depends upon it so its tracked.

Fixes #5961


src/portal/package.json
```diff
@@ -59,11 +59,11 @@
         "@types/jasmine": "~2.8.6",
         "@types/jasminewd2": "~2.0.3",
         "@types/node": "~8.9.4",
-        "bootstrap": "4.0.0-alpha.5",
         "codelyzer": "~4.2.1",
         "enhanced-resolve": "^3.0.0",
         "jasmine-core": "~2.99.1",
         "jasmine-spec-reporter": "~4.2.1",
+        "jquery": "^3.3.1",
```
zhoumeina: @gnomeontherun , If remove the bootstrap, where does the styles comes from?

PR #6005, opened 2018-10-10
Fix LDAP search error

stonezdj:


clouderati:
Hi @stonezdj – looks like the CI job failed:

```2018-10-10T08:56:31Z [ERROR] [security.go:228]: failed to authenticate testUser0001: Failed to authenticate user, due to error 'Invalid credentials'
2018-10-10T08:56:31Z [INFO] unauthorized```

(see line 1394)

aladjadj:
This change is good, 
but tests failed because UT tests run stack with db Auth and tests not match ldap use case while in production environnement this use case is restricted.
I look to help to solve without quickfix  but it's tricky for the moment

PR #6007, opened 2018-10-10
Change admin server to core in jobservice

stonezdj:
Refactor admin client, remove the dependency to adminserver to core server

### DIFF
--- a/src/core/router.go
+++ b/src/core/router.go
@@ -97,6 +97,7 @@ func initRouters() {
 	beego.Router("/api/targets/:id([0-9]+)/policies/", &api.TargetAPI{}, "get:ListPolicies")
 	beego.Router("/api/targets/ping", &api.TargetAPI{}, "post:Ping")
 	beego.Router("/api/logs", &api.LogAPI{})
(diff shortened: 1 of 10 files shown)


src/core/router.go
```diff
@@ -111,6 +111,7 @@ func initRouters() {
 
 	beego.Router("/api/internal/syncregistry", &api.InternalAPI{}, "post:SyncRegistry")
 	beego.Router("/api/internal/renameadmin", &api.InternalAPI{}, "post:RenameAdmin")
+	beego.Router("/api/internal/configurations", &api.ConfigAPI{}, "get:GetInternalConfig")
```
steven-zou: Suggest replace 'internal' to avoid set bars for the system integration.

* `/api/configs/{scope}`
  * `/api/configs/usersettings`
  * `/api/configs/systemsettings`
* `/api/configs/{scope}/{group}`
* `/api/configs/{scope}/{group}/{key}`

steven-zou: * /api/configs/usersettings/emails
  * /api/configs/usersettings/emails/url
  * /api/configs/usersettings/emails/cc

PR #6030, opened 2018-10-12
add gc e2e test case

wy65701436:


tests/resources/Harbor-Pages/Configuration.robot
```diff
@@ -276,3 +276,18 @@ Delete A Label
...
+## Garbage Collection	
+Switch To Garbage Collection
+    Sleep  1
+    Click Element  xpath=${configuration_xpath}
+    Click Element  xpath=${garbage_collection_xpath}
+
+Click GC Now
+    Sleep  1
+    Click Element  xpath=${gc_now_xpath}
+    Sleep  5
```
ywk253100: Seems the`sleep 5` here is useless as the `Wait Until Page Contains  Finished` is used in the following case.

PR #6039, opened 2018-10-15
Refactor capacity

stonezdj:
Move dependency to admin server to core api.


src/core/api/systeminfo.go
```diff
@@ -120,7 +122,8 @@ func (sia *SystemInfoAPI) validate() {
 func (sia *SystemInfoAPI) GetVolumeInfo() {
 	sia.validate()
 
-	capacity, err := config.AdminserverClient.Capacity()
+	systeminfo.Init()
```
cd1989: Please remove capacity related code from adminserver to core. One problem is you are relying env (e.g. `IMAGE_STORE_PATH`) of adminserver.

stonezdj: The IMAGE_STORE_PATH is not set in adminserver, it is not set in core api,  uses default value /data.

src/core/router.go
```diff
@@ -112,6 +112,7 @@ func initRouters() {
 
 	beego.Router("/api/internal/syncregistry", &api.InternalAPI{}, "post:SyncRegistry")
 	beego.Router("/api/internal/renameadmin", &api.InternalAPI{}, "post:RenameAdmin")
+	beego.Router("/api/internal/configurations", &api.ConfigAPI{}, "get:GetInternalConfig")
```
steven-zou: Will this be a formal route in future?

stonezdj: It is not the final version, introduce /api/internal/configurations is used only for remove dependency and remove admin container.

PR #6063, opened 2018-10-17
Use docker official way to unset dns search

wy65701436:
According docker official document, use 'dns_search= .' in the docker
compose file if you don't wish to set the search domain.

https://docs.docker.com/v17.09/engine/userguide/networking/default_network/configure-dns/


reasonerjt:
BTW I think it's safe to remove the env
`GODEBUG=netdns=cgo` from docker-compose template.

PR #6067, opened 2018-10-17
Change the rights of developers and project admin in user guide.

pureshine:
Adding developer and project admin  can download registry certificates entries.


docs/user_guide.md
```diff
@@ -309,6 +308,11 @@ Go into the repository and select the "Info" tab, and click the "EDIT" button.
 
 ![edit info](img/edit_description.png)
 
+### Download the harbor certs
+
+Developers and project admin can click the "registry certificate" link to download the registry certificate.
```
reasonerjt: So is it true that guest can not download the cert?

zhoumeina: It is a download link, we don't do any changes to block anybody of it. For the button itself, only users who login harbor can see the download link. 
I am not sure how to describe it is best. I think can be describe it as "Users can click the registry certificate link".

Issue #3876, opened 2017-12-26
Use unique replication to track all the fired jobs for once executing replication rule


steven-zou:

supereagle:
Any plan for this?

cd1989:
I have made a PR for this issue.

cd1989:
It's done already. Can be closed now.

PR #6093, opened 2018-10-21
Add op uuid to image replication

cd1989:
Add uuid to track all jobs triggered by one replication execution.

Related issue: https://github.com/goharbor/harbor/issues/3876

### DIFF
--- a/src/core/api/replication.go
+++ b/src/core/api/replication.go
@@ -17,6 +17,7 @@ package api
 import (
 	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/goharbor/harbor/src/common/dao"
 	"github.com/goharbor/harbor/src/common/models"
@@ -26,6 +27,8 @@ import (
 	"github.com/goharbor/harbor/src/replication/core"
 	"github.com/goharbor/harbor/src/replication/event/notification"
 	"github.com/goharbor/harbor/src/replication/event/topic"
+
+	"github.com/docker/distribution/uuid"
 )
 
 // ReplicationAPI handles API calls for replication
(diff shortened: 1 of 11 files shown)


steven-zou:
@ywk253100 

Let's take a look at this PR.

src/core/api/replication.go
```diff
@@ -78,16 +81,27 @@ func (r *ReplicationAPI) Post() {
...
+
+	r.Data["json"] = api_models.ReplicationResponse{
+		UUID: opUUID,
+	}
+	r.ServeJSON()
 }
 
-func startReplication(policyID int64) error {
-	return notifier.Publish(topic.StartReplicationTopic,
+// startReplication triggers a replication and return the uuid of this replication.
```
clouderati: @reasonerjt Is there a reason that we don't log here that we've started a replication job (and in this case explicitly state the UUID of the job)?

reasonerjt: One replication job is designed to handle replication of one repository,
and one replication action, given certain filter, will map to multiple jobs in job service.
And this PR make it possible to trace a job based on replication action.

cd1989:
@ywk253100 Is it possible to make this into Harbor 1.7.0 ?

ywk253100:
@cd1989 we'll discuss later, thanks.

steven-zou:
@cd1989 

For the purpose of this PR, do you have any real scenario or customer requirements?

cd1989:
@steven-zou Yes. In our product, we need to get replication results of one replication action. If users trigger a replication manually, they want to get the status of this replication action. We are not using Harbor independently, but call APIs of Harbor and embed Harbor to our product.  Also, we have customers that have similar requirements.

clouderati:
I'm definitely in favor of this. Having a UUID that we can trace throughout various internal calls is tremendously invaluable.

cd1989:
@steven-zou @clouderati What's the conclusion of this PR, do we accept it in 1.7.0 ?

clouderati:
I'm ok with this landing in v1.7. @reasonerjt and @steven-zou – can we please make sure this gets merged if it's not breaking anything? Thanks.

steven-zou:
@clouderati @cd1989 

I think it's ok to merge into master. I'll do it now.

cd1989:
thx, @steven-zou

PR #6132, opened 2018-10-24
Enhance job and jobservice loggers to support multiple backends

steven-zou:
**Enhance job and jobservice loggers to support multiple backends**
  * use go-logging in liue of common log
  * support multiple logger backends configurations
  * file and stdout/stderr logger can be enabled at the same time

### DIFF
--- a/src/jobservice/job/impl/logger/job_logger.go
+++ b/src/jobservice/job/impl/logger/job_logger.go
@@ -15,35 +15,78 @@
 package logger
 
 import (
+	"errors"
+	"fmt"
 	"os"
+	"path"
 	"strings"
 
-	"github.com/goharbor/harbor/src/common/utils/log"
+	"github.com/goharbor/harbor/src/jobservice/config"
 	"github.com/goharbor/harbor/src/jobservice/logger"
+	logging "github.com/op/go-logging"
 )
+
+var format = logging.MustStringFormatter(
+	`%{color}%{time:15:04:05.000} %{shortfunc} ▶ %{level:.4s} %{id:03x}%{color:reset} %{message}`,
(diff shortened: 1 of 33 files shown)


src/jobservice/job/impl/logger/job_logger.go
```diff
@@ -15,35 +15,78 @@
...
+			fileLeveledBackend.SetLevel(parseLevel(logger.LogLevel), moduleName)
+
+			backends = append(backends, fileLeveledBackend)
+			continue
+		}
+
+		// Should be STD outs
+		stdOut := os.Stdout
+		if logger.Kind == config.LoggerKindStdError {
+			stdOut = os.Stderr
```
reasonerjt: The naming of the var is too confusing.  how about rename to `output`

src/jobservice/job/impl/logger/job_logger.go
```diff
@@ -15,35 +15,78 @@
...
 }
 
 // New logger
-// nil might be returned
-func New(logPath string, level string) logger.Interface {
-	f, err := os.OpenFile(logPath, os.O_CREATE|os.O_WRONLY, 0644)
-	if err != nil {
-		logger.Errorf("Failed to create job logger: %s", err)
-		return nil
+func New(jobID string) (logger.Interface, error) {
```
reasonerjt: The way the job logger is configured, makes it seems there should be multiple implementation of the interface, rather than switch to go-logging and add different backends to it.

steven-zou:
@reasonerjt 

Let's build a logger framework to support logging of jobs running job service. I'll close this PR and will submit new one later.

PR #5873, opened 2018-09-11
issue 5851 support jib client

mmpei:
Signed-off-by: mmpei <peimingming1986@126.com>

add for supporting jib client.
test success.
step1.compile harbor and bring up
2.config new harbor IP and account password in pom.xml in jib block.
3.compile using jib plugin and jib will auto pushing image to harbor.
4.check the image in project in harbor. success.

PR #6152, opened 2018-10-26
5178 jib supporting

mmpei:
add for supporting jib client.
test success.
step1.compile harbor and bring up
2.config new harbor IP and account password in pom.xml in jib block.
3.compile using jib plugin and jib will auto pushing image to harbor.
4.check the image in project in harbor. success.

resubmit for PR  https://github.com/goharbor/harbor/pull/5873


mmpei:
@steven-zou  please check.

PR #6154, opened 2018-10-26
Fix gc jobs issues

wy65701436:
This commit is to fix two problems in the gc job,
1, in order to support redis cluseter, gc job should delete the keys only
for docker reigstry instead of all keys. It because that in the redis cluster,
only db 0 is supported.
2, to restore the real status of read only not just to set it to false.

### DIFF
--- a/src/jobservice/job/impl/gc/job.go
+++ b/src/jobservice/job/impl/gc/job.go
@@ -25,6 +25,7 @@ import (
 	common_http "github.com/goharbor/harbor/src/common/http"
 	"github.com/goharbor/harbor/src/common/http/modifier/auth"
 	"github.com/goharbor/harbor/src/common/registryctl"
+	"github.com/goharbor/harbor/src/common/utils"
 	reg "github.com/goharbor/harbor/src/common/utils/registry"
 	"github.com/goharbor/harbor/src/jobservice/env"
 	"github.com/goharbor/harbor/src/jobservice/logger"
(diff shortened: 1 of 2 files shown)


reasonerjt:
Is it still true that we have to only use dbindex=0?

src/jobservice/job/impl/gc/job.go
```diff
@@ -139,11 +156,51 @@ func (gc *GarbageCollector) cleanCache() error {
...
+
+func delKeys(con redis.Conn, pattern string) error {
+	iter := 0
+	keys := []string{}
+	for {
+		arr, err := redis.Values(con.Do("SCAN", iter, "MATCH", pattern))
+		if err != nil {
+			return fmt.Errorf("error retrieving '%s' keys", pattern)
+		}
+		iter, err := redis.Int(arr[0], nil)
```
ywk253100: The variable `iter` should not be created again, or the for loop will be in dead lock.

src/jobservice/job/impl/gc/job.go
```diff
@@ -67,10 +70,16 @@ func (gc *GarbageCollector) Run(ctx env.JobContext, params map[string]interface{
...
+	if err != nil {
 		return err
 	}
-	defer gc.readonly(false)
+	if readOnlyCur != true {
+		if err := gc.setReadOnly(true); err != nil {
+			return err
+		}
+	}
+	defer gc.setReadOnly(readOnlyCur)
```
ywk253100: Only needed when `readOnlyCur != true`, so should put it in the `if readOnlyCur != true` block.

PR #6165, opened 2018-10-29
Fix failure on GC test case

wy65701436:
As the changes on scan all, the default schedule was removed from source code,
hence the gc job should be the first admin job, to change to job id to 1.

### DIFF
--- a/tests/robot-cases/Group1-Nightly/Common.robot
+++ b/tests/robot-cases/Group1-Nightly/Common.robot
@@ -58,7 +58,7 @@ Test Case - Garbage Collection
     Sleep  1
     Wait Until Page Contains  Finished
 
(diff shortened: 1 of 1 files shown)


tests/robot-cases/Group1-Nightly/Common.robot
```diff
@@ -58,7 +58,7 @@ Test Case - Garbage Collection
     Sleep  1
     Wait Until Page Contains  Finished
 
-    ${rc}  ${output}=  Run And Return Rc And Output  curl -u ${HARBOR_ADMIN}:${HARBOR_PASSWORD} -s --insecure -H "Content-Type: application/json" -X GET "https://${ip}/api/system/gc/2/log"
+    ${rc}  ${output}=  Run And Return Rc And Output  curl -u ${HARBOR_ADMIN}:${HARBOR_PASSWORD} -s --insecure -H "Content-Type: application/json" -X GET "https://${ip}/api/system/gc/1/log"
```
ywk253100: How about removing the `-s` option and adding `-i` so that we can debug what happens when the API call fails?

PR #6167, opened 2018-10-29
Fix the issue of duplicated periodic job scheduling

steven-zou:

### DIFF
--- a/src/jobservice/period/enqueuer.go
+++ b/src/jobservice/period/enqueuer.go
@@ -15,6 +15,7 @@
 package period
 
 import (
+	"errors"
 	"fmt"
 	"math/rand"
 	"time"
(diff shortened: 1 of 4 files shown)


src/jobservice/period/enqueuer.go
```diff
@@ -91,19 +93,34 @@ func (pe *periodicEnqueuer) loop() {
...
 	defer conn.Close()
 
+	// Set last periodic enqueue timestamp in advance to avoid duplicated enqueue actions
+	if _, err := conn.Do("SET", utils.RedisKeyLastPeriodicEnqueue(pe.namespace), now); err != nil {
+		return err
+	}
+
+	// Avoid schedule in the same time slot.
+	expireTime := (int64)(periodicEnqueuerSleep/time.Minute) * (60 + 5)
+	slot := fmt.Sprintf("%s:%d", utils.RedisKeyLastPeriodicEnqueue(pe.namespace), now)
```
reasonerjt: 1) This chunk is repetitive with line 110~113.
2) `fmt.Sprintf("%s:%d", utils.RedisKeyLastPeriodicEnqueue(pe.namespace), now)` this may not work in the multi-instance scenario, as the value of `now` in different instances may be different.

Issue #6168, opened 2018-10-29
Retag should not be allowed for Guest role


wy65701436:
Steps to reproduce:
1, Log in as a guest.
2, Go into the library -- public project.
3, Retag is enable for me.

Expected:
Retag should be disabled for the guest user.

reasonerjt:
@wy65701436 Does the API blocks `guest`?

cd1989:
The API blocks guest, but the portal UI didn't.

steven-zou:
@cd1989 

Seems the API does not block the request fired by a user with a guest role.

PR #6170, opened 2018-10-29
Allow retag in portal only for developer or above role

cd1989:


cd1989:
https://github.com/goharbor/harbor/issues/6168

wy65701436:
@cd1989 thanks for fixing.

PR #6173, opened 2018-10-30
refactor configuration

zhoumeina:
1 remove save and cancel button from config.component.ts
2 add save and cancel logic in authentication and email and system component. 
3 getChanges function add filter to filterout other changes in configuration.


ninjadq:
btw, it's better to provide a more detailed commit msg to describe this changes

PR #6192, opened 2018-11-01
Fix: When session expire can not click anything

ninjadq:
This is casued that not handle 401 err on member routing guard


src/portal/src/app/shared/route/leaving-repository-deactivate.service.ts
```diff
@@ -37,6 +37,7 @@ export class LeavingRepositoryRouteDeactivate implements CanDeactivate<TagReposi
     // Confirmation before leaving config route
     return new Promise((resolve, reject) => {
       if (tagRepo && tagRepo.hasChanges()) {
+        console.log('haschanges');
```
zhoumeina: submit the console log?

PR #6212, opened 2018-11-05
refactor system settings and fixed bug of system settings

zhoumeina:
1 split system-setting logic from config.component.ts
2 fix error of cancel buttons not working
3 fix issue of system setting, modify the create project attribute can not save.


src/portal/lib/src/config/system/system-settings.component.ts
```diff
@@ -60,9 +95,115 @@ export class SystemSettingsComponent {
...
+
+    reset(changes: any): void {
+        if (!isEmpty(changes)) {
+            for (let prop in changes) {
+                if (this.originalConfig[prop]) {
+                    this.config[prop] = clone(this.originalConfig[prop]);
+                }
+            }
+        } else {
+            // force reset
```
ninjadq: if changes are empty, why force retrieve config?

src/portal/lib/src/config/system/system-settings.component.ts
```diff
@@ -60,9 +95,115 @@ export class SystemSettingsComponent {
...
+                .then(response => {
+                    this.onGoing = false;
+                    // API should return the updated configurations here
+                    // Unfortunately API does not do that
+                    // To refresh the view, we can clone the original data copy
+                    // or force refresh by calling service.
+                    // HERE we choose force way
+                    this.retrieveConfig();
+                    if ('read_only' in changes) {
+                        this.readOnlyChange.emit(changes['read_only']);
```
ninjadq: why consider read_only as a specific case?

zhoumeina: This is from the original code. I did not modified it. I think it because read only will have a global alert

src/portal/lib/src/config/system/system-settings.component.ts
zhoumeina: I did not add this logic. It just moved from config.component.ts

PR #6240, opened 2018-11-08
Create os trusted certificate for registry, clair, chartmuseum

stonezdj:
Fix the issue when registry using a customized s3 storage, the registry , clair, chartmuseum need to trust the certificates.

### DIFF
--- a/make/prepare
+++ b/make/prepare
@@ -531,9 +531,14 @@ else:
     print("Copied configuration file: %s" % registry_config_dir + "root.crt")
     shutil.copyfile(os.path.join(templates_dir, "registry", "root.crt"), os.path.join(registry_config_dir, "root.crt"))
 
+# Create os trusted certificate for registry, clair, chartmuseum
+common_cert_path = os.path.join(config_dir, "ca-bundle.crt")
+version_output = subprocess.check_output("grep  goharbor/registry-photon docker-compose.yml  |grep image", shell=True)
+registry_version = version_output.split(":")[2].strip()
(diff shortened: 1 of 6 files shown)


make/prepare
```diff
@@ -531,9 +531,14 @@ else:
     print("Copied configuration file: %s" % registry_config_dir + "root.crt")
     shutil.copyfile(os.path.join(templates_dir, "registry", "root.crt"), os.path.join(registry_config_dir, "root.crt"))
 
+# Create os trusted certificate for registry, clair, chartmuseum
+common_cert_path = os.path.join(config_dir, "ca-bundle.crt")
+version_output = subprocess.check_output("grep  goharbor/registry-photon docker-compose.yml  |grep image", shell=True)
+registry_version = version_output.split(":")[2].strip()
+os.system("docker run -i --rm goharbor/registry-photon:"+registry_version+" cat  /etc/pki/tls/certs/ca-bundle.crt > " + common_cert_path)
```
reasonerjt: The error is not handled.
I prefer appending the ca cert inside the container, the purpose for prepare is populating configuration files.  As this is vulnerable if the image name changes.  And it will add more dependencies to `prepare`, considering we may put it into a container.

Maybe we can talk to photon developers to see if there's more flexible way to put the cert into truststore in addition to appending the ca-bundle.crt?  
If there's no other way, I suggest we do the append when the container starts rather than in `prepare`.  If there's concern about the dup code, we can create a utility script and put them into the images during `docker build`

PR #6242, opened 2018-11-08
Update user guide screenshot

ninjadq:
update the screenshots in user guide
update document in user guide


docs/user_guide.md
```diff
@@ -480,6 +480,17 @@ Check the checkbox at the 1st column to select the specified chart versions:
 * Click the `DOWNLOAD` button to download the chart artifact file. Batch operation is not supported.
 * Click the `UPLOAD` button to upload the new chart version for the current chart
 
+#### Adding labels to/remove labels from chart versions
```
reasonerjt: We can say `charts` instead of `chart versions`

reasonerjt:
1. In `docs/img/chartrepo/add_labesl_to_chart_versions.png` the text is "add labels to image", but it should be chart.
2. In `harbor/docs/img/chartrepo/chart_details.png` I recall we discussed to rework the lables area in chart details view?

docs/user_guide.md
ninjadq: I think charts may make some confusion because the label cannot add to a chart. it can only add to a chart's version currently

PR #6258, opened 2018-11-12
Enhance: Add helm chart count number in the project list

ninjadq:
Project API return the helm chart count now, so put it to UI.


src/portal/src/i18n/lang/es-es-lang.json
```diff
@@ -154,6 +154,7 @@
         "PUBLIC_OR_PRIVATE": "Público",
         "ACCESS_LEVEL": "Nivel de acceso",
         "REPO_COUNT": "Contador de repositorios",
+        "CHART_COUNT": "Chart Count",
```
zhoumeina: Not sure what is the purpose of this , if it means the number of Helm Chart. It should be translated to "Helm Chart Number".Is my understanding right?

PR #6312, opened 2018-11-20
add document for gc

wy65701436:
This commit is to update user guide to introduce the handbook of GC


docs/user_guide.md
```diff
@@ -552,3 +553,30 @@ helm install --ca-file=ca.crt --key-file=server.key --cert-file=server.crt --use
 
 For other more helm commands like how to sign a chart, please refer to the [helm doc](https://docs.helm.sh/helm/#helm).
 
+## Online Garbage Collection
+Online Garbage Collection is to let user to trigger docker registry garbage collection by clicking button in UI.
```
reasonerjt: enables user to trigger ..... on UI.

docs/user_guide.md
```diff
@@ -552,3 +553,30 @@ helm install --ca-file=ca.crt --key-file=server.key --cert-file=server.crt --use
 
 For other more helm commands like how to sign a chart, please refer to the [helm doc](https://docs.helm.sh/helm/#helm).
 
+## Online Garbage Collection
+Online Garbage Collection is to let user to trigger docker registry garbage collection by clicking button in UI.
+
+**NOTES:** GC is not stoppable, and it deletes unused image layers.
```
reasonerjt: Sometimes it's referred as "GC" , sometimes it's "Garbage Collection", we'd better fix the inconsistency.

reasonerjt: We should explain when the images are removed the space is not freed, user needs to run GC to free up the space, etc, etc.  Or, how about add a link to explain "Garbage Collection"

Issue #6148, opened 2018-10-26
The installation guide need to be updated


ywk253100:
Some of the instructions are outdated in the installation guide, we need to go through and update it. Users get confused with the outdated guide: #6137. Here are part of them:
1. Missing properties: LDAP group ...
2. The default value changed: job workers ...
3. The way used to configure changed: storage ...

PR #6320, opened 2018-11-21
Migration guide update

kyrogue:
#6148 

Signed-off-by: zhiqiangchin kyroguex@gmail.com

### DIFF
--- a/docs/migration_guide.md
+++ b/docs/migration_guide.md
@@ -1,45 +1,36 @@
 # Harbor upgrade and database migration guide
 
-When upgrading your existing Habor instance to a newer version, you may need to migrate the data in your database and the settings in harbor.cfg. Refer to [change log](../tools/migration/changelog.md) to find out whether there is any change in the database. If there is, you should go through the database migration process. Since the migration may alter the database schema and the settings of harbor.cfg, you should **always** back up your data before any migration.
-
-*If your install Harbor for the first time, or the database version is the same as that of the lastest version, you do not need any database migration.*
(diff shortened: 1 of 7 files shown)


kyrogue:
The backup step in the original documentation was confusing in that, we were asked to backup the folder by moving it, then the next steps we are using the migrator tool to backup the folders by mounting them. Internally i missed out using the migrator tool to backup the databases, but rather just did a full copy on /data.

And step 3 which said to download the release but did not say what to do with it, in this case i assumed it is to be extracted to the original Harbor directory, since it does not create a /data directory the migrated databases directory will already be there.

I feel we should move out the backup/restore to another documentation and as there are a few notes in each step making it hard to read linearly.

docs/migration_guide.md
```diff
@@ -1,6 +1,6 @@
 # Harbor upgrade and database migration guide
 
-Refer to [change log](../tools/migration/changelog.md) to find out whether there is any change in the database. If there is, you should go through the database migration process. 
+Refer to [change log](../tools/migration/changelog.md) to find out whether there is any change in the database or harbor.cfg. If there is, you should go through the database migration process.
 
 **NOTE:**
 - From v1.6.0 onwards, Harbor migrated from MariaDB to Postgresql, combining Harbor, Notary and Clair databases into one.
```
kyrogue: We set a assumed directory, and updated the commands to be consistent.

For steps 4 and 5, i think there is confusion on which of steps should be taken, as it looks like you can perform multiple different ways of migration depending on your situation.

PR #6338, opened 2018-11-23
Fix chart ui related issues

ninjadq:
* Fix: i18n issue in add label filter header in chart version
* Fix: strange string in cmd input in the chart summary
* Fix: change label style in helm chart summary page
* Fix: upload chart version testcase failure in nightly
* Fix: helm chart results style


src/portal/lib/src/label/label-marker/label-marker.component.html
```diff
@@ -1,5 +1,5 @@
 <div>
-    <label class="dropdown-header">{{'REPOSITORY.ADD_TO_IMAGE' | translate}}</label>
+    <label class="dropdown-header">{{addLabelHeaders | translate}}</label>
```
zhoumeina: this drop down lable does not need to translate?

PR #6344, opened 2018-11-26
Bump up golang to 1.11.2

reasonerjt:


make/photon/registry/Dockerfile.binary
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.7.3
```
wy65701436: this should not be updated as docker uses 1.7.3 to build the v2.6.2 registry.

reasonerjt: OK let's shift to 1.11 when bump up to registry v2.7

PR #6371, opened 2018-11-28
6144 js logger in db

mmpei:
test cases:

1. Start a image scan job. Job successful. after job finished, check data in table Job_log in DB.
2. Start a image scan job. Job failed. after job finished, check data in table Job_log in DB.
3. Check the Job log produced in step 1 and 2, Log from step 2 checked from UI and 1 from OpenAPI.
4. Confing using sweeper，use default Duration(1 day)，Start JobService，the expired logs will be clean immediately.
5. Confing using sweeper，use default Duration(1 day)，Start JobService，the expired logs will be clean after one day.
6. delete DB data, and startup harbor again,  DB will be created and table job_log will be created too.
7. Case for limit of length of Job content. when size of log proceeds 4k retrieving log normally. No limit for log size.
8. all Unit Test cases run successfully。

### DIFF
--- a/src/common/dao/joblog.go
+++ b/src/common/dao/joblog.go
@@ -1,8 +1,8 @@
 package dao
 
 import (
-	"github.com/goharbor/harbor/src/common/models"
 	"github.com/astaxie/beego/orm"
+	"github.com/goharbor/harbor/src/common/models"
 	"time"
 )
 
--- a/src/common/dao/joblog.go
+++ b/src/common/dao/joblog.go
@@ -1,0 +1,40 @@
+package dao
+
+import (
+	"github.com/goharbor/harbor/src/common/models"
+	"github.com/astaxie/beego/orm"
+	"time"
+)
+
+// CreateOrUpdateJobLog ...
+func CreateOrUpdateJobLog(log *models.JobLog) (int64, error) {
+	o := GetOrmer()
+	count, err := o.InsertOrUpdate(log, "job_uuid")
+	if err != nil {
+		return 0, err
+	}
+
+	return count, nil
+}
+
+// GetJobLog ...
+func GetJobLog(uuid string) (*models.JobLog, error) {
+	o := GetOrmer()
+	jl := models.JobLog{UUID: uuid}
+	err := o.Read(&jl, "UUID")
+	if err == orm.ErrNoRows {
+		return nil, err
+	}
+	return &jl, nil
+}
+
+// DeleteJobLogsBefore ...
+func DeleteJobLogsBefore(t time.Time) (int64, error) {
+	o := GetOrmer()
+	sql := `delete from job_log where creation_time < ?`
+	res, err := o.Raw(sql, t).Exec()
+	if err != nil {
+		return 0, err
+	}
+	return res.RowsAffected()
+}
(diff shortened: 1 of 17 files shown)


src/common/dao/joblog.go
```diff
@@ -0,0 +1,40 @@
+package dao
+
+import (
+	"github.com/goharbor/harbor/src/common/models"
+	"github.com/astaxie/beego/orm"
+	"time"
+)
+
+// CreateOrUpdateJobLog ...
+func CreateOrUpdateJobLog(log *models.JobLog) (int64, error) {
```
reasonerjt: Does it need to return count?

src/jobservice/job/impl/context.go
```diff
@@ -95,7 +96,14 @@ func (c *Context) Init() error {
...
 	db := getDBFromConfig(configs)
 
-	return dao.InitDatabase(db)
+	err = dao.InitDatabase(db)
+	if err != nil {
+		return err
+	}
+
+	// Initialize DB finished
+	initDBCompleted()
```
reasonerjt: Is it necessary?  IMO we can assume the DB is ready when DBSweeper sweeps the log?

src/jobservice/logger/sweeper/db_sweeper.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+}
+
+// Sweep logs
+func (dbs *DBSweeper) Sweep() (int, error) {
+	// DB initialization not completed, waiting
+	WaitingDBInit()
+
+	// Start to sweep logs
+	before := time.Now().Add(time.Duration(dbs.duration) * oneDay * -1)
```
reasonerjt: time.Now().Sub?

src/jobservice/logger/known_loggers.go
```diff
@@ -79,3 +87,24 @@ func IsKnownLevel(level string) bool {
 
 	return false
 }
+
+// Return a logger name by Interface
+func GetLoggerName(l Interface) string {
```
reasonerjt: Maybe we should add `func Name() string` to logger interface? @steven-zou

src/jobservice/job/impl/context.go
mmpei: actually not, DB initialized after logger and DB initialization relay on logger.
Sweeper will be executed once as soon as logger initialization finished.

src/jobservice/logger/sweeper/db_sweeper.go
mmpei: sub accept  a parameter Time and return Duration, not applicable.

src/common/dao/joblog.go
mmpei: not necessary. always return 1 for Job_uuid is a unique index in DB.
just a result from ORM InsertOrUpdate.

src/jobservice/logger/sweeper/db_sweeper.go
reasonerjt: correct, my bad

steven-zou:
@mmpei 

Thanks a lot for such great PR. I did not see the decrease count `-0.2%` before merging. It is a little bigger (We have merging criteria, the coverage should not decrease more than `-%0.03`)  Would u please to add more UT cases to increase the testing coverage in a new PR?

mmpei:
OK, i will add some UT code and create a new PR these days.

PR #6381, opened 2018-11-29
Check the existence of name when creating replication rule

### LINKED ISSUE #5102: Can create duplicate replication rule
Version v1.5.0-bf6192f3

Can create two same replication rules

steps: create a replication rule (for example: a), then create a new with the same name, then will a hint name exist, then click cancel and click yes to cancel it, create replication rule with the same name again

replication

ywk253100:
This commits fix #5102 by checking the existence of name when creating/editing replication rule

### DIFF
--- a/src/core/api/replication_policy.go
+++ b/src/core/api/replication_policy.go
@@ -128,6 +128,18 @@ func (pa *RepPolicyAPI) Post() {
 	policy := &api_models.ReplicationPolicy{}
 	pa.DecodeJSONReqAndValidate(policy)
 
+	// check the name
+	exist, err := exist(policy.Name)
+	if err != nil {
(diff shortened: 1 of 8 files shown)


src/core/api/replication_policy.go
```diff
@@ -128,6 +128,18 @@ func (pa *RepPolicyAPI) Post() {
...
 
+	// check the name
+	exist, err := exist(policy.Name)
+	if err != nil {
+		pa.HandleInternalServerError(fmt.Sprintf("failed to check the existence of policy %s: %v", policy.Name, err))
+		return
+	}
+
+	if exist {
+		pa.HandleConflict("name is already used")
```
wy65701436: pa.HandleConflict(fmt.Sprintf("the policy name %s already exists", policy.Name))

PR #6386, opened 2018-11-29
change session cookie name to sid

### LINKED ISSUE #6355: Got error gob: name not registered for interface: "github.com/vmware/harbor/src/common/models.User"
Got error `gob: name not registered for interface: "github.com/vmware/harbor/src/common/models.User"` when upgrading 1.6.0 to the latest version as the base path is changed from `vmware` to `goharbor`. Maybe we can clean the session store when starting up.

heww:
change session cookie name to sid and this fix #6355

### DIFF
--- a/src/core/main.go
+++ b/src/core/main.go
@@ -72,6 +72,7 @@ func updateInitPassword(userID int, password string) error {
 
 func main() {
 	beego.BConfig.WebConfig.Session.SessionOn = true
+	beego.BConfig.WebConfig.Session.SessionName = "sid"
 	// TODO
 	redisURL := os.Getenv("_REDIS_URL")
 	if len(redisURL) > 0 {


clouderati:
Hi @heww – please forgive the silly question: can you please share a bit of background on why this cookie name was chosen and why we need this fix?

Thanks!

heww:
@clouderati sure.

Harbor use `beego` as web framework to provider API and others.  The session of `beego`  use `gob` to encode and decode interface{} to save and read from session store.

Before v1.6.0 the package name used by Harbor is `github.com/vmware/harbor`, so `models.User` is registered  in `gob` by name `github.com/vmware/harbor/src/common/models.User`.

And in latest version the package name of Harbor is changed to `github.com/goharbor/harbor`, and the name of `models.User` registered in `gob` is changed to `github.com/vmware/harbor/src/common/models.User`.

When upgrade Harbor from v1.6.0 to the latest version, authorized user session generated at v1.6.0 will read failed in latest version for bytes encoded with `github.com/vmware/harbor/src/common/models.User`  can not be decode (see more #6355).

That make the user can not login the portal ui until the session is expired in session store and the default session expired time is 1 hour.

The best way to solve this problem is clean all sessions after upgrade from v1.6.0 to the latest version by manual, but this is not a good experience for upgrade. 

Change the session name make `beego` ignore the session generated by v1.6.0 and regenerate new session, no manual clean for sessions required after upgrade.

clouderati:
Thanks @heww – makes sense. Very much appreciate the explanation!

PR #6396, opened 2018-11-30
Support unique job constraints (both unique queue and running jobs)

steven-zou:
Currently, the uniqueness checking only covers the job queue. The running job is not taken into consideration.
In this PR, the unique scope is expanded to job queue and running jobs. That means if the job is declared to be unique, there will be only one job running at the same time.

### DIFF
--- a/src/jobservice/pool/de_duplicator.go
+++ b/src/jobservice/pool/de_duplicator.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/goharbor/harbor/src/jobservice/errs"
 	"github.com/goharbor/harbor/src/jobservice/models"
 	"github.com/goharbor/harbor/src/jobservice/utils"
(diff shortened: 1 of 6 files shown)


src/jobservice/pool/de_duplicator.go
```diff
@@ -0,0 +1,94 @@
...
+func NewDeDuplicator(ns string, pool *redis.Pool) *DeDuplicator {
+	return &DeDuplicator{
+		namespace: ns,
+		pool:      pool,
+	}
+}
+
+// Unique checks if the job is unique and set unique flag if it is not set yet.
+func (dd *DeDuplicator) Unique(jobName string, params models.Parameters) error {
+	uniqueKey, err := redisKeyUniqueJob(dd.namespace, jobName, params)
```
wy65701436: should handle uniqueKey for "" as the redisKeyUniqueJob could return ""

PR #6404, opened 2018-12-01
Fix link to Online GC in user guide

nlowe:
The anchor tag for Online GC isn't correct in the user guide. This PR fixes that.


nlowe:
Gah I keep forgetting about the DCO check. Is there a way to set that by default for a repo? My google-fu is failing me.

PR #6408, opened 2018-12-03
Add unit test for JobService Logger

mmpei:
add unit test for JobService Logger.

### DIFF
--- a/src/jobservice/logger/factory.go
+++ b/src/jobservice/logger/factory.go
@@ -81,5 +81,9 @@ func DBFactory(options ...OptionItem) (Interface, error) {
 		}
 	}
 
+	if len(key) == 0 {
+		return nil, errors.New("missing key option of the db logger")
+	}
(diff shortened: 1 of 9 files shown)


mmpei:
@steven-zou  add UT as we talked, please review.

src/jobservice/logger/factory.go
```diff
@@ -81,5 +81,9 @@ func DBFactory(options ...OptionItem) (Interface, error) {
 		}
 	}
 
+	if len(key) == 0 {
```
ywk253100: Is there any default value for `key` if user doesn't set it in `config.yaml`?

mmpei: this is a Job UUID, not config parameters. and this is the unique index for DB log data, must exists.

ywk253100: thanks

PR #6414, opened 2018-12-04
Corrected DOCKERCMD to fix portal multistage build

SDBrett:
removed --pull from DOCKERCMD variable, missed this update with last commit.
Signed-off-by: Brett Johnson <brett@sdbrett.com>


SDBrett:
Closing for the time being as the issue that this addresses might be related to an older version of Docker being installed.

Issue #6455, opened 2018-12-05
Retag should not be effective in read-only mode 


wy65701436:
[Steps]
1, set harbor to read-only.
2, retag an image.

That's should not work as harbor has put into read-only mode.

reasonerjt:
This can be closed? @wy65701436 @cd1989

cd1989:
Yes

PR #6457, opened 2018-12-05
Block retag requests in read-only mode

cd1989:

Block retag requests in read-only mode.

Related issue: https://github.com/goharbor/harbor/issues/6455

/assign @wy65701436


src/core/filter/readonly.go
```diff
@@ -71,6 +74,21 @@ func matchRepoTagDelete(req *http.Request) bool {
 	return false
 }
 
+// matchRetag checks whether a request is a retag request, it should be blocked in read-only mode.
+func matchRetag(req *http.Request) bool {
+	if req.Method != http.MethodPost {
+		return false
+	}
+
+	re := regexp.MustCompile(tagURL)
```
wy65701436: the url of retag is not same with tag deletion, and add more UT please.

Issue #6453, opened 2018-12-05
Fail to RETAG a image to other "public" project by a user with develop role


danfengliu:
Got an unknow error when RETAG a image to other "public" project by a user with develop role, does a public project can be push images by any user?

**Product steps:**
1. Harbor already has a public project with name of library;
2. Create a user by admin, create a project by admin;
3. Add this user to the new project as a develop member, and push a image to this new project;
4. Go to this image, and perform "RETAG" operation, got unknow error.

![image](https://user-images.githubusercontent.com/44018302/49504206-2bf3cd80-f8b4-11e8-9153-16dbfb22fe43.png)

![image](https://user-images.githubusercontent.com/44018302/49504124-0c5ca500-f8b4-11e8-901a-4796da4931e7.png)

zhoumeina:
@cd1989 ， could you please take a look?

cd1989:
Sure

danfengliu:
@cd1989 I modified this issue, because I learned out that this kind of operation should be forbidden, so this issue's key point to is to give user a reasonable information, not an unknown error.

zhoumeina:
@cd1989 , correct. It should be forbidden. I think one more message should be added when user retag  the image successfully. It should also popup a green notification.

cd1989:
According to my understanding, only admin role has write permission to public project. I have fixed based on this.

reasonerjt:
Is it in the RC code? if yes, let's close this. @wy65701436 @cd1989

cd1989:
Yes, we can close

PR #6470, opened 2018-12-06
Give meaningful error messages when retag is forbidden

cd1989:

Related issue: https://github.com/goharbor/harbor/issues/6453

image


cd1989:
@zhoumeina @steven-zou Does developer role has write permission to public project? This PR assumes it's no, and lots of UT gotten failed due to this.

Issue #6450, opened 2018-12-05
Test LDAP server error without save configuration


wy65701436:
[Steps]
1, admin - configuration
2, change auth to LDAP, and set up the ldap server.
3, test it with error response.
4, save it, test it with success response

![image](https://user-images.githubusercontent.com/2841473/49501209-4d04f000-f8ad-11e8-8837-d66d9ea1fc38.png)

[Root cause]
the ladp search password is missed in the payload of the request.

reasonerjt:
Let's keep this open until the PR is merged.

reasonerjt:
And I don't think this has to be a high priority, if it requires more effort than estimated earlier, we can defer it.

stonezdj:
fixed #7328

PR #6471, opened 2018-12-06
Fix issue 6450 which can not ping ldap without save configuration

stonezdj:
fix bug #6450

### DIFF
--- a/src/common/utils/ldap/ldap.go
+++ b/src/common/utils/ldap/ldap.go
@@ -44,16 +44,18 @@ type Session struct {
 }
 
 // LoadSystemLdapConfig - load LDAP configure from adminserver
-func LoadSystemLdapConfig() (*Session, error) {
+func LoadSystemLdapConfig(skipModeCheck ...bool) (*Session, error) {
+	// skip auth_mode check for skipModeCheck
+	if !(len(skipModeCheck) > 0 && skipModeCheck[0]) {
+		authMode, err := config.AuthMode()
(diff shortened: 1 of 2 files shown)


src/common/utils/ldap/ldap.go
```diff
@@ -44,16 +44,18 @@ type Session struct {
...
 
 // LoadSystemLdapConfig - load LDAP configure from adminserver
-func LoadSystemLdapConfig() (*Session, error) {
-
-	authMode, err := config.AuthMode()
-	if err != nil {
-		log.Errorf("can't load auth mode from system, error: %v", err)
-		return nil, err
-	}
+func LoadSystemLdapConfig(ping ...bool) (*Session, error) {
```
reasonerjt: Firstly the variable name is confusing, it 'd better be `skipModeCheck`, otherwise, what if we later decide we need to check the auth_mode when user triggers `ping`?

Additionally,  I briefly checked the reference of this func, I don't quite get why we need to check it at all, this func is for loading the LDAP config, the mode should not matter.

stonezdj: We need to keep the auth_mode check because if current mode is not ldap_auth, the /api/ldap/users/search, /api/ldap/users/import should failed.

reasonerjt: Shouldn't it fail at the API level instead of `LoadLDAPConfig`?

PR #6482, opened 2018-12-07
Return unknow when about dialog cannot get harbor version

pureshine:
About dialog returns unknow instead of 1.2.0 when harbor version is not available

### DIFF
--- a/src/portal/src/app/app-config.ts
+++ b/src/portal/src/app/app-config.ts
@@ -42,7 +42,7 @@ export class AppConfig {
         this.project_creation_restriction = "everyone";
         this.self_registration = true;
         this.has_ca_root = false;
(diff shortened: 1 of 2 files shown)


src/portal/src/app/app-config.ts
```diff
@@ -42,7 +42,7 @@ export class AppConfig {
         this.project_creation_restriction = "everyone";
         this.self_registration = true;
         this.has_ca_root = false;
-        this.harbor_version = "1.2.0";
+        this.harbor_version = "unknow";
```
clouderati: s/unknow/unknown/

pureshine: Okay, I've revised it.

PR #6486, opened 2018-12-07
Modify Project name's regexp when creating new Project

micline:
When I created new project named 'api-gateway' in harbor, portal didn't show project list. It only showed error message about Internal server error. This bug was caused by new Project name started with 'api'.  I think that the name started with 'api' caused a crash against a ingress path using '/api'.
So I changed regular expression in project name check code. I changed them in portal and core.
I also changed warning-msg in English. I know only English in provided languages.


clouderati:
Good catch @micline!

clouderati:
While this is getting reviewed, can you please sign-off on your commit to make the DCO bot happy? :+1:

micline:
I redo pull request

Issue #6465, opened 2018-12-06
Unable to sign images using notary after upgrading to 1.6


srinivasa-vasu:
When I try to sign the images, on the client side I am getting the following exception,

`2c178a8093fd: Layer already exists
12c374f8270a: Layer already exists
0c3170905795: Layer already exists
df64d3292fd6: Layer already exists
Signing and pushing trust metadata
unable to reach trust server at this time: 500.`

On the server side, if I look at the logs, I could see some DB related issues,
`Dec  5 16:32:36 172.18.0.1 notary-signer[2515]: #015
Dec  5 16:32:36 172.18.0.1 notary-signer[2515]: #033[35m(pq: permission denied for sequence private_keys_id_seq)#033[0m
Dec  5 16:32:36 172.18.0.1 notary-signer[2515]: #033[33m[2018-12-05 16:32:36]#033[0m #033[31;1m #033[0m
Dec  5 16:32:36 172.18.0.1 notary-signer[2515]: #015
Dec  5 16:32:36 172.18.0.1 notary-signer[2515]: #033[35m(pq: permission denied for sequence private_keys_id_seq)#033[0m
Dec  5 16:32:36 172.18.0.1 notary-signer[2515]: #033[33m[2018-12-05 16:32:36]#033[0m #033[31;1m #033[0m
Dec  5 16:32:36 172.18.0.1 notary-signer[2515]: {"go.version":"go1.7.3","level":"error","msg":"CreateKey: failed to create key: failed to store key: failed to add private key to database: 07de3a0f172bc9a068fe76f1eb0287017962c9cf17ee1d808623e2009f2e9347","time":"2018-12-05T16:32:36Z"}`

**Harbor version: 1.6.x**

srinivasa-vasu:
Not a fresh install. Upgraded

`Dec  6 04:17:06 172.18.0.1 notary-server[2515]: {"go.version":"go1.7.3","masked.domain.com/spring/spring-music:blue":"gun","http.request.host":"masked.domain.com:4443","http.request.id":"bb526810-0254-4652-9398-719fceb472b7","http.request.method":"GET","http.request.remoteaddr":"106.51.26.84","http.request.uri":"/v2/masked.domain.com/spring/spring-music:blue/_trust/tuf/timestamp.key","http.request.useragent":"Go-http-client/1.1","level":"error","msg":"500 GET timestamp key: rpc error: code = 13 desc = Key creation failed","time":"2018-12-06T04:17:06Z"}
[...]

PR #6488, opened 2018-12-07
Fix unable to use Notary after updating to v1.6.0

wy65701436:
This commit is to fix the issue on notary migrations from mysql to pgsql.
1, alter sequence for the primary key of changeseed, this is missed in v1.6.0 migrator.
2, alter table owners from postgres to notarysigner and notaryserver.

Issue:
https://github.com/goharbor/harbor/issues/6465

Workaround:
https://github.com/goharbor/harbor/issues/6465#issuecomment-445162616

Impacted upgrade path:
1, Upgrade from version older then v1.6.0 with migrator:v1.6.0, and migrates the notarty DB.

No impacted upgrade path:
1, Upgrade from version older than v1.6.0 with migrator:v1.6.0, but without migrates the notarty DB.

Notes：
After merge this fix, we need to provide an new migrator with an new tag, like v1.6.1, and
deprecated the v1.6.0. For those who was impacted by migrator v1.6.0, will open an new PR to build
the workaround into the migrator and expose an specical command for hot-fix.


tools/migration/db/util/mysql_pgsql_data_converter.py
```diff
@@ -148,6 +149,12 @@ def write_sequence(pgsql_dump, table_name, table_columnn):
     pgsql_dump.write("SELECT setval('%s_%s_seq', max(%s)) FROM %s;\n" % (table_name, table_columnn, table_columnn, table_name))
     pgsql_dump.write("ALTER TABLE \"%s\" ALTER COLUMN \"%s\" SET DEFAULT nextval('%s_%s_seq');\n" % (table_name, table_columnn, table_name, table_columnn))
 
+def alter_sequence(pgsql_dump, table_name, table_columnn):
```
reasonerjt: In pgSQL 9.5+ there's support in `CREATE SEQUENCE IF NOT EXIST` 

https://www.postgresql.org/docs/current/sql-createsequence.html

PR #6493, opened 2018-12-10
[RE]Modify Project name's regexp when creating new Project

micline:
When I created new project named 'api-gateway' in harbor, portal didn't show project list. It only showed error message about Internal server error. This bug was caused by new Project name started with 'api'. I think that the name started with 'api' caused a crash against a ingress path using '/api'.
So I changed regular expression in project name check code. I changed them in portal and core.
I also changed warning-msg in English. I know only English in provided languages.


src/core/api/project.go
```diff
@@ -532,6 +532,12 @@ func validateProjectReq(req *models.ProjectRequest) error {
 		return fmt.Errorf("project name is not in lower case or contains illegal characters")
 	}
 
+	validProjectName = regexp.MustCompile(`api(?:[._-][a-z0-9]+)*`)
+	legal = validProjectName.MatchString(pn)
+	if legal {
+		return fmt.Errorf("project name is not started with 'api'")
```
clouderati: s/is not started with/may not start with/

clouderati: https://github.com/goharbor/harbor/blob/d9b0ef4618ceb493de046dd615e1212d17e4891e/src/core/api/project.go#L537-L539

Shouldn't this be ```if !legal```?

micline: I indent to check whether a project name is started with 'api'. "api(?:[._-][a-z0-9]+)" means regular expression started with 'api'. Though I think that using previous variable 'legal' is effective, it seems quiet confuse. So I'm going to change that section more readable.

micline:
@clouderati I modify these codes more readable. Now, legal checks both of two conditions; including 'api' check.

PR #6497, opened 2018-12-10
Modify the error message when deleting a signed images in harbor-helm

### LINKED ISSUE #6442: Error message when deleting a signed images is wrong in harbor-helm
Steps to reproduce:
1. Install harbor-helm with ingress mode
2. Enable notary in client and push a signed image
3. Delete the signed images in web console.
4. The error message displayed in the Event log is wrong: the message is not correctly rendered. the notary server is wrong. it should be notary.harbor.domain instead of core.harbor.domain:4443

pureshine:
fixed #6442 
Remove {{param}} from internationalized documents


pureshine:
image

This is the final result.

Issue #6430, opened 2018-12-04
Inconsistent http error handling.


reasonerjt:
So far I find there are different types to represent http errors:

`HTTPError` in https://github.com/goharbor/harbor/blob/master/src/common/utils/error/error.go
 and 
`Error` in https://github.com/goharbor/harbor/blob/master/src/common/http/error.go

Seems we should only need one?

PR #6516, opened 2018-12-11
Remove the duplicate http error struct

ywk253100:
There are two different types to represent http error in the current code. This commit updates the codes to keep only one.


ywk253100:
Fixes #6430.

PR #6525, opened 2018-12-13
add developer center in swagger ui way

zhoumeina:
This pr is focus on integrate swagger ui into harbor. 
In frontend ,add swagger-ui package as one of the dependency. 
Create dev-center component to deal with the integration. The portal container will convert swagger.yml to swagger.json. And serve it in nginx server. 
dev-center will invoke swagger.json and modify the host and schemes dynamically to make sure the api can be invoked in correct url


make/photon/portal/entrypoint.sh
```diff
@@ -13,4 +18,7 @@ npm run build_lib
 npm run link_lib
 
 ## Build production
-npm run release
\ No newline at end of file
+npm run release
+
+
+
```
ninjadq: why add these empty lines?

src/portal/package.json
```diff
@@ -32,18 +32,26 @@
...
+        "@ng-bootstrap/ng-bootstrap": "^2.0.0",
         "@ngx-translate/core": "^10.0.2",
         "@ngx-translate/http-loader": "^3.0.1",
         "@types/jquery": "^2.0.41",
         "@webcomponents/custom-elements": "^1.1.3",
+        "bootstrap": "^4.1.1",
+        "buffer": "^5.2.1",
         "core-js": "^2.5.4",
         "intl": "^1.2.5",
+        "jquery": "^3.3.1",
```
ninjadq: jquery is already included in types/jquery, I think we can remove it from `package.json`

src/portal/src/polyfills.ts
```diff
@@ -37,3 +37,6 @@ import 'intl/locale-data/jsonp/zh';
 
 import 'zone.js/dist/zone';
 
+(window as any).global = window;
```
ninjadq: why assign `window` to its global property?

make/photon/portal/entrypoint.sh
zhoumeina: done.

src/portal/package.json
zhoumeina: The version is different from what swagger-ui depends.

src/portal/src/polyfills.ts
zhoumeina: global is a node js window.

Issue #6540, opened 2018-12-17
Error occured when to retag a image with special character


wy65701436:
The UI should block this input, and give user the tips about the naming convention of tag.

wy65701436:
<img width="1321" alt="image" src="https://user-images.githubusercontent.com/2841473/50074201-1f596880-0216-11e9-977a-6c52c19a333c.png">

zhoumeina:
@cd1989 , I think the backend api still has something not correct. It need to always returns a json format. This is because backend returns a html page. However, the frontend still think it is a string message. don't returns a redirect page in one api. Because harbor is a SPA application. 

I think fix this issue need to 
1 Modify the backend api to return in json format. 
2 block the special chars in frontend using Regex。 
3 After the dialog closed , the input value in the dialog should be cleared. 
4 Should add tooltip to show users only can input 26 chars and numbers in the input

zhoumeina:
![image](https://user-images.githubusercontent.com/4946616/50076168-e4f2ca00-021b-11e9-9cae-7313e791f290.png)
The value still there when reopen the dialog.

zhoumeina:
![image](https://user-images.githubusercontent.com/4946616/50076232-18355900-021c-11e9-8d71-2782ea91e000.png)
the request expect backend to return a json. but backend returns a html actually.

cd1989:
I will fix it today

zhoumeina:
fixed by:#6553

PR #6547, opened 2018-12-18
Validate repo and tag names in retag

cd1989:

Related to issue: https://github.com/goharbor/harbor/issues/6540

### DIFF
--- a/src/common/utils/validate.go
+++ b/src/common/utils/validate.go
@@ -1,0 +1,24 @@
+package utils
+
+import (
+	"fmt"
+	"regexp"
+)
+
+const nameComponent = `[a-z0-9]+((?:[._]|__|[-]*)[a-z0-9]+)*`
+
+// TagRegexp is regular expression to match image tags, for example, 'v1.0'
+var TagRegexp = regexp.MustCompile(`^[\w][\w.-]{0,127}$`)
+
+// RepoRegexp is regular expression to match repo name, for example, 'busybox', 'stage/busybox'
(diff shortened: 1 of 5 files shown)


src/common/utils/validate.go
```diff
@@ -0,0 +1,24 @@
...
+	"regexp"
+)
+
+const nameComponent = `[a-z0-9]+((?:[._]|__|[-]*)[a-z0-9]+)*`
+
+// TagRegexp is regular expression to match image tags, for example, 'v1.0'
+var TagRegexp = regexp.MustCompile(`^[\w][\w.-]{0,127}$`)
+
+// RepoRegexp is regular expression to match repo name, for example, 'busybox', 'stage/busybox'
+var RepoRegexp = regexp.MustCompile(fmt.Sprintf("^%s(/%s)*$", nameComponent, nameComponent))
```
wy65701436: according to the docker official document, https://forums.docker.com/t/docker-registry-v2-spec-and-repository-naming-rule/5466

 If a repository name has two or more path components, they must be separated by a forward slash ("/").

Why do we only support two path components?

src/core/api/repository.go
```diff
@@ -463,6 +463,16 @@ func (ra *RepositoryAPI) Retag() {
 		return
 	}
 
+	if !utils.ValidateRepo(repo) {
```
wy65701436: move the validation up to line 437, it should be the first action in the API.

src/common/utils/validate.go
cd1989: You may misunderstand it, it do support multiple components.

src/core/api/repository.go
cd1989: done

PR #6611, opened 2018-12-25
Add customized config file for proxy

ninjadq:
This is a solution for product that using harbor as proxy to add customized location

### DIFF
--- a/make/prepare
+++ b/make/prepare
@@ -4,6 +4,7 @@ from __future__ import print_function, unicode_literals # We require Python 2.6 
 from string import Template
 import random
 import os
+from fnmatch import fnmatch
 import sys
 import string
 import argparse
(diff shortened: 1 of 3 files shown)


reasonerjt:
`add_additional_location_config `  I don't think we need this , but given the prepare will be reworked anyway, it's ok as long as it works in VIC.

make/prepare
```diff
@@ -308,8 +311,28 @@ if protocol == "https":
...
+    :params dst: destination file path
+    """
+    if not os.path.isfile(src):
+        return
+    print("Copying nginx configuration file {src} to {dst}".format(src=src, dst=dst))
+    shutil.copy2(src, dst)
+    mark_file(dst)
+
+nginx_template_ext_dir = os.path.join(templates_dir, 'nginx', 'ext')
+if os.path.exists(nginx_template_ext_dir):
```
reasonerjt: We can remove `add_additional_location_config`, and copy the directory to target `conf.d` directory, and call mark_file to `conf.d`.

PR #6652, opened 2018-12-30
Add a function to check python version

scaat:
Before executing the prepare script, add a check for the python version to make sure there is already a python environment on the machine.

Signed-off-by: Scaat Feng <scaat.feng@gmail.com>


scaat:
@wy65701436

PR #6671, opened 2019-01-02
Add ram pkg

heww:
Add RAM pkg to implement HasPermission in RBAC mode


src/common/ram/cabin.go
```diff
@@ -0,0 +1,110 @@
+// Copyright Project Harbor Authors
```
reasonerjt: The file name should be `casbin.go`

src/common/ram/cabin.go
```diff
@@ -0,0 +1,110 @@
...
+
+# Policy definition
+[policy_definition]
+p = sub, obj, act
+
+# Role definition
+[role_definition]
+g = _, _
+
+# Policy effect
```
reasonerjt: Add more policy effect to support deny, though we don't need it now...

src/common/ram/cabin.go
```diff
@@ -0,0 +1,110 @@
...
+[policy_effect]
+e = some(where (p.eft == allow))
+
+# Matchers
+[matchers]
+m = g(r.sub, p.sub) && r.obj == p.obj && (r.act == p.act || p.act == '*')
```
reasonerjt: Add more variation in obj matching.

src/common/ram/cabin.go
heww: file renamed

src/common/ram/cabin.go
heww: policy definition add effect

src/common/ram/cabin.go
heww: obj matching changed to keyMatch2 which suport * and :var

src/common/ram/casbin.go
```diff
@@ -0,0 +1,133 @@
...
+func (a *userAdapter) LoadPolicy(model model.Model) error {
+	for _, line := range a.getUserAllPolicyLines() {
+		persist.LoadPolicyLine(line, model)
+	}
+
+	return nil
+}
+
+func (a *userAdapter) SavePolicy(model model.Model) error {
+	return nil
```
steven-zou: Suggest return errors.New("Not implemented")

src/common/ram/ram.go
```diff
@@ -0,0 +1,113 @@
...
+// Resource the type of resource
+type Resource string
+
+func (res Resource) String() string {
+	return string(res)
+}
+
+// Subresource returns subresource
+func (res Resource) Subresource(resources ...Resource) Resource {
+	els := []string{res.String()}
```
steven-zou: Can we change `els` with another meaningful term?

heww: els changed to elements

src/common/ram/casbin.go
heww: fixed

Issue #6596, opened 2018-12-21
Get internal server error when getting the count of chart if create a project named "api"


ywk253100:
Related issue: https://github.com/goharbor/harbor-helm/issues/128

steven-zou:
Let's fix it in 1.7.1

steven-zou:
Linked to the issue of upstream project https://github.com/helm/chartmuseum/issues/201

steven-zou:
Let's wait for the fix from chartmeseum team.

steven-zou:
No solution from chartmuseum side, let's move this issue out of the SP.

petvaa01:
All project names prefixed with "api" are affected with this.

steven-zou:
@ninjadq 

Reassign this issue to you

reasonerjt:
@steven-zou @ninjadq 
I tried it and it seems pretty severe, that any project prefixed by `api` will break the functionality, we should consider fix this earlier.

anairinac:
This issue is still affecting projects being created with version 1.8. It would be at least useful to forbid names with "api-" prefix, to avoid any problems.

PR #6673, opened 2019-01-02
Fix the issue of naming project with 'api' which is a segment of URL

steven-zou:
As the `api` is part of the chartmesum URL, there will be problem when creating a project with name `api`. The url path `/api/index.yaml` will be parsed as API interface but there is no handler to handle it. 404 error will be thrown out!

See issue #6596 


steven-zou:
High risk, pending for merging. Close this PR

PR #6682, opened 2019-01-02
Update the upgrade sql to rename the duplicate records

ywk253100:
Rename the "name" colume in table "replication_policy" and "replication_target" before adding the "UNIQUE" constraint to avoid the upgrade failure


ywk253100:
@wy65701436 Have added the timestamp to the name to avoid the conflict with the existing names.

make/migrations/postgresql/0002_1.7.0_schema.up.sql
```diff
@@ -11,5 +11,48 @@ create table job_log (
...
+        UPDATE replication_target AS t
+        SET name = (
+            CASE WHEN (length(name)+length(v.seq::text)+1) > 64 
+            THEN
+                substring(name from 1 for (63-length(v.seq::text))) || '_' || v.seq
+            ELSE
+                name || '_' || v.seq
+            END
+        )
+        FROM (SELECT id, row_number() OVER (PARTITION BY name ORDER BY id) seq FROM replication_target) AS v
```
reasonerjt: Better be `AS seq` for consistency?

make/migrations/postgresql/0002_1.7.0_schema.up.sql
```diff
@@ -11,5 +11,48 @@ create table job_log (
 
 CREATE UNIQUE INDEX job_log_uuid ON job_log (job_uuid);
 
+/*
+Rename the duplicate names before adding "UNIQUE" constraint
+*/
+DO $$ 
+BEGIN
+    WHILE EXISTS (select count(*) from replication_policy group by name having count(*) > 1) LOOP
```
reasonerjt: upper-case please.

Issue #6687, opened 2019-01-03
Counter "Pulls" doesn't work when use crictl


hardoverflow:
**Expected behavior and actual behavior:**
When i use:
```
crictl pull registry.fqdn/library/image:latest
```
the counter in the webui of harbor doesn't count up. The situation is different with the following call:
```
docker pull registry.fqdn/library/image:latest
```
**Steps to reproduce the problem:**
```
crictl pull image-xyz:tag
```
**Versions:**
Please specify the versions of following systems.
- harbor version: v1.7.0-3b3311b4
- docker engine version: 18.09.0
- docker-compose version: 1.23.2, build 1110ad01

**Additional logs:**
```
tailf /var/log/harbor/proxy.log
```
```
Jan  3 12:44:54 172.18.0.1 proxy[5062]: 1.2.3.4 - "GET /v2/ HTTP/1.1" 401 87 "-" "cri-o/1.11.10 go/go1.10.2 os/linux arch/amd64" 0.005 0.005 .
```

CrystalMethod:
401 return code means "unauthorized" .. so it is understandable that the counter does not change. Are you sure the pull works as expected?

hardoverflow:
Yes that's correct. The pull works with docker and crictl.
```
1 proxy[5062]: 1.2.3.4 - "GET /v2/ HTTP/1.1" 401 87 "-" "cri-o/1.11.10 go/go1.10.2 os/linux arch/amd64" 0.006 0.006 .
Jan  3 13:02:30 172.18.0.1 proxy[5062]: 1.2.3.4 - "GET /service/token?scope=repository%3Alibrary%2Fhaproxy%3Apull&service=harbor-registry HTTP/1.1" 200 947 "-" "Go-http-client/1.1" 0.020 0.020 .
Jan  3 13:02:31 172.18.0.1 proxy[5062]: 1.2.3.4 - "GET /v2/library/haproxy/manifests/alpine HTTP/1.1" 200 946 "-" "cri-o/1.11.10 go/go1.10.2 os/linux arch/amd64" 0.027 0.027 .
Jan  3 13:02:31 172.18.0.1 proxy[5062]: 1.2.3.4 - "GET /v2/ HTTP/1.1" 401 87 "-" "cri-o/1.11.10 go/go1.10.2 os/linux arch/amd64" 0.002 0.002 .
[...]

PR #6689, opened 2019-01-03
Remove user agent check for docker notifications

### LINKED ISSUE #5222: Push events wont work with custom client
Hi,

i'm trying to interact with custom clients to the harbor registry.

After some debugging I figured out that repositories or clair push events are only triggered if I use the docker-ce client.

On the source code I found that the harbor emits events depends on the User-Agent from the HTTP call.

https://github.com/vmware/harbor/blob/8f9e2644075744e091b91d7a5fdab73973c28b50/src/ui/service/notifications/registry/handler.go#L162

Any reasons why there are such limitations? If yes a blacklist should preffered.

### LINKED ISSUE #5729: use jib-maven-plugin push image to harbor success, also can pull image from harbor, but can not find image in harbor's web ui

jkroepke:
Fixes #5729 
Fixes #5222

### DIFF
--- a/src/core/service/notifications/registry/handler.go
+++ b/src/core/service/notifications/registry/handler.go
@@ -173,12 +173,11 @@ func filterEvents(notification *models.Notification) ([]*models.Event, error) {
 }
 
 func checkEvent(event *models.Event) bool {
-	// pull and push manifest by docker-client or vic or jib
-	if (strings.HasPrefix(event.Request.UserAgent, "docker") || strings.HasPrefix(event.Request.UserAgent, vicPrefix) || strings.HasPrefix(event.Request.UserAgent, "jib")) &&
-		(event.Action == "pull" || event.Action == "push") {
+	// pull and push manifest
+	if strings.ToLower(strings.TrimSpace(event.Request.UserAgent)) != "harbor-registry-client" && (event.Action == "pull" || event.Action == "push") {
 		return true
 	}
-	// push manifest by docker-client or job-service
+	// push manifest by job-service
 	if strings.ToLower(strings.TrimSpace(event.Request.UserAgent)) == "harbor-registry-client" && event.Action == "push" {
 		return true
 	}


steven-zou:
#6687 has the same problem.

jkroepke:
@ywk253100 @steven-zou

Small chance to get this into 1.7.x ?

steven-zou:
@jkroepke 

  The PR has missed the 1.7.1 timeline. Let's see if we have other possible rel slot for this PR. And we also need more supporting cases (reasons) to determine whether it can be put into the 1.7.x rel or not.

Asgoret:
@jkroepke @steven-zou 
I wan't to build OKD with registry based on Harbor, but i can't truly use harbor features because when i push images from podman it doesn't seemed in GUI. In which version does this PR will be released?

Asgoret:
@steven-zou @ywk253100 
So.... is there any chance, that this PR will be included in 1.7.x release?

Asgoret:
@steven-zou @ywk253100
Any information?

Asgoret:
@wy65701436 
Hi! Maybe you can help me. This PR is included in release 1.7.2?

ywk253100:
@Asgoret The fix will not included in the 1.7.x releases.

Asgoret:
@ywk253100 Why? Fix is ready and it's huge bug which stop use another container runtime?

jomeier:
We need this bugfix very urgently. I cant believe that such a huge bug is in this registry.

Even automatic replication doesn't work which is an absolute showstopper for our application.

Isn't it possible to pull this PR in a 1.7.5 ?

Best regards,

Josef

jomeier:
@ywk253100
I can confirm that this PR definitely solves the problem. 

Do you have any idea when v1.8 will be released?

Greetings, Josef

ywk253100:
@jomeier At end of May if everything goes well.

reasonerjt:
@jomeier we'll try to cherrypick this commit to 1.7.5, targeting early April.

PR #6702, opened 2019-01-07
Add DB table for robot account

wy65701436:
This commit is to add DB scheme for robot account and update the db orm releated.


make/migrations/postgresql/0004_add_robot_account.up.sql
```diff
@@ -0,0 +1,13 @@
+create table robot (
```
reasonerjt: CREATE TABLE (upper case)

PR #6708, opened 2019-01-08
disable retag function if with admiral

zhoumeina:
cherry pick disable retag function from 1.7.0 release branch to master. To make them consistent.

### DIFF
--- a/src/portal/lib/package.json
+++ b/src/portal/lib/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@harbor/ui",
-    "version": "1.0.0-rc7",
+    "version": "1.7.1-rc1",
     "description": "Harbor shared UI components based on Clarity and Angular6",
     "author": "CNCF",
     "module": "index.js",
--- a/src/portal/lib/src/tag/tag.component.html
+++ b/src/portal/lib/src/tag/tag.component.html
@@ -75,7 +75,7 @@
               </div>
             </clr-dropdown-menu>
           </clr-dropdown>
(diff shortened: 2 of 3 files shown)


src/portal/lib/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@harbor/ui",
-    "version": "1.0.0-rc7",
+    "version": "1.7.1-toggle-retag",
```
steven-zou: I don't think `1.7.1-toggle-retag` is a reasonable tag, please rename it to be `1.7.1-pacth1`

src/portal/lib/src/tag/tag.component.html
```diff
@@ -75,7 +75,7 @@ <h3 class="modal-title">{{ 'REPOSITORY.RETAG' | translate }}</h3>
               </div>
             </clr-dropdown-menu>
           </clr-dropdown>
-          <button type="button" class="btn btn-sm btn-secondary" [disabled]="!(selectedRow.length===1 && guestRoleOrAbove)" (click)="retag(selectedRow)"><clr-icon shape="copy" size="16"></clr-icon>&nbsp;{{'REPOSITORY.RETAG' | translate}}</button>
+          <button type="button" class="btn btn-sm btn-secondary" [disabled]="!(selectedRow.length===1 && guestRoleOrAbove) || withAdmiral" (click)="retag(selectedRow)"><clr-icon shape="copy" size="16"></clr-icon>&nbsp;{{'REPOSITORY.RETAG' | translate}}</button>
```
steven-zou: This just disabled the button, I think what we want to do is 'hide' the button if I'm not mistaken.
The button should not be displayed.

src/portal/lib/package.json
zhoumeina: resovled

PR #6717, opened 2019-01-09
Promte version on master to v1.8.0

wy65701436:


reasonerjt:
Thanks for doing this early @wy65701436

Issue #6714, opened 2019-01-08
Update Docs for importing the Clair vulnerability data


anasypany:
The docs at https://github.com/goharbor/harbor/blob/master/docs/import_vulnerability_data.md are out of date (they were written for harbor v1.2) and do not work with Harbor v1.7+. The docs still reference the separate Clair-db container that was used before the database consolidation work done in v1.6. 

If anyone has done a vulnerability data migration to an offline install in v1.7+ feel free to share your steps and I will gladly write up a PR for the current docs.

SDBrett:
Self assign

reasonerjt:
@SDBrett I don't think the issue can be assigned to a user that is not a member of `goharbor` team, but please feel free to write up a PR.  
You help will be appreciated!

anasypany:
@SDBrett If you've been able to successfully migrate on v1.7 would you mind sharing the commands you're exporting/importing the data with?

SDBrett:
@anasypany Below are the changes in commands. Pretty much it's all the same, but 'clair-db' is replaced with 'harbor-db'

## Harbor version >= 1.6

Databased were consolidated in version 1.6 which moved the clair database to the harbor-db container and removed the clair-db container.

### Dumping vulnerability data 

- Log in to the host (that is connected to Internet) where Clair database (Postgres) is running.
- Dump Clair's vulnerability database by the following commands, two files (`vulnerability.sql` and `clear.sql`) are generated:

_NOTE: The container name 'clair-db' is a placeholder for the db container used by the internet connected instance of clair_

```
  $ docker exec clair-db /bin/sh -c  "pg_dump -U postgres -a -t feature -t keyvalue -t namespace -t schema_migrations -t vulnerability -t vulnerability_fixedin_feature" > vulnerability.sql
  $ docker exec clair-db /bin/sh -c "pg_dump -U postgres -c -s" > clear.sql
```

### Back up Harbor's Clair database
Before importing the data, it is strongly recommended to back up the Clair database in Harbor.
```
 $ docker exec harbor-db /bin/sh -c  "pg_dump -U postgres -c" > all.sql
```

### Update Harbor's Clair database
Copy the `vulnerability.sql` and `clear.sql` to the host where Harbor is running on. Run the below commands to import the data to Harbor's Clair database:

```
 $ docker exec -i harbor-db psql -U postgres < clear.sql
 $ docker exec -i harbor-db psql -U postgres < vulnerability.sql
```

### Rescanning images
[...]

PR #6721, opened 2019-01-09
Updated clair db import for Harbor 1.6+

SDBrett:
Updated documentation for importing vulnerability data into the clair DB. 

Reference issue: #6714 

Signed-off-by: Brett Johnson <brett@sdbrett.com>

### DIFF
--- a/docs/import_vulnerability_data.md
+++ b/docs/import_vulnerability_data.md
@@ -2,13 +2,13 @@
 
 Harbor has integrated with Clair to scan vulnerabilities in images. When Harbor is installed in an environment without internet connection, Clair cannot fetch data from the public vulnerability database. Under this circumstance, Harbor administrator needs to manually update the Clair database.
 
-This document provides step-by-step instructions on updating Clair vulnerability database in Harbor v1.2.
+This document provides step-by-step instructions on updating Clair vulnerability database in Harbor.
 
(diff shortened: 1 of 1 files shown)


reasonerjt:
Thanks @SDBrett for the PR.

I don't think we need to keep the <= 1.6 as the doc should reflect the current version, for older version user can reference to the doc in other branches?

SDBrett:
I'll clean up the pre 1.6 items and place a reference to previous versions of the document.

docs/import_vulnerability_data.md
```diff
@@ -29,28 +29,39 @@ B. Check whether your Clair instance has already updated the vulnerability datab
     ```
 - The phrase "finished fetching" indicates that Clair has finished a round of vulnerability update from an endpoint. Please make sure all five endpoints (rhel, alpine, oracle, debian, ubuntu) are updated correctly. 
 
+## Harbor version < 1.6
+
+If you're using a version of Harbor prior to 1.6, you can access the correct instructions for your version using the following URL.
+https://github.com/goharbor/harbor/blob/v<VERSION NUMBER>/docs/import_vulnerability_data.md
```
ywk253100: The link isn't shown correctly after the MD file is rendered.

haisum:
These instructions are still invalid. Instead of referring to harbor-db, they refer to clair-db and the dump that they create in vulnerability.sql is only DDLs without any data.

PR #6737, opened 2019-01-11
Implement the unified health check API

ywk253100:
The commit implements an unified health check API for all Harbor services

### DIFF
--- a/docs/swagger.yaml
+++ b/docs/swagger.yaml
@@ -19,6 +19,18 @@ securityDefinitions:
 security:
   - basicAuth: []
 paths:
+  /health:
+    get:
+      summary: 'Health check API'
+      description: |
+        The endpoint returns the health stauts of the system.
+      tags:
+        - Products
+      responses:
+        '200':
+          description: The system health status.
+          schema:
+            $ref: '#/definitions/OverallHealthStatus'
   /search:
     get:
       summary: 'Search for projects, repositories and helm charts'
@@ -4514,3 +4526,27 @@ definitions:
     description: A list of label
     items:
       $ref: '#/definitions/Label'
+  OverallHealthStatus:
+    type: object
+    description: The system health status
+    properties:
+      status:
+        type: string
+        description: The overall health status. It is "healthy" only when all the components' status are "healthy" 
+      components:
+        type: array
+        items:
+          $ref: '#/definitions/ComponentHealthStatus'
+  ComponentHealthStatus:
+    type: object
+    description: The health status of component
+    properties:
+      name:
+        type: string
+        description: The component name 
+      status:
+        type: string
+        description: The health status of component
+      error:
+        type: string
+        description: (optional) The error message when the status is "unhealthy"
--- a/src/core/api/health.go
+++ b/src/core/api/health.go
@@ -1,0 +1,323 @@
+// Copyright 2019 Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+import (
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/goharbor/harbor/src/common/utils"
+
+	"github.com/goharbor/harbor/src/common/dao"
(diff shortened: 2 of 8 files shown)


docs/swagger.yaml
```diff
@@ -19,6 +19,18 @@ securityDefinitions:
...
 paths:
+  /health:
+    get:
+      summary: 'Health check API'
+      description: |
+        The endpoint returns the health stauts of the system.
+      tags:
+        - Products
+      responses:
+        '200':
```
wy65701436: Does it return only 200?

src/core/api/health.go
```diff
@@ -0,0 +1,273 @@
...
+
+func (u *updater) update(status error) {
+	u.Lock()
+	defer u.Unlock()
+
+	u.status = status
+}
+
+// PeriodicHealthChecker implements a Checker to check status periodically
+func PeriodicHealthChecker(checker health.Checker, period time.Duration) health.Checker {
```
wy65701436: can we record all of error messages instead of the last one?

src/core/api/health.go
```diff
@@ -0,0 +1,273 @@
...
+var healthCheckerRegistry = map[string]health.Checker{}
+
+// HealthAPI handles the request for "/api/health"
+type HealthAPI struct {
+	BaseController
+}
+
+// CheckHealth checks the health of system
+func (h *HealthAPI) CheckHealth() {
+	healthy := true
```
wy65701436: default is healthy?

docs/swagger.yaml
ywk253100: Yes, it returns only 200.

src/core/api/health.go
ywk253100: Why do we need to record all errors? Only the current error is meaningful.

src/core/api/health.go
ywk253100: Yes. If any of the components is unhealthy, the `healthy` will be set to `false`

src/core/api/health.go
```diff
@@ -0,0 +1,273 @@
...
+
+func coreHealthChecker() health.Checker {
+	return health.CheckFunc(func() error {
+		return nil
+	})
+}
+
+func portalHealthChecker() health.Checker {
+	url := config.GetPortalURL()
+	timeout := 60 * time.Second
```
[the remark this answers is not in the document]
ywk253100: The timeout and period may be different for each component. Do we need to expose them to the users?

src/core/api/health.go
```diff
@@ -0,0 +1,273 @@
...
+
+	go func() {
+		ticker := time.NewTicker(period)
+		for {
+			u.update(checker.Check())
+			<-ticker.C
+		}
+	}()
+
+	return u
```
steven-zou: Good code. I like it

src/core/api/health.go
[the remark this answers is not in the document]
steven-zou: Suggest define them as constants with comments

src/core/api/health.go
```diff
@@ -0,0 +1,273 @@
...
+		_, err = conn.Do("PING")
+		if err != nil {
+			return fmt.Errorf("failed to run \"PING\": %v", err)
+		}
+		return nil
+	})
+	return PeriodicHealthChecker(checker, period)
+}
+
+func registerHealthCheckers() {
```
steven-zou: Can we move the checker implementations to a separate file?

src/core/api/health.go
```diff
@@ -0,0 +1,273 @@
...
+type HealthAPI struct {
+	BaseController
+}
+
+// CheckHealth checks the health of system
+func (h *HealthAPI) CheckHealth() {
+	healthy := true
+	components := []map[string]string{}
+	for name, checker := range healthCheckerRegistry {
+		err := checker.Check()
```
reasonerjt: why not make it a-sync?

src/core/api/health.go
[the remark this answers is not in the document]
reasonerjt: In addition to let `checker` define its own timeout, it would be good if there can be a general timeout, just in case some checker takes longer than expected to return the status and the connection of the API is lost.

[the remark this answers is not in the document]
ywk253100: @steven-zou The `timeout` and `period` are the internal variables of the `Checker`, how to define them depends on the implementation of each `Checker`

ywk253100:
> how we handle if some service is not enabled by users?

@renmaosheng The API only checks and returns the status of services which are enabled.

src/core/api/health.go
```diff
@@ -42,35 +65,55 @@ type HealthAPI struct {
...
-		componentStatus := map[string]string{
-			"name":   name,
-			"status": healthyStatus(componentHealthy),
-		}
-		if !componentHealthy {
-			componentStatus["error"] = err.Error()
-			healthy = false
+		go check(name, checker, timeout, c)
+	}
+	for i := 0; i < len(healthCheckerRegistry); i++ {
```
steven-zou: Suggest to use buffer chan and WG

PR #6755, opened 2019-01-15
Refactor the prepare script

ninjadq:
1. Refactor prepare structure
2. change the template engine to jinja
3. change the config file format to yaml
4. fix `/data` dir hard code issue


make/harbor.yml
```diff
@@ -0,0 +1,141 @@
...
+#This attribute is for migrator to detect the version of the .cfg file, DO NOT MODIFY!
+_version: 1.7.0
+#The IP address or hostname to access admin UI and registry service.
+#DO NOT use localhost or 127.0.0.1, because Harbor needs to be accessed by external clients.
+#DO NOT comment out this line, modify the value of "hostname" directly, or the installation will fail.
+hostname: reg.mydomain.com
+
+#The protocol for accessing the UI and token/notification service, by default it is http.
+#It can be set to https if ssl is enabled on nginx.
+ui_url_protocol: https
```
ywk253100: The default value isn't consistent with the comment

make/harbor.yml
```diff
@@ -0,0 +1,141 @@
...
+#************************BEGIN INITIAL PROPERTIES************************
+
+##The initial password of Harbor admin, only works for the first time when Harbor starts. 
+#It has no effect after the first launch of Harbor.
+#Change the admin password from UI after launching Harbor.
+harbor_admin_password: Harbor12345
+
+##By default the auth mode is db_auth, i.e. the credentials are stored in a local database.
+#Set it to ldap_auth if you want to verify a user's credentials against an LDAP server.
+auth_mode: db_auth
```
ywk253100: I think the `auth_mode `, `ldap_basedn` and `ldap_group_gid` are all user settings?

make/photon/prepare/utils/docker_compose.py
```diff
@@ -0,0 +1,46 @@
+import os
+
+from g import templates_dir
+from .jinja import render_jinja
+
+
+# render docker-compose
+VERSION_TAG = 'dev'
+REGISTRY_VERSION = 'v2.7.1'
```
wy65701436: why hard-code the version here?

make/photon/prepare/utils/core.py
```diff
@@ -0,0 +1,56 @@
...
+        chart_cache_driver=chart_cache_driver,
+        **config_dict)
+
+    # Copy Core app.conf
+    copy_core_config(core_conf_template_path, core_conf)
+
+def prepare_core_config_dir():
+    prepare_config_dir(core_config_dir)
+
+def copy_core_config(core_templates_path, core_config_path):
```
wy65701436: can we add the underscore (_) if the method is private?

ninjadq: No OOP here, so is hard to define what is `private`

PR #6761, opened 2019-01-16
[DON'T MERGE]:remove pm dependence of token service

wy65701436:
To introduce robot account, it has to support docker login with the token, but without creating a project manager in the instance of the robot security context.

But, the token service depends on the pm to issue token to docker registry, this commit is to remove the dependence as it's not necessary.


reasonerjt:
The pm is not from security context, so it doesn't impact implementing the robot security context, however, it does seem that the pm here is unnecessary after the refactor for securityCtx, let's not merge it right now, and review it after the security Ctx refactor is finished.

PR #6764, opened 2019-01-16
Make db_name and sslmode configurable through cfg-file parameter

JarleB:
When connecting to external dabases where you only admin the database and not the complete db-server it is necessary to be able to configure the name of your database to connect  to. 

Also make "verify-full" default value in case one connect to external databases.


JarleB:
I see most tests failed. Unfortunately I'm not able to relate it back to the changes the PR introduces.

PR #6779, opened 2019-01-21
Add Can method to securty.Context interface

heww:

### DIFF
--- a/src/core/api/chart_repository_test.go
+++ b/src/core/api/chart_repository_test.go
@@ -8,6 +8,8 @@ import (
 
 	"github.com/goharbor/harbor/src/chartserver"
 	"github.com/goharbor/harbor/src/common/models"
+	"github.com/goharbor/harbor/src/common/rbac"
+	"github.com/goharbor/harbor/src/common/rbac/project"
 	"github.com/goharbor/harbor/src/core/promgr/metamgr"
 )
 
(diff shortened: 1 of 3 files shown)


src/core/api/chart_repository_test.go
```diff
@@ -311,39 +313,38 @@ func (msc *mockSecurityContext) IsSolutionUser() bool {
...
 
 // HasAllPerm returns whether the user has all permissions to the project
 func (msc *mockSecurityContext) HasAllPerm(projectIDOrName interface{}) bool {
 	return msc.HasReadPerm(projectIDOrName) && msc.HasWritePerm(projectIDOrName)
 }
 
+// Can returns whether the user can do action on resource
+func (msc *mockSecurityContext) Can(action rbac.Action, resource rbac.Resource) bool {
+	namespace, err := resource.GetNamespace()
+	if err == nil {
```
ninjadq: just a recommendation, if process`err != nil` first, will decrease the nested level of your main logic code

heww: `if err != nil` only means there is no namespace from the resource, it not means that user do not have action  permission  for the resource.  So here can not return `false` if no namespace found from resource.

Of course, it can return `false` when no namespace found in this `mockSecurityContext`.

heww: improved

PR #6781, opened 2019-01-21
User permissions api

heww:


src/common/rbac/project/const.go
```diff
@@ -20,14 +20,37 @@ import (
...
 const (
-	ActionAll      = rbac.Action("*")
-	ActionPull     = rbac.Action("pull")
-	ActionPush     = rbac.Action("push")
-	ActionPushPull = rbac.Action("push+pull")
+	ActionAll                 = rbac.Action("*")
+	ActionPull                = rbac.Action("pull")
+	ActionPush                = rbac.Action("push")
+	ActionPushPull            = rbac.Action("push+pull")
+	ActionCreate              = rbac.Action("create")
```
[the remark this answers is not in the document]
heww: Action is just a string what you like to define for resource

src/common/rbac/project/visitor_role.go
```diff
@@ -22,18 +22,116 @@ import (
...
+			{Resource: ResourceHelmChart, Action: ActionDownload},
+			{Resource: ResourceHelmChart, Action: ActionList},
+
+			{Resource: ResourceHelmChartVersion, Action: ActionDownload},
+			{Resource: ResourceHelmChartVersion, Action: ActionRead},
+			{Resource: ResourceHelmChartVersion, Action: ActionList},
+			{Resource: ResourceHelmChartVersion, Action: ActionAddLabel},
+			{Resource: ResourceHelmChartVersion, Action: ActionRemoveLabel},
+
+			{Resource: ResourceConfiguration, Action: ActionRead},
```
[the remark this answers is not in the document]
heww: policies for robot added

docs/swagger.yaml
```diff
@@ -717,6 +717,35 @@ paths:
...
+        - name: scope
+          in: query
+          type: string
+          required: false
+          description: Get permissions of the scope
+        - name: relative
+          in: query
+          type: boolean
+          required: false
+          description: If true, the resources in the response are relative to the scope
```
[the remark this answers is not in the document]
heww: add eg for this param

src/common/rbac/project/const.go
```diff
@@ -20,14 +20,57 @@ import (
...
+
+	// attach label to repository tag (aka, image tag) or helm chart version
+	ActionAddLabel = rbac.Action("add-label")
+	// detach label to repository tag (aka, image tag) or helm chart version
+	ActionRemoveLabel = rbac.Action("remove-label")
+
+	// upload helm chart or helm chart version
+	ActionUpload = rbac.Action("upload")
+	// download helm chart or helm chart version
+	ActionDownload = rbac.Action("download")
```
ywk253100: Why we need so many actions? I think most of them are CRUD? 
The `ActionPull `, `ActionRead ` and `ActionDownload ` are all `Read`?
`ActionReTag ` is the combination of `Read` source `Image` and `Write` destination `Image`?

src/common/rbac/project/util.go
```diff
@@ -47,7 +115,22 @@ func policiesForPublicProject(namespace rbac.Namespace) []*rbac.Policy {
 func policiesForSystemAdmin(namespace rbac.Namespace) []*rbac.Policy {
```
ywk253100: What is the different with `GetAllPolicies ()`?

src/common/rbac/project/util.go
```diff
@@ -21,12 +21,80 @@ import (
...
+		{Resource: ResourceHelmChartVersion, Action: ActionRead},
+		{Resource: ResourceHelmChartVersion, Action: ActionList},
 	}
 
-	// subresource policies for system admin visitor
-	systemAdminProjectPolicies = []*rbac.Policy{
-		{Resource: ResourceAll, Action: ActionAll},
+	// all policies for the projects
+	allPolicies = []*rbac.Policy{
+		{Resource: ResourceSelf, Action: ActionRead},
```
ywk253100: Can we use the `Resource: *, Action *` to represent the all?

src/core/api/user.go
```diff
@@ -339,6 +341,57 @@ func (ua *UserAPI) ToggleUserAdminRole() {
 	}
 }
 
+// ListUserPermissions handles GET to /api/users/{}/permissions
+func (ua *UserAPI) ListUserPermissions() {
+	if ua.userID != ua.currentUserID {
+		log.Errorf("Current user, id: %d can not view other user's permissions", ua.currentUserID)
```
ywk253100: Use debug rather than error to log the message, or this will make users confuse when debugging the system.

src/core/api/user.go
```diff
@@ -339,6 +341,57 @@ func (ua *UserAPI) ToggleUserAdminRole() {
...
+		return
+	}
+
+	relative := ua.Ctx.Input.Query("relative") == "true"
+
+	scope := rbac.Resource(ua.Ctx.Input.Query("scope"))
+	policies := []*rbac.Policy{}
+
+	namespace, err := scope.GetNamespace()
+	if err == nil {
```
ywk253100: What about the err != nil ?

src/common/rbac/project/const.go
heww: `ActionPull` is used by docker client, `ActionDownload` now changed to `ActionRead`, `ActionReTag` is removed.

src/common/rbac/project/util.go
heww: `policiesForSystemAdmin` removed

src/core/api/user.go
heww: changed to `Warningf`, which is used by `ToggleUserAdminRole` method

src/core/api/user.go
heww: `err == nil` is ok here, otherwise the code will be like this

```go
	namespace, err := scope.GetNamespace()
	if err != nil {
		ua.Data["json"] = []
		ua.ServeJSON()
		return
	}

	switch namespace.Kind() {
	case "project":
		for _, policy := range project.GetAllPolicies(namespace) {
			if ua.SecurityCtx.Can(policy.Action, policy.Resource) {
				policies = append(policies, policy)
			}
		}
	}

	results := []map[string]string{}
	for _, policy := range policies {
		var resource rbac.Resource

		// for resource `/project/1/repository` if `relative` is `true` then the resource in response will be `repository`
		if relative {
			relativeResource, err := policy.Resource.RelativeTo(scope)
			if err != nil {
				continue
			}
			resource = relativeResource
		} else {
			resource = policy.Resource
		}

		results = append(results, map[string]string{
			"resource": resource.String(),
			"action":   policy.Action.String(),
		})
	}

	ua.Data["json"] = results
	ua.ServeJSON()
	return
```

the following block is repeated
```go
	ua.Data["json"] = results
	ua.ServeJSON()
	return
```

src/common/rbac/project/util.go
heww: all policies for project shoud be list here, so that in `/users/current/permissions` we can use `security.Context` Can method to check user's permissions

PR #6782, opened 2019-01-22
Refine the replication job

ywk253100:
1. Refine the replication job
2. Define the Transfer interface
3. Define the common models

### DIFF
--- a/src/jobservice/job/impl/replication/ng/replication.go
+++ b/src/jobservice/job/impl/replication/ng/replication.go
@@ -1,0 +1,93 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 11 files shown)


cd1989:
lgtm

src/jobservice/job/impl/replication/ng/transfer/transfer.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+// CancelFunc is a function used to check whether the task is cancelled
+type CancelFunc func() bool
+
+// Transfer defines an interface used to transfer the resource
+// from source registry to the destination.The function "isCancelled"
+// can be used to check whether the task has been cancelled during
+// the processing progress
+type Transfer interface {
```
steven-zou:
```go
type TransferHandlerFactory func(logger.Interface, CancelFunc)(Transfer,error)

type TransferHandler interface{
  Transfer(res *model.Resource, dest *model.Resource) error
}

transferRegistry = map[model.ResourceType]TransferHandlerFactory{}

## when using
handlerFac, err := transfer.Get(res.Type)
#...
hander, err:=handlerFac(jLogger, cancelFunc)
if err!=nil {
  return err
}

err=hander.Transfer(src,dest)

```

kofj:
I think we should rename the models folder **src/replication/ng/model/** to **src/replication/ng/models/**

PR #6783, opened 2019-01-22
Fix Swagger client generation for a few chartrepo endpoints

protochron:
It looks like the Swagger for `ChartVersions` and `ChartVersionDetails`
were missing fields to properly generate a client.

This fixes the problem by adding the correct `refs` so that the client
produces an output when making an API call.

In Go, this generates something like:

```Go
/*GetChartrepoRepoChartsNameVersionOK handles this case with default header values.

Successfully retrieved the chart version
*/
type GetChartrepoRepoChartsNameVersionOK struct {
	Payload *models.ChartVersionDetails
}

...

/*GetChartrepoRepoChartsNameOK handles this case with default header values.

Retrieved all versions of the specified chart
*/
type GetChartrepoRepoChartsNameOK struct {
	Payload models.ChartVersions
}
```

Signed-off-by: Dan Norris <dan.norris@netapp.com>

### DIFF
--- a/docs/swagger.yaml
+++ b/docs/swagger.yaml
@@ -2851,7 +2851,9 @@ paths:
           description: The chart name
       responses:
         '200':
-          $ref: '#/definitions/ChartVersions'
+          description: Retrieved all versions of the specified chart
+          schema:
+            $ref: '#/definitions/ChartVersions'
         '401':
           $ref: '#/definitions/UnauthorizedChartAPIError'
         '403':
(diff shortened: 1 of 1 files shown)


sdake:
@reasonerjt what is needed to get this PR moving?

Cheers
-steve

steven-zou:
@protochron 

Sorry for the late review. Seems everything is ok, I'll merge this PR now. Thanks again for your contributions to Harbor! Looking forward to more contributions and doing more collaborations in the future!

PR #6820, opened 2019-01-24
Add robot account authn & authz implementation

wy65701436:
This commit is to add the jwt token service, and do the authn & authz for robot account.


src/common/models/robot.go
```diff
@@ -45,10 +46,10 @@ type RobotQuery struct {
...
 // RobotReq ...
 type RobotReq struct {
-	Name        string             `json:"name"`
-	Description string             `json:"description"`
-	Disabled    bool               `json:"disabled"`
-	Access      []*ResourceActions `json:"access"`
+	Name        string         `json:"name"`
+	Description string         `json:"description"`
+	Disabled    bool           `json:"disabled"`
+	Policy      []*rbac.Policy `json:"access"`
```
reasonerjt: We call it `policy` in the struct but call it `access` in the request body, any reason for this inconsistency?

src/common/rbac/project/util.go
```diff
@@ -30,6 +36,20 @@ var (
 	}
 )
 
+func policiesForPublicProjectRobot(namespace rbac.Namespace) []*rbac.Policy {
```
reasonerjt: `policiesForPublicProjectRobot`,`policiesForPublicProject`,`policiesForSystemAdmin `

The code looks extremely repetitive.

src/common/token/htoken.go
```diff
@@ -0,0 +1,78 @@
...
+	"crypto/ecdsa"
+	"crypto/rsa"
+	"errors"
+	"fmt"
+	"github.com/dgrijalva/jwt-go"
+	"github.com/goharbor/harbor/src/common/utils/log"
+	"time"
+)
+
+// HToken ...
```
reasonerjt: what is an HToken?  Please add more comment.

src/core/filter/security.go
```diff
@@ -147,6 +153,50 @@ func (s *secretReqCtxModifier) Modify(ctx *beegoctx.Context) bool {
...
 }
 
+type robotAuthReqCtxModifier struct{}
+
+func (r *robotAuthReqCtxModifier) Modify(ctx *beegoctx.Context) bool {
+	robotName, robotTk, ok := ctx.Request.BasicAuth()
+	if !ok {
+		return false
+	}
+	log.Debug("got robot information via token auth")
```
reasonerjt: This is misleading, this information will be printed even when the request contains basic auth info for a regular user.

src/core/filter/security.go
```diff
@@ -147,6 +153,50 @@ func (s *secretReqCtxModifier) Modify(ctx *beegoctx.Context) bool {
...
+	}
+	log.Infof(fmt.Sprintf("got robot token header, %v", htk.Header))
+	// Do authn for robot account, as Harbor only stores the token ID, just validate the ID and disable.
+	robot, err := dao.GetRobotByID(htk.Claims.(*token.RobotClaims).TokenID)
+	if err != nil {
+		log.Errorf("failed to get robot %s: %v", robotName, err)
+		return false
+	}
+	if robot == nil {
+		log.Error("the token is not valid.")
```
reasonerjt: It should be the ID of robot does not exist.

src/core/filter/security.go
```diff
@@ -147,6 +153,50 @@ func (s *secretReqCtxModifier) Modify(ctx *beegoctx.Context) bool {
...
+	robotName, robotTk, ok := ctx.Request.BasicAuth()
+	if !ok {
+		return false
+	}
+	log.Debug("got robot information via token auth")
+	if !strings.HasPrefix(robotName, common.RobotPrefix) {
+		return false
+	}
+	rClaims := &token.RobotClaims{}
+	htk := &token.HToken{}
```
reasonerjt: This line is not needed

src/common/token/claims.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"github.com/dgrijalva/jwt-go"
+	"github.com/goharbor/harbor/src/common/rbac"
+	"github.com/pkg/errors"
+)
+
+// RobotClaims implements the interface of jwt.Claims
+type RobotClaims struct {
+	jwt.StandardClaims
+	TokenID   int64          `json:"id"`
```
reasonerjt: I think we need to revisit the structure of the token to make it more extensible, but this should not block this PR.

src/common/rbac/project/util.go
reasonerjt: not needed.

src/common/rbac/project/robot.go
```diff
@@ -0,0 +1,61 @@
...
+	// Index whether the robot is authenticated
+	IsAuthenticated() bool
+	// GetUsername returns the name of robot
+	GetUsername() string
+	// GetPolicy get the rbac policies from security context
+	GetPolicies() []*rbac.Policy
+}
+
+// robot implement the rbac.User interface for project robot account
+type robot struct {
```
reasonerjt: This struct can be moved to the package where robot security context sits, and the robotContext is not needed.  We can reuse the robot security context in the struct.

src/common/token/htoken.go
```diff
@@ -0,0 +1,78 @@
...
+type HToken struct {
+	jwt.Token
+}
+
+// NewWithClaims ...
+func NewWithClaims(claims *RobotClaims) *HToken {
```
reasonerjt: I'd prefer the claims be transparent to the api package so the api code just call NewRobotToken(id, projectID...)

src/common/token/htoken.go
```diff
@@ -0,0 +1,78 @@
...
+			Issuer:    DefaultOptions.Issuer,
+		},
+	}
+	return &HToken{
+		Token: *jwt.NewWithClaims(DefaultOptions.SignMethod, rClaims),
+	}
+}
+
+// SignedString get the SignedString.
+func (htk *HToken) SignedString() (string, error) {
```
reasonerjt: This is essentially getting the raw token string, so maybe calling it `Raw`?

src/common/token/htoken.go
```diff
@@ -0,0 +1,78 @@
+package token
```
reasonerjt: Seems there should has a way to check the whether the token is expired?

src/core/filter/security.go
reasonerjt: misleading log, remove it.

src/core/filter/security.go
reasonerjt: Should we check the expiration in this func?

src/core/filter/security.go
reasonerjt: Misleading log?

src/core/filter/security.go
```diff
@@ -147,6 +153,50 @@ func (s *secretReqCtxModifier) Modify(ctx *beegoctx.Context) bool {
...
+	htk, err := token.ParseWithClaims(robotTk, rClaims)
+	if err != nil {
+		log.Errorf("failed to decrypt robot token, %v", err)
+		return false
+	}
+	log.Infof(fmt.Sprintf("got robot token header, %v", htk.Header))
```
reasonerjt: Why print this log message?

heww:
lgtm

PR #6833, opened 2019-01-28
Implement registries manager and APIs for NG replication

cd1989:
- Implement registries manager
- Add new registry management APIs


make/migrations/postgresql/0005_ng_replication.up.sql
```diff
@@ -0,0 +1,18 @@
...
+ credential_type varchar(16),
+ access_key varchar(128),
+ access_secret varchar(1024),
+ type varchar(32),
+ insecure boolean,
+ description varchar(1024),
+ health varchar(16),
+ creation_time timestamp default CURRENT_TIMESTAMP,
+ update_time timestamp default CURRENT_TIMESTAMP,
+ CONSTRAINT unique_url UNIQUE (url)
```
ywk253100: The unique is not necessary, uses can create two registries with the same url but different credentials. The name should be unique

cd1989: OK, I will change it.

src/core/api/registry.go
```diff
@@ -0,0 +1,210 @@
...
+		t.HandleUnauthorized()
+		return
+	}
+
+	if !t.SecurityCtx.IsSysAdmin() {
+		t.HandleForbidden(t.SecurityCtx.GetUsername())
+		return
+	}
+
+	t.manager = registry.NewDefaultManager()
```
ywk253100: How about just make a single instance for `Manager`? The `Prepare` will be called for every requests. And I think we will init an global `Manager` in replication model, we can use that one

src/core/api/registry.go
```diff
@@ -0,0 +1,210 @@
...
+		log.Errorf("failed to get registry %s: %v", registry.Name, err)
+		t.CustomAbort(http.StatusInternalServerError, http.StatusText(http.StatusInternalServerError))
+	}
+
+	if reg != nil {
+		t.HandleConflict(fmt.Sprintf("name '%s' is already used", registry.Name))
+		return
+	}
+
+	reg, err = t.manager.GetByURL(registry.URL)
```
ywk253100: Remove the logic as we don't need unique url

src/core/api/registry.go
```diff
@@ -0,0 +1,210 @@
...
+			t.CustomAbort(http.StatusInternalServerError, http.StatusText(http.StatusInternalServerError))
+		}
+
+		if reg != nil {
+			t.HandleConflict("name is already used")
+			return
+		}
+	}
+
+	if registry.URL != originalURL {
```
ywk253100: Remove this one either

src/core/api/registry.go
```diff
@@ -0,0 +1,210 @@
...
+func (t *RegistryAPI) Get() {
+	id := t.GetIDFromURL()
+
+	registry, err := t.manager.Get(id)
+	if err != nil {
+		log.Errorf("failed to get registry %d: %v", id, err)
+		t.CustomAbort(http.StatusInternalServerError, http.StatusText(http.StatusInternalServerError))
+	}
+
+	if registry == nil {
```
ywk253100: The registry will not be nil as the manager will return an error if it cannot get an valid registry from database? Maybe we need the manager returns nil when it cannot get the registry to handle the 404 error in API?

src/replication/ng/registry/manager.go
```diff
@@ -15,20 +15,335 @@
...
+
+// Remove deletes a registry
+func (m *DefaultManager) Remove(id int64) error {
+	policies, err := dao.GetRepPolicyByTarget(id)
+	if err != nil {
+		log.Errorf("Get policies related to registry %d error: %v", id, err)
+		return err
+	}
+
+	if len(policies) > 0 {
```
ywk253100: I think the logic should be handle in the API level and use the global PolicyManager, it will be defined in the replication module

src/core/api/registry.go
cd1989: You are right.

src/core/api/registry.go
cd1989: I will return specific error to indicate not found.

src/replication/ng/registry/manager.go
cd1989: That's better.

src/replication/ng/registry/manager.go
```diff
@@ -15,20 +15,326 @@
...
 	// Add new registry
 	Add(*model.Registry) (int64, error)
 	// List registries, returns total count, registry list and error
 	List(...*model.RegistryQuery) (int64, []*model.Registry, error)
 	// Get the specified registry
 	Get(int64) (*model.Registry, error)
+	// GetByName gets registry by name
+	GetByName(name string) (*model.Registry, error)
+	// GetByURL gets registry by its URL
+	GetByURL(url string) (*model.Registry, error)
```
ywk253100: The function isn't needed anymore?

src/replication/ng/registry/manager.go
```diff
@@ -15,20 +15,326 @@
...
+
+// Get gets a registry by id
+func (m *DefaultManager) Get(id int64) (*model.Registry, error) {
+	registry, err := dao.GetRegistry(id)
+	if err != nil {
+		return nil, err
+	}
+
+	if registry == nil {
+		return nil, utilerr.KnownError{
```
ywk253100: This isn't consistent with the logic in `GetByName`, I think the logic in `GetByName` is better, it's easy for API level to handle 404.

src/replication/ng/registry/manager.go
```diff
@@ -15,20 +15,310 @@
...
+		return nil, nil
+	}
+
+	return fromDaoModel(registry)
+}
+
+// List lists registries according to query provided.
+func (m *DefaultManager) List(query ...*model.RegistryQuery) (int64, []*model.Registry, error) {
+	var registryQueries []*dao.ListRegistryQuery
+	for _, q := range query {
```
ywk253100: Let's only take the query[0] as the query param

cd1989: Fixed.

src/replication/ng/registry/manager.go
cd1989: Deleted

src/common/models/registry.go
```diff
@@ -0,0 +1,58 @@
...
+		// Prevent SSRF security issue #3755
+		r.URL = url.Scheme + "://" + url.Host + url.Path
+		if len(r.URL) > 64 {
+			v.SetError("endpoint", "max length is 64")
+		}
+	}
+
+	// password is encoded using base64, the length of this field
+	// in DB is 64, so the max length in request is 48
+	if len(r.AccessSecret) > 48 {
```
ywk253100: I don't think the 48 is long enough to store the secret, the length in database is 1024, why we use 48 here?

src/common/dao/registry.go
```diff
@@ -0,0 +1,107 @@
+package dao
```
ywk253100: As we discussed, the dao methods should be put under directory `replication/ng`, can we move them in this PR?

cd1989: Moved.

src/core/api/registry.go
```diff
@@ -0,0 +1,201 @@
...
+// Put updates a registry
+func (t *RegistryAPI) Put() {
+	id := t.GetIDFromURL()
+
+	registry, err := t.manager.Get(id)
+	if err != nil {
+		log.Errorf("Get registry by id %d error: %v", id, err)
+		t.CustomAbort(http.StatusInternalServerError, http.StatusText(http.StatusInternalServerError))
+	}
+
```
ywk253100: Handler 404?

src/core/api/registry.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+}
+
+// Delete deletes a registry
+func (t *RegistryAPI) Delete() {
+	id := t.GetIDFromURL()
+
+	_, err := t.manager.Get(id)
+	if err != nil {
+		if utilerr.Is(err, utilerr.ReasonNotFound) {
```
ywk253100: I think this logic isn't needed anymore? and we need to handle 404 by checking the registry returned?

src/core/api/registry.go
```diff
@@ -0,0 +1,201 @@
...
+		msg := fmt.Sprintf("Get policies related to registry %d error: %v", id, err)
+		log.Error(msg)
+		t.HandleInternalServerError(msg)
+		return
+	}
+
+	if len(policies) > 0 {
+		msg := fmt.Sprintf("Can't delete registry with replication policies, %d found", len(policies))
+		log.Error(msg)
+		t.HandleInternalServerError(msg)
```
ywk253100: Should be `http.StatusPreconditionFailed` rather than 500

src/core/api/registry.go
```diff
@@ -0,0 +1,201 @@
...
+			return
+		}
+
+		msg := fmt.Sprintf("Get registry %d error: %v", id, err)
+		log.Error(msg)
+		t.HandleInternalServerError(msg)
+		return
+	}
+
+	policies, err := dao.GetRepPolicyByTarget(id)
```
ywk253100: Add a TODO here to replace it with the method of `PolicyManager`

src/core/api/registry.go
```diff
@@ -0,0 +1,201 @@
...
+
+// Prepare validates the user
+func (t *RegistryAPI) Prepare() {
+	t.BaseController.Prepare()
+	if !t.SecurityCtx.IsAuthenticated() {
+		t.HandleUnauthorized()
+		return
+	}
+
+	if !t.SecurityCtx.IsSysAdmin() && t.Ctx.Request.Method != "GET" {
```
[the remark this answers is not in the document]
cd1989: Normal user should have 'GET' registry permission, in my understanding

src/core/api/registry.go
cd1989: My mistake

src/core/api/registry.go
cd1989: Indeed

src/core/api/registry.go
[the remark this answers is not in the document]
ywk253100: Only system admin can operate the registries.

src/replication/ng/registry/manager.go
```diff
@@ -15,20 +15,306 @@
...
+
+	if registry == nil {
+		return nil, nil
+	}
+
+	return fromDaoModel(registry)
+}
+
+// List lists registries according to query provided.
+func (m *DefaultManager) List(query ...*model.RegistryQuery) (int64, []*model.Registry, error) {
```
wy65701436: what's the purpose to return an int? It seems that the caller doesn't consume it.

cd1989: Total count

src/core/api/models/registry.go
```diff
@@ -0,0 +1,11 @@
+package models
+
+// RegistryUpdateRequest is request used to update a registry.
+type RegistryUpdateRequest struct {
+	Name           *string `json:"name"`
```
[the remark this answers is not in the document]
cd1989: To judge whether user has set this or not, if not set, we won't update it.

src/core/api/registry.go
```diff
@@ -0,0 +1,213 @@
...
+		return
+	}
+
+	if registry == nil {
+		t.HandleNotFound(fmt.Sprintf("registry %d not found", id))
+		return
+	}
+
+	// TODO: Use PolicyManager instead
+	policies, err := dao.GetRepPolicyByTarget(id)
```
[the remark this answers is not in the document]
cd1989: I didn't change policy related code, it's `GetRepPolicyByTarget ` there.

src/replication/ng/registry/manager.go
```diff
@@ -15,20 +15,306 @@
...
 	// Update the registry, the "props" are the properties of registry
 	// that need to be updated
 	Update(registry *model.Registry, props ...string) error
 	// Remove the registry with the specified ID
 	Remove(int64) error
+	// HealthCheck checks health status of all registries and update result in database
+	HealthCheck() error
+}
+
+// DefaultManager implement the Manager interface
```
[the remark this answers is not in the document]
cd1989: It's already in the `registry` package, I think it's go's style to avoid duplication.

src/replication/ng/registry/manager.go
```diff
@@ -15,20 +15,306 @@
...
+	err = registry.Ping()
+	if err != nil {
+		return Unhealthy, err
+	}
+
+	return Healthy, nil
+}
+
+// decrypt checks whether access secret is set in the registry, if so, decrypt it.
+func decrypt(registry *model.Registry) error {
```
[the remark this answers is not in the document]
cd1989: no special reason, once they are registry, I found it's not convient for encrypt later, so I change it. I will update it.

PR #6842, opened 2019-01-29
enable adminservice https if configured

danielzhanghl:
Signed-off-by: daniel zhang <danielzhanghl@gmail.com>

currently all other harbor component could be configured with https except adminserver, this code is to enable the https if configured cert and private key file.

### DIFF
--- a/src/adminserver/main.go
+++ b/src/adminserver/main.go
@@ -15,6 +15,7 @@
 package main
 
 import (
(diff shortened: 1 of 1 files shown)


danielzhanghl:
hi harbor team, 
this is small change to enable adminserver https if configured, I tried to enable harbor https internally, and found only this one does  not support https.
BR.

PR #6853, opened 2019-01-29
Remove password hash from user objects when listing

### LINKED ISSUE #6838: /api/users list exposes password hashes as part of output
It seems like there's a difference between calling `/api/users` and `/api/users/{id}` where the former will actually return the salted password hash as part of the call:
```bash
curl -u admin:$password http://harbor.default.svc/api/users 2>/dev/null | jq
[
  {
    "user_id": 3,
    "username": "test-user",
    "email": "test@example.com",
    "password": "5f0312458e714970e39b9a45a4eccc74",
    "realname": "Test User",
    "comment": "",
    "deleted": false,
    "role_name": "",
    "role_id": 0,
    "has_admin_role": false,
    "reset_uuid": "",
    "creation_time": "2019-01-28T22:51:35Z",
    "update_time": "2019-01-28T22:51:35Z"
  }
```
vs.
```bash
/ # curl -u admin:$password http://harbor.default.svc/api/users/3 2>/dev/null | jq
{
  "user_id": 3,
  "username": "test-user",
  "email": "test@example.com",
  "password": "",
  "realname": "Test User",
  "comment": "",
  "deleted": false,
  "role_name": "",
  "role_id": 0,
  "has_admin_role": false,
  "reset_uuid": "",
  "creation_time": "2019-01-28T22:51:35Z",
  "update_time": "2019-01-28T22:51:35Z"
}
```

Note how `password` in the first call is set to a value while it is empty in the second call.

I would have expected these two calls to produce identical output to populate the user object. I'm assuming that the salted password hash is supposed to be removed from `UserAPI.List()` just like it is in `UserAPI.Get()`: https://github.com/goharbor/harbor/blob/master/src/core/api/user.go#L120.

The fix is fairly straightforward and I'm happy to contribute a fix, but I'd like to make sure that the password hashes aren't being exposed intentionally (for some reason!) before I do.

See above for curl commands to reproduce.

**Versions:**
- harbor version: [1.7.1]

protochron:
The output from `/api/users` includes the full password hash that is
stored in the database for all users.

This fixes the problem by sanitizing the password field in the list of
user structs before passing it to the JSON handler.

Fixes #6838

### DIFF
--- a/src/core/api/user.go
+++ b/src/core/api/user.go
@@ -161,6 +161,10 @@ func (ua *UserAPI) List() {
 		return
 	}
 
+	for _, user := range users {
+		user.Password = ""
+	}
+
 	ua.SetPaginationHeader(total, page, size)
 	ua.Data["json"] = users
 	ua.ServeJSON()
--- a/src/core/api/user_test.go
+++ b/src/core/api/user_test.go
@@ -201,6 +201,7 @@ func TestUsersGet(t *testing.T) {
 		assert.Equal(200, code, "Get users status should be 200")
 		assert.Equal(1, len(users), "Get users record should be 1 ")
 		testUser0002ID = users[0].UserId
+		assert.Equal("", users[0].Password)
 	}
 }
 


protochron:
The test I wrote for this is failing at the moment since I'm working on getting a test environment going locally. I'll post an update when I get tests written to cover this case and a test to ensure that `api/users/{id}` doesn't hit the same issue.

wy65701436:
@protochron could you please fix the CI failure? I will merge the PR into master as we're closing release.

-

wy65701436:
hi @protochron , thanks for catching the issue and appreciate the effort on providing the fix.

As the release schedule, we have to make the fix in v1.8.0. The commit is https://github.com/goharbor/harbor/pull/7707, let me close it as fixed.  

We will put more effort into review  PR from the community, thanks again for your contribution.

PR #6858, opened 2019-01-30
Implement the replication flow controller

ywk253100:
1. Implement the replication flow controller
2. Implement the registry adapter management
3. Define the interface for PolicyManager, RegistryManager, ExecutionManager, TaskManager and Scheduler

### DIFF
--- a/src/replication/ng/adapter/adapter.go
+++ b/src/replication/ng/adapter/adapter.go
@@ -1,0 +1,80 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package adapter
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/goharbor/harbor/src/replication/ng/model"
+)
+
+var registry = map[model.RegistryType]Factory{}
+
+// Factory creates a specific Adapter according to the params
+type Factory func(*model.Registry) (Adapter, error)
+
+// Info provides base info and capability declarations of the adapter
+type Info struct {
+	Name                     model.RegistryType   `json:"name"`
+	Description              string               `json:"description"`
+	SupportedResourceTypes   []model.ResourceType `json:"supported_resource_types"`
+	SupportedResourceFilters []model.FilterType   `json:"supported_resource_filters"`
+}
+
+// Adapter interface defines the capabilities of registry
+type Adapter interface {
+	// Info return the information of this adapter
+	Info() *Info
+	// Lists the available namespaces under the specified registry with the
+	// provided credential/token
+	ListNamespaces(*model.NamespaceQuery) ([]*model.Namespace, error)
+	// Create a new namespace
+	// This method should guarantee it's idempotent
+	// And returns nil if a namespace with the same name already exists
+	CreateNamespace(*model.Namespace) error
+	// Get the namespace specified by the name, the returning value should
+	// contain the metadata about the namespace if it has
+	GetNamespace(string) (*model.Namespace, error)
+	// Fetch the content resource under the namespace by filters
+	// SUGGESTION: Adapter provider can do their own filter based on the filter pattern
+	// or call the default `DoFilter` function of the filter to complete resource filtering.
+	FetchResources(namespace []string, filters []*model.Filter) ([]*model.Resource, error)
+}
+
+// RegisterFactory registers one adapter factory to the registry
+func RegisterFactory(name model.RegistryType, factory Factory) error {
+	if !name.Valid() {
+		return errors.New("invalid adapter factory name")
+	}
+	if factory == nil {
+		return errors.New("empty adapter factory")
+	}
+	if _, exist := registry[name]; exist {
+		return fmt.Errorf("adapter factory for %s already exists", name)
+	}
+	registry[name] = factory
+	return nil
+}
+
+// GetFactory gets the adapter factory by the specified name
+func GetFactory(name model.RegistryType) (Factory, error) {
+	factory, exist := registry[name]
+	if !exist {
+		return nil, fmt.Errorf("adapter factory for %s not found", name)
+	}
+
+	return factory, nil
+}
--- a/src/replication/ng/flow/controller.go
+++ b/src/replication/ng/flow/controller.go
@@ -1,0 +1,99 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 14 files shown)


src/replication/ng/adapter/adapter.go
```diff
@@ -0,0 +1,83 @@
...
+	// Fetch the content resource under the namespace by filters
+	// SUGGESTION: Adapter provider can do their own filter based on the filter pattern
+	// or call the default `DoFilter` function of the filter to complete resource filtering.
+	FetchResources(namespace string, filters []*model.Filter) ([]*model.Resource, error)
+}
+
+// Register one adapter to the registry, the param "adapter" is the type
+// of the adapter that you want to register
+func Register(name model.RegistryType, adapter interface{}) error {
+	if len(name) == 0 {
```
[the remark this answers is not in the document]
steven-zou: Not changed?

src/replication/ng/flow/flow.go
```diff
@@ -0,0 +1,229 @@
...
+		return 0, fmt.Errorf("failed to create the execution record for replication based on policy %d: %v", policy.ID, err)
+	}
+	log.Debugf("an execution record for replication based on the policy %d created: %d", policy.ID, executionID)
+
+	// initialize
+	_, dstReg, srcAdapter, dstAdapter, err := initialize(policy, d.registryMgr)
+	if err != nil {
+		err = fmt.Errorf("failed to initialize the replication based on policy %d: %v", policy.ID, err)
+		log.Error(err.Error())
+		markExecutionFailure(d.executionMgr, executionID, err)
```
steven-zou: Can we define related state changing methods on the execution itself?

```go
func (e *execution) Failed() error{}

func (e *execution) Succeed() error{}

...
```

src/replication/ng/flow/flow.go
```diff
@@ -0,0 +1,229 @@
...
+
+// defaultController is the default implement for the Controller
+type defaultController struct {
+	registryMgr  registry.Manager
+	executionMgr execution.Manager
+	scheduler    scheduler.Scheduler
+}
+
+// Replicate according the to policy ID
+func (d *defaultController) StartReplication(policy *model.Policy) (int64, error) {
```
steven-zou: Suggest to split the method into multi meaningful stages and then reduce the combined methods like `initialize` and `createNamespace` etc.. The body of this method is too large.

steven-zou: Can we create a flow object to cover those stages?

src/replication/ng/scheduler/scheduler.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"github.com/goharbor/harbor/src/replication/ng/model"
+)
+
+// Scheduler schedules tasks to transfer resource data
+type Scheduler interface {
+	// Schedule tasks for one execution
+	Schedule([]*model.Resource, []*model.Resource) ([]*model.Task, error)
+	// Stop the task specified by ID
+	StopTask(id string) error
```
steven-zou: Is it conflicted with PR https://github.com/goharbor/harbor/pull/6886? Can we use just `Stop` as the method name?

src/replication/ng/flow/controller.go
```diff
@@ -0,0 +1,99 @@
...
+	StartReplication(policy *model.Policy) (int64, error)
+	// Stop the replication specified by the execution ID
+	StopReplication(int64) error
+}
+
+// NewController returns an instance of a Controller
+func NewController(registryMgr registry.Manager,
+	executionMgr execution.Manager, scheduler scheduler.Scheduler) (Controller, error) {
+	if registryMgr == nil || executionMgr == nil || scheduler == nil {
+		return nil, errors.New("invalid params")
```
steven-zou: The error is not accurate, seems a little general!

src/replication/ng/model/execution.go
```diff
@@ -16,17 +16,38 @@ package model
...
-	Failed     int       `json:"failed"`
-	Succeed    int       `json:"succeed"`
-	Pending    int       `json:"pending"`
-	InProgress int       `json:"in_progress"`
-	StartTime  time.Time `json:"start_time"`
-	EndTime    time.Time `json:"end_time"`
+	ID         int64           `json:"id"`
+	PolicyID   int64           `json:"policy_id"`
+	Status     ExecutionStatus `json:"status"`
+	StatusText string          `json:"status_text"`
```
mmpei: what Status and StatusText should be like?

src/replication/ng/model/execution.go
```diff
@@ -16,17 +16,38 @@ package model
...
+	Total      int             `json:"total"`
+	Failed     int             `json:"failed"`
+	Succeed    int             `json:"succeed"`
+	Pending    int             `json:"pending"`
```
mmpei: here should match the status defined above. Failed Succeed Stoped InProgress.
And i see there is a column called 'trigger_mode' in UI design.  i think we should store it in execution data if this value is mutable in policy.

src/replication/ng/model/execution.go
```diff
@@ -37,7 +58,19 @@ type Task struct {
 	SrcResource  string       `json:"src_resource"`
 	DstResource  string       `json:"dst_resource"`
 	JobID        string       `json:"job_id"`
-	Status       string       `json:"status"`
+	Status       TaskStatus   `json:"status"`
```
mmpei: define a type TaskStatus brings translation to dao.model and mananger.model and makes model could not be shared in dao and manager even they looks the same.

src/replication/ng/flow/flow.go
```diff
@@ -0,0 +1,206 @@
...
+
+	tasks, err := f.scheduler.Schedule(f.resources, dstResources)
+	if err != nil {
+		f.markExecutionFailure(err)
+		return err
+	}
+
+	allFailed := true
+	for _, task := range tasks {
+		if task.Status != model.TaskStatusFailed {
```
mmpei: when a task will be set to failed even before creating?  does action here match with the one in scheduler?

src/replication/ng/model/execution.go
```diff
@@ -16,17 +16,38 @@ package model
...
+const (
+	ExecutionStatusFailed     ExecutionStatus = "Failed"
+	ExecutionStatusSucceed    ExecutionStatus = "Succeed"
+	ExecutionStatusStopped    ExecutionStatus = "Stopped"
+	ExecutionStatusInProgress ExecutionStatus = "InProgress"
+	TaskStatusFailed          TaskStatus      = "Failed"
+	TaskStatusSucceed         TaskStatus      = "Succeed"
+	TaskStatusStopped         TaskStatus      = "Stopped"
+	TaskStatusInProgress      TaskStatus      = "InProgress"
+	TaskStatusPending         TaskStatus      = "Pending"
```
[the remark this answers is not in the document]
ywk253100: No "Retrying" status, if a job need to be retry, it is put into the queue again, so it's status is "Pending". This is not supported by jobservice for now.

src/replication/ng/flow/flow.go
ywk253100: The new changes that we talked about will be pushed in later PR.

src/replication/ng/model/execution.go
```diff
@@ -41,3 +63,15 @@ type Task struct {
...
 	EndTime      time.Time    `json:"end_time"`
 }
+
+// ExecutionQuery defines the query conditions for listing executions
+type ExecutionQuery struct {
+	PolicyID int64
+}
+
+// TaskQuery defines the query conditions for listing tasks
+type TaskQuery struct {
```
cd1989: We don't need pagination ?

cd1989:
LGTM except one comment.

src/replication/ng/model/execution.go
ywk253100: Pagination added

PR #6865, opened 2019-01-30
Remove the token attribute from robot table

wy65701436:
This commit is to remove the token attribute as harbor doesn't store the token in DB.


heww:
This will have no effect when I have installed harbor from currently master branch for 0004 is migrated in db.

PR #6917, opened 2019-02-14
scheduler module for replication

yuanshuhan:
submit the scheduler module for replication


src/replication/ng/scheduler/schedule.go
```diff
@@ -0,0 +1,117 @@
...
+		task.SrcResource = string(src)
+		dest, err := json.Marshal(destResource)
+		if err != nil {
+			log.Errorf("failed to marshal the destResource of %v.err:%s!", destResource, err.Error())
+			task.Status = "Error"
+			tasks = append(tasks, task)
+			continue
+		}
+		task.DstResource = string(dest)
+		task.Status = "Initial"
```
mmpei: we don't design a Initial status for task, maybe should use Pending?
i feel a little confused about the translate of the status.
init(Pending)->js accept and return a jobID(running)->run success(Success)
                                                                               ->run failed(Failed)
                                                                               ->retrying(Retrying) this not support now
                                                                               ->stoped(Stopped)
@steven-zou @ywk253100

src/replication/ng/scheduler/schedule.go
```diff
@@ -0,0 +1,117 @@
...
+
+		job.Name = common_job.ImageTransfer
+		job.Parameters = map[string]interface{}{
+			"src_resource": task.SrcResource,
+			"dst_resource": task.DstResource,
+		}
+		uuid, err := d.client.SubmitJob(job)
+		if err != nil {
+			log.Errorf("failed to submit the task:%v .err:%s!", task, err.Error())
+			task.Status = "Error"
```
mmpei: this status should matched with the status defined in execution.go(wenkai created).
	TaskStatusFailed     string = "Failed"
	TaskStatusSucceed    string = "Succeed"
	TaskStatusStopped    string = "Stopped"
	TaskStatusInProgress string = "InProgress"
	TaskStatusPending    string = "Pending"

src/replication/ng/scheduler/schedule.go
yuanshuhan: remove the task status from scheduler, leave the status management to flow controller

src/replication/ng/scheduler/schedule.go
yuanshuhan: remove the task status from scheduler, leave the status management to flow controller

src/replication/ng/scheduler/scheduler.go
```diff
@@ -42,3 +64,85 @@ type Scheduler interface {
 	// Stop the job specified by ID
 	Stop(id string) error
 }
+
+// Preprocess the resources and returns the item list that can be scheduled
+func (d *DefaultReplicator) Preprocess(srcResources []*model.Resource, destResources []*model.Resource) ([]*ScheduleItem, error) {
+	if len(srcResources) != len(destResources) {
+		err := errors.New("srcResources has different length with destResources")
```
steven-zou: if this error will be bubbled to the users finally, I suggest we should use some more meaningful statements to describe the problem.

src/replication/ng/scheduler/scheduler.go
```diff
@@ -42,3 +64,85 @@ type Scheduler interface {
...
+}
+
+// Schedule transfer the tasks to jobs,and then submit these jobs to job service.
+func (d *DefaultReplicator) Schedule(items []*ScheduleItem) ([]*ScheduleResult, error) {
+	var results []*ScheduleResult
+	for _, item := range items {
+		if item.TaskID == 0 {
+			err := errors.New("some tasks do not have a ID")
+			log.Errorf(err.Error())
+			return nil, err
```
ywk253100: Put the error to its result, don't mark the whole function failed, or the flow controller will mark the whole execution failed.

src/replication/ng/scheduler/scheduler.go
```diff
@@ -42,3 +64,85 @@ type Scheduler interface {
 	// Stop the job specified by ID
 	Stop(id string) error
 }
+
+// Preprocess the resources and returns the item list that can be scheduled
+func (d *DefaultReplicator) Preprocess(srcResources []*model.Resource, destResources []*model.Resource) ([]*ScheduleItem, error) {
+	if len(srcResources) != len(destResources) {
+		err := errors.New("srcResources has different length with destResources")
+		log.Errorf(err.Error())
```
ywk253100: Don't log error, just return it, the caller is responsible for that. The same to other error handle logic

ywk253100:
@yuanshuhan Please squash your commits, thanks.

PR #6933, opened 2019-02-14
Delete open_source_license because it interferes with the CNCF recognized license for Harbor

michmike:
hi everyone,

the file open_source_license is interfering with how CNCF detects and reports the license for Harbor. When that file is present, it is messing up the rendering of the license in github. As such, in the CNCF page https://landscape.cncf.io/format=card-mode&license=other, harbor is listed under `other` license instead of `Apache 2.0`

I know this file may be needed, but who needs it? If it is VMware, we need to get it out of the open source project and into our VMware CI/CD process instead.

this is what a properly rendered license file looks like once i delete this file.
image


michmike:
I will do the DCO if you all agree with this PR

reasonerjt:
@michmike 
Uploading the OSL to this repo is for making sure it's in the VIC bundle.  

I don't have disagreement, but let me discuss with Steven R before merging this PR.

steven-zou:
+1

One more thing, I'd like to suggest keeping a document to describe what upstream frameworks/libraries we're depending on to building our project for us as well as the community to get an overall picture of the dependency tree.

@michmike

michmike:
> ping a document to describe what upstream frameworks/libraries we're depending on to building our project for us

would using the same file and renaming it to a document file with explanations work?

michmike:
updated the PR. PTAL

PR #6941, opened 2019-02-15
Fix #6698: cannot create a same name replication policy after deleting it

### LINKED ISSUE #6698: cannot recreate a replication policy with identical name
If I delete a replication rule in Harbor 1.7, I cannot recreate a rule with the same name. The creation is failing with the error 500 'Internal Server Error'. Deleted replication rules are still listed in the postgres database, in the table 'replication_policy', with a 't' in the column 'deleted'. If I manually delete this entry in the postgres table, I can then recreate a replication policy with the same name. Is that an intended behaviour? Is there a reason to keep an entry in the database for a deleted replication rule?

ywk253100:
This commit fixes the issue #6698: cannot create a same name replication policy after deleting it

### DIFF
--- a/src/replication/policy/manager.go
+++ b/src/replication/policy/manager.go
@@ -196,5 +196,10 @@ func (m *DefaultManager) UpdatePolicy(policy models.ReplicationPolicy) error {
 // RemovePolicy removes the specified policy;
 // If removing failed, error will be returned.
 func (m *DefaultManager) RemovePolicy(policyID int64) error {
+	// delete replication jobs
+	if err := dao.DeleteRepJobs(policyID); err != nil {
+		return err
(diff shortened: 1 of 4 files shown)


src/replication/policy/manager.go
```diff
@@ -196,5 +196,10 @@ func (m *DefaultManager) UpdatePolicy(policy models.ReplicationPolicy) error {
 // RemovePolicy removes the specified policy;
 // If removing failed, error will be returned.
 func (m *DefaultManager) RemovePolicy(policyID int64) error {
+	// delete replication jobs
```
reasonerjt: Can we add a todo as this is not in a transaction?

PR #6947, opened 2019-02-17
Ng policy manager

kofj:
1. Define ng persist replication policy model.
2. Add ng replication policy CURD methods.
3. Implement ng policy manger.

Signed-off-by: Frank Kung <kfanjian@gmail.com>

### DIFF
--- a/src/replication/ng/dao/policy.go
+++ b/src/replication/ng/dao/policy.go
@@ -1,0 +1,88 @@
+package dao
+
+import (
+	"errors"
+	"time"
+
+	"github.com/astaxie/beego/orm"
+	common_dao "github.com/goharbor/harbor/src/common/dao"
+	"github.com/goharbor/harbor/src/replication/ng/dao/models"
+)
+
+// AddRepPolicy insert new policy to DB.
+func AddRepPolicy(policy *models.RepPolicy) (int64, error) {
+	o := common_dao.GetOrmer()
+	now := time.Now()
+	policy.CreationTime = now
+	policy.UpdateTime = now
+
+	return o.Insert(policy)
+}
+
+func filteredRepPolicyQuerySeter(name, namespace string) orm.QuerySeter {
+	var qs = common_dao.GetOrmer().QueryTable(new(models.RepPolicy))
+
+	// TODO: just filter polices by name now, and need consider how to  filter namespace.
+	qs = qs.Filter("name__icontains", name)
+
+	return qs
+}
+
+// GetTotalOfRepPolicies returns the total count of replication policies
+func GetTotalOfRepPolicies(name, namespace string) (int64, error) {
+	var qs = filteredRepPolicyQuerySeter(name, namespace)
+	return qs.Count()
+}
+
+// GetPolicies filter policies and pagination.
+func GetPolicies(name, namespace string, page, pageSize int64) (policies []*models.RepPolicy, err error) {
+	var qs = filteredRepPolicyQuerySeter(name, namespace)
+
+	// Paginate
+	if page > 0 && pageSize > 0 {
+		qs = qs.Limit(pageSize, (page-1)*pageSize)
+	}
+
+	_, err = qs.All(&policies)
+
+	return
+}
+
+// GetRepPolicy return special policy by id.
+func GetRepPolicy(id int64) (policy *models.RepPolicy, err error) {
+	policy = new(models.RepPolicy)
+	err = common_dao.GetOrmer().QueryTable(policy).
+		Filter("id", id).One(policy)
+
+	return
+}
+
+// GetRepPolicyByName return special policy by name.
+func GetRepPolicyByName(name string) (policy *models.RepPolicy, err error) {
+	policy = new(models.RepPolicy)
+	err = common_dao.GetOrmer().QueryTable(policy).
+		Filter("name", name).One(policy)
+
+	return
+}
+
+// UpdateRepPolicy update fields by props
+func UpdateRepPolicy(policy *models.RepPolicy, props ...string) (err error) {
+	var o = common_dao.GetOrmer()
+
+	if policy != nil {
+		_, err = o.Update(policy, props...)
+	} else {
+		err = errors.New("Nil policy")
+	}
+
+	return
+}
+
+// DeleteRepPolicy will hard delete database item
+func DeleteRepPolicy(id int64) error {
+	o := common_dao.GetOrmer()
+
+	_, err := o.Delete(&models.RepPolicy{ID: id})
+	return err
+}
--- a/src/replication/ng/policy/manager.go
+++ b/src/replication/ng/policy/manager.go
@@ -15,8 +15,103 @@
 package policy
 
 import (
+	"encoding/json"
+	"errors"
+	"strings"
+	"time"
+
+	"github.com/goharbor/harbor/src/replication/ng/dao"
+	persist_models "github.com/goharbor/harbor/src/replication/ng/dao/models"
 	"github.com/goharbor/harbor/src/replication/ng/model"
 )
+
+var errNilPolicyModel = errors.New("nil policy model")
+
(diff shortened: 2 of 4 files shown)


src/common/dao/replication_job.go
```diff
@@ -136,6 +137,114 @@ func AddRepPolicy(policy models.RepPolicy) (int64, error) {
 	return policyID, nil
 }
 
+// AddRepPolicyNg ...
+func AddRepPolicyNg(policy models.RepPolicyNg) (int64, error) {
```
ywk253100: We'll remove all `ng` at last, so can we rename the function name? We can create a new folder called `ng` and put the functions in the folder.

src/common/dao/replication_job.go
```diff
@@ -136,6 +137,114 @@ func AddRepPolicy(policy models.RepPolicy) (int64, error) {
...
+	o := GetOrmer()
+	sql := `insert into replication_policy_ng (name, description, creator, src_registry_id, src_namespaces,
```
ywk253100: Use the function provided by the orm rather than the raw SQL

src/common/dao/replication_job.go
```diff
@@ -136,6 +137,114 @@ func AddRepPolicy(policy models.RepPolicy) (int64, error) {
...
+	}
+
+	// WARNING: If cols is null then update all columns.
+	_, err := o.Update(policy, props...)
+
+	return err
+}
+
+// DeleteRepPolicyNg ...
+func DeleteRepPolicyNg(id int64) error {
```
ywk253100: We need to remove the record from database when one policy is deleted, the changes for master branch is submitted https://github.com/goharbor/harbor/pull/6941

src/replication/ng/policy/manager.go
```diff
@@ -32,3 +120,69 @@ type Manager interface {
...
+	var result []*model.Policy
+	for _, policy := range policies {
+		ply, err := convertFromPersistModel(policy)
+		if err != nil {
+			return 0, nil, err
+		}
+
+		result = append(result, ply)
+	}
+	return int64(len(result)), result, nil
```
ywk253100: The first returning value of the function should be the total count, but the `result` seems just a subset?

ywk253100:
Need UT for the functions

kofj:
> Need UT for the functions

I'll Add UT later.

src/common/dao/replication_job.go
kofj: Hard delete? Why we soft-delete before?

src/common/dao/replication_job.go
kofj: OK

src/common/dao/replication_job.go
kofj: I'll remove raw sql with ORM.But why you use raw sql before?

src/replication/ng/policy/manager.go
kofj: ok

src/common/dao/replication_job.go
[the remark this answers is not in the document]
kofj: `ng`  suffix will be remove lataer. The question should ask @ywk253100 : Why the file named 'replication_'`job`?

src/replication/ng/policy/manager.go
```diff
@@ -15,9 +15,97 @@
...
+	"encoding/json"
+	"strings"
+	"time"
+
+	"github.com/goharbor/harbor/src/common/dao"
+	persist_models "github.com/goharbor/harbor/src/common/models"
 	"github.com/goharbor/harbor/src/replication/ng/model"
 )
 
+func convertFromPersistModel(policy *persist_models.RepPolicyNg) (*model.Policy, error) {
```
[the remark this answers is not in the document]
kofj: I don't think so, those functions just convert model. Maybe should rename to  `convertFromORM` and `convertToORM`.

src/replication/ng/dao/policy.go
```diff
@@ -0,0 +1,115 @@
...
+func GetPolicies(name, namespace string, page, pageSize int64) ([]*model.RepPolicy, error) {
+	var qs = filteredRepPolicyQuerySeter(name, namespace)
+
+	// Paginate
+	if page > 0 && pageSize > 0 {
+		qs = qs.Limit(pageSize, (page-1)*pageSize)
+	}
+
+	var policies []*model.RepPolicy
+	qs.All(&policies)
```
ywk253100: Handle error?

src/replication/ng/dao/policy.go
```diff
@@ -0,0 +1,115 @@
...
+func GetRepPolicy(id int64) (policy *model.RepPolicy, err error) {
+	err = common_dao.GetOrmer().QueryTable(policy).
+		Filter("id", id).One(policy)
+
+	return
+}
+
+// GetRepPolicyByName return special policy by name.
+func GetRepPolicyByName(name string) (policy *model.RepPolicy, err error) {
+
```
ywk253100: Not implement?

src/replication/ng/dao/policy.go
```diff
@@ -0,0 +1,115 @@
...
+// UpdateRepPolicy TODO: update fields by props
+func UpdateRepPolicy(policy *model.RepPolicy, props ...string) (err error) {
+	var o = common_dao.GetOrmer()
+
+	if policy.ID > 0 {
+		_, err = o.Update(policy, props...)
+	} else {
+		return errors.New("policy id MUST > 0")
+	}
+
```
ywk253100: No logic to update the model.

src/replication/ng/policy/manager.go
```diff
@@ -32,3 +125,86 @@ type Manager interface {
...
+}
+
+// List returns all the policies
+func (m *DefaultManager) List(queries ...*model.PolicyQuery) (total int64, polices []*model.Policy, err error) {
+	// default query parameters
+	var name = ""
+	var namespace = ""
+	var page int64 = 1
+	var pageSize int64 = 15
+
```
ywk253100: The default values for page and size are given on the API level, so the default values for them aren't needed here.

src/replication/ng/dao/policy.go
```diff
@@ -0,0 +1,115 @@
...
+func filteredRepPolicyQuerySeter(name, namespace string) orm.QuerySeter {
+	var qs = common_dao.GetOrmer().QueryTable(new(model.RepPolicy))
+
+	var cond = orm.NewCondition()
+	// Pull-based polices in the special namespace.
+	// If dest_registry_id equal 0, dest_namespace is the namespace of Harbor.
+	pullBasedCond := cond.And("dest_registry_id", 0).And("dest_namespace", namespace)
+	// Push-based polices in the special namespace.
+	// If src_registry_id equal 0, src_namespaces only have single namespace of Harbor.
+	pushBasedCond := cond.And("src_registry_id", 0).And("src_namespaces", namespace)
```
ywk253100: Source namespaces should be an array, seems the logic here is not correct

kofj: The namespaces is array in the manager's model, but it is serialize to string stored in database, and split with comma. In push-based mode, only one namespace in the namespaces array.

src/replication/ng/dao/policy.go
kofj: Forget to commit...

src/replication/ng/dao/policy.go
kofj: fixed

src/replication/ng/dao/policy.go
ywk253100: The source namespaces may have more than one element in both pull and push mode.

kofj: Replicate multi-namespaces from harbor to another registry in one policy?I don't think it is a good idea.

kofj: By the sync meeting, we just filter name now.

Issue #6952, opened 2019-02-18
Re-introduce the DB migration process in notary


reasonerjt:
The migrate scripts were removed from notary images to fix vulnerability issue.

Although there's no immediate requirement, (no schema change in more than 2 years and no plan to indicate a schema change will happen).  We need to re-introduce the migration process before there's a schema change in notary.  

It will use the new version of the golang-migrate tool and need to handle the break change of `schema_migrations`

More details see #5863 and https://github.com/golang-migrate/migrate/commit/2e54bda0e1e126658db3b55e096ae999e40e7894

reasonerjt:
Dup to #5863

Issue #5863, opened 2018-09-10
Bump up the migrate binary in notary containers to v3.x


reasonerjt:
The `migrate` binary packaged in notary containers is out-dated.

Seems there's a name resolution issue in the `migrate` binary currently packaged in notary containers, that it tries to look up the $hostname.$searchdomain first before trying to look up the $hostname.

But I find updating the binary will introduce break change like this:
```
error: pq: column "dirty" does not exist in line 0: SELECT version, dirty FROM "schema_migrations" LIMIT 1
``` 

So I'll put this into backlog, and consider to remove the migrate binary from notary containers until there's change in the schema of upstream notary project.

reasonerjt:
We don't plan to fix it as the name resolution is not due to the migrator.

reasonerjt:
Targeting 1.7.4, this involves adding a patch to handle the breaking change, and refine the build process.

PR #6953, opened 2019-02-18
Remove DB migration from notary images

reasonerjt:
There is a high risk vulnerabity in the migrate binary we've been using
in notary images.  I tried to bump up the binary but there was a
breaking change, more details see #5863.
To provide a fix to the vulnerability in 1.7.x and given the fact that
Notary has not updated the schema for 2 years.  We decided to remove the
migrate scripts from notary images.
This commit removes it, and a new issue #6952 is opened to track the
work to re-introduce the migration process.


reasonerjt:
There's misunderstanding and the migration tool cannot be removed.

PR #6963, opened 2019-02-19
Remove everything of adminserver container

stonezdj:
- Remove source code and dockerfile.
- Update document to remove everything about adminserver.


src/common/config/manager.go
```diff
@@ -125,7 +125,7 @@ func (c *CfgManager) GetAll() map[string]interface{} {
 	metaDataList := metadata.Instance().GetAll()
 	for _, item := range metaDataList {
 		cfgValue, err := c.store.GetAnyType(item.Name)
-		if err != metadata.ErrValueNotSet && err != nil {
+		if err != nil {
```
reasonerjt: Please double check the change.

stonezdj: Updated

PR #6988, opened 2019-02-21
Implement the operation controller

ywk253100:
This commit implements the operation controller. The operation controller wraps the flow controller and execution manager to provide capabilities for the upper level

### DIFF
--- a/src/replication/ng/flow/controller.go
+++ b/src/replication/ng/flow/controller.go
@@ -55,7 +55,7 @@ type defaultController struct {
 	scheduler    scheduler.Scheduler
 }
 
-// Replicate according the to policy ID
+// Start a replication according to the policy
 func (d *defaultController) StartReplication(policy *model.Policy) (int64, error) {
 	log.Infof("starting the replication based on the policy %d ...", policy.ID)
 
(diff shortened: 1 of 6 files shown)


src/replication/ng/flow/controller.go
```diff
@@ -55,7 +55,7 @@ type defaultController struct {
 	scheduler    scheduler.Scheduler
 }
 
-// Replicate according the to policy ID
+// Start a replication according to the policy
```
cd1989: // StartReplication starts a replication according to the policy

src/replication/ng/replication.go
```diff
@@ -0,0 +1,54 @@
...
+	// RegistryMgr is a global registry manager
+	RegistryMgr registry.Manager
+	// ExecutionMgr is a global execution manager
+	ExecutionMgr execution.Manager
+	// OperationCtl is a global operation controller
+	OperationCtl operation.Controller
+)
+
+// Init the global variables
+func Init() error {
```
cd1989: Not `init` ?

src/replication/ng/flow/controller.go
ywk253100: It isn't a public function, so the comment should be ok

src/replication/ng/replication.go
ywk253100: There should be some initial work need to be done before call this method, such as initialize the database connection.

cd1989: I see, thank you.

PR #7004, opened 2019-02-22
add expiration of robot account

wy65701436:
This commit is to make the expiration of robot account configurable

1, The expiration could be set by system admin in the configuation page or
by /api/config with robot_token_expiration=60, the default value is 30 days.
2, The expiration could be shown in the robot account infor both on UI and API.

### DIFF
--- a/docs/swagger.yaml
+++ b/docs/swagger.yaml
@@ -4734,6 +4734,9 @@ definitions:
       description:
         type: string
         description: The description of robot account
+      expiration:
+        type: integer
+        description: The expiration of robot account
       project_id:
         type: integer
         description: The project id of robot account
--- a/docs/swagger.yaml
+++ b/docs/swagger.yaml
@@ -4734,9 +4734,9 @@ definitions:
       description:
         type: string
         description: The description of robot account
-      expiration:
+      expiresat:
         type: integer
-        description: The expiration of robot account
+        description: The expiration of robot account (in seconds)
       project_id:
         type: integer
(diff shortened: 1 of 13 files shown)


make/migrations/postgresql/0004_add_robot_account.up.sql
```diff
@@ -3,6 +3,7 @@ CREATE TABLE robot (
  name varchar(255),
  description varchar(1024),
  project_id int,
+ expiration int,
```
reasonerjt: https://github.com/dgrijalva/jwt-go/blob/master/claims.go#L20
This is a `int64`, so probably we should use `bigint`

src/core/config/config.go
```diff
@@ -225,6 +225,11 @@ func TokenExpiration() (int, error) {
 	return cfgMgr.Get(common.TokenExpiration).GetInt(), nil
 }
 
+// RobotTokenExpiration returns the token expiration time of robot account (in day)
+func RobotTokenExpiration() int {
```
reasonerjt: What if user want to set it to hours?  
To be consistent with the token by Harbor the unit should be minute?

docs/swagger.yaml
```diff
@@ -4734,6 +4734,9 @@ definitions:
       description:
         type: string
         description: The description of robot account
+      expiration:
+        type: integer
+        description: The expiration of robot account
```
ywk253100: It's in minute or second? It's better to describe it here.

PR #7021, opened 2019-02-26
Add execution and task manager

mmpei:


src/common/dao/execution.go
```diff
@@ -0,0 +1,376 @@
...
+	"github.com/goharbor/harbor/src/common/models"
+	"github.com/goharbor/harbor/src/common/utils/log"
+	"fmt"
+)
+
+// AddExecution ...
+func AddExecution(execution *models.Execution) (int64, error) {
+	o := GetOrmer()
+
+	sql := "insert into replication_execution (policy_id, status, status_text, total, failed, succeed, in_progress, stopped, trigger) " +
```
ywk253100: Use the methods provided by the ORM instead of raw sql.

src/replication/ng/execution/execution.go
```diff
@@ -54,3 +57,130 @@ type Manager interface {
...
+}
+
+// UpdateTaskStatus ...
+func (dm *DefaultManager) UpdateTaskStatus(taskID int64, status string, statusCondition ...string) error {
+	n, err := dao.UpdateTaskStatus(taskID, status, statusCondition ...)
+	if err != nil {
+		return err
+	}
+	if n == 0 {
+		return fmt.Errorf("Update task status failed %d: -> %s ", taskID, status)
```
ywk253100: The error message should be same with others: `fmt.Errorf("Task not found error: %d ", task.ID)`?

src/core/service/notifications/jobs/handler.go
```diff
@@ -87,3 +87,13 @@ func (h *Handler) HandleReplication() {
 		return
 	}
 }
+
+// HandleReplicationTask handles the webhook of replication task
+func (h *Handler) HandleReplicationTask() {
+	log.Debugf("received replication task status update event: task-%d, status-%s", h.id, h.status)
+	if _, err := dao.UpdateTaskStatus(h.id, h.status); err != nil {
```
ywk253100: Use the Execution Manager defined here: https://github.com/goharbor/harbor/blob/replication_ng/src/replication/ng/replication.go#L34

src/replication/ng/execution/execution.go
mmpei: condition could be triggered not only by task not found.  actually error is always not null when n equals 0, so this condition will never reached.

src/common/dao/execution.go
mmpei: start time and end time not inserted, it will be initalized by DB.

src/core/service/notifications/jobs/handler.go
mmpei: i will change.

PR #7037, opened 2019-02-27
Refactor the adapter interface

ywk253100:
This commit refactors the Adapter interface and adjust the code in the flow controller and repository handler

### DIFF
--- a/src/replication/ng/adapter/image_registry.go
+++ b/src/replication/ng/adapter/image_registry.go
@@ -1,0 +1,33 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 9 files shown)


src/replication/ng/adapter/image_registry.go
```diff
@@ -0,0 +1,33 @@
...
+
+// ImageRegistry defines the capabilities that an image registry should have
+type ImageRegistry interface {
+	FetchImages(namespaces []string, filters []*model.Filter) ([]*model.Resource, error)
+	ManifestExist(repository, reference string) (exist bool, digest string, err error)
+	PullManifest(repository, reference string, accepttedMediaTypes []string) (manifest distribution.Manifest, digest string, err error)
+	PushManifest(repository, reference, mediaType string, payload []byte) error
+	BlobExist(repository, digest string) (exist bool, err error)
+	PullBlob(repository, digest string) (size int64, blob io.ReadCloser, err error)
+	PushBlob(repository, digest string, size int64, blob io.Reader) error
```
cd1989: I think there should be a `MountBlob` method to mount blob within the same registry.

ywk253100: If we want to support the `MountBlob`, we need a mapping store to store where the blob is. I think it's too complicated for our cases. Such as we need to refresh the store in case the images are removed from the registry.

cd1989: If the `ImageRegistry` is used only in replication context, then there is no problem. By `MountBlob`, I suppose the caller to specify from where to mount the blob. So it's not useful in replication.

```
MountBlob(repository, blobID, from string) error
```

PR #7063, opened 2019-03-04
Add users search API

heww:
Add /users/search API for autocomplete input when create project member by projectAdmin.

### DIFF
--- a/src/core/api/user.go
+++ b/src/core/api/user.go
@@ -46,6 +46,11 @@ type passwordReq struct {
 	NewPassword string `json:"new_password"`
 }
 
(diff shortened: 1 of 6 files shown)


src/core/api/user.go
```diff
@@ -166,6 +171,40 @@ func (ua *UserAPI) List() {
 	ua.ServeJSON()
 }
 
+// Search ...
+func (ua *UserAPI) Search() {
```
ywk253100: Add test for the API

heww: Test added

PR #7075, opened 2019-03-05
update gc api to support raw cron string

wy65701436:


src/core/api/models/reg_gc.go
```diff
@@ -75,44 +77,33 @@ func (gr *GCReq) Valid(v *validation.Validation) {
 		return
 	}
 	switch gr.Schedule.Type {
-	case ScheduleDaily, ScheduleWeekly:
-		if gr.Schedule.Offtime < 0 || gr.Schedule.Offtime > 3600*24 {
-			v.SetError("offtime", fmt.Sprintf("Invalid schedule trigger parameter offtime: %d", gr.Schedule.Offtime))
+	case ScheduleHourly, ScheduleDaily, ScheduleWeekly, ScheduleCustom:
```
reasonerjt: Do we still need the `Type` of Schedule at all?  Since it's always `cron`?

wy65701436: It's as designed, I have discussed with Meina. It just for UI to show history, otherwise, UI has no way to know it.

PR #7105, opened 2019-03-08
Replication ng execution

mmpei:
new PR for ng replication for execution feature.


src/core/service/notifications/jobs/handler.go
```diff
@@ -66,6 +68,8 @@ func (h *Handler) Prepare() {
 		return
 	}
 	h.status = status
+
+	h.executionMgr, _ = execution.NewDefaultManager()
```
ywk253100: Use the `OperationController` defined here https://github.com/goharbor/harbor/blob/replication_ng/src/replication/ng/replication.go#L41. It's the interface for the upper level. If it contains no methods you need, just add them.

src/core/service/notifications/jobs/handler.go
```diff
@@ -87,3 +91,13 @@ func (h *Handler) HandleReplication() {
 		return
 	}
 }
+
+// HandleReplicationTask handles the webhook of replication task
+func (h *Handler) HandleReplicationTask() {
+	log.Debugf("received replication task status update event: task-%d, status-%s", h.id, h.status)
+	if err := h.executionMgr.UpdateTaskStatus(h.id, h.status); err != nil {
```
ywk253100: Also need to update the job id to the record

make/migrations/postgresql/0006_ng_replication.up.sql
```diff
@@ -12,4 +12,35 @@ CREATE TABLE registry (
  creation_time timestamp default CURRENT_TIMESTAMP,
  update_time timestamp default CURRENT_TIMESTAMP,
  CONSTRAINT unique_registry_name UNIQUE (name)
-);
\ No newline at end of file
+);
+
+create table replication_execution (
```
ywk253100: Put all sql to https://github.com/goharbor/harbor/blob/replication_ng/make/migrations/postgresql/0004_1.8.0_schema.up.sql

src/core/router.go
```diff
@@ -125,12 +125,19 @@ func initRouters() {
 	beego.Router("/api/internal/syncregistry", &api.InternalAPI{}, "post:SyncRegistry")
 	beego.Router("/api/internal/renameadmin", &api.InternalAPI{}, "post:RenameAdmin")
 
+	// next generation replication API
+	//beego.Router("/api/replication/executions", &api.ExecutionAPI{}, "get:List")
+	//beego.Router("/api/replication/executions/:id([0-9]+)", &api.ExecutionAPI{}, "get:GetExecution;put:Stop;post:Run")
+	//beego.Router("/api/replication/executions/:id([0-9]+)/jobs", &api.ExecutionAPI{}, "get:ListJobs")
+	//beego.Router("/api/replication/executions/:id([0-9]+)/logs/:logid([0-9]+)", &api.ExecutionAPI{}, "get:GetLog")
```
cd1989: Why are these endpoints commented out ?

src/replication/ng/dao/execution.go
```diff
@@ -0,0 +1,377 @@
...
+
+// AddExecution ...
+func AddExecution(execution *models.Execution) (int64, error) {
+	o := dao.GetOrmer()
+
+	sql := "insert into replication_execution (policy_id, status, status_text, total, failed, succeed, in_progress, stopped, trigger) " +
+		"values (?, ?, ?, ?, ?, ?, ?, ?, ?) RETURNING id"
+
+	var id int64
+	err := o.Raw(sql, execution.PolicyID, execution.Status, execution.StatusText, execution.Total, execution.Failed,
```
cd1989: Why not use `o.Insert` ?

mmpei: use db to initialize start_time and end_time. start_time will be set to now and end_time will be nil.

src/core/service/notifications/jobs/handler.go
mmpei: done.

src/core/service/notifications/jobs/handler.go
mmpei: no need to modify.

src/replication/ng/dao/execution.go
mmpei: in the first design execution doesn't contain the property 'status'.  and status will be generated from other properties.

Issue #7091, opened 2019-03-07
notary-server and notary-signer cannot start after upgrading to 1.7.4 


reasonerjt:
the log of notary-server container:
```
Mar  7 06:52:25 172.18.0.1 notary-server[5654]: 2019/03/07 06:52:25 Updating database.
Mar  7 06:52:25 172.18.0.1 notary-server[5654]: 2019/03/07 06:52:25 Done updating database.
Mar  7 06:52:25 172.18.0.1 notary-server[5654]: error: file does not exist
Mar  7 06:52:25 172.18.0.1 notary-server[5654]: waiting for postgres://server:password@postgresql:5432/notaryserver?sslmode=disable to come up.
Mar  7 06:52:26 172.18.0.1 notary-server[5654]: error: file does not exist
Mar  7 06:52:26 172.18.0.1 notary-server[5654]: waiting for postgres://server:password@postgresql:5432/notaryserver?sslmode=disable to come up. 
```
The ` file does not exist ` error was due to an inconsistecy introduced in migration that switch Harbor DB from mysql/mariadb to postgresql

There's no good way to fix it easily, a quick fix is to drop all tables of the `notaryserver` and `notarysigner` databases, and re-sign the images that were signed previously.
A better workaround involve some manual work is in progress.

RRAlex:
Upgraded from 1.4.0 -> 1.6.3 without issue, was about to upgrade to 1.7.4 but saw the warning...
Does the bug manifests itself with previous 1.7.x?
Or is upgrading to 1.7.3, until 1.7.5 arrives, a safe move?
Cheers!

wy65701436:
@RRAlex it happens on v1.7.4 forward, and we will provide a manual workaround, you have to run a script at your end to fix it.

wy65701436:
Just run the script(https://github.com/goharbor/harbor/blob/master/tools/notary-migration-fix.sh) to fix notary DB before launching Harbor.

Madic-:
I've upgraded from 1.6.0 to 1.8.1 and came accross the same issue and did run the notary-migration-fix.sh script but still get the "file does not exist" error. Any idea?

apallier:
Hi, I have the same issue and same behavior as @Madic- when I run the script on a 1.8.2 upgrade.
Should I run this script when Harbor is running or stopped ?

Madic-:
"Fixed" the problem by dropping and then re-creating the databases notaryserver and notarysigner.

PR #7106, opened 2019-03-08
add script to fix migration issue on notary

wy65701436:
The script is to fix the issue #7091, notary upgrade failure from mysql to pqsl.

### DIFF
--- a/tools/notary-migration-fix.sh
+++ b/tools/notary-migration-fix.sh
@@ -29,6 +29,7 @@ clean_db() {
 }
 
 wait_for_db_ready() {
+    set +e
     TIMEOUT=12
     while [ $TIMEOUT -gt 0 ]; do
         docker exec fix-notary-migration pg_isready | grep "accepting connections"
@@ -43,6 +44,7 @@ wait_for_db_ready() {
         clean_db
         exit 1
     fi
+    set -e
 }
 
 fix_notary_server() {
--- a/tools/notary-migration-fix.sh
+++ b/tools/notary-migration-fix.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+# Copyright Project Harbor Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -ex
+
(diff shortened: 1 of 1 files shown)


tools/notary-migration-fix.sh
```diff
@@ -0,0 +1,61 @@
...
+
+clean_db() {
+    docker stop fix-notary-migration
+    docker rm fix-notary-migration
+}
+
+wait_for_db_ready() {
+    TIMEOUT=12
+    while [ $TIMEOUT -gt 0 ]; do
+        docker exec fix-notary-migration pg_isready | grep "accepting connections"
```
stonezdj: If nothing is found, grep will return non-zero value and this script will exit, to avoid this, you need to add ' || true ' at the end.

tools/notary-migration-fix.sh
```diff
@@ -0,0 +1,61 @@
...
+# limitations under the License.
+#
+
+set -ex
+
+harbor_db_image=$(docker images goharbor/harbor-db --format "{{.Repository}}:{{.Tag}}")
+harbor_db_path="/data/database"
+
+launch_db() {
+    docker run -d --name fix-notary-migration -v ${harbor_db_path}:/var/lib/postgresql/data ${harbor_db_image} "postgres"
```
stonezdj: In tile env, the docker is '/var/vcap/packages/docker/bin/docker -H unix:///var/vcap/sys/run/docker/dockerd.sock', you need to extract it as variable.

DOCKER_CMD='/var/vcap/packages/docker/bin/docker -H unix:///var/vcap/sys/run/docker/dockerd.sock'
 and use variable in command.
$DOCKER_CMD run  ....

tools/notary-migration-fix.sh
wy65701436: no-zero will still in the while loop tile timeout

tools/notary-migration-fix.sh
wy65701436: I can do it, but how does the script know it's running on tile or vic env? Should I add one more env to identify it? like --env=vic?

tools/notary-migration-fix.sh
reasonerjt: @wy65701436 did you test it with `set -e`?

tools/notary-migration-fix.sh
reasonerjt: @stonezdj if you want to reuse this script in tile, you can probably do this, create another `.sh` script and add 
```sh
docker() {
/var/vcap/packages/docker/bin/docker -H unix:///var/vcap/sys/run/docker/dockerd.sock $@
}
source notary-migration-fix.sh
```
such that `docker` is over-written in your script

tools/notary-migration-fix.sh
stonezdj: set -ex means it exits when any command returns non-zero

tools/notary-migration-fix.sh
stonezdj: It is OK to me

tools/notary-migration-fix.sh
wy65701436: get your points, code updated.

PR #7110, opened 2019-03-11
Fix returning internal error when setting cron for gc

### LINKED ISSUE #7107: Get internal error on setting cron schedule of GC
When I select hourly and save it, it could be saved success. After that, I tried to set it to none, but got internal issue.

pureshine:
fix #7107 
1、If the current type is none, a post request should be sent to other types, otherwise it is a put request.
2、Modify the tooltip description of houly daily weekly

### DIFF
--- a/src/portal/lib/src/config/gc/gc.component.ts
+++ b/src/portal/lib/src/config/gc/gc.component.ts
@@ -97,7 +97,17 @@ export class GcComponent implements OnInit {
     this.disableGC = false;
   }
 
-  getcron(cron: string) {
+  private resetSchedule(cron) {
+    this.schedule = {
+      schedule: {
+        type: this.CronScheduleComponent.scheduleType,
+        cron: cron
+      }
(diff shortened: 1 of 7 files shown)


src/portal/lib/src/config/gc/gc.component.ts
```diff
@@ -97,6 +97,16 @@ export class GcComponent implements OnInit {
...
+    this.schedule = {
+      schedule: {
+        type: this.CronScheduleComponent.scheduleType,
+        cron: cron
+      }
+    };
+    this.getJobs();
+  }
+
   getcron(cron: string) {
```
zhoumeina: why this method called getcron , actually it invoke a put operation to the backend?

pureshine: Yes, because it also gets cron from the cron component, now I think scheduleGc looks more appropriate and has been modified

PR #7130, opened 2019-03-12
Implement a default image registry client

ywk253100:
Provide a default implemmentation for image registry interface, other adapters can use it directly


cd1989:
lgtm

PR #7135, opened 2019-03-13
change promise to observable

jwangyangls:
Change Technology Base：
We replace Promise Technology with Observable in rxjs in JavaScript;

### DIFF
--- a/src/portal/lib/src/repository-gridview/repository-gridview.component.ts
+++ b/src/portal/lib/src/repository-gridview/repository-gridview.component.ts
@@ -27,7 +27,7 @@ import {
     TagService
 } from '../service/index';
 import { ErrorHandler } from '../error-handler/error-handler';
-import { toPromise, CustomComparator, DEFAULT_PAGE_SIZE, calculatePage, doFiltering, doSorting, clone } from '../utils';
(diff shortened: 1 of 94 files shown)


src/portal/lib/src/repository-gridview/repository-gridview.component.ts
```diff
@@ -169,12 +171,14 @@ export class RepositoryGridviewComponent implements OnChanges, OnInit {
...
                     }
+                }, error => {
+                    this.errorHandler.error(error);
                 });
             }
         }
     }
 
-    delOperate(repo: RepositoryItem) {
+    delOperate(repo: RepositoryItem): Observable<any> | null {
```
ninjadq: when `null` returned, is the observable pipeline still work?

jwangyangls: it return null when forkin code run  in error,so it won't affect the results.

jwangyangls: > it return null when forkin code run in error,so it won't affect the results.

We didn't write a return value when we used promise, but this function actually has a return value promise object or undefined. Now I change undefined to a null pointer.

Issue #6303, opened 2018-11-19
Scheduler - features CRON-like


lwqcz:
Hi there,

I have noticed, that in issue related (#4329) is mentioned that "original" intention was like to implement CRON into Harbor to be its job scheduler.

I am testing version **1.6.0-66709daa**. 

Unfortunately, I can't set job scheduler as weekly or daily only as there is no such option.
I need to schedule registry replication from deployment Harbor instance to production one as often as every hour for example. Using current UI that period length is unavailable and even though API I can't find the option to change job parameters likewise.

Can you please consider job scheduler feature extension to allow users to set up replication running more often and even when an image with a specified label is pushed into the registry?

**Thank you a lot.**

_PS: For others that found this "feature request" useful please add 👍  to this post, thanks!_

steven-zou:
@lwqcz 

Thanks for reaching us and raising request. The [job service](https://github.com/goharbor/harbor/tree/master/src/jobservice) (scheduler) itself supports corn-like schedule but the current UI does not support that. We'll consider it. Thanks.

lwqcz:
@steven-zou 

Thank you for your answer. Can you please update this issue when you will have an idea when this UI modification will be planned for next release or so?

And by the way - is there any possibility to make scheduled jobs with an interval like an hour via API? The only way I can imagine is to call API to make replication job to be run immediately+delete itself afterward. Every hour. 😕

clouderati:
Hi @lwqcz – not sure yet as our roadmap for v1.8 is still not finalized. I'll update this issue when we know the payload for the next release. Thanks!

lweitzman:
the mockup for Cron job for vulnerability scanning is at https://vmware.invisionapp.com/share/RKOJ1Q83XNS#/336065235_Configuration_-_Vulnerability_Custom 
this "Custom" set of inputs would be available for GC and Retention Policy as well.

pureshine:
Done as pr #7138 #7083 merged

PR #7138, opened 2019-03-14
Refactoring vulnerability ui support cron

pureshine:
link as #6303 
1、vulnerability schedule support cron like gc ,api is : api/system/scanAll/schedule
2、Scan now api changed from /api/repositories to api/system/scanAll/schedule, same as gc send a post request body is {schedule: {type: "Manual"}}


zhoumeina:
please add description of this pr.

PR #7194, opened 2019-03-21
Remove error message of saving system setting to db

stonezdj:


nlowe:
What's the rationale behind this change? Is there an issue associated with it?

PR #7199, opened 2019-03-22
add authn proxy

wy65701436:
User could use the web hook token issued by k8s api server to login to harbor.
The username should add a specific prefix.


src/core/filter/security.go
```diff
@@ -194,6 +205,114 @@ func (r *robotAuthReqCtxModifier) Modify(ctx *beegoctx.Context) bool {
 	return true
 }
 
+type authProxyReqCtxModifier struct{}
+
+func (ap *authProxyReqCtxModifier) Modify(ctx *beegoctx.Context) bool {
+	// only support docker login
+	if ctx.Request.URL.Path == "/v2/" || ctx.Request.URL.Path == "/service/token" {
+		log.Error("You do not have the proper privileges to perform the action.")
```
reasonerjt: 1. Does it need `v2`?
2. The error message is incorrect.

src/core/filter/security.go
```diff
@@ -194,6 +205,114 @@ func (r *robotAuthReqCtxModifier) Modify(ctx *beegoctx.Context) bool {
...
+		return false
+	}
+
+	proxyUserName, proxyPwd, ok := ctx.Request.BasicAuth()
+	if !ok {
+		return false
+	}
+
+	match, rawUserName := ap.matchAuthProxyUserName(proxyUserName)
```
reasonerjt: conventionally the match should be second return value?  Refer to `value, ok = map[key]`

src/core/filter/security.go
```diff
@@ -194,6 +205,114 @@ func (r *robotAuthReqCtxModifier) Modify(ctx *beegoctx.Context) bool {
...
+
+	user, err := dao.GetUser(models.User{
+		Username: rawUserName,
+	})
+	if err != nil {
+		log.Errorf("fail to get user: %v", err)
+		return false
+	}
+	if user == nil {
+		log.Errorf("User: %s has not been on boarded yet.", proxyUserName)
```
reasonerjt: rawUsername.

src/core/filter/security.go
```diff
@@ -194,6 +205,114 @@ func (r *robotAuthReqCtxModifier) Modify(ctx *beegoctx.Context) bool {
...
+	err = json.Unmarshal(resRaw, &tokenReviewResponse)
+	if err != nil {
+		log.Errorf("fail to decode token review, %v", err)
+		return false
+	}
+	if rawUserName != tokenReviewResponse.Status.User.Username {
+		log.Errorf("user name doesn't match with token: %s", rawUserName)
+		return false
+	}
+	if !tokenReviewResponse.Status.Authenticated {
```
reasonerjt: You better first check `Authenticated` status.

src/core/filter/security.go
```diff
@@ -194,6 +205,114 @@ func (r *robotAuthReqCtxModifier) Modify(ctx *beegoctx.Context) bool {
...
+
+	httpAuthProxyConf, err := config.HTTPAuthProxySetting()
+	if err != nil {
+		log.Errorf("fail to get auth proxy settings, %v", err)
+		return false
+	}
+
+	// Init auth client with the auth proxy endpoint.
+	authClientCfg := &rest.Config{
+		Host: httpAuthProxyConf.Endpoint,
```
reasonerjt: We should add another attribute like `TokenReviewEndpoint`

src/core/filter/security.go
```diff
@@ -194,6 +205,114 @@ func (r *robotAuthReqCtxModifier) Modify(ctx *beegoctx.Context) bool {
...
+		Spec: k8s_api_v1beta1.TokenReviewSpec{
+			Token: proxyPwd,
+		},
+	}
+	res := authClient.Post().Body(tokenReviewRequest).Do()
+	if res.Error() != nil {
+		log.Errorf("fail to POST auth request, %v", err)
+		return false
+	}
+	resRaw, err := res.Raw()
```
reasonerjt: Let's double check the connection is closed

src/core/filter/security.go
```diff
@@ -194,6 +205,114 @@ func (r *robotAuthReqCtxModifier) Modify(ctx *beegoctx.Context) bool {
...
+		return false
+	}
+
+	match, rawUserName := ap.matchAuthProxyUserName(proxyUserName)
+	if !match {
+		log.Errorf("User name %s doesn't meet the auth proxy name pattern", proxyUserName)
+		return false
+	}
+
+	user, err := dao.GetUser(models.User{
```
reasonerjt: It would be better do call `GetUser` after the token is checked.

reasonerjt:
I think it looks good, would you add some UT in this PR?

PR #7206, opened 2019-03-22
Fix: upgrade pyyaml version to 4.2b1

ninjadq:
Because previous version has security issue


reasonerjt:
You should also pin the versions of jinja and click.

PR #7242, opened 2019-03-27
Add controller to handle oidc login

reasonerjt:
The controller will redirect user to the OIDC login page based on
configuration.
Additionally this commit add some basic code to wrap `oauth2` package
and `provider` in `go-oidc`, and fixed an issue in UT to make
InMemoryDriver for config management thread-safe.


src/core/controllers/oidc.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+// OIDCController handles requests for OIDC login, callback and user onboard
+type OIDCController struct {
+	api.BaseController
+}
+
+// RedirectLogin redirect user's browser to OIDC provider's login page
+func (oc *OIDCController) RedirectLogin() {
+	if mode, _ := config.AuthMode(); mode != common.OIDCAuth {
```
wy65701436: move StatusPreconditionFailed into prepare()

PR #7251, opened 2019-03-29
Oidc step1

jwangyangls:
OIDC step1---- configration
#7251 
image

In the configration auth mode section, we add an option, OIDC.

When the user logs in using OIDC mode, the system defaults to auth mode to select OIDC, where the user can modify the name, endpoint, scope, clientId, clientSecret and skipCertVerify of the OIDC. After the modification, the user clicks the Save button to save the changes.
#
Clicking this button will open https://openid.net/connect./
image
#

### DIFF
--- a/src/portal/src/i18n/lang/zh-cn-lang.json
+++ b/src/portal/src/i18n/lang/zh-cn-lang.json
@@ -65,13 +65,18 @@
         "ITEM_REQUIRED": "此项为必填项。",
         "NUMBER_REQUIRED": "此项为必填项且为数字。",
         "PORT_REQUIRED": "此项为必填项且为合法端口号。",
-        "CRON_REQUIRED": "此项为必填项且为cron格式。",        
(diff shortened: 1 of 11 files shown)


src/portal/src/i18n/lang/zh-cn-lang.json
```diff
@@ -682,6 +688,14 @@
             "ALWAYS_ONBOARD": "Always Onboard",
             "VERIFY_CERT": "Authentication验证证书"
         },
+        "OIDC": {
+            "OIDC_PROVIDER": "OIDC 供应商",
+            "ENDPOINT": "OIDC Endpoint",
+            "CLIENT_ID": "OIDC 客户端标识",
+            "CLIENTSECRET": "OIDC CLIENT密码",
```
reasonerjt: you should either translate `CLIENT` in both `CLIENT_ID` and `CLIENT_SECRET`, or not translate at all.

PR #7255, opened 2019-03-31
Implement native docker registry.

kofj:
Native adapter

### DIFF
--- a/src/replication/ng/adapter/native/adapter.go
+++ b/src/replication/ng/adapter/native/adapter.go
@@ -1,0 +1,89 @@
+package native
+
+import (
+	"errors"
+	"github.com/goharbor/harbor/src/common/utils/log"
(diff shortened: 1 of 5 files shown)


src/replication/ng/adapter/native/adapter.go
```diff
@@ -0,0 +1,68 @@
...
+			},
+			{
+				Type:  model.FilterTypeTag,
+				Style: model.FilterStyleTypeText,
+			},
+		},
+		SupportedTriggers: []model.TriggerType{
+			model.TriggerTypeManual,
+			model.TriggerTypeScheduled,
+			model.TriggerTypeEventBased,
```
ywk253100: Docker registry shouldn't support the event based trigger.

kofj: removed

PR #7268, opened 2019-04-01
Fix: clair env file should using empty string is not exist

ninjadq:
This is quick fix, further fixs will in the config refactor PR


reasonerjt:
non-blocking suggestion:
`{}.get('key','') == '' `

PR #7277, opened 2019-04-02
Refactor config file

ninjadq:
1. remove reload_config  [x]
2. customize_crt removed [x]
3. admiral removed it should added by vic team [x]
4. external clair db removed [x]
5. uaa config removed [x]
6. remove auth_mode [x]
7. remove email_setting [x]
8. remove ldap_url [x]
9. remove harbor_admin_password [x]
10. redis item name change [x]
11. internal database can only change password [x]
12. add external db [x]
13. add info level [x]
14. add external_url [x]
15. storage format changed [x]


make/photon/prepare/templates/nginx/nginx.http.conf.jinja
```diff
@@ -28,7 +28,7 @@ http {
   access_log /dev/stdout timed_combined;
 
   server {
-    listen 80;
+    listen {{http_port}};
```
wy65701436: should also update the ports in docker-compose.tpl

make/prepare
```diff
@@ -3,14 +3,24 @@
 # If compling source code this dir is harbor's make dir
```
wy65701436: add `set -e` to fail early.

make/prepare
```diff
@@ -23,9 +33,6 @@ mkdir -p $input_dir/common
 cp $ssl_cert_path $input_dir/nginx/server.crt
 cp $ssl_cert_key_path $input_dir/nginx/server.key
 
-# Copy secretkey to input dir
-cp -r $secretkey_path $input_dir/keys
-
 # Copy ca bundle to input dir
 if [ -f $registry_custom_ca_bundle ]
```
wy65701436: please pass the image tag($1) into prepare rather than hard-coding it to `goharbor/prepare:dev`

make/harbor.yml
```diff
@@ -1,114 +1,91 @@
...
-  port: 6379
-  # Redis connection password
-  password:
-  # Redis connection db index
-  # db_index 1,2,3 is for registry, jobservice and chartmuseum. 
-  # db_index 0 is for UI, it's unchangeable
-  db_index: 1,2,3
+# Harbor Storage settings
+storage_service:
+  # registry_custom_ca_bundle is the path to the custom root ca certificate, which will be injected into the truststore
```
reasonerjt: the comment is in correct. it's not `registry_custom_ca`

make/harbor.yml
```diff
@@ -1,114 +1,91 @@
...
+  # of registry's and chart repository's containers.  This is usually needed when the user hosts a internal storage with self signed certificate.
+  ca_bundle:
 
+  # storage backend, default is filesystem, options include filesystem, azure, gcs, s3, swift and oss
```
reasonerjt: We should explicitly tell user when they choose filesystem, what path will be used.

make/harbor.yml
```diff
@@ -1,114 +1,91 @@
...
+  location: /var/log/harbor
+
+#This attribute is for migrator to detect the version of the .cfg file, DO NOT MODIFY!
+_version: 1.7.0
 
-#If reload_config=true, all settings which present in harbor.yml take effect after prepare and restart harbor, it overwrites exsiting settings.
-#reload_config=true
-#Regular expression to match skipped environment variables
-#skip_reload_env_pattern: (^EMAIL.*)|(^LDAP.*)
+# Uncomment if using external database. Only support postgres.
```
reasonerjt: explicitly tell user when he chooses to use external user the password set under the item `database` will not be used.

make/prepare
```diff
@@ -19,12 +22,30 @@ mkdir -p $input_dir/nginx
 mkdir -p $input_dir/keys
 mkdir -p $input_dir/common
 
-# Copy nginx config file to input dir
-cp $ssl_cert_path $input_dir/nginx/server.crt
-cp $ssl_cert_key_path $input_dir/nginx/server.key
+# GCS key copy if exist
+gcs_keyfile_lines=$(grep '^[^#]*keyfile:' ${harbor_prepare_path}/harbor.yml)
```
reasonerjt: Reading a value from shell is not reliable.
We should double check in the shell script that there's only one matching line, it there's more than one we have to fail.

A potential fix is to transform the yml to json or other format so the shell can locate the attribute on host.

make/prepare
reasonerjt: +1 @ninjadq is there reason for this `set +e`?

make/prepare
reasonerjt: Please make sure this fails explicitly, when there are more than one line.

PR #7281, opened 2019-04-03
Add the OIDC onboard page

jwangyangls:
When a user logs in to Harbor for the first time through OIDC, the user will enter an onboard page, prompting the user to add the user name of Harbor. After the user name is entered, click save, and the user successfully logs in to Harbor through OIDC.
image

image


reasonerjt:
@jwangyangls let me clarify this is for onboarding users, which means after onboarding the dialog will not show for the same user.

PR #7286, opened 2019-04-03
Add controller to onboard oidc user

wy65701436:

### DIFF
--- a/src/core/controllers/oidc.go
+++ b/src/core/controllers/oidc.go
@@ -15,13 +15,18 @@
 package controllers
 
 import (
+	"encoding/json"
 	"fmt"
 	"github.com/goharbor/harbor/src/common"
+	"github.com/goharbor/harbor/src/common/dao"
+	"github.com/goharbor/harbor/src/common/models"
 	"github.com/goharbor/harbor/src/common/utils"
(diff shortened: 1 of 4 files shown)


reasonerjt:
Just dump and load serialized `oidcUserData` in session would be easier.

src/core/controllers/oidc.go
```diff
@@ -92,7 +95,42 @@ func (oc *OIDCController) Callback() {
...
+	}
+
+	oidcUser := models.OIDCUser{
+		SubIss: d.Subject + d.Issuer,
+		// TODO: get secret with secret manager.
+		Secret: utils.GenerateRandomString(),
+	}
+
+	user := models.User{
+		Username:     oc.GetString("username"),
```
reasonerjt: Should check the value of username, same rule as registration should apply.

src/core/controllers/oidc.go
```diff
@@ -92,7 +95,42 @@ func (oc *OIDCController) Callback() {
...
+		Email:        utils.GenerateRandomString() + "@harbor.com",
```
reasonerjt: Email can be read from the oidcUserData.

PR #7289, opened 2019-04-03
Add registry ping API

cd1989:

Add registry Ping API

### DIFF
--- a/src/common/utils/registry/registry.go
+++ b/src/common/utils/registry/registry.go
@@ -157,3 +157,23 @@ func (r *Registry) Ping() error {
 		Message: string(b),
 	}
 }
+
+// PingSimple checks whether the registry is available. It checks the connectivity and certificate (if TLS enabled)
+// only, regardless of credential.
+func (r *Registry) PingSimple() error {
+	err := r.Ping()
+	if err == nil {
+		return nil
+	}
+
+	httpErr, ok := err.(*commonhttp.Error)
+	if !ok {
+		return err
+	}
+
+	if httpErr.Code < 500 {
+		return nil
+	}
+
+	return httpErr
+}
--- a/src/core/api/registry.go
+++ b/src/core/api/registry.go
@@ -61,8 +61,8 @@ func (t *RegistryAPI) Ping() {
 		}
 	}
 
-	if len(r.URL) == 0 || r.Credential == nil {
-		t.CustomAbort(http.StatusBadRequest, "URL or credential emptry")
+	if len(r.URL) == 0 {
+		t.CustomAbort(http.StatusBadRequest, "URL can't be emptry")
 		return
 	}
 
@@ -72,7 +72,7 @@ func (t *RegistryAPI) Ping() {
 		return
 	}
 
-	if status != registry.Healthy {
+	if status != model.Healthy {
 		t.CustomAbort(http.StatusBadRequest, fmt.Sprintf("Ping registry %d failed", r.ID))
 	}
 	return
--- a/src/core/api/registry.go
+++ b/src/core/api/registry.go
@@ -40,6 +40,44 @@ func (t *RegistryAPI) Prepare() {
 	t.policyCtl = ng.PolicyCtl
 }
 
+// Ping checks health status of a registry
+func (t *RegistryAPI) Ping() {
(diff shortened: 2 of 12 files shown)


src/replication/ng/registry/manager.go
```diff
@@ -202,15 +202,16 @@ func (m *DefaultManager) HealthCheck() error {
...
+func CheckHealthStatus(r *model.Registry) (HealthStatus, error) {
 	// TODO(ChenDe): Support other credential type like OAuth, for the moment, only basic auth is supported.
 	if r.Credential.Type != model.CredentialTypeBasic {
 		return Unknown, fmt.Errorf("unknown credential type '%s', only '%s' supported yet", r.Credential.Type, model.CredentialTypeBasic)
 	}
 
 	// TODO(ChenDe): Support health check for other kinds of registry
 	if r.Type != model.RegistryTypeHarbor {
-		return Unknown, fmt.Errorf("unknown registry type '%s'", model.RegistryTypeHarbor)
+		return Unknown, fmt.Errorf("unknown registry type '%s'", r.Type)
```
ywk253100: As we discussed offline, we need to put the `heath check` or `ping` method to the `Adapter` interface and call the method when do ping.

src/core/api/registry.go
```diff
@@ -38,6 +38,44 @@ func (t *RegistryAPI) Prepare() {
...
+			return
+		}
+
+		if r == nil {
+			t.CustomAbort(http.StatusNotFound, fmt.Sprintf("Registry %d not found", id))
+			return
+		}
+	}
+
+	if len(r.URL) == 0 || r.Credential == nil {
```
ywk253100: I think the credential isn't necessary. For example, when I set up a replication policy that just pulling image from docker Hub.

src/replication/ng/registry/manager.go
cd1989: How about do it in another PR.

src/core/api/registry.go
cd1989: You are right, I will fix it.

src/replication/ng/registry/manager.go
ywk253100: Sure

cd1989: I will add Ping to Adapter in this PR.

cd1989:
All comment resolved. Please take a look @ywk253100

src/common/utils/registry/registry.go
```diff
@@ -157,3 +157,19 @@ func (r *Registry) Ping() error {
 		Message: string(b),
 	}
 }
+
+// PingSimple checks whether the registry is available. It checks the connectivity and certificate (if TLS enabled)
+// only, regardless of credential.
+func (r *Registry) PingSimple() error {
+	err := r.Ping()
```
ywk253100: check err != nil first?

src/replication/ng/adapter/adapter.go
```diff
@@ -67,6 +69,12 @@ func GetFactory(t model.RegistryType) (Factory, error) {
 	return factory, nil
 }
 
+// HasFactory checks whether there is given type adapter factory
```
ywk253100: Why we need this method? I think we can just call `adapter.GetFactory(r.Type)` to get the factory, if it doesn't exist, an error will be returned

src/replication/ng/adapter/harbor/adapter.go
```diff
@@ -178,6 +179,51 @@ func (a *adapter) GetNamespace(namespace string) (*model.Namespace, error) {
 	}, nil
 }
 
+func (a *adapter) HealthCheck() (model.HealthStatus, error) {
```
ywk253100: Put the method into the `DefaultImageRegistry` so that other registries can use it either.

src/common/utils/registry/registry.go
cd1989: Indeed.

src/replication/ng/adapter/adapter.go
cd1989: We can use GetFactory, but does an error of it always means not exist, I think it's not explicit in semantics.

PR #7304, opened 2019-04-04
Use policy controller in registry deletion

cd1989:

Use policy controller to get policies when delete registry.

And also refactor the GetPolicies method.

### DIFF
--- a/src/replication/ng/policy/manager/manager.go
+++ b/src/replication/ng/policy/manager/manager.go
@@ -156,25 +156,8 @@ func (m *DefaultManager) Create(policy *model.Policy) (int64, error) {
 
 // List returns all the policies
 func (m *DefaultManager) List(queries ...*model.PolicyQuery) (total int64, policies []*model.Policy, err error) {
-	// default query parameters
-	var name = ""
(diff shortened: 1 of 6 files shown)


src/replication/ng/policy/manager/manager.go
```diff
@@ -156,25 +156,20 @@ func (m *DefaultManager) Create(policy *model.Policy) (int64, error) {
...
-	var page int64 = 1
-	var pageSize int64 = 15
-
-	if len(queries) > 0 {
-		name = queries[0].Name
-		namespace = queries[0].Namespace
-		page = queries[0].Pagination.Page
-		pageSize = queries[0].Pagination.Size
+	// Generate policy query parameters, if pagination not set, set to default value.
+	query := &model.PolicyQuery{}
```
ywk253100: I don't think the default values for pagination are necessary as we have handle it in the API level. And we do need to list all policies  by some other internal component, the default values for pagination will make it complicated.

cd1989: Updated

PR #7328, opened 2019-04-09
Fix issue 6450 Test LDAP server error without save configuration

stonezdj:


src/core/api/ldap.go
```diff
@@ -51,6 +54,19 @@ func (l *LdapAPI) Prepare() {
...
+	if strings.EqualFold(l.Ctx.Request.RequestURI, "/api/ldap/ping") {
+		return
+	}
+	authMode, err := config.AuthMode()
+	if err != nil {
+		l.HandleInternalServerError(fmt.Sprintf("Can't load system configuration, error: %v", err))
+		return
+	}
+	if authMode != "ldap_auth" {
+		l.HandleInternalServerError(fmt.Sprintf("current auth_mode is not ldap_auth"))
```
reasonerjt: Have you tested this?  I'm afraid this doesn't work because the `return` will only return from `Prepare` and the request will be handled.

stonezdj: Verified, HandlerInternalServerError works in Prepare

src/core/api/ldap.go
```diff
@@ -52,6 +55,19 @@ func (l *LdapAPI) Prepare() {
...
+	if strings.EqualFold(l.Ctx.Request.RequestURI, "/api/ldap/ping") {
+		return
+	}
+	authMode, err := config.AuthMode()
+	if err != nil {
+		l.SendInternalServerError(fmt.Errorf("Can't load system configuration, error: %v", err))
+		return
+	}
+	if authMode != "ldap_auth" {
+		l.SendInternalServerError(fmt.Errorf("Can't load system configuration, error: %v", err))
```
reasonerjt: wrong error message　...

reasonerjt:
Please refine the error message and merge the change. @stonezdj

src/core/api/ldap.go
stonezdj: Update

PR #7331, opened 2019-04-09
Correct some typos and do some small adjustments

cd1989:

### DIFF
--- a/src/replication/ng/operation/flow/stage.go
+++ b/src/replication/ng/operation/flow/stage.go
@@ -59,8 +59,8 @@ func initialize(policy *model.Policy) (adp.Adapter, adp.Adapter, error) {
 
 // fetch resources from the source registry
 func fetchResources(adapter adp.Adapter, policy *model.Policy) ([]*model.Resource, error) {
-	resTypes := []model.ResourceType{}
-	filters := []*model.Filter{}
+	var resTypes []model.ResourceType
+	var filters []*model.Filter
 	for _, filter := range policy.Filters {
 		if filter.Type != model.FilterTypeResource {
 			filters = append(filters, filter)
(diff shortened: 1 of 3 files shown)


src/replication/ng/operation/flow/stage.go
```diff
@@ -59,8 +59,8 @@ func initialize(policy *model.Policy) (adp.Adapter, adp.Adapter, error) {
 
 // fetch resources from the source registry
 func fetchResources(adapter adp.Adapter, policy *model.Policy) ([]*model.Resource, error) {
-	resTypes := []model.ResourceType{}
-	filters := []*model.Filter{}
+	var resTypes []model.ResourceType
+	var filters []*model.Filter
```
cd1989: Use nil slices over empty slices: https://programming.guide/go/nil-slice-vs-empty-slice.html

src/replication/ng/operation/flow/stage.go
```diff
@@ -185,28 +186,28 @@ func filterResources(resources []*model.Resource, filters []*model.Filter) ([]*m
...
-		namespace, err := srcAdapter.GetNamespace(srcResource.Metadata.Namespace)
-		if err != nil {
-			return nil, err
-		}
-		namespaces = append(namespaces, namespace)
-	}
-
 	if len(dstNamespace) != 0 {
-		namespaces = []*model.Namespace{
+		return []*model.Namespace{
```
ywk253100: We have to set the metadata of destination namespace according to the metadata of source namespaces, so we must get the source namespaces first. And I have add a TODO in line 195.

src/replication/ng/operation/flow/stage.go
```diff
@@ -185,28 +186,28 @@ func filterResources(resources []*model.Resource, filters []*model.Filter) ([]*m
...
+	for _, srcResource := range srcResources {
+		namespace, err := srcAdapter.GetNamespace(srcResource.Metadata.Namespace)
+		if err != nil {
+			return nil, err
 		}
+		namespaces = append(namespaces, namespace)
 	}
 
-	log.Debug("assemble the destination namespaces completed")
+	log.Debugf("assemble the destination namespaces completed: %v", namespaces)
```
ywk253100: It will print just the address in memory rather than the names

cd1989: You are right

PR #7333, opened 2019-04-10
Check health status when add registry

cd1989:

Check health status when add registry, so that invalid registry won't be added.

### DIFF
--- a/src/core/api/registry.go
+++ b/src/core/api/registry.go
@@ -129,24 +129,24 @@ func (t *RegistryAPI) Ping() {
 func (t *RegistryAPI) Get() {
 	id := t.GetIDFromURL()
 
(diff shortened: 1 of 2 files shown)


src/core/api/registry.go
```diff
@@ -177,24 +177,34 @@ func (t *RegistryAPI) List() {
...
 		return
 	}
 
 	if reg != nil {
-		t.HandleConflict(fmt.Sprintf("name '%s' is already used", registry.Name))
+		t.HandleConflict(fmt.Sprintf("name '%s' is already used", r.Name))
+		return
+	}
+
+	status, err := registry.CheckHealthStatus(r)
```
ywk253100: Should check the status when updating the registry either.

PR #7335, opened 2019-04-10
OIDC E2E flow + secret support

reasonerjt:


src/core/filter/security.go
```diff
@@ -205,6 +207,46 @@ func (r *robotAuthReqCtxModifier) Modify(ctx *beegoctx.Context) bool {
...
 
+type oidcCliReqCtxModifier struct{}
+
+func (oc *oidcCliReqCtxModifier) Modify(ctx *beegoctx.Context) bool {
+	path := ctx.Request.URL.Path
+	if path != "/service/token" || strings.HasPrefix(path, "/chartrepo/") {
+		log.Debug("OIDC CLI modifer only handles request by docker CLI or helm CLI")
+		return false
+	}
+	authMode, err := config.AuthMode()
```
wy65701436: first to check auth mode should be better.

src/core/filter/security.go
```diff
@@ -205,6 +207,46 @@ func (r *robotAuthReqCtxModifier) Modify(ctx *beegoctx.Context) bool {
...
+	}
+	username, secret, ok := ctx.Request.BasicAuth()
+	if !ok {
+		return false
+	}
+
+	user, err := dao.GetUser(models.User{
+		Username: username,
+	})
+	if user == nil {
```
wy65701436: ser, err := dao.GetUser(models.User{
		Username: rawUserName,
	})
	if err != nil {
		log.Errorf("fail to get user: %v", err)
		return false
	}
	if user == nil {
		return false
}

src/common/utils/oidc/secret.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// VerifySecret verifies the secret and the token associated with it, it tries to update the token in the DB if it's
+// refreshed during the verification
+func (dm *defaultManager) VerifySecret(ctx context.Context, userID int, secret string) error {
+	key, err := dm.getEncryptKey()
+	if err != nil {
+		return fmt.Errorf("failed to load the key for encryption/decryption： %v", err)
+	}
+	oidcUser, err := dao.GetOIDCUserByUserID(userID)
```
wy65701436: first to check the user existing.

src/common/utils/oidc/secret.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// SetSecret sets the secret and token based on the ID of the user, when setting the secret the user has to be
+// onboarded to Harbor DB.
+func (dm *defaultManager) SetSecret(userID int, secret string, token *Token) error {
+	key, err := dm.getEncryptKey()
+	if err != nil {
+		return fmt.Errorf("failed to load the key for encryption/decryption： %v", err)
+	}
+	oidcUser, err := dao.GetOIDCUserByUserID(userID)
```
wy65701436: is it dup as dao.GetOIDCUserByUserID(userID) is called in VerifySecret()?

src/core/api/user.go
```diff
@@ -118,12 +118,19 @@ func (ua *UserAPI) Get() {
...
 			log.Errorf("Error occurred in GetUser, error: %v", err)
-			ua.CustomAbort(http.StatusInternalServerError, "Internal error.")
+			ua.RenderFormatedError(http.StatusInternalServerError, err)
+			return
 		}
 		u.Password = ""
 		if ua.userID == ua.currentUserID {
 			u.HasAdminRole = ua.SecurityCtx.IsSysAdmin()
 		}
+		o, err := ua.getOIDCUserInfo()
```
wy65701436: if ua.AuthMode != common.OIDCAuth {
		o, err := ua.getOIDCUserInfo() 
                ........
	}

src/common/utils/oidc/secret.go
```diff
@@ -0,0 +1,127 @@
...
+// VerifySecret verifies the secret and the token associated with it, it tries to update the token in the DB if it's
+// refreshed during the verification
+func (dm *defaultManager) VerifySecret(ctx context.Context, userID int, secret string) error {
+	key, err := dm.getEncryptKey()
+	if err != nil {
+		return fmt.Errorf("failed to load the key for encryption/decryption： %v", err)
+	}
+	oidcUser, err := dao.GetOIDCUserByUserID(userID)
+	if oidcUser == nil {
+		return fmt.Errorf("failed to get oidc user info, error: %v", err)
```
wy65701436: can we move the ORM into the Modifier? And let the dm just to set/verify secret and token?

src/core/filter/security.go
```diff
@@ -110,6 +111,7 @@ func Init() {
 	// standalone
 	reqCtxModifiers = []ReqCtxModifier{
 		&secretReqCtxModifier{config.SecretStore},
+		&oidcCliReqCtxModifier{},
```
wy65701436: we should notice that all of these modifiers are not in VIC mode.

src/common/utils/oidc/secret.go
```diff
@@ -0,0 +1,127 @@
...
+	tb, _ := json.Marshal(token)
+	encToken, _ := utils.ReversibleEncrypt(string(tb), key)
+	oidcUser.Secret = encSecret
+	oidcUser.Token = encToken
+	return dao.UpdateOIDCUser(oidcUser)
+}
+
+// VerifySecret verifies the secret and the token associated with it, it tries to update the token in the DB if it's
+// refreshed during the verification
+func (dm *defaultManager) VerifySecret(ctx context.Context, userID int, secret string) error {
```
wy65701436: the func name is so confusing as it does lots of things .....

src/common/utils/oidc/secret.go
reasonerjt: Yes when called by `VerifySecret()` there is a dup, I think I'll remove this `SetSecret` in a subsequent commit so simplify.

src/common/utils/oidc/secret.go
reasonerjt: But it has to happen together because part of the verification is refreshing the token.  And after we have a new token we want to store it to reduce future remote calls.

src/core/api/user.go
reasonerjt: it has it in the func `getOIDCUserInfo `

src/core/controllers/oidc.go
```diff
@@ -94,16 +100,33 @@ func (oc *OIDCController) Callback() {
...
-	oc.ServeFormatted()
+	u, err := dao.GetUserBySubIss(d.Subject, d.Issuer)
+	if err != nil {
+		oc.RenderFormatedError(http.StatusInternalServerError, err)
+		return
+	}
+	tokenBytes, err := json.Marshal(token)
+	if err != nil {
+		oc.RenderFormatedError(http.StatusInternalServerError, err)
+
```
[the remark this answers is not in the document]
reasonerjt: yes thanks

src/core/filter/security.go
reasonerjt: This is expected because in VIC the auth is delegated to admiral.

src/core/filter/security.go
reasonerjt: ok thanks.

src/core/filter/security.go
reasonerjt: checking authmode requires a db connection.

src/common/utils/oidc/secret.go
reasonerjt: ok

reasonerjt: fixed.

src/core/api/user.go
reasonerjt: fixed

src/core/controllers/oidc.go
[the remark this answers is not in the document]
reasonerjt: fixed.

src/core/filter/security.go
reasonerjt: fixed.

PR #7336, opened 2019-04-10
Change the login module

jwangyangls:
change the login module to fix the issue that the nav is being in page when we login harbor again,
and change harbor router .


DonAndrey:
Is this feature delivered in some harbor release?

PR #7342, opened 2019-04-11
Add Cli secret in Harbor user profile

jwangyangls:
image

user could copy the CLI secret when login through oidc, and using this cli secret as password when using docker/helm cli to access Harbor.


reasonerjt:
LGTM, for ts/html code @zhoumeina @pureshine please help review

PR #7384, opened 2019-04-15
fix deadlock when client not exist.

kofj:
When client not exist , will deadlock.


kofj:
have fixed, close.

kofj:
done.

PR #7388, opened 2019-04-15
Fix replication ui relate issues

pureshine:
1、Fix when clicking the edit policy, the source registry isn’t shown
2、Fix save button enable when not input rule name or source/destination namespaces
3、When editing the pull mode policy, generate filter and trigger according to the selected source registry.

### DIFF
--- a/src/portal/lib/src/create-edit-rule/create-edit-rule.component.ts
+++ b/src/portal/lib/src/create-edit-rule/create-edit-rule.component.ts
@@ -141,10 +141,8 @@ export class CreateEditRuleComponent implements OnInit, OnDestroy {
   }
 
   modeChange(): void {
-    if (this.isPushMode) {
-      this.setFilter([]);
-      this.initRegistryInfo(0);
-    }
+    this.setFilter([]);
+    this.initRegistryInfo(0);
   }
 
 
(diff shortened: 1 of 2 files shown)


src/portal/lib/src/create-edit-rule/create-edit-rule.component.ts
```diff
@@ -396,23 +400,29 @@ export class CreateEditRuleComponent implements OnInit, OnDestroy {
...
+        .subscribe((ruleInfo) => {
+          let srcRegistryId = ruleInfo.src_registry.id;
+          this.repService.getRegistryInfo(srcRegistryId)
+            .subscribe(adapter => {
+              this.setFilterAndTrigger(adapter);
+              this.copyUpdateForm = clone(ruleInfo);
+              // set filter value is [] if callback filter value is null.
+              this.updateForm(ruleInfo);
+              // keep trigger same value
+              this.copyUpdateForm.trigger = clone(ruleInfo.trigger);
```
zhoumeina: should moved to in the front of updateform

pureshine: It has been modified

PR #7392, opened 2019-04-15
Handle OIDC user invalidation from OIDC provider.

reasonerjt:
Ths commmit ensures that when user's token is invalidated OIDC provider, he
cannot access protected resource in Harbor with the user info in his session.
We share the code path with secret verification b/c the refresh token
can be used only once, so it has to be stored in one place.

### DIFF
--- a/src/common/utils/oidc/secret.go
+++ b/src/common/utils/oidc/secret.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"github.com/goharbor/harbor/src/common/dao"
+	"github.com/goharbor/harbor/src/common/models"
 	"github.com/goharbor/harbor/src/common/utils"
 	"github.com/goharbor/harbor/src/common/utils/log"
 	"github.com/goharbor/harbor/src/core/config"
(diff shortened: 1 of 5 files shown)


src/common/utils/oidc/secret.go
```diff
@@ -98,7 +80,18 @@ func (dm *defaultManager) VerifySecret(ctx context.Context, userID int, secret s
 	if secret != plainSecret {
 		return verifyError(errors.New("secret mismatch"))
 	}
-	tokenStr, err := utils.ReversibleDecrypt(oidcUser.Token, key)
+	return dm.VerifyToken(ctx, oidcUser)
+}
+
+func (dm *defaultManager) VerifyToken(ctx context.Context, user *models.OIDCUser) error {
```
wy65701436: please add some comments for VerifyToken, and its name is same as helper.VerfiyToken.

It confuses code reader.

src/common/utils/oidc/secret.go
```diff
@@ -116,15 +109,22 @@ func (dm *defaultManager) VerifySecret(ctx context.Context, userID int, secret s
 	if err != nil {
 		return verifyError(err)
 	}
-	err = dm.SetSecret(oidcUser.UserID, secret, t)
+	tb, _ := json.Marshal(t)
```
wy65701436: what's the reason to eat these errors? At least we can have logs.

src/core/filter/security.go
```diff
@@ -144,6 +147,20 @@ type ReqCtxModifier interface {
 	Modify(*beegoctx.Context) bool
 }
 
+// configCtxModifier populates to the configuration values to context, which are to be read by subsequent
+// filters.
+type configCtxModifier struct {
+}
+
+func (c *configCtxModifier) Modify(ctx *beegoctx.Context) bool {
+	m, err := config.AuthMode()
```
wy65701436: the AuthMode will return a `db_auth` as default value when encountering any error. 
And the request will be treated as a `db_auth` request in following, is it as expected?

reasonerjt: When error happens it will return `""`
https://github.com/goharbor/harbor/blob/master/src/common/config/manager.go#L193

src/common/utils/oidc/secret.go
reasonerjt: It returns error when the `t` can not be encoded to json, which is not true in this case, so it's ok to eat the error.
Anyway there's no harm add a log

src/common/utils/oidc/secret.go
reasonerjt: I'll add some comment.
But actually there's no need to add comment because it's a func of a struct that is not exposed.

Issue #6808, opened 2019-01-23
[UI] Create new UI to display the new replication report


steven-zou:
What can we help you?

steven-zou:
@zhoumeina 

Please refer to https://github.com/goharbor/community/blob/master/proposals/Content-Replication-ng.md#replication-policy-and-execution for more details.

pureshine:
don as pr #7393 #7224 #7172

PR #7393, opened 2019-04-15
Fix replication frontend issues

### LINKED ISSUE #7372: Wrong error message is shown to users when creating registries
The error message and the real error don't match.
create_registry

pureshine:
fix #7372 #6808 
1、Fix wrong error message is shown to users when creating registries
2、Add an override field when creating a policy
3、Execution lists and details pages show total and stopped information

### DIFF
--- a/src/portal/lib/src/replication/replication-tasks/replication-tasks.component.ts
+++ b/src/portal/lib/src/replication/replication-tasks/replication-tasks.component.ts
@@ -1,18 +1,20 @@
-import { Component, OnInit, Input } from '@angular/core';
+import { Component, OnInit, Input, OnDestroy } from '@angular/core';
 import { Router } from '@angular/router';
 import { ReplicationService } from "../../service/replication.service";
 import { TranslateService } from '@ngx-translate/core';
 import { finalize } from "rxjs/operators";
+import { Subscription, timer } from "rxjs";
 import { ErrorHandler } from "../../error-handler/error-handler";
 import { ReplicationJob, ReplicationTasks, Comparator, ReplicationJobItem, State } from "../../service/interface";
 import { CustomComparator, DEFAULT_PAGE_SIZE, calculatePage, doFiltering, doSorting } from "../../utils";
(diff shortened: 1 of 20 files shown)


src/portal/lib/src/create-edit-endpoint/create-edit-endpoint.component.html
```diff
@@ -41,7 +41,7 @@ <h3 class="modal-title">{{modalTitle}}</h3>
           <textarea type="text" class="inputWidth" row=3 name="description" [(ngModel)]="target.description"></textarea>
         </div>
         <!-- Endpoint Url -->
-        <div class="form-group">
+        <div class="form-group" [hidden]="target.type === 'dockerHub'">
```
kofj: Some registry not support too, like huawei SWR.

pureshine: So how should I judge whether it supports it?

kofj: We'll discuss this tonight in the Replication Workgroup meeting up.

pureshine: OK ,then this pr is temporarily not making changes here.

src/portal/lib/src/replication/replication-tasks/replication-tasks.component.ts
```diff
@@ -128,6 +139,24 @@ export class ReplicationTasksComponent implements OnInit {
         this.totalCount = res.length;
         this.tasks = res; // Keep the data
         this.taskItem = this.tasks.filter(tasks => tasks.resource_type !== "");
+        if (!this.timerDelay) {
+          this.timerDelay = timer(10000, 10000).subscribe(() => {
+            let count: number = 0;
+            this.tasks.forEach(tasks => {
+              if (
+                tasks.status === "InProgress"
```
zhoumeina: abstract to a const.

PR #7414, opened 2019-04-17
Optimizing native adapter

kofj:
R.T.

### DIFF
--- a/src/replication/adapter/native/image_registry.go
+++ b/src/replication/adapter/native/image_registry.go
@@ -15,10 +15,8 @@
 package native
 
 import (
-	"net/http"
 	"strings"
 
-	common_http "github.com/goharbor/harbor/src/common/http"
 	adp "github.com/goharbor/harbor/src/replication/adapter"
 	"github.com/goharbor/harbor/src/replication/model"
 	"github.com/goharbor/harbor/src/replication/util"
@@ -42,7 +40,7 @@ func (n native) FetchImages(filters []*model.Filter) ([]*model.Resource, error) 
 		return nil, err
 	}
 
-	resources := []*model.Resource{}
+	var resources []*model.Resource
 	for _, repository := range repositories {
 		tags, err := n.filterTags(repository, tagFilterPattern)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


src/common/utils/registry/repository.go
```diff
@@ -99,13 +101,6 @@ func (r *Repository) ListTag() ([]string, error) {
 		sort.Strings(tags)
 		tags = tagsResp.Tags
 
-		return tags, nil
-	} else if resp.StatusCode == http.StatusNotFound {
-
```
ywk253100: The issue of distribution is not fixed yet, why we remove the workaround here?

src/replication/adapter/native/image_registry.go
```diff
@@ -63,24 +62,21 @@ func (n native) FetchImages(filters []*model.Filter) ([]*model.Resource, error)
 		})
 	}
 
+	// replication execution should be failed if no any image found,
+	// and show the reason to user.
+	if resources == nil {
+		return nil, fmt.Errorf("Not find any image of name filter: %s, tag filter: %s", nameFilterPattern, tagFilterPattern)
```
ywk253100: Shouldn't return error here, if no resources returned, the execution will skip.

src/common/utils/registry/repository.go
kofj: This workaround will hide error message when it is true. Can we 
just hidden error in the UI to fix that's issue of harbor.

src/replication/adapter/native/image_registry.go
kofj: But the execution status will become success if nothing to do. User maybe not know reason that's execution is success but task list is empty.

ywk253100: The status text will be filled with message : there is no resources to replicate.

src/common/utils/registry/repository.go
ywk253100: No. The method is used by other functions and it's a bug of registry API, we should not expose it to upper callers

kofj: 👌, I have reverted.

PR #7421, opened 2019-04-17
Fix create ns failure in DockerHub

cd1989:

Issue: #7373


src/replication/adapter/dockerhub/adapter.go
```diff
@@ -198,6 +198,9 @@ func (a *adapter) CreateNamespace(namespace *model.Namespace) error {
 // getNamespace get namespace from DockerHub, if the namespace not found, two nil would be returned.
 func (a *adapter) getNamespace(namespace string) (*model.Namespace, error) {
 	resp, err := a.client.Do(http.MethodGet, getNamespacePath(namespace), nil)
+	if resp.StatusCode == http.StatusNotFound {
```
ywk253100: Shouldn't we check the err first? If not, we will got nil pointer exception here

cd1989: My mistake.

PR #7446, opened 2019-04-18
Update k8s-prepare

harryge00:
remove an unused char


reasonerjt:
Sorry but that script has been deprecated and will be removed soon...

harryge00:
@reasonerjt so how to deploy on K8s now?

Issue #7296, opened 2019-04-04
Cannot stop a pending job


ywk253100:
When trying to stop the job which is not running, got the error:
```
 [ERROR] [handler.go:253]: Serve http request 'POST /api/v1/jobs/734a11140d939ef700889725' error: 500 {"code":10008,"message":"Stop job failed with error","details":"job '734a11140d939ef700889725' is not a running job"}
```

ywk253100:
This impacts the `Stop execution` feature of replication

steven-zou:
Fixed. Close issue now.

Issue #7237, opened 2019-03-27
Job service is consuming the schedule in memory instead of redis


wy65701436:

reasonerjt:
@wy65701436 could you add more details regarding what issue it will cause.

steven-zou:
To avoid to retrieve policies every time from the backend with large traffic, the policies will be still kept into the mem list and sync with the backend by sub/pub.

Issue #7064, opened 2019-03-04
Improve the status webhook mechanism of job service to avoid ”zombie“ tasks


steven-zou:
Possible two ways:
* Improve the webhook to make it more stable
* Add status sync in the core side to double confirm the status syncing

steven-zou:
done!

Issue #7039, opened 2019-02-27
Jobservice intermittent restarting: "fatal error: concurrent map writes"


gunboe:
Jobservice is eventually rebooting. My system has a heavy write usage that combined with Clair's Scan scheduling and a replication to Harbor mirror, increases intermittent Jobservice reboots.

Look at Jobservice Logs:
```
-- Scan begin:
Feb 27 02:00:00 172.19.0.1 jobservice[1273]: 2019-02-27T05:00:00Z [INFO] Job incoming: IMAGE_SCAN_ALL:e0be25d2ee9ce0a579c7360f
Feb 27 02:00:00 172.19.0.1 jobservice[1273]: 2019-02-27T05:00:00Z [INFO] Scanning all the images in the registry
Feb 27 02:00:00 172.19.0.1 jobservice[1273]: 2019-02-27T05:00:00Z [INFO] Calling harbor-core API to scan image, 10216-autentikus/nginx-estatico:1
Feb 27 02:00:01 172.19.0.1 jobservice[1273]: 2019-02-27T05:00:01Z [INFO] Receive event 'register_hook' with data(unformatted): map[string]interface {}{"
job_id":"5e7a96c2dc4f69997166dbd9", "hook_url":"http://core:8080/service/notifications/jobs/scan/5743458"}
Feb 27 02:00:01 172.19.0.1 jobservice[1273]: 2019-02-27T05:00:01Z [INFO] Calling harbor-core API to scan image, 10216-autentikus/nginx-estatico:4
Feb 27 02:00:01 172.19.0.1 jobservice[1273]: 2019-02-27T05:00:01Z [INFO] Receive event 'register_hook' with data(unformatted): map[string]interface {}{"
job_id":"4f0a185414495f52a7ffa326", "hook_url":"http://core:8080/service/notifications/jobs/scan/5743459"}
Feb 27 02:23:19 172.19.0.1 jobservice[1273]: 2019-02-27T05:23:19Z [INFO] Scanning Layer: 652d56b59dcdba77d6d8237b46055f55627e5bf8aaad7e2e864f716d2673c588, path: http://registry:5000/v2/raia/raia-backend/blobs/sha256:ece7b24cf39d6e51f1bd7d53bc20942225c219b6840b79bf352ace8e2325c790
[...]

PR #7452, opened 2019-04-19
Fix issues for jobservice

steven-zou:
* Job status is feezed to a specified one (most times are ‘pending’) and is not changed anymore (status out of sync between job svc and core) (#6415, #7039, #7064)
  * enhance status webhook
  * enhance/simplify job stats persistence flow
* Failed to stop/cancel jobs (#7296)
  * remove command synchronizing channel across multiple instances
  * enhance the conditions of locating jobs which can be stopped/canceled
* Time interval precision of periodical jobs (#7237)
  * schedule periodical job instances when creating a periodical job if it already matches the scheduled judgment
* Extend the lifetime of job stats to 6 months
* Add retry to hook events sending and job status switching and persisting
* Remove the support of canceling and retrying jobs
* Remove the support of launching sub jobs inside one job


src/jobservice/config.yml
```diff
@@ -29,17 +29,14 @@ job_loggers:
   - name: "FILE"
     level: "DEBUG"
     settings: # Customized settings of logger
-      base_dir: "/tmp/job_logs"
+      base_dir: "/Users/szou/tmp/job_logs"
```
heww: should be revert back

reasonerjt:
Why `dao_test.go` is removed?
Why goharbor/harbor is copied to `vendor` folder?

steven-zou:
@heww @wy65701436 @wy65701436 

All the comments here will be fixed with another new PR tomorrow.

PR #7459, opened 2019-04-19
fix error message when auth_mod is not ldap 

stonezdj:

### DIFF
--- a/src/core/api/ldap.go
+++ b/src/core/api/ldap.go
@@ -65,7 +65,7 @@ func (l *LdapAPI) Prepare() {
 		return
 	}
 	if authMode != "ldap_auth" {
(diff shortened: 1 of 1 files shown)


src/core/api/ldap.go
```diff
@@ -65,7 +65,7 @@ func (l *LdapAPI) Prepare() {
 		return
 	}
 	if authMode != "ldap_auth" {
-		l.SendInternalServerError(fmt.Errorf("Can't load system configuration, error: %v", err))
+		l.SendInternalServerError(errors.New("system auth_mode isn't ldap_auth, please check configuration"))
```
reasonerjt: Shouldn't this be a `412` rather than `500` ?

PR #7460, opened 2019-04-19
Fix bug found in replication

ywk253100:
Call filterResources direclty if resource isn't null in copy flow


ywk253100:
This PR also fixes the issue #7461.

Issue #6876, opened 2019-01-31
v1.7.0 : tags listing => 504 gateway timeout


nvacelet:
I have large amount of tags related to one image (tags > 100). Problem is I am getting "504 Gateway Time-out" when I try to list tags from UI or from API. 

How can I increase timeout ? Or please can you add timeout setting update in new version ? 
Thx

clouderati:
My first instinct is that this has something to do with the LB configuration. Will have to investigate.

oxygenxo:
We have quite a big repositories (hundreds of tags, we're using them in CI) and we're unable to fetch them even just to delete them :) Not only on v1.7.0 though, can confirm that v1.6.0 is affected too.
Is there a way to increase request timeout?

oxygenxo:
Interesting behavior. We're using Harbor.Tagd to clean old images from Harbor. When it tries to fetch all tags it raises exception with '504 Gateway Timeout', but subsequent runs went fine. It took 10-15 runs for me to retrieve all tags from all repositories. May be some cache involved?

reasonerjt:
@nvacelet @oxygenxo what version of Harbor are you using?

oxygenxo:
Hi @reasonerjt , I've experienced this on v1.6.0 and then when I've upgraded to v1.7.4 too

lf1029698952:
same issue
+1

lf1029698952:
I try add timeout options in nginx.conf
```
      proxy_connect_timeout 900s;
      proxy_send_timeout 900s;
      proxy_read_timeout 900s;
```
but It seems useless.
when get repo tags > 100,  the harbor api server will  spend a long time until timeout

![image](https://user-images.githubusercontent.com/11751263/56415023-fda0e880-62be-11e9-8268-ef2d90d147b8.png)

cd1989:
There is a `detail` parameter, but doesn't take effect actually.

https://xxx/api/repositories/release/patroni/tags?detail=0

For tagd to work, maybe no need to fetch details.

pgkulk:
I was getting similar error for repository having around 9000 tags. 
Adding proxy_send_timeout and proxy_read_timeout to /api/ block in nginx.conf resolved the issues.

location /api/ {
      proxy_pass http://core/api/;
      proxy_set_header Host $host;
      proxy_set_header X-Real-IP $remote_addr;
      proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;

[...]

PR #7462, opened 2019-04-19
List simple tags when detail set to false

cd1989:

Enable the `detail` parameter when list tags. In some cases, only tag name is cared, no need to spend long time to fetch detail, vulnerability scanning result, notary sign results.

Related issues: https://github.com/goharbor/harbor/issues/6876

### DIFF
--- a/src/core/api/repository.go
+++ b/src/core/api/repository.go
@@ -576,9 +576,29 @@ func (ra *RepositoryAPI) GetTags() {
 		tags = ts
 	}
 
+	detail, err := ra.GetBool("detail", true)
+	if !detail && err == nil {
+		ra.Data["json"] = simpleTags(tags)
+		ra.ServeJSON()
+		return
+	}
+
 	ra.Data["json"] = assembleTagsInParallel(client, repoName, tags,
 		ra.SecurityCtx.GetUsername())
 	ra.ServeJSON()
+}
+
+func simpleTags(tags []string) []*models.TagResp {
+	var tagsResp []*models.TagResp
+	for _, tag := range tags {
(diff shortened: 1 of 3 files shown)


src/core/api/repository.go
```diff
@@ -599,11 +599,31 @@ func (ra *RepositoryAPI) GetTags() {
 		tags = ts
 	}
 
+	detail := ra.GetString("detail", "1")
```
kofj: Why not use `GetBool` ?

cd1989: Changed to bool

kofj: great!

ywk253100:
@cd1989 Please rebase the PR as some data models have changed.

reasonerjt:
Will you update the api doc? @cd1989

reasonerjt:
And this seems a break change... 
Can we make it backward-compatible?

cd1989:
> Will you update the api doc?

src/portal/lib/src/service/tag.service.ts
```diff
@@ -140,7 +140,7 @@ export class TagDefaultService extends TagService {
       queryParams = queryParams = new RequestQueryParams();
     }
 
-    queryParams = queryParams.set("detail", "1");
+    queryParams = queryParams.set("detail", "true");
```
reasonerjt: Seems we don't need this line because by default the api will return details by default?

cd1989: Yes, it's the default behavior, but I think it's clearer to set it implicitly.

cd1989:
@reasonerjt CI passed, please check it again.

PR #7479, opened 2019-04-22
Add document for config user settings

stonezdj:


reasonerjt:
@stonezdj I think when triggering the API all values have to be `string`, correct?

Additionally, could you provide one or two concrete example of curl command?  such as 
```
Update Harbor to user LDAP auth, and restrict project creation to admin only, and update the token expiration time
```

stonezdj:
> @stonezdj I think when triggering the API all values have to be `string`, correct?
> 
> Additionally, could you provide one or two concrete example of curl command? such as
> 
> ```
> Update Harbor to user LDAP auth, and restrict project creation to admin only, and update the token expiration time
> ```

Everything can be string, include bool and integer, it can be parsed to the correct type

PR #7482, opened 2019-04-22
Fix chart upload issue on event based

wy65701436:
Use chart API to load the uploaded chart file to get the name and version

### DIFF
--- a/src/chartserver/reverse_proxy.go
+++ b/src/chartserver/reverse_proxy.go
@@ -92,7 +92,7 @@ func modifyResponse(res *http.Response) error {
 		chartUploadEvent := res.Request.Context().Value(common.ChartUploadCtxKey)
 		e, ok := chartUploadEvent.(*rep_event.Event)
 		if !ok {
-			hlog.Errorf("failed to convert chart upload context to replication event, which could lead to no execution of event based chart replication policy")
+			hlog.Error("failed to convert chart upload context into replication event.")
 		} else {
 			// Todo: it used as the replacement of webhook, will be removed when webhook to be introduced.
 			go func() {
--- a/src/chartserver/reverse_proxy.go
+++ b/src/chartserver/reverse_proxy.go
@@ -92,7 +92,7 @@ func modifyResponse(res *http.Response) error {
 		// means this response is for uploading chart success.
 		chartUpload := res.Request.Context().Value(common.ChartUploadCtxKey).(string)
 		if chartUpload != "" {
-			chartUploadSplitted := strings.Split(chartUpload, ":")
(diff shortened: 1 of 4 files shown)


src/core/api/chart_repository.go
```diff
@@ -296,30 +296,13 @@ func (cra *ChartRepositoryAPI) UploadChartVersion() {
 			cra.SendInternalServerError(err)
 			return
 		}
+		if err := cra.addEventContext(formFiles, cra.Ctx.Request); err != nil {
+			hlog.Infof("Failed to add chart upload context into request, which could lead to no execution of event based chart replication policy, %v", err)
```
steven-zou: The event may be used by other non-replication components in the future, right? The message does not need to be so detailed.

src/core/api/chart_repository.go
```diff
@@ -428,6 +411,43 @@ type formFile struct {
...
+				hlog.Infof("failed to read file content for upload event, %v", err)
+				return err
+			}
+			chartOpr := chartserver.ChartOperator{}
+			chartDetails, err := chartOpr.GetChartData(Buf.Bytes())
+			if err != nil {
+				hlog.Infof("failed to get chart content for upload event, %v", err)
+				return err
+			}
+			*request = *(request.WithContext(context.WithValue(request.Context(), common.ChartUploadCtxKey, cra.namespace+
```
steven-zou: Please use standard object to replace the serial strings

src/core/api/chart_repository.go
```diff
@@ -296,30 +298,13 @@ func (cra *ChartRepositoryAPI) UploadChartVersion() {
 			cra.SendInternalServerError(err)
 			return
 		}
+		if err := cra.addEventContext(formFiles, cra.Ctx.Request); err != nil {
+			hlog.Errorf("Failed to add chart upload context into request, which could lead to no execution of event based chart replication policy, %v", err)
```
steven-zou: change the message to general not replication specified

src/chartserver/reverse_proxy.go
```diff
@@ -88,32 +87,21 @@ func director(target *url.URL, cred *Credential, req *http.Request) {
...
-					}
-					if err := replication.EventHandler.Handle(e); err != nil {
-						hlog.Errorf("failed to handle event: %v", err)
-					}
-				}()
-			}
+		chartUploadEvent := res.Request.Context().Value(common.ChartUploadCtxKey)
+		e, ok := chartUploadEvent.(*rep_event.Event)
+		if !ok {
+			hlog.Errorf("failed to convert chart upload context to replication event, which could lead to no execution of event based chart replication policy")
```
steven-zou: change the message to the general one, not the specified one

PR #7503, opened 2019-04-24
Check the task status of execution whose status is running when deleting the policy

ywk253100:
Check the task status of execution whose status is running when deleting the policy

### DIFF
--- a/src/core/api/replication_policy.go
+++ b/src/core/api/replication_policy.go
@@ -221,25 +221,64 @@ func (r *ReplicationPolicyAPI) Delete() {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


src/core/api/replication_policy.go
```diff
@@ -242,6 +238,48 @@ func (r *ReplicationPolicyAPI) Delete() {
...
+		for _, task := range tasks {
+			if isTaskRunning(task) {
+				return true, nil
+			}
+		}
+	}
+	return false, nil
+}
+
+func isTaskRunning(task *models.Task) bool {
```
wy65701436: please either rename the func or add more comments, as it's not just want to check running only.

Issue #7455, opened 2019-04-19
The change in "verify_cert" attribute does not take effect for http proxy auth mode.


reasonerjt:
Right now the cilent is initialized only once, it should reuse the transport based on the latest configuration value.

ericvmw:
also a relevant observation: the initial set on the configuration to be false via API doesn't seem to take effect while other configurations are updated.

reasonerjt:
Fixed in #7504

PR #7504, opened 2019-04-24
Update Transport of HTTP cient in auth proxy client

reasonerjt:
This commit ensures that the TLS config of the HTTP client for auth
proxy is updated when the configuration is changed.


reasonerjt:
Fixes #7455

PR #7522, opened 2019-04-26
Return immediately after creating the execution record when starting the replication

ywk253100:
When fetching resources is slow, the starting replication request may timeout, this commit returns immediately after creating the execution record when starting the replication

### DIFF
--- a/src/replication/operation/controller.go
+++ b/src/replication/operation/controller.go
@@ -42,16 +42,26 @@ type Controller interface {
 	GetTaskLog(int64) ([]byte, error)
 }
 
+const (
+	maxReplicators = 1024
+)
+
 // NewController returns a controller implementation
 func NewController(js job.Client) Controller {
-	return &controller{
+	ctl := &controller{
+		replicators:  make(chan struct{}, maxReplicators),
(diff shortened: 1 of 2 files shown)


src/replication/operation/controller.go
```diff
@@ -68,24 +68,24 @@ func (c *controller) StartReplication(policy *model.Policy, resource *model.Reso
...
-		if e := c.executionMgr.Update(&models.Execution{
-			ID:         id,
-			Status:     models.ExecutionStatusFailed,
-			StatusText: err.Error(),
-			Total:      n,
-			Failed:     n,
-		}, "Status", "StatusText", "Total", "Failed"); e != nil {
-			log.Errorf("failed to update the execution %d: %v", id, e)
+	go func() {
+		flow := c.createFlow(id, policy, resource)
```
steven-zou: Suggest to limit the number of go routine

```go
const maxRequests = 1024
tokens:=make(chan bool, maxRequests)

for i:=0;i<maxRequests;i++{
  tokens<-true
}

go func(){
  <-token
 doSomething(){
  defer func(){
    token<-true // return
  }()
 }
}()
```

PR #7530, opened 2019-04-26
add periodic job UUID to upstream job id and use execution log as the…

wy65701436:
… periodic log

### DIFF
--- a/src/common/job/client.go
+++ b/src/common/job/client.go
@@ -10,6 +10,7 @@ import (
 	commonhttp "github.com/goharbor/harbor/src/common/http"
 	"github.com/goharbor/harbor/src/common/http/modifier/auth"
 	"github.com/goharbor/harbor/src/common/job/models"
+	"github.com/goharbor/harbor/src/jobservice/job"
 )
 
 // Client wraps interface to access jobservice.
@@ -17,6 +18,7 @@ type Client interface {
 	SubmitJob(*models.JobData) (string, error)
 	GetJobLog(uuid string) ([]byte, error)
 	PostAction(uuid, action string) error
+	GetExecutions(uuid string) ([]job.Stats, error)
 	// TODO Redirect joblog when we see there's memory issue.
 }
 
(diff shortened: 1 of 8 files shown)


src/common/job/client.go
```diff
@@ -103,6 +105,36 @@ func (d *DefaultClient) GetJobLog(uuid string) ([]byte, error) {
 	return data, nil
 }
 
+// GetExecutions ...
+func (d *DefaultClient) GetExecutions(periodicJobID string) ([]job.Stats, error) {
+	url := d.endpoint + "/api/v1/jobs/" + periodicJobID + "/executions"
```
steven-zou: how about trying 'fmt.Sprintf("%s/%s/%s",1,1,1)'

src/common/job/client.go
```diff
@@ -103,6 +105,36 @@ func (d *DefaultClient) GetJobLog(uuid string) ([]byte, error) {
...
+	req, err := http.NewRequest(http.MethodGet, url, nil)
+	if err != nil {
+		return nil, err
+	}
+	resp, err := d.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
```
steven-zou: maybe better after error checking?

src/common/job/client_test.go
```diff
@@ -47,6 +47,14 @@ func TestGetJobLog(t *testing.T) {
 	assert.Contains(text, "The content in this file is for mocking the get log api.")
 }
 
+func TestGetExecutions(t *testing.T) {
+	assert := assert.New(t)
+	exes, err := testClient.GetExecutions(ID)
+	assert.Nil(err)
+	stat := exes[0]
+	assert.Equal(ID, stat.Info.JobID)
```
steven-zou: seems not `==`

src/common/job/models/models.go
```diff
@@ -28,25 +28,28 @@ type JobMetadata struct {
 
 // JobStats keeps the result of job launching.
 type JobStats struct {
```
steven-zou: Why NOT directly import job.*?

src/core/api/admin_job.go
```diff
@@ -169,8 +169,22 @@ func (aj *AJAPI) getLog(id int64) {
...
+		if err != nil {
+			aj.SendInternalServerError(err)
+			return
+		}
+		if len(exes) == 0 {
+			aj.SendNotFoundError(errors.New("no execution log "))
+			return
+		}
+		jobID = exes[0].Info.JobID
+	}
```
steven-zou: add else to void the non-necessary case?

src/common/job/client.go
wy65701436: will fix, thanks.

src/common/job/client_test.go
wy65701436: update the mock response

src/core/api/admin_job.go
wy65701436: already has

PR #7546, opened 2019-04-26
Fix vulnerability doesn't call systeminfo api by default

### LINKED ISSUE #7511: vulnerability doesn't call /api/systeminfo to get DB ready time by default
image

pureshine:
fix #7511 #7589 
1、Fix vulnerability doesn't call /api/systeminfo to get DB ready time by default
2、Add a refresh button on the gc history page
image


src/portal/lib/src/config/gc/gc-history/gc-history.component.scss
```diff
@@ -1,4 +1,20 @@
 .history-header {
     color: #000;
     margin:20px 0 6px 0;
-}
\ No newline at end of file
+}
+
+.rightPos {
```
zhoumeina: use "-" style to naming the css style rule.

zhoumeina: use mixin

Issue #7568, opened 2019-04-28
OICC authentication returns 500 when client secret is wrong


zhoumeina:
It will returns 500 when user click "grant access" button when user set wrong client secret. At this time user does not know what happened. From the log we can see: 
![image](https://user-images.githubusercontent.com/4946616/56856689-601a7880-6992-11e9-9b71-c5f6a1656c36.png)
We expected it will return the warning message said "client secret is wrong"

reasonerjt:
I'll update it to 4xx error so the error from OIDC provider will be included, however the level of detail of the error depends on the error returned by OIDC provider.

reasonerjt:
Fixed in #7590

PR #7590, opened 2019-04-29
Return more details for error in exchange token

reasonerjt:
This commit update the response off OIDC callback when there's error in exchange token.
Additionally add comments to clarify that by default 500 error will not
contain any details.


reasonerjt:
Fixes #7568

PR #7612, opened 2019-04-30
Fix tls related issues

ninjadq:


make/photon/prepare/utils/nginx.py
```diff
@@ -25,6 +30,9 @@ def render_nginx_template(config_dict):
             ssl_cert=SSL_CERT_PATH,
             ssl_cert_key=SSL_CERT_KEY_PATH)
         location_file_pattern = CUSTOM_NGINX_LOCATION_FILE_PATTERN_HTTPS
+        cert_dir.mkdir(parents=True, exist_ok=True)
+        ssl_cert_key.touch()
```
reasonerjt: We don't need touch as per #7649

reasonerjt:
Please narrow down the usage of `bind` to make sure they only apply to the paths that is set by user in `harbor.yml`

PR #7624, opened 2019-04-30
Prepare for harbor tile

ninjadq:


make/harbor.yml
```diff
@@ -45,6 +45,9 @@ data_volume: /data
 #   # for more info about this configuration please refer https://docs.docker.com/registry/configuration/
 #   filesystem:
 #     maxthreads: 100
+#   # Uncomment following when you like configure disable
```
reasonerjt: why there are two `#` ?

ninjadq: Because there are two levels of comments.

PR #7628, opened 2019-04-30
add scan all and gc schedule migration

wy65701436:

### DIFF
--- a/src/common/utils/utils.go
+++ b/src/common/utils/utils.go
@@ -26,6 +26,7 @@ import (
 	"strings"
 	"time"
 
+	common_model "github.com/goharbor/harbor/src/common/models"
 	"github.com/goharbor/harbor/src/common/utils/log"
 )
 
@@ -212,6 +213,23 @@ func ParseOfftime(offtime int64) (hour, minite, second int) {
 	minite = int(offtime / 60)
 	second = int(offtime % 60)
 	return
+}
+
+// ParseScheduleParamToCron ...
+func ParseScheduleParamToCron(param *common_model.ScheduleParam) string {
+	if param == nil {
+		return ""
+	}
+	offtime := param.Offtime
+	offtime = offtime % (3600 * 24)
+	hour := int(offtime / 3600)
+	offtime = offtime % 3600
+	minute := int(offtime / 60)
+	second := int(offtime % 60)
+	if param.Type == "Weekly" {
+		return fmt.Sprintf("%d %d %d * * %d", second, minute, hour, param.Weekday%7)
+	}
+	return fmt.Sprintf("%d %d %d * * *", second, minute, hour)
 }
 
 // TrimLower ...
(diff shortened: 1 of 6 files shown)


src/common/utils/utils.go
```diff
@@ -214,6 +215,23 @@ func ParseOfftime(offtime int64) (hour, minite, second int) {
 	return
 }
 
+// ParseScheduleParamToCron ...
+func ParseScheduleParamToCron(param *common_model.ScheduleParam) string {
```
reasonerjt: How about move this func to `models.ScheduleParam` and call it `CronString()`?

src/core/api/models/admin_job.go
```diff
@@ -139,3 +142,42 @@ func (ar *AdminJobReq) CronString() string {
...
+	if cronStr == "" {
+		return ScheduleParam{}, nil
+	}
+
+	convertedSchedule := ScheduleParam{}
+	convertedSchedule.Type = "custom"
+
+	if strings.Contains(cronStr, "parameter") {
+		scheduleModel := &common_models.ScanAllPolicy{}
+		if err := json.Unmarshal([]byte(cronStr), &scheduleModel); err != nil {
```
reasonerjt: it should not be `&scheduleModel` as it's already a pointer?

src/core/api/models/admin_job.go
```diff
@@ -139,3 +142,42 @@ func (ar *AdminJobReq) CronString() string {
...
+			return ScheduleParam{}, err
+		}
+		h, m, s := common_utils.ParseOfftime(int64(scheduleModel.Parm["daily_time"].(float64)))
+		cron := fmt.Sprintf("%d %d %d * * *", s, m, h)
+		convertedSchedule.Cron = cron
+		return convertedSchedule, nil
+	} else if strings.Contains(cronStr, "offtime") {
+		scheduleModel := &common_models.ScheduleParam{}
+		if err := json.Unmarshal([]byte(cronStr), &scheduleModel); err != nil {
```
reasonerjt: it should not be &scheduleModel as it's already a pointer?

src/core/api/models/admin_job.go
```diff
@@ -139,3 +142,42 @@ func (ar *AdminJobReq) CronString() string {
 	}
 	return string(str)
 }
+
+// ConvertSchedule converts different kinds of cron string into one standard for UI to show.
+// in the latest design, it uses {"type":"Daily","cron":"0 0 0 * * *"} as the cron item.
+// As for supporting migration from older version, it needs to convert {"parameter":{"daily_time":0},"type":"daily"}
+// and {"type":"Daily","weekday":0,"offtime":57600} into one standard.
+func ConvertSchedule(cronStr string) (converted ScheduleParam, err error) {
```
reasonerjt: No need to return the names return value here.

PR #7635, opened 2019-05-05
Return 400 status code for validation failed

### LINKED ISSUE #6880: Validate failed for Post/Patch API should return 400 status code not 404
When the body request for Post/Patch API is invalide, API should return 400 status code not 404.

eg. for `/api/labels` POST request to create a label for project, when the project which id is  project_id not exist, API  should return 400 bot 404.

heww:
Return 400 status code for validation failed which fix #6880 and this match the responses of `/api/labels` in swagger.yaml 

### DIFF
--- a/docs/swagger.yaml
+++ b/docs/swagger.yaml
@@ -515,8 +515,6 @@ paths:
           description: User need to log in first.
         '403':
           description: User in session does not have permission to the project.
-        '404':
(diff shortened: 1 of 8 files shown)


ywk253100:
@heww Why we should return 400 rather than 404? Is there any standard for this? If not, I don't think we need to do this as some other APIs(project member, replication policy, ...) return 404

PR #7639, opened 2019-05-05
Fix chart storage keyfile issue in gcs

ninjadq:
Add volumn binding on docker-compose.yml

### DIFF
--- a/make/photon/prepare/templates/docker_compose/docker-compose.yml.jinja
+++ b/make/photon/prepare/templates/docker_compose/docker-compose.yml.jinja
@@ -378,6 +378,9 @@ services:
     volumes:
       - {{data_volume}}/chart_storage:/chart_storage:z
       - ./common/config/chartserver:/etc/chartserver:z
+{% if gcs_keyfile %}
+      - {{gcs_keyfile}}:/etc/registry/gcs.key
+{% endif %}
 {%if registry_custom_ca_bundle_path %}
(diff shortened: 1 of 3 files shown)


make/photon/prepare/templates/docker_compose/docker-compose.yml.jinja
```diff
@@ -378,6 +378,9 @@ services:
     volumes:
       - {{data_volume}}/chart_storage:/chart_storage:z
       - ./common/config/chartserver:/etc/chartserver:z
+{% if gcs_keyfile %}
```
stonezdj: Please do not take gcs_keyfile as a condition, this file is generated with empty content in harbor tile env, use storage_service is gcp

reasonerjt: This attribute is not reliable:
https://github.com/goharbor/harbor/blob/master/make/photon/prepare/utils/docker_compose.py#L37

You should make sure this attribute is set only when the storage is `"gcs"`

make/photon/prepare/utils/chart.py
```diff
@@ -12,13 +12,12 @@
 def prepare_chartmuseum(config_dict):
 
     core_secret = config_dict['core_secret']
-    registry_custom_ca_bundle_path = config_dict['registry_custom_ca_bundle_path']
```
reasonerjt: How is this change relevant to this PR?  Is it necessary to remove it?

ninjadq: because it is an unused var

PR #7640, opened 2019-05-05
Remove duplicate env items

ninjadq:
some env items are duplicate in both env and config_env file


reasonerjt:
Please remove `config_env` so there is only one `env` file.

Issue #7295, opened 2019-04-04
UAA cert file problem caused by remove uaa setting from harbor.yml 


ninjadq:
In the previous version, if uaa related settings are set. prepare script will copy cert file to proper location.
But now all uaa related settings are user level settings and removed from config file. 
These will influence the tile migration. We need to solve it in harbor 1.8 release

reasonerjt:
@ninjadq this is not configurable via configurations API, we may need to update the docker-compose yaml and mount the cert into `core` container.

reasonerjt:
A new chunk like `UAA -> CA certificate` will be added to harbor.yml

reasonerjt:
Fixed in #7652

PR #7652, opened 2019-05-06
Alow user to set CA cert for UAA in harbor.yml

reasonerjt:


reasonerjt:
Fixes #7295

PR #7657, opened 2019-05-07
add swagger user guide

zhoumeina:
add swagger user guide.


reasonerjt:
It should be part of the user guide?  And we only have to let user know how to navigate to that piece of UI via Harbor's Portal.

reasonerjt:
1.  It's REST API not Swagger API.  So it should reword to "Invoke REST API from Portal" or "Try REST API via API Explorer"
2. The image looks extremely big in user guide markdown file.

zhoumeina:
> 1. It's REST API not Swagger API.  So it should reword to "Invoke REST API from Portal" or "Try REST API via API Explorer"
> 2. The image looks extremely big in user guide markdown file.

done.

zhoumeina:
> It should be part of the user guide? And we only have to let user know how to navigate to that piece of UI via Harbor's Portal.

done.

PR #7660, opened 2019-05-07
Iterate all paginations when listing projects and repositories

ywk253100:
Iterate all paginations when listing projects and repositories

### DIFF
--- a/src/common/http/client.go
+++ b/src/common/http/client.go
@@ -17,9 +17,13 @@ package http
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/url"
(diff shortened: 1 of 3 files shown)


src/common/http/client.go
```diff
@@ -168,3 +172,63 @@ func (c *Client) do(req *http.Request) ([]byte, error) {
...
+		}
+
+		res := reflect.New(elemType)
+		if err = json.Unmarshal(data, res.Interface()); err != nil {
+			return err
+		}
+		resources = reflect.AppendSlice(resources, reflect.Indirect(res))
+
+		endpoint = ""
+		link := resp.Header.Get("Link")
```
steven-zou: Suggest to use a better approach to handle the next/previous paginations.

PR #7697, opened 2019-05-07
Fix the cron tooltip in replication is not fully displayed in Firefox.

pureshine:
Using the table tag directly as the content of the tooltip, compatibility issues will appear in the Firefox browser, resulting in incomplete tooltip display.

image

solution:
Add a layer of div outside as tooltip content

image


src/portal/lib/src/create-edit-rule/create-edit-rule.component.html
```diff
@@ -147,59 +147,61 @@ <h3 class="modal-title">{{headerTitle | translate}}</h3>
...
-                            </tr>
-                            <tr>
-                                <td>{{'REPLICATION.DAY_WEEK' | translate}}</td>
-                                <td>{{'REPLICATION.YES' | translate}}</td>
-                                <td>0-6 or SUN-SAT</td>
-                                <td>* / , - ?</td>
-                            </tr>
-                        </tbody>
-                    </table>
+                    <div class="tooltip-content table-box">
```
jwangyangls: the table  content ，why not pull it out and put it in a component?

Issue #7054, opened 2019-03-01
Update migrator and docs to reflect new migration process


reasonerjt:
In 1.8.0 we will only support upgrade path 
1.6.x -> 1.8 and 1.7.x -> 1.8

So some change will be made:
1.  Remove the useless mariadb modules from the migrator image
2.  Simplify the migration guide to make it only cover the supported upgrade/migration path.

PR #7703, opened 2019-05-07
Update migrator to support only from >=v1.6.0

reasonerjt:
This commit removes code to support upgrade from <v1.6.0
It also removes packages for supporting mysql/mariadb from Dockerfile

It does not handle the optimization in the script such as run.sh after the update.


reasonerjt:
Partially fixes #7054

PR #7717, opened 2019-05-08
Redirect regular user to OIDC login page

reasonerjt:
When the auth mode is OIDC, when a user login via Harbor's login form.
If the user does not exist or the user is onboarded via OIDC, he will be
redirected to the OIDC login page.

### DIFF
--- a/src/core/controllers/base.go
+++ b/src/core/controllers/base.go
@@ -16,12 +16,15 @@ package controllers
 
 import (
 	"bytes"
+	"context"
+	"github.com/goharbor/harbor/src/core/filter"
 	"html/template"
 	"net"
 	"net/http"
 	"os"
 	"regexp"
 	"strconv"
+	"strings"
 
 	"github.com/astaxie/beego"
 	"github.com/beego/i18n"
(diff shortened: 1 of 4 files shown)


src/core/controllers/base.go
```diff
@@ -54,10 +57,43 @@ type messageDetail struct {
...
 
+func redirectForOIDC(ctx context.Context, username string) bool {
+	am, _ := ctx.Value(filter.AuthModeKey).(string)
+	if am != common.OIDCAuth {
+		return false
+	}
+	u, err := dao.GetUser(models.User{Username: username})
+	if err != nil {
+		log.Warningf("Failed to get user by name: %s, error: %v", username, err)
+	}
```
ywk253100: return ?

src/core/controllers/base.go
```diff
@@ -54,10 +57,43 @@ type messageDetail struct {
...
+		return true
+	}
+	return false
+}
+
 // Login handles login request from UI.
 func (cc *CommonController) Login() {
 	principal := cc.GetString("principal")
 	password := cc.GetString("password")
+	if redirectForOIDC(cc.Ctx.Request.Context(), principal) {
```
wy65701436: am, _ := ctx.Value(filter.AuthModeKey).(string)
if am != common.OIDCAuth {
	if redirectForOIDC(cc.Ctx.Request.Context(), principal) {
               ........
        }
}

src/core/controllers/base.go
wy65701436: +1

src/core/controllers/base.go
reasonerjt: That was checked in `redirectForOIDC `

src/core/controllers/base.go
reasonerjt: It's not necessary because if u is nil it returns `true` so when error happens the redirection will happen.

PR #7719, opened 2019-05-08
Fix chart relative url issue

ninjadq:
Support absolute path in chart museum


make/harbor.yml
```diff
@@ -61,6 +61,9 @@ jobservice:
   # Maximum number of job workers in job service  
   max_job_workers: 10
 
+chart:
```
steven-zou: add comments

PR #7720, opened 2019-05-08
Update migration_guide.md

rajboruah:


rajboruah:
Need to revert a change.

PR #7760, opened 2019-05-10
Fix public url shoud not display port is it's default value

ninjadq:
if https port is 443 or http port is 80, then only showing url


tansuomm:
i hava a problem, when setting port 5000, it's hub.domain.com:5000/project/images:tag when pulling; Because setting  from hub.domain.com to (ip:5000) proxy,How just only display hub.domain.com/project/images:tab in harbor ui.

PR #7045, opened 2019-02-28
Fix hostname command not found

John-Lin:
Fix issue #7044 

screen shot 2019-02-28 at 8 55 21 pm

Issue #7044, opened 2019-02-28
goharbor/harbor-db container hostname Command not found


John-Lin:
**Expected behavior and actual behavior:**

Expected behavior:
hostname command is able to use that docker-healthcheck.sh can work.

Actual behavior:
hostname command not found caused k8s healthcheck failed.

![hostname-command-not-found](https://user-images.githubusercontent.com/4214069/53567387-0cf69600-3b9a-11e9-8d0a-d42bb189c2ed.png)

**Steps to reproduce the problem:**
Please provide the steps to reproduce this problem.
![screen shot 2019-02-28 at 8 42 08 pm](https://user-images.githubusercontent.com/4214069/53567455-44654280-3b9a-11e9-9d51-8b6be3a39d6d.png)

**Versions:**
Please specify the versions of following systems.
- harbor version: 1.7
![screen shot 2019-02-28 at 8 50 10 pm](https://user-images.githubusercontent.com/4214069/53567523-6bbc0f80-3b9a-11e9-8176-9640f9bd30a2.png)

reasonerjt:
@ywk253100 could you help take a look at the PR?

hasonhai:
> @ywk253100 could you help take a look at the PR?
> And I'm curious why this is not discovered by our nightly test for helm chart?

I think because the pod reported unhealthy only. It didn't fail so the nightly test cannot recognize.

ywk253100:
This should be fixed by install the `hostname` when building the image.

This script can work currently because the error is ignored https://github.com/goharbor/harbor/blob/master/make/photon/db/docker-healthcheck.sh#L4

John-Lin:
There is no hostname command package since toybox has been removed.

https://github.com/goharbor/harbor/blob/f79895b6d8c07041758fe7b1d2411caf56af12e4/make/photon/db/Dockerfile#L17

and toybox hostname doesn't have `-i` command option

![screen shot 2019-03-01 at 11 36 15 am](https://user-images.githubusercontent.com/4214069/53614831-3e14ac00-3c16-11e9-8de3-04cb44de34d2.png)

ywk253100:
The unhealthy status isn't caused by the `hostname: command not found`.

If you check the code here: 
https://github.com/goharbor/harbor/blob/master/make/photon/db/docker-healthcheck.sh#L4, you can find the variable `host` will be set as `127.0.0.1` when the command not found, and the script will continue rather than exit.

The Kubernetes checks the status of pod by the exit code of the script.

PR #7798, opened 2019-05-14
fix hostname not fould

runzexia:
https://github.com/goharbor/harbor/pull/7045
It should make sense to merge the fix to 1.7 before the 1.8 release is released.


ywk253100:
@runzexia If you got any health check error, it shouldn't related with the fix, I have explained the reason here https://github.com/goharbor/harbor/issues/7044#issuecomment-468540103. So you need to check other reasons that caused the health error.

PR #7855, opened 2019-05-22
small typo correction

ckd:


wy65701436:
thanks @ckd , can you make DCO pass?

steven-zou:
@ckd 

Would you please append DCO to the commit for passing checking?

Issue #7871, opened 2019-05-22
JobService keeps restarting when Redis served by HAProxy


cd1989:
If you are reporting a problem, please make sure the following information are provided:

**Expected behavior and actual behavior:**

`jobservice` keeps restarting every 1min around with following error log:

```
May 21 17:33:29 172.20.0.1 jobservice[841]: 2019-05-21T09:33:29Z [INFO] OP commands sweeper is started
May 21 17:33:29 172.20.0.1 jobservice[841]: 2019-05-21T09:33:29Z [INFO] Redis job stats manager is started
May 21 17:33:29 172.20.0.1 jobservice[841]: 2019-05-21T09:33:29Z [INFO] Message server is started
May 21 17:33:29 172.20.0.1 jobservice[841]: 2019-05-21T09:33:29Z [INFO] Load 0 periodic job policies
May 21 17:33:29 172.20.0.1 jobservice[841]: 2019-05-21T09:33:29Z [INFO] Periodic enqueuer is started
May 21 17:33:29 172.20.0.1 jobservice[841]: 2019-05-21T09:33:29Z [INFO] Redis scheduler is started
May 21 17:33:29 172.20.0.1 jobservice[841]: 2019-05-21T09:33:29Z [INFO] Subscribe redis channel {harbor_job_service_namespace}:period:policies:notifications
May 21 17:33:29 172.20.0.1 jobservice[841]: 2019-05-21T09:33:29Z [ERROR] [enqueuer.go:81]: periodic_enqueuer.loop.enqueue:key {harbor_job_service_namespace}:period:lock is already set with value a0325172fd59d4de186e3ec3
May 21 17:33:29 172.20.0.1 jobservice[841]: 2019-05-21T09:33:29Z [INFO] Redis worker pool is started
May 21 17:33:59 172.20.0.1 jobservice[841]: 2019-05-21T09:33:59Z [INFO] Message server is stopped
May 21 17:33:59 172.20.0.1 jobservice[841]: 2019-05-21T09:33:59Z [ERROR] [redis_pool.go:160]: Message server exits with error: error occurred when receiving from pub/sub channel of message server: read tcp 172.20.0.8:37576->192.168.21.250:6379: use of closed network connection
May 21 17:34:06 172.20.0.1 jobservice[841]: 2019-05-21T09:34:06Z [INFO] Restart message server (1 times)
[...]

PR #7872, opened 2019-05-22
Config idle timeout for redis pool to avoid jobservice restarting

cd1989:

When Redis is served under HAProxy, job service keeps restarting without configure idle timeout for redis pool

Fix: https://github.com/goharbor/harbor/issues/7871

### DIFF
--- a/make/photon/prepare/templates/jobservice/config.yml.jinja
+++ b/make/photon/prepare/templates/jobservice/config.yml.jinja
@@ -20,6 +20,7 @@ worker_pool:
     #redis://[arbitrary_username:password@]ipaddress:port/database_index
     redis_url: {{redis_url}}
     namespace: "harbor_job_service_namespace"
+    idle_timeout_second: 3600
 #Loggers for the running job
 job_loggers:
   - name: "STD_OUTPUT" # logger backend name, only support "FILE" and "STD_OUTPUT"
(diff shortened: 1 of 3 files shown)


cd1989:
@kofj I think the issue already describes the problem.

mmpei:
lgtm

cd1989:
@wy65701436 @reasonerjt   I have rebased it, please check it again.

make/photon/prepare/templates/jobservice/config.yml.jinja
```diff
@@ -20,6 +20,7 @@ worker_pool:
     #redis://[arbitrary_username:password@]ipaddress:port/database_index
     redis_url: {{redis_url}}
     namespace: "harbor_job_service_namespace"
+    idle_timeout_second: 3600
```
steven-zou: I suggest this property can be hard coded and we don't need to expose it in the config YAML file

cd1989: What's the considerations ? I don't think there will be a fixed value for it, it's related to the proxy timeout settings. Won't it be more flexible to give users a chance to configure it ?

reasonerjt: I'm ok as long as it's not configurable in `harbor.yml`

KarstenSiemer: I don't understand as to why someone would want to hardcode this?
Default is 30s in most haproxy charts. So why hardcode 3600?
It would make a ton of sense to give user the chance to change this

PR #7915, opened 2019-05-28
aws driver for replication

bitsf:
Change-Id: I8792ffce2eaa5975359bb6159a1ba7b85926a925


bitsf:
use https://github.com/goharbor/harbor/pull/8016 to do merge first

src/replication/adapter/awsecr/adapter.go
```diff
@@ -0,0 +1,165 @@
...
+			model.TriggerTypeManual,
+			model.TriggerTypeScheduled,
+		},
+	}, nil
+}
+
+// HealthCheck checks health status of a registry
+func (d adapter) HealthCheck() (model.HealthStatus, error) {
+	if d.registry.Credential == nil ||
+		(len(d.registry.Credential.AccessKey) == 0 && len(d.registry.Credential.AccessSecret) == 0) {
```
steven-zou: I don't think one of them (key/secret) can be empty. Both should be existing for authing.

src/replication/adapter/awsecr/adapter.go
```diff
@@ -0,0 +1,165 @@
...
+		return model.Unhealthy, nil
+	}
+	return model.Healthy, nil
+}
+
+// PrepareForPush nothing need to do.
+func (d adapter) PrepareForPush(resources []*model.Resource) error {
+	for _, resource := range resources {
+		if resource == nil {
+			return errors.New("the resource cannot be null")
```
steven-zou: Please change `null` to `nil` in the error messages.

src/replication/adapter/awsecr/auth.go
```diff
@@ -0,0 +1,146 @@
...
+	return nil
+}
+
+func (a *awsAuthCredential) getAuthorization() (string, string, string, error) {
+	log.Infof("Aws Ecr getAuthorization %s", a.accessKey)
+	cred := credentials.NewStaticCredentials(
+		a.accessKey,
+		a.accessSecret,
+		"")
+	sess := session.Must(session.NewSession(&aws.Config{
```
steven-zou: Seems the `must` will panic the process, right?  The error can be wrapped and returned.

src/replication/adapter/awsecr/image_registry.go
```diff
@@ -0,0 +1,113 @@
...
+
+import (
+	adp "github.com/goharbor/harbor/src/replication/adapter"
+	"github.com/goharbor/harbor/src/replication/model"
+	"github.com/goharbor/harbor/src/replication/util"
+)
+
+var _ adp.ImageRegistry = adapter{}
+
+func (n adapter) FetchImages(filters []*model.Filter) ([]*model.Resource, error) {
```
steven-zou: Why the short name for `adapter` is `n`? Based on the best practice of go, it should be `a`.

src/replication/adapter/awsecr/adapter.go
bitsf: > I don't think one of them (key/secret) can be empty. Both should be existing for authing.

Yes, you are right, I just copy code from src/replication/adapter/image_registry.go.
These should not be empty

steven-zou:
`-0.6%` down too much, you need to add more test cases @bitsf

src/replication/adapter/awsecr/adapter.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+type adapter struct {
+	*adp.DefaultImageRegistry
+	registry *model.Registry
+}
+
+var _ adp.Adapter = adapter{}
+
+func (adapter) Info() (info *model.RegistryInfo, err error) {
```
heww: Please use pointer receiver and keep the same with other adapter implements.

src/replication/adapter/awsecr/adapter.go
```diff
@@ -0,0 +1,166 @@
...
+		},
+		SupportedTriggers: []model.TriggerType{
+			model.TriggerTypeManual,
+			model.TriggerTypeScheduled,
+		},
+	}, nil
+}
+
+// HealthCheck checks health status of a registry
+func (a adapter) HealthCheck() (model.HealthStatus, error) {
```
heww: Please use pointer receiver and keep the same with other adapter implements.

src/replication/adapter/awsecr/adapter.go
bitsf: ok
actually 'src/replication/adapter/native/adapter.go' also have the same problem.

bitsf:
> `-0.6%` down too much, you need to add more test cases @bitsf

fixed

JordanSussman:
Any chance this functionality will eventually support utilizing [IAM Roles](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2.html) instead of requiring the administrator to pass static credentials via access/secret keys?

bitsf:
@JordanSussman I haven't test this method yet, I think if it can do something just like the access/secret, and can be static to save in DB, and no expired timeout, it can be supported.

bitsf:
@JordanSussman And it would be much more helpful if you can give more sample on how do you think this works. Such as what the UI behavior like, what the authorize process like.

JordanSussman:
@bitsf - I'd expect the auth method to work exactly like how the docker registry handles the s3 storage driver. Note from the [registry docs](https://docs.docker.com/registry/storage-drivers/s3/):

> You can provide empty strings for your access and secret keys to run the driver on an ec2 instance and handles authentication with the instance’s credentials. If you use [IAM roles](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html), omit these keys to fetch temporary credentials from IAM.

Some code examples from the registry:

- [Reading parameters](https://github.com/docker/distribution/blob/0c394fdd8477bff93f347d44308509f41f9516d5/registry/storage/driver/s3-aws/s3.go#L172-L182)
- [Specify credentials](https://github.com/docker/distribution/blob/0c394fdd8477bff93f347d44308509f41f9516d5/registry/storage/driver/s3-aws/s3.go#L412-L423)
- [Utilize credentials](https://github.com/docker/distribution/blob/0c394fdd8477bff93f347d44308509f41f9516d5/registry/storage/driver/s3-aws/s3.go#L430)

So, if the user configures ECR replication without access/secret keys it should fall back to utilize IAM role roles setup on the instance. You do **not** need to provide the name/id of the IAM role since the EC2 instance is able to determine that itself. 

I wouldn't change anything in the UI besides maybe adding a note that the replication will attempt to utilize IAM roles if no access/secret keys are defined.

bitsf:
Do you mean if I'm install a harbor instance in an AWS VM, then I will not use any AK to call ECR registry API or native docker API ?

PR #7934, opened 2019-05-30
Fix save button always disabled in safari

### LINKED ISSUE #7829: need 2 endpoints to create replication in 1.8.0-rc1 using Safari, Edge or Internet Explorer
I expect the Save button to be enabled once all the required fields are filled in on the Mew Replication Rule Dialog, but it is not until I have explicitly selected a destination registry, and I can only make that selection if I have more than one available/

Fresh install of 1.8.0-rc1 on Ubuntu 18.04
Access the UI via Safari 12.1
Create a single endpoint
Try to create a replication rule - I have found no way to enable the save button on the New Replication Rule dialog
Note that the Destination registry field is pre-populated with the only endpoint.

Now create a second endpoint
Try again - this time the save button can be enabled by providing a Name and selecting the 2nd endpoint from the drop-down at "Destination registry".
It remains enabled after re-selecting the first entry on that list.

Please provide the steps to reproduce this problem.

**Versions:**
- harbor version: 1.8.0-rc1
- docker engine version: 18.09.6
- docker-compose version: 1.24.

**Additional context:**
[harbor.config.tar.gz](https://github.com/goharbor/harbor/files/3187355/harbor.config.tar.gz)
[harbor.logs.tar.gz](https://github.com/goharbor/harbor/files/3187356/harbor.logs.tar.gz)

jwangyangls:
add the default value to fix dest_registry cannot select  in safari;  fix #7829

### DIFF
--- a/src/portal/lib/src/create-edit-rule/create-edit-rule.component.ts
+++ b/src/portal/lib/src/create-edit-rule/create-edit-rule.component.ts
@@ -95,6 +95,7 @@ export class CreateEditRuleComponent implements OnInit, OnDestroy {
       });
       this.supportedTriggers = adapter.supported_triggers;
       this.ruleForm.get("trigger").get("type").setValue(this.supportedTriggers[0]);
+      this.clearSelectDefaultValueInSafari('dest_registry');
     }, (error: any) => {
       this.inlineAlert.showInlineError(error);
     });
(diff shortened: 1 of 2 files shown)


src/portal/lib/src/create-edit-rule/create-edit-rule.component.ts
```diff
@@ -142,6 +142,12 @@ export class CreateEditRuleComponent implements OnInit, OnDestroy {
   }
 
   pullModeChange(): void {
+
+    // add the default value to fix src_registry cannot select when this.sourceList.length==1 in safari;
+    if (this.sourceList.length) {
```
ninjadq: this condition is not enough to make sure sourceList length == 1

jwangyangls: I have change the comments

PR #7946, opened 2019-05-31
gcr driver for replication

bitsf:
Change-Id: I5a6626950d3878bfa9726b332e68bee59159269f


bitsf:
Because GCR is using project_id as part of namespace, currently we have to extremely specify it as the prefix of target namespace.

src/replication/adapter/googlegcr/adapter.go
```diff
@@ -0,0 +1,105 @@
...
+			model.TriggerTypeManual,
+			model.TriggerTypeScheduled,
+		},
+	}, nil
+}
+
+// HealthCheck checks health status of a registry
+func (d adapter) HealthCheck() (model.HealthStatus, error) {
+	var err error
+	if d.registry.Credential == nil ||
```
ywk253100: Is the credential necessary to create a GCR registry?

bitsf: Yes, no matter push-based or pull-based, both need credential.
And I also can't find an extra API that can list repos, except using registry native api /v2/_catalog, which can not search by special namespace

bitsf:
ui enhancement PR https://github.com/goharbor/harbor/pull/8028

src/common/utils/registry/registry.go
```diff
@@ -130,9 +128,18 @@ func (r *Registry) Catalog() ([]string, error) {
 	return repos, nil
 }
 
-// Ping ...
+// Ping checks by Head method
 func (r *Registry) Ping() error {
-	req, err := http.NewRequest(http.MethodHead, buildPingURL(r.Endpoint.String()), nil)
+	return r.ping(http.MethodHead)
```
steven-zou: Duplicated changes? ecr adapter also has the same change.

src/replication/adapter/googlegcr/adapter.go
```diff
@@ -0,0 +1,105 @@
...
+		(len(d.registry.Credential.AccessKey) == 0 && len(d.registry.Credential.AccessSecret) == 0) {
```
steven-zou:
```
len(d.registry.Credential.AccessKey) == 0 && len(d.registry.Credential.AccessSecret) == 0
```
=>
```
len(d.registry.Credential.AccessKey) == 0 || len(d.registry.Credential.AccessSecret) == 0
```
?

src/common/utils/registry/registry.go
bitsf: Yes, because this is two seperate PR, so I copy this code, in case they can't be merged together

steven-zou:
`-0.3%` down a little much more. @bitsf

PR #7951, opened 2019-06-02
Line 143: Syntax error. Delete letter 'i'

dmnt3d:


reasonerjt:
@dmnt3d thanks for the pr however this script is to be removed from `master` branch

PR #7952, opened 2019-06-02
Fix some bugs in the Log page

qiqizjl:
1: Fix When the log is not found, the prompt is English.
image

2:Fix When clicking the Refresh Log button is invalid. Set search box to null invalid bug

Signed-off-by: 耐小心 <qiqizjl@qq.com>


src/portal/lib/src/log/recent-log.component.ts
```diff
@@ -67,9 +67,9 @@ export class RecentLogComponent implements OnInit {
     }
 
     public doFilter(terms: string): void {
-        if (!terms) {
-            return;
-        }
+        // if (!terms) {
```
jwangyangls: it has conflict because of I has modify it  in branch master ,please update code and commit .

jwangyangls: image

jwangyangls: Bona fide reminder: Your code should use the same commit (just like the pull request you are submitting now)

src/portal/lib/src/log/recent-log.component.html
```diff
@@ -40,4 +40,4 @@
             </clr-dg-footer>
         </clr-datagrid>
     </div>
-</div>
```
zhoumeina: please don't modify this line.

jwangyangls:
I have resolved the conflict  in log.html and en.json,and approve it .

PR #8003, opened 2019-06-10
Bump up to go 1.12.5 enable go.mod

reasonerjt:
This commit bumps up the version of Go to compile the code to v1.12.5,
and shifts to go.mod for managing depedency.
Some code from "harbor/tests" to "harbor/src/testing" to avoid depedency
loop of modules.

Note that in short term we will still vendor the dependency.


reasonerjt:
The warnings from Codacy can be ignored as they are all introduced by moving the api test lib into `src` directory.

PR #8028, opened 2019-06-12
ui change: gcr/ecr driver for replication

bitsf:
ui enhance for gcr replication,
depend on https://github.com/goharbor/harbor/pull/7946

### DIFF
--- a/src/portal/lib/src/create-edit-endpoint/create-edit-endpoint.component.html
+++ b/src/portal/lib/src/create-edit-endpoint/create-edit-endpoint.component.html
@@ -44,11 +44,15 @@
         <div class="form-group">
           <label for="destination_url" class="col-md-4 form-group-label-override required">{{ 'DESTINATION.URL' |
             translate }}</label>
-          <label class="col-md-8" for="destination_url" aria-haspopup="true" role="tooltip" [class.invalid]="targetEndpoint.errors && (targetEndpoint.dirty || targetEndpoint.touched)"
-            [class.valid]="targetEndpoint.valid" class="tooltip tooltip-validation tooltip-sm tooltip-bottom-left">
-            <input type="text" id="destination_url" [disabled]="testOngoing || urlDisabled" [readonly]="!editable" [(ngModel)]="target.url"
-              size="25" name="endpointUrl" #targetEndpoint="ngModel" required placeholder="http(s)://192.168.1.1">
-            <span class="tooltip-content" *ngIf="targetEndpoint.errors && targetEndpoint.errors.required && (targetEndpoint.dirty || targetEndpoint.touched)">
+          <label class="col-md-8" for="destination_url" aria-haspopup="true" role="tooltip" [class.invalid]="targetEndpoint?.errors && (targetEndpoint?.dirty || targetEndpoint?.touched)"
+                 [class.valid]="targetEndpoint?.valid" class="tooltip tooltip-validation tooltip-sm tooltip-bottom-left">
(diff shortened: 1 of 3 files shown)


src/portal/lib/src/create-edit-endpoint/create-edit-endpoint.component.html
```diff
@@ -44,11 +44,36 @@ <h3 class="modal-title">{{modalTitle}}</h3>
...
+            <input *ngIf="target.type !=='google-gcr' && target.type !=='aws-ecr'" type="text" id="destination_url" [disabled]="testOngoing || urlDisabled" [readonly]="!editable" [(ngModel)]="target.url"
+                   size="25" name="endpointUrl" #targetEndpoint="ngModel" required placeholder="http(s)://192.168.1.1">
+            <select *ngIf="target.type =='google-gcr'" [(ngModel)]="target.url" name="endpointUrl" #targetEndpoint="ngModel">
+              <option value="https://gcr.io">https://gcr.io</option>
+              <option value="https://us.gcr.io">https://us.gcr.io</option>
+              <option value="https://eu.gcr.io">https://eu.gcr.io</option>
+              <option value="https://asia.gcr.io">https://asia.gcr.io</option>
+            </select>
+            <select *ngIf="target.type =='aws-ecr'" [(ngModel)]="target.url" name="endpointUrl" #targetEndpoint="ngModel">
+              <option value="https://api.ecr.ap-northeast-1.amazonaws.com">ap-northeast-1</option>
```
jwangyangls: Can you make sure that the data in these options is not returned from the back end but from the front end page?

src/portal/lib/src/create-edit-endpoint/create-edit-endpoint.component.ts
```diff
@@ -219,6 +220,11 @@ export class CreateEditEndpointComponent
       this.urlDisabled = false;
       this.targetForm.controls.endpointUrl.setValue("");
     }
+    if (selectValue === 'google-gcr') {
```
jwangyangls: The above condition should be the relationship of else if, not nothing!

src/portal/lib/src/create-edit-endpoint/create-edit-endpoint.component.html
jwangyangls: if  the data in these options is not returned from the front end page,then it should also be in the TS file to do some data display in the front end, rather than writing directly on the page.

src/portal/lib/src/create-edit-endpoint/create-edit-endpoint.component.ts
```diff
@@ -219,6 +220,11 @@ export class CreateEditEndpointComponent
...
+      this.targetForm.controls.access_key.setValue("_json_key");
+    }else{
```
jwangyangls: it display formatting problems in travels CI

src/portal/lib/src/create-edit-endpoint/create-edit-endpoint.component.ts
bitsf: no need

src/portal/lib/src/create-edit-endpoint/create-edit-endpoint.component.html
bitsf: ok

src/portal/lib/src/create-edit-endpoint/create-edit-endpoint.component.ts
bitsf: fixed

src/portal/lib/src/create-edit-endpoint/create-edit-endpoint.component.ts
bitsf: fixed

PR #8032, opened 2019-06-12
Support label filter for replication

ywk253100:
Only the system level labels are supported to be used as the replication policy filter


mumrau:
I wonder if this could be merged into a `1.8.2` release as it doesn't look backwards-incompatible?
I might be completely wrong though.

bitsf:
@reasonerjt will it be into 1.8.2

PR #8086, opened 2019-06-20
Add backup and restore script and document

stonezdj:

### DIFF
--- a/docs/backup_restore_guide.md
+++ b/docs/backup_restore_guide.md
@@ -1,0 +1,69 @@
+# Harbor Backup and Restore Guide
+
+When the container registry storage is local file system, the straightforward way to backup a standalone Harbor data is to backup the /data directory and recover it to /data directory.
+For some special cases, if you want more efficient way to backup the Harbor data, you can refer the guide following. 
+
+## Backup Harbor data
+
+1. Login to the Harbor host, find a disk which has enough space, create a backup directory in it, for exmaple, /backup
+2. Download script from https://github.com/goharbor/harbor/blob/backup_restore/tools/harbor-backup.sh and copy it to /backup
+3. Shutdown the Harbor instance  
+```
+docker-compose down -v
+```
+4. Backup data
+```
+cd /backup
+ls
+harbor-backup.sh
+```
+
+Stop all running containers. Backup all data
+
+```
+./harbor-backup.sh 
+```
+Or only backup database data when Harbor storage is NFS, GCP or S3
+```
+./harbor-backup.sh --dbonly
+```
+4. After the backup complete, there is a harbor.tgz file in /backup, it is the backup data, copy it to the backup storage.
+```
+ls /backup
+harbor-backup.sh    harbor.tgz
+```
+5. Start Harbor 
+```
+docker-compose up -d
+```
+## Restore Harbor data
+Install Harbor with the same version and same install options. Login to the Harbor host.
+Copy the backup data file harbor.tgz to the directory /restore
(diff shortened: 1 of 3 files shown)


reasonerjt:
Can we first clarify under what circumstances will these back/restore scripts be used?  
If this should be the single source of truth regarding back/restore, we should also modify the upgrade/migration guide (https://github.com/goharbor/harbor/blob/master/docs/migration_guide.md) to revise the process and make sure there's only one way to backup/restore Harbor.

docs/backup_restore_guide.md
```diff
@@ -0,0 +1,69 @@
+# Harbor Backup and Restore Guide
+
+When the container registry storage is local file system, the straightforward way to backup a standalone Harbor data is to backup the /data directory and recover it to /data directory.
```
michmike: what happens with in-flight data in the /data directory? is this a safe way to backup where no corruption can occur?

docs/backup_restore_guide.md
```diff
@@ -0,0 +1,69 @@
...
+For some special cases, if you want more efficient way to backup the Harbor data, you can refer the guide following. 
+
```
michmike: rephrase "you can refer to the following guide"

docs/backup_restore_guide.md
```diff
@@ -0,0 +1,69 @@
...
+## Backup Harbor data
+
+1. Login to the Harbor host, find a disk which has enough space, create a backup directory in it, for exmaple, /backup
```
michmike: the word example is mispelled

docs/backup_restore_guide.md
```diff
@@ -0,0 +1,69 @@
...
+docker-compose down -v
+```
+4. Backup data
+```
+cd /backup
+ls
+harbor-backup.sh
+```
+
+Stop all running containers. Backup all data
```
michmike: the command harbor-backup.sh is listed twice. before this step and here. do they need to run it twice?

is it possible to take a good backup without downtime of the Harbor services?

docs/backup_restore_guide.md
```diff
@@ -0,0 +1,69 @@
...
+harbor-backup.sh    harbor.tgz
+```
+5. Start Harbor 
+```
+docker-compose up -d
+```
+## Restore Harbor data
+Install Harbor with the same version and same install options. Login to the Harbor host.
+Copy the backup data file harbor.tgz to the directory /restore
+Download script from https://github.com/goharbor/harbor/blob/backup_restore/tools/harbor-restore.sh and copy it to /restore
```
michmike: we should ship the backup and restore commands with harbor so they are versioned for the exact release of Harbor they are backing up and restoring

docs/backup_restore_guide.md
```diff
@@ -0,0 +1,69 @@
...
+```
+2. Restore data
+```
+cd /restore
+ls
+harbor-restore.sh   harbor.tgz
+```
+Stop all running containers. Restore all data
+```
+./harbor-restore.sh
```
michmike: same comment as above. do they need to run harbor-restore.sh twice?

docs/backup_restore_guide.md
wToad: it is not. to avoid data corruption it is necessary to ensure that there is no activity on the platform - remember that Harbor uses PostgreSQL. to have a useful backup of Harbor, at least we would need to use `pg_dump` or similar tool for the database.

reasonerjt:
I'm closing this PR as we've discussed and we need a better tool for this purpose, the script in this PR is quite limited.  
Currently user for OSS Harbor just need to back up the "data volume"

PR #8123, opened 2019-06-21
Enable push dev images on release branch

wy65701436:

### DIFF
--- a/tests/integration.sh
+++ b/tests/integration.sh
@@ -123,7 +123,6 @@ function publishImage {
     # for release-*, will use 'release-*-dev' as the tag name, like release-v1.8.0-dev
     if [[ $DRONE_BRANCH == "master" ]]; then
       image_tag=dev
-      docker images | sed -n "s|\(goharbor/[-._a-z0-9]*\)\s*\(.*$Harbor_Assets_Version\).*|docker tag \1:\2 \1:$image_tag;docker push \1:$image_tag|p" | bash
     fi
     if [[ $DRONE_BRANCH == "release-"* ]]; then
       image_tag=$Harbor_Assets_Version-dev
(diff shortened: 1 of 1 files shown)


tests/integration.sh
```diff
@@ -119,10 +119,19 @@ function package_offline_installer {
 function publishImage {
     echo "Publishing images to Docker Hub..."
     echo "The images on the host:"
+    # for master, will use 'dev' as the tag name
+    # for release-*, will use 'release-*-dev' as the tag name, like release-v1.8.0-dev
+    if [[ $DRONE_BRANCH == "master" ]]; then
```
ywk253100: We should call the docker login first before pushing images

tests/integration.sh
```diff
@@ -119,10 +119,19 @@ function package_offline_installer {
...
+    fi
+    if [[ $DRONE_BRANCH == "release-"* ]]; then
+      image_tag=$Harbor_Assets_Version-dev
+    fi
+    # rename the images with tag "dev" and push to Docker Hub
     docker images
     docker login -u $DOCKER_HUB_USERNAME -p $DOCKER_HUB_PASSWORD
-    # rename the images with tag "dev" and push to Docker Hub
-    docker images | sed -n "s|\(goharbor/[-._a-z0-9]*\)\s*\(.*$Harbor_Assets_Version\).*|docker tag \1:\2 \1:dev;docker push \1:dev|p" | bash
+    docker images | sed -n "s|\(goharbor/[-._a-z0-9]*\)\s*\(.*$Harbor_Assets_Version\).*|docker tag \1:\2 \1:$image_tag;docker push \1:$image_tag|p" | bash
```
ywk253100: How about set different value to `image_tag ` according to the branch name and keep only one `docker images | sed -n ...`?

tests/integration.sh
wy65701436: do you mean line 133?

tests/integration.sh
wy65701436: yes, it is. The image_tag will be set to branch name plus '-dev', like v1.8.0-dev. 

I do keep only one "docker images | sed -n ..." now, not clearly understand your comments.

PR #8133, opened 2019-06-25
Add dao for quota

heww:
DAO for quota

### DIFF
--- a/make/migrations/postgresql/0010_1.9.0_schema.up.sql
+++ b/make/migrations/postgresql/0010_1.9.0_schema.up.sql
@@ -8,3 +8,25 @@ CREATE TABLE cve_whitelist (
     items text NOT NULL,
     UNIQUE (project_id)
 );
+
+/* add quota table */
+CREATE TABLE quota (
+ id SERIAL PRIMARY KEY NOT NULL,
+ reference VARCHAR(255) NOT NULL,
+ reference_id VARCHAR(255) NOT NULL,
+ hard JSONB NOT NULL,
+ creation_time timestamp default CURRENT_TIMESTAMP,
+ update_time timestamp default CURRENT_TIMESTAMP,
+ UNIQUE(reference, reference_id)
+);
+
+/* add quota usage table */
+CREATE TABLE quota_usage (
+ id SERIAL PRIMARY KEY NOT NULL,
+ reference VARCHAR(255) NOT NULL,
+ reference_id VARCHAR(255) NOT NULL,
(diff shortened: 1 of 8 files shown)


src/common/models/quota.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	return hard, nil
+}
+
+// MustHard must returns quota hard
+func (q *Quota) MustHard() QuotaHard {
+	hard, err := q.GetHard()
+	if err != nil {
+		panic("hard invalid")
```
ywk253100: Where panic here?

heww: This method assert that `Hard` value in `Quota` instance is a valid `QuotaHard`.

ywk253100: If the `Harbor` value is invalid, the whole process will exit, is it the expected behavior?

make/migrations/postgresql/0010_1.9.0_schema.up.sql
```diff
@@ -0,0 +1,21 @@
+/* add quota table */
+CREATE TABLE quota (
+ id SERIAL PRIMARY KEY NOT NULL,
+ reference VARCHAR(255),
```
wy65701436: not null?

make/migrations/postgresql/0010_1.9.0_schema.up.sql
```diff
@@ -0,0 +1,21 @@
...
+ hard JSONB NOT NULL,
+ creation_time timestamp default CURRENT_TIMESTAMP,
+ update_time timestamp default CURRENT_TIMESTAMP,
+ UNIQUE(reference, reference_id)
+);
+
+/* add quota usage table */
+CREATE TABLE quota_usage (
+ id SERIAL PRIMARY KEY NOT NULL,
+ reference VARCHAR(255),
```
wy65701436: not null?

src/common/models/quota.go
heww: `MustHard` and `MustUsed` methods are removed, provide `equalHard` and `equalUsed` in suite instead.

make/migrations/postgresql/0010_1.9.0_schema.up.sql
heww: Fixed

make/migrations/postgresql/0010_1.9.0_schema.up.sql
heww: Fixed

Issue #8085, opened 2019-06-20
Unable to set external_db_configs['harbor']['host']


lmdkfs:
when upgrade harbor to 1.8.1 form 1.7.0,   i use external db .But get the following errors when to execute prepare.
```
Traceback (most recent call last):
  File "main.py", line 62, in <module>
    main()
  File "/usr/lib/python3.6/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3.6/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3.6/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "main.py", line 31, in main
    config_dict = parse_yaml_config(conf)
  File "/usr/src/app/utils/configs.py", line 199, in parse_yaml_config
    config_dict['harbor_db_host'] = external_db_configs['harbor']['host']
KeyError: 'harbor'
```

lmdkfs:
how to resolve。

ninjadq:
are you update your harbor.yml file with your external db information?

lmdkfs:
It's been upgraded . but generate harbor.yml  has problem。 Configuration is as follows
`external_database:
  host: 10.10.104.41
  port: 5432
  db_name: registry
  username: xx
  password: xx
  ssl_mode: disable`
I fixed the configuration file manually， it resolved。
`
external_database:
  harbor:
    host: 10.10.104.41
    port: 5432
    db_name: registry
    username: xx
    password: xx
    ssl_mode: disable

  clair:
    host: 10.10.104.41
    port: 5432
    db_name: clair
    username: xx
    password: xx
    ssl_mode: disable

  notary_signer:
    host: 10.10.104.41
    port: 5432
    db_name: notary
    username: xx
    password: xx
    ssl_mode: disable
  notary_server:
    host: 10.10.104.41
    port: 5432
    db_name: notary
    username: xx
    password: xx
    ssl_mode: disable
`
An error occurred while using goharbor/prepare:v1.8.1 to upgrade the configuration file.
Hopefully it can be fixed.

PR #8139, opened 2019-06-25
Fix migration script to support exeternal db

ninjadq:
migrated config file will contains seperated db configuration


ninjadq:
fix  #8085

Issue #7700, opened 2019-05-07
helm-chart label filter shouldn't inside the table


ninjadq:
![image](https://user-images.githubusercontent.com/1069785/57289828-b4d69580-70ee-11e9-9d79-7967f6995d80.png)

PR #8165, opened 2019-06-28
fix  filter ui bug : half filter is covered when there is one or two records in a datagrid

AllForNothing:
refer to #7700

### DIFF
--- a/src/portal/src/styles.css
+++ b/src/portal/src/styles.css
@@ -87,3 +87,7 @@ body {
 .color-red {
     color: red;
 }
+
+.datagrid-table,.datagrid-header{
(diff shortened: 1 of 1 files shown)


src/portal/src/styles.css
```diff
@@ -86,4 +86,8 @@ body {
 
 .color-red {
     color: red;
+}
+
+.datagrid-table{
+    min-height: 172px;
```
AllForNothing: this code set min-height to every datagrid .this will fix the filter ui bug.

AllForNothing:
This is a framework level bug of clarity v1 .

Issue #8157, opened 2019-06-27
Should not need to clear tables in TestMain


reasonerjt:
In the `common/dao` package it calls migrator to update the schema, this should be idempotent,
so we should not need:
https://github.com/goharbor/harbor/blob/master/src/replication/dao/dao_test.go#L31

However I encountered failure in travis before adding a new element to the array, there should be somewhere the UT is manually creating tables.
Need to check.

PR #8168, opened 2019-06-28
Clean up the dao test for replication

ywk253100:
Clean up the dao test for replication


ywk253100:
Fixes #8157

Issue #7978, opened 2019-06-05
Support replication between Harbor and ACR(Azure Container Registry)


ywk253100:

s-spindler:
I would very much like to see this. Is any work being done on this already?

mumrau:
I am wondering whether ACR is that different from a "normal" v2 docker registry?
I tried the feature and got fairly disappointed in the end, the log finishing on an error code...which is... 200.

cd1989:
Auth of ACR is different from normal v2 docker registry. By normal registry, /v2 requests always got 401 without token, but with ACR, if basic auth provided, /v2 requests will work fine.

Replication of Harbor relies on 401 status code to get token.

ywk253100:
Done

PR #8172, opened 2019-06-29
Implement azure acr adapter

cd1989:

Implement Azure ACR adapter to replicate images between Harbor and Azure ACR

Issue: https://github.com/goharbor/harbor/issues/7978

### DIFF
--- a/src/replication/adapter/azurecr/adapter.go
+++ b/src/replication/adapter/azurecr/adapter.go
@@ -1,0 +1,118 @@
+package azurecr
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/goharbor/harbor/src/common/http/modifier"
+	common_http_auth "github.com/goharbor/harbor/src/common/http/modifier/auth"
+	"github.com/goharbor/harbor/src/common/utils/log"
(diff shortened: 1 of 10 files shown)


src/replication/adapter/azurecr/adapter.go
```diff
@@ -0,0 +1,118 @@
...
+	}, nil
+}
+
+// PrepareForPush no preparation needed for Azure container registry
+func (a *adapter) PrepareForPush(resources []*model.Resource) error {
+	return nil
+}
+
+// HealthCheck checks health status of a registry
+func (a adapter) HealthCheck() (model.HealthStatus, error) {
```
heww: Please use pointer receiver for `HealthCheck` to keep the same with other adapters

src/replication/adapter/azurecr/adapter.go
```diff
@@ -0,0 +1,118 @@
+package azurecr
```
ywk253100: I think the package name should be `azureacr`?

cd1989: It's `azurecr` in Azure officially.

PR #8174, opened 2019-07-01
refact replication info to support special endpoints and credential

bitsf:
Refact replication info to support special endpoints and credential.

Just backend model for review first

### DIFF
--- a/src/core/router.go
+++ b/src/core/router.go
@@ -102,6 +102,7 @@ func initRouters() {
 	beego.Router("/api/logs", &api.LogAPI{})
 
 	beego.Router("/api/replication/adapters", &api.ReplicationAdapterAPI{}, "get:List")
+	beego.Router("/api/replication/adapterinfos", &api.ReplicationAdapterAPI{}, "get:ListAdapterInfos")
 	beego.Router("/api/replication/executions", &api.ReplicationOperationAPI{}, "get:ListExecutions;post:CreateExecution")
 	beego.Router("/api/replication/executions/:id([0-9]+)", &api.ReplicationOperationAPI{}, "get:GetExecution;put:StopExecution")
 	beego.Router("/api/replication/executions/:id([0-9]+)/tasks", &api.ReplicationOperationAPI{}, "get:ListTasks")
--- a/src/replication/adapter/adapter.go
+++ b/src/replication/adapter/adapter.go
@@ -31,10 +31,13 @@ const (
 )
 
 var registry = map[model.RegistryType]Factory{}
-var adapterInfoMap = map[model.RegistryType]*model.AdapterInfo{}
+var adapterInfoMap = map[model.RegistryType]*model.AdapterPattern{}
 
 // Factory creates a specific Adapter according to the params
-type Factory func(*model.Registry) (Adapter, error)
+type Factory interface {
+	Create(*model.Registry) (Adapter, error)
+	AdapterPattern() *model.AdapterPattern
+}
 
 // Adapter interface defines the capabilities of registry
 type Adapter interface {
@@ -123,7 +126,7 @@ func (v *VTag) GetLabels() []string {
 }
 
 // RegisterFactory registers one adapter factory to the registry
-func RegisterFactory(t model.RegistryType, factory Factory, adapterInfo *model.AdapterInfo) error {
+func RegisterFactory(t model.RegistryType, factory Factory) error {
 	if len(t) == 0 {
 		return errors.New("invalid registry type")
 	}
@@ -135,6 +138,7 @@ func RegisterFactory(t model.RegistryType, factory Factory, adapterInfo *model.A
 		return fmt.Errorf("adapter factory for %s already exists", t)
 	}
 	registry[t] = factory
+	adapterInfo := factory.AdapterPattern()
 	if adapterInfo != nil {
 		adapterInfoMap[t] = adapterInfo
 	}
@@ -166,6 +170,6 @@ func ListRegisteredAdapterTypes() []model.RegistryType {
 }
 
 // ListAdapterInfos list the adapter infos
-func ListAdapterInfos() map[model.RegistryType]*model.AdapterInfo {
+func ListAdapterInfos() map[model.RegistryType]*model.AdapterPattern {
 	return adapterInfoMap
 }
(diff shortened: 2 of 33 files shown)


bitsf:
UI change https://github.com/goharbor/harbor/pull/8610

src/replication/adapter/adapter.go
```diff
@@ -121,7 +122,7 @@ func (v *VTag) GetLabels() []string {
 }
 
 // RegisterFactory registers one adapter factory to the registry
-func RegisterFactory(t model.RegistryType, factory Factory) error {
+func RegisterFactory(t model.RegistryType, factory Factory, registryInfo *model.AdapterInfo) error {
```
steven-zou: Maybe we can use one unified model to contain both Factory and registry info.

```go
package adapter

type Metadata struct {
   Name string
   Kind string // registration/self-contain
   Factory Factory
}
```

src/core/router.go
```diff
@@ -107,6 +107,7 @@ func initRouters() {
 	beego.Router("/api/logs", &api.LogAPI{})
 
 	beego.Router("/api/replication/adapters", &api.ReplicationAdapterAPI{}, "get:List")
+	beego.Router("/api/replication/adapterinfos", &api.ReplicationAdapterAPI{}, "get:ListAdapterInfos")
```
steven-zou: What's the difference between these two endpoints.

```go
beego.Router("/api/replication/adapters", &api.ReplicationAdapterAPI{}, "get:List")
	beego.Router("/api/replication/adapterinfos", &api.ReplicationAdapterAPI{}, "get:ListAdapterInfos")
```

bitsf: adapterinfos API can be used to replace the old adapters API, but because the new API return a map while old return a list, if directly remove old one, it may break user's script

src/replication/model/registry.go
```diff
@@ -116,11 +116,53 @@ type FilterStyle struct {
...
-	SupportedResourceTypes   []ResourceType `json:"-"`
-	SupportedResourceFilters []*FilterStyle `json:"supported_resource_filters"`
-	SupportedTriggers        []TriggerType  `json:"supported_triggers"`
+	Type                     RegistryType    `json:"type"`
+	Description              string          `json:"description"`
+	SupportedResourceTypes   []ResourceType  `json:"-"`
+	SupportedResourceFilters []*FilterStyle  `json:"supported_resource_filters"`
+	SupportedTriggers        []TriggerType   `json:"supported_triggers"`
+	SpecialEndpoints         []*Endpoint     `json:"special_endpoints"`
+	SpecialCredential        *CredentialInfo `json:"special_credential"`
```
ywk253100: Why we need the `SpecialEndpoints` and `SpecialCredential` for registry? They aren't used only to render the UI?

src/replication/adapter/adapter.go
```diff
@@ -122,7 +123,7 @@ func (v *VTag) GetLabels() []string {
 }
 
 // RegisterFactory registers one adapter factory to the registry
-func RegisterFactory(t model.RegistryType, factory Factory) error {
+func RegisterFactory(t model.RegistryType, factory Factory, adapterInfo *model.AdapterInfo) error {
```
ywk253100: How about changing the `Factory` to an interface and adding a method `AdapterInfo()`(or we can call it Style/Pattern) to it?
```
type Factory interface{
Create(*model.Registry)(Adapter, error)
AdapterPattern() *Pattern
}
```

src/replication/model/registry.go
```diff
@@ -116,11 +116,53 @@ type FilterStyle struct {
...
+}
+
+// AdapterInfo provides base info and capability declarations of the registry
+type AdapterInfo struct {
```
ywk253100: How about rename it to `AdapterPattern` or `AdapterStyle` as we already define a method called `Info` in interface `Adapter`?

src/replication/model/registry.go
```diff
@@ -116,11 +116,53 @@ type FilterStyle struct {
...
+	SpecialEndpoints  []*Endpoint     `json:"special_endpoints"`
+	SpecialCredential *CredentialInfo `json:"special_credential"`
```
ywk253100: My personal opinion:
```golang
type AdapterPattern struct{
  Endpoints []*EndpointPattern
  Credential *CredentialPattern
}

type EndpointPattern struct{
  Type string
  Name string
  Value string
}

type CredentialPattern struct{
  Type string
  
}

const(
  EndpointPatternTypeStandard // for most of the adapters
  EndpointPatternTypeFixed // for docker hub, helm hub, etc

  CredentialPatternTypeStandard // for most of the adapters
  CredentialPatternTypeAccessSecretOnly // cover the GCR adapter, the fixed access key should be covered in the implementation of adapter
)

```

src/replication/adapter/adapter.go
```diff
@@ -31,9 +31,13 @@ const (
 )
 
 var registry = map[model.RegistryType]Factory{}
+var adapterInfoMap = map[model.RegistryType]*model.AdapterPattern{}
```
ywk253100: Do we need the map? How about to call the AdapterPattern method of registered factory to get the pattern when needed?

src/replication/model/registry.go
```diff
@@ -124,3 +176,32 @@ type RegistryInfo struct {
 	SupportedResourceFilters []*FilterStyle `json:"supported_resource_filters"`
 	SupportedTriggers        []TriggerType  `json:"supported_triggers"`
 }
+
+// AdapterPattern provides base info and capability declarations of the registry
+type AdapterPattern struct {
+	EndpointPattern   *EndpointPattern   `json:"endpoint_pattern"`
```
ywk253100: I think we can use a list here for endpoints, for docker hub it contains only one item, and for ali it contains multiple items. It's easy for frontend as they use the same logic to render the UI

src/replication/adapter/adapter.go
```diff
@@ -31,9 +31,13 @@ const (
...
 
 // Factory creates a specific Adapter according to the params
-type Factory func(*model.Registry) (Adapter, error)
+type Factory interface {
```
steven-zou: I still has some confusion about this change here. Let's have more discussions offline. @bitsf @ywk253100

PR #8175, opened 2019-07-01
Add manager for quota

heww:
Add manager for quota

### DIFF
--- a/src/common/dao/base.go
+++ b/src/common/dao/base.go
@@ -187,3 +187,29 @@ func Escape(str string) string {
 	str = strings.Replace(str, `_`, `\_`, -1)
 	return str
 }
+
+// WithTransaction helper for transaction
+func WithTransaction(handler func(o orm.Ormer) error) error {
+	o := orm.NewOrm()
+
+	if err := o.Begin(); err != nil {
+		log.Errorf("begin transaction failed: %v", err)
+		return err
+	}
+
+	if err := handler(o); err != nil {
+		if e := o.Rollback(); e != nil {
+			log.Errorf("rollback transaction failed: %v", e)
+			return e
+		}
+
+		return err
+	}
+
+	if err := o.Commit(); err != nil {
+		log.Errorf("commit transaction failed: %v", err)
+		return err
+	}
+
+	return nil
+}
--- a/src/common/models/quota_usage.go
+++ b/src/common/models/quota_usage.go
@@ -40,7 +40,7 @@ func (u QuotaUsed) Copy() QuotaUsed {
 // QuotaUsage model for quota usage
 type QuotaUsage struct {
 	ID           int64     `orm:"pk;auto;column(id)" json:"id"`
-	Reference    string    `orm:"column(reference)" json:"reference"`
+	Reference    string    `orm:"column(reference)" json:"reference"` // The reference type for quota usage, eg: project, user
 	ReferenceID  string    `orm:"column(reference_id)" json:"reference_id"`
 	Used         string    `orm:"column(used);type(jsonb)" json:"-"`
 	CreationTime time.Time `orm:"column(creation_time);auto_now_add" json:"creation_time"`
--- a/src/common/quota/util.go
+++ b/src/common/quota/util.go
@@ -1,0 +1,65 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package quota
+
+import (
+	"fmt"
+)
+
+const (
+	// UNLIMITED unlimited quota value
+	UNLIMITED = int64(-1)
+)
+
+type unsafe struct {
+	message string
+}
+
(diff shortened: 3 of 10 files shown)


src/common/quota/util.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	if used > hard {
+		return errors.New("over the quota")
+	}
+
+	return nil
+}
+
+func withTransaction(fn func(o orm.Ormer) error) error {
```
wy65701436: move it to base of DAO, it could be reused by other API.

src/common/quota/quota.go
```diff
@@ -0,0 +1,96 @@
...
+
+// Reserve reserve resource
+func (m *Manager) Reserve(resource Resource) error {
+	return withTransaction(func(o orm.Ormer) error {
+		return m.updateUsage(o, resource, reserve)
+	})
+}
+
+// Unreserve unreserve resource
+func (m *Manager) Unreserve(resource Resource) error {
```
wy65701436: do you mean ‘release’ or 'free'?

src/common/quota/util.go
```diff
@@ -0,0 +1,89 @@
...
+	UNLIMITED = int64(-1)
+)
+
+type operate func(current int64, value int64) int64
+
+func reserve(current int64, value int64) int64 {
+	return current + value
+}
+
+func unreserve(current int64, value int64) int64 {
```
wy65701436: release or free?

src/common/quota/util.go
```diff
@@ -0,0 +1,89 @@
...
+
+		ret[key] = current
+	}
+
+	return ret, nil
+}
+
+func isSafe(hard int64, used int64) error {
+	if used < 0 {
+		return errors.New("bad used value")
```
wy65701436: print used?

src/common/quota/util.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	if hard == UNLIMITED {
+		return nil
+	}
+
+	if used > hard {
```
wy65701436: ditto

src/common/models/quota_usage.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	return used, nil
+}
+
+// MustUsed must returns quota used
+func (qu *QuotaUsage) MustUsed() QuotaUsed {
+	used, err := qu.GetUsed()
+	if err != nil {
+		panic("used invalid")
```
wy65701436: panic here breaks the main thread?

src/common/quota/util.go
heww: moved

src/common/models/quota_usage.go
heww: panic removed

src/common/dao/base.go
```diff
@@ -187,3 +187,27 @@ func Escape(str string) string {
...
+func WithTransaction(fn func(o orm.Ormer) error) error {
+	o := orm.NewOrm()
+
+	if err := o.Begin(); err != nil {
+		log.Errorf("begin transaction failed: %v", err)
+		return err
+	}
+
+	if err := fn(o); err != nil {
+		if err := o.Rollback(); err != nil {
```
ywk253100: We should declare a new variable for the error returned by `Rollback` or the `err` returned by `fn` will be overwritten by `Rollback`. The error of `fn` should be returned.

src/common/quota/quota.go
```diff
@@ -0,0 +1,112 @@
...
+	usage.SetUsed(used)
+
+	usage.UpdateTime = time.Now()
+	_, err = o.Update(usage)
+	return err
+}
+
+// UpdateQuota update the hard of quota
+func (m *Manager) UpdateQuota(hard models.QuotaHard) error {
+	return dao.WithTransaction(func(o orm.Ormer) error {
```
ywk253100: Why needs two steps to update the quota? It can be done by one step and without transaction

src/common/quota/quota.go
```diff
@@ -0,0 +1,112 @@
...
+	usage := models.QuotaUsage{Reference: m.reference, ReferenceID: m.referenceID}
+	err := o.ReadForUpdate(&usage, "reference", "reference_id")
+	if err != nil {
+		return nil, err
+	}
+
+	return &usage, nil
+}
+
+func (m *Manager) updateUsage(o orm.Ormer, resource Resource, op operate) error {
```
ywk253100: Why we don't use the optimistic concurrent control to avoid the performance issue?

src/common/quota/quota.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	usage, err := m.getUsage(o)
+	if err != nil {
+		return err
+	}
+
+	used, err := calcUsed(quota, usage, resource, op)
+	if err != nil {
+		return err
```
ywk253100: Should return a specific error for upper layers to determine it's internal server error or just because the request doesn't meet the quota requirement?

src/common/quota/quota.go
heww: `UPDATE` can not access row concurrently with `SELECT FOR UPDATE` so i will delete this method.

src/common/quota/quota.go
heww: unsafe error return when request resources failed

src/common/dao/base.go
heww: The err returns by `Rollback` will not overwritten the err returns by `fn`, see more https://goplay.space/#RuSSkIvv26J

But I think that when `Rollback` failed, we should return the error of `Rollback` which is an internal error. Otherwise, return the error of `fn`.

src/common/quota/quota.go
heww: `Reserve` renamed to `AddResources`, and `Unreserve` renamed to `SubtractResources`

PR #8176, opened 2019-07-01
Refactor LDAP usergroup

stonezdj:

### DIFF
--- a/src/common/dao/project.go
+++ b/src/common/dao/project.go
@@ -156,18 +156,19 @@ func GetProjects(query *models.ProjectQueryParam) ([]*models.Project, error) {
 
 // GetGroupProjects - Get user's all projects, including user is the user member of this project
 // and the user is in the group which is a group member of this project.
-func GetGroupProjects(groupDNCondition string, query *models.ProjectQueryParam) ([]*models.Project, error) {
+func GetGroupProjects(groupIDs []int, query *models.ProjectQueryParam) ([]*models.Project, error) {
 	sql, params := projectQueryConditions(query)
 	sql = `select distinct p.project_id, p.name, p.owner_id, 
 				p.creation_time, p.update_time ` + sql
-	if len(groupDNCondition) > 0 {
+	groupIDCondition := JoinNumberConditions(groupIDs)
+	if len(groupIDs) > 0 {
 		sql = fmt.Sprintf(
 			`%s union select distinct p.project_id, p.name, p.owner_id, p.creation_time, p.update_time  
 		     from project p 
 		     left join project_member pm on p.project_id = pm.project_id
-		     left join user_group ug on ug.id = pm.entity_id and pm.entity_type = 'g' and ug.group_type = 1
-			 where ug.ldap_group_dn in ( %s ) order by name`,
-			sql, groupDNCondition)
+		     left join user_group ug on ug.id = pm.entity_id and pm.entity_type = 'g' 
+			 where ug.id in ( %s ) order by name`,
+			sql, groupIDCondition)
 	}
 	sqlStr, queryParams := CreatePagination(query, sql, params)
 	log.Debugf("query sql:%v", sql)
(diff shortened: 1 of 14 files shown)


src/common/models/user.go
```diff
@@ -21,6 +21,12 @@ import (
 // UserTable is the name of table in DB that holds the user object
 const UserTable = "harbor_user"
 
+// GroupContext ...
+type GroupContext struct {
+	UserGroup []*UserGroup
```
reasonerjt: This should be plural or call it a list?

src/core/auth/ldap/ldap.go
```diff
@@ -112,8 +112,14 @@ func (l *Auth) Authenticate(m models.AuthModel) (*models.User, error) {
 		}
 		userGroups = append(userGroups, userGroupList[0])
 	}
-	u.GroupList = userGroups
-
+	if len(userGroups) == 0 {
+		u.GroupContext = models.GroupContext{}
```
reasonerjt: You should still set the groupType correctly

src/core/promgr/pmsdriver/local/local.go
```diff
@@ -132,23 +132,16 @@ func (d *driver) Update(projectIDOrName interface{},
 func (d *driver) List(query *models.ProjectQueryParam) (*models.ProjectQueryResult, error) {
 	var total int64
 	var projects []*models.Project
-	var groupDNCondition string
+	var gpi group.PrivilegeInterface
+	gpi = &group.DefaultGroupPrivilege{}
```
reasonerjt: Considering `CreatePrivilegeInterface` is already handling the fallback case, we don't need the defaultvalue here. so that the `DefaultGroupPrivilege` will not need to be exposed from the package?

src/common/dao/group/ldap.go
```diff
@@ -156,3 +156,28 @@ func GetGroupDNQueryCondition(userGroupList []*models.UserGroup) string {
 	}
 	return strings.Join(result, ",")
 }
+
+// LDAPGroupContext ...
+type LDAPGroupContext []*models.UserGroup
+
+// GetRoleInProject ...
+func (c *LDAPGroupContext) GetRoleInProject(projectID int64) ([]int, error) {
+	queryCondition := GetGroupDNQueryCondition([]*models.UserGroup(*c))
```
reasonerjt: In such case will we still need to export  `GetGroupDNQueryCondition` ?

src/common/dao/project.go
```diff
@@ -156,18 +156,18 @@ func GetProjects(query *models.ProjectQueryParam) ([]*models.Project, error) {
 
 // GetGroupProjects - Get user's all projects, including user is the user member of this project
 // and the user is in the group which is a group member of this project.
-func GetGroupProjects(groupDNCondition string, query *models.ProjectQueryParam) ([]*models.Project, error) {
+func GetGroupProjects(groupIDCondition string, query *models.ProjectQueryParam) ([]*models.Project, error) {
```
reasonerjt: If it only supports would it be better to input a slice of groupID and transform it into `ID Contidioin` in the func in `dao` package?

src/common/utils/utils.go
```diff
@@ -255,3 +255,8 @@ func IsContainIllegalChar(s string, illegalChar []string) bool {
 	}
 	return false
 }
+
+// JoinNumberConditions - To join number condition into string,used in sql query
```
reasonerjt: if it's just for sql query would it be more appropriate to move it to `dao` pkg?

reasonerjt:
comments added, please also handle the failure in travis-ci

src/common/utils/utils.go
stonezdj: Updated

src/common/dao/project.go
stonezdj: Updated

PR #8190, opened 2019-07-01
Merge Default ImageRegistry into the native adapter to reduce the duplicate code

ywk253100:
Merge Default ImageRegistry into the native adapter to reduce the duplicate code


cd1989:
LGTM

PR #8206, opened 2019-07-03
Refactor interceptors code with chain

wy65701436:
1, add a blob inteceptors for quota usage
2, add a manifest inteceptors for quota usage


src/core/controllers/controllers_test.go
```diff
@@ -32,7 +32,7 @@ import (
 	"github.com/goharbor/harbor/src/common/models"
 	utilstest "github.com/goharbor/harbor/src/common/utils/test"
 	"github.com/goharbor/harbor/src/core/config"
-	"github.com/goharbor/harbor/src/core/proxy"
+	"github.com/goharbor/review/harbor/src/core/proxy"
```
heww: This import path may not exist. Should it be `github.com/goharbor/src/core/proxy` ?

src/core/middlewares/chain.go
```diff
@@ -0,0 +1,126 @@
...
+				return nil
+			}
+			return constructor(next)
+		})
+	}
+	return &chain
+}
+
+func (b *DefaultCreator) getMiddleware(mName string) alice.Constructor {
+	var middleware alice.Constructor
```
heww: This `getMiddleware` method can be implemented using table-driven methods.

```go
middlewares := map[string]alice.Constructor {
  READONLY: func(next http.Handler) http.Handler { return readonly.New(next) },
  URL: func(next http.Handler) http.Handler { return url.New(next) },
  // ...
}

return middlewares[mName]
```

`middlewares` can be defined and initialized outside this method for performance.

src/core/controllers/controllers_test.go
wy65701436: ok thanks.

src/core/middlewares/regquota/handler.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// ServeHTTP PATCH manifest ...
+func (rqh regQuotaHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	match, _, _ := util.MatchManifestURL(req)
+	if match {
+		var mfSize int64
+		var mfDigest string
+		mediaType := req.Header.Get("Content-Type")
+		if req.Method == http.MethodPut && mediaType == "application/vnd.docker.distribution.manifest.v2+json" {
```
ywk253100: There are some other media types that should be supported, refer to  https://github.com/goharbor/harbor/blob/master/src/replication/transfer/image/transfer.go#L263

src/core/middlewares/util/util.go
```diff
@@ -0,0 +1,168 @@
...
+	if err != nil {
+		log.Debugf("failed to marshal json error, %v", err)
+		return msg
+	}
+	return string(str)
+}
+
+// MatchManifestURL ...
+func MatchManifestURL(req *http.Request) (bool, string, string) {
+	re := regexp.MustCompile(manifestURLPattern)
```
ywk253100: Should use `Compile` instead, the `MustCompile` panics if the pattern cannot be parsed.

PR #8210, opened 2019-07-03
Add HTTP group support

stonezdj:


src/common/dao/group/usergroup.go
```diff
@@ -84,6 +86,27 @@ func GetUserGroup(id int) (*models.UserGroup, error) {
 	return nil, nil
 }
 
+// GetHTTPGroupIDByGroupName ...
+func GetHTTPGroupIDByGroupName(groupName []string) ([]int, error) {
```
reasonerjt: 1.  This is not restricted to HTTP group
2.  Is it possible we reuse `QueryUserGroup` ?  For example you can add a list of names to the input query parm.

I hope there's only one "ListGroup" func or there's only one fundamental "listGroup" func that is reused by different funcs.

src/core/api/email_test.go
```diff
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Avoid to run this test in local env (Mac)
```
reasonerjt: You can add parm to `go test` when you run the test cases locally?

src/common/dao/group/usergroup.go
```diff
@@ -137,3 +160,15 @@ func OnBoardUserGroup(g *models.UserGroup, keyAttribute string, combinedKeyAttri
 
 	return nil
 }
+
+// GetOrCreateHTTPUserGroup - If there is a group with same name, return the group id, or create a new one and return ID
+func GetOrCreateHTTPUserGroup(groupName string) (int, error) {
```
reasonerjt: 1. This does not have to be restricted to HTTP group?
2. Is it safe in concurrent situation?  I think it would be better you call `Add` first and if there's a conflict error you get the ID.

src/core/api/usergroup.go
```diff
@@ -103,41 +114,58 @@ func (uga *UserGroupAPI) Post() {
...
-	}
-	if err == ldap.ErrDNSyntax {
-		uga.SendBadRequestError(fmt.Errorf("invalid DN syntax. DN: %v", userGroup.LdapGroupDN))
-		return
-	}
-	if err != nil {
-		uga.SendInternalServerError(fmt.Errorf("Error occurred in search user group. error: %v", err))
-		return
+
+	if userGroup.GroupType == 1 {
```
reasonerjt: use the const?

src/core/api/usergroup.go
```diff
@@ -150,13 +178,17 @@ func (uga *UserGroupAPI) Put() {
...
 	ID := uga.id
 	userGroup.GroupName = strings.TrimSpace(userGroup.GroupName)
 	if len(userGroup.GroupName) == 0 {
 		uga.SendBadRequestError(errors.New(userNameEmptyMsg))
 		return
 	}
-	userGroup.GroupType = common.LdapGroupType
+	userGroup.GroupType = common.LDAPGroupType
 	log.Debugf("Updated user group %v", userGroup)
 	err := group.UpdateUserGroupName(ID, userGroup.GroupName)
```
reasonerjt: Can we consider disable updating group name by default?  I don't think we should do that to LDAP group either, but we need to avoid breaking user's scenario.

src/core/api/usergroup.go
```diff
@@ -103,41 +114,58 @@ func (uga *UserGroupAPI) Post() {
...
-		uga.SendBadRequestError(fmt.Errorf("invalid DN syntax. DN: %v", userGroup.LdapGroupDN))
-		return
-	}
-	if err != nil {
-		uga.SendInternalServerError(fmt.Errorf("Error occurred in search user group. error: %v", err))
-		return
+
+	if userGroup.GroupType == 1 {
+		query := models.UserGroup{GroupType: userGroup.GroupType, LdapGroupDN: userGroup.LdapGroupDN}
+		result, err := group.QueryUserGroup(query)
```
reasonerjt: Querying before creating is not safe in concurrent situation?

src/core/auth/authproxy/auth.go
```diff
@@ -72,7 +83,39 @@ func (a *Auth) Authenticate(m models.AuthModel) (*models.User, error) {
...
+			return nil, err
+		}
+		if reviewResponse == nil {
+			return nil, auth.ErrAuth{}
+		}
+
+		// Attach user group ID information
+		ugList := reviewResponse.Status.User.Groups
+		log.Debugf("user groups %+v", ugList)
+		if len(ugList) > 0 {
```
reasonerjt: it should be OK to call it even the list is empty?

src/core/auth/authproxy/auth.go
```diff
@@ -72,7 +83,39 @@ func (a *Auth) Authenticate(m models.AuthModel) (*models.User, error) {
...
+		err = json.Unmarshal(data, &s)
+		if err != nil {
+			log.Errorf("failed to read session %v", err)
+		}
+
+		reviewResponse, err := a.tokenReview(s.SessionID)
+		if err != nil {
+			return nil, err
+		}
+		if reviewResponse == nil {
```
reasonerjt: In what case this is `nil` ?

src/core/auth/authproxy/auth.go
```diff
@@ -72,7 +83,39 @@ func (a *Auth) Authenticate(m models.AuthModel) (*models.User, error) {
...
+
+		// Attach user group ID information
+		ugList := reviewResponse.Status.User.Groups
+		log.Debugf("user groups %+v", ugList)
+		if len(ugList) > 0 {
+			groupIDList, err := group.GetHTTPGroupIDByGroupName(ugList)
+			if err != nil {
+				return nil, err
+			}
+			log.Debugf("current user in group list %v", groupIDList)
```
reasonerjt: the message is inaccurate?

src/common/dao/group/usergroup.go
stonezdj: 1. for future, we can make it general and add a groupType param
2. QueryUserGroup match groupName by exact name

src/common/dao/group/usergroup.go
stonezdj: 1. It can be more general.
2.It is not concurrent safe only when the group name is unique.

src/core/api/email_test.go
stonezdj: I run it in GoLand

src/core/api/usergroup.go
stonezdj: OK

src/core/api/usergroup.go
stonezdj: For LDAP group, it is allowed to update groupname.

src/core/auth/authproxy/auth.go
stonezdj: Update

src/core/auth/authproxy/auth.go
stonezdj: Double check for safety.

src/core/auth/authproxy/auth.go
stonezdj: Check for safety, because it is a pointer

make/migrations/postgresql/0010_1.9.0_schema.up.sql
```diff
@@ -7,4 +7,30 @@ CREATE TABLE cve_whitelist (
     expires_at bigint,
     items text NOT NULL,
     UNIQUE (project_id)
-);
\ No newline at end of file
+);
```
reasonerjt: this should be moved to 000x_1.8.2_xxxx.up.sql

src/common/dao/group/usergroup.go
```diff
@@ -84,6 +86,27 @@ func GetUserGroup(id int) (*models.UserGroup, error) {
 	return nil, nil
 }
 
+// GetGroupIDByGroupName ...
```
reasonerjt: This is trivial... but how about GetGroupIDByName?  The extra group is redundant...

reasonerjt: Sorry I forgot to mention this in previous review.
By viewing the code it is possible you input 10 group names but get 8 group IDs, and this may be acceptable for current logic, but we should clearly document it in the comment, thoughts?

stonezdj: Add

src/core/api/projectmember.go
```diff
@@ -243,14 +255,41 @@ func AddProjectMember(projectID int64, request models.MemberReq) (int, error) {
...
-
+		request.MemberGroup.GroupType = common.LDAPGroupType
 		// If groupname provided, use the provided groupname to name this group
 		groupID, err := auth.SearchAndOnBoardGroup(request.MemberGroup.LdapGroupDN, request.MemberGroup.GroupName)
 		if err != nil {
 			return 0, err
 		}
 		member.EntityID = groupID
 		member.EntityType = common.GroupMember
+	} else if len(request.MemberGroup.GroupName) > 0 && request.MemberGroup.GroupType == common.HTTPGroupType {
```
reasonerjt: Move this chunk into the http authenticator's implementation of `SearchAndOnboardGroup` ?

stonezdj: Updated

PR #8230, opened 2019-07-05
add dao of artifact

wy65701436:

### DIFF
--- a/make/migrations/postgresql/0010_1.9.0_schema.up.sql
+++ b/make/migrations/postgresql/0010_1.9.0_schema.up.sql
@@ -8,3 +8,65 @@ CREATE TABLE cve_whitelist (
     items text NOT NULL,
     UNIQUE (project_id)
 );
+
+CREATE TABLE blob (
+ id SERIAL PRIMARY KEY NOT NULL,
+ /*
+    digest of config, layer, manifest
+ */
+ digest varchar(255) NOT NULL,
+ content_type varchar(255) NOT NULL,
+ size int NOT NULL,
+ creation_time timestamp default CURRENT_TIMESTAMP,
+ UNIQUE (digest)
+);
+
+CREATE TABLE artifact (
+ id SERIAL PRIMARY KEY NOT NULL,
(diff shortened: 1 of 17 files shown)


make/migrations/postgresql/0010_1.9.0_schema.up.sql
```diff
@@ -9,6 +9,46 @@ CREATE TABLE cve_whitelist (
     UNIQUE (project_id)
 );
 
+CREATE TABLE Blob (
```
heww: Table name using snake case will be better and keep the same with other table names.

make/migrations/postgresql/0010_1.9.0_schema.up.sql
```diff
@@ -9,6 +9,46 @@ CREATE TABLE cve_whitelist (
...
+    digest of config, layer, manifest
+ */
+ digest varchar(255) NOT NULL,
+ content_type varchar(255) NOT NULL,
+ size int NOT NULL,
+ creation_time timestamp default CURRENT_TIMESTAMP,
+ UNIQUE (digest)
+);
+
+CREATE TABLE Artifact (
```
heww: Table name using snake case will be better and keep the same with other table names.

make/migrations/postgresql/0010_1.9.0_schema.up.sql
```diff
@@ -9,6 +9,46 @@ CREATE TABLE cve_whitelist (
...
+ */
+ kind varchar(255) NOT NULL,
+ creation_time timestamp default CURRENT_TIMESTAMP,
+ pull_time timestamp,
+ push_time timestamp,
+ CONSTRAINT unique_artifact UNIQUE (project_id, repo, tag)
+);
+
+/* add the table for relation of artifact and blob */
+CREATE TABLE ArtifactAndBlob (
```
heww: Table name using snake case will be better and keep the same with other table names.

And suggest changing this name to `artifact_blob`,  no `and` needed between `artifact` and `blob`

PR #8238, opened 2019-07-08
Enable project level CVE whitelist

reasonerjt:
This commit update the project API to support "reuse_sys_cve_whitelist"
setting in project metadata and "cve_whitelist" in project request.
Also modify the interceptor to support project level CVE whitelist if
the reuse flag is false.

### DIFF
--- a/src/core/promgr/promgr.go
+++ b/src/core/promgr/promgr.go
@@ -16,6 +16,7 @@ package promgr
 
 import (
 	"fmt"
+	"github.com/goharbor/harbor/src/pkg/scan/whitelist"
 	"strconv"
 
 	"github.com/goharbor/harbor/src/common/models"
@@ -44,6 +45,7 @@ type defaultProjectManager struct {
 	pmsDriver      pmsdriver.PMSDriver
 	metaMgrEnabled bool // if metaMgrEnabled is enabled, metaMgr will be used to CURD metadata
 	metaMgr        metamgr.ProjectMetadataManager
+	whitelistMgr   whitelist.Manager
 }
 
 // NewDefaultProjectManager returns an instance of defaultProjectManager,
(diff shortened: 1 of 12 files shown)


src/core/promgr/promgr.go
```diff
@@ -110,37 +121,40 @@ func (d *defaultProjectManager) Delete(projectIDOrName interface{}) error {
...
-		if err != nil {
+	pro, err := d.Get(projectIDOrName)
+	if err != nil {
+		return err
+	}
+	if pro == nil {
+		return fmt.Errorf("project %v not found", projectIDOrName)
+	}
+	// TODO transaction?
+	if d.metaMgrEnabled {
```
wy65701436: check with-clair?

src/pkg/scan/whitelist/validator.go
```diff
@@ -0,0 +1,60 @@
...
+	_, ok := err.(*invalidErr)
+	return ok
+}
+
+const cveIDPattern = `^CVE-\d{4}-\d+$`
+
+// Validate help validates the CVE whitelist, to ensure the CVE ID is valid and there's no duplication
+func Validate(wl models.CVEWhitelist) error {
+	m := map[string]struct{}{}
+	re := regexp.MustCompile(cveIDPattern)
```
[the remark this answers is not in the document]
reasonerjt: It should never fail as the `cveIDPattern` is a constant

src/core/promgr/promgr.go
reasonerjt: I don't think we have to do that for CVE whitelist management as it does not depend on Clair.  The API can be called to manage the whitelist without Clair and once Clair is enabled the whitelist will take effect.

For UX consistency, we may only need to hide certain components from UI if Clair is not installed.

PR #8241, opened 2019-07-08
Fix bug when no labels

jwangyangls:


src/portal/lib/src/create-edit-rule/create-edit-rule.component.ts
```diff
@@ -533,7 +532,11 @@ export class CreateEditRuleComponent implements OnInit, OnDestroy {
       }
 
       if (!findTag) {
-        filtersArray.push({ type: this.supportedFilters[i].type, value: "" });
+        if (this.supportedFilters[i].type === 'label') {
```
zhoumeina: please change "label" to use const instead.

PR #8244, opened 2019-07-09
Support helm chart replication from helmhub

mmpei:

it will invoke getChartDetail 3-5 times to make it compatible with the chart interface.
so this will degrade some performance.
@ywk253100 @steven-zou @cd1989 @kofj

### DIFF
--- a/src/replication/adapter/helmhub/adapter.go
+++ b/src/replication/adapter/helmhub/adapter.go
@@ -1,0 +1,76 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package helmhub
+
+import (
+	"errors"
+	"github.com/goharbor/harbor/src/common/utils/log"
+	adp "github.com/goharbor/harbor/src/replication/adapter"
+	"github.com/goharbor/harbor/src/replication/model"
+)
+
+func init() {
+	if err := adp.RegisterFactory(model.RegistryTypeHelmHub, func(registry *model.Registry) (adp.Adapter, error) {
+		return newAdapter(registry)
+	}); err != nil {
+		log.Errorf("failed to register factory for %s: %v", model.RegistryTypeHelmHub, err)
+		return
+	}
+	log.Infof("the factory for adapter %s registered", model.RegistryTypeHelmHub)
+}
+
+type adapter struct {
+	registry *model.Registry
+	client   *Client
+}
+
+func newAdapter(registry *model.Registry) (*adapter, error) {
+	return &adapter{
+		registry: registry,
+		client:   NewClient(registry),
+	}, nil
+}
+
+func (a *adapter) Info() (*model.RegistryInfo, error) {
+	return &model.RegistryInfo{
+		Type: model.RegistryTypeHelmHub,
+		SupportedResourceTypes: []model.ResourceType{
+			model.ResourceTypeChart,
+		},
+		SupportedResourceFilters: []*model.FilterStyle{
+			{
+				Type:  model.FilterTypeName,
+				Style: model.FilterStyleTypeText,
+			},
+			{
+				Type:  model.FilterTypeTag,
+				Style: model.FilterStyleTypeText,
+			},
+		},
+		SupportedTriggers: []model.TriggerType{
+			model.TriggerTypeManual,
+			model.TriggerTypeScheduled,
+		},
+	}, nil
+}
+
+func (a *adapter) PrepareForPush(resources []*model.Resource) error {
+	return errors.New("not supported")
+}
+
+// HealthCheck checks health status of a registry
+func (a *adapter) HealthCheck() (model.HealthStatus, error) {
+	return model.Healthy, nil
+}
--- a/src/replication/adapter/helmhub/adapter.go
+++ b/src/replication/adapter/helmhub/adapter.go
@@ -72,5 +72,9 @@ func (a *adapter) PrepareForPush(resources []*model.Resource) error {
 
 // HealthCheck checks health status of a registry
 func (a *adapter) HealthCheck() (model.HealthStatus, error) {
-	return model.Healthy, nil
+	err := a.client.checkHealthy()
+	if err == nil {
+		return model.Healthy, nil
+	}
+	return model.Unhealthy, err
 }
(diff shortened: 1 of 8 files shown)


kofj:
@mmpei Please fix: https://travis-ci.org/goharbor/harbor/jobs/556101182

src/replication/adapter/helmhub/chart_registry.go
```diff
@@ -0,0 +1,145 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, v := range versionList.Data {
+		if v.Attributes.Version == version {
+			return a.download(v)
+		}
+	}
+	return nil, nil
```
ywk253100: Should return error here to indicates that the chart is not found

src/replication/adapter/helmhub/client.go
```diff
@@ -0,0 +1,97 @@
...
+// Client is a client to interact with HelmHub
+type Client struct {
+	client *http.Client
+}
+
+// NewClient creates a new HelmHub client.
+func NewClient(registry *model.Registry) *Client {
+	return &Client{
+		client: &http.Client{
+			Transport: util.GetHTTPTransport(false),
```
ywk253100: Should read the `insecure` property of `registry` to set the parameter to `GetHTTPTransport`

src/replication/adapter/helmhub/adapter.go
```diff
@@ -0,0 +1,76 @@
...
+	}, nil
+}
+
+func (a *adapter) PrepareForPush(resources []*model.Resource) error {
+	return errors.New("not supported")
+}
+
+// HealthCheck checks health status of a registry
+func (a *adapter) HealthCheck() (model.HealthStatus, error) {
+	return model.Healthy, nil
```
ywk253100: Check the connection

src/replication/adapter/helmhub/chart_registry.go
```diff
@@ -0,0 +1,145 @@
...
+		if err = filter.DoFilter(&repositories); err != nil {
+			return nil, err
+		}
+	}
+
+	for _, repository := range repositories {
+		versionList, err := a.client.fetchChartDetail(repository.Name)
+		if err != nil {
+			log.Errorf("fetch chart detail: %v", err)
+			continue
```
ywk253100: Should return the error directly, two reasons:
1. If we keep the process logic even if got an error already, the users don't know it in the portal, the replication is marked as succeed, but actually it is not.
2.  To keep consistence with other adapters

src/replication/adapter/helmhub/client.go
mmpei: the helm hub URL is fixed. and it is from offical. so secure should be checked. also refering to dockerhub.

src/replication/adapter/helmhub/adapter.go
mmpei: ok

src/replication/adapter/helmhub/chart_registry.go
mmpei: ok

src/replication/adapter/helmhub/chart_registry.go
mmpei: ok

src/replication/adapter/helmhub/client.go
```diff
@@ -0,0 +1,116 @@
...
+	err = json.Unmarshal(body, list)
+	if err != nil {
+		return nil, fmt.Errorf("unmarshal chart detail response error: %v", err)
+	}
+
+	return list, nil
+}
+
+func (c *Client) checkHealthy() error {
+	request, err := http.NewRequest(http.MethodGet, baseURL, nil)
```
steven-zou: Shall we check the` index.yaml`  service endpoint? Seems the base URL is the root URL of the website?

src/replication/adapter/helmhub/client.go
```diff
@@ -0,0 +1,116 @@
...
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusNotFound {
```
steven-zou: How about changing to the following code?
```go
if resp.StatusCode == http.StatusNotFound {
    return nil, ErrHTTPNotFound
}

if resp.StatusCode != http.StatusOK {
   return nil, fmt.Errorf("fetch chart detail error %d: %s", resp.StatusCode, string(body))
}
```

src/replication/adapter/helmhub/chart_registry.go
```diff
@@ -0,0 +1,145 @@
...
+	for _, v := range versionList.Data {
+		if v.Attributes.Version == version {
+			return a.download(v)
+		}
+	}
+	return nil, errors.New("chart not found")
+}
+
+func (a *adapter) download(version *chartVersion) (io.ReadCloser, error) {
+	if version.Attributes.URLs == nil || len(version.Attributes.URLs) == 0 || len(version.Attributes.URLs[0]) == 0 {
```
steven-zou: `if version.Attributes.URLs == nil || len(version.Attributes.URLs) == 0 || len(version.Attributes.URLs[0]) == 0` ==>
`
if len(version.Attributes.URLs) == 0 || len(version.Attributes.URLs[0]) == 0
`

src/replication/adapter/helmhub/chart_registry.go
```diff
@@ -0,0 +1,145 @@
...
+				},
+			})
+		}
+	}
+	return resources, nil
+}
+
+func (a *adapter) ChartExist(name, version string) (bool, error) {
+	versionList, err := a.client.fetchChartDetail(name)
+	if err != nil && err == ErrHTTPNotFound {
```
steven-zou:
```go

if err!=nil {
  if err == ErrHTTPNotFound {
    err = nil
  }

  return false, err
}

File: src/replication/adapter/helmhub/adapter.go [hunk 9]
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func newAdapter(registry *model.Registry) (*adapter, error) {
+	return &adapter{
+		registry: registry,
+		client:   NewClient(registry),
+	}, nil
+}
+
+func (a *adapter) Info() (*model.RegistryInfo, error) {
```
steven-zou: From the implementation code, we can see that the helm hub adapter does not support push mode. Have we considered declaring such capability too?

@mmpei @ywk253100

src/replication/adapter/helmhub/chart_registry.go
mmpei: looks strange.  but can change to the nested struct.

src/replication/adapter/helmhub/client.go
mmpei: i don't know what you mean. but i couldn't find any related API. and i want to use `/charts` as health URL at first, but response body is too large.

src/replication/adapter/helmhub/adapter.go
mmpei: it seams that there are some restrictions in UI. maybe hard code now.
agree with @ywk253100

PR #8256, opened 2019-07-10
Retention: Implement Evaluator: Retain if created less than x days ago

### LINKED ISSUE #6661: Tag Retention: Filter Implementation: Retain tags no older than n days
> This issue is a sub-task of the Tag Retention Policies epic (#6654)

Implement the "Delete tags older than `n` days" filter

nlowe:
This ports my WIP implementation to match the updated proposal.

Fixes #6661 

Signed-off-by: Nathan Lowe <public@nlowe.me>

### DIFF
--- a/src/pkg/retention/policy/rule/lastx/evaluator_test.go
+++ b/src/pkg/retention/policy/rule/lastx/evaluator_test.go
@@ -1,0 +1,69 @@
+package lastx
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/goharbor/harbor/src/pkg/retention/policy/rule"
+	"github.com/goharbor/harbor/src/pkg/retention/res"
+	"github.com/stretchr/testify/require"
+)
+
+func TestEvaluator_New(t *testing.T) {
+	tests := []struct {
+		Name      string
+		args      rule.Parameters
+		expectedX int
+	}{
+		{Name: "Valid", args: map[string]rule.Parameter{ParameterX: 3}, expectedX: 3},
+		{Name: "Default If Negative", args: map[string]rule.Parameter{ParameterX: -3}, expectedX: DefaultX},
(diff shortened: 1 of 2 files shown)


src/pkg/retention/policy/rule/lastx/evaluator_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/goharbor/harbor/src/pkg/retention/policy/rule"
+	"github.com/goharbor/harbor/src/pkg/retention/res"
+	"github.com/stretchr/testify/require"
+)
+
+func TestEvaluator_New(t *testing.T) {
```
steven-zou: Let's drive the test cases with the testing framework `github.com/stretchr/testify` from now on.

nlowe: Is there a proposal somewhere pitching this? I'm fine changing it out but it feels like a lot of added overhead for simple tests like this.

steven-zou: @nlowe 

No proposal about changing UT approach. The reason I'd like to recommend it because we have introduced that framework a long time ago and some UT cases have implemented with it. The previous cases will be kept as they're. We'd like to apply a unified way for the newly introduced cases. Obviously, testify is better than the native test because more libraries are provided and well-formated results are supported.

I need to admit that this is a recommended approach, not mandatory as we did not raise the proposal to discuss it (actually I'm doubting that does it really need a formal proposal to discuss). If you insist on the native case way, I'm fine with that.

nlowe:
@steven-zou `continuous-integration/vic/push` is failing. I'm guessing this is because the source branch is in the main repo and not my fork, is that correct?

steven-zou:
> @steven-zou `continuous-integration/vic/push` is failing. I'm guessing this is because the source branch is in the main repo and not my fork, is that correct?

@nlowe 

Just ignore it.

PR #8317, opened 2019-07-17
implement label and regexp selectors

steven-zou:


src/pkg/retention/res/selectors/label/selector.go
```diff
@@ -38,10 +38,22 @@ type selector struct {
 	labels []string
 }
 
-// Select candidates by regular expressions
+// Select candidates by the labels
 func (s *selector) Select(artifacts []*res.Candidate) ([]*res.Candidate, error) {
-	// TODO: REPLACE SAMPLE CODE WITH REAL IMPLEMENTATION
-	return artifacts, nil
+	selected := make([]*res.Candidate, 0)
```
nlowe: `append` on a nil slice makes a new slice for you, so we don't have to initialize it here. We can also make it a named return parameter to drop this line.

Additionally, the `len` check on the input shouldn't be needed, since the for loop will exit immediately if there's nothing to loop over.

What about something like this?

```go
func (s *selector) Select(artifacts []*res.Candidate) (selected []*res.Candidate, err error) {
    for _, a := range artifacts {
        if isMatched(s.labels, a.Labels, s.decoration) {
            selected = append(selected, a)
        }
    }
}
```

src/pkg/retention/res/selectors/label/selector_test.go
```diff
@@ -0,0 +1,121 @@
...
+// TestWithLabelsUnMatched tests the selector of `with` labels but nothing matched
+func (suite *LabelSelectorTestSuite) TestWithLabelsUnMatched() {
+	withNothing := &selector{
+		decoration: With,
+		labels:     []string{"label6"},
+	}
+
+	selected, err := withNothing.Select(suite.artifacts)
+	require.NoError(suite.T(), err)
+	assert.Equal(suite.T(), 0, len(selected))
```
nlowe: We should also probably check that the correct one was selected.

src/pkg/retention/res/selectors/regexp/selector.go
```diff
@@ -15,32 +15,73 @@
...
 	pattern string
 }
 
 // Select candidates by regular expressions
 func (s *selector) Select(artifacts []*res.Candidate) ([]*res.Candidate, error) {
-	// TODO: REPLACE SAMPLE CODE WITH REAL IMPLEMENTATION
-	return artifacts, nil
+	selected := make([]*res.Candidate, 0)
+
+	if len(artifacts) == 0 {
```
nlowe: I don't think you need an explicit check for 0 length here, won't the `for` loop below simply do nothing if the input is empty?

src/pkg/retention/res/selectors/regexp/selector.go
```diff
@@ -15,32 +15,73 @@
 package regexp
 
 import (
+	"github.com/bmatcuk/doublestar"
```
nlowe: Note that this ***isn't*** a regex matching library, it does something slightly different and I'm not sure how applicable it is to Harbor. If we're going to use this then the selector should probably be renamed accordingly (along with documentation being updated)

src/pkg/retention/res/selectors/regexp/selector_test.go
```diff
@@ -0,0 +1,159 @@
...
+	assert.Equal(suite.T(), 2, len(selected))
+
+	repoMatches2 := &selector{
+		decoration: RepoMatches,
+		pattern:    "har*",
+	}
+
+	selected, err = repoMatches2.Select(suite.artifacts)
+	require.NoError(suite.T(), err)
+	assert.Equal(suite.T(), 1, len(selected))
```
nlowe: Same as the other test, we should probably check that we're matching the right thing.

src/pkg/retention/res/selectors/regexp/selector.go
steven-zou: Yes, that's not regexp. We'd like to keep the same pattern framework in the whole harbor. So, I agree with the rename.

src/pkg/retention/res/selectors/label/selector_test.go
steven-zou: Sure

src/pkg/retention/res/selectors/regexp/selector_test.go
steven-zou: Sure

steven-zou:
Merge back to the feature branch to check the CI status to avoid code rebase.

PR #8324, opened 2019-07-18
Travis ci build prepare container in local

ninjadq:
build prepare container before running ut case


reasonerjt:
Better set the tag to "prepare:travis-ci" and let `make/prepare` use this tag

PR #8333, opened 2019-07-18
Fix OnBoardGroup issue

stonezdj:
Fix issue when adding a HTTP user group to a project member, returns HTTP 500 error.


reasonerjt:
Not relevant to this commit, but I noticed a bad smell here
`err := group.OnBoardUserGroup(u, "GroupName", "GroupType")`

We should encapsulate the contraint of DB within package `group`

PR #8335, opened 2019-07-18
Add API to ping OIDC endpoint

reasonerjt:
This commit adds an API to help admin verify the OIDC endpoint is a
valid one.

### DIFF
--- a/docs/swagger.yaml
+++ b/docs/swagger.yaml
@@ -3478,6 +3478,39 @@ paths:
           description: The robot account is not found.
         '500':
           description: Unexpected internal errors.
+  '/system/oidc/ping':
+    post:
+      summary: Test the OIDC endpoint.
+      description: Test the OIDC endpoint, the setting of the endpoint is provided in the request.  This API can only
+        be called by system admin.
+      tags:
+        - Products
+        - System
+      parameters:
+        - name: endpoint
+          in: body
+          description: Request body for OIDC endpoint to be tested.
+          required: true
+          schema:
+            type: object
+            properties:
+              url:
+                type: string
+                description: The URL of OIDC endpoint to be tested.
+              verify_cert:
+                type: boolean
+                description: Whether the certificate should be verified
+      responses:
+        '200':
+          description: The specified robot account is successfully deleted.
+        '400':
+          description: The ping failed
+        '401':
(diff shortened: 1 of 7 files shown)


docs/swagger.yaml
```diff
@@ -3478,6 +3478,39 @@ paths:
...
+            properties:
+              url:
+                type: string
+                description: The URL of OIDC endpoint to be tested.
+              verify_cert:
+                type: boolean
+                description: Whether the certificate should be verified
+      responses:
+        '200':
+          description: The specified robot account is successfully deleted.
```
wy65701436: please fix the description.

docs/swagger.yaml
```diff
@@ -3478,6 +3478,39 @@ paths:
...
+        '400':
+          description: The ping failed
+        '401':
+          description: User need to log in first.
+        '403':
+          description: User does not have permission to call this API
+        '500':
```
wy65701436: and there is no 500 return in the API.

docs/swagger.yaml
stonezdj: It should be Ping OIDC success

src/core/api/oidc.go
```diff
@@ -0,0 +1,56 @@
...
+func (oa *OIDCAPI) Ping() {
+	var c oidc.Conn
+	if err := oa.DecodeJSONReq(&c); err != nil {
+		log.Error("Failed to decode JSON request.")
+		oa.SendBadRequestError(err)
+		return
+	}
+	if err := oidc.TestEndpoint(c); err != nil {
+		log.Errorf("Failed to verify connection: %+v, err: %v", c, err)
+		oa.SendBadRequestError(err)
```
wy65701436: what's the difference between there two BadRequest? Does the second one mean the endpoint is not reachable?

reasonerjt: yes, kind of, either unreachable, or it's not an OIDC provider.

docs/swagger.yaml
reasonerjt: Will do

docs/swagger.yaml
reasonerjt: Oh, yes, will do.

PR #8353, opened 2019-07-19
Make jobservice use the proxy env variables (including no_proxy)

randomswdev:
jobservice reads the proxy environment variables, but do not consider the no_proxy variable.
Modified the code to leverage the go proxy handling function that reads the env variables including no_proxy. The same function is used in all the other containers.

### DIFF
--- a/src/jobservice/hook/hook_client.go
+++ b/src/jobservice/hook/hook_client.go
@@ -21,18 +21,10 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
-	"net/url"
-	"os"
 	"strings"
 	"time"
 
 	"context"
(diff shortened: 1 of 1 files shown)


steven-zou:
@randomswdev 

Thank you very much for doing a contribution to Harbor. I think this is your 1st commit, really hope there will be more in the future!

randomswdev:
Do you plan to port this also to release 1.8? If so, are you going to cherry pick it to the branch release-1.8.2? Or do I have to cherry pick it?

steven-zou:
@randomswdev 

So far, we did not hear voices from the community about such an issue, so maybe it is ok if we only contain it in the next new release.

PR #8366, opened 2019-07-23
Implement the task management in retention manager

ywk253100:
Implement the task management in retention manager

### DIFF
--- a/make/migrations/postgresql/0010_1.9.0_schema.up.sql
+++ b/make/migrations/postgresql/0010_1.9.0_schema.up.sql
@@ -39,12 +39,12 @@ create table retention_execution
 
 create table retention_task
 (
-	id integer PRIMARY KEY NOT NULL,
+	id SERIAL NOT NULL,
 	execution_id integer,
-	rule_id integer,
-	rule_display_text varchar(255),
-	artifact varchar(255),
-	timestamp time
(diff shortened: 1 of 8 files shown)


make/migrations/postgresql/0010_1.9.0_schema.up.sql
```diff
@@ -39,12 +39,12 @@ create table retention_execution
 
 create table retention_task
 (
-	id integer PRIMARY KEY NOT NULL,
+	id SERIAL NOT NULL,
```
steven-zou: missing repository

src/pkg/retention/controllers/retention.go
```diff
@@ -155,15 +156,15 @@ func (r *RetentionAPI) ListRetentionExecHistory() {
 		r.SendInternalServerError(err)
 		return
 	}
-	query := &q.Query{
-		PageNumber: page,
-		PageSize:   size,
-	}
-	his, err := r.manager.ListHistories(eid, query)
+	tasks, err := r.manager.ListTasks(&q.TaskQuery{
```
steven-zou: should put into `retention.APIController`

ywk253100: Will change it in the next PR.

Issue #8122, opened 2019-06-21
Support nested LDAP groups


mungo312:
In our company we are using nested groups in our Active Directory. This means:
User <-> OrgGroup <-> ResourceGroup 
The resource group is configured in Harbor

At the moment the Group membership is not detected in harbor, in my tests the user did not get admin privileges. 

**Steps to reproduce the problem:**
Use a LDAP with nested Groups.

**Versions:**
Please specify the versions of following systems.
- harbor version: 1.8.0

Typositoire:
Any idea when this will be actually done ?

Typositoire:
I tried to get a patch out but I failed horribly due to my setup, I can help getting this implemented if someone has time.

Typositoire:
Actually I've got a really crud implementation if you guys are interested ?

ConorNevin:
I believe this change was rolled back - maybe this should be reopened and moved back into the Backlog?

narendrakum:
Few of the customers using harbor tile in TKGI environment requested for nested ldap group support. Spoke with engineering and reopened this issue..
Current harbor version: 2.1.1

gwiersma:
Following this issue. Our organization really needs this feature.

vikingtoby:
Agreed, this is essential for us as well.

hashCode1:
same this is important for us

lavih:
I will be more than happy for this feature as well :)

RobMokkink:
This RFE is open for a long time now, any thoughts when this will be implemented?

xcorp:
Chiming in to request this to

robertdebock:
I see the issue is in the ["v1.9 completed" row](https://github.com/orgs/goharbor/projects/1#column-4940706) of the [Harbor Project Board](https://github.com/orgs/goharbor/projects/1#card-24627074). 

On the other hand, I see this issue is assigned to the [backlog](https://github.com/goharbor/harbor/labels/backlog) if [candidate/2.7.0](https://github.com/goharbor/harbor/labels/candidate%2F2.7.0)

This issue is open however.

The [related issue](https://github.com/goharbor/harbor/issues/9492) is in the ["prioritised backlog"](https://github.com/orgs/goharbor/projects/1#column-16774322)

Is there anything we can do to improve the issue so it can be picked up?

kkrombach:
Hello, we also would love to see this feature in our company :)

anokun7:
[...]

PR #8378, opened 2019-07-23
Search for LDAP_MATCHING_RULE_IN_CHAIN groups

Typositoire:
More details here: https://ldapwiki.com/wiki/1.2.840.113556.1.4.1941

Implementation of https://github.com/goharbor/harbor/issues/8122

It works but if someone want to use it as a starting plate, go for it!

### DIFF
--- a/src/common/utils/ldap/ldap.go
+++ b/src/common/utils/ldap/ldap.go
@@ -230,9 +230,14 @@ func (session *Session) SearchUser(username string) ([]models.LdapUser, error) {
 		}
 
 		for _, groupEntry := range result.Entries {
-			nestedGroupDNList = append(nestedGroupDNList, strings.TrimSpace(groupEntry.DN))
-			log.Debugf("Found group %v", groupEntry.DN)
+			if !contains(u.GroupDNList, groupEntry.DN) {
+				nestedGroupDNList = append(nestedGroupDNList, strings.TrimSpace(groupEntry.DN))
+				log.Debugf("Found group %v", groupEntry.DN)
+			} else {
+				log.Debugf("%v is already in GroupDNList", groupEntry.DN)
+			}
 		}
+
 		u.GroupDNList = append(u.GroupDNList, nestedGroupDNList...)
 		log.Debugf("Done searching for nested groups")
 
@@ -440,4 +445,13 @@ func createNestedGroupFilter(userDN string) string {
 	filter := ""
 	filter = "(&(objectClass=group)(member:1.2.840.113556.1.4.1941:=" + userDN + "))"
 	return filter
+}
+
+func contains(s []string, e string) bool {
+	for _, a := range s {
+		if a == e {
+			return true
+		}
+	}
+	return false
 }
--- a/src/common/utils/ldap/ldap.go
+++ b/src/common/utils/ldap/ldap.go
@@ -220,6 +220,22 @@ func (session *Session) SearchUser(username string) ([]models.LdapUser, error) {
 			}
 			u.GroupDNList = groupDNList
 		}
+
+		log.Debugf("Searching for nested groups")
+		nestedGroupDNList := []string{}
+		nestedGroupFilter := createNestedGroupFilter(ldapEntry.DN)
+		result, err := session.SearchLdap(nestedGroupFilter)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, groupEntry := range result.Entries {
+			nestedGroupDNList = append(nestedGroupDNList, strings.TrimSpace(groupEntry.DN))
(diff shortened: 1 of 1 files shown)


Typositoire:
Force pushed to `amend -s`

Typositoire:
Any chances to get this reviewed ?

Typositoire:
Ping ?

reasonerjt:
@Typositoire sorry for the late response, we'll do it.

Typositoire:
Thanks!

src/common/utils/ldap/ldap.go
```diff
@@ -220,6 +220,22 @@ func (session *Session) SearchUser(username string) ([]models.LdapUser, error) {
...
+		result, err := session.SearchLdap(nestedGroupFilter)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, groupEntry := range result.Entries {
+			nestedGroupDNList = append(nestedGroupDNList, strings.TrimSpace(groupEntry.DN))
+			log.Debugf("Found group %v", groupEntry.DN)
+		}
+		u.GroupDNList = append(u.GroupDNList, nestedGroupDNList...)
```
stonezdj: Top level groups present in both u.GroupDNList and nestedGroupDNList, could you please remove duplicated items in the u.GroupDNList?
for example:
CN=cna_users,CN=Users,DC=harbor,DC=com is a top level group.
the u.GroupDNList might contains two items with DN: CN=cna_users,CN=Users,DC=harbor,DC=com
```
CN=cna_users,CN=Users,DC=harbor,DC=com 
CN=harbor_users,CN=Users,DC=harbor,DC=com 
CN=cna_users,CN=Users,DC=harbor,DC=com
```

Typositoire: Oh right! Thanks! I'm at an appointment right now but will do when i'm back.

Typositoire:
I've filtered the new search entry that were already in `u.GroupDNList` when building the list of nested groups. Debug logs output msg when duplicate are found and excluded from nested groups list.

Typositoire:
@stonezdj Do you know if this would be able to be part of 1.9.0 release ?

src/common/utils/ldap/ldap.go
```diff
@@ -220,6 +220,27 @@ func (session *Session) SearchUser(username string) ([]models.LdapUser, error) {
...
+
+		log.Debugf("Searching for nested groups")
+		nestedGroupDNList := []string{}
+		nestedGroupFilter := createNestedGroupFilter(ldapEntry.DN)
+		result, err := session.SearchLdap(nestedGroupFilter)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, groupEntry := range result.Entries {
```
reasonerjt: Thanks for the PR.  I think this will be part of 1.9.

However, it looks we only need a `difference(s1 []string, s2[]string) []string` to return the elements in `result.Entries` but not in `u.GroupDNList`?  So that the nested loop can be eliminated?

Additionally, could you write UT?

Typositoire: +1 on the nested loop, I'll fix that today. As for UT I'll try, I'm far from being a good enough developer for that xD (Not my line of work) So if you want to help on that it'd be appreciate and probably help me learn a bit :p

Typositoire: So I'm looking at that and you can't really remove the loop, since `result.Entries` is a `[]*ldap.Entry`. Unless I'm missing something you still need to go through it and get the DN from each `ldap.Entry`. I can clean it by moving it to a single function so that the above code would be just 1 line: `groupList := difference(result.Entries, u.GroupDNList)`

Typositoire: @reasonerjt  ^

reasonerjt:
Per discussion with David on Slack I'll merge the PR and he will refine it afterwards.

stonezdj:
It seems this PR caused some issues:
#9092 
#9110 
#9326 

@Typositoire could you please revise your code?

xaleeks:
Since it introduces break chain, can you please review this, otherwise we need to take this out until we have bandwidth to inspect carefully @Typositoire

Typositoire:
I guess I can create a switch to turn it off ? (Feature toggle) I don't really have the bandwidth to try and fix all those uses cases unfortunately. #9092 Looks like just bad escaping on the filter. The other are just missing the `LDAP_MATCHING_RULE_IN_CHAIN` feature which is not enabled by default in anything other than Active Directory. I'll add the toggle tomorrow and try to find some documentation on how to enable that on anything other than ActiveDirectory to add to the repo.

PR #8406, opened 2019-07-25
add ut for tag retention controller

bitsf:
Change-Id: I6d9412126e03b697d93ac512b75b5007223aa926

### DIFF
--- a/src/core/api/retention.go
+++ b/src/core/api/retention.go
@@ -131,7 +131,10 @@ func (r *RetentionAPI) GetMetadatas() {
     ]
 }
 `
-	r.WriteJSONData(data)
+	w := r.Ctx.ResponseWriter
+	w.Header().Set("Content-Type", "application/json; charset=utf-8")
+	w.WriteHeader(http.StatusOK)
+	w.Write([]byte(data))
 }
 
 // GetRetention Get Retention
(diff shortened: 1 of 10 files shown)


src/core/api/retention.go
```diff
@@ -175,13 +178,13 @@ func (r *RetentionAPI) CreateRetention() {
...
 		if proj == nil {
-			r.SendBadRequestError(fmt.Errorf("Invalid Project id %d", p.Scope.Reference))
+			r.SendBadRequestError(fmt.Errorf("invalid Project id %d", p.Scope.Reference))
 		}
 	default:
 		r.SendBadRequestError(fmt.Errorf("scope %s is not support", p.Scope.Level))
 		return
 	}
-	if err = retentionController.CreateRetention(p); err != nil {
+	if _, err = retentionController.CreateRetention(p); err != nil {
```
ywk253100: We should return the ID just like https://github.com/goharbor/harbor/blob/master/src/core/api/replication_policy.go#L104

src/core/api/retention.go
```diff
@@ -175,13 +178,13 @@ func (r *RetentionAPI) CreateRetention() {
...
 		r.SendInternalServerError(err)
 		return
 	}
```
ywk253100: Handle the error in line 193

PR #8429, opened 2019-07-26
Only listing chart repositories when chart server is enabled

ywk253100:
Only listing chart repositories when chart server is enabled

### DIFF
--- a/src/pkg/retention/launcher.go
+++ b/src/pkg/retention/launcher.go
@@ -72,20 +72,22 @@ type Launcher interface {
 func NewLauncher(projectMgr project.Manager, repositoryMgr repository.Manager,
 	retentionMgr Manager) Launcher {
 	return &launcher{
-		projectMgr:       projectMgr,
-		repositoryMgr:    repositoryMgr,
-		retentionMgr:     retentionMgr,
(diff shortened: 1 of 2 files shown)


src/core/api/base.go
```diff
@@ -116,7 +116,7 @@ func Init() error {
 
 	retentionMgr = retention.NewManager()
 
-	retentionLauncher = retention.NewLauncher(projectMgr, repositoryMgr, retentionMgr)
+	retentionLauncher = retention.NewLauncher(projectMgr, repositoryMgr, retentionMgr, config.WithChartMuseum())
```
steven-zou: we can directly call this config getting at the core side.

src/pkg/retention/launcher.go
```diff
@@ -278,7 +280,8 @@ func getProjects(projectMgr project.Manager) ([]*res.Candidate, error) {
 	return candidates, nil
 }
 
-func getRepositories(projectMgr project.Manager, repositoryMgr repository.Manager, projectID int64) ([]*res.Candidate, error) {
+func getRepositories(projectMgr project.Manager, repositoryMgr repository.Manager,
```
steven-zou: dep.Client.GetCandidates has the same problem

PR #8435, opened 2019-07-28
Aliyun ACR adapter

kofj:
1. aliyun ACR registry adapter.
2. Unit test case.

Signed-off-by: 疯魔慕薇 <kfanjian@gmail.com>


src/replication/adapter/aliacr/adapter.go
```diff
@@ -0,0 +1,260 @@
...
+		SupportedTriggers: []model.TriggerType{
+			model.TriggerTypeManual,
+			model.TriggerTypeScheduled,
+		},
+	}
+	return
+}
+
+// PrepareForPush nothing.
+func (a *adapter) PrepareForPush([]*model.Resource) error { return nil }
```
ywk253100: The functions `PrepareForPush` and `HealthCheck` are same with the native adapter, right? If they are, we should omit them

src/replication/adapter/aliacr/adapter_test.go
```diff
@@ -0,0 +1,275 @@
...
+	defer s.Close()
+	info, err := a.Info()
+	assert.Nil(t, err)
+	assert.NotNil(t, info)
+
+	assert.EqualValues(t, 1, len(info.SupportedResourceTypes))
+	assert.EqualValues(t, model.ResourceTypeImage, info.SupportedResourceTypes[0])
+}
+
+func TestAdapter_HealthCheck(t *testing.T) {
```
ywk253100: Remove it and `Test_adapter_PrepareForPush` as the two methods have been removed from the adapter

PR #8439, opened 2019-07-28
feat(helm-chart,quota): count quota support for helm chart

heww:


src/core/api/chart_repository.go
```diff
@@ -267,16 +282,34 @@ func (cra *ChartRepositoryAPI) DeleteChartVersion() {
 	chartName := cra.GetStringFromPath(nameParam)
 	version := cra.GetStringFromPath(versionParam)
 
+	key := fmt.Sprintf("chart:%s:%s:%s", cra.namespace, chartName, version)
```
ywk253100: Can we move all the code related with quota into a separate middleware or handler just like other quota handlers?

src/core/api/chart_repository.go
```diff
@@ -554,3 +651,29 @@ func chartFullName(namespace, chartName, version string) string {
 	}
 	return fmt.Sprintf("%s/%s:%s", namespace, chartName, version)
 }
+
+// Returns true if chart of version exists in namespace
+var chartVersionExists = func(namespace, chartName, version string) bool {
```
ywk253100: Why the `chartVersionExists` and `parseChart` are variables? Can't we just define the function?

src/core/api/utils.go
```diff
@@ -278,3 +280,22 @@ func repositoryExist(name string, client *registry.Repository) (bool, error) {
 	}
 	return len(tags) != 0, nil
 }
+
+func requireLock(key string) (*redis.Mutex, error) {
```
ywk253100: There is already a same function for image quota? Can we use the same one?

src/core/api/chart_repository.go
```diff
@@ -347,18 +426,36 @@ func (cra *ChartRepositoryAPI) DeleteChart() {
 	// Get other parameters from the request
 	chartName := cra.GetStringFromPath(nameParam)
 
+	key := fmt.Sprintf("chart:%s:%s", cra.namespace, chartName)
+	lock, err := requireLock(key)
```
ywk253100: Should lock to the version level?

src/core/api/chart_repository.go
```diff
@@ -267,16 +282,34 @@ func (cra *ChartRepositoryAPI) DeleteChartVersion() {
...
+	}
+
 	if err := chartController.DeleteChartVersion(cra.namespace, chartName, version); err != nil {
 		cra.ParseAndHandleError("fail to delete chart version", err)
 		return
 	}
+
+	if err := quotaMgr.SubtractResources(types.ResourceList{types.ResourceCount: 1}); err != nil {
+		cra.SendInternalServerError(fmt.Errorf("failed to subtract resources for quota, error: %v", err))
+	}
```
wy65701436: add return

src/core/api/chart_repository.go
heww: The `chart:namespace:chartName` will be locked when upload, delete the chart version now, so when delete chart, no lock needed for the chart version.

And the resource will be free when chart deleted success.

src/core/api/utils.go
heww: Move it to the `common/utils/redis` package, it's a helper function to get lock of the key using the default redis connection, and require the lock by the same function for image quota.

src/core/api/chart_repository.go
heww: They are defined as variables so that can be mocked in chart repository test case.

src/core/api/chart_repository.go
heww: It is difficulty to move this to middleware for calculating the resources required when upload chart package need to parse the chart info from archive file and chart info need to parse after `rewriteFileContent`. If move this to middleware, `rewriteFileContent` also need to move.

src/core/api/chart_repository.go
heww: Added

src/core/middlewares/chain.go
```diff
@@ -59,6 +61,7 @@ func (b *DefaultCreator) Create() *alice.Chain {
 
 func (b *DefaultCreator) geMiddleware(mName string) alice.Constructor {
 	middlewares := map[string]alice.Constructor{
+		CHART:            func(next http.Handler) http.Handler { return chart.New(next) },
```
ywk253100: Can we create a separate chain for chart as the requests sent to registry don't need to be handled by the chart handler?

src/core/middlewares/config.go
```diff
@@ -26,5 +27,8 @@ const (
 	COUNTQUOTA       = "countquota"
 )
 
+// ChartMiddlewares middlewares for chart server
+var ChartMiddlewares = []string{CHART}
```
ywk253100: Where is the middleware chain initialized for chart?

src/chartserver/reverse_proxy.go
```diff
@@ -36,20 +37,29 @@ type ProxyEngine struct {
...
+		ErrorLog: log.New(os.Stdout, "", log.Ldate|log.Ltime|log.Lshortfile),
+		Director: func(req *http.Request) {
+			director(target, cred, req)
+		},
+		ModifyResponse: modifyResponse,
+	}
+
+	if len(chains) > 0 {
+		hlog.Info("New chart server traffic proxy with middlewares")
+		engine = chains[0].Then(engine)
```
wy65701436: why use chains[0]?

src/common/utils/redis/helper.go
```diff
@@ -103,12 +107,126 @@ type Options struct {
 	maxRetry   int
 }
 
+var (
+	opt     *Options
+	optOnce sync.Once
+
+	defaultDelay    = int64(1) // 1 second
```
wy65701436: make them configurable?

src/core/middlewares/chart/handler.go
```diff
@@ -0,0 +1,71 @@
...
+// ServeHTTP manifest ...
+func (h *chartHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	interceptor := h.getInterceptor(req)
+	if interceptor == nil {
+		h.next.ServeHTTP(rw, req)
+		return
+	}
+
+	if err := interceptor.HandleRequest(req); err != nil {
+		log.Warningf("Error occurred when to handle request in count quota handler: %v", err)
```
wy65701436: log -- in chart count quota handler？

src/core/middlewares/chart/util.go
```diff
@@ -0,0 +1,118 @@
...
+	formFieldNameForChart = "chart"
+)
+
+var (
+	controller     *chartserver.Controller
+	controllerErr  error
+	controllerOnce sync.Once
+)
+
+func chartController() (*chartserver.Controller, error) {
```
ywk253100: There is an initialized chart controller https://github.com/goharbor/harbor/blob/master/src/core/api/chart_repository.go#L50, use this one?

src/core/middlewares/chart/handler.go
```diff
@@ -0,0 +1,71 @@
...
+		http.Error(rw, util.MarshalError("InternalError", fmt.Sprintf("Error occurred when to handle request in count quota handler: %v", err)),
+			http.StatusInternalServerError)
+		return
+	}
+
+	w := util.NewCustomResponseWriter(rw)
```
wy65701436: move the NewCustomResponseWriter into chart proxy ServeHTTP()?

src/core/middlewares/chart/matcher.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/goharbor/harbor/src/common/dao"
+	"github.com/goharbor/harbor/src/common/utils/log"
+	"github.com/goharbor/harbor/src/core/middlewares/interceptor"
+	"github.com/goharbor/harbor/src/core/middlewares/interceptor/quota"
+	"github.com/goharbor/harbor/src/core/middlewares/util"
+	"github.com/goharbor/harbor/src/pkg/types"
+)
+
+var (
+	deleteRe = regexp.MustCompile(`^/api/chartrepo/(?P<namespace>\w+)/charts/(?P<name>\w+)/(?P<version>[\w\d\.]+)/?$`)
```
wy65701436: deleteChartRepo or deleteChartSth .... ?

src/core/middlewares/chart/matcher.go
```diff
@@ -0,0 +1,142 @@
...
+	if len(matches) <= 1 {
+		return false
+	}
+
+	project, err := dao.GetProjectByName(matches[1])
+	if err != nil {
+		log.Errorf("Failed to get project %s, error: %v", matches[1], err)
+		return false
+	}
+
```
wy65701436: add project == nil checking?

src/core/middlewares/chain.go
heww: `DefaultCreator` is the factory to make chain, it can create chain with different middleware and it is not the creator for registry only.

src/core/middlewares/chart/matcher.go
```diff
@@ -0,0 +1,142 @@
...
+	chart, err := parseChart(req)
```
wy65701436: add chart == nil check？

src/core/middlewares/chart/util.go
wy65701436: why put this into middleware? is it possible that move it into ./src/chartserver?

src/core/middlewares/config.go
wy65701436: in the chart proxy.

src/core/middlewares/interceptor/interceptor.go
```diff
@@ -0,0 +1,34 @@
...
+	// HandleRequest ...
+	HandleRequest(*http.Request) error
+
+	// HandleResponse won't return any error
+	HandleResponse(http.ResponseWriter, *http.Request)
+}
+
+// Matcher ...
+type Matcher interface {
+	Match(*http.Request) bool
```
wy65701436: Match is actually to match and set context? add some comments here?

src/core/middlewares/config.go
heww: Sorry, fixed in `chart_repository.go` now

src/core/middlewares/interceptor/interceptor.go
```diff
@@ -0,0 +1,34 @@
...
+	SetupInterceptor(*http.Request) Interceptor
```
wy65701436: SetupInteceptor sounds like, it needs to pass an interceptor into the func.

But, it actually always returns a quota interceptor.

src/core/middlewares/chart/util.go
heww: chart controller here will be call the middleware and chart controller in `chart_repository.go` will be initialized with middleware, maybe have cycle import problem.

src/core/middlewares/interceptor/quota/quota.go
```diff
@@ -0,0 +1,147 @@
...
+type quotaInterceptor struct {
+	opts      *Options
+	resources types.ResourceList
+	mutexes   []*redis.Mutex
+}
+
+// HandleRequest ...
+func (qi *quotaInterceptor) HandleRequest(req *http.Request) (err error) {
+	defer func() {
+		if err != nil {
```
wy65701436: the err can be returned by reserve? then, we will not release the mutex?

src/core/middlewares/interceptor/interceptor.go
heww: `Matcher` renamed to be `Builder` which only has one method `Build`.

src/core/middlewares/interceptor/interceptor.go
heww: `Matcher` renamed to be `Builder` which only has one method Build.

src/core/middlewares/chart/util.go
heww: this method depend on `config` pkg, move it to `chartserver` not make sense.

src/core/middlewares/chart/matcher.go
heww: if err is nil, chart must not be nil.

src/core/middlewares/chart/handler.go
heww: chart proxy is initialized in `chartserver` pkg, move it to chartserver make the `chartserver` depend on `core` pkg.

src/common/utils/redis/helper.go
heww: configurations are implemented in `DefaultOptions` function

PR #8444, opened 2019-07-29
get execution status on the fly

bitsf:
Change-Id: I0be88c728194b815be3e8fcd24b4c2457bb64ade

### DIFF
--- a/src/pkg/retention/dao/retention.go
+++ b/src/pkg/retention/dao/retention.go
@@ -3,7 +3,6 @@ package dao
 import (
 	"errors"
 	"fmt"
-
 	"github.com/goharbor/harbor/src/common/dao"
 	"github.com/goharbor/harbor/src/pkg/retention/dao/models"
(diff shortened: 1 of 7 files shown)


src/pkg/retention/dao/retention.go
```diff
@@ -85,9 +84,32 @@ func GetExecution(id int64) (*models.RetentionExecution, error) {
...
 		return nil, err
 	}
+	if err := fillStatus(e); err != nil {
+		return nil, err
+	}
 	return e, nil
 }
 
+// fillStatus the priority is InProgress Stopped Failed Succeed
+func fillStatus(exec *models.RetentionExecution) error {
```
ywk253100: Need to fill the `StartTime` and `EndTime` either.

bitsf: Will remove EndTime, and calc on the fly
Keep StartTime

PR #8445, opened 2019-07-29
refactor index registering processes

steven-zou:

refactor index registering processes

- register action performers in the `init` func of `action` package
- register selectors in the `init` func of `selectors` package
- register rule evaluators in the `init` func of `rule` package
- register processor in the `init` func of `alg` package

pass job parameters with JSON format
- add `ToJSON` and `FromJSON` methods to `res.Repository`
- add `ToJSON` and `FromJSON` methods to `lwp.Metadata`

enhance the label selector

fix the OR processor failure issue


src/pkg/retention/policy/lwp/models.go
```diff
@@ -32,14 +34,21 @@ type Metadata struct {
 	Rules []*rule.Metadata `json:"rules"`
 }
 
-// FromMap constructs the metadata struct from map
-func (meta *Metadata) FromMap(m map[string]interface{}) error {
-	mdata, err := json.Marshal(&m)
+// ToJSON marshals metadata to JSON string
+func (m *Metadata) ToJSON() (string, error) {
```
nlowe: I'm struggling to see why this is needed (along with the `FromJSON` pair below. What does this give us that calling `json.Marshal`/`json.Unmarshal` directly doesn't?

src/pkg/retention/policy/lwp/models.go
```diff
@@ -32,14 +34,21 @@ type Metadata struct {
...
+	jsonData, err := json.Marshal(m)
 	if err != nil {
-		return err
+		return "", errors.Wrap(err, "marshal reporitory")
```
nlowe:
```suggestion
		return "", errors.Wrap(err, "marshal repository")
```

src/pkg/retention/policy/rule/index/index.go
```diff
@@ -48,13 +63,78 @@ type IndexedParam struct {
...
-	Meta *IndexMeta
+	Meta *Metadata
+
+	Factory rule.Factory
+}
 
-	Factory Factory
+func init() {
+	// Register latest pushed
+	Register(&Metadata{
```
nlowe: Ditto my comments on other `Register`-like calls in `init` functions. It feels weird to me to call these from a different package from where the thing we're registering is defined.

src/pkg/retention/policy/rule/latestpl/evaluator.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package lastpulled
+package latestpl
```
nlowe: The new package name feels more confusing to me. Was the old one too long?

src/pkg/retention/policy/action/index/index.go
```diff
@@ -12,19 +12,26 @@
...
+
 	"github.com/pkg/errors"
 )
 
 // index for keeping the mapping action and its performer
 var index sync.Map
 
+func init() {
+	// Register retain action
+	Register(action.Retain, action.NewRetainAction)
```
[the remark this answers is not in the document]
steven-zou: To fix importing cycle

src/pkg/retention/policy/alg/index/index.go
```diff
@@ -27,21 +32,26 @@ const (
 // index for keeping the mapping between algorithm and its processor
 var index sync.Map
 
+func init() {
+	// Register or
+	Register(AlgorithmOR, or.New)
```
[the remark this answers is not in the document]
steven-zou: For fixing importing cycle

src/pkg/retention/policy/lwp/models.go
steven-zou: code is not much or less, but it makes the calling easier and clear.

src/pkg/retention/policy/lwp/models.go
steven-zou: fixed

src/pkg/retention/policy/rule/index/index.go
steven-zou: all the registration now happened in their package `init` of the corresponding index file because:

- avoid importing cycle
- the `init` of other concreted packages you mentioned will not be executed because we use interface to call them and did not refer any concreted packages

src/pkg/retention/policy/rule/latestpl/evaluator.go
steven-zou: yes, too long

src/pkg/retention/job.go
```diff
@@ -125,13 +125,13 @@ func logResults(logger logger.Interface, all []*res.Candidate, results []*res.Re
 	op := func(art *res.Candidate) string {
 		if e, exists := hash[art.Hash()]; exists {
 			if e != nil {
-				return "Err"
+				return "ERR"
 			}
 
-			return "X"
+			return "DEL"
```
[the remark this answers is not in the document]
steven-zou: Only here for logging retention and they're treated as `marks` to highlight/emphasize the final action of the retention.

Screen Shot 2019-07-30 at 07 54 07

steven-zou:
> I noticed there is a lot of superfluous whitespace between imports in quite a few places. I wasn't sure what the standard was on that in the project to I didn't make any specific comments on it. Overall the renames LGTM, I did nave a few questions around `ToJSON`/`FromJSON` and moving the `Register` calls:

No blocking comments, so merge this PR first because we have other works based on it.

steven-zou:
The code is `gofmt` by the `goland` IDE, I 'm not sure if it is a `bug` of it.  Leave the space there when deleting the unused importing package.

PR #8453, opened 2019-07-29
Enable quota middleware for internal registry api call

wy65701436:

### DIFF
--- a/src/core/utils/utils.go
+++ b/src/core/utils/utils.go
@@ -19,11 +19,18 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/goharbor/harbor/src/common/http/modifier"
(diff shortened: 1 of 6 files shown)


src/core/utils/utils.go
```diff
@@ -45,6 +52,58 @@ func NewRepositoryClientForUI(username, repository string) (*registry.Repository
 	return registry.NewRepository(repository, endpoint, client)
 }
 
+// NewRepositoryClientForUIWithMiddleware creates a repository client that can only be used to
+// access the internal registry with quota middle
+func NewRepositoryClientForUIWithMiddleware(username, repository string) (*registry.Repository, error) {
```
ywk253100: Can we the client with the core URL?

PR #8469, opened 2019-07-29
fix mis-spelling in handler.go

dzzg:
fix mis-spelling in handler.go


reasonerjt:
@dzzg 

This file will be removed in another PR(#8466), so this is not needed.

I'm closing it.

PR #8478, opened 2019-07-30
Remove unneccessary certificate options for Helm CLI

jessehu:
The --cert-file and --key-file options for Helm CLI are used to identify HTTPS client, i.e. the helm CLI.
It's not mandantory and usually not needed. Most of the users don't know how to get these two files.


jessehu:
@steven-zou please take a look.

PR #8494, opened 2019-07-30
Fix redirect login page when enter harbor through global search

jwangyangls:


jwangyangls:
The pull request cannot merge,because the backend's API isnot ready.

Issue #8499, opened 2019-07-31
Redis connection pool is run out when doing migration in job service 


steven-zou:
https://pivotal-spike.atlassian.net/browse/PKS-351

reasonerjt:
Fixed in #8504

PR #8500, opened 2019-07-31
fix bug #8499: redis connection pool is run out

steven-zou:

The cause is the Redis connection pool used in the job service for the data migration is run out. That is caused by a bug in the migration process, the connection should be returned to the pool instantly, not with `defer` way.

We’ll build a patch soon and check if we resolve this issue.

### DIFF
--- a/src/jobservice/migration/migrator_v180.go
+++ b/src/jobservice/migration/migrator_v180.go
@@ -18,6 +18,8 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"strconv"
+
 	"github.com/goharbor/harbor/src/jobservice/common/rds"
 	"github.com/goharbor/harbor/src/jobservice/common/utils"
 	"github.com/goharbor/harbor/src/jobservice/job"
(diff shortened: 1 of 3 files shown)


src/jobservice/migration/migrator_v180.go
```diff
@@ -250,6 +250,11 @@ func getScoreByID(id string, conn redis.Conn, ns string) (int64, error) {
 
 // Get periodic policy object by the numeric ID
 func getPeriodicPolicy(numericID int64, conn redis.Conn, ns string) (*period.Policy, error) {
+	// close this inner connection here
+	defer func() {
+		_ = conn.Close()
```
reasonerjt: not very relevant to the fix, but shall we log the error of closing conn?

steven-zou: No, the root cause is this conn is not closed.

Remove the maxConn in the pool creation is an enhanced fix 

for the error handling, I don't think we need because the reason I add _ is dismissing the best practice tips, no risk to the code.

reasonerjt: Thanks for fixing it anyway.

Issue #7982, opened 2019-06-05
Improve the performance of replication when fetching resources


ywk253100:
Currently, the resources fetching progress is in sequence, we can enhance it by fetching them in parallel.

ywk253100:
Done

PR #8514, opened 2019-07-31
Replication performance

cd1989:
Improve FetchImages performance for registry adapters

Issue: https://github.com/goharbor/harbor/issues/7982

### DIFF
--- a/src/replication/adapter/aliacr/adapter.go
+++ b/src/replication/adapter/aliacr/adapter.go
@@ -1,14 +1,12 @@
 package aliacr
 
 import (
-	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
 	"path/filepath"
 	"regexp"
-	"sync"
 
 	"github.com/aliyun/alibaba-cloud-sdk-go/services/cr"
 	"github.com/goharbor/harbor/src/common/utils"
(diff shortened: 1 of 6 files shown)


src/replication/adapter/native/adapter.go
```diff
@@ -159,37 +162,75 @@ func (a *Adapter) FetchImages(filters []*model.Filter) ([]*model.Resource, error
...
-		}
-		for _, filter := range filters {
-			if err = filter.DoFilter(&vTags); err != nil {
-				return nil, err
+	rawResources := make([]*model.Resource, len(repositories))
+	var wg = new(sync.WaitGroup)
+	ctx, cancel := context.WithCancel(context.Background())
+	var passportsPool = utils.NewPassportsPool(adp.MaxConcurrency, ctx.Done())
+
+	for i, r := range repositories {
```
renmaosheng: Could we write a function for this for_loop? it seems we have 3 sections are similar code.

src/replication/adapter/aliacr/adapter.go
```diff
@@ -157,42 +159,81 @@ func (a *adapter) FetchImages(filters []*model.Filter) (resources []*model.Resou
...
+				}
+			} else {
+				rawResources[index] = nil
+			}
+		}(i, r)
+	}
+	wg.Wait()
+
+	err = ctx.Err()
+	cancel()
```
steven-zou: seems this `cancel` can be deferred too?

src/replication/adapter/dockerhub/adapter.go
```diff
@@ -25,16 +28,16 @@ func init() {
...
-		Credential: registry.Credential,
-		Insecure:   registry.Insecure,
-	})
+	client, err := NewClient(registry)
 	if err != nil {
 		return nil, err
 	}
 
-	dockerRegistryAdapter, err := native.NewAdapter(registry)
+	dockerRegistryAdapter, err := native.NewAdapter(&model.Registry{
```
steven-zou: what's the reason for exchanging the passed-in parameters?

src/replication/adapter/native/adapter.go
steven-zou: +1,

Can we move those duplicated codes into a unified runner? Seems most of the code are similar.

cd1989: OK

src/replication/adapter/aliacr/adapter.go
```diff
@@ -157,42 +159,81 @@ func (a *adapter) FetchImages(filters []*model.Filter) (resources []*model.Resou
...
+						Vtags:  filterTags,
+						Labels: []string{},
 					},
-					Vtags:  filterTags,
-					Labels: []string{},
-				},
-			})
+				}
+			} else {
+				rawResources[index] = nil
```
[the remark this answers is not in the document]
cd1989: To avoid use lock for the `append` op.

src/replication/adapter/aliacr/adapter.go
cd1989: Yes

src/replication/adapter/dockerhub/adapter.go
cd1989: DockerHub use too different url for v2 api and others. For example, `hub.docker.io` to list namespace, delete repo, and `registry-1.docker.io` for v2 apis.

PR #8515, opened 2019-07-31
Typo fix in en-us-lang.json

ruzickap:
Typo fix in en-us-lang.json "Creat" -> "Create"


reasonerjt:
@ruzickap please fix the DCO error.

PR #8529, opened 2019-08-01
add disable rule feature for tag retention

bitsf:
Change-Id: Ic82b41c2ed14c47c51b9be8a46376bdfc7567376

### DIFF
--- a/src/pkg/retention/launcher.go
+++ b/src/pkg/retention/launcher.go
@@ -125,6 +125,10 @@ func (l *launcher) Launch(ply *policy.Metadata, executionID int64, isDryRun bool
 	}
 
 	for _, rule := range ply.Rules {
+		if rule.Disabled {
+			log.Infof("Policy %d rule %d %s is disabled", ply.ID, rule.ID, rule.Template)
+			continue
+		}
 		projectCandidates := allProjects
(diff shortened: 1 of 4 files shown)


src/pkg/retention/launcher.go
```diff
@@ -125,6 +125,10 @@ func (l *launcher) Launch(ply *policy.Metadata, executionID int64, isDryRun bool
 	}
 
 	for _, rule := range ply.Rules {
+		if rule.Disabled {
+			log.Debugf("policy %d rule %d %s is disabled", ply.ID, rule.ID, rule.Template)
```
renmaosheng: this operation should be explicit logged, not in debug log.

src/pkg/retention/launcher.go
```diff
@@ -125,6 +125,10 @@ func (l *launcher) Launch(ply *policy.Metadata, executionID int64, isDryRun bool
 	}
 
 	for _, rule := range ply.Rules {
+		if rule.Disabled {
+			log.Infof("policy %d rule %d %s is disabled", ply.ID, rule.ID, rule.Template)
```
steven-zou: - debug?
- Start with upper case word for the message?

PR #8536, opened 2019-08-01
add task retain num

bitsf:
Change-Id: Ic3696a924118b8ea0e3897ca4aafeba30fcc82c7

### DIFF
--- a/src/pkg/retention/job.go
+++ b/src/pkg/retention/job.go
@@ -16,6 +16,7 @@ package retention
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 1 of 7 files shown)


src/pkg/retention/job.go
```diff
@@ -110,6 +111,24 @@ func (pj *Job) Run(ctx job.Context, params job.Parameters) error {
 
 	// Log stage: results with table view
 	logResults(myLogger, allCandidates, results)
+	var delNum int
```
steven-zou: actually, those codes can be put into 1 local function.

bitsf: Actually I don't like to use one string to store two integer, and UI may have to decode and validate them.

PR #8546, opened 2019-08-02
nginx: Enable HTTP/2

JakubOnderka:


renmaosheng:
could you please share more background why you need this? thanks.

JakubOnderka:
Hi, this change is not needes, but HTTP/2 has many adavtanges, for example more efficient multiplexing and header compression. Docker Registry has HTTP/2 enabled by default. For all advatages see https://http2.akamai.com/

reasonerjt:
HTTP2 has its advantage but Harbor service is not based on http2.

Closing.

JakubOnderka:
@reasonerjt Sorry, but what do you mean that "Harbor service is not based on http2". Because this pull requests add the support, so clients can use HTTP/2 when communicating with Harbor.

Issue #8432, opened 2019-07-26
group role not honored for user group support in http auth mode - overwritten by user role


ericvmw:
**Expected behavior and actual behavior:**
Expected: 
if a user with guest role in a project belongs to a project member group that has master role, the user should be able to push images since its user group has the role
Actual:
if a user with guest role in a project belongs to a project member group that has master role, the project member group role is not honor and the user is not able to push images

**Steps to reproduce the problem:**
0. assign user A to user group B on the identity provider side
1. create a Harbor project
2. assign guest role to user A in the project
3. assign master role to group B in the project
4. docker login as user A, and try to tag and push an image to the project

**Versions:**
Please specify the versions of following systems.
- harbor version: [1.8.2]

**Configurations**
Harbor auth is configured with http_auth mode

reasonerjt:
The PRs fixed #8433 , reopening this issue

reasonerjt:
@ericvmw 
Currently the design is that if a user is added to a project as a user member, his role overwrites the the role of the group he is in.

I think what you expect is also reasonable but changing this will introduce some sort of break, I hope we can have some time to discuss and decide.

xaleeks:
> @ericvmw
> Currently the design is that if a user is added to a project as a user member, his role overwrites the the role of the group he is in.
> 
> I think what you expect is also reasonable but changing this will introduce some sort of break, I hope we can have some time to discuss and decide.

This is not consistent with the resolution of #8433 right? where we take the higher privilege of the two.  

I think the deciding question is can the user do anything to override the permissions set for him by someone of higher privilege in order to gain additional access to a project that was not originally intended for him

reasonerjt:
@xaleeks I think either the user cannot gain additional access.  But yes this does not look consistent with #8433 

Currently I'm thinking we can have it fixed in 1.8.2 and write a release note to point it out.

reasonerjt:
Fixed in #8557 and #8560 
Closing.

xaleeks:
https://github.com/goharbor/harbor/blob/master/docs/manage_role_by_ldap_group.md
hey Daniel, you already changed the last section there but maybe can provide some more details on how they are merged?

PR #8557, opened 2019-08-05
Merge user roles and group roles

stonezdj:
Fix issue #8432
If a user in the project member list and also this user's group is also in the project member. merge these privileges.

### DIFF
--- a/src/common/security/local/context.go
+++ b/src/common/security/local/context.go
@@ -127,10 +127,24 @@ func (s *SecurityContext) GetProjectRoles(projectIDOrName interface{}) []int {
 			roles = append(roles, common.RoleGuest)
 		}
 	}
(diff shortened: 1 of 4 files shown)


src/common/security/local/context.go
```diff
@@ -127,10 +127,24 @@ func (s *SecurityContext) GetProjectRoles(projectIDOrName interface{}) []int {
...
 	}
-	if len(roles) != 0 {
-		return roles
+	return mergeRoles(roles, s.GetRolesByGroup(projectIDOrName))
+}
+
+func mergeRoles(rolesA, rolesB []int) []int {
+	type void struct{}
+	var roles []int
+	var placeHolder void
```
reasonerjt: What is void?

PR #8562, opened 2019-08-05
retain nothing rule

bitsf:
Change-Id: I4e7a4ecb40fe39b80e41a6d9bf8b5fb3968a41af

### DIFF
--- a/src/pkg/retention/policy/rule/index/index_test.go
+++ b/src/pkg/retention/policy/rule/index/index_test.go
@@ -84,7 +84,7 @@ func (suite *IndexTestSuite) TestGet() {
 // TestIndex tests Index
 func (suite *IndexTestSuite) TestIndex() {
 	metas := Index()
-	require.Equal(suite.T(), 6, len(metas))
+	require.Equal(suite.T(), 7, len(metas))
 	assert.Condition(suite.T(), func() bool {
 		for _, m := range metas {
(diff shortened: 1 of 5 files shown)


src/pkg/retention/policy/rule/index/index_test.go
```diff
@@ -84,7 +84,7 @@ func (suite *IndexTestSuite) TestGet() {
 // TestIndex tests Index
 func (suite *IndexTestSuite) TestIndex() {
 	metas := Index()
-	require.Equal(suite.T(), 6, len(metas))
+	require.True(suite.T(), len(metas) > 6)
```
steven-zou: The case is too general

src/pkg/retention/policy/rule/nothing/evaluator.go
```diff
@@ -0,0 +1,42 @@
...
+const (
+	// TemplateID of the always retain rule
+	TemplateID = "nothing"
+)
+
+type evaluator struct{}
+
+// Process for the "always" Evaluator simply returns the input with no error
+func (e *evaluator) Process(artifacts []*res.Candidate) ([]*res.Candidate, error) {
+	return nil, nil
```
steven-zou: not `nil`, should be an empty array

or use the declared variable in the return statement

```
Process(artifacts []*res.Candidate) (processed []*res.Candidate, err error) {
    return processed
}
```

Issue #8547, opened 2019-08-02
Removal of leading white space prior to reaching out to registry


IAmATeaPot418:
If a user copy-pastes an image tag or image when pulling using a replication rule and leading white space is part of one of those names than a 404 error is thrown. While this is an obvious and expected response, it would be useful to help prevent user error by removing the leading white space.

Documenting in case someone else has this but I think I'll be able to submit a pull request for this.

**Expected behavior and actual behavior:**
A clear and concise description of what you expected to happen and what's the actual behavior. If applicable, add screenshots to help explain your problem.

When a leading space is set with a registry replication rule for the namespace or image tag of an image then the replication rule will fail with a 404 error. The expectation would be to remove this white space and still reach out to the correct image location. Instead this fails.

**Steps to reproduce the problem:**
Please provide the steps to reproduce this problem.

Create a replication rule and add a space in front of the tag or name in the source resource filter when trying to pull from a registry. This will result in a 404 error.

**Versions:**
Please specify the versions of following systems.
- harbor version: v1.8.2-rc1

AllForNothing:
@IAmATeaPot418 Will fix it in v1.9.0 , thanks for your comment.

PR #8574, opened 2019-08-06
Removal of leading white space

AllForNothing:
#8547


src/portal/lib/src/create-edit-rule/create-edit-rule.component.html
```diff
@@ -68,7 +68,7 @@ <h3 class="modal-title">{{headerTitle | translate}}</h3>
                 </div>
                 <label *ngIf="supportedFilters[i]?.style==='input'" aria-haspopup="true" role="tooltip" class="tooltip tooltip-validation tooltip-md tooltip-bottom-left"
                   [class.invalid]='(filter.value.dirty || filter.value.touched) && filter.value.invalid'>
-                  <input type="text" #filterValue required size="14" formControlName="value" id="{{'filter_'+ supportedFilters[i]?.type}}">
+                  <input (input)="trimText($event)" type="text" #filterValue required size="14" formControlName="value" id="{{'filter_'+ supportedFilters[i]?.type}}">
```
zhoumeina: Why need trim() here. not in the form.

AllForNothing: the form is dynamic .  it will be much more complicated to use trim in the form.

PR #8589, opened 2019-08-07
Update docker-compose checker

ninjadq:
Docker compose check function need updated to 1.23.0+


cadnce:
Actually, what I probably mean is can you also update this reference as it's now incorrect.

https://github.com/goharbor/harbor/blame/master/docs/compile_guide.md#L13

PR #8592, opened 2019-08-07
check rule conflict

bitsf:
Change-Id: I8af648f22aa3fe76024240d322a7fdae560e3fbd

### DIFF
--- a/src/core/api/retention.go
+++ b/src/core/api/retention.go
@@ -1,6 +1,8 @@
 package api
 
 import (
+	"crypto/sha256"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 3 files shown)


src/core/api/retention.go
```diff
@@ -250,6 +236,23 @@ func (r *RetentionAPI) UpdateRetention() {
 	}
 }
 
+func (r *RetentionAPI) checkRuleConflict(p *policy.Metadata) error {
+	temp := make(map[string]int)
+	for n, rule := range p.Rules {
+		tid := rule.ID
+		rule.ID = 0
+		bs, _ := json.Marshal(rule)
+		s := sha256.New()
```
steven-zou: removed

steven-zou: copy:=*r

PR #8596, opened 2019-08-07
nginx: Remove TLSv1.1 support

JakubOnderka:
According to [linked article](https://raymii.org/s/tutorials/Strong_SSL_Security_On_nginx.html) from comment, TLSv1.1 is not recommended anymore and should be disabled. 

And for example, [GitLab 12 has TLSv1.1 disabled by default](https://docs.gitlab.com/omnibus/update/gitlab_11_changes.html#tls-v11-deprecation).


reasonerjt:
Let's triage this PR for 1.10 as it may break some upgrade

xaleeks:
I think it's good practice, please get rid of this from recommended list @ninjadq

PR #8597, opened 2019-08-07
refactor(quota,middleware): implement size quota by quota interceptor

heww:
1. refactor size quota and implement it by quota interceptor
2. feat release size quota when delete image


src/common/dao/artifact.go
```diff
@@ -26,6 +26,9 @@ import (
 func AddArtifact(af *models.Artifact) (int64, error) {
 	now := time.Now()
 	af.CreationTime = now
```
ywk253100: The `CreationTime` is the time when the artifact is built, so it shouldn't be set here.

src/common/dao/artifact.go
```diff
@@ -26,6 +26,9 @@ import (
...
+	af.PushTime = now
+	af.PullTime = now
```
ywk253100: The `PushTime` and `PullTime` should be set at upper level. Besides, the `PullTime` isn't the time when the artifact record is created in database.

src/core/middlewares/util/util.go
```diff
@@ -111,22 +129,104 @@ type MfInfo struct {
...
+	manifestExist     bool
+	manifestExistErr  error
+	manifestExistOnce sync.Once
+
+	// artifact the artifact indexed by (repository, tag) in DB
+	artifact     *models.Artifact
+	artifactErr  error
+	artifactOnce sync.Once
+
+	ExclusiveBlobs []*models.Blob
```
wy65701436: // ExclusiveBlobs include the blobs that belong to the manifest only and exclude the blobs that shared by other manifests in the same repo(project/repository).

please gives an explanation about what is ExclusiveBlobs.

src/core/middlewares/util/util.go
```diff
@@ -111,22 +129,104 @@ type MfInfo struct {
...
+}
+
+// MutexKey returns mutex key of the manifest
+func (info *MfInfo) MutexKey(suffix ...string) string {
```
wy65701436: I remember that we need to lock tag for count usage on intercept the put manifest request?

src/common/dao/artifact.go
heww: It should be set here, the CreationTime is the time to insert data to db.

If you want the build time of the image, i think `BuildTime` need to be added like `PushTime` and `PullTime`

heww: The assignment for `PullTime` was removed, but `PushTime` was kept for the time to create artifact is the most time to push the image to registry.

src/common/dao/artifact.go
heww: The assignment for PullTime was removed, but PushTime was kept for the time to create artifact is the most time to push the image to registry.

src/common/dao/project_blob.go
```diff
@@ -0,0 +1,158 @@
...
+	if len(blobs) == 0 {
+		return 0, nil
+	}
+
+	var pb []*models.ProjectBlob
+	for _, blob := range blobs {
+		pb = append(pb, &models.ProjectBlob{ProjectID: projectID, BlobID: blob.ID})
+	}
+
+	return GetOrmer().InsertMulti(100, pb)
```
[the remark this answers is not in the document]
heww: Changed to the length of the project blobs now.

src/common/dao/project_blob.go
```diff
@@ -0,0 +1,158 @@
...
+	var blobs []*models.Blob
+	if _, err := GetOrmer().Raw(sql, params...).QueryRows(&blobs); err != nil {
+		return nil, err
+	}
+
+	return blobs, nil
+}
+
+// GetExclusiveBlobs returns layers of repository:tag which are not shared with other repositories in the project
+func GetExclusiveBlobs(projectID int64, repository, digest string) ([]*models.Blob, error) {
```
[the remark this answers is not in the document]
heww: Test cases added now.

src/core/middlewares/util/util.go
heww: Added

src/core/middlewares/util/util.go
heww: Fixed

src/common/dao/artifact.go
heww: The build time of the artifact problem can create an issue.

src/common/dao/blob.go
```diff
@@ -50,15 +64,73 @@ func DeleteBlob(digest string) error {
...
+	if _, err := GetOrmer().Raw(sql, artifactDigest).QueryRows(&blobs); err != nil {
+		return nil, err
+	}
+
+	return blobs, nil
+}
+
+// GetExclusiveBlobs returns layers of repository:tag which are not shared with other repositories in the project
+func GetExclusiveBlobs(projectID int64, repository, digest string) ([]*models.Blob, error) {
+	blobs, err := GetBlobsByArtifact(digest)
```
ywk253100: This returns the blobs of artifact which even isn't in the project specified by the projectID, is this the expected behavior?

PR #8605, opened 2019-08-08
Fix issue width Link to license in the about dialog should not be hardcoded to master

jwangyangls:
image
image


AllForNothing:
Please let Alex know about it, as the way to show LICENSE is changed.

Issue #8202, opened 2019-07-02
Replication tasks stuck in "InProgress" status


YakirShriker:
I have some replication executions that stuck on InProgress mode
I am not able to stop them, getting Internal server error
what can I do more in order to remove the executions or mark them as Finished?
Log for execution: {"code":404,"message":"the log of task 90050 not found"}

ywk253100:
What is the version of Harbor that you are using?

YakirShriker:
1.8

ywk253100:
Could you provide more information about that? Some screenshots may be helpful. And upload all logs.

YakirShriker:
Hi! thanks for replying.
Example of one execution:
**Proxy log**
Jul  4 12:28:11 172.18.0.1 proxy[30933]: 192.168.14.196 - "GET /api/replication/executions/90198 HTTP/1.1" 200 191 "https://ctvr-dcr01.
int.liveperson.net/harbor/replications/90198/tasks" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_4) AppleWebKit/537.36 (KHTML, like Ge
cko) Chrome/73.0.3683.75 Safari/537.36" 0.005 0.005 .
Jul  4 12:28:11 172.18.0.1 proxy[30933]: 192.168.14.196 - "GET /api/replication/executions/90198/tasks HTTP/1.1" 200 210 "https://ctvr-dcr01.int.liveperson.net/harbor/replications/90198/tasks" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/73.0.3683.75 Safari/537.36" 0.004 0.004 .

**Core log**
Jul  4 12:29:01 172.18.0.1 core[30933]: 2019/07/04 09:29:01 #033[1;44m[D] [server.go:2619] | 192.168.14.196|#033[42m 200 #033[0m|   2.653121ms|   match|#033[44m GET     #033[0m /api/replication/executions/90198   r:/api/replication/executions/:id([0-9]+)#033[0m
Jul  4 12:29:01 172.18.0.1 core[30933]: 2019/07/04 09:29:01 #033[1;44m[D] [server.go:2619] | 192.168.14.196|#033[42m 200 #033[0m|   3.441975ms|   match|#033[44m GET     #033[0m /api/replication/executions/90198/tasks   r:/api/replication/executions/:id([0-9]+)/tasks#033[0m
[...]

PR #8606, opened 2019-08-08
Fix replication tasks stuck in "InProgress" issue

ywk253100:
Fix replication tasks stuck in "InProgress" issue


ywk253100:
Fixes #8202

ywk253100:
@steven-zou Using status code to do the comparation involves works of migration and there are some other tasks(scan/GC) need to do the same work, we can finish that in the future release. And I have opened the issue #8644 to track the work.

Issue #8570, opened 2019-08-05
project are not sorted by name in UI like they used to be.


tkatkov:
**Expected behavior and actual behavior:**
Projects are not sorted in UI when there are no LDAP groups in play.

**Steps to reproduce the problem:**
In this function 
https://github.com/goharbor/harbor/blob/master/src/common/dao/project.go#L159
order by name never added if there are no groups

- harbor version: [1.8.1]
* **Log files:** 
Aug  5 19:26:36 172.18.0.1 core[9700]: 2019-08-05T19:26:36Z [DEBUG] [/common/dao/project.go:173]: query sql:select distinct p.project_id, p.name, p.owner_id,
Aug  5 19:26:36 172.18.0.1 core[9700]: #011#011#011#011p.creation_time, p.update_time  from project as p where p.deleted=false

stonezdj:
Fixed

PR #8621, opened 2019-08-09
Sort project by name

stonezdj:
Fix issue #8570, sort projects when there is no user group related info


reasonerjt:
The name `func GetGroupProjects` is a little misleading

Issue #8564, opened 2019-08-05
Excessive refresh requests in History tab of GC UI


reasonerjt:
![image](https://user-images.githubusercontent.com/2390463/62479427-0b339d00-b7e0-11e9-991f-35c54ab72e67.png)

In this page above, I found the UI is refreshing every second.

This is not necessary, I don't see why we need different refresh strategy than repository page.
Let's review why such decision is made, and fix it to make things consistent with other parts of the UI.

jwangyangls:
fix it in #8623

PR #8623, opened 2019-08-09
Add paging from backend in replication task ui

jwangyangls:
fix #8623 #8564
image


src/portal/lib/src/config/gc/gc-history/gc-history.component.ts
```diff
@@ -34,7 +34,7 @@ export class GcHistoryComponent implements OnInit, OnDestroy {
       this.loading = false;
       // to avoid some jobs not finished.
       if (!this.timerDelay) {
-        this.timerDelay = timer(3000, 3000).subscribe(() => {
+        this.timerDelay = timer(10000, 10000).subscribe(() => {
```
zhoumeina: DEFINE A CONST

jwangyangls: done

PR #8631, opened 2019-08-11
Add chart and scanning event for webhook

mmpei:


src/chartserver/reverse_proxy.go
```diff
@@ -110,8 +112,44 @@ func modifyResponse(res *http.Response) error {
...
+				metaData := &n_event.ChartUploadMetaData{
+					ChartMetaData: n_event.ChartMetaData{
+						ProjectName: e.Resource.ExtendedInfo["projectName"].(string),
+						ChartName:   e.Resource.ExtendedInfo["chartName"].(string),
+						Versions:    e.Resource.Metadata.Vtags,
+						OccurAt:     time.Now(),
+						Operator:    e.Resource.ExtendedInfo["operator"].(string),
+					},
+				}
+				if err := event.Build(metaData); err != nil {
```
steven-zou: Publish should be called only and only if `event.Build` returns no error.

Check other similar codes?

src/core/api/chart_repository.go
```diff
@@ -555,3 +622,24 @@ func chartFullName(namespace, chartName, version string) string {
 	}
 	return fmt.Sprintf("%s/%s:%s", namespace, chartName, version)
 }
+
+// parseChartVersionFromFilename parse chart and version from file name
+func parseChartVersionFromFilename(filename string) (string, string) {
```
wy65701436: @steven-zou please confirm whether we can use this func to get version and name?

src/core/api/chart_repository.go
```diff
@@ -355,13 +384,32 @@ func (cra *ChartRepositoryAPI) DeleteChart() {
...
+			ChartName:   chartName,
+			Versions:    versions,
+			OccurAt:     time.Now(),
+			Operator:    cra.SecurityCtx.GetUsername(),
+		},
+	}
+	if err := event.Build(metaData); err != nil {
+		hlog.Errorf("failed to build chart delete event metadata: %v", err)
+	}
+	if err := event.Publish(); err != nil {
```
steven-zou: delete chart event should be published after `chartController.DeleteChart`?

src/core/notifier/handler/notification/chart_handler.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+// ChartPreprocessHandler preprocess chart event data
+type ChartPreprocessHandler struct {
+}
+
+// Handle preprocess chart event data and then publish hook event
+func (cph *ChartPreprocessHandler) Handle(value interface{}) error {
+	// if global notification configured disabled, return directly
+	if !config.NotificationEnable() {
```
wy65701436: if !config.WithChartMuseum() {
  return nil
}

src/core/notifier/handler/notification/processor.go
```diff
@@ -172,3 +173,30 @@ func preprocessAndSendImageHook(value interface{}) error {
...
+
+// will return nil when it failed to get data
+func getScanOverview(digest string, tag string, eventType string) *models.ImgScanOverview {
+	if len(digest) == 0 {
+		log.Debug("digest is nil")
+		return nil
+	}
+	data, err := dao.GetImgScanOverview(digest)
+	if err != nil {
+		log.Errorf("Failed to get scan result for tag:%s, digest: %s, error: %v", tag, digest, err)
```
wy65701436: return err?

src/core/notifier/handler/notification/scan_image_handler.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// Handle preprocess chart event data and then publish hook event
+func (si *ScanImagePreprocessHandler) Handle(value interface{}) error {
+	// if global notification configured disabled, return directly
+	if !config.NotificationEnable() {
+		log.Debug("notification feature is not enabled")
+		return nil
+	}
+
```
wy65701436: if !config.WithClair(){
  return nil
}

src/core/notifier/handler/notification/scan_image_handler.go
```diff
@@ -0,0 +1,127 @@
...
+	rs := strings.SplitN(job.Repository, "/", 2)
+	projectName := rs[0]
+	repoName := rs[1]
+
+	project, err := config.GlobalProjectMgr.Get(projectName)
+	if err != nil {
+		log.Errorf("failed to find project[%s] for scan image event: %v", projectName, err)
+		return err
+	}
```
wy65701436: if project == nil {
  ....
}

src/core/api/chart_repository.go
mmpei: you are right. but it will be triggered regardless of the result of DeleteChart for the action is not atomic and some charts may have been deleted even error returned.

src/core/notifier/handler/notification/scan_image_handler.go
mmpei: the logic will go through this judge statement before triggering this event.

src/core/notifier/handler/notification/processor.go
mmpei: i pick the first one between missing some part of payload and dropping the triggered event.

src/core/notifier/handler/notification/chart_handler.go
mmpei: i think it has been judged when it gets here.

src/chartserver/reverse_proxy.go
mmpei: ok i will check.

src/core/notifier/handler/notification/scan_image_handler.go
mmpei: ok. i will check.

src/chartserver/reverse_proxy.go
```diff
@@ -110,8 +112,45 @@ func modifyResponse(res *http.Response) error {
...
```
reasonerjt: This is still incorrect.  Forgot to fix?

mmpei: sorry. ide shows a wrong info when i search the invoking of Build function. i will check carefully again.

PR #8633, opened 2019-08-11
feat(quota,middleware): enable or disable quota per project by config

heww:
1. add quota_per_project_enable  config item
2. quota interceptor reserve resources only enforce resources set for the interceptor
3. quota for project creation and summary have effect only quota_per_project_enable is true

### DIFF
--- a/docs/swagger.yaml
+++ b/docs/swagger.yaml
@@ -4841,6 +4841,9 @@ definitions:
       project_creation_restriction:
         type: string
         description: This attribute restricts what users have the permission to create project.  It can be "everyone" or "adminonly".
+      quota_per_project_enable:
+        type: boolean
+        description: This attribute indicates whether quota per project enabled in harbor
       read_only:
         type: boolean
         description: '''docker push'' is prohibited by Harbor if you set it to true.   '
@@ -4938,6 +4941,9 @@ definitions:
       project_creation_restriction:
         $ref: '#/definitions/StringConfigItem'
         description: This attribute restricts what users have the permission to create project.  It can be "everyone" or "adminonly".
+      quota_per_project_enable:
+        $ref: '#/definitions/BoolConfigItem'
+        description: This attribute indicates whether quota per project enabled in harbor
       read_only:
         $ref: '#/definitions/BoolConfigItem'
         description: '''docker push'' is prohibited by Harbor if you set it to true.   '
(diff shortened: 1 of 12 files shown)


src/core/api/project.go
```diff
@@ -212,14 +215,16 @@ func (p *ProjectAPI) Post() {
...
-	if _, err := quotaMgr.NewQuota(hardLimits); err != nil {
-		p.SendInternalServerError(fmt.Errorf("failed to create quota for project: %v", err))
-		return
+	if config.QuotaPerProjectEnable() {
+		quotaMgr, err := quota.NewManager("project", strconv.FormatInt(projectID, 10))
+		if err != nil {
+			p.SendInternalServerError(fmt.Errorf("failed to get quota manager: %v", err))
+			return
+		}
+		if _, err := quotaMgr.NewQuota(hardLimits); err != nil {
```
ywk253100: If we don't add the quota and usage record into the database for the project, does the quota flow work well after we enabling the quota again?

heww: It will work well because quota manager will create the quota and quota usage for the project using default quota settings when they are not in db.

wy65701436: for the scenario, disabel -> enable, it will call the sync quota to sync all of the usages.

@heww, I will remove the quota and usage adding from the DB scheme script and leave the library only. All of the others(projects) will be handled by the sync api. 

You can see the sync PR for details.

src/core/middlewares/interceptor/quota/quota.go
```diff
@@ -113,6 +104,23 @@ func (qi *quotaInterceptor) HandleResponse(w http.ResponseWriter, req *http.Requ
...
+			return err
+		}
+		qi.mutexes = append(qi.mutexes, m)
+	}
+
+	return nil
+}
+
 func (qi *quotaInterceptor) freeMutexes() {
 	for i := len(qi.mutexes) - 1; i >= 0; i-- {
```
wy65701436: if !qi.opts.EnforceResources() {
		......
}

as it's in the defer call of HandleRequest()

docs/swagger.yaml
```diff
@@ -4841,6 +4841,9 @@ definitions:
       project_creation_restriction:
         type: string
         description: This attribute restricts what users have the permission to create project.  It can be "everyone" or "adminonly".
+      quota_per_project_enable:
```
wy65701436: I'd prefer that we use a separate API to control the disable and enable quota, and we can add logic in the scenario from disable to enable.

But, it can be stored in the table of properties.

PR #8637, opened 2019-08-12
Add ensure quota and usage after execute GC

wy65701436:

### DIFF
--- a/src/jobservice/job/impl/gc/job.go
+++ b/src/jobservice/job/impl/gc/job.go
@@ -22,10 +22,13 @@ import (
 	"github.com/garyburd/redigo/redis"
 	"github.com/goharbor/harbor/src/common"
 	"github.com/goharbor/harbor/src/common/config"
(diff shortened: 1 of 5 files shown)


src/jobservice/job/impl/gc/job.go
```diff
@@ -88,6 +91,9 @@ func (gc *GarbageCollector) Run(ctx job.Context, params job.Parameters) error {
 	if err := gc.cleanCache(); err != nil {
 		return err
 	}
+	if err := ensureQuota(); err != nil {
```
ywk253100: Why we re-calculate the storage usage in the garbage collection job? The GC job doesn't change the database records about the blobs at all, right?

wy65701436: As harbor returns the quota at the time when the user deletes the artifact, and harbor will remove the corresponding item in table project_blob. In other words, we can believe that the project_blob reflect reality.

Here just add one insurance step after GC to ensure the correct usage.

PR #8645, opened 2019-08-13
Enable usage sync when switch quota setting

wy65701436:

### DIFF
--- a/src/core/api/internal.go
+++ b/src/core/api/internal.go
@@ -15,12 +15,19 @@
 package api
 
 import (
-	"errors"
-
+	"fmt"
+	"github.com/goharbor/harbor/src/chartserver"
 	"github.com/goharbor/harbor/src/common"
 	"github.com/goharbor/harbor/src/common/dao"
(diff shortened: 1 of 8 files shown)


src/core/api/internal.go
```diff
@@ -69,3 +76,96 @@ func (ia *InternalAPI) RenameAdmin() {
...
+		}
+		pCount := int64(len(afs))
+
+		// it needs to append the chart count
+		if config.GetCfgManager().Get(common.WithChartMuseum).GetBool() {
+			chartEndpoint := strings.TrimSpace(config.GetCfgManager().Get(common.ChartRepoURL).GetString())
+			if len(chartEndpoint) == 0 {
+				return errors.New("empty chartmuseum endpoint")
+			}
+			chartCtr, err := getChartCtr(chartEndpoint)
```
ywk253100: There is a chart controller variable already in the package `api`: https://github.com/goharbor/harbor/blob/master/src/core/api/base.go#L147, you can use it directly

PR #8648, opened 2019-08-13
Normalize LDAP filter for user filter and group filter

### LINKED ISSUE #7710: LDAP User/Group filter syntax inconsistency
Version: 1.7.5 and master

There is inconsistency in configuring LDAP User and Group Filter in Configuration/Authentication or in harbor.cfg.
While the User Filter must contain round brackets at the begging and end - `(objectClass=user)`; the Group Filter can't contain brackets at the beggining and end - `objectClass=group`.

From: https://github.com/goharbor/harbor/blob/master/src/common/utils/ldap/ldap.go
User Filter constraction code:
`ldapFilter = "(&" + ldapFilter + "(" + ldapUID + "=" + filterTag + "))"`
Group Filter constraction code:
`filter = "(&(" + oldFilter + ")(" + groupNameAttribute + "=*" + groupName + "*))"`

There should be unified way how to configure LDAP filters. I am not doing PR since I will let someone from maintainers to decide what way will be used as this change will break existing configurations.

stonezdj:
Fix #7710   Normalize LDAP filter for user and group filter to make a common schema for filter

### DIFF
--- a/src/common/utils/ldap/ldap.go
+++ b/src/common/utils/ldap/ldap.go
@@ -351,13 +351,13 @@ func (session *Session) createUserFilter(username string) string {
 		filterTag = goldap.EscapeFilter(username)
 	}
 
-	ldapFilter := session.ldapConfig.LdapFilter
+	ldapFilter := normalizeFilter(session.ldapConfig.LdapFilter)
 	ldapUID := session.ldapConfig.LdapUID
 
 	if ldapFilter == "" {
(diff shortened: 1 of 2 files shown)


reasonerjt:
Do we need to update the doc?

stonezdj:
> Do we need to update the doc?

Both (xxx=???) and xxx=??? will be OK, there is no document need to change.

reasonerjt:
@stonezdj please resolve the conflict

PR #8652, opened 2019-08-13
Optimize cicd

bitsf:
optimized CICD build process, use precompiled base image, move some static files from google storage to own server, fix cert creation and so on


reasonerjt:
@bitsf 
The pre-compiled image may not have latest security fixes.  The `photon:2.0` base image on docker hub is maintained by a team continuously.
May I suggest that we revisit the process after branching 1.9 and figure out what can be improved?

reasonerjt:
@bitsf 
Let me mark this PR as change requested to block accidental merge for now.

Issue #8658, opened 2019-08-14
dockerhub replication always failed


bitsf:
Recently dockerhub Replication adapter can't be used, always report "UNAUTHORIZED"

![image](https://user-images.githubusercontent.com/188115/63001208-acd48180-bea5-11e9-9874-4f17e5860553.png)

branch:master
commit:81f4b64b45837a9bddc7aa7d6c5ce9c5ed55d820

ywk253100:
Fixed

PR #8661, opened 2019-08-14
Fix bug in native replication adapter

ywk253100:
Use the stardard token authentication work flow even though the credentail isn't provided


ywk253100:
Fixed #8658

Issue #7848, opened 2019-05-21
Security gap with generated email


Reamer:
Hi,
I'm a new harbor user and installed harbor 1.8 today.
I have a technical LDAP-User, because at the moment it's not possible to login to the same registry to multiple projects via robot-accounts.
My technical-User (jenkinssu) has no email address in my ActiveDirectory.
I was shocked, that email address will be generated to `[username]@placeholder.com`. In my case `jenkinssu@placeholder.com`
I think, i found the code lines, which produce this: https://github.com/goharbor/harbor/blob/v1.8.0/src/core/auth/authproxy/auth.go#L124-L128

If harbor sends emails to project members, `placeholder.com` gets a email with potential security sensitive and company internal information.

Harbor should be able to save users without an email.

Best Regards
Reamer

wy65701436:
@Reamer  1) For the robot account, from the security perspective, Harbor doesn't store the password(jwt token) of the account, that means, the token cannot be modified once it created. In another word, the robot account cannot be added to another project. 
2)For the LDAP auth, Harbor sets the email to `@placeholder.com` if the email is empty, refer to https://github.com/goharbor/harbor/blob/release-1.8.0/src/core/auth/ldap/ldap.go#L122, it could be discussed if you have a security concern.

Reamer:
Hi @wy65701436,
the robot account shouldn't be the topic in this issue. I mentioned this only for understanding, why I use a technical LDAP-User.

I'm not the owner of the `placeholder.com` domain. If Harbor sends to project members (also to technical server user with a `@placeholder.com` email address) notification or report emails, we have a information leak. An email is send to an email server, which shouldn't not receive this information.

Reamer:
@stonezdj & @wy65701436 
What are you thoughts of this security gap?

reasonerjt:
@Reamer Agreed that the `placeholder.com` is an issue, we should probably Allow leaving the Email as empty.

Reamer:
To correct wrong users, manual correction directly on database is needed.
Use this update-query: https://github.com/goharbor/harbor/pull/8662/files#diff-f67e7dbf48c78752ee91ecd40cb9f581R240

PR #8662, opened 2019-08-14
Set default email to null if not provided

stonezdj:
Fix issue #7848, If email address is empty, set null instead of xxx@placeholder.com to avoid any security concern.

### DIFF
--- a/src/common/dao/user.go
+++ b/src/common/dao/user.go
@@ -234,6 +234,14 @@ func OnBoardUser(u *models.User) error {
 	}
 	if created {
 		u.UserID = int(id)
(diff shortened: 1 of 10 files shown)


src/common/dao/user.go
```diff
@@ -234,6 +234,13 @@ func OnBoardUser(u *models.User) error {
 	}
 	if created {
 		u.UserID = int(id)
+		// current orm framework doesn't support insert null value
```
reasonerjt: Can we open an issue on beego's repo, and reference that issue?

stonezdj: Updated

PR #8663, opened 2019-08-14
Fix quota switch fail to get project size

wy65701436:

### DIFF
--- a/src/core/api/internal.go
+++ b/src/core/api/internal.go
@@ -76,7 +76,7 @@ func (ia *InternalAPI) RenameAdmin() {
 
 // QuotaSwitcher ...
 type QuotaSwitcher struct {
-	Disabled bool
+	Enabled bool
 }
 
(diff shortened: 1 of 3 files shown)


src/core/api/internal.go
```diff
@@ -86,14 +86,17 @@ func (ia *InternalAPI) SwitchQuota() {
 		ia.SendBadRequestError(err)
 		return
 	}
-	// From disable to enable, it needs to update the quota usage bases on the DB records.
-	if config.QuotaPerProjectEnable() == false && req.Disabled == true {
+	// quota per project from disable to enable, it needs to update the quota usage bases on the DB records.
+	if config.QuotaPerProjectEnable() == false && req.Enabled == true {
```
heww: if !config.QuotaPerProjectEnable() && req.Enabled {
   ....
}

wy65701436: sure, updated.

Issue #234, opened 2016-05-16
User password management & length


RRAlex:
There doesn't seem to be an interface to change the user's password as admin and if you delete the user to recreate it, system will say user exists!
This breaks user management when passwords need to be changed. :(

Secondly, the passwords seem to be limited to a short 20 characters...
Any reason why it couldn't be say 63 or 255?

cheers!

reasonerjt:
@RRAlex 
Currently for audit purpose, when admin delete a user, the user is just "soft deleted", i.e. user with the same username can not be created again.

So you want to enable admin to update a user's password so the username can be re-used?

RRAlex:
So deleting a user would also delete its activity log?

But yes, indeed, simply being able to update a user's password would be sufficient.
This would allow anyone to re-purpose an account: be it to rotate admins creds., correct a mistake, respect password change policy, etc.
And it's much more practical then going into MySQL to delete it manually and re-create / re-attribute the role everywhere, which is management hell and, well, bad UX.

thanks! :-)

reasonerjt:
> So deleting a user would also delete its activity log?

Currently activitly log is associate to user table via a Foreign Key, I know this is debatable, but we may use the term to "Disable" a user, not Delete.

As for enabling admin to update password, the original idea is no one should know any other's password including admin, but it makes sense to me when you mention password change policy, we'll consider implement it.

fuhbar:
@reasonerjt What about the second question @RRAlex raised, the password length? It seems to be restricted to max 20 chars. Is this on purpose? And if so, why?

bergtwvd:
The ability for the admin to reset a user password is **essential** in at least the following situation:
- no LDAP
- no mail server
- user has forgotten his password

But also in the following situation:
- no LDAP
- mail server
- user has changed email address, and forgotten his password

tronicum:
Yeah I would love if we could reset passwords via the web ui..., too

bengadbois:
It would be nice to have the length restriction presented to the user when setting the password

phenomax:
@reasonerjt What about the rather restrictive password length?

PR #8666, opened 2019-08-14
feat: Expand password length to unlimited

wilmardo:
Addresses: #234 

Extends the password length to unlimited characters.

This should be possible since:
* In [`src/common/dao/user.go`](https://github.com/goharbor/harbor/blob/master/src/common/dao/user.go#L159) the Salt never exceeds 32 characters since the [`utils.GenerateRandomString()`](https://github.com/goharbor/harbor/tree/master/src/common/utils/utils.go#L67) always return 32.
* [`utils.Encrypt(u.Password, u.Salt)`](https://github.com/goharbor/harbor/blob/master/src/common/utils/encrypt.go#L42) also always returns 32. that can be test here:
https://play.golang.org/p/NeBuv9eHm91

The database table for the password and the salt both have a `character varying(40)` type.

I updated the translations as best as I could, I trusted Google translate for most of them.

### DIFF
--- a/src/portal/src/app/account/password-setting/password-setting.component.html
+++ b/src/portal/src/app/account/password-setting/password-setting.component.html
@@ -13,7 +13,7 @@
             </clr-input-container>
             <clr-input-container>
                 <label class="required">{{'CHANGE_PWD.NEW_PWD' | translate}}</label>
-                <input clrInput type="password" id="newPassword" required pattern="^(?=.*\d)(?=.*[a-z])(?=.*[A-Z])(?!.*\s).{8,20}$" name="newPassword"
+                <input clrInput type="password" id="newPassword" required pattern="^(?=.*\d)(?=.*[a-z])(?=.*[A-Z])(?!.*\s).{8,128}$" name="newPassword"
                     [(ngModel)]="newPwd" #newPassInput="ngModel" size="40" (input)='handleValidation("newPassword", false)' (blur)='handleValidation("newPassword", true)'>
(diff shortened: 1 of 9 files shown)


wilmardo:
@ninjadq PTAL

This issue is blocking our migration to Harbor, we are using passwords up to 35 characters at the moment. Without this change a drop-in replacement is not possible. Would be nice to have it one of the upcoming releases 👍

Merlingilb:
Would be nice to include this in an upcoming release... This is stopping us from migrating to Harbor aswell

jomeier:
@ninjadq
@jwangyangls
For us this pull request is one of the most important Harbor Features. Our security officer was laughing out loud as he saw this limitation in password length. We use lots of Open Source Tools in our CI/CD chains and no one has a limitation like this.

Approve this pull request and make Harbor more secure, please.

wilmardo:
@ninjadq @jwangyangls How is the review coming along? Please let me know if any issues arise!

wilmardo:
Rebased to current master to resolve conflicts :)

xaleeks:
@ninjadq please check this on reasonable upper limit

ninjadq:
I agree that 20 is not enough. But unlimited is meaningless and may cause other issues. @wilmardo could you change the upper boundary to a concrete number like `128`.

wilmardo:
@ninjadq @jwangyangls Ready for review 🎉 Implemented the 128 upper limit :)

PR #8669, opened 2019-08-14
[Do not merge] fix(quota,replication): get blob size from body for replication

heww:
1. Contains Content-Length when push blob for replication.
2. Get blob size from body for replication of harbor 1.8 and below.

### DIFF
--- a/src/common/utils/registry/repository.go
+++ b/src/common/utils/registry/repository.go
@@ -25,7 +25,6 @@ import (
 	"sort"
 	"strconv"
 	"strings"
-	//	"time"
 
 	"github.com/docker/distribution/manifest/schema1"
 	"github.com/docker/distribution/manifest/schema2"
@@ -407,6 +406,9 @@ func (r *Repository) monolithicBlobUpload(location, digest string, size int64, d
 	if err != nil {
 		return err
 	}
+
+	// set the Content-Length for the request
+	req.ContentLength = size
 
 	resp, err := r.client.Do(req)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


src/common/utils/registry/repository.go
```diff
@@ -408,6 +407,9 @@ func (r *Repository) monolithicBlobUpload(location, digest string, size int64, d
 		return err
 	}
 
+	// set the Content-Length for the request
+	req.ContentLength = size
```
wy65701436: this works, thanks.

src/core/middlewares/sizequota/util.go
```diff
@@ -98,7 +101,27 @@ func parseBlobSize(req *http.Request, uuid string) (int64, error) {
...
-	return getUploadedBlobSize(uuid)
+	size, err = getUploadedBlobSize(uuid)
+	if err == nil && size != 0 {
+		return size, nil
+	}
+
+	// Get blob size for replication of harbor 1.8 and below
+	// which don't contain Content-Length header when do PUT request to Blob Upload URL
+	if req.Body != nil {
+		body, err := ioutil.ReadAll(req.Body)
```
wy65701436: Let's put this into the discussion as it copies the data into memory.

reasonerjt:
@wy65701436 I believe this is a must-fix for 1.9?  Could you clarify?

wy65701436:
@reasonerjt it's already fixed in v1.9, and this PR tries to enable replication between harbor 1.9 with the previous version, but it has to be discussed.

PR #8713, opened 2019-08-16
Avoid overwriting system CVE whitelist by mistake

### LINKED ISSUE #8702: System level CVE whitelist disappear after create another new project
1. Add a CVE whitelist to system level.
2. Create another project named sample3
3. Check the system CVE whitelist, found that system level CVE whitelist is empty.

reasonerjt:
Fixes #8702
Also enforce the code to mitigate the potential risk.

### DIFF
--- a/src/common/dao/cve_whitelist.go
+++ b/src/common/dao/cve_whitelist.go
@@ -21,6 +21,14 @@ import (
 	"github.com/goharbor/harbor/src/common/utils/log"
 )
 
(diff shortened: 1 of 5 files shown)


reasonerjt:
@stonezdj 
It should only be retrieved while pulling the manifest.

And I don't worry much about the length of the CVE list because it's done in one query from DB.
We'll keep an eye on performance impact

Issue #8690, opened 2019-08-15
[tag retention] corn for every 5 seconds doesn't work


wy65701436:
![image](https://user-images.githubusercontent.com/2841473/63078302-a6114180-bf6d-11e9-90b6-1cb5e03f474a.png)

ywk253100:
Fixed

Issue #8694, opened 2019-08-15
Tag Retention modify cron Internet Error 500


jwangyangls:
![image](https://user-images.githubusercontent.com/29826789/63079376-c2fb4400-bf70-11e9-9d23-f96845723aef.png)
![image](https://user-images.githubusercontent.com/29826789/63079472-fb9b1d80-bf70-11e9-8dbf-a78a3dcb6275.png)

ywk253100:
Cannot reproduce it, but we can do some enhancement to skip the error and continue the work flow if the schedule record not found

ywk253100:
Done

PR #8714, opened 2019-08-16
Fix bug found in scheduler

ywk253100:
The scheduler hook handler doesn't parse the job status struct when handling the hook. This commit fixes it.


ywk253100:
Fixes #8690 and #8694

PR #8715, opened 2019-08-16
fix tag retention schedule none error

bitsf:
Change-Id: I04cad9d4c520db751bfa413bb139317563716501


src/pkg/retention/controller.go
```diff
@@ -92,7 +92,7 @@ func (r *DefaultAPIController) GetRetention(id int64) (*policy.Metadata, error)
 func (r *DefaultAPIController) CreateRetention(p *policy.Metadata) (int64, error) {
 	if p.Trigger.Kind == policy.TriggerKindSchedule {
 		cron, ok := p.Trigger.Settings[policy.TriggerSettingsCron]
-		if ok {
+		if ok && len(p.Trigger.Settings[policy.TriggerSettingsCron].(string)) > 0 {
```
steven-zou: Please use local var to replace such a long expression,

```
// p.Trigger.Settings[policy.TriggerSettingsCron]
cron.(string)
```

PR #8725, opened 2019-08-18
fix(quota): fix computeResources method of qutoa interceptor

### LINKED ISSUE #8701: artifact count is not updated after deleting a chart
1. Upload a chart a project via UI
2. On project summary page the count increased by 1
3. Delete the chart via UI
4. On project summary page the artifact size remain unchanged.

heww:
Fix #8701

### DIFF
--- a/src/core/middlewares/chart/builder.go
+++ b/src/core/middlewares/chart/builder.go
@@ -103,20 +103,23 @@ func (*chartVersionCreationBuilder) Build(req *http.Request) (interceptor.Interc
 		return nil, fmt.Errorf("project %s not found", namespace)
 	}
 
-	chart, err := parseChart(req)
-	if err != nil {
-		return nil, fmt.Errorf("failed to parse chart from body, error: %v", err)
-	}
-	chartName, version := chart.Metadata.Name, chart.Metadata.Version
+	info, ok := util.ChartVersionInfoFromContext(req.Context())
+	if !ok {
+		chart, err := parseChart(req)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse chart from body, error: %v", err)
+		}
+		chartName, version := chart.Metadata.Name, chart.Metadata.Version
 
-	info := &util.ChartVersionInfo{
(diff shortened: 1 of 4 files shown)


src/core/middlewares/chart/builder.go
```diff
@@ -103,20 +103,23 @@ func (*chartVersionCreationBuilder) Build(req *http.Request) (interceptor.Interc
...
-		return nil, fmt.Errorf("failed to parse chart from body, error: %v", err)
-	}
-	chartName, version := chart.Metadata.Name, chart.Metadata.Version
-
-	info := &util.ChartVersionInfo{
-		ProjectID: project.ProjectID,
-		Namespace: namespace,
-		ChartName: chartName,
-		Version:   version,
+	info, ok := util.ChartVersionInfoFromContext(req.Context())
```
ywk253100: Not quite understand why we need to update the `chartVersionCreationBuilder ` instead of  `chartVersionDeletionBuilder ` as the issue #8701 is caused when deleting a chart

src/core/middlewares/interceptor/quota/quota.go
```diff
@@ -135,7 +135,8 @@ func (qi *quotaInterceptor) computeResources(req *http.Request) error {
 		return nil
 	}
 
-	if len(qi.opts.Resources) == 0 && qi.opts.OnResources != nil {
+	qi.resources = qi.opts.Resources
```
wy65701436: what's the difference?

wy65701436: delete chart is using opt.Resources(types.ResourceList{types.ResourceCount: 1}, but it still cannot satisfy the condition?

heww: quotaInterceptor reserve and unreserve will process the resources saved in quotaInterceptor struct.

This change ensure that resources will first get from the Resources of opts, if it is empty, then try to compute resources.

wy65701436: ok，got it.

PR #8726, opened 2019-08-19
refactor(quota): new error types for quota checking

### LINKED ISSUE #8693: There should be more details in the log message when the quota is exceeded
First, when the size quota is exceeded in the middle of pushing a image, docker will keep retrying.  This is a little confusing to the end user, there is another issue to track this problem

At that time, if I check the log, I see message like this:
```
Aug 15 00:16:49 172.27.0.1 core[5099]: 2019-08-15T07:16:49Z [ERROR] [/core/middlewares/interceptor/quota/quota.go:64]: Failed to add resources, error: over the quota: used 1209051826 but only hard 1073741824
```

It should have more details to help user understand the situation, for example which resource is failed to be added.  The url of the resource (artifact) so user understand this error happens when which artifact is being pushed.

heww:
More info message when quota usage overflow the hard limit, and this will fix #8693

### DIFF
--- a/src/common/quota/errors.go
+++ b/src/common/quota/errors.go
@@ -1,0 +1,111 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


src/common/quota/errors.go
```diff
@@ -0,0 +1,111 @@
...
+// Errors contains all happened errors
+type Errors []error
+
+// GetErrors gets all errors that have occurred and returns a slice of errors (Error type)
+func (errs Errors) GetErrors() []error {
+	return errs
+}
+
+// Add adds an error to a given slice of errors
+func (errs Errors) Add(newErrors ...error) Errors {
```
wy65701436: why do not use pkg/errors.wrap()？

PR #8753, opened 2019-08-20
Document offline migration

arvan-pritchard:
Add instructions for migration using offline installer.
Requested by Wang Yan in Issue 7831


ninjadq:
@arvan-pritchard please fix the confcting

xaleeks:
please also help double check @stuclem

Issue #8599, opened 2019-08-08
Core server is hung when stopping server with CTRL+C


steven-zou:
What can we help you?
<img width="1419" alt="Screen Shot 2019-08-08 at 10 58 57" src="https://user-images.githubusercontent.com/5753287/62672034-b2c0f300-b9cb-11e9-9dcd-427cd31ab45b.png">

ywk253100:
Fixed

PR #8787, opened 2019-08-22
Fix core hung when stop problem

cd1989:

Issue: https://github.com/goharbor/harbor/issues/8599


wy65701436:
@cd1989 please fix the travis.

PR #8797, opened 2019-08-22
fix #8688

### LINKED ISSUE #8688: Backend should return 400 when the quota is exceeded.
I update the quota of my project, so the current number of images exceeds the "count quota".

```
~/harbor-tst/harbor# docker push 10.184.64.157/library/postgres:jt
The push refers to repository [10.184.64.157/library/postgres]
b07a445f2bbe: Layer already exists
20ce01b5a27f: Layer already exists
10731cced086: Layer already exists
f3770cb1e616: Layer already exists
41a11ca907b1: Layer already exists
52dec37e8405: Layer already exists
fbec570c0136: Layer already exists
ce2813200d75: Layer already exists
4b1516981157: Layer already exists
235b155548bc: Layer already exists
b3cda402d3b2: Layer already exists
f471b002050c: Layer already exists
748e596c77e9: Layer already exists
d56055da3352: Layer already exists
received unexpected HTTP status: 500 Internal Server Error
```

500 indicates something wrong at the server side, so we better returns 400

wy65701436:
use the docker defined error to avoid retry pushing on quota overflow

### DIFF
--- a/src/core/middlewares/countquota/handler.go
+++ b/src/core/middlewares/countquota/handler.go
@@ -21,6 +21,7 @@ import (
 	"github.com/goharbor/harbor/src/common/utils/log"
 	"github.com/goharbor/harbor/src/core/middlewares/interceptor"
 	"github.com/goharbor/harbor/src/core/middlewares/util"
+	"strings"
 )
 
 type countQuotaHandler struct {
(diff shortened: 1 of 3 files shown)


wy65701436:
image

PR #8801, opened 2019-08-23
Allow chartserver work with NFS

hlwanghl:
NFS server doesn't allow change owner, which is the same as what we do in registry/entrypoint.sh.


ninjadq:
This PR will block us to move all containers to non-root user.  So close this pr and will add these logic to prepare script

PR #8814, opened 2019-08-23
adjust alg processor: rule first and then selector

steven-zou:


steven-zou:
No need to change

PR #8819, opened 2019-08-23
fix #8815 :add remove untagged blob record in table project_blob

### LINKED ISSUE #8815: Quota doesn't work as expected after running GC
The steps to reproduce the issue:
1. Prepare a totally fresh Harbor instance without any data
2. Create a project called `yinw` and set storage quota to `200M`
3. Push `golang:1.11.2`(280M) to `yinw`, failed and 160M blobs are pushed, the usage is changed to 160M
4. Run GC, the usage is changed to 0
5. Push again, success and the usage is changed to 120M

wy65701436:

### DIFF
--- a/src/common/dao/project_blob.go
+++ b/src/common/dao/project_blob.go
@@ -134,3 +134,57 @@ WHERE af.project_id = ?
 
 	return size, err
 }
+
+// RemoveUntaggedBlobs ...
+func RemoveUntaggedBlobs(pid int64) error {
+	var blobs []models.Blob
+	sql := `
+SELECT 
+    DISTINCT bb.digest,
+    bb.id,
+    bb.content_type,
+    bb.size,
+    bb.creation_time
+FROM artifact af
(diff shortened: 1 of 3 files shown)


src/common/dao/project_blob.go
```diff
@@ -134,3 +134,45 @@ WHERE af.project_id = ?
...
+	var pbIDs []interface{}
+	for _, pb := range projectBlobs {
+		pbIDs = append(pbIDs, pb.ID)
+	}
+
+	if len(pbIDs) == 0 {
+		return nil
+	}
+
+	sql = fmt.Sprintf(`DELETE FROM project_blob WHERE id IN (%s)`, ParamPlaceholderForIn(len(pbIDs)))
```
ywk253100: The two sql can be merged into one

wy65701436: as the pq will raise an exception when the IN or NOT IN condition array is empty.

PR #8831, opened 2019-08-26
Add placeholder in registry page when select helm-hub

jwangyangls:
image


src/portal/lib/src/create-edit-endpoint/create-edit-endpoint.component.ts
```diff
@@ -217,6 +219,9 @@ export class CreateEditEndpointComponent
     if (selectValue === 'docker-hub') {
       this.urlDisabled = true;
       this.targetForm.controls.endpointUrl.setValue(DOCKERHUB_URL);
+    } else if (selectValue === 'helm-hub') {
```
zhoumeina: Please use a const for helm-hub.

Issue #8816, opened 2019-08-23
Tag retention rule should not created when repo exclude all


ninjadq:
When rule's repo part is `exclude **` it is not affected to any repo. 
![image](https://user-images.githubusercontent.com/1069785/63583274-42ba8b80-c5cd-11e9-8695-2027c162999c.png)
Suggest restricting the user to create a rule like this.

bitsf:
This is actually harmless, why we have to check this, I don't think we have to challenge user for this.

AllForNothing:
fixed in PR #8834

Issue #8811, opened 2019-08-23
[tag retention] The conflict rules can be created by "Edit"


wy65701436:
[Steps]
1, create Rule A in UI, with retain always.
2, create Rule B in UI, with different options.
3, edit Rule B, make it exactly the same as Rule A.

After that, I get the two same rules, but not 409 in the log.

BTW, If I try to create the same rule with one the existing, the UI can index me the conflict error.

bitsf:
Not reproduced

AllForNothing:
fixed in PR #8834

Issue #8805, opened 2019-08-23
Tag Retention :the madel should not close,and it should be show the error message in  current page  when save error


jwangyangls:
![image](https://user-images.githubusercontent.com/29826789/63573108-13991f80-c5b7-11e9-91f8-e8d607c19622.png)
![image](https://user-images.githubusercontent.com/29826789/63573440-eef17780-c5b7-11e9-9d69-cd21d73bc3af.png)

bitsf:
fixed by https://github.com/goharbor/harbor/pull/8812

jwangyangls:
After discussion, it is considered that the maximum setting of 20190904 is more appropriate.

AllForNothing:
fixed in PR #8834

Issue #8804, opened 2019-08-23
add rule in tag-retention: first selection is selected by default  in Safari


AllForNothing:
![image](https://user-images.githubusercontent.com/30999793/63572674-fb74d080-c5b5-11e9-9fa4-24f6764935c5.png)

AllForNothing:
fixed in PR #8834

PR #8834, opened 2019-08-26
Improvement for rule validator of tag-retention

AllForNothing:
image
#8816 
#8811 
#8805 
#8804 
#8803
Signed-off-by: Shijun Sun (c) <sshijun@vmware.com>

### DIFF
--- a/src/portal/src/app/project/tag-retention/add-rule/add-rule.component.ts
+++ b/src/portal/src/app/project/tag-retention/add-rule/add-rule.component.ts
@@ -16,12 +16,17 @@ import {
     OnInit,
     OnDestroy,
     Output,
-    EventEmitter,
+    EventEmitter, ViewChild, Input,
 } from "@angular/core";
-import { Rule, RuleMetadate } from "../retention";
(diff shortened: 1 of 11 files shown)


src/portal/src/app/project/tag-retention/add-rule/add-rule.component.ts
```diff
@@ -163,9 +179,50 @@ export class AddRuleComponent implements OnInit, OnDestroy {
...
+        if (this.retention && this.retention.rules && this.retention.rules.length > 0) {
+            for (let i = 0; i < this.retention.rules.length; i++) {
+                if (this.isSameRule(this.retention.rules[i])) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+    isSameRule(rule: Rule): boolean {
```
zhoumeina: Why not just compare the rule object are equal? Is there some property different?

AllForNothing: according to the design, some key properties are equal, then we view these two rules are the same

PR #8837, opened 2019-08-26
Disable redis and db containers if external db enabled

ninjadq:
If depend on external redis or pg. local db and redis should not start. Therefore can save some resources.


reasonerjt:
@wy65701436 could you review?

PR #8846, opened 2019-08-27
add valid for rule

bitsf:
Change-Id: I82215a0cf1ec32a253c8db9bfafe7e25b26c9ad9

### DIFF
--- a/src/pkg/retention/policy/rule/dayspl/evaluator.go
+++ b/src/pkg/retention/policy/rule/dayspl/evaluator.go
@@ -15,6 +15,8 @@
 package dayspl
 
 import (
+	"fmt"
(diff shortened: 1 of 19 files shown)


src/pkg/retention/policy/rule/dayspl/evaluator.go
```diff
@@ -68,3 +70,22 @@ func New(params rule.Parameters) rule.Evaluator {
 
 	return &evaluator{n: DefaultN}
 }
+
+// Valid ...
+func Valid(params rule.Parameters) error {
+	if params != nil {
```
wy65701436: if params == nil {
  return nil
}
......

PR #8854, opened 2019-08-27
fix(build): max idle and open conn settings for external db

heww:


heww:
@ninjadq I do not think the migration script needs these settings for it only run in one process and do not need many db connections.

PR #8867, opened 2019-08-28
Tag retention same digest tag

bitsf:
prevent retained tag with same digest deleted by other tag


src/pkg/retention/res/candidate.go
```diff
@@ -84,7 +86,10 @@ type Candidate struct {
 
 // Hash code based on the candidate info for differentiation
 func (c *Candidate) Hash() string {
-	raw := fmt.Sprintf("%s:%s/%s:%s", c.Kind, c.Namespace, c.Repository, c.Tag)
+	if c.Digest == "" {
+		panic("Lack Digest of Candidate")
```
ywk253100: Don't panic here, panic crashes the process. Just log the error is ok

Issue #8826, opened 2019-08-23
Document CVE whitelist


stuclem:
New feature requires doc.

stuclem:
See https://github.com/goharbor/harbor/issues/7942 for the related Epic.

stuclem:
Fixed in https://github.com/goharbor/harbor/pull/8869.

Issue #8823, opened 2019-08-23
Document project quotas


stuclem:
New feature requires doc.

stuclem:
See https://github.com/goharbor/harbor/issues/8098 for the associated epic.

stuclem:
Fixed in https://github.com/goharbor/harbor/pull/8869.

Issue #8825, opened 2019-08-23
Document webhook notification


stuclem:
New feature requires doc.

stuclem:
See https://github.com/goharbor/harbor/issues/8266 for the related Epic.

stuclem:
Fixed in https://github.com/goharbor/harbor/pull/8869.

Issue #8824, opened 2019-08-23
Document tag retention


stuclem:
New feature requires doc.

stuclem:
Can't find an associated epic, but there is a label `area/tag-retention`. This is the [list of all issues targetting 1.9 with the tag-retention label](https://github.com/goharbor/harbor/issues/8266#workspaces/cncf-5b84fc036092773c0b854fe6/board?labels=target%2F1.9.0,area%2Ftag-retention&repos=50613991).

stuclem:
Fixed in https://github.com/goharbor/harbor/pull/8869.

PR #8869, opened 2019-08-28
Adding docs about webhooks, tag retention, CVE whitelists and project quotas

stuclem:
Towards https://github.com/goharbor/harbor/issues/8825.

@reasonerjt and @xaleeks can you please review or assign a reviewer? Thanks!


docs/user_guide.md
```diff
@@ -719,6 +720,76 @@ If you are a project admin, you can disable a Robot Account by clicking "Disable
...
 
+## Configure Webhook Notifications
+
+If you are a project admin, you can configure a connection from a project in Harbor to a webhook endpoint. If you configure webhooks, Harbor notifies the webhook endpoint of certain events that occur in the project. Webhooks allow you to integrate Harbor with other tools to streamline continuous integration and development processes. 
+
+The action that is taken upon receiving a notification from a Harbor project depends on your continuous integration and development process. For example, by configuring Harbor to send a `POST` request to a webhook listener at an endpoint of your choice, you can trigger a build and deployment of an application whenever there is a change to an image in the repository.
+
+### Supported Events
+
+You can define one webhook per project. Webhook notifications provide information about events in JSON format and are delivered by `HTTP` or `HTTPS POST` to an existing webhhook endpoint URL that you provide. The following table describes the events that trigger notifications and the contents of each notification.
```
xaleeks: You can configure multiple webhooks per project, but only one endpoint

docs/user_guide.md
```diff
@@ -719,6 +720,76 @@ If you are a project admin, you can disable a Robot Account by clicking "Disable
...
+|Image scan completed|`IMAGE SCAN COMPLETED`|Repository namespace name, repository name, tag scanned, image name, number of critical issues, number of major issues, number of minor issues, last scan status, scan completion time timestamp|
+|CVE found in image scan|`CVE FOUND`|Repository namespace name, repository repository, tags, vulnerability information (CVE ID, description, link to CVE, criticality, URL for any fix), username of user who performed scan|
+|Image scan failed|`IMAGE SCAN FAILED`|Repository namespace name, repository name, tag scanned, image name, error that occurred, username of user who performed scan|
+|Image replication|`IMAGE REPLICATION`|Repository name, replication type (docker image or helm chart), image/chart name, image/chart size, source registry type (Harbor or if 3rd party), destination registry type (Harbor or if 3rd party), source registry name, destination registry name, source namespace, destination namespace, authentication type, replication type (push or pull based) username of user who configured replication|
```
xaleeks: This unfortunately didn't make it to the 1.10 release, so not supported

docs/user_guide.md
```diff
@@ -719,6 +720,76 @@ If you are a project admin, you can disable a Robot Account by clicking "Disable
...
+|Delete manifest from registry|`IMAGE DELETE`|Repository namespace name, repository name, manifest digest, image name, image size, delete time timestamp, username of user who deleted image|
+|Upload Helm chart to registry|`CHART PUSH`|Repository name, chart name, chart type, chart version, chart size, tag, timestamp of push, username of user who uploaded chart|
+|Download Helm chart from registry|`CHART PULL`|Repository name, chart name, chart type, chart version, chart size, tag, timestamp of push, username of user who pulled chart|
+|Delete Helm chart from registry|`CHART DELETE`|Repository name, chart name, chart type, chart version, chart size, tag, timestamp of delete, username of user who deleted chart|
+|Image scan completed|`IMAGE SCAN COMPLETED`|Repository namespace name, repository name, tag scanned, image name, number of critical issues, number of major issues, number of minor issues, last scan status, scan completion time timestamp|
+|CVE found in image scan|`CVE FOUND`|Repository namespace name, repository repository, tags, vulnerability information (CVE ID, description, link to CVE, criticality, URL for any fix), username of user who performed scan|
```
xaleeks: CVE found has been merged into the SCAN COMPLETED as a single webhook, no longer existing on its own

docs/user_guide.md
```diff
@@ -719,6 +720,76 @@ If you are a project admin, you can disable a Robot Account by clicking "Disable
...
+|Image scan failed|`IMAGE SCAN FAILED`|Repository namespace name, repository name, tag scanned, image name, error that occurred, username of user who performed scan|
+|Image replication|`IMAGE REPLICATION`|Repository name, replication type (docker image or helm chart), image/chart name, image/chart size, source registry type (Harbor or if 3rd party), destination registry type (Harbor or if 3rd party), source registry name, destination registry name, source namespace, destination namespace, authentication type, replication type (push or pull based) username of user who configured replication|
+
+### Webhook Endpoint Requirements
```
xaleeks: Should these be called requirements? or just some kind of 'advisable'

docs/user_guide.md
stuclem: I added that you can add multiple webhooks per project, but I cannot see how to do so in the UI. @reasonerjt ?

docs/user_guide.md
stuclem: Removed, and also removed the example use case for replication.

docs/user_guide.md
stuclem: Merged into the row above.

docs/user_guide.md
stuclem: Renamed to Webhook Endpoint **Recommendations** and stated that the "endpoint that receives the webhook **should ideally** meet the following requirements"

stuclem:
Thanks @xaleeks. I incorporated your comments in https://github.com/goharbor/harbor/pull/8869/commits/8a699d16ecd8a71b5febe124e861a683e125077d. 

Outstanding questions, that @reasonerjt can hopefully help with:

- How does one upload the cert for `HTTPS POST` notifications?
- Can you please provide a sample Auth Header?
- How do you add multiple webhooks to a single project?

Thanks in advance.

reasonerjt:
I think we need more detailed JSON object example to help user write handler on the server who receives the webhooks.

docs/user_guide.md
```diff
@@ -719,6 +721,70 @@ If you are a project admin, you can disable a Robot Account by clicking "Disable
...
+1. To implement `HTTPS POST` instead of `HTTP POST`, select the **Verifiy Remote Certficate** check box.
+1. Click **Test Endpoint** to make sure that Harbor can connect to the listener.
+1. Click **Continue** to create the webhook.
+
+When you have created the webhook, you see the status of the different notifications and the timestamp of the last time each notification was triggered. You can click **Disable** to disable notifications.
+
+![Webhook Status](img/webhooks3.png)
+
+**NOTE**: You can only disable and reenable all notifications. You cannot disable and enable selected notifications.
+
```
reasonerjt: We should also mention that if the webhook fails to send or receives error http response (with non-2xx code) the webhook will be re-sent based on configuration in `harbor.yml`
@wy65701436  @mmpei please confirm

stuclem:
Towards https://github.com/goharbor/harbor/issues/8824

docs/user_guide.md
mmpei: alex means that you can define one webhook policy to subscribe multiple webhook events.
one webhook policy per project. one endpoint  for policy but multiple events for it.

docs/user_guide.md
mmpei: yes. and there is a system configuration in `configuration`->`system settings` to enable/disable webhook globally.

docs/user_guide.md
```diff
@@ -719,6 +721,70 @@ If you are a project admin, you can disable a Robot Account by clicking "Disable
...
+You can configure your continuous integration and development infrastructure so that it performs the following types of operations when it receives a webhook notification from Harbor.
+
+- Image push: 
+  - Trigger a new build immediately following a push on selected repositories s or tags.
+  - Notify services or applications that use the image that a new image is available and pull it.
+  - Scan the image using Clair.
+  - Replicate the image to remote registries.
+- Image scanning:
+  - If a vulnerability is found, rescan the image or replicate it to another registry.
+  - If the scan passes, build the image.
```
mmpei: i think you mean deploy the image when the scan passes.

docs/user_guide.md
```diff
@@ -719,6 +721,70 @@ If you are a project admin, you can disable a Robot Account by clicking "Disable
...
+|Image scan failed|`IMAGE SCAN FAILED`|Repository namespace name, repository name, tag scanned, image name, error that occurred, username of user who performed scan|
+
+### Webhook Endpoint Recommendations
+
+The endpoint that receives the webhook should ideally meet the following requirements:
+
+- Have a webhook listener that is capable of interpreting the payload and acting upon the information it contains. For example, running a shell script.
+- Support queues and retry configurations, so that failed webhook notifications are queued and retried until the request succeeds or number of retries exceeds the configured maximum.
+
+It is recommended to create a separate endpoint URL for each project.
```
mmpei: it is our mode now. but i still think it is a little weird if we add this recommendation here.
i believe there is only one endpoint URL for most users.
do you have any opinion? @reasonerjt @xaleeks

docs/user_guide.md
```diff
@@ -719,6 +721,70 @@ If you are a project admin, you can disable a Robot Account by clicking "Disable
...
+
+The action that is taken upon receiving a notification from a Harbor project depends on your continuous integration and development processes. For example, by configuring Harbor to send a `POST` request to a webhook listener at an endpoint of your choice, you can trigger a build and deployment of an application whenever there is a change to an image in the repository.
+
+### Supported Events
+
+You can define one webhook endpoint per project. Webhook notifications provide information about events in JSON format and are delivered by `HTTP` or `HTTPS POST` to an existing webhhook endpoint URL that you provide. The following table describes the events that trigger notifications and the contents of each notification.
+
+|Event|Webhook Event Type|Contents of Notification|
+|---|---|---|
+|Push manifest to registry|`IMAGE PUSH`|Repository namespace name, repository name, tags, manifest digest, image name, image size, push time timestamp, username of user who pushed image|
```
mmpei: i think `Push Image to registry` is better here and so to others. and the content is a little different from our implementation. you can refer [here](https://github.com/goharbor/community/pull/36). 
size is not included,  and there is an addtional attribution `resource URL` in content.

docs/user_guide.md
```diff
@@ -719,6 +721,200 @@ If you are a project admin, you can disable a Robot Account by clicking "Disable
...
+
+A tag retention rule has 3 filters that are applied sequentially, as described in the following table.
+
+|Order|Filter|Description|
+|---|---|---|
+|First|Repository or repositories|Identify the repository or repositories on which to apply the rule. You can identify repositories that either have a certain name or name fragment, or that do not have that name or name fragment. Wild cards (for example `*repo`, `repo*`, and `**`) are permitted.|
+|Second|Quantity to retain|Set which tags to retain either by specifying a maximum number of tags, or by specifying a maximum period for which to retain tags.|
+|Third|Tags to retain|Identify the tag or tags on which to apply the rule. You can identify tags that either have a certain name or name fragment, or that do not have that name or name fragment. Wild cards (for example `*tag`, `tag*`, and `**`) and regular expressions are permitted.|
```
steven-zou: Regular expressions are not supported. I suggest adding the hyperlink of the `double star` we adopted for [reference](https://github.com/bmatcuk/doublestar#patterns).

docs/user_guide.md
```diff
@@ -719,6 +721,200 @@ If you are a project admin, you can disable a Robot Account by clicking "Disable
...
+#### Example 1
+
+- You have 5 repositories in a project, repositories A to E.
+  - Repository A has 100 image tags, all of which have been pulled in the last week.
+  - Repositories B to E each have 6 images, none of which have been pulled in the last month.
+- You set the repository filter to `**`, meaning that all repositories in the project are included.
+- You set the retention policy to retain the 10 most recently pulled images in each repository.
+- You set the tag filter to `**`, meaning that all tags in the repository are included.
+
+In this example the rule retains the 10 most recently pulled images in repository A, and all of the images in repositories B to E, retaining a total of 34 image tags in the project. In other words, the rule does not treat all of the images in repositories A to E as a single pool from which to choose the 10 most recent images. So, even if the 11th to 100th tags in repository A have been pulled more recently than any of the tags in repositories B to E, all of the tags in repositories B to E are retained, because each of those repositories has fewer than 10 tags.
```
steven-zou: `34` -> `24`?

docs/user_guide.md
```diff
@@ -719,6 +721,200 @@ If you are a project admin, you can disable a Robot Account by clicking "Disable
...
+
+#### Example 2
+
+This example uses the same project and repositories as example 1, but sets the retention policy to retain the images in each repository that have been pulled in the last 7 days.
+
+In this case, all of the images in repository A are retained because they have been pulled in the last 7 days. None of the images in repositories B to E are retained, because none of them has been pulled in the last week. In this example, 100 images are retained, as opposed to 34 images in example 1.
```
steven-zou: `34` -> `24` too?

docs/user_guide.md
```diff
@@ -719,6 +721,200 @@ If you are a project admin, you can disable a Robot Account by clicking "Disable
...
+
+### How Tag Retention Rules Work
+
+You define tag retention rules on repositories, not on projects. This allows for greater granularity when defining your retention rules. As the name suggests, when you define a retention rule for a repository, you are identifying which tags to retain. You do not define rules to explicitly remove tags. Rather, when you set a rule, any tags in a repository that are not identified as being eligible for retention are discarded. 
+
+A tag retention rule has 3 filters that are applied sequentially, as described in the following table.
+
+|Order|Filter|Description|
+|---|---|---|
+|First|Repository or repositories|Identify the repository or repositories on which to apply the rule. You can identify repositories that either have a certain name or name fragment, or that do not have that name or name fragment. Wild cards (for example `*repo`, `repo*`, and `**`) are permitted.|
```
xaleeks: @stuclem Hi Stuart, this statement is correct but I want to point something out about repository matching since it behaves differently from tag matching.  The matched repos are the ones that the retention policies are actually performed on, the non-matched repos are not deleted/discarded, they are simply ignored (from this rule).  In other words, there is an ordering to this matching; repo matching happens first and these are 'earmarked' for further matching based on the tag criteria.  This was a design choice to minimize the area of images to be delete.  Not sure if it's worth pointing this out in the user_guide.

docs/user_guide.md
```diff
@@ -719,6 +721,200 @@ If you are a project admin, you can disable a Robot Account by clicking "Disable
...
+|Second|Quantity to retain|Set which tags to retain either by specifying a maximum number of tags, or by specifying a maximum period for which to retain tags.|
+|Third|Tags to retain|Identify the tag or tags on which to apply the rule. You can identify tags that either have a certain name or name fragment, or that do not have that name or name fragment. Wild cards (for example `*tag`, `tag*`, and `**`) and regular expressions are permitted.|
+
+The repository filter is applied first to mark the repositories to which to apply the retention rule. 
+
+#### Example 1
```
xaleeks: Just want to say this behavior is correct.  want @steven-zou to double check these examples carefully as well.

docs/user_guide.md
```diff
@@ -719,6 +721,200 @@ If you are a project admin, you can disable a Robot Account by clicking "Disable
...
+
+#### Example 2
+
+This example uses the same project and repositories as example 1, but sets the retention policy to retain the images in each repository that have been pulled in the last 7 days.
+
+In this case, all of the images in repository A are retained because they have been pulled in the last 7 days. None of the images in repositories B to E are retained, because none of them has been pulled in the last week. In this example, 100 images are retained, as opposed to 34 images in example 1.
+
+### Combining Rules on a Respository
+
+You can define up to 15 rules per project. You can apply multiple rules to a repository or set of repositories. When you apply multiple rules to a repository, they are applied with `OR` logic rather than with `AND` logic. In this way, there is no prioritization of application of the rules on a given repository.
```
xaleeks: maybe useful to add the following consequences that 1)these are running concurrently in the background which means shorter running time, 2) the resulting sets from each rule are then union'ed.

docs/user_guide.md
```diff
@@ -719,6 +721,200 @@ If you are a project admin, you can disable a Robot Account by clicking "Disable
...
+You can define up to 15 rules per project. You can apply multiple rules to a repository or set of repositories. When you apply multiple rules to a repository, they are applied with `OR` logic rather than with `AND` logic. In this way, there is no prioritization of application of the rules on a given repository. 
+
+#### Example 3
+
+This example uses the same project and repositories as examples 1 and 2, but sets two rules:
+
+- Rule 1: Retain all of the images in each repository that have been pulled in the last 7 days.
+- Rule 2: Retain a maximum number of 10 images in each repository.
+
+In this case, repository A only retains the 10 most recent images, even though images 11 to 100 have been pulled in the last week. Repositories B to E retain no images, because none of those images have been pulled in the last week. In this example, 10 images are retained, as opposed to 34 images in example 1 and 100 in example 2.
```
steven-zou: Not clear and seems something wrong here.

For repository A, rule 1 will retain all of the images as they have been pulled in the last week;
For repository A, rule 2 will retain 10 pulled images

In summary, with `OR` relationship, for repository A, all 100 images are retained.

For repository B-E, rule 1 will retain 0 images as no images are pulled in the last week;
For repository B-E, rule 2 will retain 6 (6<10) images

In summary, for repository B-E, each repository will keep 6 images there.

docs/user_guide.md
```diff
@@ -719,6 +721,200 @@ If you are a project admin, you can disable a Robot Account by clicking "Disable
...
+  |---|---|---|
+  |`2.1-your_repo-prod`|`2.1-your_repo-rc`|`2.1-your_repo-release`|
+  |`2.2-your_repo-prod`|`2.2-your_repo-rc`|`2.2-your_repo-release`|
+  |`3.1-your_repo-prod`|`3.1-your_repo-rc`|`3.1-your_repo-release`|
+  |`4.4-your_repo-prod`|`4.4-your_repo-rc`|`4.4-your_repo-release`| 
+
+- You define three tag retention rules on this repository:
+  - Retain the 10 most recently pushed image tags that start with `2`.
+  - Retain the 10 most recently pushed image tags that end with `-prod`.
+  - Exclude the image tag `2.1-your_repo-prod` from the list of tags to retain.
```
steven-zou: This rule is not clear as it's not defined with the retention rule pattern (shown in the UI, not the backend meanings).

docs/user_guide.md
```diff
@@ -719,6 +721,200 @@ If you are a project admin, you can disable a Robot Account by clicking "Disable
...
+1. (Optional) Click **Add Rule** to add more rules, up to a maximum of 15 per project.
+1. (Optional) Under Schedule, click **Edit** and select how often to run the rule.
+  ![Select retention criteria](img/tag-retention4.png)
+  If you select **Custom**, enter a cron job command to schedule the rule. 
+  
+  **NOTE**: If you define multiple rules, the schedule is applied to all of the rules. You cannot schedule different rules to run at different times. 
+1. (Optional) Click **Dry Run** to test the rule or rules that you have defined.
+1. (Optional) Click **Run Now** to run the rule immediately.
+
+To modify an existing rule, click the three vertical dots next to a rule to disable, edit, or delete that rule.
```
steven-zou: Add one more screenshot?

docs/user_guide.md
```diff
@@ -719,6 +721,200 @@ If you are a project admin, you can disable a Robot Account by clicking "Disable
...
+  - `**` to apply the rule to all of the tags in the project. 
+  
+  If you selected **matching**, the rule is applied to the tags you identified. If you selected **excluding**, the rule is applied to all of the tags in the repository except for the ones that you identified.
+1. Click **Add** to save the rule.
+1. (Optional) Click **Add Rule** to add more rules, up to a maximum of 15 per project.
+1. (Optional) Under Schedule, click **Edit** and select how often to run the rule.
+  ![Select retention criteria](img/tag-retention4.png)
+  If you select **Custom**, enter a cron job command to schedule the rule. 
+  
+  **NOTE**: If you define multiple rules, the schedule is applied to all of the rules. You cannot schedule different rules to run at different times.
```
steven-zou: Add `Warning` to indicate that the retention is a destroyed operation, suggest running dry run before actually triggered every time.

docs/user_guide.md
```diff
@@ -719,6 +721,200 @@ If you are a project admin, you can disable a Robot Account by clicking "Disable
...
+
+This example uses the same project and repositories as example 1, but sets the retention policy to retain the images in each repository that have been pulled in the last 7 days.
+
+In this case, all of the images in repository A are retained because they have been pulled in the last 7 days. None of the images in repositories B to E are retained, because none of them has been pulled in the last week. In this example, 100 images are retained, as opposed to 34 images in example 1.
+
+### Combining Rules on a Respository
+
+You can define up to 15 rules per project. You can apply multiple rules to a repository or set of repositories. When you apply multiple rules to a repository, they are applied with `OR` logic rather than with `AND` logic. In this way, there is no prioritization of application of the rules on a given repository. 
+
+#### Example 3
```
xaleeks: Pretty sure this is incorrect, since there are 2 runs they are run independent of each other.  The first will indeed retain all 100 images in repo A, which will get be union'ed with the 10 images in repo A as a result of running the 2nd policy.  In the end, all 100 are retained.  please double check @steven-zou

docs/user_guide.md
xaleeks: I agree with @steven-zou .  @stuclem

docs/user_guide.md
```diff
@@ -719,6 +721,200 @@ If you are a project admin, you can disable a Robot Account by clicking "Disable
...
+- `2.2-your_repo-rc`
+- `2.2-your_repo-release`
+- `3.1-your_repo-prod`
+- `4.4-your_repo-prod`
+
+### How Tag Retention Rules Interact with Project Quotas
+
+The system administrator can set a maximum on the number of artifacts that a project can contain and the amount of storage that it can consume. For information about project quotas, see [Configure Project Quotas](LINK). 
+
+Project quota settings take precedence over tag retention rules. If you set a tag retention rule to keep the 5 most recent images in a repository, but the 5th image takes the total amount of storage consumed by the project above the project quota for storage, the 5th image will be deleted.
```
xaleeks: I don't think this will happen because retention policy only trims down the repo, in other words, running tag retention will not violate any quotas set that was not already in violation of to begin with.  Does that make sense? @stuclem

docs/user_guide.md
stuclem: OK, that's what I now have in the doc:

"You can define one webhook endpoint per project. Webhook notifications provide information about events in JSON format and are delivered by `HTTP` or `HTTPS POST` to an existing webhhook endpoint URL that you provide."

stuclem:
> I think we need more detailed JSON object example to help user write handler on the server who receives the webhooks.

Added the push image example from the PRD in https://github.com/goharbor/harbor/pull/8869/commits/6a38dd45172ec5369c650513c1e747333ebb43de

docs/user_guide.md
stuclem: Added both statements in https://github.com/goharbor/harbor/pull/8869/commits/d86e39bccf0dc1c5fa1260d602ccba47fa405393.

docs/user_guide.md
stuclem: Fixed in https://github.com/goharbor/harbor/pull/8869/commits/e1103b753e11b4d6a794d752620d84c3cccff54f.

docs/user_guide.md
stuclem: Deleted this sentence in https://github.com/goharbor/harbor/pull/8869/commits/e1103b753e11b4d6a794d752620d84c3cccff54f. We can restore it if necessary.

docs/user_guide.md
stuclem: Fixed in https://github.com/goharbor/harbor/pull/8869/commits/e1103b753e11b4d6a794d752620d84c3cccff54f.

docs/user_guide.md
stuclem: I think that this info is useful. Added the following in https://github.com/goharbor/harbor/pull/8869/commits/bb32b5dbb292ef429652f14bca203d19b4d282bd:

The repository filter is applied first to mark the repositories to which to apply the retention rule. The identified repositories are earmarked for further matching based on the tag criteria. No action is taken on the nonspecified repositories at this stage.

docs/user_guide.md
stuclem: OK, leaving this conversation unresolved pending confirmation from @steven-zou.

docs/user_guide.md
stuclem: Added "Rules run concurrently in the background, and the resulting sets from each rule are combined at the end of the run" in https://github.com/goharbor/harbor/pull/8869/commits/617028d798279b4c7fe9c3556ce86d9f3408f36d.

docs/user_guide.md
stuclem: Fixed in https://github.com/goharbor/harbor/pull/8869/commits/a1f64ff5f2c4e2c02337a9d19fa4dee03fafe4db.

docs/user_guide.md
stuclem: @steven-zou I'm not sure what you mean here. Can you please explain a bit more?

docs/user_guide.md
stuclem: Updated to the following in https://github.com/goharbor/harbor/pull/8869/commits/f15d012a403726b6f59604e2418e11c0793ab5e1:

For repository A, rule 1 retains all of the images because they have all been pulled in the last week. Rule 2 retains the 10 most recently pulled images. So, since the two rules are applied with an `OR` relationship, all 100 images are retained in repository A.

For repositories B-E, rule 1 will retain 0 images as no images are pulled in the last week. Rule 2 will retain all 6 images because 6 < 10. So, since the two rules are applied with an `OR` relationship, for repositories B-E, each repository will keep all 6 images.

In this example, all of the images are retained.

docs/user_guide.md
stuclem: I'm not sure. 

- Repo A has 100 images, of which 10 are retained.
- Repos B to E each have 6 images, of which all are retained. 6 x 4 = 24. 
- Total for project: 24 + 10 = 34

The wording wasn't very clear, so I have changed it as follows in https://github.com/goharbor/harbor/pull/8869/commits/f15d012a403726b6f59604e2418e11c0793ab5e1:

"In this example the rule retains the 10 most recently pulled images in repository A, and all 6 of the images in each of the 4 repositories B to E. So, a total of 34 image tags are retained in the project."

docs/user_guide.md
stuclem: See above.

docs/user_guide.md
stuclem: Added in https://github.com/goharbor/harbor/pull/8869/commits/21c394d6d68057c165165bac5086987d5e8e0a2c

docs/user_guide.md
stuclem: I think that this is addressed in https://github.com/goharbor/harbor/commit/617028d798279b4c7fe9c3556ce86d9f3408f36d.

docs/user_guide.md
stuclem: Reworded as follows in https://github.com/goharbor/harbor/pull/8869/commits/38092ca3d28b250ea29e305bb264417c0125a63c:

"Project quota settings take precedence over tag retention rules. If you set a quota on a project, this quota cannot be exceeded. The quota is applied to a project even if you set a retention rule that would exceed it. In other words, you cannot use retention rules to bypass quotas."

docs/user_guide.md
stuclem: Added "**WARNING**: You cannot revert a rule after you run it. It is strongly recommended to perform a dry run before you run rules".

stuclem:
@xaleeks @reasonerjt @mmpei @steven-zou I have incorporated your comments. Please can you check those conversations that are still marked as unresolved?

In addition, I posted some questions above about Webhooks:

> * How does one upload the cert for `HTTPS POST` notifications?
> * Can you please provide a sample Auth Header?

Can you all please take a last look before we merge? 

The doc is more easily readable here:
https://github.com/stuclem/harbor/blob/docs-webhooks/docs/user_guide.md#tag-retention-rules
https://github.com/stuclem/harbor/blob/docs-webhooks/docs/user_guide.md#webhook-notifications

Thanks!

stuclem:
https://github.com/goharbor/harbor/pull/8869/commits/9d2fa63a8dead80358fbbe22c4e663c4b58218e7 fixes https://github.com/goharbor/harbor/issues/8823.

docs/user_guide.md
xaleeks: lgtm

docs/user_guide.md
xaleeks: lgtm!

docs/user_guide.md
xaleeks: lgtm

docs/user_guide.md
xaleeks: yep, lgtm!

stuclem:
https://github.com/goharbor/harbor/pull/8869/commits/534e40b9af4cad5805a2b1894b40c45495813227 fixes https://github.com/goharbor/harbor/issues/8826

docs/user_guide.md
xaleeks: I think this is fine for now, I'm just a little hesitant to define the precedence because it can get tricky.  I think 'cannot use retention rules to bypass quotas' captures it very well.  LGTM for now, thanks for the thoughtful wording.

stuclem:
Easily readable versions of the project quota and CVE whitelist docs are here:

- https://github.com/stuclem/harbor/blob/docs-webhooks/docs/user_guide.md#set-project-quotas
- https://github.com/stuclem/harbor/blob/docs-webhooks/docs/user_guide.md#configure-cve-whitelists

stuclem:
@xaleeks @reasonerjt I think that this PR now includes doc updates for all of the new 1.9 features.

docs/user_guide.md
steven-zou: I mean this rule statement`Exclude the image tag `2.1-your_repo-prod` from the list of tags to retain.` is not following the existing retention rule pattern as it misses some parts, it might be something like:

`Retain always tags that excludes 2.1-your_repo-prod`

docs/user_guide.md
steven-zou: Got it. Resolve the comment.

docs/user_guide.md
steven-zou: Got it. Resolve the comment.

docs/user_guide.md
```diff
@@ -252,6 +256,77 @@ The images can be filtered by labels:
...
+   ![Project quotas](img/project-quota3.png)
+   1. Click **OK**.
+1. To set quotas on an individual project, click the 3 vertical dots next to a project and select **Edit**.
+   ![Project quotas](img/project-quota4.png)
+   1. For **Default artifact count**, enter the maximum number of artifacts that this individual project can contain, or enter `-1` to set the default to unlimited. 
+   1. For **Default storage consumption**, enter the maximum quantity of storage that this individual project can consume, selecting `MB`, `GB`, or `TB` from the drop-down menu.
+
+After you set quotas, the you can see how much of their quotas each project has consumed in the **Project Quotas** tab.
+
+![Project quotas](img/project-quota5.png)
```
wy65701436: Some additional info that needs to notify user,

[Background]
1, Harbor computes size on putting blob/manifest from docker client
2, Harbor computes count on putting manifest from docker client.
3, Shared blob will be only computed once per project. In docker, the definition of blob sharing is globally, but in Harbor, we narrow it down to the project.


1, For `Retag`, it will also book/release resources. 

-  Retag image A from project A to project A, with a different tag. The artifact count increase one, but disk do not change as no more new blobs/manifest introduced.
-  Retag image A from project A to project B. The artifact count and disk are updated accordingly.

2, For `GC`, docker push will put the config/blobs firstly, and finally put the manifest.

- If the count reaches the limit, the blobs of the image can be pushed into the project and the storage usage will be updated accordingly. We can call these blob as `untagged blobs`, and these can be removed by GC, and the storage usage will be returned after GC.

3, The overall storage may larger than the actual disk capacity as mentioned in background #3.

4, Chart size is not calculated(Only for artifact count)

reasonerjt:
The section "Configure CVE Whitelists" looks good to me.

stuclem:
> The section "Configure CVE Whitelists" looks good to me.

Thanks @reasonerjt !

docs/user_guide.md
stuclem: Added the above information in https://github.com/goharbor/harbor/pull/8869/commits/78c7238cdfa7fec16200db9fd9827fd6f07d9436.

docs/user_guide.md
```diff
@@ -252,6 +256,77 @@ The images can be filtered by labels:
...
+To exercise control over resource use, as a system administrator you can set  quotas on projects. You can limit the number of artifacts that a project can contain and limit the amount of storage capacity that a project can consume. You can set default quotas that apply to all projects globally. You can also set quotas on individual projects. If you set global default quotas and also set different quotas on individual projects, the per-project quotas override the global default quotas.
+
+By default, all projects have unlimited quotas for both artifacts and storage use. 
+
+1. Go to **Configuration** > **Project Quotas**.
+   ![Project quotas](img/project-quota1.png)
+1. To set global default quotas on all projects, click **Edit**.
+   ![Project quotas](img/project-quota2.png)
+   1. For **Default artifact count**, enter the maximum number of artifacts that any project can contain, or enter `-1` to set the default to unlimited. 
+   1. For **Default storage consumption**, enter the maximum quantity of storage that any project can consume, selecting `MB`, `GB`, or `TB` from the drop-down menu.
```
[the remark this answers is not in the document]
stuclem: Fixed in https://github.com/goharbor/harbor/pull/8869/commits/78c7238cdfa7fec16200db9fd9827fd6f07d9436.

docs/user_guide.md
stuclem: @steven-zou is this OK now? ^^^^

docs/user_guide.md
stuclem: @steven-zou is this correct, given the comments above?

docs/user_guide.md
stuclem: @steven-zou is this OK now? ^^^^

docs/user_guide.md
stuclem: @xaleeks is this OK now? ^^^^

docs/user_guide.md
stuclem: @wy65701436 can you please review https://github.com/goharbor/harbor/commit/78c7238cdfa7fec16200db9fd9827fd6f07d9436 ? Thanks!

docs/user_guide.md
```diff
@@ -719,6 +810,237 @@ If you are a project admin, you can disable a Robot Account by clicking "Disable
...
+             ]
+}
+```
+
+### Webhook Endpoint Recommendations
+
+The endpoint that receives the webhook should ideally meet the following requirements:
+
+- Have a webhook listener that is capable of interpreting the payload and acting upon the information it contains. For example, running a shell script.
+- Support queues and retry configurations, so that failed webhook notifications are queued and retried until the request succeeds or number of retries exceeds the configured maximum.
```
xaleeks: This is not a requirement of the endpoint, the endpoint just catches the payload.  This is just a general requirement of the webhook mechanism on the harbor side, support if it wasn't clear from the prd. @stuclem

docs/user_guide.md
```diff
@@ -719,6 +810,237 @@ If you are a project admin, you can disable a Robot Account by clicking "Disable
...
+
+### Example Use Cases
+
+You can configure your continuous integration and development infrastructure so that it performs the following types of operations when it receives a webhook notification from Harbor.
+
+- Image push: 
+  - Trigger a new build immediately following a push on selected repositories s or tags.
```
xaleeks: there's a typo here, should be 'repositories or tags'

docs/user_guide.md
```diff
@@ -252,6 +256,95 @@ The images can be filtered by labels:
...
+   ![Add system CVEs](img/cve-whitelist3.png)
+1. Click **Save** at the bottom of the page to save your settings.
+   
+After you have created a system whitelist, you can remove CVE IDs from the list by clicking the delete button next to it in the list. You can click **Add** to add more CVE IDs to the system whitelist.
+
+![Add and remove system CVEs](img/cve-whitelist4.png)
+
+### Configure a Per-Project CVE Whitelist
+
+By default, the system whitelist is applied to all projects. You can configure different CVE whitelists for individual projects, that override the system whitelist. Per-project CVE whitelists only apply to the projects in which you create them.
```
xaleeks: 'Per-project CVE whitelists only apply to the projects in which you create them.', I wonder if this is necessary?  Is this self-evident?  Just thinking out loud here. @stuclem

docs/user_guide.md
```diff
@@ -252,6 +256,95 @@ The images can be filtered by labels:
...
+   Either use a comma-separated list or newlines to add multiple CVE IDs to the list.
+1. Click **Add** at the bottom of the window to add the CVEs to the project whitelist.
+1. Optionally uncheck the **Never expires** checkbox and use the calendar selector to set an expiry date for the whitelist.
+1. Click **Save** at the bottom of the page to save your settings.
+
+After you have created a project whitelist, you can remove CVE IDs from the list by clicking the delete button next to it in the list. You can click **Add** at any time to add more CVE IDs to the whitelist for this project. 
+
+If CVEs are added to the system whitelist after you have created a project whitelist, click **Add System** to add the new entries from the system whitelist to the project whitelist. 
+
+**NOTE**: If CVEs are deleted from the system whitelist after you have created a project whitelist, and if you added the system whitelist to the project whitelist, you must manually remove the deleted CVEs from the project whitelist. If you click **Add System** after CVEs have been deleted from the system whitelist, the deleted CVEs are not automatically removed from the project whitelist.
```
xaleeks: regarding 'If CVEs are deleted from the system whitelist after you have created a project whitelist, and if you added the system whitelist to the project whitelist, you must manually remove the deleted CVEs from the project whitelist.'  Should this be the correct behavior?  I wonder if these should be automatically removed from the project, ie 'de-merged' from the project whitelist at runtime.  Or is that not possible?  @reasonerjt

docs/user_guide.md
```diff
@@ -252,6 +256,95 @@ The images can be filtered by labels:
...
+After you set quotas, the you can see how much of their quotas each project has consumed in the **Project Quotas** tab.
+
+![Project quotas](img/project-quota5.png)
+
+### How Harbor Calculates Resource Usage
+
+When setting project quotas, it is useful to know how Harbor calculates artifact numbers and storage use, especially in relation to image pushing, retagging, and garbage collection.
+
+- Harbor computes image size when blobs and manifests are pushed from the Docker client.
+- Harbor computes artifact counts when manifests are pushed from the Docker client.
```
xaleeks: There's a salient point here which is manifest is pushed at the end after all blobs have been pushed successfully so in the case of pushing 2 images at once when there's only one 'spot' left, they don't get rejected right away which obviates a lot of the unnecessary pushing of blobs but only after the faster of the two is pushed.  Not sure if worth mentioning. @stuclem

docs/user_guide.md
```diff
@@ -252,6 +256,95 @@ The images can be filtered by labels:
...
+- Shared blobs are only computed once per project. In Docker, blob sharing is defined globally. In Harbor, blob sharing is defined at the project level. As a consequence, overall storage usage can be greater than the actual disk capacity.
+- Retagging images reserves and releases resources: 
+  -  If you retag an image within a project, the artifact count increases by one, but storage usage does not change because there are no new blobs or manifests.
+  - If you retag an image from one project to another, the artifact count and storage usage both increase.
```
xaleeks: So we can safely say that artifact count = tag count, correct? @wy65701436

docs/user_guide.md
stuclem: @steven-zou, so if in the UI I select tag *excluding* and enter `2.1-your_repo-prod` does that mean that all tags except for `2.1-your_repo-prod` are retained? Or that only `2.1-your_repo-prod` is retained?

docs/user_guide.md
stuclem: @xaleeks I deleted "Project quota settings take precedence over tag retention rules" and left the rest as it was.

docs/user_guide.md
stuclem: @xaleeks  I deleted that bullet and merged the previous bullet into the first sentence.

docs/user_guide.md
stuclem: Fixed.

docs/user_guide.md
stuclem: Agreed. Deleted.

docs/user_guide.md
stuclem: @xaleeks and @reasonerjt I tested as follows:

- Created a per-project list
- Created a system-wide list
- In the project list, clicked Add System and saw the CVEs from the system list appear in the project list.
- Deleted a CVE from the system list.
- Went back to the project list and clicked Add System again to see if that updated the project list, but the CVE that I removed from the system list remained in the project list

docs/user_guide.md
stuclem: @xaleeks it's your call. If you think it's necessary, I'll add it. I suppose the question is "will customers notice and be confused about why one of their images didn't push?"

docs/user_guide.md
stuclem: If artifact == tag, we should state that somewhere. If things other than tags count as artifacts, we should state that and list them. @wy65701436 can you please clarify?

docs/user_guide.md
xaleeks: lgtm

docs/user_guide.md
xaleeks: Thx for verifying.  I'll make a note on this.  Doc-wise, this is fine, lgtm

docs/user_guide.md
xaleeks: Hey Stuart, if you can make an attempt at wording this in a manner which answers the question you pointed out, would be appreciated.  @stuclem

docs/user_guide.md
xaleeks: @stuclem It is the same thing, I verified with @wy65701436 today

docs/user_guide.md
stuclem: @xaleeks how's this?

**NOTE**: When users push an image, the manifest is pushed last, after all of the associated blobs have been pushed successfully to the registry. If several images are pushed concurrently and if there is an insufficient number of artifacts left in the quota for all of them, images are accepted in the order that their manifests arrive. Consequently, an attempt to push an image might not be immediately rejected for exceeding the quota. This is because there was availability in the artifact quota when the push was initiated, but by the time the manifest arrived the quota had been exhausted.

docs/user_guide.md
stuclem: Replaced "artifact" with "tag" as appropriate.

docs/user_guide.md
xaleeks: perfect! Thx @stuclem

docs/user_guide.md
stuclem: @xaleeks has looked at this section since the update, so I think that we can resolve this one.

stuclem:
@xaleeks @reasonerjt @wy65701436 @steven-zou I believe that I have now integrated all of your comments. The only things outstanding before we can merge this one are:

- For Webhooks, how does one upload the cert for HTTPS POST notifications?
- Can you please provide a sample webhook Auth Header?
- @steven-zou can you please clarify about how the **excluding** rule for tags works, per https://github.com/goharbor/harbor/pull/8869#discussion_r321313303? 

Thanks in advance!

docs/user_guide.md
```diff
@@ -252,6 +256,97 @@ The images can be filtered by labels:
...
+
+If CVEs are added to the system whitelist after you have created a project whitelist, click **Add System** to add the new entries from the system whitelist to the project whitelist. 
+
+**NOTE**: If CVEs are deleted from the system whitelist after you have created a project whitelist, and if you added the system whitelist to the project whitelist, you must manually remove the deleted CVEs from the project whitelist. If you click **Add System** after CVEs have been deleted from the system whitelist, the deleted CVEs are not automatically removed from the project whitelist.
+
+## Set Project Quotas
+
+To exercise control over resource use, as a system administrator you can set  quotas on projects. You can limit the number of tags that a project can contain and limit the amount of storage capacity that a project can consume. You can set default quotas that apply to all projects globally. You can also set quotas on individual projects. If you set global default quotas and also set different quotas on individual projects, the per-project quotas override the global default quotas.
```
heww: `You can set default quotas that apply to all projects globally.` there is something not clearly for this.

The default quotas only have effect for the new created projects, once the project created, its quotas will be confirmed. The Changing of the default quota settings will not have effect for the projects already exist.

As the above description, there is no scenario to mean `If you set global default quotas and also set different quotas on individual projects, the per-project quotas override the global default quotas.`,  set different quotas on individual projects is just update quotas to the individual projects.

docs/user_guide.md
```diff
@@ -252,6 +256,97 @@ The images can be filtered by labels:
...
+- Shared blobs are only computed once per project. In Docker, blob sharing is defined globally. In Harbor, blob sharing is defined at the project level. As a consequence, overall storage usage can be greater than the actual disk capacity.
+- Retagging images reserves and releases resources: 
+  -  If you retag an image within a project, the tag count increases by one, but storage usage does not change because there are no new blobs or manifests.
+  - If you retag an image from one project to another, the tag count and storage usage both increase.
+- For garbage collection, `docker push` adds the config and blobs first, and then adds the manifest.
```
heww: When run garbage collection, harbor will free the untagged blobs storage usage in the project.

stuclem:
Changed CVE whitelist button from **Add System** to **Copy from System** in https://github.com/goharbor/harbor/pull/8869/commits/67105a4153c996c8ab1319e0f3eee42bf2ac4ece, per https://github.com/goharbor/harbor/pull/9038.

docs/user_guide.md
stuclem: Thanks @heww I reworded this in https://github.com/goharbor/harbor/pull/8869/commits/bb795ef05621fb35aa482e19187476799501e57c. Is this OK now?

docs/user_guide.md
stuclem: @heww I fixed this in https://github.com/goharbor/harbor/pull/8869/commits/bb795ef05621fb35aa482e19187476799501e57c. Is this OK now?

heww: It's ok now, thanks!

reasonerjt:
@stuclem could you squash the commits

stuclem:
> @stuclem could you squash the commits

Yes, I will squash when I merge. I'm still waiting on one bit of feedback.

docs/user_guide.md
xaleeks: @stuclem  He's not referring to the behavior.  He's referring to the format, that all policies start with 'retain ...', so 'Exclude the image tag `2.1-your_repo-prod` from the list of tags to retain.' should perhaps be better stated as 'Retain all tags excluding ...'   To answer your question in the last comment,  if in the UI you select tag excluding and enter 2.1-your_repo-prod, then all tags except 2.1-your_repo-prod are retained.

@steven-zou I don't think your way is necessarily better because this is a user guide, it should follow natural speech pattern, but maybe ok for consistency sake. 

You guys can comment on which is actually better. I just wanted to clear up some misunderstanding.

PR #8891, opened 2019-08-29
Fix: prepare permission issue

ninjadq:
1. recursivele change ownership for all prepare dir
2. database file permission fix


reasonerjt:
@ninjadq I see there some code change in `/src/pkg/retention` is that intentional?

PR #8907, opened 2019-08-30
Handle error on OIDC callback

airadier:
Same PR as https://github.com/goharbor/harbor/pull/8820 but rebased on **master** (although would be nice to merge also PR 8820 into 1.8.x).

If wrong OIDC scopes are defined, or there are some configuration errors, the OIDC callback query string might contain "error=..." with an error message. Intercept this case and show an error to the user instead of trying to exchange the token with a missing "code" parameter.


reasonerjt:
Thanks @airadier

PR #8910, opened 2019-09-01
fix(quota): correct size quota for image with foreign layers

heww:
1. Sync blobs from manifest for image with foreign layers.
2. Ignore size of foreign layers when compute size quota.
3. Fix repo info of artifact when upgrade from 1.8 version.

### DIFF
--- a/src/common/dao/blob.go
+++ b/src/common/dao/blob.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/docker/distribution"
 	"github.com/goharbor/harbor/src/common/models"
 	"github.com/goharbor/harbor/src/common/utils/log"
 )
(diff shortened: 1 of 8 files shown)


src/common/dao/blob.go
```diff
@@ -64,6 +65,107 @@ func DeleteBlob(digest string) error {
...
+		}
+
+		return err
+	}
+
+	return nil
+}
+
+// CleanupBlobs remove blobs which are not referenced by artifact blob
+func CleanupBlobs() error {
```
wy65701436: The IN could in an empty condition? that could lead to a PG error.

src/core/middlewares/sizequota/builder.go
```diff
@@ -119,6 +119,11 @@ func (*manifestCreationBuilder) Build(req *http.Request) (interceptor.Intercepto
 	// Replace request with manifests info context
 	*req = *req.WithContext(util.NewManifestInfoContext(req.Context(), info))
 
+	// Sync manifest layers to blobs for foreign layers not pushed and they are not in blob table
+	if err := info.SyncBlobs(); err != nil {
```
wy65701436: Could we just log err instead of blocking PUT manifest? As all of the required information has been inserted into DB at this time. 

Sync err will not break any further action?

src/common/dao/blob.go
```diff
@@ -64,6 +65,107 @@ func DeleteBlob(digest string) error {
...
+	mp := make(map[string]*models.Blob, len(existing))
+	for _, blob := range existing {
+		mp[blob.Digest] = blob
+	}
+
+	var missing, updating []*models.Blob
+	for _, reference := range references {
+		if blob, found := mp[reference.Digest.String()]; found {
+			if blob.ContentType == "" {
+				blob.ContentType = reference.MediaType
```
wy65701436: if blob.ContentType == "" || blob.ContentType != reference.MediaType {
  ....
}

src/jobservice/job/impl/gc/job.go
```diff
@@ -95,6 +95,9 @@ func (gc *GarbageCollector) Run(ctx job.Context, params job.Parameters) error {
 	if err := gc.ensureQuota(); err != nil {
 		gc.logger.Warningf("failed to align quota data in gc job, with error: %v", err)
 	}
+	if err := dao.CleanupBlobs(); err != nil {
```
wy65701436: Is it a required action?

src/common/dao/blob.go
heww: changed to `if blob.ContentType != reference.MediaType {`

src/core/middlewares/sizequota/builder.go
heww: change to only log the error

src/jobservice/job/impl/gc/job.go
heww: removed

src/common/dao/blob.go
heww: this sql will work well even artifact_blob is empty, but this func removed for the job of gc will not call it.

src/core/api/quota/registry/registry.go
```diff
@@ -423,7 +423,7 @@ func infoOfRepo(pid int64, repo string) (quota.RepoData, error) {
 		}
 		af := &models.Artifact{
 			PID:          pid,
-			Repo:         strings.Split(repo, "/")[1],
+			Repo:         repo,
```
wy65701436: good catch, buddy.

PR #8912, opened 2019-09-02
Add no-cache to index.html

ninjadq:
shouldn't cache index.html for access fresh page after upgrade.


bitsf:
image
The change will also make other document no-cache, we should only control "/"

bitsf:
I think this can work.
```
        location = /index.html {
            add_header Cache-Control "no-store, no-cache, must-revalidate";
        }
```

PR #8936, opened 2019-09-03
add swagger for tag retention

bitsf:
1. add swagger for tag retention
2. enhance some validation
3. add e2e lib for tag retention

### DIFF
--- a/docs/swagger.yaml
+++ b/docs/swagger.yaml
@@ -80,7 +80,7 @@ paths:
           type: integer
           format: int32
           required: false
-          description: 'The page nubmer, default is 1.'
+          description: 'The page number, default is 1.'
         - name: page_size
           in: query
           type: integer
@@ -280,7 +280,7 @@ paths:
           type: integer
           format: int32
           required: false
-          description: 'The page nubmer, default is 1.'
+          description: 'The page number, default is 1.'
         - name: page_size
           in: query
           type: integer
@@ -679,7 +679,7 @@ paths:
           type: integer
           format: int32
           required: false
-          description: 'The page nubmer, default is 1.'
+          description: 'The page number, default is 1.'
         - name: page_size
           in: query
           type: integer
@@ -793,7 +793,7 @@ paths:
           type: integer
           format: int32
           required: false
-          description: 'The page nubmer, default is 1.'
+          description: 'The page number, default is 1.'
         - name: page_size
           in: query
           type: integer
@@ -1035,7 +1035,7 @@ paths:
           type: integer
           format: int32
           required: false
-          description: 'The page nubmer, default is 1.'
+          description: 'The page number, default is 1.'
         - name: page_size
           in: query
           type: integer
@@ -1598,7 +1598,7 @@ paths:
           type: integer
           format: int32
           required: false
-          description: 'The page nubmer, default is 1.'
+          description: 'The page number, default is 1.'
         - name: page_size
           in: query
           type: integer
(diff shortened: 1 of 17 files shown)


src/pkg/retention/manager.go
```diff
@@ -111,7 +111,7 @@ func (d *DefaultManager) GetPolicy(id int64) (*policy.Metadata, error) {
 	p1, err := dao.GetPolicy(id)
 	if err != nil {
 		if err == orm.ErrNoRows {
-			return nil, nil
+			return nil, fmt.Errorf("No such Retention policy with id %v", id)
```
ywk253100: How can we distinguish the 404 and other error in the API level?

src/core/api/retention.go
```diff
@@ -151,6 +151,10 @@ func (r *RetentionAPI) CreateRetention() {
 		r.SendBadRequestError(err)
 		return
 	}
+	if len(p.Rules) > 15 {
+		r.SendBadRequestError(errors.New("there can be at most 15 rules"))
```
steven-zou: This error message `"there can be at most 15 rules"` seems not very smooth.

“at most, only 15 rules are allowed”?

src/pkg/retention/manager.go
steven-zou: the error should start with lower case

`"No such`-> `no such`

src/pkg/retention/policy/models.go
```diff
@@ -53,27 +53,24 @@ type Metadata struct {
...
-
-	// The max number of rules in a policy
-	Capacity int `json:"cap"`
 }
 
 // Valid Valid
 func (m *Metadata) Valid(v *validation.Validation) {
 	if m.Trigger == nil {
-		_ = v.SetError("Trigger", "Trigger is required")
+		_ = v.SetError("Trigger", "Can not be empty")
```
steven-zou: Seems `"Can not be empty"` is an error message, then it should start with lower case letter

docs/swagger.yaml
```diff
@@ -3968,6 +3968,213 @@ paths:
           description: User have no permission to list webhook jobs of the project.
         '500':
           description: Unexpected internal errors.
+
+  '/retentions/metadatas':
+    get:
+      summary: Get Retention Metadatas
+      description: Get Retention Metadatas.
+      tags:
+        - Products
```
steven-zou: why not add a `retention` tag?

docs/swagger.yaml
```diff
@@ -3968,6 +3968,213 @@ paths:
...
+      responses:
```
steven-zou: No any error cases?

docs/swagger.yaml
```diff
@@ -3968,6 +3968,213 @@ paths:
...
+          description: Retention ID.
+      responses:
+        '200':
+          description: Get Retention Policy successfully.
+          schema:
+            type: object
+            $ref: '#/definitions/RetentionPolicy'
+    put:
+      summary: Update Retention Policy
+      description: Update Retention Policy.
```
steven-zou: Suggest providing more detailed description info in the `description `.

docs/swagger.yaml
bitsf: This just return string, no error cases

src/pkg/retention/policy/models.go
bitsf: Here I use upper case is because beego's default error message is using this.

image

wy65701436:
@bitsf please fix the Codacy

src/pkg/retention/policy/models.go
steven-zou: ok, let's keep consistent with beego. For our own created error, should be lower case error message.

docs/swagger.yaml
steven-zou: I mean, for API, we should handle the error cases, like 400, 404, 403, 409, 500 etc.

PR #8960, opened 2019-09-05
Upgrade hash alg for pswd

ninjadq:
Use sha256 for newer encryption but the previous sha1 will still be used for the old password

### DIFF
--- a/make/migrations/postgresql/0010_1.9.0_schema.up.sql
+++ b/make/migrations/postgresql/0010_1.9.0_schema.up.sql
@@ -186,3 +186,5 @@ create table notification_policy (
 ALTER TABLE replication_task ADD COLUMN status_revision int DEFAULT 0;
 DELETE FROM project_metadata WHERE deleted = TRUE;
 ALTER TABLE project_metadata DROP COLUMN deleted;
+ALTER TABLE harbor_user ADD COLUMN password_version varchar(16) Default 'sha256';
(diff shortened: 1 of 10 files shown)


make/migrations/postgresql/0010_1.9.0_schema.up.sql
```diff
@@ -185,4 +185,5 @@ create table notification_policy (
 
 ALTER TABLE replication_task ADD COLUMN status_revision int DEFAULT 0;
 DELETE FROM project_metadata WHERE deleted = TRUE;
-ALTER TABLE project_metadata DROP COLUMN deleted;
\ No newline at end of file
+ALTER TABLE project_metadata DROP COLUMN deleted;
+ALTER TABLE harbor_user ADD COLUMN password_version int Default 1;
```
reasonerjt: 1. Default to `1` is problematic because we should default to `2`, and mark the old data as version `1` in migration.
 

2. Comparing to introducing an int version,  how about adding a prefix to the hash like `<alg-sha1>` or `<alg-sha256>` which will be more explanatory?  Even setting the alg as a `varchar` is better to me, think how `jwt` encode the alg in the token, instead of obscure int

3. we may consider add `Match password` to the user model, or find better way to map the `version` to algorithm in the code, so we don't need to append `if version == xxxx { ....}` in future

PR #9004, opened 2019-09-06
Documented that Robo accounts can push/pull helm charts

stuclem:
Updated screenshot to include the Helm chart options, and modified text slightly.

@reasonerjt and @xaleeks is this sufficient, or do we need to provide more detail?


docs/user_guide.md
```diff
@@ -688,10 +688,10 @@ new ID token and the CLI secret will work again.
 Robot Accounts are accounts created by project admins that are intended for automated operations. They have the following limitations:
 
 1, Robot Accounts cannot login Harbor portal
-2, Robot Accounts can only perform `docker push`/`docker pull` operations with a token.
+2, Robot Accounts can only perform `docker push` / `docker pull` operations with a token.
```
reasonerjt: Could we rephrase to robot accounts can only perform operations using `docker` and `helm` CLI?  What we want to imply is that it's not eligible for API calls currently

xaleeks: agreed

PR #9063, opened 2019-09-11
Disable users with GuestRole to project logs

henderjm:
Guests can have access to sensitive information through logs like usernames/emails.


reasonerjt:
This is a break-change, we'll need more discussion before merging this one

xaleeks:
we're creating a new role called limited guest that will implement this exactly, limiting visibility into other group members an logs. closing this.

see prd here:
https://docs.google.com/document/d/14sCk6kRkPL8mt2klP2Z35AEA9hCcZa7C8rErdx2_LDI/edit

PR #9099, opened 2019-09-16
Xaleeks changed edit quotas permissions

xaleeks:


xaleeks:
can we get this approved first?  I'll need to make changes to security_disclosure anyway.  But the permissions.md should be correct.  Otherwise I have to submit separate PRs again.  Apologies for mixing these 2 completely unrelated issues.

PR #9123, opened 2019-09-17
Add DAO for immutable tags

stonezdj:
Add CRUD operation on immutable tag rules

### DIFF
--- a/make/migrations/postgresql/0015_1.10.0_schema.up.sql
+++ b/make/migrations/postgresql/0015_1.10.0_schema.up.sql
@@ -1,0 +1,9 @@
+/** Add table for immutable tag  **/
+CREATE TABLE immutable_tag_rule
+(
+  id            SERIAL PRIMARY KEY NOT NULL,
+  project_id    int NOT NULL,
+  tag_filter   text,
+  enabled       boolean default true NOT NULL,
+  creation_time timestamp default CURRENT_TIMESTAMP
+)
--- a/src/common/dao/immutable.go
+++ b/src/common/dao/immutable.go
@@ -1,0 +1,73 @@
+package dao
+
+import (
+	"fmt"
+
+	"github.com/astaxie/beego/orm"
+	"github.com/goharbor/harbor/src/common/models"
+)
+
+// CreateImmutableRule creates the Immutable Rule
+func CreateImmutableRule(ir *models.ImmutableRule) (int64, error) {
+	ir.Enabled = true
+	o := GetOrmer()
+	return o.Insert(ir)
+}
+
+// UpdateImmutableRule update the immutable rules
+func UpdateImmutableRule(projectID int, ir *models.ImmutableRule) (int64, error) {
+	o := GetOrmer()
+	return o.Update(ir, "TagFilter")
+}
+
+// ToggleImmutableRule enable/disable immutable rules
+func ToggleImmutableRule(id int64, enabled bool) (int64, error) {
+	o := GetOrmer()
+	ir := &models.ImmutableRule{ID: id, Enabled: enabled}
+	return o.Update(ir, "Enabled")
+}
+
+// GetImmutableRule get immutable rule
+func GetImmutableRule(id int64) (*models.ImmutableRule, error) {
+	o := GetOrmer()
+	ir := &models.ImmutableRule{ID: id}
+	err := o.Read(ir)
+	if err == orm.ErrNoRows {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, err
+	}
+	return ir, nil
+}
+
+// QueryImmutableRuleByProjectID get all immutable rule by project
+func QueryImmutableRuleByProjectID(projectID int64) ([]models.ImmutableRule, error) {
+	o := GetOrmer()
+	qs := o.QueryTable(&models.ImmutableRule{}).Filter("ProjectID", projectID)
+	var r []models.ImmutableRule
+	_, err := qs.All(&r)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get immutable tag rule by projectID %d, error: %v", projectID, err)
+	}
+	return r, nil
+}
+
+// QueryEnabledImmutableRuleByProjectID get all enabled immutable rule by project
+func QueryEnabledImmutableRuleByProjectID(projectID int64) ([]models.ImmutableRule, error) {
+	o := GetOrmer()
+	qs := o.QueryTable(&models.ImmutableRule{}).Filter("ProjectID", projectID).Filter("Enabled", true)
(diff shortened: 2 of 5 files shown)


make/migrations/postgresql/0015_1.10.0_schema.up.sql
```diff
@@ -0,0 +1,12 @@
+/** Add table for immutable tag  **/
+CREATE TABLE immutable_tag_rule
+(
+  id            SERIAL PRIMARY KEY NOT NULL,
+  project_id    int NOT NULL,
+  repo_filter  text,
+  tag_filter   text,
+  decoration    varchar(100) NOT NULL DEFAULT 'matches',
```
wy65701436: the decoration is for repo or tag? each of them should have one `decoration`?

Or, we can use a JSON data to store all of the parameters?

src/common/dao/immutable.go
```diff
@@ -0,0 +1,94 @@
...
+	ir.Enabled = true
+	ir.Decoration = getValidDecoration(ir.Decoration)
+	if exceedMaxImmutableRuleCount(ir.ProjectID) {
+		return 0, fmt.Errorf("immutable tag rule count exceed %v", common.MaxRuleCount)
+	}
+	o := GetOrmer()
+	return o.Insert(ir)
+}
+
+func getValidDecoration(decoration string) string {
```
wy65701436: validation should be in API part?

src/common/dao/immutable.go
```diff
@@ -0,0 +1,94 @@
...
+
+// UpdateImmutableRule update the immutable rules
+func UpdateImmutableRule(projectID int, ir *models.ImmutableRule) (int64, error) {
+	o := GetOrmer()
+	ir.Decoration = getValidDecoration(ir.Decoration)
+	return o.Update(ir, "RepoFilter", "TagFilter", "Decoration")
+}
+
+// EnableImmutableRule enable/disable immutable rules
+func EnableImmutableRule(id int64, enabled bool) (int64, error) {
```
wy65701436: the name is confusing? If I want to disable any of rule, I still have to call EnableImmutableRule()

stonezdj: change to ToggleImmutableRule

make/migrations/postgresql/0015_1.10.0_schema.up.sql
stonezdj: Remove decoration

src/common/dao/immutable.go
stonezdj: remove it

src/common/models/immutable.go
```diff
@@ -0,0 +1,15 @@
+package models
+
+// ImmutableRule - rule which filter image tags should be immutable.
+type ImmutableRule struct {
+	ID         int64  `orm:"pk;auto;column(id)" json:"id,omitempty"`
+	ProjectID  int64  `orm:"column(project_id)" json:"project_id,omitempty"`
+	RepoFilter string `orm:"column(repo_filter)" json:"repo_filter,omitempty"`
+	TagFilter  string `orm:"column(tag_filter)" json:"tag_filter,omitempty"`
```
ywk253100: How about defining a filter list that can contain kinds of different filters？In this way, the supported filters can be extended easily

src/common/dao/immutable.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/goharbor/harbor/src/common"
+	"github.com/goharbor/harbor/src/common/models"
+)
+
+// CreateImmutableRule creates the Immutable Rule
+func CreateImmutableRule(ir *models.ImmutableRule) (int64, error) {
+	ir.Enabled = true
+	if exceedMaxImmutableRuleCount(ir.ProjectID) {
+		return 0, fmt.Errorf("immutable tag rule count exceed %v", common.MaxRuleCount)
+	}
```
ywk253100: I think the checking rule count should be handled in the upper level. The dao should only handle the data access related operations

src/common/dao/immutable.go
```diff
@@ -0,0 +1,84 @@
...
+	o := GetOrmer()
+	var count int
+	o.Raw("select  count(*) cnt from immutable_tag_rule where project_id = ? ", projectID).QueryRow(&count)
+	return count >= common.MaxRuleCount
+}
+
+// UpdateImmutableRule update the immutable rules
+func UpdateImmutableRule(projectID int, ir *models.ImmutableRule) (int64, error) {
+	o := GetOrmer()
+	return o.Update(ir, "RepoFilter", "TagFilter")
```
ywk253100: Update the `UpdateTime` field?

src/common/dao/immutable.go
```diff
@@ -0,0 +1,84 @@
...
+	var r []models.ImmutableRule
+	_, err := qs.All(&r)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get immutable tag rule by projectID %d, error: %v", projectID, err)
+	}
+	return r, nil
+}
+
+// QueryEnabledImmutableRuleByProjectID get all enabled immutable rule by project
+func QueryEnabledImmutableRuleByProjectID(projectID int64) ([]models.ImmutableRule, error) {
```
ywk253100: Merging `QueryImmutableRuleByProjectID ` and `QueryEnabledImmutableRuleByProjectID ` as `QueryImmutableRules(q *Query)`, the query conditions can be specified in the `Query` object

src/common/dao/immutable.go
```diff
@@ -0,0 +1,84 @@
+package dao
+
+import (
+	"fmt"
+
+	"github.com/astaxie/beego/orm"
+	"github.com/goharbor/harbor/src/common"
+	"github.com/goharbor/harbor/src/common/models"
+)
+
```
ywk253100: Define a interface for the dao of immutable, so that the unit test cases will not depend on the database

src/common/dao/immutable.go
stonezdj: One API is query all, another API is just query enabled rule

src/common/models/immutable.go
stonezdj: Remove the repoFilter, all information store ini tagfilter

src/common/dao/immutable.go
stonezdj: Removed

src/common/dao/immutable.go
stonezdj: Removed

src/common/models/immutable.go
stonezdj: Remove repo filter, all filter stored in tagfilter

PR #9125, opened 2019-09-17
refactor selector of retention

wy65701436:
extract select from pkg/retention, move it to pkg/reselector to make it usable by immutable tag

Signed-off-by: wang yan wangyan@vmware.com


src/pkg/artselector/selector.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package res
+package artselector
```
steven-zou: `src/pkg/art/selectors/`

PR #9134, opened 2019-09-18
support pluggable scanner

steven-zou:
support pluggable scanner

- add DAO layer for scanner registration
- add CURD manager for scanner registration
- add API controller for plug scanner
- add REST APIs for CURD of plug scanner
- add migration sql:0011_1.10.0
- add scan interface definition (no implementations)
- add related UT cases with testify


src/pkg/plug/scanner/dao/scanner/model.go
```diff
@@ -0,0 +1,120 @@
...
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+// Registration represents a named configuration for invoking a scanner via its adapter.
+// UUID will be used to track the scanner.Endpoint as unique ID
+type Registration struct {
```
heww: Will `Scanner` be better than `Registration`?

src/core/api/plug_scanners.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+
+	query := &q.Query{
+		PageSize:   pz,
+		PageNumber: p,
+	}
+
+	// Get query key words
+	kws := make(map[string]string)
+	kw := sa.GetString("q")
```
heww: query by `q` is not friendly for swagger

src/core/api/plug_scanners.go
```diff
@@ -0,0 +1,350 @@
...
+	if len(kw) > 0 {
+		ws := strings.Split(kw, ",")
```
heww: what will happen if value itself contains `,`?

src/pkg/plug/scanner/dao/scanner/model.go
```diff
@@ -0,0 +1,120 @@
...
+	// Basic information
+	// int64 ID is kept for being aligned with previous DB schema
+	ID          int64  `orm:"pk;auto;column(id)" json:"-"`
+	UUID        string `orm:"unique;column(uid)" json:"uid"`
```
heww: will `uuid` be better than `uid` for `uid` often short for `user id`?

src/pkg/plug/scanner/dao/scanner/model.go
```diff
@@ -0,0 +1,120 @@
...
+	Name        string `orm:"unique;column(name);size(128)" json:"name"`
+	Description string `orm:"column(description);null;size(1024)" json:"description"`
+	URL         string `orm:"column(url);unique;size(512)" json:"url"`
+	Disabled    bool   `orm:"column(disabled);default(true)" json:"disabled"`
+	IsDefault   bool   `orm:"column(is_default);default(false)" json:"isDefault"`
```
heww: maybe `json:"is_default"` for other apis all use the snake case in the response

make/migrations/postgresql/0011_1.10.0_schema.up.sql
```diff
@@ -0,0 +1,35 @@
...
+    name VARCHAR(128) UNIQUE NOT NULL,
+    description VARCHAR(1024) NULL,
+    auth VARCHAR(16) NOT NULL,
+    access_cred VARCHAR(512) NULL,
+    adapter VARCHAR(128) NOT NULL,
+    vendor VARCHAR(128) NOT NULL,
+    version VARCHAR(32) NOT NULL,
+    disabled BOOLEAN NOT NULL DEFAULT FALSE,
+    is_default BOOLEAN NOT NULL DEFAULT FALSE,
+    skip_cert_verify BOOLEAN NOT NULL DEFAULT TRUE,
```
reasonerjt: It has to be false, i.e. by default it should always verify the cert.

src/core/api/plug_scanners_test.go
```diff
@@ -0,0 +1,384 @@
...
+// TearDownTest clears test case env
+func (suite *ScannerAPITestSuite) TearDownTest() {
+	// List
+	l := make([]*scanner.Registration, 0)
+	err := handleAndParse(&testingRequest{
+		url:        rootRoute,
+		method:     http.MethodGet,
+		credential: sysAdmin,
+	}, &l)
+	require.NoError(suite.T(), err)
```
heww: Suite object has assertion methods, maybe `suite.NoError(err)` will be better.

src/core/api/plug_scanners_test.go
```diff
@@ -0,0 +1,384 @@
...
+		url:        fmt.Sprintf("%s/%s", rootRoute, uid),
+		method:     http.MethodPut,
+		credential: sysAdmin,
+		bodyJSON:   r,
+	}, rr)
+	require.NoError(suite.T(), err)
+	require.NotNil(suite.T(), rr)
+
+	assert.Equal(suite.T(), "apiTestUpdate", rr.Name)
+	assert.Equal(suite.T(), true, rr.Disabled)
```
heww: Suite object has assertion methods, `suite.Equal(true, rr.Disabled)`

make/migrations/postgresql/0011_1.10.0_schema.up.sql
```diff
@@ -0,0 +1,35 @@
...
+CREATE TABLE scanner_report
+(
+    id SERIAL PRIMARY KEY NOT NULL,
+    digest VARCHAR(256) NOT NULL,
+    endpoint_id VARCHAR(64) NOT NULL,
+    vendor VARCHAR(32) NOT NULL,
+    job_id VARCHAR(32),
+    status VARCHAR(16) NOT NULL,
+    status_code INTEGER DEFAULT 0,
+    report JSONB,
```
reasonerjt: What's the reason to choose JSONB over `text` or `JSON` do we want to index the value in this column?

make/migrations/postgresql/0011_1.10.0_schema.up.sql
```diff
@@ -0,0 +1,35 @@
...
+    create_time TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+    update_time TIMESTAMP DEFAULT CURRENT_TIMESTAMP
+);
+
+/*Table for keeping the scanner report. The report details are stored as JSONB*/
+CREATE TABLE scanner_report
+(
+    id SERIAL PRIMARY KEY NOT NULL,
+    digest VARCHAR(256) NOT NULL,
+    endpoint_id VARCHAR(64) NOT NULL,
```
reasonerjt: does it map to the `uid` in `scanner_registration` table?

make/migrations/postgresql/0011_1.10.0_schema.up.sql
```diff
@@ -0,0 +1,35 @@
...
+    vendor VARCHAR(32) NOT NULL,
```
reasonerjt: is 32 enough?

make/migrations/postgresql/0011_1.10.0_schema.up.sql
```diff
@@ -0,0 +1,35 @@
...
+    job_id VARCHAR(32),
+    status VARCHAR(16) NOT NULL,
```
reasonerjt: this `status` is NOT NULL `job_id` should also be NOT NULL.

src/core/api/plug_scanners.go
```diff
@@ -0,0 +1,350 @@
...
+		sa.SendNotFoundError(errors.Errorf("scanner registration: %s", uid))
+		return
+	}
+
+	sa.Data["json"] = deleted
+	sa.ServeJSON()
+}
+
+// SetAsDefault sets the given registration as default one
+func (sa *ScannerAPI) SetAsDefault() {
```
heww: Do we need a special API to set the scanner as default? This function can be done by update scanner API.

src/core/api/plug_scanners.go
```diff
@@ -0,0 +1,350 @@
...
+	// URL changed?
+	if r.URL != rr.URL {
+		if !sa.checkDuplicated("url", rr.URL) {
+			return
+		}
+	}
+
+	getChanges(r, rr)
+
+	if err := sa.c.UpdateRegistration(r); err != nil {
```
heww: What will happen when update a scanner with the true value of `isDefault`? Maybe there will be more than one default scanners.

src/core/api/plug_scanners.go
```diff
@@ -0,0 +1,350 @@
...
+		return
+	}
+
+	// Explicitly check if conflict
+	if !sa.checkDuplicated("name", r.Name) ||
+		!sa.checkDuplicated("url", r.URL) {
+		return
+	}
+
+	uid, err := sa.c.CreateRegistration(r)
```
heww: Maybe there will be more than one default scanners when call this API with the true value of `isDefault`.

src/pkg/plug/scanner/api/registration.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+// UpdateRegistration ...
+func (bc *basicController) UpdateRegistration(registration *dscanner.Registration) error {
+	return bc.manager.Update(registration)
+}
+
+// SetDefaultRegistration ...
+func (bc *basicController) DeleteRegistration(registrationUUID string) (*dscanner.Registration, error) {
+	registration, err := bc.manager.Get(registrationUUID)
```
heww: err not handle when get the registration got internal server error

src/pkg/plug/scanner/api/registration.go
```diff
@@ -0,0 +1,170 @@
...
+
+	// First, get it from the project metadata
+	m, err := bc.proMetaMgr.Get(projectID, proScannerMetaKey)
+	if err != nil {
+		return nil, errors.Wrap(err, "get project scanner")
+	}
+
+	if len(m) > 0 {
+		if registrationID, ok := m[proScannerMetaKey]; ok && len(registrationID) > 0 {
+			registration, err := bc.manager.Get(registrationID)
```
heww: When registration of registrationID deleted in the system, maybe return the default one?

make/migrations/postgresql/0011_1.10.0_schema.up.sql
```diff
@@ -0,0 +1,35 @@
+/*Table for keeping the plug scanner registration*/
+CREATE TABLE scanner_registration
+(
+    id SERIAL PRIMARY KEY NOT NULL,
+    uid VARCHAR(64) UNIQUE NOT NULL,
```
heww: Can we use the native uuid type of postgres here?

make/migrations/postgresql/0011_1.10.0_schema.up.sql
```diff
@@ -0,0 +1,35 @@
...
+    url VARCHAR(256) UNIQUE NOT NULL,
+    name VARCHAR(128) UNIQUE NOT NULL,
+    description VARCHAR(1024) NULL,
+    auth VARCHAR(16) NOT NULL,
+    access_cred VARCHAR(512) NULL,
+    adapter VARCHAR(128) NOT NULL,
```
reasonerjt: As for `adapter` `vendor` `version`, which can be retrieved via metadata API of the scanner adapter, we may not want to store them in DB, in out-of-tree scenario, the scanner adapter has independent lifecycle of Harbor, so it can be tricky when they are upgraded or switched to other scanner

src/pkg/plug/scanner/dao/scanner/model.go
reasonerjt: +1

src/core/router.go
```diff
@@ -192,6 +192,13 @@ func initRouters() {
 		beego.Router("/api/chartrepo/:repo/charts/:name/:version/labels/:id([0-9]+)", chartLabelAPIType, "delete:RemoveLabel")
 	}
 
+	// Add routes for plugin scanner management
+	scannerAPI := &api.ScannerAPI{}
+	beego.Router("/api/scanners", scannerAPI, "post:Create;get:List")
+	beego.Router("/api/scanners/:uid", scannerAPI, "get:Get;delete:Delete;put:Update;patch:SetAsDefault")
```
reasonerjt: Why introduce both `put` and `patch`?

src/core/api/harborapi_test.go
```diff
@@ -206,6 +206,13 @@ func init() {
 	beego.Router("/api/internal/switchquota", &InternalAPI{}, "put:SwitchQuota")
 	beego.Router("/api/internal/syncquota", &InternalAPI{}, "post:SyncQuota")
 
+	// Add routes for plugin scanner management
+	scannerAPI := &ScannerAPI{}
+	beego.Router("/api/scanners", scannerAPI, "post:Create;get:List")
```
reasonerjt: conventionally such resource should better be put under `/api/system/` domain as it's system setting?

src/pkg/plug/scanner/api/controller.go
```diff
@@ -0,0 +1,157 @@
+// Copyright Project Harbor Authors
```
reasonerjt: what's `plug` in package path stands for? it doesn't make much sense as the code here is not per plugin, it's the unified shared code.  

Additionally,  there has been `/pkg/scan/` we probably should consider put all scan-related stuff under one sub-dir

make/migrations/postgresql/0011_1.10.0_schema.up.sql
steven-zou: In case we have the query requirement in future

make/migrations/postgresql/0011_1.10.0_schema.up.sql
steven-zou: `endpoint_id` should be changed to `registration_id`

make/migrations/postgresql/0011_1.10.0_schema.up.sql
steven-zou: I think so. If you're afraid of the risk, I can change it to `128`

make/migrations/postgresql/0011_1.10.0_schema.up.sql
steven-zou: `job_id` is updated after job launching, not at the time when creating this record. So it should not add `NOT NULL` here.

src/core/api/plug_scanners.go
steven-zou: Yes, we need. There is only one default scanner in the system. We need to update the specified one as default and mark the previous default one to not default. More clear.

This method follows the design proposal.

src/core/api/plug_scanners.go
steven-zou: Default can only be set by the default API, the update API will not pick up that property.

src/core/api/plug_scanners.go
steven-zou: Correct, add guard code here. All the newly created registration should be none default ones.

src/pkg/plug/scanner/api/registration.go
steven-zou: ok

src/pkg/plug/scanner/api/registration.go
steven-zou: Not get your point.

make/migrations/postgresql/0011_1.10.0_schema.up.sql
steven-zou: No, decouple with the database driver.

make/migrations/postgresql/0011_1.10.0_schema.up.sql
steven-zou: We have discussed this. I'd like to show a little more info to the users not just name and endpoint URL. This info is mostly like a cache of part of the metadata. We'll update it if there are any changes introduced when triggering scanner.

I'm planning to test the connection by calling the metadata API before launch the real job. Maybe the cache can be updated if there are changes.

src/core/router.go
steven-zou: Here `put` does a full updating.
`patch` does property piece updating.

src/pkg/plug/scanner/scan/controller.go
```diff
@@ -0,0 +1,48 @@
...
+import "github.com/goharbor/harbor/src/pkg/plug/scanner/dao/scan"
+
+// Options object for the scan action
+type Options struct{}
+
+// Option for scan action
+type Option interface {
+	// Apply option to the passing in options
+	Apply(options *Options) error
+}
```
heww: Maybe function will simpler than interface?  

https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

https://github.com/goharbor/harbor/blob/master/src/core/middlewares/interceptor/quota/options.go#L25

src/core/api/harborapi_test.go
steven-zou: I suggest using explicit API paths, not adding too many context fragments in the path. `system` is a context subpath, not related to the resources themselves.

src/pkg/plug/scanner/scan/controller.go
steven-zou: I'm totally fine with function options. Seems writing a function is much easier than implementing an interface.

src/pkg/plug/scanner/api/controller.go
steven-zou: I'm ok. I'll move the packages to`/scan` and unify the related definitions if necessary.

make/migrations/postgresql/0011_1.10.0_schema.up.sql
steven-zou: Agree.

src/pkg/plug/scanner/dao/scanner/model.go
steven-zou: ok

src/pkg/plug/scanner/dao/scanner/model.go
steven-zou: agree

src/pkg/plug/scanner/dao/scanner/model.go
steven-zou: Let's keep using `Registration` which is proposed in the design proposal

src/core/api/plug_scanners.go
steven-zou: The design idea is support as many properties as possible and not need to extract and specify the concreted property names. Like fuzzy query.

src/core/api/plug_scanners.go
steven-zou: good point.

src/core/api/plug_scanners.go
heww: But this can be done in update API. When the instance updated and `IsDefault` field changed from false to true, we can update all records' whose `IsDefault` value are true exclude the record currently update and change their `IsDefault` from true to false.

This can be implemented in the manager or DAO of the scanner if it's sure that there is only one default scanner in the system.

src/core/api/harborapi_test.go
heww: agree `/api/scanners`

src/pkg/plug/scanner/api/registration.go
heww: eg, for project `library` set the scanner `clair` before, then system administrator delete the scanner, this function will return `nil` scanner.

src/pkg/plug/scanner/dao/scanner/model.go
heww: ok

src/core/api/plug_scanners_test.go
steven-zou: Reply together with the above comment, `require` will exit and assert will not. So, need different methods for the different cases.

Let's keep them as they were.

src/core/api/plug_scanners.go
steven-zou: changed

make/migrations/postgresql/0011_1.10.0_schema.up.sql
steven-zou: JSON

src/core/api/plug_scanners.go
steven-zou: > But this can be done in update API. When the instance updated and `IsDefault` field changed from false to true, we can update all records' whose `IsDefault` value are true exclude the record currently update and change their `IsDefault` from true to false.
> 
> This can be implemented in the manager or DAO of the scanner if it's sure that there is only one default scanner in the system.

What you described is done in the separate method and API. Meanwhile, this method is proposed by the design proposal and I'm fine with it. So let's keep it as it is now.

src/core/api/plug_scanners.go
steven-zou: It is not possible to update `IsDefault` via the update API.

src/core/api/plug_scanners.go
steven-zou: added

src/pkg/plug/scanner/api/registration.go
steven-zou: handled

src/pkg/plug/scanner/api/registration.go
steven-zou: Actually that's the logic of this method. Seems there still are some gaps. Will enhance the code to be more strong.

steven-zou:
```go
if len(m) > 0 {
		if registrationID, ok := m[proScannerMetaKey]; ok && len(registrationID) > 0 {
			registration, err := bc.manager.Get(registrationID)
			if err != nil {
				return nil, errors.Wrap(err, "get project scanner")
			}

			if registration == nil {
				// Not found
				// Might be deleted by the admin, the project scanner ID reference should be cleared
				if err := bc.proMetaMgr.Delete(projectID, proScannerMetaKey); err != nil {
					return nil, errors.Wrap(err, "get project scanner")
				}
			} else {
				return registration, nil
			}
		}
	}```

File: make/migrations/postgresql/0011_1.10.0_schema.up.sql (hunk 19 above)
Comment (steven-zou) [reply make/migrations/postgresql/0011_1.10.0_schema.up.sql]: Keep it as it is now.

File: src/pkg/plug/scanner/api/controller.go (hunk 22 above)
Comment (steven-zou) [reply src/pkg/plug/scanner/api/controller.go]: Further package path refactoring will be performed after code completion.

File: src/pkg/plug/scanner/scan/controller.go (hunk 23 above)
Comment (steven-zou) [reply src/pkg/plug/scanner/scan/controller.go]: I'll change it in the implementation PR later. Will not do any changes to the current interface definitions.

File: src/pkg/scan/scanner/api/registration.go [hunk 24]
```diff
@@ -0,0 +1,194 @@
...
+
+// ListRegistrations ...
+func (bc *basicController) ListRegistrations(query *q.Query) ([]*scanner.Registration, error) {
+	return bc.manager.List(query)
+}
+
+// CreateRegistration ...
+func (bc *basicController) CreateRegistration(registration *scanner.Registration) (string, error) {
+	// TODO: Get metadata from the adapter service first
+	l, err := bc.manager.List(nil)
```
heww: This call will be very expensive when there are many registrations in the system, suggest to list the registrations limit to one.

steven-zou: Actually, the registrations will not be on a large scale "state". Anyway, I'll take care of it. As it's a perf issue, if no other obvious issues, I'll change it in the later PR.

steven-zou:
```go
// Check if there is any registrations already existing.
	l, err := bc.manager.List(&q.Query{
		PageSize:   1,
		PageNumber: 1,
	})
	if err != nil {
		return "", err
	}```

File: src/pkg/scan/scanner/dao/scan/report.go [hunk 25]
```diff
@@ -0,0 +1,43 @@
...
+package scan
+
+import "time"
+
+// Report of the scan
+// Identified by the `digest` and `endpoint_id`
+type Report struct {
+	ID               int64     `orm:"pk;auto;column(id)"`
+	Digest           string    `orm:"column(digest)"`
+	ReregistrationID string    `orm:"column(registration_id)"`
```
heww: `ReregistrationID` here is confused, when I see this I think it is the id column of `Registration` which type is int64, but the real reference column is `uuid` of `Registration`.

Btw, there is a spelling mistake for `ReregistrationID`.

make/migrations/postgresql/0011_1.10.0_schema.up.sql
reasonerjt: After offline discussion, we agreed such dynamic info should not be tracked in Harbor DB.

make/migrations/postgresql/0011_1.10.0_schema.up.sql
reasonerjt: ok

src/core/api/harborapi_test.go
reasonerjt: OK since it's a 2:1, I'm moving on with disagreement.

src/pkg/scan/scanner/dao/scan/report.go
steven-zou:
```go

// Report of the scan
// Identified by the `digest` and `endpoint_id`
type Report struct {
	ID               int64     `orm:"pk;auto;column(id)"`
	Digest           string    `orm:"column(digest)"`
	RegistrationUUID string    `orm:"column(registration_uuid)"`
	JobID            string    `orm:"column(job_id)"`
	Status           string    `orm:"column(status)"`
	StatusCode       int       `orm:"column(status_code)"`
	Report           string    `orm:"column(report);type(json)"`
	StartTime        time.Time `orm:"column(start_time);auto_now_add;type(datetime)"`
	EndTime          time.Time `orm:"column(end_time);type(datetime)"`
}
```

Tracked. Submitted by the follow-up PR

make/migrations/postgresql/0011_1.10.0_schema.up.sql
steven-zou: YES

PR #9142, opened 2019-09-18
Support OIDC groups

reasonerjt:
This commit enable project admin to add group as project member when
Harbor is configured against OIDC as AuthN backend.

It populates the information of groups from ID Token based on the claim
that is set in OIDC settings.


src/core/controllers/oidc.go
```diff
@@ -114,6 +116,7 @@ func (oc *OIDCController) Callback() {
 		oc.SendInternalServerError(err)
 		return
 	}
+	d.GroupIDs, err = group.GetGroupIDByGroupName(oidc.GroupsFromToken(idToken), common.OIDCGroupType)
```
stonezdj: There is an err, it should be handled, or it will be overwrite by next line

reasonerjt: I'll just log this error but set empty list in the model.

src/core/controllers/oidc.go
```diff
@@ -175,6 +182,7 @@ func (oc *OIDCController) Onboard() {
 		oc.SendBadRequestError(errors.New("Failed to get OIDC user info from session"))
 		return
 	}
+	defer oc.DelSession(userInfoKey)
```
wy65701436: in the previous implementation, line 223 will not delete the session.  But in this PR, it will.

reasonerjt: Will fix

PR #9154, opened 2019-09-19
[stage2]support pluggable scanner

steven-zou:
[stage2]support the pluggable scanner

- add scanner rest API v1 spec
- implement v1 client which is used to talk to scanner adapter
- adjust data/orm models
- adjust code package structure
- implement scan client which is used to talk to scanner adapter
- implement scan job which take the work of communicating with scanner
- update scanner mgmt API routes
- add corresponding UT cases
- change error collection in scan job
- add dead client checking in client pool
- change key word type to interface{} for q.Query
- update bearer authorizer


src/pkg/scan/rest/auth/auth.go
```diff
@@ -0,0 +1,54 @@
...
+	APIKey = "X-ScannerAdapter-API-Key"
+)
+
+// Authorizer defines operation for authorizing the requests
+type Authorizer interface {
+	Authorize(req *http.Request) error
+}
+
+// GetAuthorizer is a factory method for getting an authorizer based on the given auth type
+func GetAuthorizer(auth, cred string) (Authorizer, error) {
```
heww: I do not think we need so many authorizers, such `apiKeyAuthorizer`, `basicAuthorizer` and `bearerAuthorizer` and there are many common code in these implements.

```go
type defaultAuthorizer struct {
    prefix string
    value string
}
```

`defaultAuthorizer` (maybe other name) can cover apikey, base, bearer scenarios.

For basic mode, the value can be encoded as base64 before the creating of the authorizer.

The validation of `cred` can be done in GetAuthorizer, currently the behave to validate the `cred` is not same in these three authorizers.

src/pkg/scan/rest/auth/api_key_auth.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+// apiKeyAuthorizer authorize by adding a header `X-ScannerAdapter-API-Key` with value "credential"
+type apiKeyAuthorizer struct {
+	typeID     string
+	accessCred string
+}
+
+// Authorize the requests
+func (aa *apiKeyAuthorizer) Authorize(req *http.Request) error {
```
heww: `accessCred` is validated

src/pkg/scan/rest/auth/basic_auth.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+// basicAuthorizer authorizes the request by adding `Authorization Basic base64(credential)` header
+type basicAuthorizer struct {
+	typeID     string
+	accessCred string
+}
+
+// Authorize requests
+func (ba *basicAuthorizer) Authorize(req *http.Request) error {
```
heww: `accessCred` is validated

src/pkg/scan/rest/auth/bearer_auth.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+// bearerAuthorizer authorizes the request by adding `Authorization Bearer credential` header
+type bearerAuthorizer struct {
+	typeID     string
+	accessCred string
+}
+
+// Authorize requests
+func (ba *bearerAuthorizer) Authorize(req *http.Request) error {
```
heww: `accessCred` is not validated

src/pkg/scan/rest/v1/spec.go
```diff
@@ -0,0 +1,109 @@
...
+	MimeTypeScanRequest = "application/vnd.scanner.adapter.scan.request+json; version=1.0"
+	// MimeTypeScanResponse defines the mime type for scan response
+	MimeTypeScanResponse = "application/vnd.scanner.adapter.scan.response+json; version=1.0"
+)
+
+// RequestResolver is a function template to modify the API request, e.g: add headers
+type RequestResolver func(req *http.Request)
+
+// Definition for API
+type Definition struct {
```
heww: Do we really need `Definition` here?  In the client, some methods call `Resolver` but some methods do not, this is very strange.

src/pkg/scan/rest/v1/spec.go
```diff
@@ -0,0 +1,109 @@
...
+
+// Spec of the API
+// Contains URL and possible headers.
+type Spec struct {
+	baseRoute string
+}
+
+// NewSpec news V1 spec
+func NewSpec(base string) *Spec {
+	baseRoute := "http://localhost"
```
heww: Will the default base route be accessible in the client? If it is not accessible, I think we do need to provide the default base route here.

src/pkg/scan/rest/v1/client_pool.go
```diff
@@ -0,0 +1,128 @@
...
+
+		// Cache it
+		bcp.pool.Store(k, nc)
+		c = nc
+	}
+
+	return c.(Client), nil
+}
+
+func key(r *scanner.Registration) string {
```
heww: What about use a prefix and uuid as the cache key which will use less cpu than encode string to base64?

And the cache invalid problem for the update of the registration, we can save the client and registration update time in the cache (sync.Map). The after fetch the client from the cache, compare the update time between the cache and registration, recreate new client and save to cache when they are not equal.

When use the base64 string as cache key, old client will never cleanup after registration info updated.

src/pkg/scan/rest/auth/bearer_auth.go
steven-zou: NOT?

```
f req != nil && len(ba.accessCred) > 0 {
		req.Header.Add(authorization, fmt.Sprintf("%s %s", ba.typeID, ba.accessCred))
	}
```

src/pkg/scan/rest/v1/spec.go
steven-zou: So far, all the methods have their resolvers. Even in the future, there are some methods not need resolvers, a nil resolver can also be ok. It is explicitly aligned with the API spec and easy for using.

src/pkg/scan/rest/auth/auth.go
steven-zou: The design idea is aligned with the `Auth` property in the registration model. With auth factory, it is easy to get one by that property. And explicitly declaring the different kinds of authorizer will more flexible and clear (Although some duplicated code generated).

Moreover, it is easier for API consumer (like the portal API ) to consume and mapping the data.

It is a design pattern, and I don't think your proposed merged authorizer is better than the current one.

For the cred validation, it should be done by the authorizer themselves. They may have different requirements. (At present, the no authorizer will not validate the cred).

src/pkg/scan/rest/v1/spec.go
steven-zou: we can remove this one.

src/pkg/scan/rest/v1/client_pool.go
steven-zou: We can remove BASE64 encoding for key generating. The commented code can be reintroduced for the registration expiration

```
// func (bcp *basicClientPool) deadCheck(key string, item *poolItem) {
//	// Run in a separate goroutine
//	go func() {
//		// As we do not have a global context, let's watch the system signal to
//		// exit the goroutine correctly.
//		sig := make(chan os.Signal, 1)
//		signal.Notify(sig, os.Interrupt, syscall.SIGTERM, os.Kill)
//
//		tk := time.NewTicker(bcp.config.DeadCheckInterval)
//		defer tk.Stop()
//
//		for {
//			select {
//			case t := <-tk.C:
//				if item.timestamp.Add(bcp.config.ExpireTime).Before(t.UTC()) {
//					// Expired
//					bcp.pool.Delete(key)
//					return
//				}
//			case <-sig:
//				// Terminated by system
//				return
//			}
//		}
//	}()
// }
```

src/pkg/scan/job.go
```diff
@@ -0,0 +1,337 @@
...
+	}
+
+	// Exit signal
+	exit := make(chan bool, 1)
+	// Done signal
+	done := make(chan bool, 1)
+	// Collect errors
+	eChan := make(chan error, 1)
+
+	go func() {
```
heww: It's heavy to get the errors from get report goroutines.

src/pkg/scan/rest/auth/auth.go
heww: ok

src/pkg/scan/rest/auth/bearer_auth.go
steven-zou: Correctly, I'll fix it.

src/pkg/scan/job.go
steven-zou: Agree.

src/pkg/scan/rest/auth/bearer_auth.go
steven-zou: validated now!

src/pkg/scan/rest/v1/spec.go
steven-zou: changed to:

```go
s := &Spec{}

	if len(base) > 0 {
		if strings.HasSuffix(base, "/") {
			s.baseRoute = base[:len(base)-1]
		} else {
			s.baseRoute = base
		}
	}

	return s
}
```

PR #9192, opened 2019-09-22
return err when replication schema v1 image

### LINKED ISSUE #9170: Pull based replication not support schema 1 image

I want config harbor replication a old registry, and the old registry have some schema 1 images, all schema 1 images replication failed.

try pull based replication a schema 1 image

**Versions:**

- harbor version: 1.8.3
- docker engine version: 19.03.2

**Additional context:**

```
2019-09-19T15:27:31Z [INFO] [/replication/transfer/image/transfer.go:331]: pushing the manifest of image base/dns_checker:v1 ...
2019-09-19T15:27:31Z [ERROR] [/replication/transfer/image/transfer.go:339]: failed to push manifest of image base/dns_checker:v1: http error: code 500, message {"errors":[{"code":"UNKNOWN","message":"unknown error","detail":{}}]}
2019-09-19T15:27:31Z [ERROR] [/replication/transfer/image/transfer.go:143]: http error: code 500, message {"errors":[{"code":"UNKNOWN","message":"unknown error","detail":{}}]}
```

liuxu623:
Fixes #9170

### DIFF
--- a/src/replication/transfer/image/transfer.go
+++ b/src/replication/transfer/image/transfer.go
@@ -284,34 +284,35 @@ func (t *transfer) handleManifest(manifest distribution.Manifest, repository, di
 		t.logger.Errorf("failed to call the payload method for manifest of %s:%s: %v", repository, digest, err)
 		return nil, "", err
 	}
-	// manifest
-	if mediaType == schema1.MediaTypeManifest ||
-		mediaType == schema1.MediaTypeSignedManifest ||
-		mediaType == schema2.MediaTypeManifest {
+	switch mediaType {
+	case schema1.MediaTypeManifest, schema1.MediaTypeSignedManifest:
+		return nil, "", errors.New("schema v1 image is not supported")
+	case schema2.MediaTypeManifest:
 		return manifest, digest, nil
-	}
-	// manifest list
-	t.logger.Info("trying abstract a manifest from the manifest list...")
-	manifestlist, ok := manifest.(*manifestlist.DeserializedManifestList)
-	if !ok {
-		err := fmt.Errorf("the object isn't a DeserializedManifestList")
-		t.logger.Errorf(err.Error())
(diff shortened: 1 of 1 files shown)


chlins:
@liuxu623 Thanks, How about squash your commits and check the ci error ?

bitsf:
schema v1 is seldom used now, especially we're going to implement the OCI spec now, and this change may also break quota function, as it will decode manifest to get blob size info.
suggest not support such case now, it will increase code maintain cost

liuxu623:
> schema v1 is seldom used now, especially we're going to implement the OCI spec now, and this change may also break quota function, as it will decode manifest to get blob size info.
> suggest not support such case now, it will increase code maintain cost

or should we skip replication schema v1 image ?

xaleeks:
like @bitsf said, this is not happening since schema 1 is almost completely deprecated

bitsf:
@liuxu623  If skipped, you will not know that some images fail replicate from UI event log, I think current behavior is not quite a problem.

liuxu623:
> @liuxu623 If skipped, you will not know that some images fail replicate from UI event log, I think current behavior is not quite a problem.

```
/ / copy contents between the source and destination registries
	switch manifest.(type) {
	case *schema1.SignedManifest:
		return errors.New("skip copy schema1 image")
	default:
		for _, content := range manifest.References() {
			if err = t.copyContent(content, srcRepo, dstRepo); err != nil {
				return err
			}
		}
	}
```

bitsf:
I think the error message maybe something like "schema v1 image is not supported"

@xaleeks @ywk253100 what do you think, if we need change this now ?

xaleeks:
@bitsf @ywk253100  please check and merge if ok

bitsf:
this fix is conflict with the OCI implement, hold on to see how to deal with this.

Issue #8717, opened 2019-08-16
Tag Retention: unwanted deletion behavior


xaleeks:
There's an issue with the current retention policy implementation (due to native docker tag deletion behavior) where if you have multiple tags referring to the same sha digest and a subset of the tags are deleted due to a configured retention policy, all remaining tags will also get wiped out which in turn violates the retention policy
eg.
imagine following tags (listed by order of push time) all refer to the same digest

harbor-1.8,  		pushed 8/14/2019 12:00am
harbor-release, 	pushed 8/14/2019 03:00am
harbor-nightly, 	pushed 8/14/2019 06:00am
harbor-latest,		pushed 8/14/2019 09:00am

And there's retention policy configured to retain latest 2 tags matching 'harbor-*', so that 'harbor-rc' and 'harbor-latest' will be deleted, which triggers the deletion the other 2 tags 'harbor-1.8' and 'harbor-release' as well

This is due to the underlying issue highlighted here:
https://github.com/goharbor/harbor/issues/4868
https://github.com/goharbor/harbor/issues/6994

I don't characterize this as a bug (of tag retention feature) but it's definitely an unwanted side effect.  There's a soft solution in ignoring the deletion altogether in this case but that's also not what the user wanted.

yusuf-gunaydin:
A solution might be overwriting the given tag with a dummy image and then deleting the digest of that dummy image. I have used this workaround in the past quite a few times.

Unfortunately Docker team does not want to solve this issue as its open for years.

xaleeks:
We either need to implement a simple workaround like @yusuf-gunaydin  suggested or a full blown solution of deletable Harbor tags in the next release 1.11.  If we go with the workaround, need to think about where this dummy manifest sits as to minimize the overhead associated with each delete operation, all deletes can leverage this single lightweight dummy manifest.  @steven-zou @wy65701436 

In the meantime, I'm already working with @stuclem to address this in the docs for tag retention.

MartinEmrich:
This one just bit me hard... I expected a behaviour similar to Linux/UNIX hard links. This breaks our current cleanup procedure for older images/tags.
And I could almost swear that the behaviour with harbor < 1.8 was different?

xaleeks:
@MartinEmrich this will be fixed in upcoming release 1.11 since we will solve the underlying issue of being able to delete image tags

xaleeks:
[...]

PR #9215, opened 2019-09-23
Documenting tag retention issue

stuclem:
@xaleeks do we want to add the workaround from https://github.com/goharbor/harbor/issues/8717#issuecomment-528235134?

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -904,7 +904,7 @@ For example, you have following tags, listed according to their push time, and a
 - `harbor-nightly`, pushed 8/14/2019 06:00am
 - `harbor-latest`, pushed 8/14/2019 09:00am
 
-You configure a retention policy to retain the two latest tags that match `harbor-*`, so that `harbor-rc` and `harbor-latest` are deleted. However, since all tags refer to the same SHA digest, this also policy would also delete the tags `harbor-1.8` and `harbor-release`, so all tags are retained.
+You configure a retention policy to retain the two latest tags that match `harbor-*`, so that `harbor-rc` and `harbor-latest` are deleted. However, since all tags refer to the same SHA digest, this policy would also delete the tags `harbor-1.8` and `harbor-release`, so all tags are retained.
 
 ### Combining Rules on a Respository
 
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -895,7 +895,7 @@ In this case, all of the images in repository A are retained because they have b
 
 #### Tag Retention Rules and Native Docker Tag Deletion
 
-**WARNING**: Due to native Docker tag deletion behavior, there is an issue with the current retention policy implementation. If you have multiple tags that refer to the same SHA digest, and if a subset of these tags are deleted by a configured retention policy, all of the remaining tags are also deleted. This violates the retention policy. 
+**WARNING**: Due to native Docker tag deletion behavior, there is an issue with the current retention policy implementation. If you have multiple tags that refer to the same SHA digest, and if a subset of these tags are marked for deletion by a configured retention policy, all of the remaining tags would also be deleted. This violates the retention policy, so in this case all of the tags are retained. This issue will be addressed in a future update release, so that tag retention policies can delete tags without deleting the digest and other shared tags.
 
 For example, you have following tags, listed according to their push time, and all of them refer to the same SHA digest:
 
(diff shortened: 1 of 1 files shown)


docs/user_guide.md
```diff
@@ -893,6 +893,19 @@ This example uses the same project and repositories as example 1, but sets the r
...
+**WARNING**: Due to native Docker tag deletion behavior, there is an issue with the current retention policy implementation. If you have multiple tags that refer to the same SHA digest, and if a subset of these tags are deleted by a configured retention policy, all of the remaining tags are also deleted. This violates the retention policy. 
+
+For example, you have following tags, listed according to their push time, and all of them refer to the same SHA digest:
+
+- `harbor-1.8`, pushed 8/14/2019 12:00am
+- `harbor-release`, pushed 8/14/2019 03:00am
+- `harbor-nightly`, pushed 8/14/2019 06:00am
+- `harbor-latest`, pushed 8/14/2019 09:00am
+
+You configure a retention policy to retain the two latest tags that match `harbor-*`, so that `harbor-rc` and `harbor-latest` are deleted. However, since all tags refer to the same SHA digest, this also policy deletes the tags `harbor-1.8` and `harbor-release`.
```
xaleeks: Hi Stuart, right now actually none of them are deleted precisely because of this.  My ticket wasn't very clear on this.  So all of them are actually retained instead of deleted because delete can't be rolled back.  So we retain more than the policy should

xaleeks:
> @xaleeks do we want to add the workaround from [#8717 (comment)](https://github.com/goharbor/harbor/issues/8717#issuecomment-528235134)?

His workaround works but only for a manual delete.  It's impossible to do this if tag retention is deleting large bulks at scheduled intervals.  So leave it out for now, but maybe mention that this issue is being addressed, ie we are looking for a way to delete tags without deleting the digest and other shared tags.

xaleeks:
@reasonerjt is pto this week so please assign all docs related to @renmaosheng and we'll triage to the right person.  @stuclem

docs/user_guide.md
```diff
@@ -904,7 +904,7 @@ For example, you have following tags, listed according to their push time, and a
 - `harbor-nightly`, pushed 8/14/2019 06:00am
 - `harbor-latest`, pushed 8/14/2019 09:00am
 
-You configure a retention policy to retain the two latest tags that match `harbor-*`, so that `harbor-rc` and `harbor-latest` are deleted. However, since all tags refer to the same SHA digest, this also policy deletes the tags `harbor-1.8` and `harbor-release`.
+You configure a retention policy to retain the two latest tags that match `harbor-*`, so that `harbor-rc` and `harbor-latest` are deleted. However, since all tags refer to the same SHA digest, this also policy would also delete the tags `harbor-1.8` and `harbor-release`, so all tags are retained.
```
xaleeks: slight typo, should say 'this policy would also delete'

xaleeks: maybe 'so all tags are retained instead.' ?

xaleeks: lgtm

docs/user_guide.md
```diff
@@ -893,6 +893,19 @@ This example uses the same project and repositories as example 1, but sets the r
...
+**WARNING**: Due to native Docker tag deletion behavior, there is an issue with the current retention policy implementation. If you have multiple tags that refer to the same SHA digest, and if a subset of these tags are marked for deletion by a configured retention policy, all of the remaining tags would also be deleted. This violates the retention policy, so in this case all of the tags are retained. This issue will be addressed in a future update release, so that tag retention policies can delete tags without deleting the digest and other shared tags.
+
+For example, you have following tags, listed according to their push time, and all of them refer to the same SHA digest:
+
+- `harbor-1.8`, pushed 8/14/2019 12:00am
+- `harbor-release`, pushed 8/14/2019 03:00am
+- `harbor-nightly`, pushed 8/14/2019 06:00am
+- `harbor-latest`, pushed 8/14/2019 09:00am
+
+You configure a retention policy to retain the two latest tags that match `harbor-*`, so that `harbor-rc` and `harbor-latest` are deleted. However, since all tags refer to the same SHA digest, this policy would also delete the tags `harbor-1.8` and `harbor-release`, so all tags are retained.
```
michmike: one important thing to document is how does a user identify this situation from the logs? meaning how does Harbor let the user know that we did not delete a tag that was marked for retention due to this situation?

stuclem: @michmike good question. @xaleeks do you have the answer? Thanks!

Issue #9186, opened 2019-09-20
Harbor Core is failing at quota sync after update to 1.9


mkjoerg:
If you are reporting a problem, please make sure the following information are provided:

**Expected behavior and actual behavior:**
Harbor is upgraded by a helm upgrade to the newest version (1.9) from (1.8) and runs smooth.

**Steps to reproduce the problem:**
helm upgrade with values updated to 1.9

**Versions:**
Please specify the versions of following systems.

- harbor version: 1.9.0
- Kubernetes version: 1.15.3

**Additional context:**

Maybe some data in the registry is corrupted? We are using an external Postgres database and everything else is provisioned by harbor. Database is looking fine. Registry is connected to S3

```sh
NAME                                         READY   STATUS             RESTARTS   AGE
acid-harbor-cluster-0                        2/2     Running            0          111m
acid-harbor-cluster-1                        2/2     Running            0          117m
harbor-harbor-chartmuseum-6644d76bdd-8nq5h   1/1     Running            0          32m
harbor-harbor-chartmuseum-6644d76bdd-j7jxf   1/1     Running            0          31m
harbor-harbor-chartmuseum-6644d76bdd-qpbs8   1/1     Running            0          31m
harbor-harbor-clair-7bf7d7cb76-gr9st         1/1     Running            0          31m
harbor-harbor-clair-7bf7d7cb76-zglhm         1/1     Running            0          32m
harbor-harbor-core-7c75b55f68-97p6c          0/1     CrashLoopBackOff   4          3m7s
harbor-harbor-jobservice-689b758974-dn277    0/1     CrashLoopBackOff   13         32m
harbor-harbor-jobservice-749779f679-mlck2    0/1     CrashLoopBackOff   13         38m
harbor-harbor-portal-b97448bc6-6s2n6         1/1     Running            0          32m
harbor-harbor-portal-b97448bc6-g8xfc         1/1     Running            0          31m
harbor-harbor-redis-0                        1/1     Running            0          31m
harbor-harbor-registry-6c7989df76-5w8rn      2/2     Running            0          3m54s
```

```log
# Harbor Core Logs
 logs -f harbor-harbor-core-7c75b55f68-97p6c
2019-09-20T16:57:51Z [INFO] [/replication/adapter/native/adapter.go:44]: the factory for adapter docker-registry registered
2019-09-20T16:57:51Z [INFO] [/replication/adapter/harbor/adapter.go:42]: the factory for adapter harbor registered
[...]

PR #9221, opened 2019-09-24
patch registry fix of issue 2553

wy65701436:
This commit is target to fix harbor issue #9186, which root cause is mentioned by
docker/distribution#2553, and fixed by docker/distribution#2879.

As the latest distribution release(v2.7.1) does not contain this fix, but it will break the quota migraion process on S3 storage, we have to path this fix into Harbor regsitry binary.

[Tag Version]
It uses the issue number(2553) as the tag naming convention, like v2.7.1-patch-2553, means that we patch the fix of issue 2553 into v2.7.1.

[Note]
So far, this fix is only targets on docker regsitry v2.7.1. If the registry has this fix in new release, we'll move on.


make/photon/registry/builder
```diff
@@ -22,6 +22,13 @@ cur=$PWD
 TEMP=`mktemp -d /$TMPDIR/distribution.XXXXXX`
 git clone -b $VERSION https://github.com/docker/distribution.git $TEMP
 
+# add patch 2879
```
renmaosheng: is it 2553 or 2879? it seems difference from the REGISTRYVERSION above.

wy65701436: 2879 is the pull request number,
2553 is the github issue number, we will the issue number as tag, it's better for user to get more detail and background.

reasonerjt: Can you explain this in the comment?  Plus, let's also clarify this is only needed for registry 271

reasonerjt:
The change in registry looks good to me, let's add some comment and merge this PR

make/photon/registry/builder
wy65701436: added in the commit-msg.

PR #9238, opened 2019-09-24
Implement an unified task manager

ywk253100:
This commit introduces an unified task manager to handle the job related operations to reduce the complexity, such as stopping and status updating


make/migrations/postgresql/0015_1.10.0_schema.up.sql
```diff
@@ -6,4 +6,26 @@ CREATE TABLE immutable_tag_rule
...
+
+CREATE TABLE task
+(
+  id SERIAL PRIMARY KEY NOT NULL,
+  job_id varchar(64),
+  status varchar(16),
+  status_code int,
+  status_revision int,
+  message varchar(1024),
+  option varchar(256),
```
steven-zou: add `group id`

src/pkg/task/controller.go
```diff
@@ -0,0 +1,233 @@
...
+// All Jobservice job related operations should be covered by the controller to reduce
+// the complexity
+type Controller interface {
+	// Submit creates the task record in database and submits a job to the Jobservice
+	Submit(job *model.Job, option ...*model.Option) (int64, error)
+	Get(id int64) (*model.Task, error)
+	Stop(id int64) error
+	Delete(id int64) error
+	GetLog(id int64) ([]byte, error)
+	HandleHook(id int64, sc *job.StatusChange) error
```
steven-zou: move out `HandleHook `

src/pkg/task/controller.go
```diff
@@ -0,0 +1,233 @@
...
+	// Calculate the status of a group of tasks specified by the ids.
+	// If the status is a final status, an end time is returned either
+	CalculateTaskGroupStatus(ids []int64) (string, time.Time, error)
```
steven-zou: change the argument to group id and return a rich data model with overall status and stats info

src/pkg/task/model/task.go
```diff
@@ -0,0 +1,125 @@
...
+	"time"
+
+	"github.com/goharbor/harbor/src/jobservice/job"
+	"github.com/goharbor/harbor/src/pkg/task/dao"
+)
+
+// Task is the object mapping to the job of Jobservice, it holds
+// status and check in data
+type Task struct {
+	ID     int64  `json:"id"`
```
steven-zou: try to define a base model and let the dao model and api model inherit from it

src/pkg/task/model/task.go
```diff
@@ -0,0 +1,125 @@
...
+
+// Job object submitted to Jobservice
+type Job struct {
+	Name       string         `json:"name"`
+	Parameters job.Parameters `json:"parameters"`
+	Metadata   *job.Metadata  `json:"metadata"`
+}
+
+// Option for Job
+type Option struct {
```
steven-zou: rename to `Options`

```
type Option func(options *Options) error

func CheckInDataMode(append bool) Option {
   return func(options *Options) error{
        options. AppendCheckInData = append
  }
}
```

PR #9256, opened 2019-09-25
make/photon/prepare: Respect directories defined in g

g-braeunlich:
* Respect g.data_dir in utils.jobservice
* Add g.compose_location (to be used in g.config_file_path and in utils.docker_compose.docker_compose_yml_path)

Signed-off-by: Gerhard Bräunlich <g.braeunlich@disroot.org>


g-braeunlich:
What about this?
Still think this is usefull

g-braeunlich:
> What about this?
> Still think this is usefull

Same again

PR #9257, opened 2019-09-25
make/photon/prepare: Use yaml.safe_load instead of yaml.load

g-braeunlich:
See: https://seclists.org/oss-sec/2018/q2/240

Signed-off-by: Gerhard Bräunlich <g.braeunlich@disroot.org>

### DIFF
--- a/make/photon/prepare/utils/configs.py
+++ b/make/photon/prepare/utils/configs.py
@@ -53,7 +53,7 @@ def parse_versions():
     if not versions_file_path.is_file():
         return {}
     with open('versions') as f:
-        versions = yaml.load(f)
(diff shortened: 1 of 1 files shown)


heww:
currently pyyaml version 4.2b1 is used by prepare, according to the oss-sec, `load` is safe to use.

g-braeunlich:
OK. True. But still I do not see any reason to use load. According to the docs,
safe_load recognizes only standard YAML tags and cannot construct an arbitrary Python object in contrast to load.
See https://pyyaml.org/wiki/PyYAMLDocumentation

ninjadq:
closed because `load` is safe in the current version.

PR #9274, opened 2019-09-26
refactor immutable dao code to align the new structure under pkg

wy65701436:
1, add manager
2, move model dao to /pkg/dao


src/pkg/immutabletag/dao/model/rule.go
```diff
@@ -1,4 +1,4 @@
-package models
+package model
```
ywk253100: Move this into immutabletag/model package if the manager and dao use the same one

src/pkg/immutabletag/model/rule.go
```diff
@@ -0,0 +1,64 @@
...
+	Kind string `json:"kind" valid:"Required;Match(doublestar)"`
+
+	// Decorated the selector
+	// for "doublestar" : "matching" and "excluding"
+	// for "label" : "with" and "without"
+	Decoration string `json:"decoration" valid:"Required"`
+
+	// Param for the selector
+	Pattern string `json:"pattern" valid:"Required"`
+}
```
stonezdj: There is a similar Metadata struct in src/pkg/retention/policy/rule/models.go, should we reuse this struct?

src/pkg/immutabletag/manager.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func (drm *defaultRuleManager) UpdateImmutableRule(projectID int64, ir *model.Metadata) (int64, error) {
+	daoRule := &dao_model.ImmutableRule{}
+	data, _ := json.Marshal(ir)
+	daoRule.TagFilter = string(data)
+	return drm.dao.UpdateImmutableRule(projectID, daoRule)
+}
+
+func (drm *defaultRuleManager) EnableImmutableRule(id int64, enabled bool) (int64, error) {
```
stonezdj: the name is confusing? If I want to disable any of rule, I still have to call EnableImmutableRule()

src/pkg/immutabletag/controller.go
```diff
@@ -0,0 +1,70 @@
...
+	// UpdateImmutableRule ...
+	UpdateImmutableRule(pid int64, m *model.Metadata) error
+
+	// ListImmutableRules ...
+	ListImmutableRules(pid int64) ([]model.Metadata, error)
+}
+
+// DefaultAPIController ...
+type DefaultAPIController struct {
+	manager Manager
```
stonezdj: The controller just delegate the request for manager. do we need controller for all modules?

src/pkg/immutabletag/model/rule.go
wy65701436: discussed with SZ, we will not share any model between components under ./pkg..

src/pkg/immutabletag/dao/model/rule.go
wy65701436: it's for dao only

Issue #9174, opened 2019-09-20
upgrade v1.8.3-v1.9.0 error


liuxin638507:
[root@harbor data]#  docker run -it --rm -v /mnt/harbor/harbor.yml:/harbor-migration/harbor-cfg/harbor.cfg -v /data/harbor/harbor.yml:/harbor-migration/harbor-cfg-out/harbor.yml goharbor/harbor-migrator:v1.9.0 --cfg up
Please backup before upgrade, 
Enter y to continue updating or n to abort: y
The path of the migrated harbor.cfg is not set, the input file will be overwritten.
Command for config file migration: python ./cfg/run.py --input /harbor-migration/harbor-cfg/harbor.cfg
input version: 1.8.0, migrator chain: ['1.9.0']
migrating to version 1.9.0
Traceback (most recent call last):
  File "./cfg/run.py", line 70, in <module>
    main()
  File "./cfg/run.py", line 46, in main
    m.migrate(curr_input_path, curr_output_path)
  File "/harbor-migration/cfg/migrator_1_9_0/__init__.py", line 21, in migrate
    f.write(tpl.render(**config_dict))
  File "/usr/lib/python2.7/site-packages/jinja2/environment.py", line 1008, in render
    return self.environment.handle_exception(exc_info, True)
  File "/usr/lib/python2.7/site-packages/jinja2/environment.py", line 780, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/harbor-migration/cfg/migrator_1_9_0/harbor.yml.jinja", line 122, in top-level template code
    absolute_url: {{ chart.absolute_url if chart.absolute_url == 'enabled' else 'disabled' }}
  File "/usr/lib/python2.7/site-packages/jinja2/environment.py", line 430, in getattr
    return getattr(obj, attribute)
jinja2.exceptions.UndefinedError: 'chart' is undefined

MrFloppy:
I have the same error

kedarnok:
Same here when upgrading from 1.7.5

PualCN:
same here where upgrading from 1.7.4

corganfuzz:
same when upgrading from 1.7.0

deep-cds:
same here where upgrading from 1.7.4

thatsk:
same here where upgrading from 1.7.5

thatsk:
might be issue with instruction

thatsk:
Hello guies you can this following image for migration.
[...]

PR #9276, opened 2019-09-26
#9174 config migration fails with python error: jinja2.exceptions.UndefinedError: 'chart' is undefined

das-peter:
Possible fix for: https://github.com/goharbor/harbor/issues/9174
As far as I'm aware chart is only available if the stack was started with `./install.sh --with-chartmuseum` so I guess it's fair to add a check if the property is available.


ninjadq:
thanks for your pr, please sign you name on git commit.
``` bash
git commit --amend --signoff
```
 Because DCO is not pass

ninjadq:
hi, @das-peter could you open another PR to cherry-pick this to `release-1.9.0` branch. many thanks

PR #9297, opened 2019-09-28
Fixes #9090 - Custom logo for portal loaded from correct place

### LINKED ISSUE #9090: Custom images for Portal look & feel loaded from wrong directory
If you modify `src/portal/src/setting.json` (following [the look & feel doc](https://github.com/goharbor/harbor/blob/master/docs/customize_look%26feel_guide.md)) and add custom images to the folder `src/portal/src/images` (as advised in the document), building a custom portal image will result in the web interface being broken because the custom images are loaded from `/static/images/<Image Name>` instead of `/images/` (where the `src/portal/src/images` folder is located in the final image).

- Customize the portal via `setting.json` by adding a custom image to the `images` folder and referencing it in the config file.
- Rebuild the portal image.
- Open it in a browser => custom background and logo images will be broken.
- Bonus: Replace `/static/images` with `/images` (e.g. with browser dev tools) and see that the images have been indeed included in the custom container image.

**Versions:**

- harbor version: 1.9.0-rc2
- docker engine version: 17.12.1-ce
- docker-compose version: -

**Additional context:**

- Seems to be related to these two lines:
    - https://github.com/goharbor/harbor/blob/v1.9.0-rc2/src/portal/src/app/sign-in/sign-in.component.html#L5
    - https://github.com/goharbor/harbor/blob/v1.9.0-rc2/src/portal/src/app/base/navigator/navigator.component.html#L6

cadnce:
Hopefully this sorts it in a way you're happy with.

It did make me wonder that if a custom logo was provided, should that actually override the harbor one rather than appearing as well as?

### DIFF
--- a/src/portal/src/app/base/navigator/navigator.component.html
+++ b/src/portal/src/app/base/navigator/navigator.component.html
@@ -3,7 +3,7 @@
         <a href="javascript:void(0)" class="nav-link" (click)="homeAction()">
             <!-- <clr-icon shape="vm-bug"  *ngIf="!customStyle?.headerLogo"></clr-icon> -->
             <img [src]="'images/harbor-logo.svg'" class="harbor-logo" />
-            <img [attr.src]="'static/images/'+customStyle?.headerLogo" *ngIf="customStyle?.headerLogo" class="headerLogo">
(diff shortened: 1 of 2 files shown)


cadnce:
I guess you decided to re implement this and fix in #9332 instead.

PR #9298, opened 2019-09-28
Fixes #5446 - allows you to change project name in top left

### LINKED ISSUE #5446: Changing "Harbor" text in top left corner possible?
I want to change the text in the top left corner from "Harbor" to something else.

I'm aware I can change the look and feel via the `setting.json` file, as [documented here](https://github.com/vmware/harbor/blob/master/docs/customize_look%26feel_guide.md), but alas the file in the documentation and the [actual file](https://github.com/vmware/harbor/blob/master/src/ui_ng/src/setting.json) differ. (E.g. `product` is now `project`, different entries, ...)

The file in the documentation has a `name` entry that contains `Harbor`, while the actual file is missing this. I tried changing all the values in the actual file, and while the changes work (e.g. the text above the login form, the logo, etc.), none of these entries change the "Harbor" text in the top left corner. I also tried adding `"name": "<Harbor changed>"` into various sections of the `setting.json` file, but it's never applied either.

Can this be changed somewhere else? Or has this functionality been removed? Why are the files in the documentation and the file in the source code different? Am I missing something? Is there a workaround for me in the meantime? :confused: 

The easiest way would be to have a dedicated entry for this in the `setting.json` file.

---
PS: 
Also, please add an option to remove the ™ symbol after the text above the login form, i.e. the value in `appTitle`. Or remove it by default if the value is changed from `VMware Harbor`. But that's not as important right now.

cadnce:
Looks like this change got missed out when there was some refactoring done to setting.json

https://github.com/goharbor/harbor/commit/7845405862a1f762d35b60768cdad039fba96e99

Hopefully this works for you.

### DIFF
--- a/src/portal/src/app/base/navigator/navigator.component.ts
+++ b/src/portal/src/app/base/navigator/navigator.component.ts
@@ -61,8 +61,8 @@ export class NavigatorComponent implements OnInit {
         // custom skin
         let customSkinObj = this.skinableConfig.getSkinConfig();
         if (customSkinObj) {
-            if (customSkinObj.projects) {
-                this.customProjectName = customSkinObj.projects;
+            if (customSkinObj.project) {
+                this.customProjectName = customSkinObj.project;
             }
             this.customStyle = customSkinObj;
         }


cadnce:
I guess you decided to re implement this and fix in #9332 instead.

Issue #8422, opened 2019-07-26
[feature request]Add option to use username from oidc


annProg:
When login with oidc, user can change their username. This may cause confusion inside the company

**Describe the solution you'd like**
Add an option to use username from oidc. don't allow user to change it.

Hokwang:
Please refer to #8017 .

reasonerjt:
@annProg are you suggesting that we should add an option to disable user to update his username during onboard?

annProg:
> @annProg are you suggesting that we should add an option to disable user to update his username during onboard?

disable user to update his username when user first login with oidc. 

![image](https://user-images.githubusercontent.com/4145852/61935632-7e0d6e80-afbd-11e9-9bce-faa0bcf617f9.png)

reasonerjt:
@annProg 
This is essentially requiring disable user from setting his own username, I don't think it's a very reasonable requirement, esp.  bear in mind that there's no guarantee the username in the OIDC provider will be presented in the token or stay unchanged.  
If you want some unchanged attribute, you should use `sub` according to the spec:
https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims

And the sub is stored in Harbor's DB, so that's the only reliable way to trace who is who.

jlegrone:
This [came up in slack](https://cloud-native.slack.com/archives/CC1E09J6S/p1568667588010500) recently:

> I was wondering if there is a way to "pre-configure" an OIDC user. Ideally, I would like to bypass the popup on first sign-in asking a user for their harbor username. Is there a way to have Harbor use the username passed back in the OIDC response?

I have a feeling that it is a very common use case for organizations self-hosting Harbor to want to enforce some user naming convention which can be unified with existing systems and trusted to be correct. For example, an administrator granting push access to `jane.smith` should know that they're not actually giving access to someone who happened to be masquerading as Jane.

Certainly the OIDC subject + issuer should be used as a primary key. However many OIDC provider implementations don't supply subject in a human-readable format, and defaulting usernames to this would make for a very poor end user experience.

I see two possible approaches, but there are likely others as well.

1. Allow administrators to configure custom policy around acceptable usernames (ie. must equal the text before `@` character in email). We could provide a set of default conventions to choose from, or make this configurable with a scripting language like Lua or [open policy agent](https://www.openpolicyagent.org/).
[...]

PR #9311, opened 2019-10-01
Add options for automatic onboarding and username claim

airadier:
In our company, as well as #8422 and #9995, we need to skip the onboarding screen. Also, OIDC is integrated with ADFS (Active Directory Federation Services) and we need to get the username from a different claim name (sAMAccountName) instead of "name" claim. This PR addresses these 2 issues, allowing to select "Automatic onboarding" from the OIDC configuration and use a custom claim name for the username. 

- Add an option in the UI to enable or disable the automatic user onboarding
- Add an option to specify the claim name where the username is retrieved from.

Signed-off-by: Iradier, AlvaroJose <AlvaroJose.Iradier@adidas.com>

### DIFF
--- a/src/common/utils/oidc/helper.go
+++ b/src/common/utils/oidc/helper.go
@@ -321,11 +321,11 @@ func userInfoFromClaims(c claimsProvider, g, u string) (*UserInfo, error) {
 			return nil, err
 		}
 
-		if username, ok := allClaims[u].(string); !ok {
-			log.Warningf("OIDC. Failed to recover Username from claim. Claim '%s' is empty", u)
-		} else {
-			res.Username = username
+		username, ok := allClaims[u].(string)
+		if !ok {
+			return nil, fmt.Errorf("OIDC. Failed to recover Username from claim. Claim '%s' is invalid or not a string", u)
 		}
+		res.Username = username
 
 	}
 	res.Groups, res.hasGroupClaim = GroupsFromClaims(c, g)
--- a/src/common/utils/oidc/helper.go
+++ b/src/common/utils/oidc/helper.go
@@ -19,16 +19,17 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
-	gooidc "github.com/coreos/go-oidc"
-	"github.com/goharbor/harbor/src/common/models"
-	"github.com/goharbor/harbor/src/core/config"
-	"github.com/goharbor/harbor/src/lib/log"
-	"golang.org/x/oauth2"
 	"net/http"
 	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
+
+	gooidc "github.com/coreos/go-oidc"
+	"github.com/goharbor/harbor/src/common/models"
+	"github.com/goharbor/harbor/src/core/config"
+	"github.com/goharbor/harbor/src/lib/log"
+	"golang.org/x/oauth2"
 )
 
 const (
@@ -294,7 +295,7 @@ func userInfoFromRemote(ctx context.Context, token *Token, setting models.OIDCSe
 	if err != nil {
 		return nil, err
 	}
-	return userInfoFromClaims(u, setting.GroupsClaim)
+	return userInfoFromClaims(u, setting.GroupsClaim, setting.UserClaim)
 }
 
 func userInfoFromIDToken(ctx context.Context, token *Token, setting models.OIDCSetting) (*UserInfo, error) {
@@ -305,13 +306,27 @@ func userInfoFromIDToken(ctx context.Context, token *Token, setting models.OIDCS
 	if err != nil {
 		return nil, err
 	}
-	return userInfoFromClaims(idt, setting.GroupsClaim)
+
+	return userInfoFromClaims(idt, setting.GroupsClaim, setting.UserClaim)
 }
 
-func userInfoFromClaims(c claimsProvider, g string) (*UserInfo, error) {
+func userInfoFromClaims(c claimsProvider, g, u string) (*UserInfo, error) {
 	res := &UserInfo{}
 	if err := c.Claims(res); err != nil {
 		return nil, err
+	}
+	if u != "" {
+		allClaims := make(map[string]interface{})
+		if err := c.Claims(&allClaims); err != nil {
+			return nil, err
+		}
+
+		if username, ok := allClaims[u].(string); !ok {
+			log.Warningf("OIDC. Failed to recover Username from claim. Claim '%s' is empty", u)
+		} else {
+			res.Username = username
+		}
+
 	}
 	res.Groups, res.hasGroupClaim = GroupsFromClaims(c, g)
 	return res, nil
--- a/src/core/controllers/oidc.go
+++ b/src/core/controllers/oidc.go
@@ -17,9 +17,10 @@ package controllers
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/goharbor/harbor/src/common/dao/group"
 	"net/http"
 	"strings"
+
+	"github.com/goharbor/harbor/src/common/dao/group"
 
 	"github.com/goharbor/harbor/src/common"
 	"github.com/goharbor/harbor/src/common/dao"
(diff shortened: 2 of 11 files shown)


reasonerjt:
@airadier  thanks for the PR 

But the "auto onboarding" flow seems a bit opinionate, I personally don't think it should be merged to master.

airadier:
Hi @reasonerjt, although it can seem opinionated, the fact is that we are using SAML or OIDC integrations in some of our tools (we expect to migrate every tool in the future), and this is the first time I see an onboarding screen instead of a configuration option to grab the username from some specific attribute or claim.

From an enterprise point of view, allowing the user to onboard a choose a custom username is messy. As commented on #8422 this is a common use case for many companies. This PR keeps the default behavior of having an onboarding step, but includes 2 additional options:

- The ability to choose a claim name other than "name" to grab the username from.
- The ability to totally skip the onboarding screen.

I would really like to see this merged, as in another case we will need to keep our own fork of the project to support this use case.

By the way, I am also following up on the group support for OIDC, as it is also a requirement for our company. I will be testing dev branch in the following weeks and helping with the implementation if required.

Regards.

reasonerjt:
@airadier
Thanks for your reply, I think it makes sense, I'll add a comment in #8422 and you can chime in and discuss with other community users about your design, if we reach agreement we can try to refine the PR and merge it.

Please note that the FC for 1.10 is pretty close (before end of Oct) so this change may be delayed to future release.

xaleeks:
don't merge this yet.  What does the screen look if opting out of setting own username?  are there diagrams?

xaleeks:
@airadier , Hi Alvaro, thanks for the PR, this is a slightly more opinionated approach than we like so we will have more discussions on if there's a more uniform solution.  Don't have bandwidth to test this fully for 1.10 so most likely will be in the following release.  Please see comments in https://github.com/goharbor/harbor/issues/8422.  In the meantime, I would recommend using the current workflow of entering the username manually on first login instead of maintaining a separate fork of this.

airadier:
Hi @xaleeks, @reasonerjt, there is no diagram for the new flow. If the "Automatic onboarding" option is set, the onboarding screen is just skipped. There are in fact 2 (3 if you also include UI changes) big changes in this PR:

- The first change is including the option (in the UI + mapping to the OIDC config in core) to set a claim name for grabbing the username. If this option is set, if will unmarshal the ID Token Json to a map[string]interface{} instead of using the JSON tags in the struct, then it will retrieve the username from the specific claim. This can work independently from the "Auto onboard" feature, so if this is enabled, the user will see the onboarding screen, but the preset username will be taken from this claim instead of the hardcoded one.
- If the "Auto onboard" option is checked, the login in the OIDC callback handler, when the user cannot be retrieved from the OpenID identifier, it will not redirect the user to the Onboard screen, but instead a new User is created (the previously described username login applies), and the onboard step is skipped.
- The user onboard/creation login has been extracted to a new *userOnboard* function that is called from both the controller OIDC callback and the *Onboard* function that is called from the Onboard screen, as the user onboard/creation is the same in both cases. The only difference is the username is provided by the user, or provided by the claim with no user intervention.

Regards!

xaleeks:
@reasonerjt I looked at his PR and I think it makes sense.  We just need to make sure that if the username claim on the oidc side coming through to violates some conditions (ie length, some disallowed characters etc), then there are warnings to the user to reconfigure on the OIDC side before trying again.  Another thing I prefer we just take it from the claim for everyone moving forward, so make the auto onboarding mandatory.  I think this use case is the majority, not the minority.  Please take a look and see if we can get it merged per a few modifications maybe.

renmaosheng:
Hi Alvaro,
       Sorry for the late reply, we are reviewing the code, found there are some conflicts with current files, could you please do a rebase of your changes? thanks.
Steven

airadier:
Sure, let me try to find some time this week[end].

Regards.

xaleeks:
@airadier  This is very useful but we have a few questions, we could jump on a zoom call to discuss if you're available.

airadier:
Sure, is it possible to do it during the weekend? I am in Spain, any time between 11AM-11PM CEST should be ok for me. Also Monday from 11AM-5PM should be ok. I'll try to review the changes because I made this PR some months ago for a different company, and now I don't remember the exact details :)

xaleeks:
let's work it out over email, can you please ping me at xalex@vmware.com ?  I'm sure we can work out a time amenable to you

airadier:
Hi, I rebased the PR on latest _master_ branch. 

Basically it adds two options to the configuration screen of OIDC:

image

The **Automatic onboarding** changes the flow to skip the onboard screen after first login through OIDC, where the user can modify its username. It will directly onboard the user without providing the option to edit the username.

The **OIDC Username Claim** option, if specified, overrides the default _claim_ which is used for the username field (currently it is mapped to "name", so for example in Azure it gets the full user name). If the option is blank, it doesn't alter the current behaviour. If set to anything, it will take the username from the specified _claim_ instead of from "name" _claim._ If the claim does not exist, a warning error is logged, and the username is still retrieved from the "name" _claim._

Both options can work independently.

The new _userOnboard_ function encapsulates the business logic of user onboarding, which was previously contained in the Onboard controller, as now the same logic can be used from the Onboard screen, or directly from the OIDC callback.

src/common/utils/oidc/helper.go
```diff
@@ -305,14 +306,28 @@ func userInfoFromIDToken(ctx context.Context, token *Token, setting models.OIDCS
...
 		return nil, err
 	}
+	if u != "" {
+		allClaims := make(map[string]interface{})
+		if err := c.Claims(&allClaims); err != nil {
+			return nil, err
+		}
+
+		if username, ok := allClaims[u].(string); !ok {
+			log.Warningf("OIDC. Failed to recover Username from claim. Claim '%s' is empty", u)
```
reasonerjt, line 325: It's also possible that claim is non-string.

Additionally, we'd better check the string is valid for a username, we may want to refactor and share the logic with this:
https://github.com/goharbor/harbor/blob/master/src/core/api/user.go#L598

src/core/controllers/oidc.go
```diff
@@ -122,30 +123,100 @@ func (oc *OIDCController) Callback() {
...
-		gids, err := group.PopulateGroup(models.UserGroupsFromName(info.Groups, common.OIDCGroupType))
-		if err != nil {
-			log.Warningf("Failed to populate groups, error: %v, user will have empty group list, username: %s", err, info.Username)
-		}
-		u.GroupIDs = gids
-		oidcUser, err := dao.GetOIDCUserByUserID(u.UserID)
-		if err != nil {
-			oc.SendInternalServerError(err)
+		// Recover the username from d.Username by default
+		username := info.Username
```
reasonerjt, line 134: `username` may be empty string here?

src/core/controllers/oidc.go
```diff
@@ -122,30 +123,100 @@ func (oc *OIDCController) Callback() {
...
+
+		// Fix blanks in username
+		username = strings.Replace(username, " ", "_", -1)
+
+		// If automatic onboard is enabled, skip the onboard page
+		if oidcSettings.AutoOnboard {
+			log.Debug("Doing automatic onboarding\n")
+			user, onboarded := userOnboard(oc, info, username, tokenBytes)
+			if onboarded == false {
```
reasonerjt, line 143: I don't think this is right, in this case the `/callback` will return 200 and nothing happens, we need to explicitly let user know something is wrong.

Maybe when the `info` does not have `username` or user failed to onboard, log the error and redirect to the onboard dialog so user can input the username and onboard?

src/core/controllers/oidc.go
```diff
@@ -122,30 +123,100 @@ func (oc *OIDCController) Callback() {
...
+	if err != nil {
+		if strings.Contains(err.Error(), dao.ErrDupUser.Error()) {
+			oc.RenderError(http.StatusConflict, "Conflict, the user with same username or email has been onboarded.")
+			return nil, false
 		}
-		oc.PopulateUserSession(*u)
-		oc.Controller.Redirect("/", http.StatusFound)
+
+		oc.SendInternalServerError(err)
+		oc.DelSession(userInfoKey)
```
reasonerjt, line 215: you have the userinfo removed in the `Onboard`  func so this seems not necessary

src/portal/src/i18n/lang/en-us-lang.json
```diff
@@ -899,6 +901,8 @@
             "CLIENTSECRET": "OIDC Client Secret",
             "SCOPE": "OIDC Scope",
             "OIDC_VERIFYCERT": "Verify Certificate",
+            "OIDC_AUTOONBOARD": "Automatic onboarding",
+            "USER_CLAIM": "OIDC Username Claim",
```
reasonerjt, line 905: I think we can just call it `Username Claim`

src/common/utils/oidc/helper.go
airadier: Hi @reasonerjt, in case the claim is not a string, it is not valid, the warning will be logged and the user won't be recovered (same case if the claim is not a string, does not exist, etc.). I might just change the warning message to tell that claim is empty or invalid.

About validating the user, I don't think this is the right moment to perform the user validation. No validation is done if the user is retrieved from the default claim, so I see no sense in doing the validation if it is retrieved from a customized claim. I thin the validation is performed later, when we try to onboard the user in https://github.com/goharbor/harbor/blob/426bce6ce4db66dde6e87ea3cf932798dea87174/src/core/controllers/oidc.go#L258, so the user is validated no matter if a custom user token is set, or the default token is used.

src/core/controllers/oidc.go
airadier: I'll test that use case again, but I think you get an error and onboard fails if the username is empty, which could both happen with a custom claim or with the default claim.

src/core/controllers/oidc.go
airadier: You are right, but I would prefer to use the oc.SendInternalServerError method instead. If you enable AutoOnboard, definitely you don't want to give the user the change to type a username. So in case the automatic onboarding fails I think informing with an HTTP 500 error makes more sense. Let me test this case and see what happens, then provide a better error handling.

src/core/controllers/oidc.go
airadier: Do you mean that oc.DelSession(userInfoKey) is already in Onboard func? It is, but only when the onboard is successful, so userOnboard returns (user, true). But if the onboard fails it won't be removed. But I think as this session key is used only for the "manual onboarding" screen, it makes sense to remove that from userOnboard func and remove that key in the Onboard func instead, whether the onboard fails to not. I'll make that change.

airadier: Fixed now. In fact, it can only be removed when the onboard is successful, so if there are errors and the username is not accepted (i.e. already in use), the information needs to be present in the session.

So now it is only removed in `Onboard` func after successful onboarding.

src/core/controllers/oidc.go
airadier: After some tests, the "userOnboard" function is already handling the error and displaying an error via oc.RenderError (for duplicate user) or just an Internal Server Error for other cases.

src/core/controllers/oidc.go
airadier: After checking, there is not much difference from the previous code. info.Username could be empty string if the default username claim is empty.

src/common/utils/oidc/helper.go
airadier: I made a change to trigger an error instead of just logging a Warning in the log. So in case the user configures an invalid username claim (claim does not exist, or it is not a string), an error is raised now.

I like the idea of adding some validation, but it shouldn't be at this point, but on the userOnboard function, and outside the scope of this patch. in fact, even without this patch, it is possible to type a username like " " (a single blank) in the onboard screen, and it will be accepted.

Vad1mo:
I really like this feature, I tested it with Auth0, and Azure AD B2C in both cases a provided username is used.

This is a much needed feature and I it is backward compatible and on top flexible enough to support different IDP settings. would be nice if string could be also expressed as path `custom_meta.domain.username`

airadier:
> I really like this feature, I tested it with Auth0, and Azure AD B2C in both cases a provided username is used.
> 
> This is a much needed feature and I it is backward compatible and on top flexible enough to support different IDP settings. would be nice if string could be also expressed as path `custom_meta.domain.username`

I guess we could use something like https://github.com/tidwall/gjson or improve the way we recover custom claims now (as we just unmarshal to a `map[string]interface{}`, so we would need to iterate on the interface{} to support nesting.

The same path settings should apply for the _groups_ claim. I could work on such a feature, but I'd prefer to do it in a separate PR once this gets merged.

Regards.

Vad1mo:
> The same path settings should apply for the groups claim. I could work on such a feature, but I'd prefer to do it in a separate PR once this gets merged.

agree the current state is already a big leap forward.

n0rig:
Looking forward to this :)

martinDDe:
> Looking forward to this :)

We all do! Do you have any estimates when it will finally arrive?

src/core/controllers/oidc.go
```diff
@@ -122,30 +123,99 @@ func (oc *OIDCController) Callback() {
...
+	if err != nil {
+		log.Warningf("Failed to populate groups, error: %v, user will have empty group list, username: %s", err, info.Username)
+	}
+	u.GroupIDs = gids
+	oidcUser, err := dao.GetOIDCUserByUserID(u.UserID)
+	if err != nil {
+		oc.SendInternalServerError(err)
+		return
+	}
+	_, t, err := secretAndToken(tokenBytes)
```
wy65701436, line 167: no error handling here?

src/core/controllers/oidc.go
```diff
@@ -122,30 +123,99 @@ func (oc *OIDCController) Callback() {
...
+	if err := dao.UpdateOIDCUser(oidcUser); err != nil {
+		oc.SendInternalServerError(err)
+		return
+	}
+	oc.PopulateUserSession(*u)
+	oc.Controller.Redirect("/", http.StatusFound)
+
+}
+
+func userOnboard(oc *OIDCController, info *oidc.UserInfo, username string, tokenBytes []byte) (*models.User, bool) {
```
wy65701436, line 178: we already have a func Onboard, can we have a better name to distinguish them?or at least add some comments here.

src/core/controllers/oidc.go
```diff
@@ -122,30 +123,99 @@ func (oc *OIDCController) Callback() {
...
-		}
-		u.GroupIDs = gids
-		oidcUser, err := dao.GetOIDCUserByUserID(u.UserID)
-		if err != nil {
-			oc.SendInternalServerError(err)
+		// Recover the username from d.Username by default
+		username := info.Username
+
+		// Fix blanks in username
+		username = strings.Replace(username, " ", "_", -1)
```
wy65701436, line 137: can we log the original name here?

src/core/controllers/oidc.go
```diff
@@ -122,30 +123,99 @@ func (oc *OIDCController) Callback() {
...
+
+		// If automatic onboard is enabled, skip the onboard page
+		if oidcSettings.AutoOnboard {
+			log.Debug("Doing automatic onboarding\n")
+			user, onboarded := userOnboard(oc, info, username, tokenBytes)
+			if onboarded == false {
+				log.Error("User not onboarded\n")
```
wy65701436, line 144: and it's better to just return err in the `userOnboard`, it can be seen as a until func. And define http return value at here?

src/core/controllers/oidc.go
```diff
@@ -122,30 +123,99 @@ func (oc *OIDCController) Callback() {
...
+				return
+			}
+			log.Debug("User automatically onboarded\n")
```
wy65701436, line 147: also log the `username`?

PR #9329, opened 2019-10-08
feat(replication): support for jfrog artifactory docker image replica…

chlins:

### DIFF
--- a/src/replication/adapter/jfrog/adapter.go
+++ b/src/replication/adapter/jfrog/adapter.go
@@ -1,0 +1,310 @@
+package jfrog
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"strconv"
+	"strings"
+
+	common_http "github.com/goharbor/harbor/src/common/http"
+	"github.com/goharbor/harbor/src/common/http/modifier"
+	"github.com/goharbor/harbor/src/common/utils/log"
+	"github.com/goharbor/harbor/src/common/utils/registry/auth"
+	adp "github.com/goharbor/harbor/src/replication/adapter"
+	"github.com/goharbor/harbor/src/replication/adapter/native"
+	"github.com/goharbor/harbor/src/replication/model"
+	"github.com/goharbor/harbor/src/replication/util"
+)
+
+func init() {
+	err := adp.RegisterFactory(model.RegistryTypeJfrogArtifactory, AdapterFactory)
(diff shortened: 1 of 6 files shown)


src/replication/adapter/jfrog/adapter.go
```diff
@@ -0,0 +1,312 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var (
+		modifiers = []modifier.Modifier{
+			&auth.UserAgentModifier{
+				UserAgent: adp.UserAgentReplication,
+			}}
+		authorizer modifier.Modifier
```
cd1989: There is no need to define this variable. How about 

```
if registry.Credential != nil {
		modifiers = append(modifiers, auth.NewBasicAuthCredential(
			registry.Credential.AccessKey,
			registry.Credential.AccessSecret))
	}
```

src/replication/adapter/jfrog/adapter.go
```diff
@@ -0,0 +1,312 @@
...
+			}
+		}
+	}
+
+	return nil
+}
+
+func (a *adapter) getLocalRepositories() ([]*repository, error) {
+	var repositories []*repository
+	url := fmt.Sprintf("%s/artifactory/api/repositories?type=local&packageType=docker", a.registry.URL)
```
cd1989: It will get all repositories, without pagination ?

src/replication/adapter/jfrog/adapter_test.go
```diff
@@ -0,0 +1,133 @@
...
+	"bytes"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	adp "github.com/goharbor/harbor/src/replication/adapter"
+	"github.com/goharbor/harbor/src/replication/model"
+	"github.com/stretchr/testify/assert"
+
```
cd1989: Suggest to remove this blank line.

src/replication/adapter/jfrog/adapter.go
chlins: I think it will get all，from offical api document, this interface just provide two query parameters(type and packageType), without pagination definition.

chlins: https://www.jfrog.com/confluence/display/RTF/Artifactory+REST+API#ArtifactoryRESTAPI-GetRepositories

src/replication/adapter/jfrog/adapter_test.go
chlins: thanks

PR #9378, opened 2019-10-11
add visible attribute to robot account

wy65701436:


src/pkg/robot/model/robot.go
```diff
@@ -24,6 +24,7 @@ type Robot struct {
 	ProjectID    int64     `orm:"column(project_id)" json:"project_id"`
 	ExpiresAt    int64     `orm:"column(expiresat)" json:"expires_at"`
 	Disabled     bool      `orm:"column(disabled)" json:"disabled"`
+	Visible      bool      `orm:"column(visible)" json:"visible"`
```
ywk253100: Seems the `visible` is invisible for users? It it is, how about set the json annotation to `-` and don't expose it to users

steven-zou: +1

PR #9393, opened 2019-10-14
add immutable match

wy65701436:


src/pkg/immutabletag/match/rule/match.go
```diff
@@ -0,0 +1,83 @@
...
+
+		// match repositories according to the repository selectors
+		var repositoryCandidates []*art.Candidate
+		for _, repositorySelector := range r.ScopeSelectors["repository"] {
+			selector, err := index.Get(repositorySelector.Kind, repositorySelector.Decoration,
+				repositorySelector.Pattern)
+			if err != nil {
+				return false, err
+			}
+			repositoryCandidates, err = selector.Select(cands)
```
ywk253100: Should use the output of the first selector as the input of next selector?

PR #9397, opened 2019-10-14
Disabled 2 scan e2e test cases

danfengliu:
2 e2e test cases should be disabled due to scan code was upgraded, but the way of comment them off is wrong, so correct it in this PR.

Signed-off-by: Danfeng Liu (c) <danfengl@vmware.com>


danfengliu:
It's no longer needed.

PR #9400, opened 2019-10-14
Inject certs to non root

ninjadq:
* Registry user to non-root
* Registryctl user to non-root
* Clair user to non-root
* Chartmuseum user to non-root

### DIFF
--- a/make/photon/chartserver/Dockerfile
+++ b/make/photon/chartserver/Dockerfile
@@ -2,21 +2,26 @@ FROM photon:2.0
 
 RUN tdnf install -y shadow sudo >>/dev/null\
     && tdnf clean all \
-    && mkdir /chartserver/ \
-    && mkdir /harbor/ \
-    && groupadd -r -g 10000 chartuser \
-    && useradd --no-log-init -m -r -g 10000 -u 10000 chartuser
-COPY ./make/photon/chartserver/binary/chartm /chartserver/
-COPY ./make/photon/chartserver/docker-entrypoint.sh /docker-entrypoint.sh
-COPY ./make/photon/common/install_cert.sh /harbor
+    && groupadd -r -g 10000 chart \
+    && useradd --no-log-init -m -g 10000 -u 10000 chart
+
+COPY ./make/photon/chartserver/binary/chartm /home/chart/
+COPY ./make/photon/chartserver/docker-entrypoint.sh /home/chart/
+COPY ./make/photon/common/install_cert.sh /home/chart/
+
+RUN chmod -R 777 /etc/pki/tls/certs \
+    && chown -R chart:chart /home/chart \
+    && chmod u+x /home/chart/chartm \
+    && chmod u+x /home/chart/docker-entrypoint.sh \
+    && chmod u+x /home/chart/install_cert.sh
+
+USER chart
+
+WORKDIR /home/chart
+
+ENTRYPOINT ["./docker-entrypoint.sh"]
 
 VOLUME ["/chart_storage"]
 EXPOSE 9999
 
-RUN chown -R 10000:10000 /chartserver \
-    && chmod u+x /chartserver/chartm \
-    && chmod u+x /docker-entrypoint.sh
-
 HEALTHCHECK --interval=30s --timeout=10s --retries=3 CMD curl -sS 127.0.0.1:9999/health || exit 1
-
-ENTRYPOINT ["/docker-entrypoint.sh"]
--- a/make/photon/chartserver/Dockerfile
+++ b/make/photon/chartserver/Dockerfile
@@ -9,7 +9,7 @@ COPY ./make/photon/chartserver/binary/chartm /home/chart/
 COPY ./make/photon/chartserver/docker-entrypoint.sh /home/chart/
 COPY ./make/photon/common/install_cert.sh /home/chart/
 
-RUN chmod -R 777 /etc/pki/tls/certs \
+RUN chown -R chart:chart /etc/pki/tls/certs \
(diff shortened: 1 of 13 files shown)


make/photon/chartserver/Dockerfile
```diff
@@ -2,21 +2,26 @@ FROM photon:2.0
...
-COPY ./make/photon/chartserver/docker-entrypoint.sh /docker-entrypoint.sh
-COPY ./make/photon/common/install_cert.sh /harbor
+    && groupadd -r -g 10000 chart \
+    && useradd --no-log-init -m -g 10000 -u 10000 chart
+
+COPY ./make/photon/chartserver/binary/chartm /home/chart/
+COPY ./make/photon/chartserver/docker-entrypoint.sh /home/chart/
+COPY ./make/photon/common/install_cert.sh /home/chart/
+
+RUN chmod -R 777 /etc/pki/tls/certs \
```
cadnce: Is chmod 777 really the security conscious thing to be doing to the sensitive certs directory?

make/photon/clair/Dockerfile
```diff
@@ -2,24 +2,25 @@ FROM photon:2.0
...
 
 VOLUME /config
 
 EXPOSE 6060 6061
 
-RUN chown -R 10000:10000 /clair \
-    && chmod u+x /clair/clair \
-    && chmod u+x /docker-entrypoint.sh \
-    && chmod +x /dumb-init
+RUN chmod -R 777 /etc/pki/tls/certs \
```
cadnce: Again

make/photon/common/install_cert.sh
```diff
@@ -2,14 +2,14 @@
 
 set -e
 
-if [ ! -f /etc/pki/tls/certs/ca-bundle.crt.original ]; then
-    cp /etc/pki/tls/certs/ca-bundle.crt /etc/pki/tls/certs/ca-bundle.crt.original
+if [ ! -f ~/ca-bundle.crt.original ]; then
```
cadnce: Is this supposed to be "-/ca-bundle.crt" looks like a typo with the leading -

ninjadq: No, ~ means user's home directory

make/photon/chartserver/Dockerfile
ninjadq: 777 is not secure, so change the owner to another user

make/photon/common/install_cert.sh
cadnce: Nvm, wasn't wearing my glasses 😂

make/photon/chartserver/Dockerfile
```diff
@@ -2,21 +2,26 @@ FROM photon:2.0
 
 RUN tdnf install -y shadow sudo >>/dev/null\
```
reasonerjt: why do we still need sudo?

make/photon/chartserver/Dockerfile
```diff
@@ -2,21 +2,26 @@ FROM photon:2.0
...
+    && chown -R chart:chart /home/chart \
+    && chmod u+x /home/chart/chartm \
+    && chmod u+x /home/chart/docker-entrypoint.sh \
+    && chmod u+x /home/chart/install_cert.sh
+
+USER chart
+
+WORKDIR /home/chart
+
+ENTRYPOINT ["./docker-entrypoint.sh"]
```
reasonerjt: You are still calling sudo in this script

make/photon/clair/docker-entrypoint.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 set -e
 
-/harbor/install_cert.sh
-sudo -E -H -u \#10000 sh -c "/dumb-init -- /clair/clair -config /etc/clair/config.yaml $*"
+/home/clair/install_cert.sh
+/home/clair/dumb-init -- /home/clair/clair -config /etc/clair/config.yaml $*
```
reasonerjt: You have tested it right? not sure whether dumb-init does anything that requires higher privilege

make/photon/common/install_cert.sh
reasonerjt: you don't need the `~` as the workdir is the home directory.

make/photon/registryctl/start.sh
```diff
@@ -4,19 +4,10 @@ set -e
...
 
-if [ -d /storage ]; then
-    if ! stat -c '%u:%g' /storage | grep -q '10000:10000' ; then
-        # 10000 is the id of harbor user/group.
-        # Usually NFS Server does not allow changing owner of the export directory,
-        # so need to skip this step and requires NFS Server admin to set its owner to 10000.
-        chown 10000:10000 -R /storage
-    fi
-fi
+/home/harbor/install_cert.sh
```
reasonerjt: `install_cert.sh` has to make sure that the injected custom-ca crt can be read by user 10000 otherwise this script will fail

This comment applies to other entrypoint script that injects ca cert

make/photon/chartserver/Dockerfile
ninjadq: because some cmd will install with sudo

make/photon/clair/docker-entrypoint.sh
reasonerjt: resolve it as it is tested

make/photon/registryctl/start.sh
reasonerjt: Please add TODO to make sure `prepare` knows the directory on host.

PR #9402, opened 2019-10-14
fix duration caculation in package report

beautytiger:


reasonerjt:
@steven-zou we may want to cover this in UT...

steven-zou:
@beautytiger 

Thanks for the PR. Actually we have found the bug and made a fix already, see PR https://github.com/goharbor/harbor/pull/9528. To avoid conflicts, I'll close this PR. You can check later when the code is merged.

beautytiger:
> @beautytiger
> 
> Thanks for the PR. Actually we have found the bug and made a fix already, see PR #9528. To avoid conflicts, I'll close this PR. You can check later when the code is merged.

Got it, glad to hear that, I will close this.

PR #9413, opened 2019-10-15
Add new cli secret ui in profile

### LINKED ISSUE #9407: Allow user to set CLI secret on UI
1.  The UI work flow should adopt to the new API  to let user input a secret and `PUT` to backend, some checking has to be performed.
2.  UI should keep the generate secret feature, instead it should generate the secret at frontend and send it to backend.

jwangyangls:
fix #9407

### DIFF
--- a/src/portal/src/app/account/account-settings/account-settings-modal-service.service.ts
+++ b/src/portal/src/app/account/account-settings/account-settings-modal-service.service.ts
@@ -10,8 +10,8 @@ import { throwError as observableThrowError, Observable } from 'rxjs';
 export class AccountSettingsModalService {
 
   constructor(private http: HttpClient) { }
-  generateCli(userId): Observable<any> {
-    return this.http.post(`/api/users/${userId}/gen_cli_secret`, {}).pipe( map(response => response)
+  saveNewCli(userId, secretObj): Observable<any> {
+    return this.http.put(`/api/users/${userId}/cli_secret`, secretObj).pipe( map(response => response)
     , catchError(error => observableThrowError(error)));
   }
 }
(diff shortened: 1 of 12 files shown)


src/portal/src/app/account/account-settings/account-settings-modal-service.service.ts
```diff
@@ -10,8 +11,8 @@ import { throwError as observableThrowError, Observable } from 'rxjs';
 export class AccountSettingsModalService {
 
   constructor(private http: HttpClient) { }
-  generateCli(userId): Observable<any> {
-    return this.http.post(`/api/users/${userId}/gen_cli_secret`, {}).pipe( map(response => response)
+  saveNewCli(userId, secretObj): Observable<any> {
+    return this.http.put(`/api/users/${userId}/cli_secret`, secretObj, HTTP_JSON_OPTIONS).pipe( map(response => response)
```
AllForNothing: HTTP_JSON_OPTIONS  is no longer needed, since we are using angular8

src/portal/src/app/shared/shared.utils.ts
```diff
@@ -75,4 +75,30 @@ export const maintainUrlQueryParmas = function (uri: string, key: string, value:
...
 };
+/**
+  * the password or secret must longer than 8 chars with at least 1 uppercase letter, 1 lowercase letter and 1 number
+  * @param randomFlag
+  * @param min
+  * @param max
+  * @returns {string}
+  */
+ export function randomWord(randomFlag, min, max) {
+    let str = "0aA",
```
AllForNothing: "the password or secret must longer than 8 chars with at least 1 uppercase letter, 1 lowercase letter and 1 number":  all the random password should begin with "0aA"? if not,  this function can be improved.

PR #9415, opened 2019-10-15
do changes to let the vul policy check compatible with new framework

steven-zou:
- update the scan/scanner controller
- enhance the report summary generation
- do changes to the vulnerable handler
- remove the unused Clair related code
- add more UT cases
- update the scan webhook event
- drop the unused tables/index/triggers in SQL schema


make/migrations/postgresql/0015_1.10.0_schema.up.sql
```diff
@@ -44,4 +44,17 @@ CREATE TABLE immutable_tag_rule
   creation_time timestamp default CURRENT_TIMESTAMP
 );
 
-ALTER TABLE robot ADD COLUMN visible boolean DEFAULT true NOT NULL;
\ No newline at end of file
+ALTER TABLE robot ADD COLUMN visible boolean DEFAULT true NOT NULL;
+
+/** Drop the unused vul related tables **/
```
reasonerjt: Is this change intentionally put in this PR?

steven-zou: Yes, together with the deletions of DAO related files.

src/core/middlewares/vulnerable/handler.go
```diff
@@ -37,44 +41,100 @@ func New(next http.Handler) http.Handler {
...
+	// Invalid project ID
+	if wl.ProjectID == 0 {
+		err := errors.Errorf("project verification error: project %s", img.ProjectName)
+		vh.sendError(err, rw)
+
+		return
+	}
+
+	// Is scanner available?
+	yes, err := scanner.DefaultController.IsScannerAvailable(wl.ProjectID)
```
reasonerjt: Is it possible the scanner becomes unavailable after scanning?  In that case the policy should still take effect, correct?

src/core/middlewares/vulnerable/handler.go
```diff
@@ -37,44 +41,100 @@ func New(next http.Handler) http.Handler {
...
+		err = errors.Errorf("no scan report existing for the artifact: %s:%s@%s", img.Repository, img.Reference, img.Digest)
+		vh.sendError(err, rw)
+
+		return
+	}
+
+	summary := rawSummary.(*vuln.NativeReportSummary)
+
+	// Do judgement
+	if summary.Severity.Code() >= projectVulnerableSeverity.Code() {
```
reasonerjt: Here may be a problem, if the severity is `unknown` or "N/A", the `Code()` returns -1 and the image can be pulled, maybe we should see `unknown` or `N/A` a more severe vulnerability, even higher than `Critical`.

src/pkg/scan/vuln/summary.go
```diff
@@ -21,13 +21,14 @@ import (
...
-	Duration   int64                 `json:"duration"`
-	Summary    *VulnerabilitySummary `json:"summary"`
-	StartTime  time.Time             `json:"start_time"`
-	EndTime    time.Time             `json:"end_time"`
+	ReportID    string                `json:"report_id"`
+	ScanStatus  string                `json:"scan_status"`
+	Severity    Severity              `json:"severity"`
+	Duration    int64                 `json:"duration"`
+	Summary     *VulnerabilitySummary `json:"summary"`
+	CVEBypassed []string              `json:"-"`
```
reasonerjt: It may be even better to return the CVE-Bypassed in the API such that UI can highlight what vulnerabilities in the whitelist, but not a must for this PR.

src/pkg/scan/vuln/severity.go
```diff
@@ -37,3 +37,21 @@ const (
...
+	case Low:
+		return 1
+	case Medium:
+		return 2
+	case High:
+		return 3
+	case Critical:
+		return 4
+	default:
+		return -1
```
reasonerjt: See my comment regarding how we handle "Unknown"

src/pkg/scan/report/summary.go
```diff
@@ -24,33 +24,67 @@ import (
 	"github.com/pkg/errors"
 )
 
+// CVESet defines the CVE whitelist with a hash set way for easy query.
```
reasonerjt: `CVESet` should be moved to /pkg/scan or un-exported.

src/pkg/scan/report/summary.go
```diff
@@ -24,33 +24,67 @@ import (
...
 	}
 
-	return g(r)
+	return g(r, options...)
 }
 
 // SummaryGenerator is a func template which used to generated report
 // summary for relevant mime type.
-type SummaryGenerator func(r *scan.Report) (interface{}, error)
+type SummaryGenerator func(r *scan.Report, options ...Option) (interface{}, error)
```
reasonerjt: This is beyond the scope of this PR but why let this func return a `interface{}` rather than a summary?

src/core/service/notifications/registry/handler.go
```diff
@@ -158,13 +161,16 @@ func (n *NotificationHandler) Post() {
...
 
 			if autoScanEnabled(pro) {
-				last, err := clairdao.GetLastUpdate()
-				if err != nil {
-					log.Errorf("Failed to get last update from Clair DB, error: %v, the auto scan will be skipped.", err)
-				} else if last == 0 {
-					log.Infof("The Vulnerability data is not ready in Clair DB, the auto scan will be skipped, error %v", err)
-				} else if err := coreutils.TriggerImageScan(repository, tag); err != nil {
-					log.Warningf("Failed to scan image, repository: %s, tag: %s, error: %v", repository, tag, err)
+				artifact := &v1.Artifact{
```
reasonerjt: Will this Artifact be moved out of `/scan/***` package?

src/core/service/notifications/jobs/handler.go
```diff
@@ -92,27 +93,36 @@ func (h *Handler) Prepare() {
...
+			log.Error(errors.Wrap(err, "scan job hook handler: event publish"))
 		} else {
-			log.Errorf("failed to build image scanning event metadata: %v", err)
+			e := &event.Event{}
+			metaData := &event.ScanImageMetaData{
+				Artifact: req.Artifact,
+				Status:   h.status,
+			}
+
+			if err := e.Build(metaData); err == nil {
```
reasonerjt: This is trivial but generally we should avoid nested `if`
```
if err:= e.Build(); err!= nil {
    ..
    return
}
if err:= e.Publish(); err != nil {
    ... 
   return
}
```
is a better approach

src/core/service/notifications/registry/handler.go
steven-zou: I marked it as a TODO item, I think there will be a global `Artifact` model after OCI is ready. So far, we do not need to do the movement.

src/pkg/scan/report/summary.go
steven-zou: It's a design consideration. It is a func template for handling different format summary. So, we need a general interface.

src/pkg/scan/report/summary.go
```diff
@@ -24,33 +24,67 @@ import (
...
 
 // GenerateNativeSummary generates the report summary for the native report.
-func GenerateNativeSummary(r *scan.Report) (interface{}, error) {
+func GenerateNativeSummary(r *scan.Report, options ...Option) (interface{}, error) {
```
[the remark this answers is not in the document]
steven-zou: This method is not designed only for CVESet. CVESet may influence the summary result as it is filtered by CVESet. It is an optional case, so use options way is better.

src/pkg/scan/report/summary.go
steven-zou: I'd prefer to keep it is here because it's close to the WithCVEWhitelist options although it is also used in the vulnerable hander. It's ok to refer it with the package `report` because it will cause influences to the "report" summary.

src/pkg/scan/vuln/severity.go
steven-zou: Agree with this comment, let's assign it the highest code.

src/pkg/scan/vuln/summary.go
steven-zou: @reasonerjt 
You can raise an issue for this because it needs UI work. For the current case, it follows the previous behavior, only prints them in the log.

src/core/middlewares/vulnerable/handler.go
steven-zou: Of course, we can only do the check based on the current status. I think it's ok for the case you mentioned, there will be corresponding errors returned in the log.

src/pkg/scan/report/summary.go
steven-zou: `CVESet` here is also defined as an alias for the model `map[string]interface{}` defined in the whitelist for easy referring and appending utility method.

src/core/middlewares/vulnerable/handler.go
reasonerjt: But that means, if the scanner is down, the vulnerable images can be pulled even project admin sets the policy?

I don't think it's correct.

src/pkg/scan/report/summary.go
reasonerjt: CVESet is utility to represent the whitelist, we are confusing the boundary in terms of report .vs. whitelist

src/core/service/notifications/registry/handler.go
reasonerjt: I'll write up an issue to make sure it's done.

src/core/middlewares/vulnerable/handler.go
steven-zou: @reasonerjt 
You should check the code.

I think you misunderstood it. Obviously, the availability is not related to the health of the scanner.  Here I only check the scanner metadata.

I think this check can be removed. It can be depended on the get summary method.

src/pkg/scan/report/summary.go
steven-zou: @reasonerjt 

Actually `CVESet` can replace the general one `map[string]interface{}` in the whitelist package. 

If you care about the package, I can do the change to replace the original one. I did NOT want to change many packages originally.

src/core/service/notifications/registry/handler.go
steven-zou: Sure!

PR #9434, opened 2019-10-16
build(clair): internal clair adapter when install with clair

heww:
1. build support for clair adapter in makefile
2. clair-adapter support in prepare stage
3. create the internal clair adapter when install harbor with clair flag
4. modify test scripts and swagger doc to make the tests match the API changed about scan
5. fix some issues when call scan request to adapter 


make/photon/prepare/templates/core/env.jinja
```diff
@@ -36,6 +36,7 @@ CORE_URL={{core_url}}
 CORE_LOCAL_URL={{core_local_url}}
 JOBSERVICE_URL={{jobservice_url}}
 CLAIR_URL={{clair_url}}
+CLAIR_ADAPTER_URL={{clair_adapter_url}}
```
ywk253100: I think the CLAIR_DB_xxx are all useless?

ywk253100: The same to `CLAIR_URL `

heww: clair adapter need the url of the clair and harbor core need the url of the clair adapter

make/photon/clair-adapter/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+RUN tdnf install -y sudo >>/dev/null\
+    && tdnf clean all \
+    && mkdir /clair-adapter/ \
+    && groupadd -r -g 10000 clair-adapter \
+    && useradd --no-log-init -m -r -g 10000 -u 10000 clair-adapter
+
+COPY ./make/photon/clair-adapter/binary/harbor-scanner-clair /clair-adapter/clair-adapter
+
+RUN chown -R 10000:10000 /clair-adapter \
```
ywk253100: Should we run this container as non-root user?

heww: Yes

steven-zou:
General comment:

Please fill in meaningful info into the PR description. @heww

heww:
@steven-zou description updated for the pr

make/photon/clair-adapter/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+    && chmod u+x /clair-adapter/clair-adapter
+
+EXPOSE 8080
+
+HEALTHCHECK --interval=30s --timeout=10s --retries=3 CMD curl -sS 127.0.0.1:8080/healthy || exit 1
```
steven-zou: You should use `/metadata` API endpoint to do check. No healthy endpoint.

make/photon/clair-adapter/builder
```diff
@@ -0,0 +1,39 @@
...
+
+cd `dirname $0`
+cur=$PWD
+
+# the temp folder to store distribution source code...
+TEMP=`mktemp -d ${TMPDIR-/tmp}/clair-adapter.XXXXXX`
+git clone https://github.com/danielpacak/harbor-scanner-clair.git $TEMP
+cd $TEMP; git checkout $VERSION; cd -
+
+echo 'build the clair adapter binary bases on the golang:1.12...'
```
steven-zou: no const value defined for golang version?

src/core/main.go
```diff
@@ -146,6 +150,39 @@ func quotaSync() error {
 	return nil
 }
 
+func initClairAdapter(clairAdapterEndpoint string) error {
```
steven-zou: Move to the `/pkg/scan` package, `main` is not a good place for it. We should keep the `main` as simple as possible.

src/pkg/scan/api/scan/base_controller.go
```diff
@@ -102,6 +102,14 @@ func NewController() Controller {
 	}
 }
 
+func (bc *basicController) jobClient() cj.Client {
```
steven-zou: This is ok for now. Just heads up, I'll update it after the rebase.

make/photon/clair-adapter/Dockerfile
heww: there is `/healthy` in the clair adapter 
```go
probeRouter.Methods(http.MethodGet).Path("/healthy").HandlerFunc(handler.GetHealthy)
```

`/metadata` maybe need authorization if the adapter implemented the auth.

make/photon/clair-adapter/builder
heww: changed, the `golang:1.12.5` was used in builder stage.

src/core/main.go
heww: Moved

make/photon/clair-adapter/Dockerfile
steven-zou: It is not defined in the open API spec

heww: Yes, it is not in the open API spec, this is the image of clair adapter and `/healthy` is defined by clair adapter i think it is ok to use this for the health checking.

When there is another adapter eg named `foo` maybe its health checking endpoint is `/foo` and then the `HEALTHCHECK` in its docker image maybe `HEALTHCHECK --interval=30s --timeout=10s --retries=3 CMD curl -sS 127.0.0.1:8080/foo || exit 1`

This image care the clair adapter itself, not care the open API spect defined by harbor with other scanners.

steven-zou: Ok, but we need to make sure the `healthy` check endpoint will be always there.

src/pkg/scan/init.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/goharbor/harbor/src/common/utils/log"
+	"github.com/goharbor/harbor/src/pkg/q"
+	sc "github.com/goharbor/harbor/src/pkg/scan/api/scanner"
+	"github.com/goharbor/harbor/src/pkg/scan/dao/scanner"
+)
+
+// EnsureScanner ensure the scanner which specially name exists in the system
+func EnsureScanner(registration *scanner.Registration) error {
+	q := &q.Query{
+		Keywords: map[string]interface{}{"url": registration.URL},
```
steven-zou: Apppend `PageNumber` and `PageSize` for better performance?

heww: They are not needed here for url in unique in registration table.

PR #9452, opened 2019-10-17
All certs in /harbor_cust_certs will appended to ca_bundle

ninjadq:
Every file in specific dir `/harbor_cust_cert` will be appended to the end of the system ca_bundle

### DIFF
--- a/make/photon/common/install_cert.sh
+++ b/make/photon/common/install_cert.sh
@@ -2,16 +2,29 @@
 
 set -e
 
-if [ ! -f /etc/pki/tls/certs/ca-bundle.crt.original ]; then
-    cp /etc/pki/tls/certs/ca-bundle.crt /etc/pki/tls/certs/ca-bundle.crt.original
+if [ ! -f ~/ca-bundle.crt.original ]; then
+    cp /etc/pki/tls/certs/ca-bundle.crt ~/ca-bundle.crt.original
 fi
 
-if [ -f /harbor_cust_cert/custom-ca-bundle.crt ]; then
+cp ~/ca-bundle.crt.original /etc/pki/tls/certs/ca-bundle.crt
+
+if [ "$(ls -A /harbor_cust_cert)" ]; then
     if grep -q "Photon" /etc/lsb-release; then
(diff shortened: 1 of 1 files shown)


heww:
Does ca_bundle in storage_service still work after this PR?

make/photon/common/install_cert.sh
```diff
@@ -2,17 +2,21 @@
...
 
-if [ -f /harbor_cust_cert/custom-ca-bundle.crt ]; then
+cp ~/ca-bundle.crt.original /etc/pki/tls/certs/ca-bundle.crt
+
+if [ "$(ls -A /harbor_cust_cert)" ]; then
     if grep -q "Photon" /etc/lsb-release; then
-        echo "Appending custom ca bundle ..."
-        cp /etc/pki/tls/certs/ca-bundle.crt.original /etc/pki/tls/certs/ca-bundle.crt
-        cat /harbor_cust_cert/custom-ca-bundle.crt >> /etc/pki/tls/certs/ca-bundle.crt
+        for z in /harbor_cust_cert/*; do
```
reasonerjt: Could we add more sanity check  to `z`?
for example it has to be a file and it has to match some pattern?

make/photon/common/install_cert.sh
```diff
@@ -2,17 +2,30 @@
...
     if grep -q "Photon" /etc/lsb-release; then
-        echo "Appending custom ca bundle ..."
-        cp /etc/pki/tls/certs/ca-bundle.crt.original /etc/pki/tls/certs/ca-bundle.crt
-        cat /harbor_cust_cert/custom-ca-bundle.crt >> /etc/pki/tls/certs/ca-bundle.crt
-        echo "Done."
+        echo "Appending trust CA to ca-bundle ..."
+        for z in /harbor_cust_cert/*; do
+            case ${z} in
+                *.crt | *.ca | *.ca-bundle | *.pem)
+                    if [ -d "$z" ]; then
```
reasonerjt: I think use `-f` to check is a better approach.

PR #9497, opened 2019-10-21
Disable self registration by default (cherry-pick)

stonezdj:
Update default self_registration=false


reasonerjt:
@stonezdj for this break change , please make sure open an issue for awareness of doc writer.

Issue #9326, opened 2019-10-07
Searching nested groups in wrong basedn


martinbalint:
**Expected behavior and actual behavior:**
Actual behavior:
"Searching for nested groups" searches in LDAP Base DN.
This is very slow in our environment, search query "`(&(objectClass=group)(member:1.2.840.113556.1.4.1941:=CN=e-Martin.Balint,OU=Users,OU=xxx,DC=yyy,DC=zzz,DC=net))`" takes 15 seconds.

Expected behavior:
It should search in LDAP Group Base DN
Same query takes 2 seconds

**Steps to reproduce the problem:**
Setup LDAP authentication

**Versions:**
- harbor version: [1.9.0]
- docker engine version: [18.09.6]

stonezdj:
It seems that the nested group search very slow and sometimes return errors, we will create a configure option to enable/disable nested groups. do you use nested group?

Mr-iX:
I have the same problem. After upgrading to Harbor 1.9.0 the login isn't possible anymore, because of this ldap query which times out.
I compared the debug core.log with the previous version 1.8.2 and noticed that this ldap query is only executed in Harbor 1.9.0. The issue seems to be introduced with the PR #8378
To get the login working again, we would need an option to disable this query.

jomeier:
Same here. Harbor 1.9.0 LDAP based Login doesn't work anymore.

martinbalint:
>do you use nested group?
We do not need nested groups in our environment.
But basically the issue with speed seems to be with basedn, where those nested groups are searched in. If you use groupbasedn instead if basedn, it is 10x faster here.

stonezdj:
hi, @martinbalint  what is type of your LDAP server, is it a Active Directory or Open LDAP?

Issue #9110, opened 2019-09-16
LDAP Search failing during nested group search with "Bad search filter"


charlottestjohn:
Trying to setup Harbor LDAP authentication, and running into an issue that I'm not sure I understand.

Here's my configuration values: 

Key | Value
------------ | -------------
Auth Mode | LDAP
LDAP URL | ldap://oururl:389
LDAP Search DN | blank
LDAP Search Password | blank
LDAP Base DN | ou=people,dc=company,dc=local
LDAP Filter | blank
LDAP UID | uid 
LDAP Scope | OneLevel
LDAP Group Base DN | ou=groups,dc=company,dc=local
LDAP Group Filter | blank
LDAP Group GID | cn 
LDAP Group Admin DN | blank
LDAP Group Membership | memberOf
LDAP Group Scope | OneLevel
LDAP Verify Certificate | checked

When my users log in, the logs show this:
```
2019-09-16T20:42:30Z [DEBUG] [/common/dao/user.go:277]: Check if user myusername is super user
2019-09-16T20:42:30Z [DEBUG] [/core/auth/authenticator.go:139]: Current AUTH_MODE is ldap_auth
2019-09-16T20:42:30Z [ERROR] [/common/config/manager.go:192]: failed to get key ldap_search_dn, error: the configure value is not set
2019-09-16T20:42:30Z [ERROR] [/common/config/manager.go:192]: failed to get key ldap_group_admin_dn, error: the configure value is not set
2019-09-16T20:42:30Z [DEBUG] [/common/utils/ldap/ldap.go:363]: ldap filter :(uid=myusername)
2019-09-16T20:42:30Z [DEBUG] [/common/utils/ldap/ldap.go:299]: Membership attribute: memberOf
2019-09-16T20:42:30Z [DEBUG] [/common/utils/ldap/ldap.go:315]: Search ldap with filter:(uid=myusername)
2019-09-16T20:42:30Z [DEBUG] [/common/utils/ldap/ldap.go:330]: Found entries:1
2019-09-16T20:42:30Z [DEBUG] [/common/utils/ldap/ldap.go:203]: Current ldap entry attr name: uid
2019-09-16T20:42:30Z [DEBUG] [/common/utils/ldap/ldap.go:203]: Current ldap entry attr name: cn
2019-09-16T20:42:30Z [DEBUG] [/common/utils/ldap/ldap.go:203]: Current ldap entry attr name: mail
[...]

Issue #9092, opened 2019-09-13
LDAP authentication failed (invalid characters)


lukwgnd:
If you are reporting a problem, please make sure the following information are provided:

**Expected behavior and actual behavior:**
I can not log into Harbor using LDAP as a provider. I receive the error message "Invalid user name or password". The same LDAP configuration works with Harbor 1.8.

**Steps to reproduce the problem:**

- update / install harbor 1.9.0rc2 and configure ldap authentication.
- try to log-in with ad-credentials

**Versions:**
Please specify the versions of following systems.

- harbor version: 1.9.0rc2
- docker engine version: 19.03.1
- docker-compose version: 1.24.0

**Additional context:**

- **Harbor config files:** You can get them by packaging `harbor.yml` and files in the same directory, including subdirectory.
- **Log files:** You can get them by package the `/var/log/harbor/` .
```
[ERROR] [/common/config/manager.go:192]: failed to get key ldap_filter, error: the configure value is not set
[WARNING] [/core/auth/ldap/ldap.go:65]: ldap search fail: LDAP Result Code 201 "": ldap: invalid characters for escape in filter
[ERROR] [/core/controllers/base.go:107]: Error occurred in UserLogin: LDAP Result Code 201 "": ldap: invalid characters for escape in filter
```

wy65701436:
hi @lukwgnd, is it fresh install or upgrade from the previous harbor?

lukwgnd:
Hey @wy65701436 , I tried both. Can I help with more information?

stonezdj:
Is there any special character in the username? such as space or ` * ( ) \`

lukwgnd:
@stonezdj: no special character in the username. 

LDAP Filter: empty
LDAP Group Filter: objectClass=group

This works in harbor 1.8.2.

lukwgnd:
Similar / same problem as in https://github.com/goharbor/harbor/issues/9110

stonezdj:
The #9110 ldap return code is 12, and in your case is 201, 
Could you please update the common/config/core/env 
LOG_LEVEL=debug
and restart harbor
and retry login with ldap user, and upload the detail trace log?

Szauka:
Hi @stonezdj 

I also see the same issue.

key  | Value
---- | ------
LDAP Filter | "objectClass=*" or "blank"
LDAP Group Filter | objectclass=group

Harbor
Version v1.9.0-d13cf448

This are the logs:
```
[...]

PR #9498, opened 2019-10-21
Remove the nested group search -- cherrypick 1.9

stonezdj:
Remove the code change in #8378, because the previous code change caused issues: #9092, #9110, #9326


reasonerjt:
LGTM, btw I think this is a cherrypick right?

PR #9506, opened 2019-10-21
Enable robot account to support scan pull case

wy65701436:
1, the commit is for the internal robot to support scan pull case, like vul and signature checking.
2, add a bool attribute into registry token, decode it in the harbor core and add the status into request context.
3, add a bool attribut for robot API controller, but API will not use it.


src/common/security/admiral/context.go
```diff
@@ -64,6 +64,11 @@ func (s *SecurityContext) IsSysAdmin() bool {
 	return s.ctx.IsSysAdmin()
 }
 
+// PolicyCheck ...
```
reasonerjt: I don't think we need this, we can modify the `Can` func for robot security context and make it return true for a special action.

src/core/service/token/authutils.go
```diff
@@ -100,7 +106,7 @@ func filterAccess(access []*token.ResourceActions, ctx security.Context,
 }
 
 // MakeToken makes a valid jwt token based on parms.
-func MakeToken(username, service string, access []*token.ResourceActions) (*models.Token, error) {
+func MakeToken(username, service string, access []*token.ResourceActions, policyCheck bool) (*models.Token, error) {
```
reasonerjt: I think @steven-zon mentioned the scan is just a reource action?

src/common/security/context.go
```diff
@@ -27,6 +27,8 @@ type Context interface {
 	GetUsername() string
 	// IsSysAdmin returns whether the user is system admin
 	IsSysAdmin() bool
+	// PolicyCheck returns whether the middlerwares apply to the request
```
reasonerjt: I don't think we need this, we can modify the Can func for robot security context and make it return true for a special action.

src/core/middlewares/config.go
```diff
@@ -26,13 +26,14 @@ const (
 	SIZEQUOTA        = "sizequota"
 	COUNTQUOTA       = "countquota"
 	IMMUTABLE        = "immutable"
+	REGTOKEN         = "REGTOKEN"
```
reasonerjt: Why only this is upper-case?

src/core/middlewares/regtoken/handler.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/goharbor/harbor/src/common/utils/log"
+	"github.com/goharbor/harbor/src/core/config"
+	"github.com/goharbor/harbor/src/core/middlewares/util"
+	pkg_token "github.com/goharbor/harbor/src/pkg/token"
+	"github.com/goharbor/harbor/src/pkg/token/claim"
+	"net/http"
+	"strings"
+)
+
+type regTokenHandler struct {
```
reasonerjt: Please add comment to explain what this handler does.

src/core/service/token/authutils.go
```diff
@@ -42,6 +42,12 @@ func init() {
 	privateKey = config.TokenPrivateKeyPath()
 }
 
+// ClaimSet ...
+type ClaimSet struct {
+	token.ClaimSet
+	PolicyCheck bool `json:"policy_check"`
```
reasonerjt: I don't think we should add `policy check` into the token, though we add this field to bypass the scan.
I think the token needs re-design, please discuss with @steven-zou I recall he mentioned there should be a `scan` in the claim, that makes more sense, because that means the token is for scan the image.

reasonerjt:
I do think this PR needs some re-work, please see my inline comments.

src/core/service/token/authutils.go
wy65701436: + @steven-zou, please add comments

src/core/middlewares/util/util.go
```diff
@@ -49,6 +49,8 @@ type contextKey string
 const (
 	// ImageInfoCtxKey the context key for image information
 	ImageInfoCtxKey = contextKey("ImageInfo")
+	// PolicyCheckCtxKey the context key for image information
```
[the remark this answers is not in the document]
wy65701436: updated.

src/core/middlewares/vulnerable/handler.go
```diff
@@ -52,6 +52,13 @@ func (vh vulnerableHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request)
 		return
 	}
 
+	// Token bypass policy check
+	policyCheck := req.Context().Value(util.PolicyCheckCtxKey)
+	if policyCheck != nil && !policyCheck.(bool) {
```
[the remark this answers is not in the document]
wy65701436: ok, will update.

src/core/middlewares/config.go
wy65701436: updated

src/core/middlewares/regtoken/handler.go
wy65701436: sure, added.

src/core/service/token/authutils.go
wy65701436: + @steven-zou please add comments

src/core/service/token/authutils.go
wy65701436: updated

src/common/security/context.go
wy65701436: updated.

src/core/service/token/authutils.go
wy65701436: updated

src/core/middlewares/regtoken/handler.go
```diff
@@ -0,0 +1,77 @@
...
+
+// ServeHTTP ...
+func (r *regTokenHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+
+	imgRaw := req.Context().Value(util.ImageInfoCtxKey)
+	if imgRaw == nil || !config.WithClair() {
+		r.next.ServeHTTP(rw, req)
+		return
+	}
+	img, _ := req.Context().Value(util.ImageInfoCtxKey).(util.ImageInfo)
```
steven-zou: Code here can be simplified as the image info has already been extracted out.

src/core/middlewares/regtoken/handler.go
```diff
@@ -0,0 +1,77 @@
...
+	return &regTokenHandler{
+		next: next,
+	}
+}
+
+// ServeHTTP ...
+func (r *regTokenHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+
+	imgRaw := req.Context().Value(util.ImageInfoCtxKey)
+	if imgRaw == nil || !config.WithClair() {
```
steven-zou: Should not depend on `config.WithClair()`, now it should depend on the `GetProjectScanner` method in the scanner ctl.

src/core/middlewares/regtoken/handler.go
```diff
@@ -0,0 +1,77 @@
...
+	parts := strings.Split(req.Header.Get("Authorization"), " ")
+	if len(parts) != 2 || strings.ToLower(parts[0]) != "bearer" {
+		r.next.ServeHTTP(rw, req)
+		return
+	}
+	rawToken := parts[1]
+	opt := pkg_token.DefaultTokenOptions()
+	regTK, err := pkg_token.Parse(opt, rawToken, &registry.Claim{})
+	if err != nil {
+		log.Debugf("failed to decode reg token: %v, the error is skipped and round the request to native registry.", err)
```
steven-zou: Should use Errorf here

steven-zou: As it is already a not valid token, why we still let it pass to the next step?

src/core/middlewares/regtoken/handler.go
```diff
@@ -0,0 +1,77 @@
...
+		},
+		Action: rbac.ActionScannerPull.String(),
+	})
+
+	accessSet := regTK.Claims.(*registry.Claim).GetAccessSet()
+	for _, access := range accessItems {
+		if accessSet.Contains(access) {
+			ctx := context.WithValue(req.Context(), util.ByPassPolicyCheckCtxKey, true)
+			req = req.WithContext(ctx)
+			r.next.ServeHTTP(rw, req)
```
steven-zou: duplicated code?

src/pkg/token/claims/registry/registry.go
```diff
@@ -0,0 +1,42 @@
...
+
+// Claim implements the interface of jwt.Claims
+type Claim struct {
+	jwt.StandardClaims
+	Access []*token.ResourceActions `json:"access"`
+}
+
+// Valid valid the standard claims
+func (rc *Claim) Valid() error {
+	stdErr := rc.StandardClaims.Valid()
```
steven-zou: return `rc.StandardClaims.Valid()`

src/pkg/token/options.go
```diff
@@ -76,3 +59,20 @@ func (o *Options) GetKey() (interface{}, error) {
 		return nil, fmt.Errorf(fmt.Sprintf("unsupported sign method, %s", o.SignMethod))
 	}
 }
+
+// DefaultTokenOptions ...
+func DefaultTokenOptions() *Options {
+	privateKeyFile := config.TokenPrivateKeyPath()
+	privateKey, err := ioutil.ReadFile(privateKeyFile)
+	if err != nil {
+		log.Errorf(fmt.Sprintf("failed to read private key %v", err))
```
steven-zou: A general function, maybe an error should be returned, not logged

src/pkg/token/token.go
```diff
@@ -0,0 +1,77 @@
...
+	key, err := opt.GetKey()
+	if err != nil {
+		return nil, err
+	}
+	token, err := jwt.ParseWithClaims(rawToken, claims, func(token *jwt.Token) (interface{}, error) {
+		if token.Method.Alg() != opt.SignMethod.Alg() {
+			return nil, errors.New("invalid signing method")
+		}
+		switch k := key.(type) {
+		case *rsa.PrivateKey:
```
steven-zou: case *rsa.PrivateKey, *ecdsa.PrivateKey as they'll return the same value?

src/common/rbac/const.go
```diff
@@ -28,7 +28,8 @@ const (
 	ActionDelete = Action("delete")
 	ActionList   = Action("list")
 
-	ActionOperate = Action("operate")
+	ActionOperate     = Action("operate")
+	ActionScannerPull = Action("scannerpull") // for robot account created by scanner to pull image, bypass the policy check
```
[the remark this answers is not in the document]
heww: `scanner-pull` +1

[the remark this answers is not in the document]
wy65701436: updated

src/core/middlewares/regtoken/handler.go
wy65701436: removed clair check.

src/core/middlewares/regtoken/handler.go
wy65701436: just let the registry to give a real response.

src/pkg/token/options.go
wy65701436: the token handler can throw the real issue.

src/common/security/robot/robot.go
```diff
@@ -60,3 +56,20 @@ func filterPolicies(namespace rbac.Namespace, policies []*rbac.Policy) []*rbac.P
...
+	for _, policy := range project.GetAllPolicies(namespace) {
+		mp[policy.String()] = true
+	}
+
+	scannerPull := &rbac.Policy{Resource: rbac.ResourceRepository, Action: rbac.ActionCreate}
+	robotExternalPolicies := []*rbac.Policy{}
+	robotExternalPolicies = append(robotExternalPolicies, scannerPull)
+	for _, policy := range robotExternalPolicies {
+		mp[policy.String()] = true
+	}
```
heww: 1. loop is not need here
2. action is not correct in line 67 it should be `rbac.ActionScannerPull`

src/core/middlewares/regtoken/handler.go
heww: There is a function `ImageInfoFromContext` in `middlewares/util` pkg.

src/core/middlewares/contenttrust/handler.go
```diff
@@ -49,6 +49,14 @@ func (cth contentTrustHandler) ServeHTTP(rw http.ResponseWriter, req *http.Reque
...
 		return
 	}
+	// Token bypass policy check
+	if bypassPC := req.Context().Value(util.ByPassPolicyCheckCtxKey); bypassPC != nil {
+		bypassPolicyCheck, ok := bypassPC.(bool)
+		if ok && bypassPolicyCheck {
+			cth.next.ServeHTTP(rw, req)
+			return
+		}
+	}
```
heww: maybe add a function `BypassPolicyCheckFromContext` then
```go

if bypass, ok := BypassPolicyCheckFromContext(req.Context()); ok && bypass {
        cth.next.ServeHTTP(rw, req)
        return
}
```

src/core/middlewares/regtoken/handler.go
```diff
@@ -0,0 +1,76 @@
...
+			Type: rbac.ResourceRepository.String(),
+			Name: img.Repository,
+		},
+		Action: rbac.ActionScannerPull.String(),
+	})
+
+	accessSet := regTK.Claims.(*registry.Claim).GetAccessSet()
+	for _, access := range accessItems {
+		if accessSet.Contains(access) {
+			ctx := context.WithValue(req.Context(), util.ByPassPolicyCheckCtxKey, true)
```
heww: maybe add a function `NewBypassPolicyCheckContext` and put it together with `BypassPolicyCheckFromContext` so that user no need care about the `ByPassPolicyCheckCtxKey` when set and get it.

src/core/middlewares/regtoken/handler.go
```diff
@@ -0,0 +1,76 @@
...
+			req = req.WithContext(ctx)
+			r.next.ServeHTTP(rw, req)
+			return
```
heww: these two lines are not needed for at the end of this function `r.next.ServeHTTP` called.

src/core/middlewares/url/handler.go
```diff
@@ -37,7 +37,7 @@ func New(next http.Handler) http.Handler {
 
 // ServeHTTP ...
 func (uh urlHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
-	log.Debugf("in url handler, path: %s", req.URL.Path)
+	log.Infof("in url handler, path: %s", req.URL.Path)
```
heww: Is this log useful to customers? If not, suggest to keep it same before as debug message.

src/core/middlewares/vulnerable/handler.go
```diff
@@ -52,6 +52,15 @@ func (vh vulnerableHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request)
 		return
 	}
 
+	// Token bypass policy check
+	if bypassPC := req.Context().Value(util.ByPassPolicyCheckCtxKey); bypassPC != nil {
+		bypassPolicyCheck, ok := bypassPC.(bool)
```
heww: maybe add a function `BypassPolicyCheckFromContext`?

src/core/service/token/creator.go
```diff
@@ -180,6 +180,9 @@ func (rep repositoryFilter) filter(ctx security.Context, pm promgr.ProjectManage
 	} else if ctx.Can(rbac.ActionPull, resource) {
 		permission = "R"
 	}
+	if ctx.Can(rbac.ActionScannerPull, resource) {
+		permission = fmt.Sprintf("%s%s", permission, "S")
+	}
```
heww: Here is very abrupt. What about `ActionScannerPull` as a special action for registry which means `RS` like `RW` for `ActionPush` ? So that it will be like this

```go
	if ctx.Can(rbac.ActionPush, resource) && ctx.Can(rbac.ActionPull, resource) {
		permission = "RWM"
	} else if ctx.Can(rbac.ActionPush, resource) {
		permission = "RW"
	} else if ctx.Can(rbac.ActionScannerPull, resource) {
		permission = "RS"
	} else if ctx.Can(rbac.ActionPull, resource) {
		permission = "R"
	}
```

And maybe we can create a robot which only has (Resource, ScannerPull) policy.

src/pkg/robot/model/robot.go
```diff
@@ -45,12 +45,13 @@ type RobotQuery struct {
...
-	Description string         `json:"description"`
-	Disabled    bool           `json:"disabled"`
-	Visible     bool           `json:"-"`
-	Access      []*rbac.Policy `json:"access"`
+	Name              string         `json:"name"`
+	ProjectID         int64          `json:"pid"`
+	Description       string         `json:"description"`
+	Disabled          bool           `json:"disabled"`
+	Visible           bool           `json:"-"`
+	ByPassPolicyCheck bool           `json:"-"`
```
heww: `ByPassPolicyCheck` maybe not needed here, the scanner can create robot account with (Resource, ScannerPull) access.

src/core/api/robot.go
```diff
@@ -108,6 +108,7 @@ func (r *RobotAPI) Post() {
 		return
 	}
 	robotReq.Visible = true
+	robotReq.ByPassPolicyCheck = false
```
heww: If `ByPassPolicyCheck` not needed in robotReq, here is not needed too.

src/pkg/robot/controller.go
```diff
@@ -69,20 +72,36 @@ func (d *DefaultAPIController) CreateRobotAccount(robotReq *model.RobotCreate) (
 		ExpiresAt:   expiresAt,
 		Visible:     robotReq.Visible,
 	}
+	if robotReq.ByPassPolicyCheck {
+		robotReq.Access = append(robotReq.Access, &rbac.Policy{
+			Resource: rbac.NewProjectNamespace(robotReq.ProjectID).Resource(rbac.ResourceRepository),
+			Action:   rbac.ActionScannerPull,
+		})
+	}
```
heww: No needed here, scanner can provide (Repository, ScannerPull) access when create a robot.

src/core/middlewares/regtoken/handler.go
wy65701436: updated.

src/core/middlewares/regtoken/handler.go
wy65701436: removed

src/common/security/robot/robot.go
wy65701436: 1, in case we have other exceptions. leave the array here?
2, updated, thanks.

src/core/middlewares/regtoken/handler.go
wy65701436: removed, thanks.

src/core/middlewares/url/handler.go
wy65701436: resolved, thanks.

src/core/service/token/creator.go
wy65701436: agree, updated.

src/core/middlewares/vulnerable/handler.go
wy65701436: added.

src/core/middlewares/contenttrust/handler.go
wy65701436: updated.

src/core/api/robot.go
wy65701436: removed.

src/pkg/robot/controller.go
wy65701436: removed

src/pkg/robot/model/robot.go
wy65701436: removed

src/core/middlewares/regtoken/handler.go
wy65701436: updated

steven-zou:
A general comment, shall we rename the `bypass` policy to avoid evil risks?

Some comments are not resolved yet.

@wy65701436

PR #9521, opened 2019-10-22
fix syntax error

JohnJAS:


xaleeks:
this will be deprecated, we are using helm charts now.

PR #9528, opened 2019-10-22
improve the scan controlling

steven-zou:
- add LCM control to the robot account generated for scanning
- improve the scan webhook
- remove reprots when related artifact is deleted
- update report manager/scan controller and other components to support above cases
- add artifact manager/comtroller to list artifacts


src/core/api/scan_all.go
```diff
@@ -88,3 +94,12 @@ func (sc *ScanAllAPI) Get() {
 func (sc *ScanAllAPI) List() {
 	sc.list(common_job.ImageScanAllJob)
 }
+
+func isScanEnabled() (bool, error) {
+	l, err := scanner.DefaultController.ListRegistrations(nil)
```
heww: this will list all scanners

steven-zou: get the default one

PR #9530, opened 2019-10-22
Disable beego color

ninjadq:
The color can not disable because beego's bug

I create a harbor log adapter which can use harbor's logger to write log msg

Acctually this commit can not disable color in core log but will reduce some useless infos

### DIFF
--- a/src/common/utils/log/beegoadapter.go
+++ b/src/common/utils/log/beegoadapter.go
@@ -1,0 +1,90 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


heww:
If the requirement is only to disable the color of the log, maybe the following will work.

```go
	beego.SetLogger(logs.AdapterConsole, "{\"color\": false}")
```

ninjadq:
@heww That method doesn't work. Because Beego's console adapter has a bug when creating a console adapter. 
And beego's log level doesn't controlled by Harbor config and it also not consistent with harbor. So this pr also resolved this.

src/common/utils/log/beegoadapter.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/astaxie/beego/logs"
+)
+
+// BeeLevel2Hbr mapping beego's log level to harbor's log lever
```
reasonerjt: Do we have to export this variable?

ninjadq: I think import logs are the most convenient way or I need using integers as the key which is not good for readability.

PR #9538, opened 2019-10-23
Enable https by default

ninjadq:
1. Umcomment https related configs
2. Remove the https prepare related thing in ci

### DIFF
--- a/make/harbor.yml
+++ b/make/harbor.yml
@@ -10,12 +10,12 @@ http:
   port: 80
 
 # https related config
-# https:
-#   # https port for harbor, default is 443
-#   port: 443
-#   # The path of cert and key files for nginx
-#   certificate: /your/certificate/path
-#   private_key: /your/private/key/path
+https:
+  # https port for harbor, default is 443
+  port: 443
+  # The path of cert and key files for nginx
+  certificate: /your/certificate/path
+  private_key: /your/private/key/path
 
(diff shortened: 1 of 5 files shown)


make/harbor.yml
```diff
@@ -10,12 +10,12 @@ http:
...
-#   # https port for harbor, default is 443
-#   port: 443
-#   # The path of cert and key files for nginx
-#   certificate: /your/certificate/path
-#   private_key: /your/private/key/path
+https:
+  # https port for harbor, default is 443
+  port: 443
+  # The path of cert and key files for nginx
+  certificate: /data/cert/server.crt
```
reasonerjt: Why not keep it as is?

reasonerjt:
I understand the http will be redirected, but I'd suggest that we comment that section.

reasonerjt:
Can we handle the upgrade correctly with this change?

make/harbor.yml
wy65701436: +1 

or

`/PATH_TO_YOUR_CERTIFICATE`
`/PATH_TO_YOUR_PRIVATEKEY`

wy65701436: But, if we change back, do you need to add an installation check in the install.sh?

And, also, we need to update the travis and nightly accordingly?

reasonerjt:
Closing as we won't ship this change in 1.9.x

PR #9579, opened 2019-10-24
update immutable rule API

wy65701436:
1, unify disable and enable
2, fix update rule error

### DIFF
--- a/src/pkg/immutabletag/dao/immutable.go
+++ b/src/pkg/immutabletag/dao/immutable.go
@@ -12,7 +12,7 @@ import (
 type ImmutableRuleDao interface {
 	CreateImmutableRule(ir *model.ImmutableRule) (int64, error)
 	UpdateImmutableRule(projectID int64, ir *model.ImmutableRule) (int64, error)
-	ToggleImmutableRule(id int64, enabled bool) (int64, error)
+	ToggleImmutableRule(id int64, status bool) (int64, error)
 	GetImmutableRule(id int64) (*model.ImmutableRule, error)
 	QueryImmutableRuleByProjectID(projectID int64) ([]model.ImmutableRule, error)
 	QueryEnabledImmutableRuleByProjectID(projectID int64) ([]model.ImmutableRule, error)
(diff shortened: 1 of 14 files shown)


src/pkg/immutabletag/dao/immutable.go
```diff
@@ -76,7 +76,7 @@ func (i *immutableRuleDao) QueryImmutableRuleByProjectID(projectID int64) ([]mod
 // QueryEnabledImmutableRuleByProjectID get all enabled immutable rule by project
 func (i *immutableRuleDao) QueryEnabledImmutableRuleByProjectID(projectID int64) ([]model.ImmutableRule, error) {
 	o := dao.GetOrmer()
-	qs := o.QueryTable(&model.ImmutableRule{}).Filter("ProjectID", projectID).Filter("Enabled", true)
+	qs := o.QueryTable(&model.ImmutableRule{}).Filter("ProjectID", projectID).Filter("Disabled", true)
```
stonezdj: the filter should changed to disabled=false

src/pkg/immutabletag/dao/immutable_test.go
```diff
@@ -62,7 +62,7 @@ func (t *immutableRuleDaoTestSuite) TestEnableImmutableRule() {
 	newIr, err := t.dao.GetImmutableRule(id)
 
 	t.require.Nil(err)
-	t.require.False(newIr.Enabled, "Failed to disable the immutable rule")
+	t.require.False(newIr.Disabled, "Failed to disable the immutable rule")
```
stonezdj: It should be True?

src/pkg/immutabletag/manager_test.go
```diff
@@ -110,7 +110,7 @@ func (m *managerTestingSuite) TestQueryImmutableRuleByProjectID() {
 		{
 			ID:        1,
 			ProjectID: 1,
-			Enabled:   true,
+			Disabled:  true,
```
stonezdj: changed to Disabled=false

src/pkg/immutabletag/dao/immutable.go
wy65701436: ok, thanks.

src/pkg/immutabletag/dao/immutable_test.go
wy65701436: updated

PR #9588, opened 2019-10-25
Fix User Group Search SQL error

stonezdj:
User Group Query SQL error in some cases

### DIFF
--- a/src/common/dao/group/usergroup.go
+++ b/src/common/dao/group/usergroup.go
@@ -15,7 +15,6 @@
 package group
 
 import (
-	"strings"
 	"time"
 
 	"github.com/goharbor/harbor/src/common/utils"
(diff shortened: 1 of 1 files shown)


src/common/dao/group/usergroup.go
```diff
@@ -101,17 +101,17 @@ func GetUserGroup(id int) (*models.UserGroup, error) {
...
+	var qm []string
 	if len(groupName) == 0 {
 		return retGroupID, nil
 	}
-	for _, gName := range groupName {
-		con := "'" + gName + "'"
-		conditions = append(conditions, con)
+	for i := 0; i < len(groupName); i++ {
+		qm = append(qm, "?")
 	}
```
heww: There is a function `ParamPlaceholderForIn` can do this.

stonezdj: Updated

PR #9599, opened 2019-10-25
Fix quota sync crash on getting the unknown mainfest

wy65701436:
1, eat the unknown manifest error, and log it. The migration process will not crashed on it.
2, enable to persist DB of sync quota API.
3, add empty project support.

append empty project

### DIFF
--- a/src/core/api/quota/registry/registry.go
+++ b/src/core/api/quota/registry/registry.go
@@ -86,6 +86,11 @@ func (rm *Migrator) Dump() ([]quota.ProjectInfo, error) {
 			repoMap[pro.Name] = repos
 		}
 	}
+	repoMap, err = rm.appendEmptyProject(repoMap)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


src/core/api/quota/registry/registry.go
```diff
@@ -143,6 +148,24 @@ func (rm *Migrator) Dump() ([]quota.ProjectInfo, error) {
 	return projects, nil
 }
 
+// As catalog api cannot list the empty projects in harbor, here it needs to append the empty projects into repo infor
+// so that quota syncer can add 0 usage into quota usage.
+func (rm *Migrator) appendEmptyProject(repoMap map[string][]string) (map[string][]string, error) {
+	var withEmptyProjects map[string][]string
+	all, err := dao.GetProjects(nil)
+	if err != nil {
+		return withEmptyProjects, err
```
heww: Maybe we should return the repositories from the catalog when failed to get projects from db rather than return empty repositories?

PR #9607, opened 2019-10-28
Fix the error when delete some images

### LINKED ISSUE #9603: Unorganized error warning --test   immutable
I have created an immutable-tag ,and push images,  when delete an image that has contact with immutable tag , it can not delete  this image, it is right ,but  the error message is wrong

image

### LINKED ISSUE #9604: delete images error
When I delete multiple images, which contains images that cannot be deleted, the delete display fails, but the background has been deleted. The foreground is not refreshed.
image

jwangyangls:
fix #9604  fix #9603

### DIFF
--- a/src/portal/lib/src/shared/shared.utils.ts
+++ b/src/portal/lib/src/shared/shared.utils.ts
@@ -1,3 +1,5 @@
+import { errorSrcWithoutHttpClient } from "ngx-markdown";
+
 // Copyright (c) 2017 VMware, Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


zhoumeina:
image

zhoumeina:
image

zhoumeina:
this text is not vertical center. And text should be 13px.

src/portal/lib/src/shared/shared.utils.ts
```diff
@@ -22,19 +22,43 @@ export const errorHandler = function (error: any): string {
...
     try {
         return JSON.parse(error.error).message;
     } catch (err) { }
+    // Not a standard error return Basically not used cover unknown error
     if (typeof error.error === "string") {
         return error.error;
     }
     if (error.error && error.error.message) {
-        return error.error.message;
+        // handle docker client response error
```
zhoumeina: Add unit test for this logic.

AllForNothing:
In the future , we need to take more care of  batch operation

PR #9621, opened 2019-10-28
[cherry-pick] Fix quota sync crash on getting the unknown mainfest

wy65701436:
1, eat the unknown manifest error, and log it. The migration process will not crashed on it.
2, enable to persist DB of sync quota API.
3, add empty project support.

### DIFF
--- a/src/core/api/internal.go
+++ b/src/core/api/internal.go
@@ -172,7 +172,8 @@ func (ia *InternalAPI) SyncQuota() {
 			cfgMgr.Save()
 		}()
 		log.Info("start to sync quota(API), the system will be set to ReadOnly and back it normal once it done.")
-		err := quota.Sync(ia.ProjectMgr, false)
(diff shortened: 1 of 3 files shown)


src/core/api/internal.go
```diff
@@ -172,7 +172,8 @@ func (ia *InternalAPI) SyncQuota() {
 			cfgMgr.Save()
 		}()
 		log.Info("start to sync quota(API), the system will be set to ReadOnly and back it normal once it done.")
-		err := quota.Sync(ia.ProjectMgr, false)
+		// As the sync function ignores all of duplicate error, it's safe to enable persist DB.
```
renmaosheng: could you please elaborate this a little bit?

wy65701436: it just an enhancement for sync quota API, which is designed for sys-admin used only for some urgent case to correct the quota usage data.

Originally, it will not persist the artifact related data as a duplicate failure. Here just enable it as the duplicate errors are all ignored in the previous fix.

PR #9623, opened 2019-10-28
fix: add spec of oidc settings in swagger.yaml

chlins:
Issue: https://github.com/goharbor/harbor/issues/9624


reasonerjt:
Thanks for pointing it out, but could you specify in the description?  For example the endpoint of oidc has to be https, and scope should be separated by comma, etc.

chlins:
@reasonerjt I have changed the descriptions, maybe need you take time to look again. thanks.

PR #9626, opened 2019-10-28
Upgrade script 190 to 110

ninjadq:
Add migrator for harbor 1.10.0


reasonerjt:
This should also be updated:
https://github.com/goharbor/harbor/blob/master/make/harbor.yml#L99

Issue #9600, opened 2019-10-25
[OIDC] Conflict in username, the user with same username has been onboarded.


annProg:
**Expected behavior and actual behavior:**
Second user login with oidc will see `Conflict in username, the user with same username has been onboarded.`

harbor-core log:

```
2019-10-25T09:51:21Z [DEBUG] [/core/controllers/oidc.go:178]: User info string: {"iss":"https://oidc.scloud.xx.com","sub":"CgRoZWFuEglsZWNvbW1vbnM","name":"john","email":""}
2019-10-25T09:51:21Z [ERROR] [/common/dao/oidc_user.go:135]: fail to insert user, pq: duplicate key value violates unique constraint "harbor_user_email_key"
2019/10/25 09:51:21 [D] [server.go:2774] |   10.58.102.45| 409 |    6.83201ms|   match| POST     /c/oidc/onboard   r:/c/oidc/onboard
2019-10-25T09:51:25Z [DEBUG] [/core/filter/security.go:227]: OIDC CLI modifier only handles request by docker CLI or helm CLI
2019-10-25T09:51:25Z [DEBUG] [/core/filter/security.go:442]: can not get user information from session
2019-10-25T09:51:25Z [DEBUG] [/core/filter/security.go:511]: user information is nil
2019-10-25T09:51:25Z [DEBUG] [/core/filter/security.go:525]: using local database project manager
2019-10-25T09:51:25Z [DEBUG] [/core/filter/security.go:527]: creating local database security context...
```

It say `duplicate key value violates unique constraint "harbor_user_email_key"`, and the email of first user login with oidc is empty. （Second user's email is empty too: `User info string: {"iss":"https://oidc.scloud.xx.com","sub":"CgRoZWFuEglsZWNvbW1vbnM","name":"john","email":""}` )

oidc claims with email address

```
Claims:

{
  "iss": "https://oidc.scloud.xx.com",
  "sub": "CgRoZWFuEglsZWNvbW1vbnM",
  "aud": "example-app",
  "exp": 1572085737,
  "iat": 1571999337,
  "at_hash": "B3Ipt0n9usQscDA4d-jjZw",
  "email": "john@qq.com",
  "email_verified": true,
  "name": "john"
}
```

**Steps to reproduce the problem:**
Please provide the steps to reproduce this problem.

**Versions:**
Please specify the versions of following systems.

- harbor version: [1.9.0]

**Additional context:**

[...]

PR #9643, opened 2019-10-29
FetchUserInfo retrieving details about the logged-in user

### LINKED ISSUE #7885: Incorrect Email (placeholder) is shown when using Gitlab as OIDC provider
**Docker Version**
Docker version 18.09.6, build 481bc77
**Docker Compose Version**
docker-compose version 1.23.2, build 1110ad01
**Harbor Version**
1.8.0
**Gitlab Version**
11.9.7-ee (782806f)

The signup process itself using Gitlab as OIDC Provider is working without any issues. However, when a user is logging in a user account is created but the associated email address is of the format
`5eea2h6bzid4qgh95enqpbjfo8f9497x@placeholder.com`
**Scopes set in Gitlab**
image
**Configuration in Harbor**
image

jsimomaa:
According to the openidconnect spec when `response_type=code` `id_token` does not need to include user detail claims (e.g. `email`) if Access Token is being issued.

> In this case the Claims requested by the `profile, email, address, and phone` scope values are returned from the UserInfo Endpoint, as described in Section 5.3.2

http://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims

This solution first unmarshals claims from `userInfo` (if present) and then overrides the values with possible claims from `IdToken` therefore keeping backwards compatibility.

Should fix #7885 and #9600

Signed-off-by: jsimomaa <jani.simomaa@gmail.com>

### DIFF
--- a/src/common/utils/oidc/helper.go
+++ b/src/common/utils/oidc/helper.go
@@ -208,6 +208,24 @@ func RefreshToken(ctx context.Context, token *Token) (*Token, error) {
 	return &Token{Token: *t, IDToken: it}, nil
 }
 
+// FetchUserInfo retrieving details about the logged-in user
+// https://connect2id.com/products/server/docs/api/userinfo
+// 
+// The Claims requested by the profile, email, address, and phone scope values are returned from the UserInfo Endpoint,
+// as described in Section 5.3.2, when a response_type value is used that results in an Access Token being issued. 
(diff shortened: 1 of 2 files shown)


reasonerjt:
@jsimomaa Thanks for the PR actually I have another PR leveraging the userinfo endpoint to fetch and populate user info and group membership, so this PR will be closed.

Issue #9081, opened 2019-09-12
Task: CII badging req - makefile for native binaries


xaleeks:
> This issue is a sub-task of the CII badging to-do tasks epic (#9076)

Req: Build systems for native binaries MUST honor the relevant compiler and linker (environment) variables passed in to them (e.g., CC, CFLAGS, CXX, CXXFLAGS, and LDFLAGS) and pass them to compiler and linker invocations. A build system MAY extend them with additional flags; it MUST NOT simply replace provided values with its own. [build_standard_variables]

wy65701436:
PR: https://github.com/goharbor/harbor/pull/9649

wy65701436:
close as PR is merged.

PR #9649, opened 2019-10-29
add ldflags for harbor compiler and linker

wy65701436:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,9 +71,6 @@ TOOLSPATH=$(BUILDPATH)/tools
 CORE_PATH=$(BUILDPATH)/src/core
 PORTAL_PATH=$(BUILDPATH)/src/portal
 CHECKENVCMD=checkenv.sh
-GITCOMMIT := $(shell git rev-parse --short=8 HEAD)
-RELEASEVERSION := $(shell cat VERSION)
-PKG_PATH=github.com/goharbor/harbor/src/pkg
 
 # parameters
 REGISTRYSERVER=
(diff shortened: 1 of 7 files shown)


xaleeks:
to fix : https://github.com/goharbor/harbor/issues/9081

Makefile
```diff
@@ -143,8 +142,10 @@ GODEP=$(GOTEST) -i
 GOFMT=gofmt -w
 GOBUILDIMAGE=golang:1.12.12
 GOBUILDPATH=/harbor
-GOIMAGEBUILDCMD=/usr/local/go/bin/go
-GOIMAGEBUILD=$(GOIMAGEBUILDCMD) build -mod vendor
+GOIMAGEBUILDCMD=/usr/local/go/bin/go build -mod vendor
+GOIMAGEBUILD_COMMON=$(GOIMAGEBUILDCMD) --ldflags "$(ADDITIONAL_LDFLAGS)"
```
reasonerjt: if look at this https://github.com/docker/distribution/blob/master/Makefile#L91

There's BUILD FLAGS this is also needed because build flags are not limited to ldflags

A simple search at golang.org you can see there are more than that:
```
-asmflags '[pattern=]arg list'
	arguments to pass on each go tool asm invocation.
-buildmode mode
	build mode to use. See 'go help buildmode' for more.
-compiler name
	name of compiler to use, as in runtime.Compiler (gccgo or gc).
-gccgoflags '[pattern=]arg list'
	arguments to pass on each gccgo compiler/linker invocation.
-gcflags '[pattern=]arg list'
	arguments to pass on each go tool compile invocation.
.....
```

wy65701436: updated.

PR #9662, opened 2019-10-30
[Cherry pick]fix(quota): order by quotas only on support resources

heww:
1. Limit `ListQuotas` and `ListQuotaUsages` sort only on support resources.


xaleeks:
"order by quotas only on support resources" - can you write a more detailed description on what this is trying to accomplish, it's not very clear from this. @heww

PR #9673, opened 2019-10-31
return more clear error message for scan related API

### LINKED ISSUE #9668: Return more reasonable error for the scanning status conflicts
Change 500 to 409

steven-zou:
- add a common error pkg to support error with code and AsError check
- replace some errors in scan with coded errors
- fix #9668


reasonerjt:
For this particular issue, I think the common infrastructure code better be discussed thoroughly before introduction.  Considering the point of time, I wish to suggest either keep the scope of err package to scan or return a common http error which also satisfy your requirement.

steven-zou:
> I don't think we should introduce the package like `src/pkg/errs` as there will be enhancement in go1.13

I know the updates. Will we upgrade the golang version in 1.10?  If not, I need such error format.

steven-zou:
@reasonerjt 

Moved `errs` pkg to `scan` package.

Issue #9740, opened 2019-11-05
use config variable instead of hardcore local url in Replication adapter


DanielQujun:
Replication got "500 Internal Server error" when harbor-core listening on  a non-default httpPort, This was Related to  [#7718](https://github.com/goharbor/harbor/pull/7718)

```
// when the adapter is created for local Harbor, returns the "http://127.0.0.1:8080"
// as URL to avoid issue https://github.com/goharbor/harbor-helm/issues/222
// when harbor is deployed on Kubernetes
func (a *adapter) getURL() string {
	if a.registry.Type == model.RegistryTypeHarbor && a.registry.Name == "Local" {
		return "http://127.0.0.1:8080"
	}
	return a.url
}
```

Since we already have "CORE_LOCAL_URL"  environment variable, "http://127.0.0.1:8080"  should  replace by "config.LocalCoreURL()"

DanielQujun:
@ywk253100

ywk253100:
@DanielQujun The port of core is used internally, it cannot be configured. The port can be configured in `harbor.yml` is for the nginx. 

The `127.0.0.1` is used to fix issue https://github.com/goharbor/harbor-helm/issues/222, it cannot be replaced by `config.LocalCoreURL()`

DanielQujun:
Sorry for  misunderstood `config.LocalCoreURL()` , but to fix kubernetes hairpin mode disabled no need have to be  "127.0.0.1", any ip configured in that pod will work. that why I thought  `config.LocalCoreURL()` would be useful.

In my case ,8080 is used by another service, and I  edited httpport section in `app.conf` the configure file of core. It worked well except replication  error as I post above.

The  8080 port is used very common , I suggest the port of core should be configured.

ywk253100:
Could you explain more about `but to fix kubernetes hairpin mode disabled no need have to be "127.0.0.1", any ip configured in that pod will work`?

For `In my case ,8080 is used by another service`, I didn't get your point why you have to change the default port

DanielQujun:
[ #7718](https://github.com/goharbor/harbor/pull/7718): Fix issue: replication cannot work when Harbor is deployed on Kubernetes whose hairpin mode is disabled

[...]

PR #9741, opened 2019-11-05
replication 500 server error when harbor-core listening on a non-defa…

DanielQujun:
fix issue [goharbor/harbor-helm#9740](https://github.com/goharbor/harbor/issues/9740): use config variable instead of hardcore local url for Replication adapter


DanielQujun:
ping @ywk253100

PR #9746, opened 2019-11-06
Add configuration "case sensitive" to HTTP auth proxy

reasonerjt:
This commit make case sensitivity configurable when the authentication
backend is auth proxy.
When the "http_authproxy_case_sensitive" is set to false, the name of
user/group will be converted to lower-case when onboarded to Harbor, so
as long as the authentication is successful there's no difference regardless
upper or lower case is used.  It will be mapped to one entry in Harbor's
User/Group table.

### DIFF
--- a/src/common/const.go
+++ b/src/common/const.go
@@ -104,6 +104,7 @@ const (
 	HTTPAuthProxyTokenReviewEndpoint = "http_authproxy_tokenreview_endpoint"
 	HTTPAuthProxyVerifyCert          = "http_authproxy_verify_cert"
 	HTTPAuthProxySkipSearch          = "http_authproxy_skip_search"
+	HTTPAuthProxyCaseSensitive       = "http_authproxy_case_sensitive"
 	OIDCName                         = "oidc_name"
 	OIDCEndpoint                     = "oidc_endpoint"
 	OIDCCLientID                     = "oidc_client_id"
--- a/src/core/auth/authproxy/auth.go
+++ b/src/core/auth/authproxy/auth.go
@@ -40,11 +40,14 @@ import (
 const refreshDuration = 2 * time.Second
 const userEntryComment = "By Authproxy"
 
-var secureTransport = &http.Transport{}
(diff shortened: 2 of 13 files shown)


src/core/auth/authproxy/auth.go
```diff
@@ -140,6 +158,9 @@ func (a *Auth) tokenReview(sessionID string) (*k8s_api_v1beta1.TokenReview, erro
 
 // OnBoardUser delegates to dao pkg to insert/update data in DB.
 func (a *Auth) OnBoardUser(u *models.User) error {
+	if !a.CaseSensitive {
+		u.Username = strings.ToLower(u.Username)
+	}
```
stonezdj: It is better to extract a function so that the a.CaseSesnsitive appear everywhere.

src/common/const.go
```diff
@@ -104,6 +104,7 @@ const (
 	HTTPAuthProxyTokenReviewEndpoint = "http_authproxy_tokenreview_endpoint"
 	HTTPAuthProxyVerifyCert          = "http_authproxy_verify_cert"
 	HTTPAuthProxySkipSearch          = "http_authproxy_skip_search"
+	HTTPAuthProxyCaseSensitive       = "http_authproxy_case_sensitive"
```
stonezdj: The case sensitive might be used by other auth mod, for example, LDAP

src/core/auth/authproxy/auth.go
reasonerjt: If we do that we still need to call this function in different funcs in `core/auth/authproxy/auth.go` , so I don't think it we need to do it as this piece of code is quite short.

src/common/const.go
reasonerjt: Is LDAP case-insensitive

src/core/auth/authproxy/auth.go
reasonerjt: fixed

PR #9747, opened 2019-11-06
Increase memory space when ng build

jwangyangls:

### DIFF
--- a/make/photon/portal/Dockerfile
+++ b/make/photon/portal/Dockerfile
@@ -20,7 +20,7 @@ COPY src/portal /build_dir
 RUN ls -la \
     && npm run build_lib \
     && npm run link_lib \
-    && npm run release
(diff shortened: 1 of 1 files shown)


make/photon/portal/Dockerfile
```diff
@@ -20,7 +20,7 @@ COPY src/portal /build_dir
 RUN ls -la \
     && npm run build_lib \
     && npm run link_lib \
-    && npm run release
+    && node --max_old_space_size=8192 'node_modules/@angular/cli/bin/ng' build --prod
```
reasonerjt: Do we really need 8GB of heap size of V8 VM?

jwangyangls: By default the memory limit in Node.js is 512 mb.So we only need to specify a quantity more than it can, I am now trying to specify this number as 2G.

PR #9761, opened 2019-11-07
use local tdnf server to build harbor images

wy65701436:


.drone.yml
```diff
@@ -55,6 +55,8 @@ pipeline:
       - npm_registry
       - harbor_sign_key
       - harbor_sign_key_id
+      - use_internal_tdnf
```
reasonerjt: I don't think  we need the flag `use_internal_tdnf` how about we just check if the `harbor_tdnf_repo` is set  ` ! -z $HARBOR_TDNF_REPO ]`

tests/integration.sh
```diff
@@ -105,6 +106,8 @@ function uploader {
 
 function package_offline_installer {
     echo "Package Harbor offline installer."
+    sed "s/harbor_tdnf_repo/$HARBOR_TDNF_REPO/" -i make/photon/tdnf/photon-local.repo
```
reasonerjt: run `sed` when $HARBOR_TDNF_REP is non-empty.

make/photon/chartserver/Dockerfile
```diff
@@ -1,10 +1,25 @@
 FROM photon:2.0
 
+COPY ./make/photon/tdnf/photon.internal ./make/photon/tdnf/photon-updates.internal /etc/yum.repos.d/
+
+# use the internal tdnf repo server
+ARG USE_INTERNAL_TDNF
```
reasonerjt: We don't need this flag, we can check if the photon.internal and photon-updates.internal is updated or not.

make/photon/chartserver/Dockerfile
```diff
@@ -1,10 +1,25 @@
...
+RUN if [ "$USE_INTERNAL_TDNF" = "true" ] ; then mv /etc/yum.repos.d/photon.repo /etc/yum.repos.d/photon.repo.bk ; fi \
```
reasonerjt: seems we only need one `if ... fi`

make/photon/chartserver/Dockerfile
```diff
@@ -1,10 +1,25 @@
...
+
 RUN tdnf install -y shadow sudo >>/dev/null\
     && tdnf clean all \
     && groupadd -r -g 10000 chart \
     && useradd --no-log-init -m -g 10000 -u 10000 chart
 
+# roll back tdnf repo server
+RUN if [ "$USE_INTERNAL_TDNF" = "true" ] ; then mv /etc/yum.repos.d/photon.repo.bk /etc/yum.repos.d/photon.repo ; fi \
+    && if [ "$USE_INTERNAL_TDNF" = "true" ] ; then mv /etc/yum.repos.d/photon-updates.repo.bk /etc/yum.repos.d/photon-updates.repo ; fi \
+    && if [ "$USE_INTERNAL_TDNF" != "true" ] ; then rm /etc/yum.repos.d/photon.internal ; fi \
```
reasonerjt: we can remove the .internal files regardless if internal TDNF repo is to be used.

reasonerjt:
Generally I don't think we need the extra `use_internal` flag, we just customize the repo url per configuration when it's set.

reasonerjt:
We'll build base images so local tdnf cache is not needed.

PR #9810, opened 2019-11-08
Populate group from auth provider to Harbor when user login

stonezdj:
Change include LDAP auth, OIDC auth, HTTP auth

### DIFF
--- a/src/common/dao/group/usergroup.go
+++ b/src/common/dao/group/usergroup.go
@@ -19,8 +19,6 @@ import (
 
 	"github.com/goharbor/harbor/src/common/utils"
 
-	"fmt"
-
 	"github.com/goharbor/harbor/src/common"
 	"github.com/goharbor/harbor/src/common/dao"
 	"github.com/goharbor/harbor/src/common/models"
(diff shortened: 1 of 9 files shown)


src/common/dao/group/usergroup.go
```diff
@@ -97,23 +97,46 @@ func GetUserGroup(id int) (*models.UserGroup, error) {
...
+	cnt, err := o.Raw(sql, groupKeys, groupType).QueryRows(&retGroupID)
 	if err != nil {
 		return retGroupID, err
 	}
 	log.Debugf("Found rows %v", cnt)
 	return retGroupID, nil
 }
 
+// SyncGroupByGroupKey -  Return the group ID by given group name. if not exist in Harbor DB, create one
+func SyncGroupByGroupKey(groupKeys []string, groupType int) ([]int, error) {
```
reasonerjt: it accept a list of groupentries so that it will not care about the group type.

reasonerjt:
I wish to suggest rename the commit message and PR, because it's not really syncing, it's just populate based on the returning value from auth backend.

src/common/dao/group/usergroup.go
```diff
@@ -97,21 +95,16 @@ func GetUserGroup(id int) (*models.UserGroup, error) {
...
-		return retGroupID, nil
-	}
-	sql := fmt.Sprintf("select id from user_group where group_name in ( %s ) and group_type = ? ", dao.ParamPlaceholderForIn(len(groupName)))
-	log.Debugf("GetGroupIDByGroupName: statement sql is %v", sql)
-	o := dao.GetOrmer()
-	cnt, err := o.Raw(sql, groupName, groupType).QueryRows(&retGroupID)
-	if err != nil {
-		return retGroupID, err
+// SyncGroup -  Return the group ID by given group name. if not exist in Harbor DB, create one
+func SyncGroup(userGroups []models.UserGroup) ([]int, error) {
```
reasonerjt: I don't think it's appropriate to call it sync, it's just batch onboarding.

src/common/dao/group/usergroup.go
```diff
@@ -172,3 +165,12 @@ func OnBoardUserGroup(g *models.UserGroup) error {
 	}
 	return onBoardCommonUserGroup(g, "GroupName", "GroupType")
 }
+
+// GetModelsFromName ...
+func GetModelsFromName(groupNames []string, groupType int) []models.UserGroup {
```
reasonerjt: If this is a exported func, wouldn't it be moved to models pkg? as it has nothing to do with DB operation?

PR #9813, opened 2019-11-08
Re-implemented LDAP nested group searching

Typositoire:
This is a take 2.

Re-implemented the feature, but behind an `experimental` toggle that is off by default.

I've tested this for basic functionality and also simplified the code.

Last thing to do is add test cases but I'm getting `POSTGRES` errors .. What's the setup for that? It's not in the `contributing` documentation. I'll go back to previous ticket and try to account for already known issues.

This is kind of a deal breaker for us, so I'd really appreciate your help to make this works this time!

harbor-toggle


Typositoire:
What would be the best time to talk and get me up with the testing setup? My new test fails which I expected cuz it's wrong lol But I'd like to be able to test them locally before going through Travis.

reasonerjt:
I don't think we can merge it in 1.10 as we are so close to cutting an RC.

xaleeks:
yep this is out of scope for 1.10 but we can potentially merge for the next release. @Typositoire

Typositoire:
Still have no idea how to run Unit Tests... This really doesn't help for contributions.

Typositoire:
We dropped Harbor here. So let's forget about this.

PR #9816, opened 2019-11-11
Add role list in project response

### LINKED ISSUE #9771: Project member role display not match the real permissions for users have multi groups in OIDC
1. Set OIDC as auth mode
2. Create two member groups `group1` and `group2` in library project, group1 as guest role, group2 as master role
3. Login harbor with `user1` which have `group1` and `group2` in OIDC

The role for `user1` in library project should be master, but actual is guest.

reasonerjt:
This commit fixes #9771

It compares the roles to return the one with highest permission in the
response of `GET /api/projects`.
In addition to that, it adds the role list to the response, because a
user can have multiple roles in a project.
It also removes the togglable attribute as it's not used anywhere.

### DIFF
--- a/API/harbor/swagger.yaml
+++ b/API/harbor/swagger.yaml
@@ -4871,7 +4871,14 @@ definitions:
         description: Correspond to the UI about whether the project's publicity is  updatable (for UI)
       current_user_role_id:
         type: integer
-        description: The role ID of the current user who triggered the API (for UI)
+        description: The role ID with highest permission of the current user who triggered the API (for UI)
+        deprecated: true
+      current_user_role_ids:
+        type: array
+        items:
+          type: integer
+          format: int32
+        description: The list of role ID of the current user who triggered the API (for UI)
       repo_count:
         type: integer
         description: The number of the repositories under this project.
--- a/src/core/api/project.go
+++ b/src/core/api/project.go
@@ -460,14 +460,8 @@ func (p *ProjectAPI) List() {
 func (p *ProjectAPI) populateProperties(project *models.Project) error {
 	if p.SecurityCtx.IsAuthenticated() {
 		roles := p.SecurityCtx.GetProjectRoles(project.ProjectID)
-		if len(roles) != 0 {
-			project.Role = roles[0]
-		}
-
-		if project.Role == common.RoleProjectAdmin ||
-			p.SecurityCtx.IsSysAdmin() {
-			project.Togglable = true
-		}
+		project.RoleList = roles
+		project.Role = highestRole(roles)
 	}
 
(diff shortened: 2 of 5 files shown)


src/core/api/project.go
```diff
@@ -712,3 +707,34 @@ func getProjectMemberSummary(projectID int64, summary *models.ProjectSummary) {
...
+		}
+	}
+	roles = roles[:ind]
+	if len(roles) == 0 {
+		return 0
+	}
+	sort.Slice(roles, func(x, y int) bool {
+		return rolePower[roles[x]] > rolePower[roles[y]]
+	})
+	return roles[0]
```
heww: Maybe compare role one by one will be better than sort.

```go
var ( 
        highest
        highestPower
) int

for _, role := range roles {
        power := rolePower[role]
        if power > highestPower {
               highest = role
               highestPower  = power
        }
}

return highest
```

And line 715-717 not needed when compare role one by one.

API/harbor/swagger.yaml
```diff
@@ -4871,7 +4871,14 @@ definitions:
         description: Correspond to the UI about whether the project's publicity is  updatable (for UI)
       current_user_role_id:
         type: integer
-        description: The role ID of the current user who triggered the API (for UI)
+        description: The role ID with highest permission of the current user who triggered the API (for UI)
+        deprecated: true
+      current_user_role_id_list:
```
heww: I thinks `current_user_role_ids` better than `current_user_role_id_list`. Just like compare with `projects` and `project_list`, `projects` will be better.

reasonerjt: Sure, I was just thinking `current_user_role_ids` looks too similar to `current_user_role_id`

src/core/api/project.go
reasonerjt: leaving 715-717 may just make it a little bit more robust?

src/core/api/project.go
```diff
@@ -712,3 +706,39 @@ func getProjectMemberSummary(projectID int64, summary *models.ProjectSummary) {
...
+		common.RoleGuest:        20,
+		common.RoleLimitedGuest: 10,
+	}
+	var ind int
+	for _, r := range roles {
+		if _, ok := rolePower[r]; ok {
+			roles[ind] = r
+			ind++
+		}
+	}
```
heww: This will change the input value of `roles`, this is not a good practice.

If only want to ensure that the value in roles is valid, it can process in line 737.

```go
p, ok := rolePower[role]
if ok && p > highestPower {
        highest = role
        highestPower = p
}
```

reasonerjt: I was thinking it's the right thing to do because it filters out the illegal values if any, but yeah ...

PR #9820, opened 2019-11-11
Populate user groups during OIDC CLI secret verification

reasonerjt:
This commit refactors the flow to populate user info and verify CLI
secret in OIDC authentication.

It will call the `userinfo` backend of OIDC backend and fallback to
using the ID token if userinfo is not supported by the backend.

### DIFF
--- a/src/common/utils/oidc/helper.go
+++ b/src/common/utils/oidc/helper.go
@@ -31,7 +31,13 @@ import (
 	"time"
 )
 
-const googleEndpoint = "https://accounts.google.com"
+const (
+	googleEndpoint = "https://accounts.google.com"
+)
+
+type claimsProvider interface {
+	Claims(v interface{}) error
+}
 
 type providerHelper struct {
 	sync.Mutex
@@ -106,7 +112,18 @@ var insecureTransport = &http.Transport{
 // Token wraps the attributes of a oauth2 token plus the attribute of ID token
 type Token struct {
 	oauth2.Token
-	IDToken string `json:"id_token,omitempty"`
+	RawIDToken string `json:"id_token,omitempty"`
+}
+
+// UserInfo wraps the information that is extracted via token.  It will be transformed to data object that is persisted
+// in the DB
+type UserInfo struct {
+	Issuer        string   `json:"iss"`
+	Subject       string   `json:"sub"`
+	Username      string   `json:"name"`
+	Email         string   `json:"email"`
+	Groups        []string `json:"groups"`
+	hasGroupClaim bool
 }
 
 func getOauthConf() (*oauth2.Config, error) {
@@ -115,7 +132,7 @@ func getOauthConf() (*oauth2.Config, error) {
 		return nil, err
 	}
 	setting := provider.setting.Load().(models.OIDCSetting)
-	scopes := []string{}
+	scopes := make([]string, 0)
 	for _, sc := range setting.Scope {
 		if strings.HasPrefix(p.Endpoint().AuthURL, googleEndpoint) && sc == gooidc.ScopeOfflineAccess {
 			log.Warningf("Dropped unsupported scope: %s ", sc)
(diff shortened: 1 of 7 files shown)


reasonerjt:
Some rebase will be done after #9810 is merged.

src/common/utils/oidc/secret.go
```diff
@@ -60,80 +62,91 @@ func (dm *defaultManager) getEncryptKey() (string, error) {
...
+	// Persist the token to DB.  Assuming there's low possibility for concurrent refresh scenario
+	tb, err := json.Marshal(newToken)
 	if err != nil {
 		log.Warningf("Failed to encode the refreshed token, error: %v", err)
 	}
 	encToken, _ := utils.ReversibleEncrypt(string(tb), key)
-	user.Token = encToken
-	err = dao.UpdateOIDCUser(user)
+	oidcUser.Token = encToken
+	err = dao.UpdateOIDCUser(oidcUser)
```
heww: The update of oidc user  here is not needed when token is not refreshed.

```go
// RefreshToken refreshes the token passed in parameter, and return the new token.
func RefreshToken(ctx context.Context, token *Token) (*Token, error) {
	oauth, err := getOauthConf()
	if err != nil {
		log.Errorf("Failed to get OAuth configuration, error: %v", err)
		return nil, err
	}
	setting := provider.setting.Load().(models.OIDCSetting)
	ctx = clientCtx(ctx, setting.VerifyCert)
	ts := oauth.TokenSource(ctx, &token.Token)
	t, err := ts.Token()
	if err != nil {
		return nil, err
	}

	it, ok := t.Extra("id_token").(string)
	if !ok {
		log.Debugf("id_token not exist in refresh response")
	}
	return &Token{Token: *t, IDToken: it}, nil
}
```

`t, err := ts.Token()` may return old token when it's still valid.


```go
// Token returns the current token if it's still valid, else will
// refresh the current token (using r.Context for HTTP client
// information) and return the new one.
func (s *reuseTokenSource) Token() (*Token, error) {
	s.mu.Lock()
	defer s.mu.Unlock()
	if s.t.Valid() {
		return s.t, nil
	}
	t, err := s.new.Token()
	if err != nil {
		return nil, err
	}
	s.t = t
	return t, nil
}
```

reasonerjt: No longer valid after the rework.

src/common/utils/oidc/helper.go
```diff
@@ -186,55 +216,120 @@ func clientCtx(ctx context.Context, verifyCert bool) context.Context {
...
+	remote, err := userInfoFromRemote(ctx, token, setting)
+	if err != nil {
+		log.Warningf("Failed to get userInfo by calling remote userinfo endpoint, error: %v ", err)
+	}
+	if remote != nil && local != nil {
+		if remote.Subject != local.Subject {
+			return nil, fmt.Errorf("the subject from userinfo: %s does not match the subject from ID token: %s, probably a security attack happened", remote.Subject, local.Subject)
+		}
+		return mergeUserInfo(remote, local), nil
+	} else if remote != nil && local == nil {
```
stonezdj: it is better to put into mergeUserInfo()

src/common/utils/oidc/helper.go
```diff
@@ -186,55 +216,120 @@ func clientCtx(ctx context.Context, verifyCert bool) context.Context {
...
+		return remote, nil
+	} else if local != nil && remote == nil {
+		log.Debugf("Fall back to user data from ID token.")
+		return local, nil
+	}
```
stonezdj: It is better to put into mergeUserInfo()

reasonerjt: it is a private func, I don't think it will make any difference

src/core/filter/security.go
```diff
@@ -289,7 +279,11 @@ func (it *idTokenReqCtxModifier) Modify(ctx *beegoctx.Context) bool {
 		log.Warning("User matches token's claims is not onboarded.")
 		return false
 	}
-	groupNames := oidc.GroupsFromToken(claims)
+	settings, err := config.OIDCSetting()
+	if err != nil {
+		log.Errorf("Failed to get OIDC settings, error: %v", err)
```
stonezdj: If failed to get OIDC settings, should skip to PopulateGroup

src/core/filter/security.go
```diff
@@ -289,7 +279,11 @@ func (it *idTokenReqCtxModifier) Modify(ctx *beegoctx.Context) bool {
...
+	}
+	groupNames, _ := oidc.GroupsFromClaims(claims, settings.GroupsClaim)
```
stonezdj: If groupNames, ok := oidc.GroupsFromClaims(claims, settings.GroupClaim); ok {
...
}
because it is useless to process an empty groupNames

src/core/filter/security.go
reasonerjt: That's true, I'll make update.

src/core/filter/security.go
reasonerjt: This is trivial, but I'll make the change,.

PR #9825, opened 2019-11-11
Avoid to create duplicated immutable tag rules in the same project

### LINKED ISSUE #9681: Immutable rule API can create duplicate rules
We have to handle the conflict rules in the API level, not just in UI.

stonezdj:
Fix #9681, add check before insert rules

### DIFF
--- a/make/migrations/postgresql/0015_1.10.0_schema.up.sql
+++ b/make/migrations/postgresql/0015_1.10.0_schema.up.sql
@@ -44,7 +44,8 @@ CREATE TABLE immutable_tag_rule
   project_id int NOT NULL,
   tag_filter text,
   disabled BOOLEAN NOT NULL DEFAULT FALSE,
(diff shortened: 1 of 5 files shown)


src/pkg/immutabletag/dao/immutable.go
```diff
@@ -30,9 +31,30 @@ type immutableRuleDao struct{}
 func (i *immutableRuleDao) CreateImmutableRule(ir *model.ImmutableRule) (int64, error) {
 	ir.Disabled = false
 	o := dao.GetOrmer()
+	exist, err := i.checkExist(ir)
```
ywk253100: This cannot avoid the duplicated records when running in parallel. How about adding the `unique` constraint on the database table, and try to catch the dup error when inserting the row?

wy65701436: +1 to use `unique tag_filter`

PR #9833, opened 2019-11-12
Solve the problem that the style cannot be neat because the page is not refreshed

jwangyangls:
1. remove toggle public project code;
2. modify the way to show role in public project list;
3. Change the way the change detection is default and remove actively refreshed code;
environment：
https://10.192.162.161/
image


src/portal/src/app/project/list-project/list-project.component.spec.ts
```diff
@@ -14,14 +14,15 @@ import { StatisticHandler } from "../../shared/statictics/statistic-handler.serv
...
+import { delay } from 'rxjs/operators';
 describe('ListProjectComponent', () => {
     let component: ListProjectComponent;
     let fixture: ComponentFixture<ListProjectComponent>;
     const mockProjectService = {
         listProjects: () => {
             return of({
                 body: []
-            });
+            }).pipe(delay(0));
```
AllForNothing: you can open auto detect:  
        { provide: ComponentFixtureAutoDetect, useValue: true }

PR #9856, opened 2019-11-13
Refine the implementation of replication execution API

ywk253100:
Remove the duplicated code in replication execution API

### DIFF
--- a/src/core/api/replication_execution.go
+++ b/src/core/api/replication_execution.go
@@ -30,6 +30,8 @@ import (
 // ReplicationOperationAPI handles the replication operation requests
 type ReplicationOperationAPI struct {
 	BaseController
+	execution *models.Execution
+	task      *models.Task
 }
 
 // Prepare ...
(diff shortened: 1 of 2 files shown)


src/core/api/replication_execution.go
```diff
@@ -146,68 +188,21 @@ func (r *ReplicationOperationAPI) CreateExecution() {
...
-	}
-
-	if execution == nil {
-		r.SendNotFoundError(fmt.Errorf("execution %d not found", executionID))
-		return
-	}
-
-	if err := replication.OperationCtl.StopReplication(executionID); err != nil {
-		r.SendInternalServerError(fmt.Errorf("failed to stop execution %d: %v", executionID, err))
+	if err := replication.OperationCtl.StopReplication(r.execution.ID); err != nil {
```
stonezdj: Should check the r.execution is nil to avoid nil pointer

src/core/api/replication_execution.go
```diff
@@ -146,68 +188,21 @@ func (r *ReplicationOperationAPI) CreateExecution() {
...
-		return
-	}
-	if execution == nil {
-		r.SendNotFoundError(fmt.Errorf("execution %d not found", executionID))
-		return
-	}
-
 	query := &models.TaskQuery{
-		ExecutionID:  executionID,
+		ExecutionID:  r.execution.ID,
```
stonezdj: Should check the r.execution is nil to avoid nil pointer

src/core/api/replication_execution.go
```diff
@@ -232,53 +227,22 @@ func (r *ReplicationOperationAPI) ListTasks() {
...
-		return
-	}
-
-	logBytes, err := replication.OperationCtl.GetTaskLog(taskID)
+	logBytes, err := replication.OperationCtl.GetTaskLog(r.task.ID)
 	if err != nil {
 		if httpErr, ok := err.(*common_http.Error); ok {
 			if ok && httpErr.Code == http.StatusNotFound {
-				r.SendNotFoundError(fmt.Errorf("the log of task %d not found", taskID))
+				r.SendNotFoundError(fmt.Errorf("the log of task %d not found", r.task.ID))
```
stonezdj: r.task could be nil

src/core/api/replication_execution.go
ywk253100: It is checked in the prepare

src/core/api/replication_execution.go
ywk253100: It is checked in the prepare

PR #9857, opened 2019-11-13
Modify the way to handle 409 error

### LINKED ISSUE #9849: Remove the unnecessary error pop-up when click 'scan' button
while testing the 1.10 dev build, when users click the 'scan' button for the second time, if the first one is still running, there is an error pop up which is not necessary. 
Screen Shot 2019-11-13 at 8 34 14 AM

AllForNothing:
Fix #9849

### DIFF
--- a/src/portal/lib/src/tag/tag.component.ts
+++ b/src/portal/lib/src/tag/tag.component.ts
@@ -745,12 +745,9 @@ export class TagComponent implements OnInit, AfterViewInit {
   }
   // Trigger scan
   scanNow(): void {
-    if (this.selectedRow && this.selectedRow.length) {
-      this.selectedRow.forEach((data: any) => {
-        let tagId = data.name;
+    if (this.selectedRow && this.selectedRow.length === 1) {
         this.onSendingScanCommand = true;
-        this.channel.publishScanEvent(this.repoName + "/" + tagId);
-      });
+        this.channel.publishScanEvent(this.repoName + "/" + this.selectedRow[0].name);
     }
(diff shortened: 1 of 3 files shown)


src/portal/lib/src/tag/tag.component.ts
```diff
@@ -745,7 +745,7 @@ export class TagComponent implements OnInit, AfterViewInit {
   }
   // Trigger scan
   scanNow(): void {
-    if (this.selectedRow && this.selectedRow.length) {
+    if (this.selectedRow && this.selectedRow.length === 1) {
```
jwangyangls: There is a problem with the logic

src/portal/lib/src/vulnerability-scanning/result-bar-chart.component.ts
```diff
@@ -138,7 +138,11 @@ export class ResultBarChartComponent implements OnInit, OnDestroy {
                 }
             }, error => {
                 this.onSubmitting = false;
-                this.errorHandler.error(error);
+                if (error && error.error && error.error.code === 409) {
+                     console.log(error.error.message);
+                } else {
+                     this.errorHandler.error(error);
```
jwangyangls: If ’this.onSubmitting = false;‘   is placed here, it will be more reasonable.

src/portal/lib/src/tag/tag.component.ts
AllForNothing: Fixed

PR #9858, opened 2019-11-13
Refine the implementation of replication execution API

ywk253100:
Remove the duplicated code in replication execution API

### DIFF
--- a/src/core/api/replication_execution.go
+++ b/src/core/api/replication_execution.go
@@ -30,6 +30,8 @@ import (
 // ReplicationOperationAPI handles the replication operation requests
 type ReplicationOperationAPI struct {
 	BaseController
+	execution *models.Execution
+	task      *models.Task
 }
 
 // Prepare ...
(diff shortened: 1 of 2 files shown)


src/core/api/replication_execution.go
```diff
@@ -45,6 +47,46 @@ func (r *ReplicationOperationAPI) Prepare() {
...
+			if err != nil || taskID <= 0 {
+				r.SendBadRequestError(fmt.Errorf("invalid task ID: %s", taskIDStr))
+				return
+			}
+			task, err := replication.OperationCtl.GetTask(taskID)
+			if err != nil {
+				r.SendInternalServerError(fmt.Errorf("failed to get task %d: %v", taskID, err))
+				return
+			}
+			if task == nil || task.ExecutionID != executionID {
```
wy65701436: the RC of deletion a non-existing objectd should be 200 as deletion is idempotent

PR #9865, opened 2019-11-13
add quota exceed event imple

wy65701436:


src/core/middlewares/countquota/handler.go
```diff
@@ -59,6 +59,7 @@ func (h *countQuotaHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request)
 	if err := interceptor.HandleRequest(req); err != nil {
 		log.Warningf("Error occurred when to handle request in count quota handler: %v", err)
 		if _, ok := err.(quota.Errors); ok {
+			util.FireQuotaEvent(req, 1, err.Error())
```
heww: In `util.FireQuotaEvent(req, 1, err.Error())`   `1` is a magic number here, suggest to change it to a named constants.

PR #9881, opened 2019-11-15
Fix datagrid text overflow issue

jwangyangls:
Solve the problem of overlapping the line due to newly created inline characters.
environment ： https://10.192.162.161


src/portal/src/app/project/tag-retention/tag-retention.component.scss
```diff
@@ -81,3 +81,12 @@
   flex: 0 0 8.3%;
   padding-left: 0.5rem;
 }
+::ng-deep {
```
AllForNothing: using " :host::ng-deep" is a more safe way

Issue #9889, opened 2019-11-15
replication from aws will not receive all tags if there're more than 100 tags in one repo


bitsf:
replication from aws will not receive all tags if there're more than 100 tags in one repo

bitsf:
fixed by #10083

PR #9892, opened 2019-11-15
fix list repos/tags count limitation

### LINKED ISSUE #9888: replication from aws/gcr will fail if there're over 1000 repos

bitsf:
fix #9888 and #9889


MrMEEE:
I'm also facing this issue..

MrMEEE:
@bitsf Could we fix this, seems to be the trivial, as it's only fails on the order of the responses?

MrMEEE:
I have fixed the test, so it checks on the content of the array, not the order:
https://github.com/goharbor/harbor/pull/9943

PR #9950, opened 2019-11-21
Fix ui issue about the third round test

jwangyangls:
1.modify input size
2. list tag . text align
3.modify webhook test 
10.192.162.161

### DIFF
--- a/src/portal/lib/src/tag/tag.component.scss
+++ b/src/portal/lib/src/tag/tag.component.scss
@@ -254,11 +254,17 @@ clr-datagrid {
   height: 100%;
 }
 
-.ml-5 {
-  margin-left: 8px !important;
+.ml-8 {
+  margin-left: 8px;
 }
 .immutable {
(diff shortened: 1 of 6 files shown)


src/portal/lib/src/tag/tag.component.scss
```diff
@@ -254,11 +254,17 @@ clr-datagrid {
   height: 100%;
 }
 
-.ml-5 {
+.position-immutable {
   margin-left: 8px !important;
```
AllForNothing: remove !improtant

src/portal/lib/src/tag/tag.component.scss
```diff
@@ -254,11 +254,17 @@ clr-datagrid {
...
 }
 .immutable {
-  padding-right: 80px;
+  padding-right: 94px;
+  position: relative;
+  .position-immutable {
```
AllForNothing: use span instead of .position-immutable

PR #9999, opened 2019-11-26
Update minimum length of project name

### LINKED ISSUE #9946: Harbor project name length
Harbor project name requirement is as below, one char project name is not allowed
const projectNameMaxLen int = 255
const projectNameMinLen int = 2
const restrictedNameChars = `[a-z0-9]+(?:[._-][a-z0-9]+)*`

One char namespace is allowed in K8. In case there is one to one mapping between namespace and Harbor project, this will result in project creation failure. So if it is possible, we propose to allow one char project name

One to one mapping between K8 namespace and Harbor project

const projectNameMinLen int = 1

**Additional context**

reasonerjt:
This commit fixes #9946, that when creating a project the minimum length
should be 1, not 2.

This commit should be cherry picked to 1.9.x and 1.10.x branch .

We need to double check if this change impacts the creation of replication
rule.


src/portal/src/app/project/create-project/create-project.component.html
```diff
@@ -8,7 +8,7 @@ <h3 class="modal-title">{{'PROJECT.NEW_PROJECT' | translate}}</h3>
                 <div class="clr-control-container" [class.clr-error]="!isNameValid">
                     <div class="clr-input-wrapper">
                         <input type="text" id="create_project_name" [(ngModel)]="project.name" name="create_project_name" class="clr-input input-width"
-                               required pattern="^[a-z0-9]+(?:[._-][a-z0-9]+)*$" minlength="2" #projectName="ngModel" autocomplete="off"
+                               required pattern="^[a-z0-9]+(?:[._-][a-z0-9]+)*$" minlength="1" #projectName="ngModel" autocomplete="off"
```
jwangyangls: if minlength is 1 ，don't need this attribute

src/portal/src/i18n/lang/pt-br-lang.json
```diff
@@ -568,7 +567,7 @@
         "DELETE_ENABLED": "Enabled this policy",
         "REPLICATE_IMMEDIATE": "Replicar imagens existentes imediatamente",
         "NEW": "Novo",
-        "NAME_TOOLTIP": "nome da regra de replicação deve conter ao menos 2 caracteres sendo caracteres minusculos, números e  ._- e devem iniciar com letras e números.",
+        "NAME_TOOLTIP": "nome da regra de replicação deve conter ao menos 1~255 caracteres sendo caracteres minusculos, números e  ._- e devem iniciar com letras e números.",
```
jwangyangls: Need recovery

Issue #9236, opened 2019-09-24
fix Harbor version for Admiral


xaleeks:
Use fixed version 1.7.x for Admiral and retire admiral-related code from Harbor code base starting with v1.10 so do not get flooded with  admiral logs for non-admiral based deployments.

reasonerjt:
@renmaosheng Does it mean we want to remove admiral related code from Harbor in 1.10?

renmaosheng:
Yes, if time allows, let's remove it in v1.10.

michmike:
added this to 1.10 backlog

reasonerjt:
This involves both backend and UI code, we'll target to 1.11 to deliver the complete removal.

xaleeks:
please make sure this is done in 2.0, in first patch release if possible @reasonerjt

reasonerjt:
Fixed by https://github.com/goharbor/harbor/pull/10034 (in master)

We also have downgrade the log-level of some admiral related message in the 1.10 and 1.9 patch releases.

Closing.

PR #10013, opened 2019-11-26
perf(rbac): add permission evaluator to improve performance

heww:
1. Introduce Evaluator interface which do the permission checking.
2. `rbac` and `namespace` are implemented the Evaluator interface.
3. Do permission checking in security context by `Evaluator`.
4. Cache the regexp in rbac evaluator for casbin.
5. Cache project evaluator in namespace evaluator to improve performance
when do permission checking in security context for same namespace
multiple times.

After this refactor, the following is the performance for the /api/users/current/permissions of 90%ile (sec) with 80 concurrency:

User | Before | After
-- | -- | --
System admin | 45.026 | 3.017
Project admin | 48.802 | 5.304


src/common/security/admiral/context.go
```diff
@@ -71,19 +75,20 @@ func (s *SecurityContext) IsSolutionUser() bool {
...
+		s.evaluator = rbac.NewNamespaceEvaluator("project", func(ns rbac.Namespace) rbac.Evaluator {
 			projectID := ns.Identity().(int64)
-			isPublicProject, _ := s.pm.IsPublic(projectID)
-			projectNamespace := rbac.NewProjectNamespace(projectID, isPublicProject)
-			user := project.NewUser(s, projectNamespace, s.GetProjectRoles(projectID)...)
-			return rbac.HasPermission(user, resource, action)
-		}
-	}
+			proj, err := s.pm.Get(projectID)
+			if err != nil || proj == nil {
```
reasonerjt: Let's explicitly log the error, b/c it seems all the subsequent `Can` will return false.

src/common/rbac/casbin.go
```diff
@@ -53,12 +57,86 @@ e = some(where (p.eft == allow)) && !some(where (p.eft == deny))
 m = g(r.sub, p.sub) && keyMatch2(r.obj, p.obj) && (r.act == p.act || p.act == '*')
 `
 
-// keyMatch2 determines whether key1 matches the pattern of key2, its behavior most likely the builtin KeyMatch2
-// except that the match of ("/project/1/robot", "/project/1") will return false
-func keyMatch2(key1 string, key2 string) bool {
-	key2 = strings.Replace(key2, "/*", "/.*", -1)
+type regexpEntry struct {
+	data      *regexp.Regexp
+	isAliveAt int64
```
reasonerjt: This is trivial as it's not exported, but why not use `time.Time` and call `isAliveAt` as `expiration`

src/common/rbac/evaluator.go
```diff
@@ -0,0 +1,83 @@
...
+func (e *rbacEvaluator) HasPermission(resource Resource, action Action) bool {
+	e.once.Do(func() {
+		e.enforcer = enforcerForUser(e.user)
+	})
+
+	return e.enforcer.Enforce(e.user.GetUserName(), resource.String(), action.String())
+}
+
+// NewEvaluator returns Evaluator for the rbac user
+func NewEvaluator(user User) Evaluator {
```
reasonerjt: NewRbacEvaluator Or call it `userEvaluator` and `NewUserEvaluator`?

src/common/security/admiral/context.go
```diff
@@ -71,19 +75,20 @@ func (s *SecurityContext) IsSolutionUser() bool {
 
 // Can returns whether the user can do action on resource
 func (s *SecurityContext) Can(action rbac.Action, resource rbac.Resource) bool {
```
wy65701436: can we set false always for admiral?

src/common/security/admiral/context.go
heww: Fixed, the error message was send to log when failed to get project info.

src/common/rbac/casbin.go
heww: Changed to `time.Time` and renamed to `expiratedAt`

src/common/rbac/evaluator.go
heww: Changed

src/common/security/admiral/context.go
heww: To keep admiral works as before although we do not need it now. Let's remove it at #9236

src/common/rbac/casbin.go
```diff
@@ -53,12 +56,85 @@ e = some(where (p.eft == allow)) && !some(where (p.eft == deny))
...
+		s.entries.Delete(key)
+	}
+}
+
+var (
+	store = &regexpStore{lifeDuration: time.Hour}
+)
 
+func init() {
+	startRegexpStorePurging(store, time.Hour)
```
wy65701436: enlarge it? To make it simple, remvoe the expiratedAt, just force refresh the cache?

src/common/rbac/casbin.go
heww: regexEntry was removed because we force to purge the cache every 24 hours

src/common/rbac/casbin.go
heww: Force to purge the cache every 24 hours now.

PR #10031, opened 2019-11-28
Fix bug when trying to get the external URL

ywk253100:
Fix bug when trying to get the external URL


reasonerjt:
Could you add a UT

reasonerjt:
@wy65701436 what's blocking you from reviewing this PR?

PR #10034, opened 2019-11-28
Clean up admiral-related code

ywk253100:
Clean up admiral-related code as it's useless


src/common/dao/repository.go
```diff
@@ -183,8 +183,6 @@ func repositoryQueryConditions(query ...*models.RepositoryQuery) (string, []inte
 	}
 
 	if len(q.ProjectName) > 0 {
-		// use "like" rather than "table joining" because that
```
reasonerjt: Do you recall the background of this comment?
Do we need to move back to "table joining"?

reasonerjt:
Thanks a lot!
I think we can merge it now but one more comment 
Do we still need to keep the concept project mamager?  
Do we still need it in security ctx or would it be better we replace it with more flexible approach?

ywk253100:
Open issue https://github.com/goharbor/harbor/issues/10140 to the follow up works

PR #10047, opened 2019-11-29
optimize the makefile process

bitsf:
Change-Id: Ibe806972a19cd69bfd90be051cdc340c4d7c6afb


reasonerjt:
Could we add more meaningful messages to the commit?

Makefile
```diff
@@ -158,23 +160,24 @@ GOIMAGEBUILDCMD=/usr/local/go/bin/go build -mod vendor
...
+GOBUILDPATH_MIGRATEPATCH=$(GOBUILDPATHINCONTAINER)/src/cmd/migrate-patch
+GOBUILDMAKEPATH=make
 GOBUILDMAKEPATH_CORE=$(GOBUILDMAKEPATH)/photon/core
 GOBUILDMAKEPATH_JOBSERVICE=$(GOBUILDMAKEPATH)/photon/jobservice
 GOBUILDMAKEPATH_REGISTRYCTL=$(GOBUILDMAKEPATH)/photon/registryctl
 GOBUILDMAKEPATH_NOTARY=$(GOBUILDMAKEPATH)/photon/notary
 
 # binary
-CORE_BINARYPATH=$(MAKEDEVPATH)/core
+CORE_BINARYPATH=$(BUILDPATH)/$(GOBUILDMAKEPATH_CORE)
```
wy65701436: remove `MAKEDEVPATH` also?

Makefile
```diff
@@ -254,14 +257,14 @@ PACKAGE_OFFLINE_PARA=-zcvf harbor-offline-installer-$(PKGVERSIONTAG).tgz \
 					$(HARBORPKG)/prepare \
 					$(HARBORPKG)/LICENSE $(HARBORPKG)/install.sh \
 					$(HARBORPKG)/common.sh \
-					$(HARBORPKG)/harbor.yml
+					$(HARBORPKG)/harbor.yml.tmpl
```
wy65701436: we need to add the `harbor.yml.tmpl` existence validation in `prepare`

bitsf: added check for harbor.yml in prepare file

Makefile
bitsf: removed

Makefile
```diff
@@ -64,13 +64,13 @@
...
 BUILDPATH=$(CURDIR)
 MAKEPATH=$(BUILDPATH)/make
-MAKEDEVPATH=$(MAKEPATH)/dev
 MAKE_PREPARE_PATH=$(MAKEPATH)/photon/prepare
 SRCPATH=./src
 TOOLSPATH=$(BUILDPATH)/tools
 CORE_PATH=$(BUILDPATH)/src/core
 PORTAL_PATH=$(BUILDPATH)/src/portal
 CHECKENVCMD=checkenv.sh
+OSNAME=$(shell uname)
```
wy65701436: I don't see where is used for?

make/photon/prepare/utils/nginx.py
```diff
@@ -59,6 +59,7 @@ def render_nginx_template(config_dict):
             nginx_conf,
             uid=DEFAULT_UID,
             gid=DEFAULT_GID,
+            https_redirect='$host' + ('https_port' in config_dict and (":" + str(config_dict['https_port'])) or ""),
```
ninjadq: why add https code here?

make/photon/prepare/templates/nginx/nginx.https.conf.jinja
```diff
@@ -162,6 +162,6 @@ http {
   server {
       listen 8080;
       #server_name harbordomain.com;
-      return 308 https://$host$request_uri;
+      return 308 https://{{https_redirect}}$request_uri;
```
ninjadq: why add https code here?

Makefile
bitsf: I intent to fix "sed -i" on MacOS with this env, but finally not use it.
Removed it

PR #10126, opened 2019-12-04
fix connection leak

lazybetrayer:
Signed-off-by: Wang Zhen <lazybetrayer@gmail.com>

### DIFF
--- a/src/common/utils/registry/repository.go
+++ b/src/common/utils/registry/repository.go
@@ -277,9 +277,9 @@ func (r *Repository) MountBlob(digest, from string) error {
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


src/common/utils/registry/repository.go
```diff
@@ -277,9 +277,9 @@ func (r *Repository) MountBlob(digest, from string) error {
 	if err != nil {
 		return err
 	}
+	defer resp.Body.Close()
```
reasonerjt: with this the connection will be closed if the status code is 200?

lazybetrayer: This maybe not a connection leak. But we should call `resp.Body.Close()` regardless of the status code.

PR #10135, opened 2019-12-04
upgrade clair to v2.1.1

bitsf:
Change-Id: Idb2ad0470a51666d75895d8c5e68d80a67e05276


make/photon/clair/Dockerfile.binary
```diff
@@ -1,7 +1,7 @@
 FROM golang:1.13.4
 
-ADD .   /go/src/github.com/coreos/clair/
-WORKDIR /go/src/github.com/coreos/clair/
+ADD .   /go/src/github.com/quay/clair/
+WORKDIR /go/src/github.com/quay/clair/
 
-RUN export CLAIR_VERSION=$(git describe --tag --always --dirty) && \
-	go build -ldflags "-X github.com/coreos/clair/pkg/version.Version=$CLAIR_VERSION" github.com/coreos/clair/cmd/clair
+RUN export CLAIR_VERSION=$(git describe --tag --always --dirty) GO111MODULE=on && \
```
reasonerjt: Any particular reason we set `GO111MODULE` to `on`?

bitsf: The default behavior auto is not same for golang 1.12 and 1.13, so we set this to on in order to auto download the dependencies

PR #10136, opened 2019-12-04
Get rid of case-sensitivity in authproxy setting

reasonerjt:
This commit removes the attribute to control case-sensitivity from
authproxy setting.
The result in token review status will be used as the single source of
truth, regardless the case of the letters in group names and user names.


reasonerjt:
Travis failed due to env issue
Closed and re-opened the PR to trigger the CI again.

PR #10138, opened 2019-12-04
Create the models for OCI supporting

ywk253100:
This commits does some basic init work for supporting OCI:
1. Create the artifact and tag model
2. Create database tables

### DIFF
--- a/make/migrations/postgresql/0030_1.11.0_schema.up.sql
+++ b/make/migrations/postgresql/0030_1.11.0_schema.up.sql
@@ -1,0 +1,42 @@
+/* TODO remove the table artifact_2 and use the artifact instead after finishing the upgrade work */
+CREATE TABLE artifact_2
+(
+  id            SERIAL PRIMARY KEY NOT NULL,
+  /* image, chart, etc */
+  type          varchar(255),
+  media_type    varchar(255),
+  manifest_media_type varchar(255),
+  project_id    int NOT NULL,
+  repository_id int NOT NULL,
+  digest        varchar(255) NOT NULL,
+  size          bigint,  
+  push_time     timestamp default CURRENT_TIMESTAMP,
+  platform      varchar(255),
+  extra_attrs   text,
+  annotations   jsonb,
+  /* when updating the data the revision MUST be checked and updated */
+  revision      varchar(64) NOT NULL,
+  CONSTRAINT unique_artifact_2 UNIQUE (repository_id, digest)
+);
+
+CREATE TABLE tag
+(
+  id            SERIAL PRIMARY KEY NOT NULL,
+  repository_id int NOT NULL,
+  artifact_id   int NOT NULL,
+  name          varchar(255),
+  push_time     timestamp default CURRENT_TIMESTAMP,
+  pull_time     timestamp,
+  /* when updating the data the revision MUST be checked and updated */
+  revision      varchar(64) NOT NULL,
+  CONSTRAINT unique_tag UNIQUE (repository_id, name)
+);
+
+/* artifact_reference records the child artifact referenced by parent artifact */
+CREATE TABLE artifact_reference
+(
+  id            SERIAL PRIMARY KEY NOT NULL,
+  parent_id   int NOT NULL,
+  child_id  int NOT NULL,
+  CONSTRAINT unique_reference UNIQUE (parent_id, child_id)
+);
--- a/src/pkg/artifact/manager/dao/model.go
+++ b/src/pkg/artifact/manager/dao/model.go
@@ -1,0 +1,78 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dao
+
+import (
+	"time"
+
+	"github.com/astaxie/beego/orm"
+)
+
+func init() {
+	orm.RegisterModel(&Artifact{})
+	orm.RegisterModel(&Tag{})
+	orm.RegisterModel(&ArtifactReference{})
+}
+
+// Artifact model in database
+type Artifact struct {
+	ID                int64     `orm:"pk;auto;column(id)"`
+	Type              string    `orm:"column(type)"`                // image or chart
+	MediaType         string    `orm:"column(media_type)"`          // the media type of artifact
+	ManifestMediaType string    `orm:"column(manifest_media_type)"` // the media type of manifest/index
+	ProjectID         int64     `orm:"column(project_id)"`          // needed for quota
+	RepositoryID      int64     `orm:"column(repository_id)"`
+	Digest            string    `orm:"column(digest)"`
+	Size              int64     `orm:"column(size)"`
+	PushTime          time.Time `orm:"column(push_time)"`
+	Platform          string    `orm:"column(platform)"`                // json string
+	ExtraAttrs        string    `orm:"column(extra_attrs)"`             // json string
+	Annotations       string    `orm:"column(annotations);type(jsonb)"` // json string
+	Revision          string    `orm:"column(revision)"`                // record data revision, when updating the data the revision MUST be checked and updated
+}
+
+// TableName for artifact
+func (a *Artifact) TableName() string {
+	// TODO use "artifact" after finishing the upgrade/migration work
+	return "artifact_2"
+}
+
+// Tag model in database
+type Tag struct {
+	ID           int64     `orm:"pk;auto;column(id)"`
+	RepositoryID int64     `orm:"column(repository_id)"` // tags are the resources of repository, one repository only contains one same name tag
+	ArtifactID   int64     `orm:"column(artifact_id)"`   // the artifact ID that the tag attaches to, it changes when pushing a same name but different digest artifact
+	Name         string    `orm:"column(name)"`
+	PushTime     time.Time `orm:"column(push_time)"`
+	PullTime     time.Time `orm:"column(pull_time)"`
+	Revision     string    `orm:"column(revision)"` // record data revision, when updating the data the revision MUST be checked and updated
+}
+
+// TableName for tag
+func (t *Tag) TableName() string {
+	return "tag"
+}
+
+// ArtifactReference records the child artifact referenced by parent artifact
+type ArtifactReference struct {
+	ID       int64 `orm:"pk;auto;column(id)"`
+	ParentID int64 `orm:"column(parent_id)"`
+	ChildID  int64 `orm:"column(child_id)"`
+}
+
+// TableName for artifact reference
+func (a *ArtifactReference) TableName() string {
+	return "artifact_reference"
+}
--- a/src/pkg/artifact/model/model.go
+++ b/src/pkg/artifact/model/model.go
@@ -1,0 +1,165 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package model
+
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/goharbor/harbor/src/common/models"
+	"github.com/goharbor/harbor/src/common/utils/log"
+	"github.com/goharbor/harbor/src/pkg/artifact/manager/dao"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+// Artifact is the abstract object managed by Harbor. It hides the
+// underlying concrete detail and provides an unified artifact view
+// for all users.
+type Artifact struct {
+	ID                int64
+	Type              string // image, chart, etc
+	MediaType         string // the media type of artifact. Mostly, it's the value of `manifest.config.mediatype`
+	ManifestMediaType string // the media type of manifest/index
+	Repository        *models.RepoRecord
+	Tags              []*Tag // the list of tags that attached to the artifact
+	Digest            string
+	Size              int64
+	PushTime          time.Time
+	Platform          *v1.Platform               // when the parent of the artifact is an index, populate the platform information here
+	ExtraAttrs        map[string]interface{}     // only contains the simple attributes specific for the different artifact type, most of them should come from the config layer
+	SubResourceLinks  map[string][]*ResourceLink // the resource link for build history(image), values.yaml(chart), dependency(chart), etc
+	Annotations       map[string]string
+	References        []int64 // child artifacts referenced by the parent artifact if the artifact is an index
+	Revision          string  // record data revision
+	// TODO: As the labels and signature aren't handled inside the artifact module,
+	// we should move it to the API level artifact model rather than
+	// keeping it here. The same to scan information
+	// Labels                  []*models.Label
+	// Signature               *Signature                 // add the signature in the artifact level rather than tag level as we cannot make sure the signature always apply to tag
+}
+
+// From converts the database level artifact to the business level object
+func (a *Artifact) From(art *dao.Artifact) {
+	a.ID = art.ID
+	a.Type = art.Type
+	a.MediaType = art.MediaType
+	a.ManifestMediaType = art.ManifestMediaType
+	a.Repository = &models.RepoRecord{
+		ProjectID:    art.ProjectID,
+		RepositoryID: art.RepositoryID,
+	}
+	a.Digest = art.Digest
+	a.Size = art.Size
+	a.PushTime = art.PushTime
+	a.ExtraAttrs = map[string]interface{}{}
+	a.Annotations = map[string]string{}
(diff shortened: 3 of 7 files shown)


src/pkg/artifact/model/model.go
```diff
@@ -0,0 +1,68 @@
...
+	Signature             *Signature                 `json:"signature"`               // add the signature in the artifact level rather than tag level as we cannot make sure the signature always apply to tag
+	ExtraAttrs            map[string]interface{}     `json:"extra_attrs"`             // only contains the simple attributes specific for the different artifact type, most of them should come from the config layer
+	AdditionResourceLinks map[string][]*ResourceLink `json:"addition_resource_links"` // the resource link for build history(image), scan result(image), values.yaml(chart), dependency(chart), etc
+	Annotations           map[string]string          `json:"annotations"`
+	References            []*Reference               `json:"references"` // child artifacts referenced by the parent artifact if the artifact is an index
+}
+
+// ResourceLink is a link via that a resource can be fetched
+type ResourceLink struct {
+	HREF string `json:"href"`
```
reasonerjt: Will we potentially support relative and absolute URIs?
so probably add a flag or an attribute called `type`?

src/pkg/artifact/model/model.go
```diff
@@ -0,0 +1,68 @@
...
+	"time"
+
+	"github.com/goharbor/harbor/src/common/models"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+// Artifact is the abstract object managed by Harbor. It hides the
+// underlying concrete detail and provides an unified artifact view
+// for all users.
+type Artifact struct {
```
reasonerjt: Wouldn't such models be generated by swagger?

src/pkg/artifact/manager/dao/model.go
```diff
@@ -0,0 +1,143 @@
+// Copyright Project Harbor Authors
```
reasonerjt: Let's sync about the programming model and path conventions or put that in the proposal or somewhere?

make/migrations/postgresql/0030_1.11.0_schema.up.sql
```diff
@@ -0,0 +1,26 @@
+/* TODO remove the table artifact_2 and use the artifact instead after finishing the upgrade work */
+CREATE TABLE artifact_2
```
reasonerjt: We need to decide the uuid .vs. numeric id stuff...

src/pkg/artifact/model/model.go
ywk253100: No, the API models should be generated by swagger

make/migrations/postgresql/0030_1.11.0_schema.up.sql
ywk253100: There is already an artifact table introduced by quota, it's numeric ID, the `artifact_2` is a temp one, so it's ID is numeric. And for the consistence, I think the numeric ID for table `tag` is better

src/pkg/artifact/model/model.go
ywk253100: Updated.

make/migrations/postgresql/0030_1.11.0_schema.up.sql
wy65701436: we should consider the migration issue. 
1 from non-artifact scheme to artifact_2
2 from artifact to artifact_2.

make/migrations/postgresql/0030_1.11.0_schema.up.sql
```diff
@@ -0,0 +1,37 @@
...
+  repository_id int NOT NULL,
+  artifact_id   int NOT NULL,
+  name          varchar(255),
+  upload_time   timestamp default CURRENT_TIMESTAMP,
+  latest_download_time timestamp,
+  CONSTRAINT unique_tag UNIQUE (repository_id, name)
+);
+
+/* reference records the child artifact referenced by parent artifact */
+CREATE TABLE reference
```
wy65701436: how about to name it as artifact_reference

make/migrations/postgresql/0030_1.11.0_schema.up.sql
```diff
@@ -0,0 +1,37 @@
...
+CREATE TABLE artifact_2
+(
+  id            SERIAL PRIMARY KEY NOT NULL,
+  /* image, chart, etc */
+  type          varchar(255),
+  project_id    int NOT NULL,
+  repository_id int NOT NULL,
+  media_type    varchar(255),
+  digest        varchar(255) NOT NULL,
+  size          int,
```
wy65701436: the `size` is stand for the artifact itself or the sum of its blobs?

wy65701436: use `bigint` instead?

make/migrations/postgresql/0030_1.11.0_schema.up.sql
ywk253100: Updated

make/migrations/postgresql/0030_1.11.0_schema.up.sql
ywk253100: The `Artifact` is the concept of Harbor, not the manifest or index, so it is the size of image or chart

ywk253100: Updated

make/migrations/postgresql/0030_1.11.0_schema.up.sql
ywk253100: What's the meaning of `from non-artifact scheme to artifact_2`?

ywk253100: @wy65701436 Will cover this in the upgrade&migration part

ywk253100: @reasonerjt We can update the type of ID when we decide to use uuid or numeric ID, this is a temp table

src/pkg/artifact/manager/dao/model.go
ywk253100: updated

src/pkg/artifact/model/model.go
```diff
@@ -0,0 +1,73 @@
...
+	ID                      int64
+	Type                    string // image, chart, etc
+	Repository              *models.RepoRecord
+	Tags                    []*Tag // the list of tags that attached to the artifact
+	MediaType               string // the specific media type
```
steven-zou: Still thinking the positions of `Type` and `MediaType`. `Type` is generated based on `MediaType`, so maybe no need to store both two, just a type function is ok to get `Type` from `MediaType`.

src/pkg/artifact/model/model.go
```diff
@@ -0,0 +1,73 @@
...
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+// Artifact is the abstract object managed by Harbor. It hides the
+// underlying concrete detail and provides an unified artifact view
+// for all users.
+type Artifact struct {
+	ID                      int64
+	Type                    string // image, chart, etc
+	Repository              *models.RepoRecord
```
steven-zou: Convention pattern, `models.RepoRecord` => `repository.X` / `registry.Repository`?

src/pkg/artifact/model/model.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/goharbor/harbor/src/common/models"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+// Artifact is the abstract object managed by Harbor. It hides the
+// underlying concrete detail and provides an unified artifact view
+// for all users.
+type Artifact struct {
+	ID                      int64
```
steven-zou: So UUID string has been out of the table? number ID finalized?

src/pkg/artifact/model/model.go
```diff
@@ -0,0 +1,73 @@
...
+	Type                    string // image, chart, etc
+	Repository              *models.RepoRecord
+	Tags                    []*Tag // the list of tags that attached to the artifact
+	MediaType               string // the specific media type
+	Digest                  string
+	Size                    int64
+	UploadTime              time.Time
```
steven-zou: Do we care about the original `Creation Time`?

src/pkg/artifact/model/model.go
```diff
@@ -0,0 +1,73 @@
...
+type ResourceLink struct {
+	HREF     string
+	Absolute bool // specify the href is an absolute URL or not
+}
+
+// Reference records the child artifact information referenced by
+// other parent artifact
+type Reference struct {
+	ArtifactID int64
+	Platform   *v1.Platform
```
steven-zou: Why we define such special info `*v1.Platform` here?

src/pkg/artifact/model/model.go
```diff
@@ -0,0 +1,73 @@
...
+// type Signature struct {
+// 	Signatures map[string]bool // tag: signed or not
+// }
+
+// Tag belongs to one repository and can only be attached to a single
+// one artifact under the repository
+type Tag struct {
+	ID                 int64
+	Name               string
+	UploadTime         time.Time
```
steven-zou: As the tag is just a `label` of the artifact, I think the `AttachedTime` or `CreationTime` may be better than `UploadTime ` which should be used in `Artifact` which is uploaded to the registry.

src/pkg/artifact/model/model.go
```diff
@@ -0,0 +1,73 @@
...
+	LatestDownloadTime time.Time
```
steven-zou: For `upload`/`download`,  as we're trying to follow the OCI spec, so every artifact can be Pulled/Pushed/Deleted and then replacing those times with `PullTime` or `PushTime` are more accurate.

src/pkg/artifact/model/model.go
```diff
@@ -0,0 +1,73 @@
...
+	"time"
+
+	"github.com/goharbor/harbor/src/common/models"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+// Artifact is the abstract object managed by Harbor. It hides the
+// underlying concrete detail and provides an unified artifact view
+// for all users.
+type Artifact struct {
```
steven-zou: Have you considered if there are some common harbor properties in the artifact? Not just OCI related things? e.g: ownership, immutability?

Just a reminder, it does not mean I have concrete ideas already.

src/pkg/artifact/model/model.go
```diff
@@ -0,0 +1,73 @@
...
+	ID                      int64
+	Type                    string // image, chart, etc
+	Repository              *models.RepoRecord
+	Tags                    []*Tag // the list of tags that attached to the artifact
+	MediaType               string // the specific media type
+	Digest                  string
+	Size                    int64
+	UploadTime              time.Time
+	ExtraAttrs              map[string]interface{}     // only contains the simple attributes specific for the different artifact type, most of them should come from the config layer
```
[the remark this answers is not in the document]
ywk253100: Don't get your point, there is a `AdditionalResourceLinks`, is this what you talking about?

src/pkg/artifact/model/model.go
```diff
@@ -0,0 +1,73 @@
...
+	AdditionalResourceLinks map[string][]*ResourceLink // the resource link for build history(image), values.yaml(chart), dependency(chart), etc
+	Annotations             map[string]string
+	References              []*Reference // child artifacts referenced by the parent artifact if the artifact is an index
+	// TODO: As the labels and signature aren't handled inside the artifact module,
```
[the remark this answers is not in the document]
ywk253100: If the dao and controller model are same, we should share them, so at lease 2 models: API and controller

src/pkg/artifact/model/model.go
ywk253100: Agree. This is tech debt, will rename it when refactoring repository part

src/pkg/artifact/model/model.go
ywk253100: Not yet, will discuss later and update it in other PR when finalized.

src/pkg/artifact/model/model.go
ywk253100: It's the one property of the index: https://github.com/opencontainers/image-spec/blob/master/image-index.md#image-index-property-descriptions
we may show them to users

src/pkg/artifact/model/model.go
ywk253100: You mean the build time if the artifact is an image? It is in the `ExtraAttrs`

src/pkg/artifact/model/model.go
ywk253100: For some kinds of artifacts, such as helm chart and single manifest image, the type can be generated by the `manifest.config.mediatype`, but for index there is still no stardard to tell us that what the artifact is. So the media type here is the media type of top level manifest, not the config's. Just in case we need it

ywk253100: Add both `MediaType` and `ManifestMediaType`

src/pkg/artifact/model/model.go
ywk253100: Move the `Platform` into `Artifact`

src/pkg/artifact/model/model.go
ywk253100: Updated

src/pkg/artifact/model/model.go
ywk253100: Updated

src/pkg/artifact/manager/dao/model.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// TableName for artifact
+func (a *Artifact) TableName() string {
+	// TODO use "artifact" after finishing the upgrade/migration work
+	return "artifact_2"
+}
+
+// To converts the artifact to the business level object
+func (a *Artifact) To() *model.Artifact {
```
steven-zou: One more comment here: with this `To` function, lower package will refer to the top-level package? Maybe the from convert function can be defined in the top-level package?

ywk253100: Updated

make/migrations/postgresql/0030_1.11.0_schema.up.sql
```diff
@@ -0,0 +1,40 @@
+/* TODO remove the table artifact_2 and use the artifact instead after finishing the upgrade work */
+CREATE TABLE artifact_2
+(
+  id            SERIAL PRIMARY KEY NOT NULL,
+  /* image, chart, etc */
+  type          varchar(255),
```
reasonerjt: I don't think we should store one semantic attribute in 2 columns

make/migrations/postgresql/0030_1.11.0_schema.up.sql
```diff
@@ -0,0 +1,40 @@
...
+  media_type    varchar(255),
+  manifest_media_type varchar(255),
+  project_id    int NOT NULL,
+  repository_id int NOT NULL,
+  digest        varchar(255) NOT NULL,
+  size          bigint,  
+  push_time     timestamp default CURRENT_TIMESTAMP,
+  platform      varchar(255),
+  extra_attrs   text,
+  annotations   text,
```
reasonerjt: This may be trivial

Let's think if we need to query based on an annotation, i.e. will we need to query artifacts with certain annotations "xxxx"?

If the answer is yes we need to use `JSONB` or  create a new table as we did for labels

make/migrations/postgresql/0030_1.11.0_schema.up.sql
```diff
@@ -0,0 +1,40 @@
...
+  revision      int NOT NULL,
```
reasonerjt: let's use varchar to store the revision and explicitly document that when updating the data the revision MUST be checked and updated.

make/migrations/postgresql/0030_1.11.0_schema.up.sql
```diff
@@ -0,0 +1,40 @@
...
+
+CREATE TABLE tag
+(
+  id            SERIAL PRIMARY KEY NOT NULL,
+  repository_id int NOT NULL,
+  artifact_id   int NOT NULL,
+  name          varchar(255),
+  push_time     timestamp default CURRENT_TIMESTAMP,
+  pull_time     timestamp,
+  revision      int NOT NULL,
```
reasonerjt: same comment for `revision` column

make/migrations/postgresql/0030_1.11.0_schema.up.sql
```diff
@@ -0,0 +1,40 @@
...
+  CONSTRAINT unique_tag UNIQUE (repository_id, name)
+);
+
+/* artifact_reference records the child artifact referenced by parent artifact */
+CREATE TABLE artifact_reference
+(
+  id            SERIAL PRIMARY KEY NOT NULL,
+  artifact_id   int NOT NULL,
```
reasonerjt: let's clarify which column is for parent artifact which column point to child.

renmaosheng: could we call the parent id directly?

make/migrations/postgresql/0030_1.11.0_schema.up.sql
ywk253100: It's easy for query. Thinking about listing only images, if we don't store the `type`, we need to get the corresponding media types from `Artifact Resolver` first(there are several media types for image), and use the sql something like `select * from artifact where media_type in ('media_type_1', 'media_type_2')` to query

And we need to define methods for `Artifact Resolver` to get type according to the media type and to get the media types according to type.

But if we store it, all these convert works can be avoid

make/migrations/postgresql/0030_1.11.0_schema.up.sql
ywk253100: Updated

make/migrations/postgresql/0030_1.11.0_schema.up.sql
ywk253100: done

make/migrations/postgresql/0030_1.11.0_schema.up.sql
ywk253100: Updated

make/migrations/postgresql/0030_1.11.0_schema.up.sql
ywk253100: Updated to jsonb

Issue #10094, opened 2019-12-03
scan job item listing should not be variable


xaleeks:
please see issue described on the screenshot

<img width="1523" alt="Screen Shot 2019-11-19 at 4 13 18 AM" src="https://user-images.githubusercontent.com/51319901/70020831-4bbbe100-155c-11ea-8bf1-c6b04abcdf0b.png">

reasonerjt:
This is irrelevant to scan.

We should make sure the order of the tags stay unchanged.

PR #10165, opened 2019-12-05
Sort the tag before returning the list when calling API

ywk253100:
Sort the tag before returning the list when calling API list tag API


xaleeks:
is this in reference to the issue I filed that order is random? please link to it if it is.

ywk253100:
Fixes https://github.com/goharbor/harbor/issues/10094

PR #10171, opened 2019-12-06
Add loading in summary and immutable page

jwangyangls:


xaleeks:
can't really understand work done from this description?  can we add more detail in the future to these kinds of tickets? @jwangyangls

PR #10207, opened 2019-12-09
[WIP]add storage driver-related code into registryctl

wy65701436:
add storage driver-related code into registryctl

The storage source is copied from distribution, the registry of harbor is using.
In order to do the non-readonly GC, harbor has to have the capability to delete blobs directly, which will call for the storage source code.


reasonerjt:
Is it possible not to copy the code into the repo? 

Is there something blocking you from importing it?

PR #10242, opened 2019-12-12
Stastics API should handle group members

### LINKED ISSUE #10230: statistic API is not handling the group membership correctly.
The statistic API returns the number of public and private projects.
While counting the project numbers it does not take into account the groups membership.

i.e. if `userA` is a member of `groupA` and `projectA` only has `groupA` as a member, the API will ignore this project.

reasonerjt:
statistic API use security Context to list project rather than calling
projectmanager directly, such that the group membership will be taken
into account.
fixes #10230

It should be cherry picked to 1.9.x and 1.10.x branches

### DIFF
--- a/src/core/api/statistic.go
+++ b/src/core/api/statistic.go
@@ -67,7 +67,7 @@ func (s *StatisticAPI) Get() {
 	if len(pubProjs) == 0 {
 		statistic[PubRC] = 0
 	} else {
-		ids := []int64{}
+		ids := make([]int64, 0)
 		for _, p := range pubProjs {
 			ids = append(ids, p.ProjectID)
 		}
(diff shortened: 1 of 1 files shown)


src/core/api/statistic.go
```diff
@@ -101,30 +101,26 @@ func (s *StatisticAPI) Get() {
...
-		value := false
-		result, err := s.ProjectMgr.List(&models.ProjectQueryParam{
-			Public: &value,
-			Member: &models.MemberQuery{
-				Name: s.username,
-			},
-		})
+		// including the public ones
+		myProjects, err := s.SecurityCtx.GetMyProjects()
+		privProjectIDs := make([]int64, 0)
```
ywk253100: Put this after the error checking?

PR #10248, opened 2019-12-13
Define the interface for artifact manager

ywk253100:
1. Define the interface for artifact manager
    2. Provide a null implementation for artifact manager


src/pkg/artifact/controller.go
```diff
@@ -13,3 +13,103 @@
...
+	Delete(ctx context.Context, id int64) error
+	// GetSubResource returns the sub resource of the artifact
+	// The sub resource is different according to the artifact type:
+	// build history for image; values.yaml, readme and dependencies for chart, etc
+	GetSubResource(ctx context.Context, artifactID int64, resource string) ([]byte, error)
+	// Attach the tags to the artifact. If the tag doesn't exist, will create it.
+	// If the tag is already attached to another artifact, will detach it first
+	Attach(ctx context.Context, artifact *model.Artifact, tags ...string) error
+	// UpdateTag updates the tag
+	UpdateTag(ctx context.Context, tag *model.Tag) error
```
wy65701436: what's the updateTag for?

src/pkg/artifact/model/model.go
```diff
@@ -163,3 +163,18 @@ func (t *Tag) To() *dao.Tag {
...
+	Digest       string
+	Tag          string
+	Type         string // image, chart, etc
+	Page         int64
+	PageSize     int64
+}
+
+// Option for pruning artifact records
+type Option struct {
+	KeepUntagged bool // keep the untagged artifacts or not
```
wy65701436: it's a global variable?

src/pkg/artifact/controller.go
```diff
@@ -13,3 +13,103 @@
...
+	// DeleteTag deletes the specified tag
+	DeleteTag(ctx context.Context, tagID int64) error
+	// Prune removes the useless artifact records. The underlying registry data will
+	// be removed during garbage collection
+	Prune(context.Context, *model.Option) error
```
wy65701436: Is it called only in GC?

src/pkg/artifact/controller.go
ywk253100: When pushing a same name but different digest image, we need to update the artifact_id field of the tag to point to the new image

src/pkg/artifact/model/model.go
ywk253100: Don't get your point

src/pkg/artifact/controller.go
ywk253100: I think so

reasonerjt:
Do you mean interface?

src/pkg/artifact/controller.go
```diff
@@ -13,3 +13,103 @@
 // limitations under the License.
 
 package artifact
```
steven-zou: As we discussed, the controller will be in the `api/artifact` folder.

I'll complete the programming model tomorrow.

src/pkg/artifact/controller.go
```diff
@@ -13,3 +13,103 @@
...
+var (
+	// Ctl is a global Controller instance
+	Ctl = NewController()
+)
+
+// Controller is the only interface of artifact module to provide the
+// management functions for artifacts
+type Controller interface {
+	// List artifacts according to the query
+	List(context.Context, ...*model.Query) (total int64, artifacts []*model.Artifact, err error)
```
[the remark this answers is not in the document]
ywk253100: How about the specific query only works for artifacts?

src/pkg/artifact/dao/model.go
```diff
@@ -67,9 +65,10 @@ func (t *Tag) TableName() string {
 
 // ArtifactReference records the child artifact referenced by parent artifact
 type ArtifactReference struct {
-	ID       int64 `orm:"pk;auto;column(id)"`
-	ParentID int64 `orm:"column(parent_id)"`
-	ChildID  int64 `orm:"column(child_id)"`
+	ID       int64  `orm:"pk;auto;column(id)"`
+	ParentID int64  `orm:"column(parent_id)"`
+	ChildID  int64  `orm:"column(child_id)"`
+	Platform string `orm:"column(platform)"`
```
steven-zou: A string can cover the platform related info? I remembered there were other information related to the `platform`

src/pkg/artifact/manager.go
```diff
@@ -0,0 +1,112 @@
...
+var (
+	// Mgr is a global artifact manager instance
+	Mgr = NewManager()
+)
+
+// Manager is the only interface of artifact module to provide the
+// management functions for artifacts
+type Manager interface {
+	// List artifacts according to the query
+	List(ctx context.Context, query ...*q.Query) (total int64, artifacts []*Artifact, err error)
```
steven-zou: IMO, suggest using `query  *q.Query` to replace `query ...*q.Query` and append more comments to indicate that query can be nil (that means no query parameters)

src/pkg/artifact/manager.go
```diff
@@ -0,0 +1,112 @@
...
+	Delete(ctx context.Context, id int64) error
+	// UpdatePullTime updates the pull time of artifact. If the tag is provided, update the pull time of tag as well
+	UpdatePullTime(ctx context.Context, artifact *Artifact, tag *Tag, time time.Time) error
+	// GetSubResource returns the sub resource of the artifact
+	// The sub resource is different according to the artifact type:
+	// build history for image; values.yaml, readme and dependencies for chart, etc
+	GetSubResource(ctx context.Context, artifactID int64, resource string) ([]byte, error)
+	// Attach creates a new tag and attaches it to the artifact. If the tag is already attached
+	// to other artifact, an error will be returned
+	Attach(ctx context.Context, artifact *Artifact, tag string) error
```
steven-zou: As this is an artifact manager, I suggest using `AttachTag` and `DetachTag` for the method names.

src/pkg/artifact/manager.go
steven-zou: Any way to query out all the useless artifacts (no tags attached)?

src/pkg/artifact/manager.go
```diff
@@ -0,0 +1,112 @@
...
+	// Create the artifact specified by the digest under the repository,
+	// returns the artifact ID and error. If tags are provided, attach them
+	// to the artifact. If the tag is already attached to other artifact,
+	// will detach it first
+	Create(ctx context.Context, repository *models.RepoRecord, digest string, tags ...string) (id int64, err error)
+	// Delete just deletes the artifact record. The underlying data of registry will be
+	// removed during garbage collection
+	Delete(ctx context.Context, id int64) error
+	// UpdatePullTime updates the pull time of artifact. If the tag is provided, update the pull time of tag as well
+	UpdatePullTime(ctx context.Context, artifact *Artifact, tag *Tag, time time.Time) error
```
steven-zou: It seems not consistent here, why use `artifact *Artifact`?  An ID is not enough?

src/pkg/artifact/manager.go
```diff
@@ -0,0 +1,112 @@
...
+	// GetSubResource returns the sub resource of the artifact
+	// The sub resource is different according to the artifact type:
+	// build history for image; values.yaml, readme and dependencies for chart, etc
+	GetSubResource(ctx context.Context, artifactID int64, resource string) ([]byte, error)
```
steven-zou: How about

```go
GetSubResources(ctx context.Context, artifactID int64, resources ...string) (map[string][]byte, error)
```

In case if I want to get them all or more than one?
?

src/pkg/artifact/dao/model.go
ywk253100: It's the database model, the platform is a json string

src/pkg/artifact/manager.go
ywk253100: We can support that by adding a query keyword such as `without-tag=true` if we have the requirement

src/pkg/artifact/manager.go
ywk253100: OK

src/pkg/artifact/manager.go
ywk253100: Will update to ArtifactID and TagID

src/pkg/artifact/manager.go
steven-zou: ok, keep `GetSubResource ` as it is

src/api/artifact/controller.go
```diff
@@ -0,0 +1,63 @@
...
+	Delete(ctx context.Context, id int64) (err error)
+	// DeleteTag deletes the tag specified by ID
+	DeleteTag(ctx context.Context, id int64) (err error)
+	// UpdatePullTime updates the pull time for the artifact. If the tag is provides, update the pull
+	// time of the tag as well
+	UpdatePullTime(ctx context.Context, artifactID int64, tag string, time time.Time) (err error)
+	// GetSubResource returns the sub resource of the artifact
+	// The sub resource is different according to the artifact type:
+	// build history for image; values.yaml, readme and dependencies for chart, etc
+	GetSubResource(ctx context.Context, artifactID int64, resource string) ([]byte, error)
```
reasonerjt: Does it mean the API should keep the returning value untouched and let the http handler return the byte array in the response?  If so, how to determine the content-type?

src/api/artifact/model.go
```diff
@@ -0,0 +1,48 @@
...
+
+package artifact
+
+import (
+	"github.com/goharbor/harbor/src/pkg/artifact"
+	"github.com/goharbor/harbor/src/pkg/tag/model/tag"
+)
+
+// Artifact is the overall view of artifact
+type Artifact struct {
```
reasonerjt: we should reuse the swagger generated models in this layer.  Please make sure definition is removed once we have a concrete workflow to gen the code.

src/api/artifact/controller.go
ywk253100: Introduced a new `Resource` type

src/api/artifact/model.go
ywk253100: `TODO` added

PR #10250, opened 2019-12-13
[doc] Update doc for migration

ninjadq:
Migration guild is a lit bit confusion

Add copy after migrate config file


ninjadq:
already migrated to 110

Issue #10012, opened 2019-11-26
Admin Permission not revoked when removed from LDAP Admin Group


sluetze:
If you are reporting a problem, please make sure the following information are provided:

**Expected behavior and actual behavior:**
When using LDAP Integration with Harbor I want to delegate permissions in LDAP. 
When I remove a ldap-user from the ldap-group for admin permissions I expect the permissions in harbor to be revoked.
Currently when I remove a ldap-user from the ldap-group the permissions persist, since the user gets the "is_admin" flag in harbor on initial log in

This is security relevant, since a permission revocation has to be done at multiple points (LDAP & *every* Harbor Instance)

**Steps to reproduce the problem:**
1. Configure Harbor for LDAP
1. Create a LDAP-User
1. Put LDAP-User in LDAP-Group for Admins
1. Log in with LDAP-User
1. Recognize "Admin Flag" and the possibility to be an Admin ("Configuration")
1. Log out with LDAP-User
1. Remove LDAP-User from LDAP-Group for Admins
1. Wait for Replication (when you have multi DC)
1. Log in with LDAP-User
1. Recognize "Admin Flag" and the possibility to be an Admin ("Configuration")

**Versions:**
- harbor version: 1.9.3

**Additional context:**
- none

**Fix**
don't set "is admin" flag for ldap users, since the permissions for ldap users are generated at runtime.

**workaround**
Regularly check users for admin flag and revoke admin flag

may be related to  #8052

jacekn:
I'm also affected by this.
Another path to see the issue is:
* log in as user who is member of the group set as `LDAP Group Admin DN`
* user gets admin. Log out
* change `LDAP Group Admin DN` to another group
* log in as original user, user still has admin

A workaround I found is to "Revoke admin" from all users in the users tab. After that admin LDAP group membership works correctly (members get admin, non members don't)

stonezdj:
There are two ways to make an user admin 
1. User is assigned admin privilege.
2. User is in the ldap group admin DN group 
The problem is when the user is in that LDAP group, isadmin is persistent in the database

xaleeks:
these permissions are now updated at runtime and not only during user creation.  this needs a mention in the release notes @stuclem

stonezdj:
fixed

PR #10261, opened 2019-12-13
Fix admin permission not revoked when removed from LDAP admin group(Do Not Merge)

stonezdj:
Separate the HasAdminRole(In DB) with the privileges from external auth, and use user.HasAdminPrivilege to check if the current user has admin privilege.
Fix issue: #10012

### DIFF
--- a/src/common/dao/dao_test.go
+++ b/src/common/dao/dao_test.go
@@ -626,12 +626,41 @@ func TestGetRoleByID(t *testing.T) {
 	}
 }
 
+// isAdminRole returns whether the user is admin.
+func isAdminRole(userIDOrUsername interface{}) (bool, error) {
+	u := models.User{}
+
+	switch v := userIDOrUsername.(type) {
+	case int:
+		u.UserID = v
+	case string:
+		u.Username = v
+	default:
+		return false, fmt.Errorf("invalid parameter, only int and string are supported: %v", userIDOrUsername)
+	}
+
+	if u.UserID == NonExistUserID && len(u.Username) == 0 {
+		return false, nil
+	}
+
+	user, err := GetUser(u)
+	if err != nil {
+		return false, err
+	}
+
+	if user == nil {
+		return false, nil
+	}
+
+	return user.SysAdminFlag, nil
+}
+
 func TestToggleAdminRole(t *testing.T) {
 	err := ToggleUserAdminRole(currentUser.UserID, true)
 	if err != nil {
 		t.Errorf("Error in toggle ToggleUserAdmin role: %v, user: %+v", err, currentUser)
 	}
-	isAdmin, err := IsAdminRole(currentUser.UserID)
+	isAdmin, err := isAdminRole(currentUser.UserID)
 	if err != nil {
 		t.Errorf("Error in IsAdminRole: %v, user id: %d", err, currentUser.UserID)
 	}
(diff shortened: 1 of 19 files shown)


src/common/dao/role.go
```diff
@@ -70,7 +70,7 @@ func IsAdminRole(userIDOrUsername interface{}) (bool, error) {
 		return false, nil
 	}
 
-	return user.HasAdminRole, nil
+	return user.HasAdminPrivilege(), nil
```
reasonerjt: dao pkg would better only reflect the data in DB, can we use securityCtx ?

src/common/models/user.go
```diff
@@ -34,15 +34,23 @@ type User struct {
...
-	//	RoleList     []Role `json:"role_list"`
-	HasAdminRole bool      `orm:"column(sysadmin_flag)" json:"has_admin_role"`
-	ResetUUID    string    `orm:"column(reset_uuid)" json:"reset_uuid"`
-	Salt         string    `orm:"column(salt)" json:"-"`
-	CreationTime time.Time `orm:"column(creation_time);auto_now_add" json:"creation_time"`
-	UpdateTime   time.Time `orm:"column(update_time);auto_now" json:"update_time"`
-	GroupIDs     []int     `orm:"-" json:"-"`
-	OIDCUserMeta *OIDCUser `orm:"-" json:"oidc_user_meta,omitempty"`
+	Role         int  `orm:"-" json:"role_id"`
+	HasAdminRole bool `orm:"column(sysadmin_flag)" json:"has_admin_role"`
```
reasonerjt: Can we check how "has_admin_role" in the JSON response is used, I think may be it's better we create another attribute, "sysadmin_flag" to 1:1 map to the DB column and use "HasAdminRole" to reflect to consolidated privilege?

src/core/api/user.go
```diff
@@ -151,7 +151,9 @@ func (ua *UserAPI) Get() {
 		}
 		u.Password = ""
 		if ua.userID == ua.currentUserID {
-			u.HasAdminRole = ua.SecurityCtx.IsSysAdmin()
+			// Because UI is rendered with the condition of HasAdminRole,
+			// copy admin privilege to this field to REST response for UI to render it correctly
+			u.SysAdminFlag = ua.SecurityCtx.IsSysAdmin()
```
reasonerjt: This is wrong, the systemAdminFlag should only reflects the column in DB, but `IsSysAdmin` is the merged result.

src/common/dao/dao_test.go
```diff
@@ -626,6 +626,35 @@ func TestGetRoleByID(t *testing.T) {
 	}
 }
 
+// IsAdminRole returns whether the user is admin.
```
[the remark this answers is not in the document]
stonezdj: this function retrieves the new state from DB

src/common/security/local/context.go
```diff
@@ -52,13 +52,18 @@ func (s *SecurityContext) GetUsername() string {
 	return s.user.Username
 }
 
+// User get the current user
+func (s *SecurityContext) User() *models.User {
```
ywk253100: Is this function used in anywhere?

ywk253100: Why we need it?

xaleeks:
what version is this fixed for? can we add a version label on PRs? @stonezdj

PR #10267, opened 2019-12-16
Improve search function for replication and tags

AllForNothing:
1.Improve search function : The old search function did not work perfectly and would cause nightly  case failure sometime.
2. Improve css for cron-component , filter-component and image-name-input.component.

Env: https://10.161.79.96:8092

### DIFF
--- a/src/portal/src/lib/components/cron-schedule/cron-schedule.component.scss
+++ b/src/portal/src/lib/components/cron-schedule/cron-schedule.component.scss
@@ -8,10 +8,6 @@
     .normal-wrapper {
       position: absolute;
       width: 700px;
-      > span:first-child {
-        width: 200px;
-      }
-    
       > span:not(:first-child) {
         margin-right: 18px;
       }
@@ -20,18 +16,14 @@
       }
     }
     button {
-      margin: 35px 20px 0 200px;
+      margin: 35px 20px 0;
     }
   }
   
   .setting-wrapper {
     position: relative;
-    > span:first-child {
-      width: 200px;
-    }
-
     .confirm-button {
-      margin: 20px 0px 0 200px;
+      margin: 20px 0 0;
     }
   
     *:not(:first-child) {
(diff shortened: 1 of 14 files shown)


src/portal/src/lib/components/cron-schedule/cron-schedule.component.scss
```diff
@@ -20,18 +16,14 @@
...
   
   .setting-wrapper {
     position: relative;
-    > span:first-child {
-      width: 200px;
-    }
-
     .confirm-button {
-      margin: 20px 0px 0 200px;
+      margin: 20px 0 0;
```
zhoumeina: Why just specify 3 parameter instead of 4?

src/portal/src/lib/components/image-name-input/image-name-input.component.ts
```diff
@@ -71,7 +93,7 @@ export class ImageNameInputComponent implements OnInit, OnDestroy {
                 } else {
                     this.errorHandler.error("not Observable type");
                 }
-            });
+            });*/
```
zhoumeina: Why comment out the old code? Not remove them?

src/portal/src/lib/components/cron-schedule/cron-schedule.component.scss
AllForNothing: here is the fourth parameter :  @Input() buttonMarginLeft: string = '200px';

src/portal/src/lib/components/image-name-input/image-name-input.component.ts
AllForNothing: fixed

src/portal/src/lib/components/cron-schedule/cron-schedule.component.scss
AllForNothing: default value  is 200px  ,  and it can be customized

PR #10280, opened 2019-12-17
enhance[cicd] introduce github action for CICD

bitsf:
In order to replace travis.
Implement 5 CI jobs
- UTTEST
- APITEST_DB
- APITEST_LDAP
- OFFLINE
- UI_UT

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,0 +1,277 @@
+name: CI
+env:
+   POSTGRESQL_HOST: localhost
+   POSTGRESQL_PORT: 5432
+   POSTGRESQL_USR: postgres
+   POSTGRESQL_PWD: root123
+   POSTGRESQL_DATABASE: registry
+   ADMINSERVER_URL: http://127.0.0.1:8888
+   DOCKER_COMPOSE_VERSION: 1.23.0
+   HARBOR_ADMIN: admin
+   HARBOR_ADMIN_PASSWD: Harbor12345
+   CORE_SECRET: tempString
+   KEY_PATH: "/data/secret/keys/secretkey"
+   REDIS_HOST: localhost
+   REG_VERSION: v2.7.1-patch-2819-2553
+   UI_BUILDER_VERSION: 1.6.0
+
+on: 
+  pull_request:
+  push:
+
+jobs:
+  UTTEST:
+    env:
+       UTTEST: true
+    runs-on: 
+      #- self-hosted
+      - ubuntu-latest
+    steps:
+      - name: Set up Go 1.13
+        uses: actions/setup-go@v1
+        with:
+           go-version: 1.13
+        id: go
+      - name: setup Docker
+        uses: docker-practice/actions-setup-docker@0.0.1
+        with:
+          docker_version: 18.09
+          docker_channel: stable
+      - uses: actions/checkout@v1
+        with:
+         fetch-depth: 1
+         path: src/github.com/goharbor/harbor
+      - name: setup env
+        run: |
+          pwd
+          go env
+          echo "::set-env name=GOPATH::$(go env GOPATH):$RUNNER_WORKSPACE"
+          echo "::add-path::$(go env GOPATH)/bin"
+          echo "::set-env name=TOKEN_PRIVATE_KEY_PATH::${GITHUB_WORKSPACE}/tests/private_key.pem"
+        shell: bash
+      - name: before_install
+        run: |
+          set -x
+          pwd
+          env
+          #sudo apt install -y xvfb
+          #xvfb-run ls
+          curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > docker-compose
+          chmod +x docker-compose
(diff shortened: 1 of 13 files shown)


wy65701436:
do we need to disable travis as well?

.github/workflows/CI.yml
```diff
@@ -0,0 +1,277 @@
...
+          IP=`hostname -I | awk '{print $1}'`
+          #echo '{"insecure-registries" : ["'$IP':5000"]}' | sudo tee /etc/docker/daemon.json
+          echo "::set-env name=IP::$IP"
+          sudo cp ./tests/harbor_ca.crt /usr/local/share/ca-certificates/
+          sudo update-ca-certificates
+          #sudo service docker restart
+      - name: script
+        run: |
+          echo IP: $IP
+          bash ./tests/showtime.sh ./tests/travis/ui_ut_run.sh
```
heww: Please use job matrix to simpler the definitions of jobs.

bitsf: I separate each job definition in order to optimize each one later. Matrix didn't introduce much benefits here.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,277 @@
...
+          echo "::set-env name=GOPATH::$(go env GOPATH):$RUNNER_WORKSPACE"
+          echo "::add-path::$(go env GOPATH)/bin"
+          echo "::set-env name=TOKEN_PRIVATE_KEY_PATH::${GITHUB_WORKSPACE}/tests/private_key.pem"
+        shell: bash
+      - name: before_install
+        run: |
+          set -x
+          pwd
+          env
+          #sudo apt install -y xvfb
```
reasonerjt: Is it commented?

tests/travis/ut_install.sh
```diff
@@ -14,9 +14,10 @@ go get github.com/stretchr/testify
 go get golang.org/x/tools/cmd/cover
 go get github.com/mattn/goveralls
 go get -u github.com/client9/misspell/cmd/misspell
-sudo service postgresql stop
+sudo service postgresql stop || echo no postgresql need to be stopped
```
reasonerjt: This would eat the error? not only that there's not pgsql?

src/core/api/utils.go
```diff
@@ -64,7 +64,7 @@ func SyncRegistry(pm promgr.ProjectManager) error {
 	}
 
 	if len(reposToAdd) > 0 {
-		log.Debugf("Start adding repositories into DB... ")
+		log.Infof("Start adding repositories into DB %v ... ", len(reposToAdd))
```
reasonerjt: How is this change necessary for switching CI?

tests/coverage4gotest.sh
```diff
@@ -21,13 +21,15 @@ function listDeps(){
...
 packages=$(go list ./... | grep -v -E 'vendor|tests|testing')
+echo testing packages: $packages
 
 for package in $packages
 do
 	listDeps $package
 
 #    echo "DEBUG: testing package $package"
-	go test -race -cover -coverprofile=profile.tmp -coverpkg "$deps" $package
+	echo go test -race -v -cover -coverprofile=profile.tmp -coverpkg "$deps" $package
```
reasonerjt: Do we still need this `echo`?
I remember some of the output was commented to reduce the size of the output

tests/docker-compose.test.yml
```diff
@@ -24,6 +24,6 @@ services:
     image: goharbor/redis-photon:4.0
     restart: always
     volumes:
-      - /data/redis:/data
+      - /data/redis:/var/lib/redis
```
reasonerjt: So redis in `docker-compose.test.yml` was not working previously?

reasonerjt:
> do we need to disable travis as well?

Let's not, let's keep the two running for a while and verify action is way better than travis.

.github/workflows/CI.yml
bitsf: yes, this is actually not used in my test.
and the installation here for ubuntu will also have problem, so I comment it out

tests/travis/ut_install.sh
bitsf: this is what we need, because postgresql is supposed to be not enable, so actually I didn't install it

src/core/api/utils.go
bitsf: This message should be info level, or it will be hard to known how many repositories being imported, so I change it  by the way.
Of course this is not for CI

tests/coverage4gotest.sh
bitsf: I think this is just the benefits for testing log, that we can easily check the command running, or it's hard for us to debug locally when case failed

tests/docker-compose.test.yml
bitsf: it's working previously, just not use a sharable folder
and actually this redis is using wrong tag, it should not be 4.0

bitsf: actually we should not use any data volume in such test yml, it will make the test env dirty and can't be reusable. We can fix this later

bitsf:
image

PR #10324, opened 2019-12-20
add OCI error format support

wy65701436:
1, Leverage go v1.13 new error feature
2, Define genernal error OCI format, so that /v2 API could return a OCI compatible error

### DIFF
--- a/src/common/api/base.go
+++ b/src/common/api/base.go
@@ -20,11 +20,12 @@ import (
 	"net/http"
 	"strconv"
 
+	"errors"
 	"github.com/astaxie/beego"
 	"github.com/astaxie/beego/validation"
 	commonhttp "github.com/goharbor/harbor/src/common/http"
 	"github.com/goharbor/harbor/src/common/utils/log"
-	"github.com/pkg/errors"
+	internal_errors "github.com/goharbor/harbor/src/internal/error"
 )
 
 const (
@@ -247,4 +248,49 @@ func (b *BaseAPI) SendPreconditionFailedError(err error) {
 // SendStatusServiceUnavailableError sends service unavailable error to the client.
 func (b *BaseAPI) SendStatusServiceUnavailableError(err error) {
 	b.RenderFormattedError(http.StatusServiceUnavailable, err.Error())
+}
+
+// SendError return the error defined in OCI spec: https://github.com/opencontainers/distribution-spec/blob/master/spec.md#errors
+// {
+//	"errors:" [{
+//			"code": <error identifier>,
+//			"message": <message describing condition>,
+//			// optional
+//			"detail": <unstructured>
+//		},
+//		...
+//	]
+// }
+func (b *BaseAPI) SendError(err error) {
+	var statusCode int
+	var send error
+
+	var e *internal_errors.Error
+	if errors.As(err, &e) {
+		code := e.Code
+		statusCode = b.getStatusCode(code)
+		if statusCode == 0 {
+			statusCode = http.StatusInternalServerError
+		}
+		send = e
+
+	} else {
+		statusCode = http.StatusInternalServerError
+		send = internal_errors.UnknownError(err)
+	}
+
+	b.RenderError(statusCode, internal_errors.NewErrs(send).Error())
+}
+
+func (b *BaseAPI) getStatusCode(code string) int {
+	statusCodeMap := map[string]int{
+		internal_errors.NotFoundCode:     http.StatusNotFound,
+		internal_errors.ConflictCode:     http.StatusConflict,
+		internal_errors.UnAuthorizedCode: http.StatusUnauthorized,
+		internal_errors.BadRequestCode:   http.StatusBadRequest,
+		internal_errors.ForbiddenCode:    http.StatusForbidden,
+		internal_errors.PreconditionCode: http.StatusPreconditionFailed,
+		internal_errors.GeneralCode:      http.StatusInternalServerError,
+	}
+	return statusCodeMap[code]
 }
--- a/src/internal/error/errors.go
+++ b/src/internal/error/errors.go
@@ -1,0 +1,147 @@
+package error
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/goharbor/harbor/src/common/utils/log"
+	"strings"
+)
+
+// Error ...
+type Error struct {
+	Cause   error  `json:"-"`
+	Code    string `json:"code"`
+	Message string `json:"message"`
+}
+
+// Error returns a human readable error.
+func (e *Error) Error() string {
+	return fmt.Sprintf("%v, %s, %s", e.Cause, e.Code, e.Message)
+}
+
+// WithMessage ...
+func (e *Error) WithMessage(msg string) *Error {
+	e.Message = msg
+	return e
+}
+
+// WithCode ...
+func (e *Error) WithCode(code string) *Error {
+	e.Code = code
+	return e
+}
+
+// Unwrap ...
+func (e *Error) Unwrap() error { return e.Cause }
+
+// Errors ...
+type Errors []error
+
+var _ error = Errors{}
+
+// Error converts slice of error
+func (errs Errors) Error() string {
+	var tmpErrs struct {
+		Errors []Error `json:"errors,omitempty"`
+	}
+
+	for _, e := range errs {
+		var err error
+		switch e.(type) {
+		case *Error:
+			err = e.(*Error)
+		default:
+			err = UnknownError(e).WithMessage(err.Error())
+		}
+		tmpErrs.Errors = append(tmpErrs.Errors, *err.(*Error))
+	}
+
+	msg, err := json.Marshal(tmpErrs)
+	if err != nil {
+		log.Error(err)
+		return "{}"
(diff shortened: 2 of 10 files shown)


src/pkg/error/errors.go
```diff
@@ -0,0 +1,184 @@
...
+	return "{}"
+}
+
+// Len returns the current number of errors.
+func (errs Errors) Len() int {
+	return len(errs)
+}
+
+// Es ...
+func Es(err error) Errors {
```
heww: `NewErrors` or `FormatError` will be better than `Es`.

src/pkg/error/errors.go
```diff
@@ -0,0 +1,184 @@
...
+
+	// ErrorPrecondition ...
+	ErrorPrecondition = errors.New("precondition failed")
+
+	// ErrUnknown ...
+	ErrUnknown = errors.New("unknown")
+)
+
+// Code ...
+type Code string
```
heww: `Int` is better than `string` for `Code`

src/pkg/error/errors.go
```diff
@@ -0,0 +1,184 @@
...
+	ForbiddenErrorCode
+	// PreconditionErrorCode ...
+	PreconditionErrorCode
+	// GeneralErrorCode ...
+	GeneralErrorCode
+)
+
+// NotFoundError is error for the case of object not found
+func NotFoundError(err error) error {
+	return E(ErrNotFound, Code(strconv.Itoa(ObjectNotFoundErrorCode)), M(err.Error()), D("not found"))
```
heww: When `Code` is int type,  type conversion not needed here.

src/pkg/error/errors.go
```diff
@@ -0,0 +1,184 @@
...
+	ErrorPrecondition = errors.New("precondition failed")
+
+	// ErrUnknown ...
+	ErrUnknown = errors.New("unknown")
+)
+
+// Code ...
+type Code string
+
+// M ...
```
ywk253100: Please add comment to explain what it is, the same for other definitions

src/pkg/error/errors.go
```diff
@@ -0,0 +1,184 @@
...
+		Message: msg,
+		Detail:  e.Detail,
+	}
+}
+
+// Unwrap ...
+func (e Error) Unwrap() error { return e.Err }
+
+// E ...
+func E(args ...interface{}) error {
```
ywk253100: Why we need this? What is it used for?

src/pkg/error/errors.go
```diff
@@ -0,0 +1,184 @@
...
+
+import (
+	"encoding/json"
+	"errors"
+	"strconv"
+)
+
+var (
+	// ErrNotFound ...
+	ErrNotFound = errors.New("not found")
```
ywk253100: Why we need these error definitions? I think the error code is enough to identify the error

src/pkg/error/errors.go
```diff
@@ -0,0 +1,184 @@
...
+type M string
+
+// D ...
+type D string
+
+// Error ...
+type Error struct {
+	Err     error `json:"-"`
+	Code    Code  `json:"code"`
+	Message M     `json:"message"`
```
ywk253100: Why don't we use `string` as the type of `Message` and `Detail`?

src/pkg/error/errors.go
```diff
@@ -0,0 +1,173 @@
...
+	return &Error{
+		Err:     err,
+		Code:    ObjectConflictErrorCode,
+		Message: err.Error(),
+		Detail:  "conflict",
+	}
+}
+
+// UnauthorizedError is error for the case of unauthorized accessing
+func UnauthorizedError(err error) error {
```
reasonerjt: if the input parm is an instance of `Error` struct, would it be better we `unwrap` the error first?

src/common/api/base.go
```diff
@@ -248,3 +249,47 @@ func (b *BaseAPI) SendPreconditionFailedError(err error) {
...
+//		},
+//		...
+//	]
+// }
+func (b *BaseAPI) SendError(err error) {
+	var statusCode int
+	var send error
+
+	var e pkg_errors.Error
+	if errors.Is(err, e) {
```
reasonerjt: looks like we should use `errors.As` here.

heww: And I have doubts that in controller  how can I check the `error` is a `not found` error returned by dao.

Use `errors.As` and check the code of the error? This is very tricky.

src/pkg/error/errors.go
```diff
@@ -0,0 +1,173 @@
...
+	GeneralErrorCode
+)
+
+// NotFoundError is error for the case of object not found
+func NotFoundError(err error) error {
+	return Error{
+		Err:     err,
+		Code:    ObjectNotFoundErrorCode,
+		Message: err.Error(),
+		Detail:  "not found",
```
ywk253100: The `Message` is a summary message for the error and the `Detail` contains the detail information

ywk253100: The `Message` should be const string for the specific code.

ywk253100: And I should be able to customize the detail information for the specific pre-defined error

src/pkg/error/errors.go
```diff
@@ -0,0 +1,173 @@
...
+// Error returns a human readable error.
+func (e Error) Error() string {
+	if data, err := json.Marshal(e); err == nil {
+		return string(data)
+	}
+	return "{}"
+}
+
+// WithMessage ...
+func (e Error) WithMessage(msg string) Error {
```
ywk253100: Seems this function is useless

src/pkg/error/errors.go
```diff
@@ -0,0 +1,173 @@
...
+	Err     error  `json:"-"`
+	Code    Code   `json:"code"`
+	Message string `json:"message"`
+	Detail  string `json:"detail,omitempty"`
+}
+
+// Error returns a human readable error.
+func (e Error) Error() string {
+	if data, err := json.Marshal(e); err == nil {
+		return string(data)
```
ywk253100: Log the error if marshal failed

src/pkg/error/errors.go
```diff
@@ -0,0 +1,173 @@
...
+
+	for _, daErr := range errs {
+		var err error
+		switch daErr.(type) {
+		case Error:
+			err = daErr.(Error)
+		default:
+			err = UnknownError(daErr).(Error).WithMessage(err.Error())
+		}
+		tmpErrs.Errors = append(tmpErrs.Errors, Error{
```
ywk253100: Just append the `err` is fine

src/pkg/error/errors.go
wy65701436: docker client redenders the message instead of detail, that's why give the error to message.

wy65701436: why message should contain code, use can get it from the code field. And the `withMessage` fun c can let user to customize messages, is it what you want?

src/common/api/base.go
```diff
@@ -248,3 +249,48 @@ func (b *BaseAPI) SendPreconditionFailedError(err error) {
...
+
+	} else {
+		statusCode = http.StatusInternalServerError
+		send = internal_errors.UnknownError(err)
+	}
+
+	b.RenderError(statusCode, internal_errors.NewErrs(send).Error())
+}
+
+func (b *BaseAPI) getStatusCode(code string) int {
```
renmaosheng: capital Get?

src/internal/error/errors.go
```diff
@@ -0,0 +1,147 @@
...
+	// NotFoundCode is code for the error of no object found
+	NotFoundCode = "NOT_FOUND"
+	// ConflictCode ...
+	ConflictCode = "CONFLICT"
+	// UnAuthorizedCode ...
+	UnAuthorizedCode = "UNAUTHORIZED"
+	// BadRequestCode ...
+	BadRequestCode = "BAD_REQUEST"
+	// ForbiddenCode ...
+	ForbiddenCode = "FORBIDDER"
```
renmaosheng: FORBIDDEN?

src/internal/error/errors.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// ForbiddenError is error for the case of forbidden
+func ForbiddenError(err error) *Error {
+	return New(err).WithCode(ForbiddenCode).WithMessage("forbidden")
+}
+
+// PreconditionFailedError is error for the case of precondition failed
+func PreconditionFailedError(err error) *Error {
+	return New(err).WithCode(PreconditionCode).WithMessage("preconfition")
```
renmaosheng: precondition?

src/common/api/base.go
ywk253100: It's a internal method, so the capital Get isn't needed

src/internal/error/errors.go
ywk253100: Fix included in https://github.com/goharbor/harbor/pull/10366

src/internal/error/errors.go
ywk253100: Fix included in https://github.com/goharbor/harbor/pull/10366

PR #10329, opened 2019-12-23
add expiration data time when to create a robot account

wy65701436:
Update API of creating robot accout, user can specify expiration time per account.

### DIFF
--- a/api/harbor/swagger.yaml
+++ b/api/harbor/swagger.yaml
@@ -6239,6 +6239,9 @@ definitions:
       description:
         type: string
         description: The description of robot account
+      expires_at:
+        type: integer
+        description: The expiration time on or after which the JWT MUST NOT be accepted for processing.
       access:
         type: array
         description: The permission of robot account
--- a/src/pkg/robot/controller.go
+++ b/src/pkg/robot/controller.go
@@ -58,17 +58,19 @@ func (d *DefaultAPIController) GetRobotAccount(id int64) (*model.Robot, error) {
 func (d *DefaultAPIController) CreateRobotAccount(robotReq *model.RobotCreate) (*model.Robot, error) {
 
 	var deferDel error
-	// Token duration in minutes
-	tokenDuration := time.Duration(config.RobotTokenDuration()) * time.Minute
-	expiresAt := time.Now().UTC().Add(tokenDuration).Unix()
 	createdName := common.RobotPrefix + robotReq.Name
+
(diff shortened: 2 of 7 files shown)


src/pkg/robot/controller.go
```diff
@@ -58,17 +58,22 @@ func (d *DefaultAPIController) GetRobotAccount(id int64) (*model.Robot, error) {
...
 	var deferDel error
-	// Token duration in minutes
+
 	tokenDuration := time.Duration(config.RobotTokenDuration()) * time.Minute
 	expiresAt := time.Now().UTC().Add(tokenDuration).Unix()
+
 	createdName := common.RobotPrefix + robotReq.Name
 
+	if robotReq.ExpiresAt == 0 {
+		robotReq.ExpiresAt = expiresAt
```
heww: Suggest to move these two lines in this if statement.

```go	
tokenDuration := time.Duration(config.RobotTokenDuration()) * time.Minute
expiresAt := time.Now().UTC().Add(tokenDuration).Unix()
```

api/harbor/swagger.yaml
```diff
@@ -6236,6 +6236,9 @@ definitions:
       description:
         type: string
         description: The description of robot account
+      expires_at:
+        type: integer
+        description: The expiration of robot account (in seconds)
```
reasonerjt: This description is misleading.

I think you mean the number of seconds since 1970-01-01?  I recall it's called epoch time?  Please double check.

wy65701436: updated

src/pkg/robot/controller.go
wy65701436: ok

src/pkg/robot/model/robot.go
```diff
@@ -67,6 +68,9 @@ func (rq *RobotCreate) Valid(v *validation.Validation) {
 	if utils.IsContainIllegalChar(rq.Name, []string{",", "~", "#", "$", "%"}) {
 		v.SetError("name", "robot name contains illegal characters")
 	}
+	if rq.ExpiresAt < 0 {
+		v.SetError("expires_at", "robot account must set a expiration time.")
```
reasonerjt: Incorrect error message, and the API caller can leave this attribute unset.

wy65701436: updated.

PR #10337, opened 2019-12-25
feat(orm): add orm support with context

heww:
1. Get and set orm from context.
2. Add WithTransaction decorator make func run in transaction.
3. Support nested transaction by Savepoint.

### DIFF
--- a/src/core/filter/orm.go
+++ b/src/core/filter/orm.go
@@ -1,0 +1,30 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package filter
+
+import (
+	"github.com/astaxie/beego/context"
+	o "github.com/astaxie/beego/orm"
+	"github.com/goharbor/harbor/src/internal/orm"
+)
+
+// OrmFilter set orm.Ormer instance to the context of the http.Request
+func OrmFilter(ctx *context.Context) {
+	if ctx == nil || ctx.Request == nil {
+		return
+	}
+
+	ctx.Request = ctx.Request.WithContext(orm.NewContext(ctx.Request.Context(), o.NewOrm()))
+}
--- a/src/internal/orm/orm.go
+++ b/src/internal/orm/orm.go
@@ -1,0 +1,68 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package orm
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/astaxie/beego/orm"
+	"github.com/goharbor/harbor/src/common/utils/log"
+)
+
+type ormKey struct{}
+
+// FromContext returns orm from context
+func FromContext(ctx context.Context) (orm.Ormer, bool) {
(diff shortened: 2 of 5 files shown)


src/core/filter/orm.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/goharbor/harbor/src/internal/orm"
+)
+
+// OrmFilter set orm.Ormer instance to the context of the http.Request
+func OrmFilter(ctx *context.Context) {
+	if ctx == nil {
+		return
+	}
+
+	req := ctx.Request
```
steven-zou: simplify this code?

```go
//  Validate context
if ctx ==nil || ctx.Request == nil {
  return
}
```

src/internal/orm/orm.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/astaxie/beego/orm"
+	"github.com/goharbor/harbor/src/common/utils/log"
+)
+
+type ormKey struct{}
```
steven-zou: This key seems too general, why not use a specified key?

src/internal/orm/orm.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+
+		tx := ormerTx{Ormer: o}
+		if err := tx.Begin(); err != nil {
+			log.Errorf("begin transaction failed: %v", err)
+			return err
+		}
+
+		if err := f(ctx); err != nil {
+			if e := tx.Rollback(); e != nil {
```
steven-zou: Suggest wrapping the error and then return it. Overriding error or discard error may lose some error context.

src/internal/orm/tx.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	return string(b)
+}
+
+var getTimestamp = func() int64 {
+	return time.Now().Unix()
+}
+
+func getSavepoint() string {
```
steven-zou: Seems the algorithm you did is just creating a random serial, why not directly use the UUID?

src/internal/orm/tx.go
```diff
@@ -0,0 +1,90 @@
...
+
+// ormerTx transaction which support savepoint
+type ormerTx struct {
+	orm.Ormer
+	savepoint string
+}
+
+func (o *ormerTx) Begin() error {
+	err := o.Ormer.Begin()
+	if err != nil && err == orm.ErrTxHasBegan {
```
steven-zou: if err != nil && err == orm.ErrTxHasBegan =>

`if err := o.Ormer.Begin(); err == orm.ErrTxHasBegan { ... }` is enough

src/internal/orm/orm.go
ywk253100: Leverage our common error

heww: Not everything needed to be common error, this is only the raw error from the sql driver, i think it's ok for now.

The API error handler will transform this error to 500 internal server error.

src/internal/orm/tx.go
heww: This is not enough, the func will return the err when it is not `ErrTxHasBegan`.

src/internal/orm/tx.go
heww: Yes, uuid is uniq and can be used as a savepoint name.

But I want the savepoint name orderly, maybe I should change the last part to auto increment number.

src/core/filter/orm.go
heww: But  I want to use short name of req in the line 34, otherwise the line will be

```go
ctx.Request = ctx.Request.WithContext(orm.NewContext(ctx.Request.Context(), o.NewOrm()))
```

src/internal/orm/orm.go
heww: This is internal type, the size of empty struct is zero and it is  addressable. So it's okay. 

And this method is better, see more https://medium.com/random-go-tips/on-context-context-keys-uniqueness-738aa7a5c87f

src/internal/orm/tx.go
heww: Changed to use uuid

src/internal/orm/orm.go
reasonerjt: IMO we only need to make sure the error is properly wrapped before the API returns, otherwise when the code stack is deep it will be wrapped more than necessarily.

src/internal/orm/tx.go
heww: Move operations about savepoint to methods which made this more more simpler.

src/internal/orm/orm.go
steven-zou: 👌

src/internal/orm/orm.go
steven-zou:
```go
func load(){
if err:=call(); err!=nil{
  return errors.Wrap(err, "my context for load") // then the error string will be `error string : my context for load`; easy for tracing error becasue error may happen in diffeent caller contexts
}
}

This is just a suggestion, not a required term
depends on you

```

PR #10338, opened 2019-12-25
Modify system admin role from has_admin_role to admin_role_in_auth and sysadmin_flag

jwangyangls:


src/portal/src/app/config/config.component.ts
```diff
@@ -59,7 +59,7 @@ export class ConfigurationComponent implements OnInit, OnDestroy {
 
     public get hasAdminRole(): boolean {
         return this.session.getCurrentUser() &&
-            this.session.getCurrentUser().has_admin_role;
```
zhoumeina: Please define a model and use it in the get current user service to avoid use a backend properties in so many places.

PR #10339, opened 2019-12-25
Add memory cache for cfgMap

stonezdj:
Add memory cache for properties
Default cache interval is 20 seconds
Disable cache by setting cfg_cache_interval_seconds=0


reasonerjt:
Is the cache in mem?  How do you handle the case when there are multiple `harbor-core`?

renmaosheng:
could you please elaborate why we need this? thanks.

stonezdj:
> could you please elaborate why we need this? thanks.
There are lots of query on the properties table.  it might cause performance issue when using external DB.

PR #10350, opened 2019-12-26
Onboard user when doing token review

reasonerjt:
This commit will make the "tokenreview" security filter onboard
user if the request carries a valid token.  If the "skipsearch" flag in
http_auth setting is set to false the onboard will fail.

### DIFF
--- a/src/core/filter/security.go
+++ b/src/core/filter/security.go
@@ -318,17 +318,26 @@ func (ap *authProxyReqCtxModifier) Modify(ctx *beegoctx.Context) bool {
 		log.Errorf("user name doesn't match with token: %s", rawUserName)
 		return false
 	}
-
 	user, err := dao.GetUser(models.User{
 		Username: rawUserName,
 	})
 	if err != nil {
-		log.Errorf("fail to get user: %v", err)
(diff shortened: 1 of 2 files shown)


src/core/filter/security.go
```diff
@@ -318,17 +318,26 @@ func (ap *authProxyReqCtxModifier) Modify(ctx *beegoctx.Context) bool {
...
-	if user == nil {
-		log.Errorf("User: %s has not been on boarded yet.", rawUserName)
-		return false
+	if user == nil { // onboard user if it's not yet onboarded.
+		uid, err := auth.SearchAndOnBoardUser(rawUserName)
+		if err != nil {
+			log.Errorf("Failed to search and onboard user, username: %s, error: %v", rawUserName, err)
+			return false
+		}
+		user, err = dao.GetUser(models.User{
```
stonezdj: No need to query  user again, the auth.SearchAndOnBoardUser is already return the uid, if it is non-zero,  then this user must be created.

reasonerjt: This is for getting the instance of the user for generating the security context.

PR #10367, opened 2019-12-27
Remove the manifest list middleware

ywk253100:
We'll support manifest list, so this commit removes the manifest list middleware who blocks the manifest list pushing


Adirio:
Despite being merged on 31st Dec 2019 and the last release (1.10) being from the 14th Feb 2020, this commit is not part of the 1.10 release. When is it planned to get this in a release? It is stopping us from upgrading from 1.7.5.

PR #10369, opened 2019-12-27
Populate basic auth information for registry

reasonerjt:
This commit updates `prepare` and templates to populate the credential
for registry for basic authentication.

A temporary flag `registry_use_basic_auth` was added to avoid breakage.
It MUST be removed before the release.

### DIFF
--- a/make/photon/prepare/main.py
+++ b/make/photon/prepare/main.py
@@ -35,7 +35,7 @@ def main(conf, with_notary, with_clair, with_chartmuseum):
     try:
         validate(config_dict, notary_mode=with_notary)
     except Exception as e:
-        logging.info('Error happend in config validation...')
+        logging.info('Error happened in config validation...')
         logging.error(e)
         sys.exit(-1)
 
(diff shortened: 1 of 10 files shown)


make/photon/prepare/main.py
```diff
@@ -35,10 +35,12 @@ def main(conf, with_notary, with_clair, with_chartmuseum):
     try:
         validate(config_dict, notary_mode=with_notary)
     except Exception as e:
-        logging.info('Error happend in config validation...')
+        logging.info('Error happened in config validation...')
         logging.error(e)
         sys.exit(-1)
 
+    gen_registry_credentials(config_dict)
```
ninjadq: Why not put this to `configs.py` file. Currently, all logics related to parsing  and changing the config_dict should be placed in `parse_yaml_config` function

reasonerjt: Sure, I've made change to move it to `parse_yaml_config` in a follow-up commit.

PR #10379, opened 2019-12-30
Implement the artifact abstractor and resolver

ywk253100:
1. Define the interface for artifact abstractor and resolver
2. Implement the artifact abstractor
3. Implement the resolver for image with manifest v2

### DIFF
--- a/src/api/artifact/abstractor/abstractor.go
+++ b/src/api/artifact/abstractor/abstractor.go
@@ -1,0 +1,118 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package abstractor
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"github.com/docker/distribution/manifest/manifestlist"
+	"github.com/docker/distribution/manifest/schema1"
+	"github.com/docker/distribution/manifest/schema2"
+	"github.com/goharbor/harbor/src/api/artifact/abstractor/blob"
+	"github.com/goharbor/harbor/src/api/artifact/abstractor/resolver"
+	"github.com/goharbor/harbor/src/pkg/artifact"
+	"github.com/goharbor/harbor/src/pkg/repository"
+	"github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+// Abstractor abstracts the specific information for different types of artifacts
+type Abstractor interface {
+	// Abstract the specific information for the specific artifact type into the artifact model,
+	// the information can be got from the manifest or other layers referenced by the manifest.
+	Abstract(ctx context.Context, artifact *artifact.Artifact) error
+}
+
+// NewAbstractor returns an instance of the default abstractor
+func NewAbstractor() Abstractor {
+	return &abstractor{
+		repoMgr:     repository.Mgr,
+		blobFetcher: blob.Fcher,
+	}
+}
+
+type abstractor struct {
+	repoMgr     repository.Manager
+	blobFetcher blob.Fetcher
+}
+
+// TODO try CNAB, how to forbid CNAB
+
+// TODO add white list for supported artifact type
+func (a *abstractor) Abstract(ctx context.Context, artifact *artifact.Artifact) error {
+	repository, err := a.repoMgr.Get(ctx, artifact.RepositoryID)
+	if err != nil {
+		return err
+	}
+	// read manifest content
+	manifestMediaType, content, err := a.blobFetcher.FetchManifest(repository.Name, artifact.Digest)
+	if err != nil {
+		return err
+	}
+	artifact.ManifestMediaType = manifestMediaType
+
+	switch artifact.ManifestMediaType {
+	// docker manifest v1
+	case "", "application/json", schema1.MediaTypeSignedManifest:
+		// TODO as the manifestmediatype isn't null, so add not null constraint in database
+		// unify the media type of v1 manifest to "schema1.MediaTypeSignedManifest"
+		artifact.ManifestMediaType = schema1.MediaTypeSignedManifest
+		// as no config layer in the docker v1 manifest, use the "schema1.MediaTypeSignedManifest"
+		// as the media type of artifact
+		artifact.MediaType = schema1.MediaTypeSignedManifest
+		// there is no layer size in v1 manifest, doesn't set the artifact size
+	// OCI manifest/docker manifest v2
+	case v1.MediaTypeImageManifest, schema2.MediaTypeManifest:
+		manifest := &v1.Manifest{}
+		if err := json.Unmarshal(content, manifest); err != nil {
+			return err
+		}
+		// use the "manifest.config.mediatype" as the media type of the artifact
+		artifact.MediaType = manifest.Config.MediaType
+		// set size
+		artifact.Size = int64(len(content)) + manifest.Config.Size
+		for _, layer := range manifest.Layers {
+			artifact.Size += layer.Size
+		}
+		// set annotations
+		artifact.Annotations = manifest.Annotations
+	// OCI index/docker manifest list
+	case v1.MediaTypeImageIndex, manifestlist.MediaTypeManifestList:
+		// the identity of index is still in progress, only handle image index for now
+		// and use the manifestMediaType as the media type of artifact
+		// If we want to support CNAB, we should get the media type from annotation
+		artifact.MediaType = artifact.ManifestMediaType
+
+		index := &v1.Index{}
+		if err := json.Unmarshal(content, index); err != nil {
+			return err
+		}
+		// the size for image index is meaningless, doesn't set it for image index
+		// but it is useful for CNAB or other artifacts, set it when needed
+
+		// set annotations
+		artifact.Annotations = index.Annotations
+		// TODO handle references in resolvers
+	default:
+		return fmt.Errorf("unsupported manifest media type: %s", artifact.ManifestMediaType)
+	}
+
+	resolver, err := resolver.Get(artifact.MediaType)
+	if err != nil {
+		return err
+	}
+	artifact.Type = resolver.ArtifactType()
+	return resolver.Resolve(ctx, content, artifact)
+}
--- a/src/api/artifact/abstractor/abstractor_test.go
+++ b/src/api/artifact/abstractor/abstractor_test.go
@@ -1,0 +1,299 @@
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package abstractor
+
+import (
+	"context"
+	"github.com/docker/distribution/manifest/schema1"
+	"github.com/docker/distribution/manifest/schema2"
+	"github.com/goharbor/harbor/src/api/artifact/abstractor/resolver"
+	"github.com/goharbor/harbor/src/common/models"
+	"github.com/goharbor/harbor/src/pkg/artifact"
+	htesting "github.com/goharbor/harbor/src/testing"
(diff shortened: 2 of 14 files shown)


src/api/artifact/resolver/resolver.go
```diff
@@ -0,0 +1,75 @@
...
+
+// Register resolver, one resolver can handle multiple media types for one kind of artifact
+func Register(resolver Resolver, mediaTypes ...string) error {
+	for _, mediaType := range mediaTypes {
+		_, exist := registry[mediaType]
+		if exist {
+			return fmt.Errorf("resolver to handle media type %s already exists", mediaType)
+		}
+		registry[mediaType] = resolver
+		log.Debugf("resolver to handle media type %s registered", mediaType)
```
renmaosheng: it is better to use info, so that the common workflow can be logged.

src/api/artifact/abstractor.go
```diff
@@ -0,0 +1,112 @@
...
+func (a *abstractor) Abstract(ctx context.Context, repoFullName string, artifact *artifact.Artifact) error {
+	// read manifest content
+	manifestMediaType, content, err := a.blobReader.Read(repoFullName, artifact.Digest, true)
+	if err != nil {
+		return err
+	}
+	artifact.ManifestMediaType = manifestMediaType
+
+	switch artifact.ManifestMediaType {
+	// docker manifest v1
```
reasonerjt: What if we do not support it?

src/api/artifact/blob/cache.go
```diff
@@ -0,0 +1,18 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package blob
+
+// TODO add cache
```
reasonerjt: If we enable cache in registry, there's already a cache of manifest in redis.

src/api/artifact/blob/reader.go
```diff
@@ -0,0 +1,67 @@
...
+		schema1.MediaTypeSignedManifest,
+		schema2.MediaTypeManifest,
+		v1.MediaTypeImageManifest,
+		manifestlist.MediaTypeManifestList,
+		v1.MediaTypeImageIndex,
+	}
+)
+
+// Reader read blob content
+type Reader interface {
```
reasonerjt: Conventionally when you call a type xxxReader it implements `io.Reader`

src/api/artifact/blob/reader.go
```diff
@@ -0,0 +1,67 @@
...
+	// Read the content of blob under the repository. Set the "manifest" as true if whether the blob is a manifest
+	Read(repoFullName, digest string, manifest bool) (mediaType string, content []byte, err error)
```
reasonerjt: Since it does not implement `io.Reader` we should add different funcs like `GetManifest` `GetConfig`...  it's easier for testing.

src/api/artifact/resolver/resolver.go
```diff
@@ -0,0 +1,75 @@
...
+func init() {
+	register(image.NewManifestV2Resolver(), image.ManifestV2ResolverMediaTypes...)
+	register(image.NewManifestV1Resolver(), image.ManifestV1ResolverMediaTypes...)
+	register(image.NewIndexResolver(), image.IndexResolverMediaTypes...)
+}
+
+// Resolver resolves the detail information for a specific kind of artifact
+type Resolver interface {
+	// ArtifactType returns the type of artifact that the resolver handles
+	ArtifactType(context.Context) string
```
reasonerjt: Why is this needed?

src/api/artifact/resolver/resolver.go
```diff
@@ -0,0 +1,75 @@
...
+	// Resolve receives the manifest content, resolves the detail information
+	// from the manifest or the layers referenced by the manifest, and populates
+	// the detail information into the artifact
+	// The layers can be read by reader
+	// The "repoFullName" is the "project_name/repository_name" that the artifact belongs to
+	Resolve(ctx context.Context, manifest []byte, repoFullName string, artifact *artifact.Artifact, reader blob.Reader) error
+}
+
+func register(resolver Resolver, mediaTypes ...string) {
```
reasonerjt: calling a public func from a private one???

src/api/artifact/resolver/resolver.go
```diff
@@ -0,0 +1,75 @@
...
+	if err := Register(resolver, mediaTypes...); err != nil {
+		log.Errorf("failed to register resolver for artifact %s: %v", resolver.ArtifactType(nil), err)
+	}
+}
+
+// Register resolver, one resolver can handle multiple media types for one kind of artifact
+func Register(resolver Resolver, mediaTypes ...string) error {
```
reasonerjt: If it's public it has to be thread-safe.

src/api/artifact/blob/cache.go
ywk253100: It's the cache for both manifest and config layer

src/api/artifact/resolver/resolver.go
ywk253100: `ArtifactType` indicates what artifact type of the resolver handles. For `image.v1ManifestResolver`, `image.v2ManifestResolver` and `image.indexManifestResolver`, the artifact type are all `IMAGE`, the `abstractor` needs the information to populate artifact

src/api/artifact/resolver/resolver.go
ywk253100: `register` is just a util function, it wraps the `Register`

src/api/artifact/abstractor.go
ywk253100: If we don't support it, the images with v1 manifest cannot be pushed into Harbor, and there may be some issue when upgrading if the images with v1 manifest already exist in Harbor 1.x.

src/api/artifact/resolver/resolver.go
```diff
@@ -0,0 +1,75 @@
...
+	for _, mediaType := range mediaTypes {
+		_, exist := registry[mediaType]
+		if exist {
+			return fmt.Errorf("resolver to handle media type %s already exists", mediaType)
```
wy65701436: why not just log and return nil here?

src/api/artifact/blob/reader.go
```diff
@@ -0,0 +1,67 @@
...
+	// TODO read from cache first
+	client, err := coreutils.NewRepositoryClientForLocal("admin", repoFullName)
+	if err != nil {
+		return "", nil, err
+	}
+	if manifest {
+		_, mediaType, payload, err := client.PullManifest(digest, accept)
+		return mediaType, payload, err
+	}
+	_, reader, err := client.PullBlob(digest)
```
wy65701436: `reader` here is confusing with the struct name.

src/api/artifact/resolver/resolver.go
```diff
@@ -0,0 +1,75 @@
...
+	// Resolve receives the manifest content, resolves the detail information
+	// from the manifest or the layers referenced by the manifest, and populates
+	// the detail information into the artifact
+	// The layers can be read by reader
+	// The "repoFullName" is the "project_name/repository_name" that the artifact belongs to
+	Resolve(ctx context.Context, manifest []byte, repoFullName string, artifact *artifact.Artifact, reader blob.Reader) error
+}
+
+func register(resolver Resolver, mediaTypes ...string) {
+	if err := Register(resolver, mediaTypes...); err != nil {
```
wy65701436:
```
if len(mediaTypes) == 0 {
  return nil
}
...
```

src/api/artifact/resolver/resolver.go
```diff
@@ -0,0 +1,75 @@
...
+		log.Errorf("failed to register resolver for artifact %s: %v", resolver.ArtifactType(nil), err)
+	}
+}
+
+// Register resolver, one resolver can handle multiple media types for one kind of artifact
+func Register(resolver Resolver, mediaTypes ...string) error {
+	for _, mediaType := range mediaTypes {
```
wy65701436: if len(mediaTypes) == 0 {
  return nil
}
...

heww:
Should these source codes move to pkg as `pkg/artifact/abstractor`?

src/api/artifact/resolver/resolver.go
```diff
@@ -0,0 +1,75 @@
...
+	"context"
+	"fmt"
+	"github.com/goharbor/harbor/src/api/artifact/blob"
+	"github.com/goharbor/harbor/src/api/artifact/resolver/image"
+	"github.com/goharbor/harbor/src/common/utils/log"
+	"github.com/goharbor/harbor/src/pkg/artifact"
+)
+
+var (
+	registry = map[string]Resolver{}
```
heww: rename `registry` to `resolvers`?

src/api/artifact/resolver/resolver.go
ywk253100: Updated

src/api/artifact/blob/reader.go
ywk253100: Renamed it to `Fetcher`

src/api/artifact/blob/reader.go
ywk253100: Updated

src/api/artifact/resolver/resolver.go
ywk253100: Done

src/api/artifact/blob/reader.go
ywk253100: Updated

src/api/artifact/resolver/resolver.go
heww: It should panic when buildin resolvers register failed. There is not any help to log error info when this problem happened.

src/api/artifact/resolver/resolver.go
reasonerjt: So the values are static, for that case I don't think ctx is needed as a parm.

src/api/artifact/resolver/resolver.go
reasonerjt: I was thinking we can make it private and expose it only when needed.  But that's trivial...

src/api/artifact/abstractor.go
renmaosheng: this is good question, I noticed that docker has deprecated the schemav1 support.

src/api/artifact/blob/cache.go
steven-zou: Is it expensive for sending http request directly to get the data which is not large? If it is not, suggest removing cache out of the table so far. If one day, there is some perf issue appeared, then we can consider it at that time.

src/api/artifact/blob/fetcher.go
```diff
@@ -0,0 +1,75 @@
...
+// Fetcher fetches the content of blob
+type Fetcher interface {
+	// FetchManifest the content of manifest under the repository
+	FetchManifest(repoFullName, digest string) (mediaType string, content []byte, err error)
+	// FetchLayer the content of layer under the repository
+	FetchLayer(repoFullName, digest string) (content []byte, err error)
+}
+
+// NewFetcher returns an instance of the default blob fetcher
+func NewFetcher() Fetcher {
```
steven-zou: If no cache enabled, fetcher is becoming unnecessary.

src/api/artifact/resolver/resolver.go
```diff
@@ -0,0 +1,79 @@
...
+// Resolver resolves the detail information for a specific kind of artifact
+type Resolver interface {
+	// ArtifactType returns the type of artifact that the resolver handles
+	ArtifactType(context.Context) string
+	// Resolve receives the manifest content, resolves the detail information
+	// from the manifest or the layers referenced by the manifest, and populates
+	// the detail information into the artifact
+	// The layers can be fetched by fetcher
+	// The "repoFullName" is the "project_name/repository_name" that the artifact belongs to
+	Resolve(ctx context.Context, manifest []byte, repoFullName string, artifact *artifact.Artifact, fetcher blob.Fetcher) error
```
steven-zou: Based the function of `resovle` method as well as method meaning, I don't think so many parameters are required,

The following pattern is enough.
```go
Resolve(ctx context.Context, manifest []byte, artifact *artifact.Artifact) error
```

The `Fetcher` is just a util tool.

src/api/artifact/resolver/resolver.go
steven-zou: we don't need it.

for loop will check it/

src/api/artifact/blob/cache.go
ywk253100: As there is no OCI registry driver interface yet, we can keep the blob fetcher for now, and remove it util we define the driver interface

ywk253100:
> Should these source codes move to pkg as `pkg/artifact/abstractor`?

No, the abstractor/resolver calls functions inside the pkg

src/api/artifact/resolver/resolver.go
ywk253100: Updated

PR #10402, opened 2020-01-06
feature(oci):implement oci v2 driver

bitsf:
implement oci v2 driver
not support v1 schema
simplify registry api and auth api


src/pkg/ocidriver/ocidriver_test.go
```diff
@@ -0,0 +1,336 @@
...
+	"testing"
+)
+
+type testSuite struct {
+	suite.Suite
+	mockRegistry *httptest.Server
+}
+
+func (t *testSuite) TestCase() {
+	reg, err := NewRegistry(t.mockRegistry.URL, nil)
```
bitsf: todo: transport用于实现auth相关逻辑

src/pkg/ocidriver/ocidriver.go
```diff
@@ -0,0 +1,297 @@
...
+// NewRegistry ...
+func NewRegistry(baseURL string, transport http.RoundTripper) (*OciRegistry, error) {
+	return &OciRegistry{
+		baseURL:   baseURL,
+		transport: transport,
+	}, nil
+}
+
+// List ...
+func (r *OciRegistry) List( /*filter*/ ) ([]*OciRepository, error) {
```
bitsf: 此处List() 对应之前的catalog()方法，命名上采用的是资源操作的风格，大家可以讨论下用哪种命名更合适

bitsf: filter 可按名称过滤，以及分页

renmaosheng:
> not support v1 schema
Wenkai's https://github.com/goharbor/harbor/pull/10379 declare support v1 schema, I think we need to make it consistent.

src/pkg/ocidriver/ocidriver.go
```diff
@@ -0,0 +1,293 @@
...
+// OciRepository ...
+type OciRepository struct {
+	Registry *OciRegistry
+
+	Name string
+}
+
+// OciImage ...
+type OciImage struct {
+	Registry   *OciRegistry
```
renmaosheng: what's the difference between 'Registry' defined here and 'Registry' inside OCIRepository?

src/pkg/ocidriver/ocidriver.go
```diff
@@ -0,0 +1,293 @@
...
+
+// OciManifest ...
+type OciManifest struct {
+	*schema2.DeserializedManifest
+
+	repository *OciRepository
+}
+
+// OciManifestList ...
+type OciManifestList struct {
```
renmaosheng: should we define this as index instead of Manifestlist to align with OCI spec?

src/pkg/ocidriver/ocidriver.go
bitsf: 这个不是继承，是引用，因为在操作 image api时，比如pull，需要用到 Registry中的信息和方法

src/pkg/ocidriver/ocidriver.go
```diff
@@ -0,0 +1,293 @@
...
+	*manifestlist.DeserializedManifestList
+
+	repository *OciRepository
+}
+
+// NewRegistry ...
+func NewRegistry(baseURL string, transport http.RoundTripper) (*OciRegistry, error) {
```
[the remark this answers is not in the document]
bitsf: Do you mean we just use modifer... ?
But we still need to set other settings like https, proxy which is needed setting in transport

src/pkg/ocidriver/ocidriver.go
bitsf: change to OciManifestIndex

src/pkg/ocidriver/ocidriver.go
```diff
@@ -0,0 +1,293 @@
...
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"sort"
+	"strings"
+)
+
+// OciRegistry ...
+type OciRegistry struct {
```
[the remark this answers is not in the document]
bitsf: done

src/pkg/ocidriver/ocirepository.go
```diff
@@ -0,0 +1,257 @@
...
+		return false, nil, nil, nil, err
+	}
+	ref, err := reference.WithDigest(repoNameRef, d)
+	if err != nil {
+		return false, nil, nil, nil, err
+	}
+	return rp.getImage(ref)
+}
+
+func (rp *OciRepository) getImage(ref reference.Named) (isList bool, manifest *OciManifest, manifestList *OciManifestIndex, manifestv1 *OciManifestv1, err error) {
```
ywk253100: In generally, I suggest that if we want to reuse the docker registry client, we can just reuse the concepts of docker registry client directly

src/pkg/ocidriver/ocimanifest.go
```diff
@@ -0,0 +1,33 @@
+package ocidriver
```
ywk253100: Rename the package name please.
The `OCI` is the name of an organization not the registry

src/pkg/ocidriver/ociimage.go
```diff
@@ -0,0 +1,22 @@
+package ocidriver
+
+import (
+	"github.com/opencontainers/go-digest"
+)
+
+// OciImage ...
+type OciImage struct {
```
ywk253100: There is no resource called `image` in the registry API, only `manifest` and `blob`

src/pkg/ocidriver/ocirepository.go
```diff
@@ -0,0 +1,257 @@
...
+	size, err = b.Seek(0, io.SeekEnd)
+	if err != nil {
+		return 0, nil, err
+	}
+	_, err = b.Seek(0, io.SeekStart)
+	return size, b, err
+}
+
+// GetImageByTag ...
+func (rp *OciRepository) GetImageByTag(tag string) (isList bool, manifest *OciManifest, manifestList *OciManifestIndex, manifestv1 *OciManifestv1, err error) {
```
ywk253100: The returning values of this method aren't clear, it's hard for callers to use

src/pkg/ocidriver/ocirepository.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+	ref, err := reference.WithTag(repoNameRef, tag)
+	if err != nil {
+		return false, nil, nil, nil, err
+	}
+	return rp.getImage(ref)
+}
+
+// GetImageByDigest ...
+func (rp *OciRepository) GetImageByDigest(d digest.Digest) (isList bool, manifest *OciManifest, manifestList *OciManifestIndex, manifestv1 *OciManifestv1, err error) {
```
ywk253100: Should expose `Get(ref string)`, the `ref` can be tag or digest, the underlying implementation can be `getByTag` and `getByDigest`

src/pkg/ocidriver/ocirepository.go
ywk253100: Same comment here, no `image`, should be `manifest`

src/pkg/ocidriver/ocirepository.go
```diff
@@ -0,0 +1,257 @@
...
+// OciRepository ...
+type OciRepository struct {
+	Registry *OciRegistry
+
+	Name string
+}
+
+// List ...
+func (rp *OciRepository) List( /*filter*/ ) ([]*OciImage, error) {
+	repoNameRef, err := reference.WithName(rp.Name)
```
ywk253100: It's list tag, not list image

src/pkg/ocidriver/ocirepository.go
```diff
@@ -0,0 +1,257 @@
...
+	if err != nil {
+		return false, nil, nil, nil, err
+	}
+	ms, err := repo.Manifests(nil)
+	if err != nil {
+		return false, nil, nil, nil, err
+	}
+	var m1 distribution.Manifest
+	if digested, isDigested := ref.(reference.Canonical); isDigested {
+		m1, err = ms.Get(nil, digested.Digest())
```
ywk253100: Set the accepting manifest media type here, or the registry will return the manifest as schema1.

bitsf:
current implement of registry client is enough to support OCI, close this first

PR #10421, opened 2020-01-08
Doc reorg for Harbor 1.10

stuclem:
Per request from @jonasrosland and @michmike, this is an initial push of the updated doc structure. I haven't touched the existing docs, so they are still in place alongside these new ones. This is just to see how it looks when these files are pulled into Netlify. 

@xaleeks @michmike can you please approve so that we can merge? Thanks!


michmike:
thanks Stuart

PR #10434, opened 2020-01-09
Add ng-swagger-gen to automatically generate models and services with swagger.yaml

AllForNothing:
To automatically generate models and services with swagger.yaml

### DIFF
--- a/src/portal/package.json
+++ b/src/portal/package.json
@@ -4,6 +4,7 @@
   "description": "Harbor UI with Clarity",
   "angular-cli": {},
   "scripts": {
+    "postinstall": "node scripts/convert-yaml-to-json.js && ng-swagger-gen -i ng-swagger-gen/swagger.json -o ng-swagger-gen && node scripts/delete-swagger-json.js",
     "start": "ng serve --ssl true --ssl-key ssl/server.key --ssl-cert ssl/server.crt --host 0.0.0.0 --proxy-config proxy.config.json",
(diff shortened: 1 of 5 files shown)


jwangyangls:
I think such a modification greatly facilitates the development of the front end; the disadvantage is that ng-swagger-gen cannot support the yaml format file for the time being.

jwangyangls:
Automatically generated files I recommend against uploading to GitHub

src/portal/package.json
```diff
@@ -16,7 +16,11 @@
     "pree2e": "webdriver-manager update",
     "e2e": "protractor",
     "build": "ng build --aot",
-    "release": "ng build --prod"
+    "release": "ng build --prod",
+    "generate_swagger_json": "node scripts\\convert-yaml-to-json.js",
```
jwangyangls: '\\'  maybe not work in some environments

AllForNothing: > “\” “\” maybe not work in some environments

Fixed

heww:
Generated go files from swagger.yaml are excluded in git, I think generated typescript files also need to exclude in git.

heww:
And for legacy APIs we will not update them as little as possible so generate models and client for legacy APIs is not cost effective.

AllForNothing:
> Please exclude generated typescript files in git and not generate files for legacy APIs.

If excluding generated typescript files in git , we need to modify all the related CI for building and running.

@zhoumeina Please help to make a decision

PR #10441, opened 2020-01-10
Delete blob & manifets API

wy65701436:
* add API to delete blob & manifest

Enable registry contronller to delete blob and manifest via digest


reasonerjt:
This is by no means a p0 for 2.0, let's keep it here and review it later.

PR #10443, opened 2020-01-10
Add express module

AllForNothing:
1. Add express module , so we can quickly set up a Api-server from which we can get mocked data


heww:
There are many tools to do APIs mock when APIs are not ready, why do it in portal?

AllForNothing:
> There are many tools to do APIs mock when APIs are not ready, why do it in portal?

Easy to use for js/ts coders

PR #10456, opened 2020-01-13
feat(cicd):stop use travis

bitsf:
stop using travis, instead by github action

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -70,6 +70,8 @@ jobs:
           env
           bash ./tests/showtime.sh ./tests/travis/ut_install.sh
       - name: script
+        env:
+          COVERALLS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -70,6 +70,8 @@ jobs:
           env
           bash ./tests/showtime.sh ./tests/travis/ut_install.sh
       - name: script
+        env:
+          COVERALLS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           echo IP: $IP
           bash ./tests/showtime.sh ./tests/travis/ut_run.sh $IP
```
wy65701436: update the script path to ./tests/ci/*** as we're not rely on travis now.

renmaosheng:
is github actions stable enough to go? thanks.

.github/workflows/CI.yml
bitsf: that fix would be big, leave it to future

PR #10481, opened 2020-01-14
Get the ormer from context instead of creating a new one

ywk253100:
This commit updates the dao methods with the ormer got from context

### DIFF
--- a/src/internal/orm/orm.go
+++ b/src/internal/orm/orm.go
@@ -16,8 +16,7 @@ package orm
 
 import (
 	"context"
-	"fmt"
-
+	"errors"
 	"github.com/astaxie/beego/orm"
 	"github.com/goharbor/harbor/src/common/utils/log"
 )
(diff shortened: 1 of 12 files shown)


src/internal/orm/orm.go
```diff
@@ -16,31 +16,36 @@ package orm
...
 
 type ormKey struct{}
 
 // FromContext returns orm from context
-func FromContext(ctx context.Context) (orm.Ormer, bool) {
+func FromContext(ctx context.Context) (orm.Ormer, error) {
 	o, ok := ctx.Value(ormKey{}).(orm.Ormer)
-	return o, ok
+	if !ok {
+		return nil, errors.New("cannot get the ORM from context")
```
heww: What about to introduct func `MustFromContext(ctx context.Context) orm.Ormer` in this function panic when ormer not found in context?

```go
func MustFromContext(ctx context.Context) orm.Ormer {
        o, ok := FromContext(ctx)
        if !ok {
                panic("ormer required in context")
        }

        return o
}
```



> A panic typically means something went unexpectedly wrong. Mostly we use it to fail fast on errors that shouldn’t occur during normal operation, or that we aren’t prepared to handle gracefully. https://gobyexample.com/panic 

This can make our dao layer easier and fail early.  Because the orm middleware will be added in global, so during normal operation we must can get ormer from ctx, if not it's a big issue.

And of course, `Recover` middleware will also be added for our APIs.

heww: The `orm. NewOrm`  also will panic when database alias named `default` not registered.

https://github.com/astaxie/beego/blob/aa90c67a7586bfb89d6bb195a32a822335c31e43/orm/orm.go#L535-L545

ywk253100: As for as I know, the panic cannot be recovered by a different goroutine, we have such scenario:  handle the request in the handler goroutine and add access log in another goroutine, the panic in the access log goroutine cannot be covered by the middleware.

So seems that we should not use the `orm.NewOrm()` directly, the `orm.NewOrmWithDB()` should be used instead, it doesn't panic

PR #10482, opened 2020-01-14
add code for catalog and list tag API

wy65701436:

### DIFF
--- a/src/server/registry/catalog/catalog.go
+++ b/src/server/registry/catalog/catalog.go
@@ -14,19 +14,117 @@
 
 package catalog
 
-import "net/http"
+import (
+	"encoding/json"
+	"fmt"
+	ierror "github.com/goharbor/harbor/src/internal/error"
+	"github.com/goharbor/harbor/src/pkg/repository"
+	reg_error "github.com/goharbor/harbor/src/server/registry/error"
+	"github.com/goharbor/harbor/src/server/registry/util"
+	"net/http"
+	"sort"
+	"strconv"
+)
 
 // NewHandler returns the handler to handler catalog request
-func NewHandler() http.Handler {
-	return &handler{}
+func NewHandler(repoMgr repository.Manager) http.Handler {
+	return &handler{
+		repoMgr: repoMgr,
+	}
 }
 
-type handler struct{}
+type handler struct {
+	repoMgr repository.Manager
+}
 
+// ServeHTTP ...
 func (h *handler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
-	// TODO implement
+	switch req.Method {
+	case http.MethodGet:
+		h.get(w, req)
+	}
 }
 
-type catalogResponse struct {
+func (h *handler) get(w http.ResponseWriter, req *http.Request) {
+	var maxEntries int
+	var err error
+
+	reqQ := req.URL.Query()
+	lastEntry := reqQ.Get("last")
+	withN := reqQ.Get("n") != ""
+	if withN {
+		maxEntries, err = strconv.Atoi(reqQ.Get("n"))
+		if err != nil || maxEntries < 0 {
+			err := ierror.New(err).WithCode(ierror.BadRequestCode).WithMessage("the N must be a positive int type")
+			reg_error.Handle(w, req, err)
+			return
+		}
+	}
+
+	var repoNames []string
+	// get all repositories
+	// ToDo filter out the untagged repos
+	total, repoRecords, err := h.repoMgr.List(req.Context(), nil)
+	if err != nil {
+		reg_error.Handle(w, req, err)
+		return
+	}
+	if total <= 0 {
+		h.sendResponse(w, req, repoNames)
+		return
+	}
+	for _, r := range repoRecords {
+		repoNames = append(repoNames, r.Name)
+	}
+	sort.Strings(repoNames)
+	if !withN {
+		h.sendResponse(w, req, repoNames)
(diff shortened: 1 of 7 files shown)


src/server/registry/catalog/catalog.go
```diff
@@ -14,19 +14,109 @@
...
+
+func (h *handler) get(w http.ResponseWriter, req *http.Request) {
+	var withPagination = true
+
+	reqQ := req.URL.Query()
+	lastEntry := reqQ.Get("last")
+	maxEntries, err := strconv.Atoi(reqQ.Get("n"))
+	if err != nil || maxEntries < 0 {
+		err := ierror.New(err).WithCode(ierror.BadRequestCode).WithMessage("the N must be a positive int type")
+		reg_error.Handle(w, req, err)
```
ywk253100: The response data should be in the format of `catalogAPIResponse`

src/server/registry/catalog/catalog.go
```diff
@@ -14,19 +14,109 @@
...
+
+	var repos []string
+	// get all repositories
+	total, repoRecords, err := h.repoMgr.List(req.Context(), nil)
+	if err != nil {
+		error.Handle(w, req, err)
+		return
+	}
+	if total <= 0 {
+		h.sendResponse(w, req, repos)
```
ywk253100: The response data should be in the format of `catalogAPIResponse`

src/server/registry/catalog/catalog.go
```diff
@@ -14,19 +14,109 @@
...
+	}
+	for _, r := range repoRecords {
+		repos = append(repos, r.Name)
+	}
+	sort.Strings(repos)
+
+	// with "last", if last page, no link header
+	// without "last", if one page covers "n", no link header
+	if lastEntry != "" {
+		lastEntryIndex := util.IndexString(repos, lastEntry)
```
ywk253100: Check `if lastEntryIndex == -1` ?

src/server/registry/tag/tag.go
```diff
@@ -15,22 +15,165 @@
...
+	repoID, err = h.getRepoID(req)
+	if err != nil {
+		reg_error.Handle(w, req, err)
+		return
+	}
+
+	// get tags ...
+	total, tags, err := h.tagMgr.List(req.Context(), &q.Query{
+		Keywords: map[string]interface{}{
+			"repository_id": repoID,
```
ywk253100: Use `RepositoryID` instead. `repository_id` is column name of the database table, it shouldn't be aware by manager level

src/server/registry/tag/tag.go
```diff
@@ -15,22 +15,165 @@
...
+		},
+	})
+	if err != nil {
+		return 0, err
+	}
+	if total <= 0 {
+		err := ierror.New(errors.New("repositoryNotFound")).WithCode(ierror.NotFoundCode)
+		return 0, err
+	}
+	if total >= 1 {
```
ywk253100: The `if` is useless

src/server/registry/tag/tag.go
```diff
@@ -15,22 +15,165 @@
...
+	total, repoRecord, err := h.repoMgr.List(req.Context(), &q.Query{
+		Keywords: map[string]interface{}{
+			"name": h.repositoryName,
+		},
+	})
+	if err != nil {
+		return 0, err
+	}
+	if total <= 0 {
+		err := ierror.New(errors.New("repositoryNotFound")).WithCode(ierror.NotFoundCode)
```
ywk253100: This is where the error generated, so I think the original cause should be nil:
`ierror.New(nil).WithCode(xxx).WithMessage(xxx)`?
Specify what repository name is in the massage

src/server/registry/tag/tag.go
wy65701436: udpated.

src/server/registry/tag/tag.go
wy65701436: updated.

src/server/registry/catalog/catalog.go
wy65701436: yes, it is.

src/server/registry/tag/tag.go
wy65701436: did

src/server/registry/catalog/catalog.go
```diff
@@ -14,19 +14,117 @@
...
+	}
+}
+
+func (h *handler) get(w http.ResponseWriter, req *http.Request) {
+	var maxEntries int
+	var err error
+
+	reqQ := req.URL.Query()
+	lastEntry := reqQ.Get("last")
+	withN := reqQ.Get("n") == ""
```
heww: `withN := reqQ.Get("n") != ""` ?

wy65701436: good catch.

src/server/registry/catalog/catalog.go
```diff
@@ -14,19 +14,117 @@
...
+		return
+	}
+	for _, r := range repoRecords {
+		repoNames = append(repoNames, r.Name)
+	}
+	sort.Strings(repoNames)
+	if !withN {
+		h.sendResponse(w, req, repoNames)
+		return
+	}
```
heww: This will cause the API returns all repositories when `last` in the query but `n` not in it.

wy65701436: we will not handle the query only with `last`

PR #10490, opened 2020-01-15
move readonly middleware to new v2 handler

wy65701436:

### DIFF
--- a/src/server/middleware/readonly.go
+++ b/src/server/middleware/readonly.go
@@ -1,0 +1,27 @@
+package middleware
+
+import (
+	"github.com/goharbor/harbor/src/common/utils/log"
+	"github.com/goharbor/harbor/src/core/config"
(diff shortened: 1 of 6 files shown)


src/server/middleware/readonly.go
```diff
@@ -0,0 +1,36 @@
...
+		next: next,
+	}
+}
+
+// ServeHTTP ...
+// it should be applied into http.MethodDelete, http.MethodPost, http.MethodPatch, http.MethodPut
+func (rh readonlyHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	if config.ReadOnly() {
+		log.Warningf("The request is prohibited in readonly mode, url is: %s", req.URL.Path)
+		pkgE := internal_errors.New(nil).WithCode("DENIED").WithMessage("The system is in read only mode. Any modification is prohibited.")
```
ywk253100: Let's define the error code in the common package as a const

PR #10517, opened 2020-01-17
Close scan plugin popup message to avoid interfereing logout operation

danfengliu:
Signed-off-by: Danfeng Liu (c) <danfengl@vmware.com>


danfengliu:
This pr need to be verified more times before merge action.

PR #10519, opened 2020-01-17
Wrap the beego router and provide a unified view for users to register routes

ywk253100:
Wrap the beego router and provide a unified view for users to register routes

### DIFF
--- a/src/core/main.go
+++ b/src/core/main.go
@@ -17,14 +17,6 @@ package main
 import (
 	"encoding/gob"
 	"fmt"
-	"net/http"
-	"os"
-	"os/signal"
-	"strconv"
-	"strings"
-	"syscall"
-	"time"
-
 	"github.com/astaxie/beego"
 	_ "github.com/astaxie/beego/session/redis"
 	"github.com/goharbor/harbor/src/common/dao"
@@ -55,8 +47,16 @@ import (
 	"github.com/goharbor/harbor/src/pkg/types"
 	"github.com/goharbor/harbor/src/pkg/version"
 	"github.com/goharbor/harbor/src/replication"
+	"github.com/goharbor/harbor/src/server"
 	"github.com/goharbor/harbor/src/server/middleware/orm"
 	"github.com/goharbor/harbor/src/server/middleware/requestid"
+	"net/http"
+	"os"
+	"os/signal"
+	"strconv"
+	"strings"
+	"syscall"
+	"time"
 )
 
 const (
@@ -254,7 +254,7 @@ func main() {
 	beego.InsertFilter("/*", beego.BeforeRouter, filter.SecurityFilter)
 	beego.InsertFilter("/*", beego.BeforeRouter, filter.ReadonlyFilter)
 
-	initRouters()
+	server.RegisterRoutes()
 
 	syncRegistry := os.Getenv("SYNC_REGISTRY")
 	sync, err := strconv.ParseBool(syncRegistry)
--- a/src/server/router/router.go
+++ b/src/server/router/router.go
@@ -1,0 +1,120 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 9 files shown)


src/core/main.go
```diff
@@ -251,7 +252,7 @@ func main() {
 	beego.InsertFilter("/*", beego.BeforeRouter, filter.SecurityFilter)
 	beego.InsertFilter("/*", beego.BeforeRouter, filter.ReadonlyFilter)
```
steven-zou: TODO REMOVE?

src/server/router/router.go
```diff
@@ -0,0 +1,111 @@
...
+			beego.Delete(r.path, filterFunc)
+		case http.MethodOptions:
+			beego.Options(r.path, filterFunc)
+		}
+	}
+}
+
+// HandlerFunc sets the handler function that handles the request
+func (r *Route) HandlerFunc(f http.HandlerFunc) {
+	r.Handler(http.HandlerFunc(f))
```
steven-zou: `f` has type `http.HandlerFunc`, why need a conversion again?

src/server/router/router.go
```diff
@@ -0,0 +1,111 @@
...
+
+// Route stores the information that matches a request
+type Route struct {
+	methods     []string
+	path        string
+	middlewares []middleware.Middleware
+}
+
+// Methods sets the methods that the route matches
+func (r *Route) Methods(methods ...string) *Route {
```
steven-zou: As you have a method chain way, why need to pass method array here? 

`Method(m string)` seems is enough.

steven-zou: `WithMethod` is more suitable as it is op.

src/server/router/router.go
```diff
@@ -0,0 +1,111 @@
...
+	r.methods = methods
+	return r
+}
+
+// Path sets the path that the route matches. Path uses the beego router path pattern
+func (r *Route) Path(path string) *Route {
```
steven-zou: `WithPath()`?

src/server/router/router.go
```diff
@@ -0,0 +1,111 @@
...
+	r.path = path
+	return r
+}
+
+// Middleware sets the middleware that executed when handling the request
+func (r *Route) Middleware(middleware middleware.Middleware) *Route {
```
steven-zou: `WithMiddleware()`?

I think you just need to keep a front pointer` of the middleware, here 

`m(middleware)` is ok for chaining them together.

src/server/router/router.go
```diff
@@ -0,0 +1,111 @@
...
+	r.middlewares = append(r.middlewares, middleware)
+	return r
+}
+
+// Handler sets the handler that handles the request
+func (r *Route) Handler(handler http.Handler) {
+	filterFunc := beego.FilterFunc(func(ctx *beegocontext.Context) {
+		ctx.Request = ctx.Request.WithContext(
+			context.WithValue(ctx.Request.Context(), contextKeyInput{}, ctx.Input))
+		// TODO remove the WithMiddlewares?
```
steven-zou: If middleware has been chained together, then code here is not required.

src/server/router/router.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+}
+
+// HandlerFunc sets the handler function that handles the request
+func (r *Route) HandlerFunc(f http.HandlerFunc) {
+	r.Handler(http.HandlerFunc(f))
+}
+
+// GetInput returns the input object from the context
+func GetInput(context context.Context) (*beegocontext.BeegoInput, error) {
```
steven-zou: Is it possible to provide more wrapped easy-using methods like Parameters(k string), not just a high-level entry object?

src/server/router/router.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	input, ok := context.Value(contextKeyInput{}).(*beegocontext.BeegoInput)
+	if !ok {
+		return nil, errors.New("input not found in the context")
+	}
+	return input, nil
+}
+
+// Middleware registers the global middleware that executed for all requests that match the path
+func Middleware(path string, middleware middleware.Middleware) {
```
steven-zou: Is it possible to provide a route with path `*` to cover this requirement with the same Route model and flow?

src/server/router/router.go
ywk253100: `Param` function added

src/server/router/router.go
ywk253100: Beego doesn't support specify the path, I'll refactor this function after weiwei'PR merged

src/core/main.go
ywk253100: Issue https://github.com/goharbor/harbor/issues/10532 opened to track the work

PR #10524, opened 2020-01-17
feat(beego): upgrade beego to v1.12 which support middleware

heww:
1. Upgrade beego to v1.12.0
2. Add RequestID middleware to all HTTP requests.
3. Add Orm middleware to v2 and v2.0 APIs.
4. Remove OrmFilter from all HTTP requests.
5. Fix some test cases which cause panic in API controllers.
6. Enable XSRF for test cases of CommonController.
7. Imporve ReadOnly middleware.


src/server/middleware/middleware.go
```diff
@@ -29,3 +29,36 @@ func WithMiddlewares(handler http.Handler, middlewares ...Middleware) http.Handl
...
+// Returning true skips processing the middleware.
+type Skipper func(*http.Request) bool
+
+// DefaultSkipper returns false which processes the middleware.
+func DefaultSkipper(*http.Request) bool {
+	return false
+}
+
+// New make a middleware from fn which type is func(w http.ResponseWriter, r *http.Request, next http.Handler)
+func New(fn func(http.ResponseWriter, *http.Request, http.Handler), skippers ...Skipper) func(http.Handler) http.Handler {
```
steven-zou: From the purpose of this func, you can split the fn to `f (rw, r), http.Handler`, it's more clear

src/server/middleware/orm.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"net/http"
+
+	o "github.com/astaxie/beego/orm"
+	"github.com/goharbor/harbor/src/internal/orm"
+)
+
+// OrmConfig defines the config for Orm middleware.
+type OrmConfig struct {
```
steven-zou: Is `Creator` counted as a cfg?

src/server/middleware/readonly.go
```diff
@@ -1,27 +1,55 @@
...
 	"github.com/goharbor/harbor/src/common/utils/log"
 	"github.com/goharbor/harbor/src/core/config"
 	internal_errors "github.com/goharbor/harbor/src/internal/error"
-	"net/http"
 )
 
-type readonlyHandler struct {
-	next http.Handler
+// ReadOnlyConfig defines the config for ReadOnly middleware.
+type ReadOnlyConfig struct {
```
steven-zou: Similar Q with above ORM middleware?

steven-zou:
@heww @ywk253100 

For creating/wrapping middleware func, we only need one place/utility method to handle such things.

src/server/middleware/middleware.go
heww: `New` func convert `func(http.ResponseWriter, *http.Request, http.Handler)` to a middleware.

src/server/middleware/readonly.go
heww: Yes, `ReadOnly func(*http.Request)` as configuration, so that this middleware can be used in many different scenarios and also make it more easy to test.

heww:
> @heww @ywk253100
> 
> For creating/wrapping middleware func, we only need one place/utility method to handle such things.

The `New` func in this PR is used to create middleware more easy, it was used to wrap ` fn func(http.ResponseWriter, *http.Request, http.Handler)` to be a middleware and its input is fn and output is a middleware.

The `WithMiddlewares` func in this file was used to apply middlewares to a handler, its input is `http.Handler` and many middlewares and output is `http.Handler`.

`New` and `WithMiddlewares` are different things.

src/server/middleware/middleware.go
```diff
@@ -29,3 +29,36 @@ func WithMiddlewares(handler http.Handler, middlewares ...Middleware) http.Handl
 	}
 	return handler
 }
+
+// Skipper defines a function to skip middleware.
+// Returning true skips processing the middleware.
+type Skipper func(*http.Request) bool
```
ywk253100: We have two kinds of Middlewares: global middleware and route specific middleware.
For the route specific ones, the skipper isn't neeeded.
For the global ones, is it possible that we implement a function `RegisterGlobalMiddleware(path string, middleware middleware.Middleware)` based on what we have?

heww: In my opinion, `Skipper` looks more beautiful than `RegisterGlobalMiddleware `. 

For example, Orm and Transaction middlewares are both only needed by `/v2/*` and `/api/v2.0/*`, then create a skipper to skip other URLs for the middlewares,  it looks very  naturally.  For `RegisterGlobalMiddleware` how to deal with this scenario?  Do we need another middleware to decide to skip Orm or Transaction middleware fo the URL?

And there is a middleware named `Readonly` which will block requests when harbor system in readonly mode but pass the requests if its method is not `POST`, `PUT` or `DELETE`. This scenario can be easily to deal with Skipper, but it is difficult with RegisterGlobalMiddleware.

src/server/middleware/middleware.go
```diff
@@ -29,3 +29,49 @@ func WithMiddlewares(handler http.Handler, middlewares ...Middleware) http.Handl
...
+
+// DefaultSkipper returns false which processes the middleware.
+func DefaultSkipper(*http.Request) bool {
```
reasonerjt: If skipper is optional, we don't need a skipper always returns false.  
We just do not set the skipper

src/server/middleware/middleware.go
```diff
@@ -29,3 +29,49 @@ func WithMiddlewares(handler http.Handler, middlewares ...Middleware) http.Handl
...
+	return false
+}
+
+// SafeMethodSkipper returns false when the request method is safe methods
+func SafeMethodSkipper(r *http.Request) bool {
+	// See more for safe method at https://developer.mozilla.org/en-US/docs/Glossary/safe
+	safeMethods := []string{http.MethodGet, http.MethodHead, http.MethodOptions}
```
reasonerjt: For perf consideration.
Use a preset map?

src/server/middleware/middleware.go
```diff
@@ -29,3 +29,49 @@ func WithMiddlewares(handler http.Handler, middlewares ...Middleware) http.Handl
...
+
+	return false
+}
+
+// New make a middleware from fn which type is func(w http.ResponseWriter, r *http.Request, next http.Handler)
+func New(fn func(http.ResponseWriter, *http.Request, http.Handler), skippers ...Skipper) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		var f func(w http.ResponseWriter, r *http.Request)
+
+		if len(skippers) == 0 {
```
reasonerjt: Why only handle one skipper?

src/server/middleware/orm.go
heww: This make the middleware configurable and easy to test.

src/server/middleware/middleware.go
```diff
@@ -29,3 +29,33 @@ func WithMiddlewares(handler http.Handler, middlewares ...Middleware) http.Handl
...
+// Skipper defines a function to skip middleware.
+// Returning true skips processing the middleware.
+type Skipper func(*http.Request) bool
+
+// New make a middleware from fn which type is func(w http.ResponseWriter, r *http.Request, next http.Handler)
+func New(fn func(http.ResponseWriter, *http.Request, http.Handler), skippers ...Skipper) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		var f func(w http.ResponseWriter, r *http.Request)
+
+		if len(skippers) == 0 {
```
reasonerjt: What's the point of this `if`?
You can always call a for loop when skippers is empty it still works as expected?

PR #10525, opened 2020-01-19
Implement the resolver for helm chart

ywk253100:
Implement the resolver for helm chart

### DIFF
--- a/src/api/artifact/abstractor/resolver/chart/chart.go
+++ b/src/api/artifact/abstractor/resolver/chart/chart.go
@@ -1,3 +1,7 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
 //    http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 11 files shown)


src/api/artifact/abstractor/resolver/chart/chart.go
```diff
@@ -9,3 +13,68 @@
...
+	// ArtifactTypeChart defines the artifact type for helm chart
+	ArtifactTypeChart = "CHART"
+)
+
+func init() {
+	resolver := &resolver{
+		repoMgr:     repository.Mgr,
+		blobFetcher: blob.Fcher,
+	}
+	if err := resolv.Register(resolver, "application/vnd.cncf.helm.config.v1+json"); err != nil {
```
steven-zou: For such value, we should have a set of const values for easy reference. Even including the value like `ArtifactTypeChart = "CHART"`

src/api/artifact/abstractor/resolver/chart/chart.go
```diff
@@ -9,3 +13,68 @@
...
+	m := &v1.Manifest{}
+	if err := json.Unmarshal(manifest, m); err != nil {
+		return err
+	}
+	digest := m.Config.Digest.String()
+	layer, err := r.blobFetcher.FetchLayer(repository.Name, digest)
+	if err != nil {
+		return err
+	}
+	// TODO should we abstract all values?
```
steven-zou: I think we should extract the data similar to the previous chart info block and put the all the left ones into the extra properties which are rendered as k-v pairs to the users

steven-zou: Seems most of the info abstracted from the yaml can only be stored in the extra properties

src/api/artifact/abstractor/resolver/chart/chart.go
ywk253100: The media types of images are imported from distribution/oci packages, we didn't define them in Harbor.

For the `ArtifactTypeChart`, it's defined by the resolvers themselves, so define it in the resolver's package is reasonable

PR #10526, opened 2020-01-19
feat(swagger): implement SendError for BaseAPI handler

heww:
implement SendError for BaseAPI handler

### DIFF
--- a/src/server/error/error.go
+++ b/src/server/error/error.go
@@ -16,8 +16,11 @@ package error
 
 import (
 	"errors"
-	ierror "github.com/goharbor/harbor/src/internal/error"
 	"net/http"
+
+	"github.com/go-openapi/runtime"
+	"github.com/go-openapi/runtime/middleware"
+	ierror "github.com/goharbor/harbor/src/internal/error"
 )
 
 var (
@@ -34,14 +37,7 @@ var (
 
 // APIError generates the HTTP status code and error payload based on the input err
 func APIError(err error) (int, string) {
-	var e *ierror.Error
-	statusCode := 0
-	if errors.As(err, &e) {
-		statusCode = getHTTPStatusCode(e.Code)
-	} else {
-		statusCode = http.StatusInternalServerError
-	}
-	return statusCode, ierror.NewErrs(err).Error()
+	return getHTTPStatusCode(ierror.ErrCode(err)), ierror.NewErrs(err).Error()
 }
 
 func getHTTPStatusCode(errCode string) int {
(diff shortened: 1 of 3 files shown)


src/server/error/error.go
```diff
@@ -51,3 +47,33 @@ func getHTTPStatusCode(errCode string) int {
...
+	rw.WriteHeader(getHTTPStatusCode(code))
+
+	var e *ierror.Error
+	if !errors.As(r.err, &e) {
+		e = &ierror.Error{
+			Code:    code,
+			Message: r.err.Error(),
+		}
+	}
+
```
ywk253100: Call `APIError()` instead?

src/server/error/error.go
```diff
@@ -51,3 +47,33 @@ func getHTTPStatusCode(errCode string) int {
...
+}
+
+// WriteResponse ...
+func (r *ErrResponder) WriteResponse(rw http.ResponseWriter, producer runtime.Producer) {
+	code := ierror.ErrCode(r.err)
+	rw.WriteHeader(getHTTPStatusCode(code))
+
+	var e *ierror.Error
+	if !errors.As(r.err, &e) {
+		e = &ierror.Error{
```
ywk253100: The response should be an error array, is this covered by other logic?

src/server/v2.0/handler/base.go
```diff
@@ -30,6 +31,5 @@ func (*BaseAPI) Prepare(ctx context.Context, operation string, params interface{
 
 // SendError returns response for the err
 func (*BaseAPI) SendError(ctx context.Context, err error) middleware.Responder {
-	// TODO: implement send error
-	return nil
+	return errs.NewErrResponder(err)
```
ywk253100: Let's move this file out of v2.0 as it is shared by all versions of APIs

src/server/error/error.go
heww: No, management APIs (v2.0) not needed keep the same as the oci error response.

src/server/v2.0/handler/base.go
heww: Yes, it is shared by all versions of APIs (swagger generated APIs) so there is a `ErrResponder` in `server/error` pkg, `SendError` in `BaseAPI` is a shortcut.

src/server/error/error.go
ywk253100: I think @wy65701436 has already implemented to return the error array for Harbor APIs in PR https://github.com/goharbor/harbor/pull/10324?

I'm OK for both these two solutions, but we must be consistent.

PR #10535, opened 2020-01-19
add immutable tag middleware into new v2 handler

wy65701436:

### DIFF
--- a/src/server/middleware/immutable/deletemf.go
+++ b/src/server/middleware/immutable/deletemf.go
@@ -1,0 +1,101 @@
+package immutable
+
+import (
+	"errors"
+	"fmt"
+	common_util "github.com/goharbor/harbor/src/common/utils"
+	internal_errors "github.com/goharbor/harbor/src/internal/error"
+	"github.com/goharbor/harbor/src/pkg/art"
+	"github.com/goharbor/harbor/src/pkg/artifact"
+	"github.com/goharbor/harbor/src/pkg/immutabletag/match/rule"
+	"github.com/goharbor/harbor/src/pkg/q"
+	"github.com/goharbor/harbor/src/pkg/repository"
+	"github.com/goharbor/harbor/src/pkg/tag"
+	"github.com/goharbor/harbor/src/server/middleware"
+	"net/http"
+)
+
+// MiddlewareDelete ...
+func MiddlewareDelete() func(http.Handler) http.Handler {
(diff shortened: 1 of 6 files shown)


heww:
Two implementions for immutable tag middleware looks not very beatulful.

```go
type Config {
	Extactor func(*http.Request) <-chan Tag
}

func NewWithConfig(config Config) func(http.Handler) http.Handler {
	return middelware.New(func(w http.ResponseWriter, r *http.Request, next http.Handler) {
		info, ok := middleware.ManifestInfoFromContext(req.Context())
		if !ok {
			// handle error
			return
		}

		projectID := info.ProjectID
		_, repoName := common_util.ParseRepository(mf.Repository)

		tags := config.Extactor(r)
		for _, tag := range tags {
			matched, err := rule.NewRuleMatcher(mf.ProjectID).Match(art.Candidate{
				Repository:  repoName,
				Tag:         tag.Name,
				NamespaceID: ProjectID,
			})
			if err != nil {
				return err
			}
			if matched {
				// handle for immutable tag
				return
			}
		}

		next.ServeHTTP(w, r)
	})
}


func ManifestDeletionExtactor(r *http.Request)  []Tag {
	// Get all tags for the manifest which will be deleted
	ch := make(chan int)
	artifact := getArtifactInRepoByDigest(repo, digest)

	return getTagsOfActifact(actifact.ID)
}

func TagDeletionExtactor(r *http.Request) []Tag {
	// Get the tag info which will be deleted by the v2.0 API
}

func ManifestCreationExtactor(r *http.Request) []Tag {
	// Get the tag info from the manifest
}

```

src/server/middleware/immutable/deletemf.go
```diff
@@ -0,0 +1,102 @@
...
+		},
+	})
+	if err != nil {
+		return err
+	}
+	if total == 0 {
+		return nil
+	}
+
+	total, afs, err := artifact.Mgr.List(req.Context(), &q.Query{
```
ywk253100: Should call the list of artifact controller, it returns all the tags and immutable status. As the immutable status isn't populated into the returning result of the function yet, open an issue https://github.com/goharbor/harbor/issues/10545 to track the refactor work.

wy65701436:
@heww this middelware will be refactored, but it depends on the code of artifact controller. Just let it be, I will update code after the artifact controller code is ready.

PR #10536, opened 2020-01-19
Add ut coverage report

### LINKED ISSUE #10405: Integrate coverage report in git hub action for frontend
Since we migrate the unit test from travis to git hub action. It is better to show frontend coverage report there. We need to investigate how to show coverage report.

AllForNothing:
Fix #10405 
Add ut coverage reports provided by codecov for every PR(see below report from a robot account)
The coverage is the sum of ui-ut coverage and backend-ut coverage(can not be provided individually so far). We can see individual reports on detail page

### DIFF
--- a/.github/codecov-ut-config/codecov.yml
+++ b/.github/codecov-ut-config/codecov.yml
@@ -1,0 +1,25 @@
+codecov:
+  require_ci_to_pass: no
+
+coverage:
+  precision: 2
+  round: down
+  range: "50...100"
+
+  status:
+    project: yes
+    patch: yes
+    changes: no
+
+parsers:
+  gcov:
+    branch_detection:
+      conditional: yes
+      loop: yes
+      method: no
+      macro: no
+
+comment:
+  layout: "reach,diff,flags,tree"
+  behavior: default
+  require_changes: no
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -14,8 +14,9 @@ env:
    REDIS_HOST: localhost
    REG_VERSION: v2.7.1-patch-2819-2553
    UI_BUILDER_VERSION: 1.6.0
+   CODECOV_TOKEN: 89c7b95b-dbce-431e-8cb3-55691a9dcc64 #codecov token: visit https://github.com/codecov/codecov-action to learn more
 
-on: 
+on:
   pull_request:
   push:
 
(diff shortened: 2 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -14,16 +14,17 @@ env:
    REDIS_HOST: localhost
    REG_VERSION: v2.7.1-patch-2819-2553
    UI_BUILDER_VERSION: 1.6.0
+   CODECOV_TOKEN: 89c7b95b-dbce-431e-8cb3-55691a9dcc64 #codecov token: visit https://github.com/codecov/codecov-action to learn more
```
steven-zou: It's a plaintext secret

AllForNothing: because I can not think of a better solution

steven-zou: @AllForNothing 

Have you tried other possible ways?

AllForNothing: https://github.community/t5/GitHub-Actions/Make-secrets-available-to-builds-of-forks/m-p/30685#M511    I found the answer here.  It indicate there is no other way so far

.github/codecov-ut-config/codecov.yml
```diff
@@ -0,0 +1,25 @@
+codecov:
+  require_ci_to_pass: yes
```
bitsf: i think we don't need ci to pass to show this coverage

.github/codecov-ut-config/codecov.yml
```diff
@@ -0,0 +1,25 @@
...
+
+coverage:
+  precision: 2
+  round: down
+  range: "70...100"
```
bitsf: range can be 50 to 100

.github/workflows/CI.yml
bitsf: it seems the only risk is hacker may post a fake test coverage report with this token

bitsf:
you should not use fork branch AllForNothing:master to submit pr, it is easily overwritten by mistake

.github/codecov-ut-config/codecov.yml
AllForNothing: modified

.github/codecov-ut-config/codecov.yml
AllForNothing: modified

PR #10537, opened 2020-01-19
Implement the listing artifact API

ywk253100:
Implement the listing artifact API

### DIFF
--- a/api/v2.0/swagger.yaml
+++ b/api/v2.0/swagger.yaml
@@ -18,21 +18,52 @@ securityDefinitions:
 security:
   - basicAuth: []
 paths:
-  /projects/{project_id}/repositories/{repository_id}/artifacts/{artifact_id}:
+  /projects/{project_name}/repositories/{repository_name}/artifacts/{digest}:
     get:
-      summary: Read artifact by id
-      description: endpoint returns artifact by id
+      summary: Get the specific artifact
+      description: Get the artifact specified by the digest under the project and repository
       tags:
         - artifact
-      operationId: readArtifact
+      operationId: getArtifact
       parameters:
         - $ref: '#/parameters/requestId'
-        - $ref: '#/parameters/projectId'
-        - $ref: '#/parameters/repositoryId'
-        - $ref: '#/parameters/artifactId'
+        - $ref: '#/parameters/projectName'
+        - $ref: '#/parameters/repositoryName'
+        - $ref: '#/parameters/digest'
+        - name: with_tag
+          in: query
+          description: Specify whether the tags are inclued inside the returning artifacts
+          type: boolean
+          required: false
+          default: true
+        - name: with_label
+          in: query
+          description: Specify whether the labels are inclued inside the returning artifacts
+          type: boolean
+          required: false
+          default: false
+        - name: with_scan_overview
+          in: query
+          description: Specify whether the scan overview is inclued inside the returning artifacts
+          type: boolean
+          required: false
+          default: false
+        # should be in tag level
+        - name: with_signatrue
+          in: query
+          description: Specify whether the signature is inclued inside the returning artifacts
+          type: boolean
+          required: false
+          default: false
+        - name: with_immutable_status
+          in: query
+          description: Specify whether the immutable status is inclued inside the tags of the returning artifacts. Only works when setting "with_tag=true"
+          type: boolean
+          required: false
+          default: false
       responses:
         '200':
-          description: OK
+          description: Success
           schema:
             $ref: '#/definitions/Artifact'
         '401':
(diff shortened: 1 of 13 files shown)


api/v2.0/swagger.yaml
```diff
@@ -18,21 +18,52 @@ securityDefinitions:
...
+  /projects/{project_name}/repositories/{repository_name}/artifacts/{digest}:
     get:
-      summary: Read artifact by id
-      description: endpoint returns artifact by id
+      summary: Get the specific artifact
+      description: Get the artifact specified by the digest under the project and repository
       tags:
         - artifact
-      operationId: readArtifact
+      operationId: get
```
heww: The operationId **MUST** be unique among all operations described in the API, operationIds of `get`, `delete` and `list` in `artifact` group are not good practical here, it means other API groups can not use operationId like `get`, `delete` and `list`.

Let's keep the operationIds like the before.

See more https://github.com/go-swagger/go-swagger/commit/8910d8f1bccae424ccbd8b4bb0f4749578a37284 and https://github.com/go-swagger/go-swagger/issues/104

src/server/v2.0/handler/handler.go
```diff
@@ -14,6 +14,8 @@
 
 package handler
 
+// TODO move this file out of v2.0 folder as this is common for all versions of API
```
heww: Why this is common for all versions of APIs ?  

`restapi.HandlerAPI` and `restapi.Config` are both related to special API version.

src/server/v2.0/handler/artifact.go
```diff
@@ -16,32 +16,157 @@ package handler
...
-	return operation.NewReadArtifactOK()
+func (a *artifactAPI) Delete(ctx context.Context, params operation.DeleteParams) middleware.Responder {
+	// TODO implement
+	return operation.NewDeleteOK()
 }
 
-// NewArtifactAPI returns API of artifacts
-func NewArtifactAPI() *ArtifactAPI {
-	return &ArtifactAPI{}
+func convert(art *artifact.Artifact) *models.Artifact {
```
heww: Suggest to rename `convert` to `convertArtifact` because maybe there are others APIs also needed to convert the resource from controller models to restful api models.

src/server/v2.0/handler/handler.go
ywk253100: My bad, put it on the wrong file, the file that I said is `base.go`

api/v2.0/swagger.yaml
ywk253100: Updated

src/server/v2.0/handler/artifact.go
ywk253100: Updated

src/server/v2.0/handler/artifact.go
```diff
@@ -16,32 +16,157 @@ package handler
...
+	if params.Type != nil {
+		query.Keywords["Type"] = *(params.Type)
+	}
+	if params.Page != nil {
+		query.PageNumber = *(params.Page)
+	}
+	if params.PageSize != nil {
+		query.PageSize = *(params.PageSize)
+	}
+	repository, err := a.repoMgr.GetByName(ctx, fmt.Sprintf("%s/%s", params.ProjectName, params.RepositoryName))
```
heww: Maybe needed to the controller of respository here.

ywk253100: Open issue https://github.com/goharbor/harbor/issues/10552 for that

src/api/artifact/model.go
```diff
@@ -50,13 +50,14 @@ type ResourceLink struct {
 type Option struct {
```
steven-zou: We have already had so many options, again, maybe the function option is more suitable now.

https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

src/pkg/repository/manager.go
```diff
@@ -65,6 +67,18 @@ func (m *manager) Get(ctx context.Context, id int64) (*models.RepoRecord, error)
 	return m.dao.Get(ctx, id)
 }
 
+func (m *manager) GetByName(ctx context.Context, name string) (repository *models.RepoRecord, err error) {
+	_, repositories, err := m.List(ctx, &q.Query{
+		Keywords: map[string]interface{}{
+			"Name": name,
```
steven-zou: Actually the key of the keywords should be aligned with the property name in the JSON model.

steven-zou: lower case : "name"?

src/server/v2.0/handler/artifact.go
steven-zou: Another option might refer to the repo mgr and project mgr in the artifact controller, not in the handler.s

src/server/v2.0/handler/artifact.go
```diff
@@ -16,32 +16,157 @@ package handler
...
-	return operation.NewReadArtifactOK()
+func (a *artifactAPI) DeleteArtifact(ctx context.Context, params operation.DeleteArtifactParams) middleware.Responder {
+	// TODO implement
+	return operation.NewDeleteArtifactOK()
 }
 
-// NewArtifactAPI returns API of artifacts
-func NewArtifactAPI() *ArtifactAPI {
-	return &ArtifactAPI{}
+func convertArtifact(art *artifact.Artifact) *models.Artifact {
```
steven-zou: Did I remember this function is defined as a method of the API model?

src/testing/repository_manager.go
```diff
@@ -38,6 +38,12 @@ func (f *FakeRepositoryManager) Get(ctx context.Context, id int64) (*models.Repo
 	return args.Get(0).(*models.RepoRecord), args.Error(1)
 }
 
+// GetByName ...
+func (f *FakeRepositoryManager) GetByName(ctx context.Context, name string) (*models.RepoRecord, error) {
+	args := f.Called()
+	return args.Get(0).(*models.RepoRecord), args.Error(1)
```
steven-zou: For such a simple return statement, you need to confirm no error case happened in your case otherwise `args.Get(0).(*models.RepoRecord)` will be panic

reasonerjt:
For all the TODOs in the code, are there issues opened for each of them?

src/server/v2.0/handler/artifact.go
ywk253100: This is a different one, it converts the api model to the one generated by swagger

src/pkg/repository/manager.go
ywk253100: No json annotation defined for the artifact model used by the controller. What users can see is only the property name of the model. The convert is handled by beego orm

ywk253100:
> For all the TODOs in the code, are there issues opened for each of them?

@reasonerjt , not every TODO items has its issue, but we'll go through all the items later and make sure every item is handled correctly in the end

PR #10541, opened 2020-01-20
Try to parse the type of the artifact based on the media type when no resolver found

ywk253100:
Try to parse the type of the artifact based on the media type when no resolver found, if parse failed, set it's type to unknown. This won't block the pushing for artifacts that has no resolver registered in Harbor

### DIFF
--- a/src/api/artifact/abstractor/abstractor.go
+++ b/src/api/artifact/abstractor/abstractor.go
@@ -26,7 +26,14 @@ import (
 	"github.com/goharbor/harbor/src/pkg/artifact"
 	"github.com/goharbor/harbor/src/pkg/repository"
 	"github.com/opencontainers/image-spec/specs-go/v1"
+	"regexp"
+	"strings"
 )
+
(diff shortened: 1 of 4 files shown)


src/api/artifact/abstractor/abstractor.go
```diff
@@ -109,10 +116,25 @@ func (a *abstractor) Abstract(ctx context.Context, artifact *artifact.Artifact)
...
+	artifact.Type = parseArtifactType(artifact.MediaType)
+	// still can not get the artifact type, set it to unknown
+	if len(artifact.Type) == 0 {
+		artifact.Type = ArtifactTypeUnknown
+	}
+	return nil
+}
+
+func parseArtifactType(mediaType string) string {
+	strs := artifactTypeRegExp.FindStringSubmatch(mediaType)
```
reasonerjt: It would be slightly better if you let `parseArtifactType ` return unknown. And you should clearly print a log like `xxxx is not recognized returning unknown artifact type`

PR #10542, opened 2020-01-20
add regtoken middleware in new v2 handler

wy65701436:
To move the regtoken middlware to new v2 handler framework, it parses the docker pull bearer token and check whether it's a scanner pull.

### DIFF
--- a/src/server/middleware/util.go
+++ b/src/server/middleware/util.go
@@ -9,6 +9,8 @@ type contextKey string
 const (
 	// manifestInfoKey the context key for manifest info
 	manifestInfoKey = contextKey("ManifestInfo")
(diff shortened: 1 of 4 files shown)


src/server/middleware/util.go
```diff
@@ -29,3 +31,8 @@ func ManifestInfoFromContext(ctx context.Context) (*ManifestInfo, bool) {
 	info, ok := ctx.Value(manifestInfoKey).(*ManifestInfo)
 	return info, ok
 }
+
+// NewScannerPullContext returns context with policy check info
+func NewScannerPullContext(ctx context.Context, scannerPull bool) context.Context {
+	return context.WithValue(ctx, ScannerPullCtxKey, scannerPull)
+}
```
heww: Suggest to move it to `regtoken` pkg

PR #10549, opened 2020-01-20
add Middleware policycheck into v2 handler

wy65701436:
1, add content trust middleware in new v2 handler
2, add vulnable middlware in new v2 hanlder


wy65701436:
UT will be in another PR, already tested the middleware on my testing env.

src/go.mod
```diff
@@ -35,12 +35,7 @@ require (
 	github.com/garyburd/redigo v1.6.0
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-openapi/errors v0.19.2
-	github.com/go-openapi/loads v0.19.3
 	github.com/go-openapi/runtime v0.19.5
-	github.com/go-openapi/spec v0.19.3
-	github.com/go-openapi/strfmt v0.19.3
-	github.com/go-openapi/swag v0.19.5
-	github.com/go-openapi/validate v0.19.3
```
heww: These pkgs used by swagger generated APIs and we should not delete them from go.mod.

src/internal/error/errors.go
```diff
@@ -93,6 +93,9 @@ const (
 
 	// DENIED it's used by middleware(readonly, vul and content trust) and returned to docker client to index the request is denied.
 	DENIED = "DENIED"
+
+	// PROJECTPOLICYVIOLATION ...
+	PROJECTPOLICYVIOLATION = "PROJECT_POLICY_VIOLATION"
```
ywk253100: Use `camel case` for the variable name, the same to `DENIED`

src/internal/error/errors.go
```diff
@@ -92,6 +92,8 @@ const (
 	GeneralCode = "UNKNOWN"
 	// DENIED it's used by middleware(readonly, vul and content trust) and returned to docker client to index the request is denied.
 	DENIED = "DENIED"
+	// PROJECTPOLICYVIOLATION ...
+	PROJECTPOLICYVIOLATION = "PROJECTPOLICYVIOLATION"
```
ywk253100: ProjectPolicyViolation = "PROJECT_POLICY_VIOLATION"

ywk253100: The same to the `DENIED`, it should be `DeniedCode`

ywk253100: And add them to the HTTP status code map https://github.com/goharbor/harbor/blob/master/src/server/error/error.go#L27

ywk253100: As the error code and HTTP status code are defined in two files, seems we should take the suggestion of @steven-zou that introduce a `Descriptor` for each common error, but this is out of the scope of this PR

wy65701436: the code here is just an specification(not a http code) for docker client. No need to add it.

wy65701436: and I will enhance error handling as issue mentioned. https://github.com/goharbor/harbor/issues/10553

PR #10550, opened 2020-01-20
feat(middleware): add transaction middleware for v2 and v2.0 APIs

heww:
Add transaction middleware for v2 and v2.0 APIs.

### DIFF
--- a/src/core/main.go
+++ b/src/core/main.go
@@ -17,6 +17,12 @@ package main
 import (
 	"encoding/gob"
 	"fmt"
+	"os"
+	"os/signal"
+	"strconv"
+	"syscall"
+	"time"
+
 	"github.com/astaxie/beego"
 	_ "github.com/astaxie/beego/session/redis"
 	"github.com/goharbor/harbor/src/common/dao"
@@ -48,15 +54,6 @@ import (
 	"github.com/goharbor/harbor/src/pkg/version"
 	"github.com/goharbor/harbor/src/replication"
 	"github.com/goharbor/harbor/src/server"
-	"github.com/goharbor/harbor/src/server/middleware/orm"
-	"github.com/goharbor/harbor/src/server/middleware/requestid"
-	"net/http"
-	"os"
-	"os/signal"
-	"strconv"
-	"strings"
-	"syscall"
-	"time"
 )
 
 const (
@@ -292,21 +289,5 @@ func main() {
 
 	log.Infof("Version: %s, Git commit: %s", version.ReleaseVersion, version.GitCommit)
 
-	middlewares := []beego.MiddleWare{
-		requestid.Middleware(),
-		orm.Middleware(legacyAPISkipper),
-	}
-	beego.RunWithMiddleWares("", middlewares...)
-
-}
-
-// legacyAPISkipper skip middleware for legacy APIs
-func legacyAPISkipper(r *http.Request) bool {
-	for _, prefix := range []string{"/v2/", "/api/v2.0/"} {
-		if strings.HasPrefix(r.URL.Path, prefix) {
-			return false
-		}
-	}
-
-	return true
+	beego.RunWithMiddleWares("", middlewares.MiddleWares()...)
 }
--- a/src/core/middlewares/middlewares.go
+++ b/src/core/middlewares/middlewares.go
@@ -1,0 +1,56 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package middlewares
+
+import (
+	"net/http"
+	"regexp"
+	"strings"
+
+	"github.com/astaxie/beego"
+	"github.com/docker/distribution/reference"
+	"github.com/goharbor/harbor/src/server/middleware"
+	"github.com/goharbor/harbor/src/server/middleware/orm"
+	"github.com/goharbor/harbor/src/server/middleware/requestid"
+	"github.com/goharbor/harbor/src/server/middleware/transaction"
+)
+
+var (
+	blobURLRe = regexp.MustCompile("^/v2/(" + reference.NameRegexp.String() + ")/blobs/" + reference.DigestRegexp.String())
+
+	// fetchBlobAPISkipper skip transaction middleware for fetch blob API
+	// because transaction use the ResponseBuffer for the response which will degrade the performance for fetch blob
+	fetchBlobAPISkipper = middleware.MethodAndPathSkipper(http.MethodGet, blobURLRe)
+)
+
+// legacyAPISkipper skip middleware for legacy APIs
+func legacyAPISkipper(r *http.Request) bool {
+	for _, prefix := range []string{"/v2/", "/api/v2.0/"} {
+		if strings.HasPrefix(r.URL.Path, prefix) {
+			return false
+		}
+	}
+
+	return true
+}
+
+// MiddleWares returns global middlewares
+func MiddleWares() []beego.MiddleWare {
+	return []beego.MiddleWare{
+		requestid.Middleware(),
+		orm.Middleware(legacyAPISkipper),
+		transaction.Middleware(legacyAPISkipper, fetchBlobAPISkipper),
+	}
+}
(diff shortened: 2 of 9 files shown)


src/server/middleware/transaction/transaction.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+
+		if err := orm.WithTransaction(h)(r.Context()); err != nil && err != errNonSuccess {
+			// 500 response for db error
+			w.WriteHeader(http.StatusInternalServerError)
+		}
+	}, skippers...)
+}
+
+type statusRecorder struct {
```
ywk253100: Use `httptest.ResponseRecorder` instead, there are some constraints for the methods of `http.ResponseWriter` that `statusRecorder` doesn't meet

src/server/middleware/transaction/transaction.go
```diff
@@ -0,0 +1,83 @@
...
+			if !config.Acceptor(rec.statusCode) {
+				return errNonSuccess
+			}
+
+			return nil
+		}
+
+		if err := orm.WithTransaction(h)(r.Context()); err != nil && err != errNonSuccess {
+			// 500 response for db error
+			w.WriteHeader(http.StatusInternalServerError)
```
ywk253100: The HTTP status code has been set by the handler, the second call for `WriteHeader` doesn't work. Just log the error?

src/server/middleware/transaction/transaction.go
```diff
@@ -0,0 +1,83 @@
...
+// Config defines the config for Transaction middleware.
+type Config struct {
+	// Acceptor defines a function to commit transaction
+	Acceptor func(statusCode int) bool
+}
+
+var (
+	// DefaultConfig default transaction config
+	DefaultConfig = Config{
+		// acceptor which will commit the transaction when status code between 200 to 308 as success
```
ywk253100: You mean 399?

src/core/main.go
```diff
@@ -295,6 +296,7 @@ func main() {
 	middlewares := []beego.MiddleWare{
 		requestid.Middleware(),
 		orm.Middleware(legacyAPISkipper),
+		transaction.Middleware(legacyAPISkipper),
```
ywk253100: Shouldn't the orm and transaction be in a same middleware?

src/server/middleware/transaction/transaction.go
heww: `httptest.ResponseRecorder` use `bytes.Buffer` to store the body content, when the status is success we need to copy the content from `bytes.Buffer` to `http.ResponseWriter`.

This is ok for API's response because its size is not very large, but the response of fetch blob API in v2 may be very very large, the copy may degrade the performance.

src/server/middleware/transaction/transaction.go
heww: There are two different scenarios for this error:
1. Begin the transaction failed, `WriteHeader`works.
2. Commit or rollback the transaction failed, `WriteHeader` does not work, it will be ignore by the ResponseWriter.

En, I will log the error.

ywk253100: Got it, could you add some comment to explain it? And we should response with standard API error here

src/server/middleware/transaction/transaction.go
ywk253100: Understood. The `httptest.ResponseRcorder` isn't the correct solution.
But your implement must meet the constraint of the `http.ResponseWriter` interface:
Should rewrite the `Write` method as it will set the HTTP status code as well?
Make sure the HTTP status  code can only be set once.

How about implement it in the `internal` package and make it as a util so that others can use it if needed?

heww: According to the comments of `Write` method in the `http.ResponseWriter`, it writes the data to the connection as part of an HTTP reply. If `WriteHeader` has not yet been called, Write calls WriteHeader(http.StatusOK) before writing the data.

https://github.com/golang/go/blob/master/src/net/http/server.go#L118-L121

It is ok for statusRecorder when we only need to get the status code of the response.

src/core/main.go
heww: When orm middleware is not applied for the request but the transaction middleware is applied, the transaction middleware will be failed. I think this is expected result.

src/server/middleware/transaction/transaction.go
heww: The ResponseWriter for the next handler changed to ResponseBuffer, so this problem was resolved.

src/server/middleware/transaction/transaction.go
heww: Changed to use ResponseBuffer.

src/core/middlewares/middlewares.go
```diff
@@ -0,0 +1,60 @@
...
+		if strings.HasPrefix(r.URL.Path, prefix) {
+			return false
+		}
+	}
+
+	return true
+}
+
+var (
+	blobURLRe = regexp.MustCompile(`^/v2/((?:[a-z0-9]+(?:[._-][a-z0-9]+)*/)+)blobs/([\w][\w.:-]{0,127})`)
```
bitsf: how about reuse src/common/utils/registry/auth/path.go:29

src/server/middleware/transaction/transaction.go
```diff
@@ -0,0 +1,84 @@
...
+			// statusCode between 200 to 399 as success
+			return statusCode >= http.StatusOK && statusCode < http.StatusBadRequest
+		},
+	}
+
+	errNonSuccess = errors.New("non success status code")
+)
+
+// Middleware middleware which add transaction for the http request with default config
+func Middleware(skippers ...middleware.Skipper) func(http.Handler) http.Handler {
```
reasonerjt: looks like config related stuff is package internal?  
Even the skippers I don't think you need to expose it to the caller to make them configurable.

Also the Acceptor -> Config -> DefaultConfig is a bit tedious, I don't see in near future there's need to customize that.

src/core/middlewares/middlewares_test.go
```diff
@@ -0,0 +1,43 @@
...
+func Test_fetchBlobAPISkipper(t *testing.T) {
+	type args struct {
+		r *http.Request
+	}
+	tests := []struct {
+		name string
+		args args
+		want bool
+	}{
+		{"fetch blob", args{httptest.NewRequest(http.MethodGet, "/v2/library/photon/blobs/sha256:6e0447537050cf871f9ab6a3fec5715f9c6fff5212f6666993f1fc46b1f717a3", nil)}, true},
```
bitsf: should also add a case that url does not match blob url

src/server/middleware/transaction/transaction.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	return middleware.New(func(w http.ResponseWriter, r *http.Request, next http.Handler) {
+		res := internal.NewResponseBuffer(w)
+		defer res.Flush()
+
+		h := func(ctx context.Context) error {
+			next.ServeHTTP(res, r)
+
+			if !config.Acceptor(res.StatusCode()) {
```
reasonerjt: There is res.Success()

src/server/middleware/transaction/transaction.go
```diff
@@ -0,0 +1,84 @@
...
+				return errNonSuccess
+			}
+
+			return nil
+		}
+
+		if err := orm.WithTransaction(h)(r.Context()); err != nil && err != errNonSuccess {
```
bitsf: it seems err only can be nil or errNonSuccess

heww: err may be a db transaction error

reasonerjt:
@heww As we discussed, adding tx to all requests seems problematic.  
There maybe case the DB SHOULD be updated even the response is 4xx
I'll mark it request change .

src/core/middlewares/middlewares_test.go
heww: "get manifest" case

src/core/middlewares/middlewares.go
heww: The path in that pkg (`xxxxx/auth`) is not publiced, and it is not suitable to use a pkg named `auth`   here.

But it's a good point the keep the regex the same with it, it will change it.

reasonerjt: I also made some change so the middleware layer can share the same set of Regular expressions...

src/server/middleware/transaction/transaction.go
heww: Acceptor in the Config of the middleware need the status code, so that can user of the middleware can define their own logic to  decide which status code is successful.

heww:
> @heww As we discussed, adding tx to all requests seems problematic.
> There maybe case the DB SHOULD be updated even the response is 4xx
> I'll mark it request change .

Now there are two ways to deal this problem:
1. The Acceptor in the Config now provide `http.Request` so that Acceptor can decide which status code is acceptable one request by request.
2. The transaction pkg provide a `MustCommit(ctx context.Context)` function which can mark ctx from `http.Request` as committed so that transaction middleware commit transaction and ignore the status code of the response.

Issue #10553, opened 2020-01-21
Unify the method/style to handle error in handler/middleware


ywk253100:
Currently, we have different ways to write the error to the response:
https://github.com/goharbor/harbor/blob/master/src/server/registry/error/error.go#L24
https://github.com/goharbor/harbor/blob/master/src/server/v2.0/handler/base.go#L34
https://github.com/goharbor/harbor/blob/master/src/server/middleware/readonly/readonly.go#L72
https://github.com/goharbor/harbor/blob/master/src/core/api/project.go#L104
Let's unify them into one style

PR #10561, opened 2020-01-22
Implement the get/delete handler for registry API

ywk253100:
Implement the get/delete handler for registry API

### DIFF
--- a/src/api/artifact/abstractor/blob/fetcher.go
+++ b/src/api/artifact/abstractor/blob/fetcher.go
@@ -18,9 +18,14 @@ import (
 	"github.com/docker/distribution/manifest/manifestlist"
 	"github.com/docker/distribution/manifest/schema1"
 	"github.com/docker/distribution/manifest/schema2"
+	"github.com/goharbor/harbor/src/common/utils/registry"
+	"github.com/goharbor/harbor/src/common/utils/registry/auth"
+	"github.com/goharbor/harbor/src/core/config"
+	"github.com/goharbor/harbor/src/core/service/token"
 	coreutils "github.com/goharbor/harbor/src/core/utils"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"io/ioutil"
+	"net/http"
 )
 
 var (
@@ -54,7 +59,7 @@ type fetcher struct{}
 // TODO re-implement it based on OCI registry driver
 func (f *fetcher) FetchManifest(repository, digest string) (string, []byte, error) {
 	// TODO read from cache first
-	client, err := coreutils.NewRepositoryClientForLocal("admin", repository)
+	client, err := newRepositoryClient(repository)
 	if err != nil {
 		return "", nil, err
 	}
@@ -75,4 +80,20 @@ func (f *fetcher) FetchLayer(repository, digest string) ([]byte, error) {
 	}
 	defer reader.Close()
 	return ioutil.ReadAll(reader)
+}
+
+func newRepositoryClient(repository string) (*registry.Repository, error) {
+	uam := &auth.UserAgentModifier{
+		UserAgent: "harbor-registry-client",
+	}
+	authorizer := auth.NewRawTokenAuthorizer("admin", token.Registry)
+	transport := registry.NewTransport(http.DefaultTransport, authorizer, uam)
+	client := &http.Client{
+		Transport: transport,
+	}
+	endpoint, err := config.RegistryURL()
+	if err != nil {
+		return nil, err
+	}
+	return registry.NewRepository(repository, endpoint, client)
 }
--- a/src/api/artifact/controller.go
+++ b/src/api/artifact/controller.go
@@ -18,7 +18,8 @@ import (
 	"context"
 	"fmt"
 	"github.com/goharbor/harbor/src/api/artifact/abstractor"
-	// register image resolvers
+	"github.com/opencontainers/go-digest"
+	// registry image resolvers
 	_ "github.com/goharbor/harbor/src/api/artifact/abstractor/resolver/image"
 	// register chart resolver
 	_ "github.com/goharbor/harbor/src/api/artifact/abstractor/resolver/chart"
@@ -48,6 +49,9 @@ type Controller interface {
 	List(ctx context.Context, query *q.Query, option *Option) (total int64, artifacts []*Artifact, err error)
 	// Get the artifact specified by ID, specify the properties returned with option
 	Get(ctx context.Context, id int64, option *Option) (artifact *Artifact, err error)
+	// Get the artifact specified by repository name and reference, the reference can be tag or digest,
+	// specify the properties returned with option
+	GetByReference(ctx context.Context, repository, reference string, option *Option) (artifact *Artifact, err error)
 	// Delete the artifact specified by ID. All tags attached to the artifact are deleted as well
 	Delete(ctx context.Context, id int64) (err error)
 	// Tags returns the tags according to the query, specify the properties returned with option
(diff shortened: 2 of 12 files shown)


ywk253100:
The UT case for registry handlers will be covered in a separate PR

src/api/artifact/abstractor/blob/fetcher.go
```diff
@@ -76,3 +81,19 @@ func (f *fetcher) FetchLayer(repository, digest string) ([]byte, error) {
 	defer reader.Close()
 	return ioutil.ReadAll(reader)
 }
+
+func newRepositoryClient(repository string) (*registry.Repository, error) {
```
reasonerjt: why?

src/pkg/artifact/manager.go
```diff
@@ -99,24 +96,6 @@ func (m *manager) Create(ctx context.Context, artifact *Artifact) (int64, error)
...
-	})
-	if err != nil {
-		return err
-	}
-	if len(references) > 0 {
-		var ids []string
-		for _, reference := range references {
-			ids = append(ids, strconv.FormatInt(reference.ParentID, 10))
-		}
-		return ierror.PreconditionFailedError(nil).
```
reasonerjt: How do you make sure 412 is returned?

src/server/registry/handler.go
```diff
@@ -32,6 +27,9 @@ import (
 	"github.com/goharbor/harbor/src/server/registry/manifest"
 	"github.com/goharbor/harbor/src/server/registry/tag"
 	"github.com/gorilla/mux"
+	"net/http"
```
reasonerjt: This will be removed later right?

src/server/registry/manifest/manifest.go
```diff
@@ -17,37 +17,35 @@ package manifest
...
+	artCtl  artifact.Controller
+	proxy   *httputil.ReverseProxy
 }
 
 func (h *handler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	switch req.Method {
-	case http.MethodHead:
-		h.head(w, req)
-	case http.MethodGet:
+	case http.MethodHead, http.MethodGet:
```
reasonerjt: Does it comply with the spec when you write the response body in HEAD?

src/api/artifact/abstractor/blob/fetcher.go
ywk253100: Change is that connect the docker registry directly rather than via core.
When pushing a new image, we need to fetch the manifest to abstract the data

src/pkg/artifact/manager.go
ywk253100: This is covered by the foreign key constraint, 412 will returned when got such error

src/server/registry/handler.go
ywk253100: yes, all handers will be bound to beego router direclty

src/server/registry/manifest/manifest.go
ywk253100: Will not write any data to the response body in HEAD as the request will be proxied to the backed docker registry. What the `get` method done is just check the existence of the artifact in database

src/api/artifact/abstractor/blob/fetcher.go
reasonerjt: So why not make it a shared func?  Or we need to do it after switching to basic auth?

src/api/artifact/abstractor/blob/fetcher.go
```diff
@@ -54,7 +59,7 @@ type fetcher struct{}
 // TODO re-implement it based on OCI registry driver
 func (f *fetcher) FetchManifest(repository, digest string) (string, []byte, error) {
```
steven-zou: From the signature of this method, is it possible to return the Manifest object instead of bytes?

src/api/artifact/controller.go
```diff
@@ -48,6 +49,9 @@ type Controller interface {
 	List(ctx context.Context, query *q.Query, option *Option) (total int64, artifacts []*Artifact, err error)
 	// Get the artifact specified by ID, specify the properties returned with option
 	Get(ctx context.Context, id int64, option *Option) (artifact *Artifact, err error)
+	// Get the artifact specified by repository name and reference, the reference can be tag or digest,
+	// specify the properties returned with option
+	GetByReference(ctx context.Context, repository, reference string, option *Option) (artifact *Artifact, err error)
```
steven-zou: WHY add this method? What's the use case?

src/api/repository/controller.go
```diff
@@ -17,38 +17,42 @@ package repository
...
 // Controller defines the operations related with repositories
 type Controller interface {
-	// Ensure the repository specified by the "name" exists under the project,
-	// creates it if it doesn't exist. The "name" should contain the namespace part.
-	// The "created" will be set as true when the repository is created
-	Ensure(ctx context.Context, projectID int64, name string) (created bool, id int64, err error)
+	// Ensure the repository specified by the "name" exists, creates it if it doesn't exist.
+	// The "name" should contain the namespace part. The "created" will be set as true
+	// when the repository is created
+	Ensure(ctx context.Context, name string) (created bool, id int64, err error)
```
steven-zou: Why do we need to know if it is a newly created?

src/server/registry/handler.go
```diff
@@ -53,9 +51,9 @@ func New(url *url.URL) http.Handler {
 	// handle manifest
 	// TODO maybe we should split it into several sub routers based on the method
 	manifestRouter := rootRouter.Path("/v2/{name:.*}/manifests/{reference}").Subrouter()
-	manifestRouter.NewRoute().Methods(http.MethodGet).Handler(middleware.WithMiddlewares(manifest.NewHandler(project.Mgr, proxy), manifestinfo.Middleware(), regtoken.Middleware()))
-	manifestRouter.NewRoute().Methods(http.MethodHead).Handler(manifest.NewHandler(project.Mgr, proxy))
-	manifestRouter.NewRoute().Methods(http.MethodDelete).Handler(middleware.WithMiddlewares(manifest.NewHandler(project.Mgr, proxy), readonly.Middleware(), manifestinfo.Middleware(), immutable.MiddlewareDelete()))
+	manifestRouter.NewRoute().Methods(http.MethodGet).Handler(middleware.WithMiddlewares(manifest.NewHandler(proxy), manifestinfo.Middleware(), regtoken.Middleware()))
+	manifestRouter.NewRoute().Methods(http.MethodHead).Handler(manifest.NewHandler(proxy))
```
steven-zou: The current methods are not clear, we'll refine them later, right?

src/server/registry/manifest/manifest.go
```diff
@@ -56,42 +54,59 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
-
 // make sure the artifact exist before proxying the request to the backend registry
 func (h *handler) get(w http.ResponseWriter, req *http.Request) {
-	// TODO check the existence
+	// check the existence in the database first
+	vars := mux.Vars(req)
+	reference := vars["reference"]
+	artifact, err := h.artCtl.GetByReference(req.Context(), vars["name"], reference, nil)
+	if err != nil {
+		error.Handle(w, req, err)
```
steven-zou: This method can be wrapped as the field method of the handler to avoid passing so many params

src/api/artifact/abstractor/blob/fetcher.go
ywk253100: There are three concrete Manifest objects at least: manifest version 1, manifest version2 and manifest list. The caller unmarshal the bytes into their own manifest object

src/api/artifact/controller.go
ywk253100: The name rather than ID is used in both Harbor management APIs and OCI registry APIs, such as:
Listing artifact API: `GET /api/v2.0/projects/{project_name}/repositories/{repository_name}/artifacts/{reference}`
Get manifest API: `GET /v2/{repository}/manifests/{reference}`
Provide this method can help implement such APIs easier.

Besides, the middlewares(immutable, content trust, vulnerability, etc.) can use this method directly to get the artifact information.

Without this method, every implementation should call:
1. RepositoryManager.List(repository_name) to get repository ID
2. TagManager.List(repository_name, tag_name) to get the artifact ID that the tag attached to
3. ArtifactManager.Get(artifact_id) to get the artifact information

src/api/repository/controller.go
ywk253100: As the `Ensure` may create the artifact or not, return the `created` can distinguish such cases, but it isn't used yet in our use cases

ywk253100: It is there already and isn't introduced by this PR

src/server/registry/handler.go
ywk253100: Yes, another PR will be submit to remove the mux and refine the routes

src/server/registry/manifest/manifest.go
ywk253100: You mean something like that `func (h *handler) sendError()`? But the params `w`, `req` and `err` are all not the fields of handler

Anyway, we'll refine the way to handle error in all handlers, issue https://github.com/goharbor/harbor/issues/10553 is opened to track this

src/api/artifact/abstractor/blob/fetcher.go
steven-zou: Not sure

src/api/repository/controller.go
steven-zou: What I mean is it seems we do NOT need to distinguish newly created and existing ones. That  parameter is unnecessary.

src/server/registry/manifest/manifest.go
steven-zou: Should have such method in the structure who has rw and r

PR #10571, opened 2020-01-24
Add middlewares for permission checking for v2 API

reasonerjt:
When the registry shifts from token auth to basic auth, we'll use the middleware to check permission.
This commit add middlewares for populate the artifact info and check
permission based on request to /v2/* api via security context

The change in `/src/core/xxx` can be disregarded because in the end all middlewares will be moved to `/src/server/middleware`

### DIFF
--- a/src/server/middleware/artifactinfo/artifact_info.go
+++ b/src/server/middleware/artifactinfo/artifact_info.go
@@ -1,0 +1,124 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package artifactinfo
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/url"
+	"regexp"
+	"strings"
+
+	"github.com/goharbor/harbor/src/common/utils/log"
+	ierror "github.com/goharbor/harbor/src/internal/error"
+	"github.com/goharbor/harbor/src/server/middleware"
+	reg_err "github.com/goharbor/harbor/src/server/registry/error"
+	"github.com/opencontainers/go-digest"
+)
+
+const (
+	blobMountQuery  = "mount"
+	blobFromQuery   = "from"
+	blobMountDigest = "blob_mount_digest"
+	blobMountRepo   = "blob_mount_repo"
+)
+
+var (
+	urlPatterns = map[string]*regexp.Regexp{
+		"manifest":    middleware.V2ManifestURLRe,
+		"tag_list":    middleware.V2TagListURLRe,
+		"blob_upload": middleware.V2BlobUploadURLRe,
+		"blob":        middleware.V2BlobURLRe,
+	}
+)
+
+// Middleware gets the information of artifact via url of the request and inject it into the context
+func Middleware() func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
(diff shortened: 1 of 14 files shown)


src/server/middleware/artifactinfo/artifact_info.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+var (
+	urlPatterns = []*regexp.Regexp{
+		middleware.V2ManifestURLRe, middleware.V2TagListURLRe, middleware.V2BlobUploadURLRe, middleware.V2BlobURLRe,
+	}
+)
+
+// Middleware gets the information of artifact via url of the request and inject it into the context
+func Middleware() func(http.Handler) http.Handler {
```
ywk253100: This middleware can be merged into the manifest info middleware https://github.com/goharbor/harbor/blob/master/src/server/middleware/manifestinfo/manifest_info.go?

src/server/middleware/artifactinfo/artifact_info.go
```diff
@@ -0,0 +1,86 @@
...
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+
+			path := req.URL.Path
+			log.Debugf("in artifact info middleware, path: %s", path)
+			m, ok := parse(path)
+			if !ok {
+				next.ServeHTTP(rw, req)
```
ywk253100: return?

src/server/middleware/artifactinfo/artifact_info_test.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/goharbor/harbor/src/core/middlewares/util"
+	"github.com/goharbor/harbor/src/server/middleware"
+	"github.com/stretchr/testify/assert"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"testing"
+)
+
+func TestMain(m *testing.M) {
```
ywk253100: Why we need the `TestMain` here?

src/server/middleware/v2authz/authz.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/goharbor/harbor/src/common/utils/log"
+	"github.com/goharbor/harbor/src/core/config"
+	"github.com/goharbor/harbor/src/core/filter"
+	"github.com/goharbor/harbor/src/core/promgr"
+	ierror "github.com/goharbor/harbor/src/internal/error"
+	"github.com/goharbor/harbor/src/server/middleware"
+	reg_err "github.com/goharbor/harbor/src/server/registry/error"
+	"net/http"
+)
+
```
ywk253100: We should handle the request of blob mount

src/server/middleware/artifactinfo/artifact_info.go
reasonerjt: Yes,  but they are a little different, I opened  #10574 and will sync with Yan to merge them.

src/server/middleware/v2authz/authz.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return err
+	}
+	if a, ok := middleware.ArtifactInfoFromContext(req.Context()); ok {
+		action := getAction(req)
+		if action == "" {
+			return nil
+		}
+		log.Debugf("action: %s, repository: %s", action, a.Repository)
+		p, err := rc.pm.Get(a.ProjectName)
```
heww: project may not found here

reasonerjt: If it's not found it will return an error

reasonerjt: Thanks I was thinking the pm will return an error if project is not found.

src/server/middleware/v2authz/authz.go
reasonerjt: Yep, thanks.

src/server/middleware/artifactinfo/artifact_info_test.go
reasonerjt: We don't.  I'll remove.

heww:
Is there another PR to apply these middlewares to v2?

reasonerjt:
Yes, there will be.

PR #10581, opened 2020-01-28
Add query string to listing artifact API to support specify only show the tagged artifacts

ywk253100:
Specify the query string "tags=NOT_NULL" to the listing artifact API to only return the tagged artifacts

### DIFF
--- a/api/v2.0/swagger.yaml
+++ b/api/v2.0/swagger.yaml
@@ -34,6 +34,11 @@ paths:
           description: Query the artifacts by type. Valid values can be "IMAGE", "CHART", etc.
           type: string
           required: false
+        - name: tags
+          in: query
+          description: Without the "tags" query string, both tagged and untagged artifacts will be returned. Specify it as "*" to return only tagged artifacts; specify it as "nil" to return only untagged artifacts.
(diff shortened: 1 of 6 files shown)


src/pkg/artifact/dao/dao.go
```diff
@@ -181,3 +197,27 @@ func (d *dao) DeleteReferences(ctx context.Context, parentID int64) error {
 	_, err = qs.Delete()
 	return err
 }
+
+func querySetter(ctx context.Context, query *q.Query) (beegoorm.QuerySeter, error) {
```
wy65701436: generally, it's a little tricky to use "tags" to judege return all or tagged.

api/v2.0/swagger.yaml
```diff
@@ -110,6 +110,12 @@ paths:
           description: Query the artifacts by type. Valid values can be "IMAGE", "CHART", etc.
           type: string
           required: false
+        - name: tags
```
steven-zou: Is it better to treat tags/untags as sub-resources of artifacts? Provide new API endpoints to handle such request?

PR #10587, opened 2020-01-29
Register registry handler with the new methods of Route

ywk253100:
Register registry handler with the new methods of Route


src/core/config/config.go
```diff
@@ -217,7 +217,11 @@ func SelfRegistration() (bool, error) {
 
 // RegistryURL ...
 func RegistryURL() (string, error) {
-	return cfgMgr.Get(common.RegistryURL).GetString(), nil
+	url := os.Getenv("REGISTRY_URL")
```
wy65701436: what's the purpose to modify it?

ywk253100: The registry URL cannot be read before the database started if use the config manager. This isn't needed

src/server/registry/catalog.go
```diff
@@ -109,12 +100,12 @@ func (h *handler) get(w http.ResponseWriter, req *http.Request) {
...
 	}
 
-	h.sendResponse(w, req, resRepos)
+	r.sendResponse(w, req, resRepos)
 	return
 }
 
 // sendResponse ...
-func (h *handler) sendResponse(w http.ResponseWriter, req *http.Request, repositoryNames []string) {
+func (r *repositoryHandler) sendResponse(w http.ResponseWriter, req *http.Request, repositoryNames []string) {
```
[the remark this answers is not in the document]
ywk253100: Currently it is not, we'll introduce a common method later to writer JSON to response directly

PR #10588, opened 2020-01-29
Implement artifact/tag related API

ywk253100:
Implement APIs:
1. Get artifact
2. Delete artifact
3. Create tag
4. Delete tag

### DIFF
--- a/api/v2.0/swagger.yaml
+++ b/api/v2.0/swagger.yaml
@@ -18,80 +18,6 @@ securityDefinitions:
 security:
   - basicAuth: []
 paths:
-  /projects/{project_name}/repositories/{repository_name}/artifacts/{digest}:
-    get:
-      summary: Get the specific artifact
-      description: Get the artifact specified by the digest under the project and repository
-      tags:
-        - artifact
-      operationId: getArtifact
-      parameters:
-        - $ref: '#/parameters/requestId'
-        - $ref: '#/parameters/projectName'
-        - $ref: '#/parameters/repositoryName'
-        - $ref: '#/parameters/digest'
-        - name: with_tag
-          in: query
-          description: Specify whether the tags are inclued inside the returning artifacts
-          type: boolean
-          required: false
-          default: true
-        - name: with_label
-          in: query
-          description: Specify whether the labels are inclued inside the returning artifacts
-          type: boolean
-          required: false
-          default: false
-        - name: with_scan_overview
-          in: query
-          description: Specify whether the scan overview is inclued inside the returning artifacts
-          type: boolean
-          required: false
-          default: false
-        # should be in tag level
-        - name: with_signatrue
-          in: query
-          description: Specify whether the signature is inclued inside the returning artifacts
-          type: boolean
-          required: false
-          default: false
-        - name: with_immutable_status
-          in: query
-          description: Specify whether the immutable status is inclued inside the tags of the returning artifacts. Only works when setting "with_tag=true"
-          type: boolean
-          required: false
-          default: false
-      responses:
-        '200':
-          description: Success
-          schema:
-            $ref: '#/definitions/Artifact'
-        '401':
-          $ref: '#/responses/401'
-        '403':
-          $ref: '#/responses/403'
-        '500':
-          $ref: '#/responses/500'
-    delete:
-      summary: Delete the specific artifact
-      description: Delete the artifact specified by the digest under the project and repository
-      tags:
-        - artifact
-      operationId: deleteArtifact
-      parameters:
-        - $ref: '#/parameters/requestId'
-        - $ref: '#/parameters/projectName'
-        - $ref: '#/parameters/repositoryName'
-        - $ref: '#/parameters/digest'
-      responses:
-        '200':
-          description: Success
-        '401':
-          $ref: '#/responses/401'
-        '403':
-          $ref: '#/responses/403'
-        '500':
-          $ref: '#/responses/500'
   /projects/{project_name}/repositories/{repository_name}/artifacts:
     get:
       summary: List artifacts
@@ -129,7 +55,7 @@ paths:
           required: false
           default: false
         # should be in tag level
-        - name: with_signatrue
+        - name: with_signature
           in: query
           description: Specify whether the signature is inclued inside the returning artifacts
           type: boolean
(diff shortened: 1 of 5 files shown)


api/v2.0/swagger.yaml
```diff
@@ -141,21 +131,80 @@ paths:
...
+        - $ref: '#/parameters/repositoryName'
+        - $ref: '#/parameters/reference'
+        - name: tag
+          in: body
+          description: The JSON object of tag.
+          required: true
+          schema:
+            $ref: '#/definitions/Tag'
+      responses:
+        '201':
```
wy65701436: no 400?

src/api/artifact/controller.go
```diff
@@ -56,6 +56,8 @@ type Controller interface {
 	Delete(ctx context.Context, id int64) (err error)
 	// Tags returns the tags according to the query, specify the properties returned with option
 	Tags(ctx context.Context, query *q.Query, option *TagOption) (total int64, tags []*Tag, err error)
+	// CreateTag creates a tag
+	CreateTag(ctx context.Context, tag *Tag) (id int64, err error)
```
wy65701436: to use "attach" and "detach" instead will be easy to understand?

src/server/v2.0/handler/artifact.go
```diff
@@ -103,11 +85,98 @@ func (a *artifactAPI) ListArtifacts(ctx context.Context, params operation.ListAr
...
+	}
+	tag := &artifact.Tag{}
+	tag.RepositoryID = art.RepositoryID
+	tag.ArtifactID = art.ID
+	tag.Name = params.Tag.Name
+	tag.PushTime = time.Now()
+	if _, err = a.artCtl.CreateTag(ctx, tag); err != nil {
+		return a.SendError(ctx, err)
+	}
+	// TODO set location header?
```
wy65701436: I think we need the URI in the header.

ywk253100: In the TODO list, will set it in a separate PR

src/api/artifact/controller.go
ywk253100: The `ArtifactID` is one property of the `Tag` object, when creating a new tag, it is attached to the artifact automatically. So `Create` and `Delete` is more reasonable

api/v2.0/swagger.yaml
ywk253100: Added

src/server/v2.0/handler/artifact.go
```diff
@@ -103,11 +85,98 @@ func (a *artifactAPI) ListArtifacts(ctx context.Context, params operation.ListAr
...
+	}
+	return operation.NewGetArtifactOK().WithPayload(artifact.ToSwagger())
 }
 
 func (a *artifactAPI) DeleteArtifact(ctx context.Context, params operation.DeleteArtifactParams) middleware.Responder {
-	// TODO implement
+	artifact, err := a.artCtl.GetByReference(ctx, fmt.Sprintf("%s/%s", params.ProjectName, params.RepositoryName), params.Reference, nil)
+	if err != nil {
+		return a.SendError(ctx, err)
+	}
```
wy65701436: what about to remove the query and call the delete directly?

ywk253100: Deletion is only supported with ID for now.

src/api/artifact/controller.go
```diff
@@ -56,6 +56,8 @@ type Controller interface {
...
 	// DeleteTag deletes the tag specified by tagID
 	DeleteTag(ctx context.Context, tagID int64) (err error)
```
steven-zou: Is this method really delete the tags or just `remove` the tags from the artifact?

src/server/v2.0/handler/artifact.go
steven-zou: Why not let the caller provide the ID? That means the caller can call some `get`/`list` method to have the related ID and then trigger deletion with that ID.

src/server/v2.0/handler/artifact.go
```diff
@@ -103,11 +85,98 @@ func (a *artifactAPI) ListArtifacts(ctx context.Context, params operation.ListAr
...
+	tag.PushTime = time.Now()
+	if _, err = a.artCtl.CreateTag(ctx, tag); err != nil {
+		return a.SendError(ctx, err)
+	}
+	// TODO set location header?
+	return operation.NewCreateTagCreated()
+}
+
+func (a *artifactAPI) DeleteTag(ctx context.Context, params operation.DeleteTagParams) middleware.Responder {
+	artifact, err := a.artCtl.GetByReference(ctx, fmt.Sprintf("%s/%s", params.ProjectName, params.RepositoryName),
```
steven-zou: So many methods depend on this utility method `GetByReference`, we need to check if there is something not well designed.

steven-zou: So many methods depend on this utility method `GetByReference`, we need to check if there is something not well designed.

src/server/v2.0/handler/artifact.go
```diff
@@ -103,11 +85,98 @@ func (a *artifactAPI) ListArtifacts(ctx context.Context, params operation.ListAr
...
+			"tag %s attached to artifact %d not found", params.TagName, artifact.ID)
+		return a.SendError(ctx, err)
+	}
+	if err = a.artCtl.DeleteTag(ctx, id); err != nil {
+		return a.SendError(ctx, err)
+	}
+	return operation.NewDeleteTagOK()
+}
+
+func option(withTag, withImmutableStatus, withLabel, withScanOverview, withSignature *bool) *artifact.Option {
```
steven-zou: the option looks like a little messy

src/api/artifact/controller.go
ywk253100: It really deletes the tag

src/server/v2.0/handler/artifact.go
ywk253100: This is in the API handler already, there is no upper caller

src/api/artifact/controller.go
```diff
@@ -56,6 +56,8 @@ type Controller interface {
 	Delete(ctx context.Context, id int64) (err error)
 	// Tags returns the tags according to the query, specify the properties returned with option
 	Tags(ctx context.Context, query *q.Query, option *TagOption) (total int64, tags []*Tag, err error)
```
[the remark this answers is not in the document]
heww: `ListTags` will be better.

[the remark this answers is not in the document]
ywk253100: Done

src/server/v2.0/handler/artifact.go
ywk253100: If we have different opinions for the method, could we open an issue and discuss later, and let's the PR move forward to avoid blocking other works?

Issue #10604, opened 2020-01-31
Revisit errors from registry after switching to basic auth.


reasonerjt:
Some test cases had to be updated as the docker client prints different error messages.  
Though the correct status code and response body is returned, we should double check how docker daemon handles errors to avoid unexpected breakage.

reasonerjt:
After enabling the test cases and manually verified against different CLIs. so far everything looks good.  
I will open new issues if new problems emerge.

PR #10605, opened 2020-01-31
Switch to basic authentication for registry

reasonerjt:
1. Add basic authorizer for registry which modify the request
to add basic authorization header to request based on configuration.
2. Set basic auth header for proxy when accessing registry
3. Switche the registry to use basic auth by default and use the basic
authorizer to access Harbor.
4. Make necessary change to test cases, particularly
"test_robot_account.py" and "docker_api.py", because the error is
changed after siwtched to basic auth from token auth.  #10604 is opened
to track the follow up work.

### DIFF
--- a/src/server/middleware/v2auth/auth.go
+++ b/src/server/middleware/v2auth/auth.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package authz
+package v2auth
 
 import (
 	"fmt"
@@ -24,7 +24,9 @@ import (
 	ierror "github.com/goharbor/harbor/src/internal/error"
 	"github.com/goharbor/harbor/src/server/middleware"
 	reg_err "github.com/goharbor/harbor/src/server/registry/error"
+	"golang.org/x/net/context"
 	"net/http"
+	"sync"
 )
 
 type reqChecker struct {
@@ -32,6 +34,10 @@ type reqChecker struct {
 }
 
 func (rc *reqChecker) check(req *http.Request) error {
+	if rc.hasRegistryCred(req) {
+		// TODO: May consider implement a local authorizer for registry, more details see #10602
+		return nil
+	}
 	securityCtx, err := filter.GetSecurityContext(req)
 	if err != nil {
 		return err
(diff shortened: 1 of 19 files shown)


src/server/middleware/v2auth/auth.go
```diff
@@ -62,6 +68,9 @@ func (rc *reqChecker) check(req *http.Request) error {
 		}
 	} else if len(middleware.V2CatalogURLRe.FindStringSubmatch(req.URL.Path)) == 1 && !securityCtx.IsSysAdmin() {
 		return fmt.Errorf("unauthorized to list catalog")
+	} else if req.URL.Path == "/v2/" && !securityCtx.IsAuthenticated() {
+		ctx := context.WithValue(req.Context(), middleware.SkipInjectRegistryCredKey, true)
+		*req = *(req.WithContext(ctx))
```
ywk253100: We should return 401 and the supported auth method here or in our own handler.

reasonerjt: Good point, but IMO we may defer this, we can always make sure the registry setting is consistent with the auth setting as the communication between registry and docker or other container runtime is very well tested.

Opened #10610 but it's likely to be deferred to future.

src/server/middleware/v2auth/auth.go
```diff
@@ -98,16 +113,24 @@ func getAction(req *http.Request) rbac.Action {
...
 
 // Middleware checks the permission of the request to access the artifact
 func Middleware() func(http.Handler) http.Handler {
+	once.Do(func() {
+		if checker.pm == nil { // for UT, where pm has been set to a mock value
+			checker = reqChecker{
+				pm: config.GlobalProjectMgr,
+			}
+		}
+	})
```
heww: Changing the value of the `checker` at the beginning of the UT doesn't work ? If it doesn't work, the right place of `once.Do` is  in `check` method of `reqChecker` not in the middleware.

reasonerjt: The problem is that config.GlobalProjectMgr is nil when checker is assigned.  So when this middleware was added to the handler I found it didn't work.

Issue #10598, opened 2020-01-30
Make sure replication works as expected after registry is switched to basic auth


reasonerjt:
Currently the harbor adapter depends on native adapter which is shared by other adapters.

Some changes will be needed to make sure replication still works.

reasonerjt:
Should be fixed in #10612 I'll verify and close it.

Issue #8211, opened 2019-07-03
Unable to add a registry endpoint when provider is docker-registry


bonniechen1119:
1.We have build a docker private registry, such as https://docker-registry.miotech.com:5000,
and docker login docker-registry.miotech.com:5000 successfully
2.However when we add new Registry Endpoint and select Provider as docker-registry
3.When Endpoint URL set as https://docker-registry.miotech.com:5000 and input Access ID and Access Secret, it meet error: registry https://docker-registry.miotech.com:5000 is unhealthy: unhealthy
4. We check the log of our private registry, it say:level=warning msg="error authorizing context: basic authentication challenge for realm "Registry Realm": invalid authorization credential"
5..When Endpoint URL set as docker-registry.miotech.com:5000 and input Access ID and Access Secret, it also meet error: registry https://docker-registry.miotech.com:5000 is unhealthy: unhealthy 
6. And check  the log of our private registry again, it say:http: TLS handshake error from 18.162.133.113:34860: tls: first record does not look like a TLS handshake

bonniechen1119:
Harbor version is 1.8

ywk253100:
What is the auth provider for your docker registry? https://github.com/docker/distribution/blob/master/docs/configuration.md#auth

bonniechen1119:
> What is the auth provider for your docker registry? https://github.com/docker/distribution/blob/master/docs/configuration.md#auth

We use htpasswd

willmao:
same issue with htpasswd auth mode set, seems currently harbor only support token auth mode when replication

kofj:
> same issue with htpasswd auth mode set, seems currently harbor only support token auth mode when replication

Yeah, it not support now. We will support others auth mode in the future.

mumrau:
What is the ETA of such mechanism, from my understanding it doesn't look that complex, and I believe it could be quite disturbing to new users that just want to pull their entire "old" registry which probably uses htpasswd mechanism (which was (is?) the de-facto docker registry self-hosted auth mechanism).

joekohut:
Bump for getting this added ASAP. This is a huge roadblock as stated by mumrau.

inyee786:
I am also facing similar kind of issue but there is one difference in my docker registry configuration 
I am using Nginx as a `basic auth` provider and `htpasswd` to generate the password for my Nginx configuration

If I put the empty `Access ID` and `Access Secret` the test is `passed`, but if put the credentials it `Failed` the ping test

core log>
```
[...]

PR #10612, opened 2020-02-02
Do enhancement for the registry authorizer

ywk253100:
This commit introduces a new wrapper authorizer which can authorize the request according to the auth scheme automatically

### DIFF
--- a/src/common/utils/registry/auth/authorizer.go
+++ b/src/common/utils/registry/auth/authorizer.go
@@ -1,0 +1,123 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package auth
+
+import (
+	"fmt"
+	"github.com/docker/distribution/registry/client/auth/challenge"
+	"github.com/goharbor/harbor/src/common/http/modifier"
+	"net/http"
+	"net/url"
+	"strings"
+	"sync"
+)
+
+// NewAuthorizer creates an authorizer that can handle different auth schemes
+func NewAuthorizer(credential Credential, client ...*http.Client) modifier.Modifier {
+	authorizer := &authorizer{
+		credential: credential,
+	}
+	if len(client) > 0 {
+		authorizer.client = client[0]
+	}
+	if authorizer.client == nil {
+		authorizer.client = http.DefaultClient
+	}
+
+	return authorizer
+}
+
+// authorizer authorizes the request with the provided credential.
+// It determines the auth scheme of registry automatically and calls
+// different underlying authorizers to do the auth work
+type authorizer struct {
+	sync.Mutex
+	client     *http.Client
+	url        *url.URL          // registry URL
+	authorizer modifier.Modifier // the underlying authorizer
+	credential Credential
+}
+
+func (a *authorizer) Modify(req *http.Request) error {
+	// Nil URL means this is the first time the authorizer is called
+	// Try to connect to the registry and determine the auth method
+	if a.url == nil {
+		// to avoid concurrent issue
+		a.Lock()
+		defer a.Unlock()
+		if err := a.initialize(req.URL); err != nil {
+			return err
+		}
+	}
+
+	// check whether the request targets the registry
+	if !a.isTarget(req) {
+		return nil
+	}
+
+	return a.authorizer.Modify(req)
+}
+
+func (a *authorizer) initialize(u *url.URL) error {
+	if a.url != nil {
+		return nil
+	}
+	url, err := url.Parse(u.Scheme + "://" + u.Host + "/v2/")
+	if err != nil {
+		return err
+	}
+	a.url = url
+	resp, err := a.client.Get(a.url.String())
+	if err != nil {
+		return err
+	}
+	challenges := ParseChallengeFromResponse(resp)
+	// no challenge, mean no auth
+	if len(challenges) == 0 {
+		a.authorizer = &nullAuthorizer{}
+		return nil
+	}
+	cm := map[string]challenge.Challenge{}
+	for _, challenge := range challenges {
+		cm[challenge.Scheme] = challenge
+	}
+	if _, exist := cm["basic"]; exist {
+		a.authorizer = a.credential
+		return nil
+	}
+
+	if _, exist := cm["bearer"]; exist {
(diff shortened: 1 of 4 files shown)


ywk253100:
Fixes https://github.com/goharbor/harbor/issues/10598 and https://github.com/goharbor/harbor/issues/8211

src/common/utils/registry/auth/authorizer.go
```diff
@@ -0,0 +1,122 @@
...
+	"fmt"
+	"github.com/docker/distribution/registry/client/auth/challenge"
+	"github.com/goharbor/harbor/src/common/http/modifier"
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+// NewAuthorizer creates an authorizer that can handle different auth schemes
+func NewAuthorizer(credential Credential, client ...*http.Client) modifier.Modifier {
```
reasonerjt: 1. Can we have a better name?
2. The attribute "url" is a little tricky, why not just set the registry endpoint when creating this authorizer?

reasonerjt:
You should also re-enable the test case.

src/common/utils/registry/auth/authorizer.go
ywk253100: For 2, we handle the URL for a long time with this method and it works well. 
Besides this is the way how docker handles such case

ywk253100:
> Could you re-enable the test case `test_add_replication_rule.py` ?

Done

src/go.mod
```diff
@@ -76,6 +76,7 @@ require (
 	github.com/stretchr/testify v1.4.0
 	github.com/theupdateframework/notary v0.6.1
 	golang.org/x/crypto v0.0.0-20190617133340-57b3e21c3d56
+	golang.org/x/net v0.0.0-20190827160401-ba9fcec4b297
```
bitsf: 这个增加了新包后，怎么vendor目录中没有变化呢

src/common/utils/registry/auth/authorizer.go
```diff
@@ -0,0 +1,122 @@
...
+		return err
+	}
+	challenges := ParseChallengeFromResponse(resp)
+	// no challenge, mean no auth
+	if len(challenges) == 0 {
+		a.authorizer = &nullAuthorizer{}
+		return nil
+	}
+	cm := map[string]challenge.Challenge{}
+	for _, challenge := range challenges {
```
steven-zou: Seem you can directly put two `if` statements here and don't need to construct a map as only one of `basic`/`bearer` existing.

src/common/utils/registry/auth/authorizer.go
```diff
@@ -0,0 +1,122 @@
...
+// Check whether the request targets to the registry.
+// If doesn't, the request shouldn't be handled by the authorizer.
+// e.g. the requests sent to backend storage(s3, etc.)
+func (a *authorizer) isTarget(req *http.Request) bool {
+	index := strings.Index(req.URL.Path, "/v2/")
+	if index == -1 {
+		return false
+	}
+	if req.URL.Host != a.url.Host || req.URL.Scheme != a.url.Scheme ||
+		req.URL.Path[:index+4] != "/v2/" {
```
steven-zou: `index := strings.Index(req.URL.Path, "/v2/")` !=? `req.URL.Path[:index+4] != "/v2/"`?

src/common/utils/registry/auth/authorizer.go
```diff
@@ -0,0 +1,122 @@
...
+	if len(challenges) == 0 {
+		a.authorizer = &nullAuthorizer{}
+		return nil
+	}
+	cm := map[string]challenge.Challenge{}
+	for _, challenge := range challenges {
+		cm[challenge.Scheme] = challenge
+	}
+	if _, exist := cm["basic"]; exist {
+		a.authorizer = a.credential
```
steven-zou: No underlying basic authorizer definition? We should follow a similar authorizer interface.

src/common/utils/registry/auth/authorizer.go
```diff
@@ -0,0 +1,122 @@
...
+		return false
+	}
+	if req.URL.Host != a.url.Host || req.URL.Scheme != a.url.Scheme ||
+		req.URL.Path[:index+4] != "/v2/" {
+		return false
+	}
+	return true
+}
+
+type nullAuthorizer struct{}
```
steven-zou: Move to a separate file?

steven-zou: ../auth/baisc
../auth/null
../auth/token
???

src/common/utils/registry/auth/authorizer.go
ywk253100: Don't get your point

src/common/utils/registry/auth/authorizer.go
ywk253100: With map, we can determine the sequence of authorizers that apply to

src/common/utils/registry/auth/authorizer.go
ywk253100: This is tricky. The `credential` is an interface, it can be either basic auth authorier or secret authorizer. The secret is used in the replication  job to call registry APIs.

src/common/utils/registry/auth/authorizer.go
reasonerjt: Could you make sure this authorizer is not for concurrent use?  Explicitly document it in the comment?

ywk253100: Add lock to avoid the concurrent issue

src/common/utils/registry/auth/authorizer.go
steven-zou: I mean `index := strings.Index(req.URL.Path, "/v2/")` is duplicated check with the below statement `req.URL.Path[:index+4]`, right?

src/common/utils/registry/auth/authorizer.go
steven-zou: `credential ` implements the authorizer interface? It's a little weird, why not new basic/secret authorized with passing the credential as inputs of `New()`？

src/common/utils/registry/auth/authorizer.go
steven-zou: What I mean is the logic `index := strings.Index(req.URL.Path, "/v2/")` doing the same check with the statement `req.URL.Path[:index+4] != "/v2/"`, right?

src/common/utils/registry/auth/authorizer.go
steven-zou: As only one authorizer can be enabled, so what I mean is 

```go
for _, challenge := range challenges {
    if challenge.Scheme == "basic" {
         a.authorizer = a.credential
         return nil
    }
    if challenge.Scheme =="Bearer"  {
         a.authorizer = NewStandardTokenAuthorizer(a.client, a.credential)
        return nil
   }
}
```

The sequence is guaranteed by the if statement. Your map just behaviors as a hash key for if checking.

src/common/utils/registry/auth/authorizer.go
ywk253100: Opened issue to track the work https://github.com/goharbor/harbor/issues/10705

PR #10617, opened 2020-02-03
refactor(security): add `NewContext` and `FromContext` to security pkg

heww:
1. Add `NewContext` and `FromContext` funcs in security pkg.
2. Add `Name` func in `security.Context` interface to make the checking
for the `/api/internal/configurations` API clear.
3. Get the security from the context to prepare change the security
filter to middleware.
4. Remove `GetSecurityContext` in filter pkg.

### DIFF
--- a/src/common/security/context.go
+++ b/src/common/security/context.go
@@ -15,12 +15,16 @@
 package security
 
 import (
+	"context"
+
 	"github.com/goharbor/harbor/src/common/models"
 	"github.com/goharbor/harbor/src/common/rbac"
 )
 
 // Context abstracts the operations related with authN and authZ
 type Context interface {
+	// Name returns the name of the security context
+	Name() string
 	// IsAuthenticated returns whether the context has been authenticated or not
 	IsAuthenticated() bool
 	// GetUsername returns the username of user related to the context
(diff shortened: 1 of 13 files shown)


src/common/security/context.go
```diff
@@ -36,3 +40,16 @@ type Context interface {
 	// Can returns whether the user can do action on resource
 	Can(action rbac.Action, resource rbac.Resource) bool
 }
+
+type securityKey struct{}
+
+// NewContext returns context with security context
+func NewContext(ctx context.Context, security Context) context.Context {
+	return context.WithValue(ctx, securityKey{}, security)
```
bitsf: 这个securityKey 用法挺奇怪，现在是流行这么写，而不用 string 了吗

src/common/security/local/context.go
```diff
@@ -38,6 +38,11 @@ func NewSecurityContext(user *models.User, pm promgr.ProjectManager) *SecurityCo
 	}
 }
 
+// Name returns the name of the security context
+func (s *SecurityContext) Name() string {
+	return "robot"
```
bitsf: 这个Name 是不是应该叫 local

src/common/security/context.go
ninjadq: @bitsf  empty struct will not allocate space in memory

@heww But I think no matter which style is used we should only use one style in our code base.

heww: https://github.com/golang/go/issues/17826#issuecomment-258946985

Issue #10573, opened 2020-01-26
Return the consistent error format in swagger handlers


ywk253100:
Got 401 error if no basic auth header provided. The auth checker can be removed as this is handled by the handler logic:
```shell
curl -i http://127.0.0.1/api/v2.0/projects/library/repositories/hello-world/artifacts
HTTP/1.1 401 Unauthorized
Server: nginx
Date: Sun, 26 Jan 2020 02:08:24 GMT
Content-Type: application/json
Content-Length: 94
Connection: keep-alive
Set-Cookie: sid=1fea807c48c72f39532228cd4af95b39; Path=/; HttpOnly
Www-Authenticate: Basic realm="API"
X-Request-Id: 9539103c-48a3-466f-9482-a4674cc45e12

unauthenticated for invalid credentials, UNAUTHORIZED, unauthenticated for invalid credentials
```

Got 422 error when the query string format doesn't meet the constraint. The returned error message should be consistent with others(currently is the JSON string of an error array ):
```
yinw-a01:ansible yinw$ curl -i -u a:a http://127.0.0.1/api/v2.0/projects/library/repositories/hello-world/artifacts?tags=1
HTTP/1.1 422 Unprocessable Entity
Server: nginx
Date: Sun, 26 Jan 2020 02:49:52 GMT
Content-Type: application/json
Content-Length: 106
Connection: keep-alive
Set-Cookie: sid=36ffcbe3335bcc0fec51c47988142a42; Path=/; HttpOnly
X-Request-Id: b39ea78b-e2b3-4885-9dd8-c66e5772e8da

tags in query should be one of [NOT_NULL], UNPROCESSABLE_ENTITY, tags in query should be one of [NOT_NULL]
```

Issue #10546, opened 2020-01-20
United error response format for management APIs (legacy and v2.0 APIs)


heww:
Use the OCI APIs error format or management APIs error format ?

OCI APIs error format
```json
{
    "errors": [
        { "code": "UNKNOW", "message": "Blob unknow" }
    ]
}
```

Managment APIs error format
```json
{
    "code": "BAD_REQUEST", 
    "message": "Name must be unique" 
}
```

ywk253100:
We should estimate the effort to update the legacy APIs to use the error array format. If it is huge effort, I suggest use the single error format.

Besides, the legacy API uses the error with integer error code, we should update it as string

ywk253100:
Fixed by #11070

PR #10626, opened 2020-02-04
Unify the method/style to handle error in handler/middleware

ywk253100:
This commit provides a "SendError" method to unify the way to handle error in handlers/middlewares


ywk253100:
Fixes https://github.com/goharbor/harbor/issues/10553

heww:
This make the management APIs return the same error format with OCI spec.  @steven-zou 

And if this PR merged, swagger.yaml and ServeError in handler for v2.0 also need to update to match the error format.

ywk253100:
Fixes https://github.com/goharbor/harbor/issues/10573

ywk253100:
@heww Let's use the issue https://github.com/goharbor/harbor/issues/10546 to discuss the error format. It's easy to change between these two formats as there is only one place to update.

PR #10628, opened 2020-02-04
set the immutable status on getting/listting tag

wy65701436:

### DIFF
--- a/src/api/artifact/controller.go
+++ b/src/api/artifact/controller.go
@@ -18,6 +18,10 @@ import (
 	"context"
 	"fmt"
 	"github.com/goharbor/harbor/src/api/artifact/abstractor"
+	"github.com/goharbor/harbor/src/common/utils"
+	"github.com/goharbor/harbor/src/pkg/art"
+	"github.com/goharbor/harbor/src/pkg/immutabletag/match"
+	"github.com/goharbor/harbor/src/pkg/immutabletag/match/rule"
 	"github.com/opencontainers/go-digest"
 	// registry image resolvers
 	_ "github.com/goharbor/harbor/src/api/artifact/abstractor/resolver/image"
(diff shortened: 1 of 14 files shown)


src/api/artifact/model.go
```diff
@@ -88,7 +88,8 @@ func (a *Artifact) ToSwagger() *models.Artifact {
 // Tag is the overall view of tag
 type Tag struct {
 	tag.Tag
-	// TODO add other attrs: signature, label, immutable status, etc
+	Immutable bool
```
ywk253100: Add the property in swagger and handle it in the upper function `func (a *Artifact) ToSwagger() *models.Artifact`

src/api/artifact/controller.go
```diff
@@ -374,8 +377,28 @@ func (c *controller) assembleTag(ctx context.Context, tag *tm.Tag, option *TagOp
 		return t
 	}
 	if option.WithImmutableStatus {
-		// TODO populate immutable status
+		repo, err := c.repoMgr.Get(ctx, tag.RepositoryID)
+		if err != nil {
+			log.Error(err)
```
ywk253100: Set the immutable to true when get erorr in case it is deleted by mistake?

src/api/artifact/controller.go
```diff
@@ -374,8 +377,28 @@ func (c *controller) assembleTag(ctx context.Context, tag *tm.Tag, option *TagOp
...
+func isImmutable(projectID int64, repo string, tag string) bool {
+	_, repoName := utils.ParseRepository(repo)
+	matched, err := rule.NewRuleMatcher(projectID).Match(art.Candidate{
+		Repository:  repoName,
+		Tag:         tag,
+		NamespaceID: projectID,
+	})
+	if err != nil {
+		log.Error(err)
+		return false
```
ywk253100: Same comment here

src/api/artifact/model.go
wy65701436: done.

src/api/artifact/controller.go
wy65701436: it's the original design, do we need to change?

PR #10650, opened 2020-02-06
fix(feat) fix prepare version in installer

### LINKED ISSUE #10569: The VERSIONTAG in prepare script is not set in install.sh
Currently in the latest build of master branch.

If I run install.sh the tags of `docker-compose.yml` is set to `dev`

seems due to this line:
https://github.com/goharbor/harbor/blob/master/make/prepare#L62

I have to export VERSIONTAG before running install.sh
```
export VERSIONTAG=v1.11.0-build.2854
```

In previous versions this is not necessary.

bitsf:
fix #10569 
The VERSIONTAG in prepare script is not set in install.sh, so revert to old code, fix this when package the installer

### DIFF
--- a/make/prepare
+++ b/make/prepare
@@ -32,8 +32,6 @@ fi
 
 data_path=$(grep '^[^#]*data_volume:' $input_dir/harbor.yml | awk '{print $NF}')
 
-echo "VERSIONTAG:${VERSIONTAG:=dev}"
-
 # If previous secretkeys exist, move it to new location
 previous_secretkey_path=/data/secretkey
 previous_defaultalias_path=/data/defaultalias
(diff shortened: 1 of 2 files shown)


wy65701436:
please give the details in the commit message.

make/prepare
```diff
@@ -59,7 +57,7 @@ docker run --rm -v $input_dir:/input:z \
                     -v $config_dir:/config:z \
                     -v $secret_dir:/secret:z \
                     -v /:/hostfs:z \
-                    goharbor/prepare:${VERSIONTAG} $@
+                    goharbor/prepare:dev $@
```
wy65701436: we do need the `${VERSIONTAG}` here, as we will release prepare image with the release builds. @ninjadq right?

ninjadq: It changed the dev tag at line `341`. the regex pattern like we used in the past.
but why we need change it back?

bitsf: 因为 VERSIONTAG 从来没有传进来，所以导致安装时找不到对应的image
而且如果要改成传进来的话，改的地方有点多了，还不如先用原来的方法直接在 package 时替换掉

PR #10652, opened 2020-02-06
Use controller rather than manager in the API handler and middleware

wy65701436:

### DIFF
--- a/src/server/middleware/immutable/deletemf.go
+++ b/src/server/middleware/immutable/deletemf.go
@@ -3,14 +3,9 @@ package immutable
 import (
 	"errors"
 	"fmt"
+	"github.com/goharbor/harbor/src/api/artifact"
 	common_util "github.com/goharbor/harbor/src/common/utils"
 	internal_errors "github.com/goharbor/harbor/src/internal/error"
-	"github.com/goharbor/harbor/src/pkg/art"
-	"github.com/goharbor/harbor/src/pkg/artifact"
-	"github.com/goharbor/harbor/src/pkg/immutabletag/match/rule"
-	"github.com/goharbor/harbor/src/pkg/q"
-	"github.com/goharbor/harbor/src/pkg/repository"
-	"github.com/goharbor/harbor/src/pkg/tag"
 	"github.com/goharbor/harbor/src/server/middleware"
 	"net/http"
 )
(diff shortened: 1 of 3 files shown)


src/server/middleware/immutable/deletemf.go
```diff
@@ -43,46 +40,21 @@ func handleDelete(req *http.Request) error {
...
 		return err
 	}
-	if total == 0 {
+	if af == nil {
 		return nil
 	}
 
-	for _, tag := range tags {
+	_, repoName := common_util.ParseRepository(mf.Repository)
+	for _, tag := range af.Tags {
```
ywk253100: Set with immutable as true, the immutable of tags status will be returned back. So we don't need to do the match again

src/server/middleware/immutable/pushmf.go
```diff
@@ -59,51 +56,21 @@ func handlePush(req *http.Request) error {
...
-			continue
-		}
-		for _, tag := range tags {
-			// push a existing immutable tag, reject the request
-			if tag.Name == mf.Tag {
-				return NewErrImmutable(repoName, mf.Tag)
-			}
+	for _, tag := range af.Tags {
+		// push a existing immutable tag, reject the request
+		if tag.Name == mf.Tag {
```
ywk253100: Same here. We can get the artifact with immutable set as true first and check whether the pushing tag already exists and is immutable. 
This can reduce the duplicate code

src/server/middleware/immutable/deletemf.go
```diff
@@ -43,46 +40,21 @@ func handleDelete(req *http.Request) error {
...
+		WithTag: true,
 	})
 	if err != nil {
+		if internal_errors.IsErr(err, internal_errors.NotFoundCode) {
+			return nil
+		}
 		return err
 	}
-	if total == 0 {
+	if af == nil {
```
heww: `af` will never be `nil` return by `artifact.Ctl.GetByReference` when `err` is not `nil`.

src/server/middleware/immutable/pushmf.go
```diff
@@ -59,51 +56,21 @@ func handlePush(req *http.Request) error {
...
-			"RepositoryID": repos[0].RepositoryID,
-		},
+	af, err := artifact.Ctl.GetByReference(req.Context(), mf.Repository, mf.Tag, &artifact.Option{
+		WithTag: true,
 	})
 	if err != nil {
 		return err
 	}
-	if total == 0 {
+	if af == nil {
```
heww: `af` will never be `nil` return by `artifact.Ctl.GetByReference` when `err` is not `nil`.

src/server/middleware/util.go
```diff
@@ -70,50 +67,16 @@ type ManifestInfo struct {
...
-				"RepositoryID": repos[0].RepositoryID,
-			},
-		})
+		af, err := artifact.Ctl.GetByReference(ctx, info.Repository, info.Tag, nil)
 		if err != nil {
 			info.manifestExistErr = err
 			return
 		}
-		if total == 0 {
+		if af == nil {
```
heww: `af` will never be `nil` return by `artifact.Ctl.GetByReference` when `err` is not `nil`.

src/server/middleware/immutable/deletemf.go
wy65701436: ok, removed.

src/server/middleware/immutable/pushmf.go
wy65701436: ditto

src/server/middleware/util.go
wy65701436: ditto

src/server/middleware/immutable/deletemf.go
wy65701436: updated.

PR #10656, opened 2020-02-06
Implement get addition API for image

ywk253100:
This commit implements the API to get build history of image with manifest version 2 and populates the addition links when listing/getting the artifact

### DIFF
--- a/api/v2.0/swagger.yaml
+++ b/api/v2.0/swagger.yaml
@@ -249,6 +249,39 @@ paths:
           $ref: '#/responses/404'
         '500':
           $ref: '#/responses/500'
+  /projects/{project_name}/repositories/{repository_name}/artifacts/{reference}/additions/{addition}:
+    get:
+      summary: Get the addition of the specific artifact
+      description: Get the addition of the artifact specified by the reference under the project and repository.
+      tags:
+        - artifact
+      operationId: getAddition
+      parameters:
+        - $ref: '#/parameters/requestId'
+        - $ref: '#/parameters/projectName'
+        - $ref: '#/parameters/repositoryName'
+        - $ref: '#/parameters/reference'
+        - name: addition
+          in: path
+          description: The addition name, "build_history" for images; "values.yaml", "readme", "dependencies" for charts
+          type: string
+          enum: [build_history, values.yaml, readme, dependencies]
+          required: true
+      responses:
+        '200':
+          description: Success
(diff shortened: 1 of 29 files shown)


api/v2.0/swagger.yaml
```diff
@@ -229,6 +241,39 @@ paths:
...
+      parameters:
+        - $ref: '#/parameters/requestId'
+        - $ref: '#/parameters/projectName'
+        - $ref: '#/parameters/repositoryName'
+        - $ref: '#/parameters/reference'
+        - name: addition
+          in: path
+          description: The addition name, "BUILD_HISTORY" for images; "VALUES.YAML", "README", "DEPENDENCIES" for charts
+          type: string
+          enum: [BUILD_HISTORY, VALUES.YAML, README, DEPENDENCIES]
```
steven-zou: These UPPER words are in the URL? Maybe lower cases are better and consistent with the other path segments.

Backend const values can be different from the query parameters.

src/api/artifact/abstractor/abstractor.go
```diff
@@ -40,6 +42,12 @@ type Abstractor interface {
 	// Abstract the specific information for the specific artifact type into the artifact model,
 	// the information can be got from the manifest or other layers referenced by the manifest.
 	Abstract(ctx context.Context, artifact *artifact.Artifact) error
+	// GetAddition returns the addition of the artifact.
+	// The additions are different for different artifacts:
+	// build history for image; values.yaml, readme and dependencies for chart, etc
+	GetAddition(ctx context.Context, artifact *artifact.Artifact, additionType string) (addition *resolver.Addition, err error)
```
steven-zou: Why addition related operations are defined in abstractor, not in the artifact(mgr)?

steven-zou: The abstractor extracts the additional contents to the artifact and the related getting/listing OPs should be from artifact side.

src/api/artifact/abstractor/abstractor.go
```diff
@@ -40,6 +42,12 @@ type Abstractor interface {
...
+	// ListSupportedAdditions lists the supported additions of the specific kind of artifact
+	ListSupportedAdditions(ctx context.Context, mediaType string) (additions []string)
```
steven-zou: `ListAdditions` is enough

PR #10674, opened 2020-02-07
fix(middleware): escape and clean path for the skipper

heww:
1. Escape and clean request path for `legacyAPISkipper`.
2. Escape and clean request path for `MethodAndPathSkipper`.

### DIFF
--- a/src/core/middlewares/middlewares.go
+++ b/src/core/middlewares/middlewares.go
@@ -16,6 +16,7 @@ package middlewares
 
 import (
 	"net/http"
+	"path"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


src/core/middlewares/middlewares.go
```diff
@@ -37,8 +38,9 @@ var (
 
 // legacyAPISkipper skip middleware for legacy APIs
 func legacyAPISkipper(r *http.Request) bool {
+	path := path.Clean(r.URL.EscapedPath())
```
bitsf: 变量名就别用path了吧，用 p 呗

heww: I think var name of `path` is ok here.

PR #10682, opened 2020-02-10
Support add/remove label to/from artifact

ywk253100:
This commit add supporting for adding/removing label to/from artifacts and populates labels when listing artifacts

### DIFF
--- a/src/api/artifact/controller.go
+++ b/src/api/artifact/controller.go
@@ -26,6 +26,7 @@ import (
 	"github.com/goharbor/harbor/src/pkg/art"
 	"github.com/goharbor/harbor/src/pkg/immutabletag/match"
 	"github.com/goharbor/harbor/src/pkg/immutabletag/match/rule"
+	"github.com/goharbor/harbor/src/pkg/label"
 	"github.com/opencontainers/go-digest"
 	"strings"
 
@@ -77,6 +78,10 @@ type Controller interface {
 	// The addition is different according to the artifact type:
 	// build history for image; values.yaml, readme and dependencies for chart, etc
 	GetAddition(ctx context.Context, artifactID int64, additionType string) (addition *resolver.Addition, err error)
+	// AddLabel to the specified artifact
+	AddLabel(ctx context.Context, artifactID int64, labelID int64) (err error)
+	// RemoveLabel from the specified artifact
+	RemoveLabel(ctx context.Context, artifactID int64, labelID int64) (err error)
 	// TODO move this to GC controller?
 	// Prune removes the useless artifact records. The underlying registry data will
 	// be removed during garbage collection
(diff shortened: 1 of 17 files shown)


src/api/artifact/controller.go
```diff
@@ -407,30 +429,46 @@ func (c *controller) assembleTag(ctx context.Context, tag *tm.Tag, option *TagOp
...
 	if err != nil {
 		log.Error(err)
-		return false
+		return
 	}
-	return matched
+	tag.Immutable = matched
+}
+
+func (c *controller) populateScanOverview(ctx context.Context, art *Artifact) {
```
steven-zou: Is it possible to define a standard populate interface with kinds of implementations for such cases?

```
func (c *controller) populateScanOverview(ctx context.Context, art *Artifact) {
	// TODO implement
}

func (c *controller) populateSignature(ctx context.Context, art *Artifact) {
	// TODO implement
	// TODO populate signature on artifact or tag level?
}	}
```

src/api/artifact/model.go
```diff
@@ -82,14 +84,27 @@ func (a *Artifact) ToSwagger() *models.Artifact {
 			Href:     link.HREF,
 		}
 	}
+	for _, label := range a.Labels {
+		art.Labels = append(art.Labels, &models.Label{
```
steven-zou: Should the labels in the artifact be a full one (with all properties)?

src/pkg/label/manager.go
```diff
@@ -0,0 +1,94 @@
...
+	ierror "github.com/goharbor/harbor/src/internal/error"
+	"github.com/goharbor/harbor/src/pkg/q"
+	"time"
+)
+
+// Mgr is a global instance of label manager
+var Mgr = New()
+
+// Manager manages the labels and references between label and resource
+type Manager interface {
```
steven-zou: label's CRUD and reference management should be two parts, right? 
label mgr for label CURD
labelMgr.ReferenceMgr() returns the ref mgr that is responsible for reference management?

src/api/artifact/model.go
ywk253100: Yes, almost all of the properties is needed when rendering the UI

src/api/artifact/controller.go
ywk253100: Issue https://github.com/goharbor/harbor/issues/10741 opened to track the work

src/pkg/label/manager.go
ywk253100: I'll submit another PR to cover this part for both label manager and artifact manager

src/api/artifact/model.go
steven-zou: that's weird

PR #10706, opened 2020-02-11
Enable tls on all components

ninjadq:
Enable tls on all components


make/photon/prepare/templates/docker_compose/docker-compose.yml.jinja
```diff
@@ -4,7 +4,7 @@ services:
     image: goharbor/harbor-log:{{version}}
     container_name: harbor-log
     restart: always
-    dns_search: .
+    dns_search: ""
```
stonezdj: from the doc 
```
Use --dns-search=. if you don't wish to set the search domain.
```
what is the meaning change  . to "" ?

make/photon/prepare/utils/configs.py
```diff
@@ -104,12 +106,12 @@ def parse_yaml_config(config_file_path, with_notary, with_clair, with_chartmuseu
         configs = yaml.load(f)
 
     config_dict = {
-        'adminserver_url': "http://adminserver:8080",
-        'registry_url': "http://registry:5000",
-        'registry_controller_url': "http://registryctl:8080",
-        'core_url': "http://core:8080",
-        'core_local_url': "http://127.0.0.1:8080",
-        'token_service_url': "http://core:8080/service/token",
+        'adminserver_url': 'http://adminserver:8080',
```
stonezdj: the adminserver_url should be remoted because there is no adminserver.

src/jobservice/api/server.go
```diff
@@ -73,20 +74,11 @@ func NewServer(ctx context.Context, router Router, cfg ServerConfig) *Server {
...
-			CipherSuites: []uint16{
-				tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
-				tls.TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA,
-				tls.TLS_RSA_WITH_AES_256_GCM_SHA384,
-				tls.TLS_RSA_WITH_AES_256_CBC_SHA,
-			},
+		logger.Infof("https enabled, load trustCAs")
+		srv.TLSConfig = &tls.Config{
+			ClientAuth: tls.RequireAndVerifyClientCert,
+			ClientCAs:  commonhttp.GetInternalCA(nil),
```
stonezdj: Does the default tls.Config cover these previous cipherSuites?

ninjadq: The default value is different from the value we need.

ninjadq:
> Is there any utility tools that can generate all certificate if user doesn't provide it?

Yes, uploaded now https://github.com/goharbor/harbor/pull/10706/files#diff-c67bc9654beda26e39c1e635a532adc1

Makefile
```diff
@@ -85,6 +85,7 @@ NPM_REGISTRY=https://registry.npmjs.org
 # enable/disable chart repo supporting
 CHARTFLAG=false
 BUILDTARGET=build
+GEN_TLS=
```
reasonerjt: If in prepare it's called `gencert` why in Makefile it's called `gentls`?

make/harbor.yml.tmpl
```diff
@@ -17,6 +17,8 @@ https:
   certificate: /your/certificate/path
   private_key: /your/private/key/path
 
+# internal_tls: /etc/harbor/tls/internal
```
reasonerjt: comment how to put the certs?

make/photon/prepare/commands/gencerts.py
```diff
@@ -0,0 +1,26 @@
...
+from subprocess import check_call, PIPE, STDOUT
+
+from utils.cert import openssl_installed
+from utils.misc import get_realpath
+
+gen_tls_script = pathlib.Path(__file__).parent.parent.joinpath('scripts/gencert.sh').absolute()
+
+@click.command()
+@click.argument('path')
+def gencert(path):
```
reasonerjt: This is a little trivial but can we provide help to each sub-command?

make/photon/prepare/commands/gencerts.py
```diff
@@ -0,0 +1,26 @@
...
+        raise(Exception('openssl not installed'))
+
+    click.echo("start generate internal tls certs")
+    if not os.path.exists(path):
+        click.echo('path {} not exist, create it...'.format(path))
+        os.makedirs(path, exist_ok=True)
+
+    shell_stat = check_call([gen_tls_script], stdout=PIPE, stderr=STDOUT, cwd=path)
+    if shell_stat != 0:
+        click.echo('Can not generate internal tls certs')
```
reasonerjt: If it fails, it should exit with non-zero?

make/photon/prepare/scripts/gencert.sh
```diff
@@ -0,0 +1,118 @@
+#! /bin/bash
+
+# CA key and certificate
+openssl req -x509 -nodes -days 365 -newkey rsa:4096 \
```
reasonerjt: It would be better if the script has flexibility to allow user to cert his CA.

But I think that can be another PR.

make/photon/prepare/templates/chartserver/env.jinja
```diff
@@ -1,5 +1,12 @@
 ## Settings should be set
+{%if internal_tls.enabled %}
+PORT=9443
+TLS_CERT=/etc/harbor/ssl/chartmuseum.crt
```
reasonerjt: Have you checked the helm chart?  If the paths to the cert and keys are hardcoded would it be easier for user to enable tls on chart?

Makefile
ninjadq: The whole name should be `generate_internal_tls_certificates`. So I think both are correct

make/photon/prepare/commands/gencerts.py
ninjadq: help will automatically generated by `click`

src/core/main.go
```diff
@@ -232,8 +234,23 @@ func main() {
 
 	server.RegisterRoutes()
 
-	log.Infof("Version: %s, Git commit: %s", version.ReleaseVersion, version.GitCommit)
+	iTLSEnabled := os.Getenv("INTERNAL_TLS_ENABLED")
```
reasonerjt: reuse the one in `commonhttp`?

src/pkg/registry/client.go
```diff
@@ -53,6 +54,13 @@ var (
 		schema2.MediaTypeManifest,
 		schema1.MediaTypeSignedManifest,
 	}
+
+	localRegistryURL = map[string]bool{
```
reasonerjt: I don't think we can use hard-code here.

src/registryctl/main.go
```diff
@@ -37,25 +38,15 @@ func (s *RegistryCtl) Start() {
...
-			MinVersion:               tls.VersionTLS12,
-			CurvePreferences:         []tls.CurveID{tls.CurveP521, tls.CurveP384, tls.CurveP256},
-			PreferServerCipherSuites: true,
-			CipherSuites: []uint16{
-				tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
-				tls.TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA,
-				tls.TLS_RSA_WITH_AES_256_GCM_SHA384,
-				tls.TLS_RSA_WITH_AES_256_CBC_SHA,
-			},
+	if s.ServerConf.Protocol == "https" {
```
reasonerjt: I think it's good practice to add supported cipher suite, why remove it?

src/common/http/client.go
```diff
@@ -28,41 +28,80 @@ import (
...
-	modifiers []modifier.Modifier
-	client    *http.Client
-}
+const (
+	// DefaultTransport used to get the default http Transport
+	DefaultTransport = iota
+	// InsecureTransport used to get the insecure http Transport
+	InsecureTransport
+	// InternalTransport used to get the internal secure http Transport
+	InternalTransport
```
reasonerjt: Would it be simpler and more consistent if we merge `internal` and `secure` transport into one `secure transport`?

This `secure transport` should trust the CA in the truststore OS as well as the Harbor internal CA if it's configured?

src/common/http/client.go
```diff
@@ -28,41 +28,80 @@ import (
...
+				InsecureSkipVerify: true,
+			},
+		}
 	}
-	if insecure[0] {
-		return insecureHTTPTransport
+}
+
+// GetHTTPTransport returns HttpTransport based on insecure configuration
+func GetHTTPTransport(clientType uint) *http.Transport {
```
reasonerjt: Please test in mass replication scenario, I remember  each time you create a new transport there will be a new connection ,and it will be pending at "TIME_WAIT" until you set some timeout values.

ninjadq: There is no new connection when you create a new Transport. The new connection only starts when you start a request with new Transport. In this situation. These changes will improve our concurrency in some bit. If it caused too many time_wait we can use a connection pool or config the kernel parameters. I think in many companies ops guy will config the kernel configuration to cut down the `TIME_WAIT` time.

src/pkg/registry/client.go
ninjadq: oh, you mean hardcode here. We can initialize this map from config.

src/registryctl/main.go
ninjadq: remove it to make it more general. Because if you communicate with another component like Nginx or other third-party services, the ciphers you both specify are different will cause a mismatch.
I think the best practice is to open a new PR to upgrade supported TLS on our components to version 1.3. It's faster and more secure.

stonezdj:
Could you please add some document to the install guide ? for example enable/disable internal TLS, rotate certificate etc.

make/photon/prepare/models.py
```diff
@@ -0,0 +1,135 @@
...
+        'clair_adapter.crt', 'clair_adapter.key',
+        'clair.crt', 'clair.key'
+    }
+
+    notary_certs_filename = {
+        'notary_signer.crt', 'notary_signer.key',
+        'notary_server.crt', 'notary_server.key'
+    }
+
+    chart_museum_filename = {
```
stonezdj: Because we are not going to enable TLS in chartmuseum, should remove it?

ninjadq: TLS still supported in the chart. just not enable mutual TLS in it.

src/core/main.go
```diff
@@ -161,8 +162,22 @@ func main() {
...
 	server.RegisterRoutes()
 
-	log.Infof("Version: %s, Git commit: %s", version.ReleaseVersion, version.GitCommit)
+	if common_http.InternalTLSEnabled() {
+		log.Info("internal TLS enabled, Init TLS ...")
+		iTLSKeyPath := os.Getenv("INTERNAL_TLS_KEY_PATH")
+		iTLSCertPath := os.Getenv("INTERNAL_TLS_CERT_PATH")
+
+		log.Infof("load client key: %s client cert: %s", iTLSKeyPath, iTLSCertPath)
+		// uncomment following if harbor2 is ready
```
reasonerjt: confusing comment

src/jobservice/hook/hook_client.go
```diff
@@ -54,6 +55,13 @@ func NewClient(ctx context.Context) Client {
 		ExpectContinueTimeout: 1 * time.Second,
 		Proxy:                 http.ProxyFromEnvironment,
 	}
+	if commonhttp.InternalTLSEnabled() {
+		tlsConfig, err := commonhttp.GetInternalTLSConfig()
+		if err != nil {
+			panic(err)
```
reasonerjt: panic???

ninjadq: yes, if you enabled tls but not provided cert file correctly. service should fail immediately.

src/common/http/tls.go
```diff
@@ -0,0 +1,96 @@
...
+	internalVerifyClientCert = "INTERNAL_VERIFY_CLIENT_CERT"
+	internalTLSKeyPath       = "INTERNAL_TLS_KEY_PATH"
+	internalTLSCertPath      = "INTERNAL_TLS_CERT_PATH"
+	internalTrustCAPath      = "INTERNAL_TLS_TRUST_CA_PATH"
+)
+
+// InternalTLSEnabled returns if internal TLS enabled
+func InternalTLSEnabled() bool {
+	iTLSEnabled := os.Getenv(internalTLSEnable)
+	if strings.ToLower(iTLSEnabled) == "true" {
```
reasonerjt: why not directly 
`return strings.ToLower(os.Getenv(xxx)) == "true" `

src/common/http/tls.go
```diff
@@ -0,0 +1,96 @@
...
+		return true
+	}
+	return false
+}
+
+// InternalEnableVerifyClientCert returns if mTLS enabled
+func InternalEnableVerifyClientCert() bool {
+	enabled := os.Getenv(internalVerifyClientCert)
+	if strings.ToLower(enabled) == "true" {
```
reasonerjt: why not directly 
`return strings.ToLower(os.Getenv(xxx)) == "true" `

src/core/main.go
ninjadq: just a note. because if we can enable mTLS in for if it's ready.

src/jobservice/hook/hook_client.go
reasonerjt: but it's the client, is it called when the service bootstraps?

src/core/main.go
reasonerjt: I think we can remove it for now.  Otherwise please update the comment to clarify.  I don't think `if harbor2 is ready` is a clear criteria...

ninjadq: ok let's remove them

reasonerjt:
For the go code part:

- The way to generate tls config need refinement.
- The secure transport doesn't seem correct.

For the prepare part"

- Can we allow user to generate certs using his own CA?
- Can you double check the custom CA can be applied along with the internal CA?

src/common/http/tls.go
```diff
@@ -0,0 +1,94 @@
...
+// InternalEnableVerifyClientCert returns if mTLS enabled
+func InternalEnableVerifyClientCert() bool {
+	if strings.ToLower(os.Getenv(internalVerifyClientCert)) == "true" {
+		return true
+	}
+	return false
+}
+
+// GetInternalCA used to get internal cert file from Env
+func GetInternalCA(caPool *x509.CertPool) *x509.CertPool {
```
reasonerjt: let's remove this.

src/jobservice/api/server.go
```diff
@@ -72,21 +73,11 @@ func NewServer(ctx context.Context, router Router, cfg ServerConfig) *Server {
 	}
 
 	// Initialize TLS/SSL config if protocol is https
-	if cfg.Protocol == config.JobServiceProtocolHTTPS {
```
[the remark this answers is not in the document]
reasonerjt: Let's open an issue to upgrade 1.3

src/replication/adapter/harbor/adapter.go
```diff
@@ -296,6 +303,9 @@ func (a *adapter) getProject(name string) (*project, error) {
 // when harbor is deployed on Kubernetes
 func (a *adapter) getURL() string {
 	if a.registry.Type == model.RegistryTypeHarbor && a.registry.Name == "Local" {
+		if common_http.InternalTLSEnabled() {
+			return "https://core:10443"
```
reasonerjt: Let's make sure it's 8443

src/replication/adapter/awsecr/adapter.go
```diff
@@ -245,11 +245,18 @@ func (a *adapter) createRepository(repository string) error {
 	if a.region == "" {
 		return errors.New("no region parsed")
 	}
+	var tr *http.Transport
```
reasonerjt: let's extract a func

PR #10714, opened 2020-02-12
Fix bugs when pushing image(with index) and CNAB

ywk253100:
1. As "List" method of artifact DAO doesn't return the artifacts that referenced by other and without tag, so we introduce a new method "GetByDigest" to check the existence of artifact
2. The "Www-Authenticate" header is needed to be returned when the request is unauthorized. This is required in the OCI distribution spec and is needed when pushing CNAB

### DIFF
--- a/src/server/middleware/v2auth/auth.go
+++ b/src/server/middleware/v2auth/auth.go
@@ -15,7 +15,7 @@
 package v2auth
 
 import (
-	"context"
+	"errors"
 	"fmt"
 	serror "github.com/goharbor/harbor/src/server/error"
 	"net/http"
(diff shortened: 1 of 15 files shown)


src/server/middleware/v2auth/auth.go
```diff
@@ -70,8 +70,7 @@ func (rc *reqChecker) check(req *http.Request) error {
 	} else if len(middleware.V2CatalogURLRe.FindStringSubmatch(req.URL.Path)) == 1 && !securityCtx.IsSysAdmin() {
 		return fmt.Errorf("unauthorized to list catalog")
 	} else if req.URL.Path == "/v2/" && !securityCtx.IsAuthenticated() {
-		ctx := context.WithValue(req.Context(), middleware.SkipInjectRegistryCredKey, true)
```
reasonerjt: Could you explain why skip setting the basic auth header will drop the www-authenticate heander from response?

PR #10720, opened 2020-02-13
Get addition properties for chart

wy65701436:
1, Get readme.md content into addition
2, Get dependency of chart
3, Get values of chart


src/api/artifact/abstractor/resolver/chart/chart.go
```diff
@@ -86,7 +90,44 @@ func (r *resolver) ResolveMetadata(ctx context.Context, manifest []byte, artifac
 }
 
 func (r *resolver) ResolveAddition(ctx context.Context, artifact *artifact.Artifact, addition string) (*resolv.Addition, error) {
-	// TODO implement
+	if addition != AdditionTypeValues {
```
ywk253100: There are three additions for helm chart: `readme.md`, `values.yaml` and `dependencies`. We should return the separated one according to the request specified by `addition` parameter

src/api/artifact/abstractor/resolver/chart/chart.go
```diff
@@ -86,7 +90,61 @@ func (r *resolver) ResolveMetadata(ctx context.Context, manifest []byte, artifac
...
+			var additionContent []byte
+			var additionContentType string
+
+			switch addition {
+			case AdditionTypeValues:
+				additionContent = []byte(chartDetails.Files[AdditionTypeValues])
+				additionContentType = "text/plain; charset=utf-8"
+			case AdditionTypeReadme:
+				additionContent = []byte(chartDetails.Files[AdditionTypeReadme])
+				additionContentType = "text/plain; charset=utf-8"
```
ywk253100: Use `text/markdown` instead

src/api/artifact/abstractor/resolver/chart/chart.go
```diff
@@ -86,7 +90,61 @@ func (r *resolver) ResolveMetadata(ctx context.Context, manifest []byte, artifac
...
+				return nil, err
+			}
+
+			var additionContent []byte
+			var additionContentType string
+
+			switch addition {
+			case AdditionTypeValues:
+				additionContent = []byte(chartDetails.Files[AdditionTypeValues])
+				additionContentType = "text/plain; charset=utf-8"
```
ywk253100: Use `text/yaml`? There is no standard for yaml yet

src/api/artifact/abstractor/resolver/chart/chart.go
```diff
@@ -30,17 +32,18 @@ import (
 const (
 	// ArtifactTypeChart defines the artifact type for helm chart
 	ArtifactTypeChart        = "CHART"
-	AdditionTypeValues       = "VALUES.YAML"
-	AdditionTypeReadme       = "README"
+	AdditionTypeValues       = "values.yaml"
+	AdditionTypeReadme       = "README.md"
```
ywk253100: It's better to use Upper case for const definitions

heww: I think [pascal case](http://wiki.c2.com/?PascalCase) here is ok.

Here is the const definitions in `distribution`:

```go
const (
	// MediaTypeManifest specifies the mediaType for the current version. Note
	// that for schema version 1, the the media is optionally "application/json".
	MediaTypeManifest = "application/vnd.docker.distribution.manifest.v1+json"
	// MediaTypeSignedManifest specifies the mediatype for current SignedManifest version
	MediaTypeSignedManifest = "application/vnd.docker.distribution.manifest.v1+prettyjws"
	// MediaTypeManifestLayer specifies the media type for manifest layers
	MediaTypeManifestLayer = "application/vnd.docker.container.image.rootfs.diff+x-gtar"
)
```

How terrible when we use these values if they are in upper case?

ywk253100: I mean the enumerated const values. They are used as enumerated values.

@wy65701436 It doesn't work here. The addition types defined here are converted into lower case when populating the addition links: https://github.com/goharbor/harbor/blob/master/src/api/artifact/controller.go#L442.
And they are converted into upper case when calling the resolver.ResolveAddition: https://github.com/goharbor/harbor/blob/master/src/server/v2.0/handler/artifact.go#L186

heww: > I mean the enumerated const values. They are used as enumerated values.
> 

I got it.  +1

src/api/artifact/abstractor/resolver/chart/chart.go
wy65701436: resolved.

src/api/artifact/abstractor/resolver/chart/chart.go
wy65701436: let's keep it

src/api/artifact/abstractor/resolver/chart/chart.go
wy65701436: use the upper in const values.

src/api/artifact/abstractor/resolver/chart/chart.go
```diff
@@ -31,16 +33,17 @@ const (
 	// ArtifactTypeChart defines the artifact type for helm chart
 	ArtifactTypeChart        = "CHART"
 	AdditionTypeValues       = "VALUES.YAML"
-	AdditionTypeReadme       = "README"
+	AdditionTypeReadme       = "README.MD"
```
ywk253100: Please update the swagger file if change it to `README.MD`: https://github.com/goharbor/harbor/blob/master/api/v2.0/swagger.yaml#L275

PR #10727, opened 2020-02-14
Bump up legacy api version to v2.0

ywk253100:
This PR includes both frontend and backend changes


api/v2.0/legacy_swagger.yaml
```diff
@@ -2,12 +2,12 @@ swagger: '2.0'
 info:
   title: Harbor API
   description: These APIs provide services for manipulating Harbor project.
-  version: 1.10.0
+  version: '2.0'
```
zhoumeina: If this version number changed to 2. Where is the swagger for 1.0? Still in this file or move to somewhere?

ywk253100: There is no 1.0 anymore, all API version is v2.0 now

PR #10733, opened 2020-02-16
Implement copy artifact API

ywk253100:
Copy artifact into the repository from the specified artifact

### DIFF
--- a/src/api/artifact/controller.go
+++ b/src/api/artifact/controller.go
@@ -20,6 +20,7 @@ import (
 	"github.com/goharbor/harbor/src/api/artifact/abstractor"
 	"github.com/goharbor/harbor/src/api/artifact/abstractor/resolver"
 	"github.com/goharbor/harbor/src/api/artifact/descriptor"
+	"github.com/goharbor/harbor/src/common/models"
 	"github.com/goharbor/harbor/src/common/utils"
 	"github.com/goharbor/harbor/src/internal"
 	"github.com/goharbor/harbor/src/pkg/art"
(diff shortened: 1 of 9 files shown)


src/api/artifact/controller.go
```diff
@@ -64,6 +65,8 @@ type Controller interface {
 	GetByReference(ctx context.Context, repository, reference string, option *Option) (artifact *Artifact, err error)
 	// Delete the artifact specified by ID. All tags attached to the artifact are deleted as well
 	Delete(ctx context.Context, id int64) (err error)
+	// Copy the artifact whose ID is specified by "from" into the repository
+	Copy(ctx context.Context, repositoryID, from int64) (id int64, err error)
```
steven-zou: from(int64) is better than from(*Artifact)?

`from(int64)` will also need to call `get()` internally

src/server/v2.0/handler/artifact.go
```diff
@@ -36,15 +38,61 @@ func newArtifactAPI() *artifactAPI {
...
-	repoMgr repository.Manager
+	repoCtl repository.Controller
+}
+
+// TODO immutable, quota, readonly middlewares should cover this API
+func (a *artifactAPI) CopyArtifact(ctx context.Context, params operation.CopyArtifactParams) middleware.Responder {
+	if err := a.RequireProjectAccess(ctx, params.ProjectName, rbac.ActionCreate, rbac.ResourceArtifact); err != nil {
+		return a.SendError(ctx, err)
+	}
+	srcRepo, srcRef, err := parse(params.From)
```
reasonerjt: The API allows copy from artifact in other project, is there any security check?

src/api/artifact/controller.go
```diff
@@ -65,6 +66,8 @@ type Controller interface {
...
```
reasonerjt: why put src behind dst?

reasonerjt:
The `Copy` takes two int64 parm the first one is artifact ID and the second one is repo ID， which can be confusing, but this is not blocker, I'll approve.

PR #10734, opened 2020-02-16
Refactor the logic of deleting artifact

ywk253100:
Delete the child artifacts along with the parent when deleting an artifact

### DIFF
--- a/src/api/artifact/controller.go
+++ b/src/api/artifact/controller.go
@@ -274,30 +274,76 @@ func (c *controller) getByTag(ctx context.Context, repository, tag string, optio
 }
 
 func (c *controller) Delete(ctx context.Context, id int64) error {
-	// remove labels added to the artifact
-	if err := c.labelMgr.RemoveAllFrom(ctx, id); err != nil {
+	return c.deleteDeeply(ctx, id, true)
+}
+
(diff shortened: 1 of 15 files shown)


src/api/artifact/controller.go
```diff
@@ -269,26 +269,73 @@ func (c *controller) getByTag(ctx context.Context, repository, tag string, optio
...
 
 func (c *controller) Delete(ctx context.Context, id int64) error {
-	// delete all tags that attached to the artifact
-	_, tags, err := c.tagMgr.List(ctx, &q.Query{
+	return c.deleteDeeply(ctx, id, true)
+}
+
+// "isRoot" is used to specify whether the artifact is the root parent artifact
+// the error handling logic for the root parent artifact and others is different
+func (c *controller) deleteDeeply(ctx context.Context, id int64, isRoot bool) error {
```
steven-zou: Why not move the logic of checking if it requires deeply in the `Delete()` method? Then deleteDeeply only deeply deleted.

ywk253100: But the checking is also needed for the child artifacts

wy65701436: +1

PR #10735, opened 2020-02-16
Scan reorganize

heww:
1.  Move scan and scanner controllers from pkg to api.
2. Move artifact query from scan all job to its notification handler to ensure that the components in pkg will not call controllers in api.


Makefile
```diff
@@ -405,7 +405,7 @@ gofmt:
 
 commentfmt:
 	@echo checking comment format...
-	@res=$$(find . -type d \( -path ./src/vendor -o -path ./tests \) -prune -o -name '*.go' -print | xargs egrep '(^|\s)\/\/(\S)'); \
+	@res=$$(find . -type d \( -path ./src/vendor -o -path ./tests \) -prune -o -name '*.go' -print | xargs grep -P '(^|\s)\/\/(?!go:generate\s)(\S)'); \
```
steven-zou: Is this change related to this PR? It is not good for change tracking.

src/api/scan/base_controller.go
```diff
@@ -112,11 +112,57 @@ func NewController() Controller {
 				return "", errors.Errorf("configuration option %s not defined", cfg)
 			}
 		},
+		metadata: func(r *scanner.Registration) (*v1.ScannerAdapterMetadata, error) {
```
steven-zou: As the controller has been moved to here, does this callback func still need? I think it can all the scanner manager to get the scanner metadata.

src/api/scan/base_controller.go
```diff
@@ -112,11 +112,57 @@ func NewController() Controller {
...
+				return nil, errors.Wrap(err, "scan controller")
+			}
+			meta, err := client.GetMetadata()
+			if err != nil {
+				return nil, errors.Wrap(err, "scan controller")
+			}
+
+			return meta, nil
+		},
+		robot: func(projectID int64) (*model.Robot, error) {
```
steven-zou: Similar case with scanner meta.

src/api/scan/base_controller.go
```diff
@@ -63,31 +61,33 @@ type configGetter func(cfg string) (string, error)
...
 func NewController() Controller {
 	return &basicController{
 		// New report manager
 		manager: report.NewManager(),
-		// Refer to the default scanner controller
-		sc: sc.DefaultController,
-		// Refer to the default robot account controller
-		rc: robot.RobotCtr,
 		// Refer to the default job service client
 		jc: func() cj.Client {
```
steven-zou: The reason for this callback is because the initialization of the default jobservice client may be after this scan controller. If we have another better choice, maybe we can remove it too.

src/api/scan/base_controller.go
```diff
@@ -254,39 +297,33 @@ func (bc *basicController) Scan(artifact *v1.Artifact, options ...Option) error
 }
 
 // GetReport ...
-func (bc *basicController) GetReport(artifact *v1.Artifact, mimeTypes []string) ([]*scan.Report, error) {
+func (bc *basicController) GetReport(r *scanner.Registration, artifact *v1.Artifact, mimeTypes []string) ([]*scan.Report, error) {
```
steven-zou: If changeing the method signature to this mode, it seems that we should get the scanner meta in the API handler by scanner ctl and then pass it to the scan ctl.

Base on current programming principles, a ctl can call any components under pkg package and provide a full function. That means for the scan ctl, it can still provide scan (art, options) method.

src/api/scan/base_controller.go
```diff
@@ -467,8 +481,7 @@ func (bc *basicController) launchScanJob(trackID string, artifact *v1.Artifact,
 	if auth == authorizationBearer {
 		authorization, err = makeBearerAuthorization(artifact.Repository, fmt.Sprintf("%s:%s", registration.Name, registration.UUID))
```
steven-zou: We'll keep it here, right?

src/api/scan/base_controller.go
heww: Get registration of the project depend on scanner manager and project metadata manager,  so the `GetRegistrationByProject` can't be implemented in scanner manager.

If we process GetRegistrationByProject  in scan controller , it will  be duplicated with scanner controller.

src/api/scan/base_controller.go
heww: Will be removed in another PR.

wy65701436:
just one comment, the execution time of scan all job is too short(in one second). The system may crash if the admin repeats to trigger it.

PR #10739, opened 2020-02-17
Test reference for swagger client

danfengliu:


danfengliu:
It's only for a PR test, it's no longer useful anymore, so close it.

PR #10746, opened 2020-02-18
Move the scan overview populating logic to API handler

ywk253100:
Move the scan overview populating logic to API handler to avoid importing cycle

### DIFF
--- a/src/server/v2.0/handler/artifact.go
+++ b/src/server/v2.0/handler/artifact.go
@@ -25,7 +25,7 @@ import (
 	"github.com/goharbor/harbor/src/pkg/project"
 	"github.com/goharbor/harbor/src/pkg/q"
 	"github.com/goharbor/harbor/src/pkg/repository"
-	"github.com/goharbor/harbor/src/server/v2.0/models"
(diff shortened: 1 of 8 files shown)


src/server/v2.0/handler/artifact.go
```diff
@@ -98,14 +101,17 @@ func (a *artifactAPI) GetArtifact(ctx context.Context, params operation.GetArtif
...
 
 	// get the artifact
 	artifact, err := a.artCtl.GetByReference(ctx, fmt.Sprintf("%s/%s", params.ProjectName, params.RepositoryName), params.Reference, option)
 	if err != nil {
 		return a.SendError(ctx, err)
 	}
-	return operation.NewGetArtifactOK().WithPayload(artifact.ToSwagger())
+	art := &model.Artifact{}
+	art.Artifact = *artifact
+	a.assembleArtifact(ctx, art, params.WithScanOverview)
```
steven-zou: Maybe we can provide a utility method to handle scan summary and possible signature etc.

`ContentBuilder.Traget(*Arrtifact).WithScanSummary().WithSignature().WithLabels().FillIn()`

PR #10758, opened 2020-02-18
Copy over your config file

dvaerum:
I don't know if you forgot this step or it is intentional, but I forgot to copy over the harbor.yml file then trying to migrate to v1.10.1 so I think you should add this step :)


bitsf:
doc changed

PR #10792, opened 2020-02-20
Fix URL issue introduced by api version

ninjadq:
Fix chart API for v2.0

### DIFF
--- a/src/chartserver/handler_manipulation.go
+++ b/src/chartserver/handler_manipulation.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/ghodss/yaml"
+	"github.com/goharbor/harbor/src/common/api"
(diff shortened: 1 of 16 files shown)


src/chartserver/handler_manipulation.go
```diff
@@ -69,7 +69,7 @@ func (c *Controller) DeleteChartVersion(namespace, chartName, version string) er
 		return errors.New("invalid chart for deleting")
 	}
 
-	url := fmt.Sprintf("/api/chartrepo/%s/charts/%s/%s", namespace, chartName, version)
+	url := fmt.Sprintf("/api/v2.0/chartrepo/%s/charts/%s/%s", namespace, chartName, version)
```
ywk253100: How about define a const for the API version as we need to change it when bump up the API version again?

PR #10803, opened 2020-02-21
Revise the GC job flow

wy65701436:
1, set harbor to readonly
2, select the candidate artifacts from Harbor DB.
3, call registry API(--delete-untagged=false) to delete manifest bases on the results of #2
4, clean keys of redis DB of registry, clean artifact trash and untagged from DB.
5, roll back readonly.

### DIFF
--- a/src/core/api/reg_gc.go
+++ b/src/core/api/reg_gc.go
@@ -48,12 +48,18 @@ func (gc *GCAPI) Prepare() {
 //  "schedule": {
 //    "type": "Daily",
 //    "cron": "0 0 0 * * *"
+//  },
+//  "parameters": {
+//    "delete_untagged": true
 //  }
 //	}
 // create a manual trigger for GC
 // 	{
 //  "schedule": {
 //    "type": "Manual"
+//  },
+//  "parameters": {
+//    "delete_untagged": true
 //  }
 //	}
 func (gc *GCAPI) Post() {
@@ -64,9 +70,7 @@ func (gc *GCAPI) Post() {
 		return
 	}
 	ajr.Name = common_job.ImageGC
-	ajr.Parameters = map[string]interface{}{
-		"redis_url_reg": os.Getenv("_REDIS_URL_REG"),
-	}
+	ajr.Parameters["redis_url_reg"] = os.Getenv("_REDIS_URL_REG")
 	gc.submit(&ajr)
 	gc.Redirect(http.StatusCreated, strconv.FormatInt(ajr.ID, 10))
 }
@@ -77,6 +81,9 @@ func (gc *GCAPI) Post() {
 //  "schedule": {
 //    "type": "None",
 //    "cron": ""
+//  },
+//  "parameters": {
+//    "delete_untagged": true
 //  }
 //	}
 func (gc *GCAPI) Put() {
@@ -87,9 +94,7 @@ func (gc *GCAPI) Put() {
 		return
 	}
 	ajr.Name = common_job.ImageGC
-	ajr.Parameters = map[string]interface{}{
-		"redis_url_reg": os.Getenv("_REDIS_URL_REG"),
-	}
+	ajr.Parameters["redis_url_reg"] = os.Getenv("_REDIS_URL_REG")
 	gc.updateSchedule(ajr)
 }
 
--- a/src/jobservice/job/impl/default_context.go
+++ b/src/jobservice/job/impl/default_context.go
@@ -17,6 +17,8 @@ package impl
 import (
 	"context"
 	"errors"
+	o "github.com/astaxie/beego/orm"
+	"github.com/goharbor/harbor/src/internal/orm"
 	"github.com/goharbor/harbor/src/jobservice/job"
 	"github.com/goharbor/harbor/src/jobservice/logger"
 )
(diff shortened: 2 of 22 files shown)


src/core/api/reg_gc.go
```diff
@@ -87,9 +94,7 @@ func (gc *GCAPI) Put() {
 		return
 	}
 	ajr.Name = common_job.ImageGC
-	ajr.Parameters = map[string]interface{}{
-		"redis_url_reg": os.Getenv("_REDIS_URL_REG"),
-	}
+	ajr.Parameters["redis_url_reg"] = os.Getenv("_REDIS_URL_REG")
```
ywk253100: Why not set the `delete_untagged` parameter here?

src/jobservice/job/impl/default_context.go
```diff
@@ -49,7 +51,7 @@ func (dc *DefaultContext) Build(t job.Tracker) (job.Context, error) {
 	}
 
 	jContext := &DefaultContext{
-		sysContext: dc.sysContext,
+		sysContext: orm.NewContext(dc.sysContext, o.NewOrm()),
```
ywk253100: Is transaction needed here? If it is, could you add a `TODO` here or open an issue?

src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -66,6 +71,20 @@ func (gc *GarbageCollector) Validate(params job.Parameters) error {
...
 
 // Run implements the interface in job/Interface
+// The workflow of GC is:
+// 1, set harbor to readonly
+// 2, select the candidate artifacts from Harbor DB.
+// 3, call registry API(--delete-untagged=false) to delete manifest bases on the results of #2
+// 4, clean keys of redis DB of registry, clean artifact trash and untagged from DB.
+// 5, roll back readonly.
+// More details:
+// 1, why disable delete untagged when to call registry API
```
ywk253100: A reason that isn't so corner case is that we introduce `Harbor tag` in v2.0, Harbor tag is stored in database and there is no corresponding data in registry.
Please add this as well in case we forget in future

src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -169,60 +198,58 @@ func (gc *GarbageCollector) cleanCache() error {
...
-	}
-	for _, project := range projects {
-		pSize, err := dao.CountSizeOfProject(project.ProjectID)
-		if err != nil {
-			gc.logger.Warningf("error happen on counting size of project:%d by artifact, error:%v, just skip it.", project.ProjectID, err)
-			continue
+	// Sweep
+	for _, art := range required {
+		if err := deleteManifest(gc.registryURL, art.RepositoryName, art.Digest); err != nil {
+			flushTrash = false
```
ywk253100: Handle 404 error

src/core/api/reg_gc.go
wy65701436: it passed by UI.

src/jobservice/job/impl/default_context.go
wy65701436: so far, no need transaction here, will add a TODO

src/jobservice/job/impl/gc/garbage_collection.go
wy65701436: already did in the `deleteManifest`

src/jobservice/job/impl/gc/garbage_collection.go
wy65701436: will do

src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -169,60 +198,58 @@ func (gc *GarbageCollector) cleanCache() error {
...
-			gc.logger.Errorf("cannot delete untagged blobs of project: %d, err: %v, just skip it.", project.ProjectID, err)
-			continue
+	}
+	if gc.deleteUntagged {
+		for _, art := range untagged {
+			repo, err := gc.repoMgr.Get(ctx.SystemContext(), art.RepositoryID)
+			if err != nil {
+				return err
+			}
+			if err := deleteManifest(gc.registryURL, repo.Name, art.Digest); err != nil {
```
ywk253100: Only the manifest of the parent artifact is deleted, doesn't cover the children's.
I suggest that if the untagged artifacts are the candidates, delete them first, and this will move them(both parent and children) into the trash table, then you can  handle the trash

src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -169,60 +198,58 @@ func (gc *GarbageCollector) cleanCache() error {
...
-		iter, err = redis.Int(arr[0], nil)
-		if err != nil {
-			return fmt.Errorf("unexpected type for Int, got type %T", err)
-		}
-		k, err := redis.Strings(arr[1], nil)
-		if err != nil {
-			return fmt.Errorf("converts an array command reply to a []string %v", err)
-		}
-		keys = append(keys, k...)
+	}(flushTrash)
```
ywk253100: If you pass `flushTrash` as parameter, it will be always `true`

src/jobservice/job/impl/gc/garbage_collection.go
wy65701436: updated

PR #10817, opened 2020-02-24
Reimplement the registry client

ywk253100:
This commit reimplements the registry client under directory src/pkg/registry and removes the useless code


src/pkg/registry/client.go
```diff
@@ -48,54 +58,379 @@ var (
...
+		// no next page, end the loop
+		if len(url) == 0 {
+			break
+		}
+		// relative URL
+		if !strings.Contains(url, "://") {
+			url = c.url + url
+		}
+	}
+	sort.Strings(tags)
```
wy65701436: API already sorted the tags.

Issue #10760, opened 2020-02-19
Disable anonymous user


MaGaoju:
we are using Harbor internally and using OIDC provider for the auth. we don't want anonymous access for this instance. we'd like to keep all public project only accessible by authed users. it's possible to make this configurable ?   if code change needed, anyone can help to share this changes? 

thx!

cafeliker:
@reasonerjt I am reading the Harbor codes, could you pls help check if we can remove this section to disable the anonymous access?
![image](https://user-images.githubusercontent.com/6274204/74898190-1f359980-53d4-11ea-98fe-870f7051f7fc.png)

https://github.com/goharbor/harbor/blob/v1.10.1/src/core/api/project.go#L406

Thanks
ye

steven-zou:
@xaleeks 

Obviously, this is a requirement and then needs the decision from PM.

cafeliker:
Anonymous access is good for community purpose; and for Enterprise, the public repo should be only visible for authenticated people to protect the content and IP - so pls add the configuration item to enable/disable anonymous.

Before it's available, can we quickly fork and make some changes as I proposed above? Do you think it's the only place I need to change?

Thank you again for such great open source software.

ye

renmaosheng:
I got a similar request from "Xing ye Yin hang".

cafeliker:
I submitted a PR https://github.com/goharbor/harbor/pull/10825.

Thanks
ye

xaleeks:
agreed, makes sense for a lot of customers

reasonerjt:
I agree it's a valid requirement but we should put it in the pipeline and write proposal, then target it to a release, no?

cafeliker:
I will make the document, thanks for your information.

cafeliker:
@xaleeks I created the document with a PR in the harbor community repo at https://github.com/goharbor/community/pull/124.

Thanks
ye

xaleeks:
@cafeliker Thanks, I think it looks good.

sdileepkumarreddy:
Hi Any update in this feature? Would be nice to have this for handling some analytics

shreekanthDev:
Hello Team, any update on this feature implementation?

PR #10825, opened 2020-02-24
Enable configuration on disable anonymous user

cafeliker:
This is the proposed change on the issue #10760 

The administrator will be able to configure the Harbor instance to enable/disable the anonymous user with the UI like below:
image

Welcome for comments and feedback.

Thanks
ye


reasonerjt:
IMO, it would be a better design if we add more level of control in project's public policy

cafeliker:
> IMO, it would be a better design if we add more level of control in project's public policy

@reasonerjt Daniel, thanks for your comments, and can you be more specific?

ye

cafeliker:
Add anonymous check in the search API

cafeliker:
@steven-zou how will you merge this PR? I don't want to change every time you make a upgrade. Again, it should be the valid cyber security requirements if running Harbor in enterprise, specially when the enterprise Harbor service is internet facing.

cafeliker:
@steven-zou I updated the codes to work with harbor v2.0.1, could you pls help review it again?

Thanks
ye

src/common/config/metadata/metadatalist.go
```diff
@@ -121,6 +121,7 @@ var (
 
 		{Name: common.ProjectCreationRestriction, Scope: UserScope, Group: BasicGroup, EnvKey: "PROJECT_CREATION_RESTRICTION", DefaultValue: common.ProCrtRestrEveryone, ItemType: &ProjectCreationRestrictionType{}, Editable: false},
 		{Name: common.ReadOnly, Scope: UserScope, Group: BasicGroup, EnvKey: "READ_ONLY", DefaultValue: "false", ItemType: &BoolType{}, Editable: false},
+		{Name: common.DisableAnonymous, Scope: UserScope, Group: BasicGroup, EnvKey: "DISABLE_ANONYMOUS", DefaultValue: "false", ItemType: &BoolType{}, Editable: false},
```
stonezdj, line 124: To make it straightforward,  can we use AllowAnonymousUser and default value is true?

cafeliker: Do you mean allow sounds more positive than disable? ;-) I am ok with the change; and for now, the default value of DisableAnonymous is false.

steven-zou: I think we should keep consistent with the previous experience, allow anonymous. With this feature, the user can disable it via configuration.

cafeliker:
I just did the rebase to pass DCO.

PR #10831, opened 2020-02-24
add readonly to beego middleware

wy65701436:

### DIFF
--- a/src/core/middlewares/middlewares.go
+++ b/src/core/middlewares/middlewares.go
@@ -15,6 +15,7 @@
 package middlewares
 
 import (
+	"github.com/goharbor/harbor/src/server/middleware/readonly"
 	"net/http"
 	"path"
 	"regexp"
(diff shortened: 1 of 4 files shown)


src/server/middleware/readonly/readonly.go
```diff
@@ -67,9 +67,11 @@ func MiddlewareWithConfig(config Config, skippers ...middleware.Skipper) func(ht
...
 	return middleware.New(func(w http.ResponseWriter, r *http.Request, next http.Handler) {
 		if config.ReadOnly(r) {
-			pkgE := internal_errors.New(nil).WithCode(internal_errors.DENIED).WithMessage("The system is in read only mode. Any modification is prohibited.")
-			serror.SendError(w, pkgE)
-			return
+			if r.Method == http.MethodDelete || r.Method == http.MethodPost || r.Method == http.MethodPatch || r.Method == http.MethodPut {
+				pkgE := internal_errors.New(nil).WithCode(internal_errors.DENIED).WithMessage("The system is in read only mode. Any modification is prohibited.")
+				serror.SendError(w, pkgE)
+				return
+			}
```
heww: These changes are not needed here, only need to join the custom's skipers and the safeMethodSkipper .

wy65701436: thanks, updated.

src/server/registry/route.go
```diff
@@ -33,7 +33,8 @@ func RegisterRoutes() {
 	root := router.NewRoute().
 		Path("/v2").
 		Middleware(artifactinfo.Middleware()).
-		Middleware(v2auth.Middleware())
+		Middleware(v2auth.Middleware()).
+		Middleware(readonly.Middleware())
```
ywk253100: The `/v2/xxx` requests should already be covered by the global middleware?

src/core/middlewares/middlewares.go
```diff
@@ -29,11 +30,28 @@ import (
...
+		middleware.MethodAndPathSkipper(http.MethodPost, match("^/c/userExists")),
+		middleware.MethodAndPathSkipper(http.MethodPost, match("^/c/oidc/onboard")),
+		middleware.MethodAndPathSkipper(http.MethodPost, match("^/service/notifications/jobs/adminjob/"+numericRegexp.String())),
+		middleware.MethodAndPathSkipper(http.MethodPost, match("^/service/notifications/jobs/replication/"+numericRegexp.String())),
+		middleware.MethodAndPathSkipper(http.MethodPost, match("^/service/notifications/jobs/replication/task/"+numericRegexp.String())),
+		middleware.MethodAndPathSkipper(http.MethodPost, match("^/service/notifications/jobs/webhook/"+numericRegexp.String())),
+		middleware.MethodAndPathSkipper(http.MethodPost, match("^/service/notifications/jobs/retention/task/"+numericRegexp.String())),
+		middleware.MethodAndPathSkipper(http.MethodPost, match("^/service/notifications/jobs/schedules/"+numericRegexp.String())),
+		middleware.MethodAndPathSkipper(http.MethodPost, match("^/service/notifications/jobs/webhook/"+numericRegexp.String())),
+	}
```
ywk253100: How about just cover the requests whose path has prefix of `/v2/` or `/api/` ? Just like what the `legacyAPISkipper` does

Seems the `legacyAPISkipper` should be renamed...

src/server/registry/route.go
wy65701436: yes, will remove it.

PR #10841, opened 2020-02-25
Migrate artifact data in 2.0

ywk253100:
Abstract extra attributes and annotations for artifacts stored in database

### DIFF
--- a/make/migrations/postgresql/0030_2.0.0_schema.up.sql
+++ b/make/migrations/postgresql/0030_2.0.0_schema.up.sql
@@ -1,4 +1,26 @@
+/*
+table artifact:
+  id            SERIAL PRIMARY KEY NOT NULL,
+  type          varchar(255) NOT NULL,
+  media_type    varchar(255) NOT NULL,
+  manifest_media_type varchar(255) NOT NULL,
+  project_id    int NOT NULL,
+  repository_id int NOT NULL,
+  repository_name varchar(255) NOT NULL,
+  digest        varchar(255) NOT NULL,
+  size          bigint,
+  push_time     timestamp default CURRENT_TIMESTAMP,
+  pull_time     timestamp,
+  extra_attrs   text,
+  annotations   jsonb,
+  CONSTRAINT unique_artifact_2 UNIQUE (repository_id, digest)
+*/
+
 ALTER TABLE admin_job ADD COLUMN job_parameters varchar(255) Default '';
+
+/*record the data version to decide whether the data migration should be skipped*/
+ALTER TABLE schema_migrations ADD COLUMN data_version int;
+
 ALTER TABLE artifact ADD COLUMN repository_id int;
 ALTER TABLE artifact ADD COLUMN media_type varchar(255);
 ALTER TABLE artifact ADD COLUMN manifest_media_type varchar(255);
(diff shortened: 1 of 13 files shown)


src/migration/artifact.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/goharbor/harbor/src/pkg/repository"
+)
+
+func upgradeData(ctx context.Context) error {
+	abstractor := abstractor.NewAbstractor()
+	pros, err := project.Mgr.List()
+	if err != nil {
+		return err
+	}
+	for _, pro := range pros {
```
wy65701436: it has four levels for iteration, it's hard to maintain.

make/migrations/postgresql/0030_2.0.0_schema.up.sql
```diff
@@ -1,4 +1,8 @@
 ALTER TABLE admin_job ADD COLUMN job_parameters varchar(255) Default '';
+
+/*record the data version to decide whether the data migration should be skipped*/
```
wy65701436: replease `data_version` with `cur_version`? it's easy to understand.

src/migration/migration.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	ctx := orm.NewContext(context.Background(), beegorm.NewOrm())
+	dataVersion, err := getDataVersion(ctx)
+	if err != nil {
+		return err
+	}
+	log.Debugf("current data version: %v", dataVersion)
+	// the abstract logic already done before, skip
+	if dataVersion == 30 {
```
wy65701436: we do need to update the number at each release?

src/migration/artifact.go
```diff
@@ -0,0 +1,87 @@
...
+					log.Errorf("failed to update the artifact %s@%s: %v, skip", repo.Name, art.Digest, err)
+					continue
+				}
+			}
+			log.Debugf("artifact metadata under repository %s abstracted", repo.Name)
+		}
+	}
+
+	// update data version
+	return setDataVersion(ctx, 30)
```
wy65701436: can we define a global varaible for `30`?

make/migrations/postgresql/0030_2.0.0_schema.up.sql
```diff
@@ -1,4 +1,8 @@
...
+ALTER TABLE schema_migrations ADD COLUMN data_version int;
```
heww: Can data version use the same version of schema because data migrations will  appear with schema migration when it exist?

make/migrations/postgresql/0030_2.0.0_schema.up.sql
ywk253100: What does the `cur_version` stand for? Is it current version?
The table `schema_migrations` contains a column `version` which indicates the schema version. Use `data_version` to distinguish it
```
CREATE TABLE schema_migrations(version bigint not null primary key, dirty boolean not null);
```

make/migrations/postgresql/0030_2.0.0_schema.up.sql
ywk253100: Seems can't. The migrations for schema and data are two separated processes. We cannot control the migration process of schema. It's possible that the schema migration success but the data migration fails

src/migration/migration.go
ywk253100: No, only update it when the data version needs to be bumped up

src/migration/artifact.go
ywk253100: sure

src/migration/artifact.go
ywk253100: What I'm considering is that as the artifact model contains server properties, if we put all artifacts into the memory, it will consume huge space. But we can do it per repository, this will reduce one level for iteration

make/migrations/postgresql/0030_2.0.0_schema.up.sql
heww: I got it

PR #10854, opened 2020-02-26
Fix transaction issue

ywk253100:

### DIFF
--- a/src/api/artifact/controller.go
+++ b/src/api/artifact/controller.go
@@ -172,57 +172,44 @@ func (c *controller) ensureArtifact(ctx context.Context, repository, digest stri
 	// populate the artifact type
 	artifact.Type = descriptor.GetArtifactType(artifact.MediaType)
 
-	// create it
-	id, err := c.artMgr.Create(ctx, artifact)
+	// create the artifact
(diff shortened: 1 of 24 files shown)


heww:
The following is the implemention of the `ReadOrCreate` in beego,  I think this method will not fix the duplicate issue in  transaction, we should wrap the `Insert` method in nested transaction (savepoint). 

``` go
func (o *orm) ReadOrCreate(md interface{}, col1 string, cols ...string) (bool, int64, error) {
	cols = append([]string{col1}, cols...)
	mi, ind := o.getMiInd(md, true)
	err := o.alias.DbBaser.Read(o.db, mi, ind, o.alias.TZ, cols, false)
	if err == ErrNoRows {
		// Create
		id, err := o.Insert(md)
		return (err == nil), id, err
	}

	id, vid := int64(0), ind.FieldByIndex(mi.fields.pk.fieldIndex)
	if mi.fields.pk.fieldType&IsPositiveIntegerField > 0 {
		id = int64(vid.Uint())
	} else if mi.fields.pk.rel {
		return o.ReadOrCreate(vid.Interface(), mi.fields.pk.relModelInfo.fields.pk.name)
	} else {
		id = vid.Int()
	}

	return false, id, err
}
```

PR #10857, opened 2020-02-26
feat(oci) remove dead code pkg/art

bitsf:


src/pkg/artifact/candidate.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package art
+package artifact
```
ywk253100: According to the "single responsibility principle" for pkg, please move the selector related code into a separated package "src/pkg/selector"

ywk253100:
@bitsf Please resolve the conflicts

renmaosheng:
just curios, why the coverage is dropping?

bitsf:
> just curios, why the coverage is dropping?
image

It seems some UI case dropped, but I also don't know why yet, this PR did nothing with UI code, maybe the calculate comparison commit is not correct

PR #10858, opened 2020-02-26
Remove docs folder from CI runs

jonasrosland:
This will remove the docs folder from triggering CI runs.

Signed-off-by: jonasrosland <jrosland@vmware.com>

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -18,6 +18,8 @@ env:
 on:
   pull_request:
   push:
+    paths-ignore:
+    - 'docs/**'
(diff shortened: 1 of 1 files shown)


bitsf:
I think pull_request can also remove this path

bitsf:
oh, seems can't remove from pull_request, as it is a mandatory check before merge now.

jonasrosland:
Any blocking issues with this @wy65701436 ? Removing the docs here will make sure our CI systems don't spend a ton of time building and testing when we just have a small documentation change done.

PR #10859, opened 2020-02-26
Trigger a new Netlify build when new docs are merged

jonasrosland:
This will trigger a new build of the main website as soon as new docs are merged to master.
The required secret needs to be added to the project.

Signed-off-by: jonasrosland <jrosland@vmware.com>


jonasrosland:
The required secret has been added by @steven-zou, so this should be able to get merged now. It will help as it triggers an automatic new website build and publish whenever the docs are edited.

PR #10863, opened 2020-02-27
Fix Api cennter

### LINKED ISSUE #10634: Swagger changes to align v2 api
According the v2 api changes, we will have 2 version of apis. We need to figure out if we will use 1 swagger yaml file to show both or seperate files. 
Make sure all apis should be showed in the dev denter correctly.

AllForNothing:
Fix #10634 
Env.:https://10.161.79.96:8083

### DIFF
--- a/src/portal/src/lib/utils/utils.ts
+++ b/src/portal/src/lib/utils/utils.ts
@@ -597,3 +597,34 @@ export function formatSize(tagSize: string): string {
         return size + "B";
     }
 }
+
+/**
+ * Simple object check.
+ * @param item
+ * @returns {boolean}
+ */
+export function isObject(item): boolean {
+    return (item && typeof item === 'object' && !Array.isArray(item));
+}
+
+/**
+ * Deep merge two objects.
+ * @param target
+ * @param ...sources
+ */
+export function mergeDeep(target, ...sources) {
+    if (!sources.length) { return target; }
+    const source = sources.shift();
+
+    if (isObject(target) && isObject(source)) {
+        for (const key in source) {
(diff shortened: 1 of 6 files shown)


src/portal/src/lib/utils/utils.ts
```diff
@@ -597,3 +597,34 @@ export function formatSize(tagSize: string): string {
...
     }
 }
+
+/**
+ * Simple object check.
+ * @param item
+ * @returns {boolean}
+ */
+export function isObject(item): boolean {
+    return (item && typeof item === 'object' && !Array.isArray(item));
```
jwangyangls: if item is a function?

src/portal/src/lib/utils/utils.ts
```diff
@@ -597,3 +597,34 @@ export function formatSize(tagSize: string): string {
...
+        for (const key in source) {
+            if (isObject(source[key])) {
+                if (!target[key]) { Object.assign(target, { [key]: {} }); }
+                mergeDeep(target[key], source[key]);
+            } else {
+                Object.assign(target, { [key]: source[key] });
+            }
+        }
+    }
+    return mergeDeep(target, ...sources);
```
jwangyangls: Do you need···

src/portal/src/lib/utils/utils.ts
AllForNothing: will return "function"

src/portal/src/lib/utils/utils.ts
AllForNothing: needed, because "source"s can be one more objects

PR #10864, opened 2020-02-27
Set the "Link" and "location" header

ywk253100:
Set the "Link" and "location" header

### DIFF
--- a/src/server/v2.0/handler/base.go
+++ b/src/server/v2.0/handler/base.go
@@ -19,7 +19,10 @@ package handler
 import (
 	"context"
 	"errors"
+	"github.com/goharbor/harbor/src/internal"
 	ierror "github.com/goharbor/harbor/src/internal/error"
+	"net/url"
+	"strconv"
 
 	"github.com/go-openapi/runtime/middleware"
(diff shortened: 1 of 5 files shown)


src/server/v2.0/handler/base.go
```diff
@@ -95,3 +98,32 @@ func (b *BaseAPI) RequireProjectAccess(ctx context.Context, projectIDOrName inte
 	}
 	return ierror.ForbiddenError(nil)
 }
+
+// Links return Links based on the provided pagination information
+// The parameter url is value rather the reference to avoid overwrite the original values
+func (b *BaseAPI) Links(ctx context.Context, url url.URL, total, pageNumber, pageSize int64) internal.Links {
```
heww: `url` parameter can be a pointer here so that every where to call this method not needed to covert the URL of the http.Request t to non pointer.

ywk253100: As the comment of the method said: `The parameter url is value rather the reference to avoid overwrite the original values`

heww: Why not copy the URL in the method?

wy65701436: no test?

PR #10868, opened 2020-02-27
Add "Docker-Distribution-Api-Version" header for the 401 response of registry API

ywk253100:
This is needed for "docker manifest" commands: https://github.com/docker/cli/issues/989


renmaosheng:
good catch！

PR #10881, opened 2020-02-28
feat(artifact): add Walk method to artifact controller

heww:
Add Walk method to artifact controller which walks the artifact tree rooted at root, calling walkFn for each artifact in the tree, including root.

### DIFF
--- a/src/api/artifact/controller.go
+++ b/src/api/artifact/controller.go
@@ -15,8 +15,12 @@
 package artifact
 
 import (
+	"container/list"
 	"context"
 	"fmt"
+	"strings"
+	"time"
+
 	"github.com/goharbor/harbor/src/api/artifact/abstractor"
 	"github.com/goharbor/harbor/src/api/artifact/abstractor/resolver"
 	"github.com/goharbor/harbor/src/api/artifact/descriptor"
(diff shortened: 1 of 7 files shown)


src/api/artifact/controller.go
```diff
@@ -507,6 +512,48 @@ func (c *controller) RemoveLabel(ctx context.Context, artifactID int64, labelID
...
+		artifact := elem.Value.(*Artifact)
+		if err := walkFn(artifact); err != nil {
+			return err
+		}
+
+		if !artifact.HasChildren() {
+			continue
+		}
+
+		refs, err := c.artMgr.ListReferences(ctx, q.New(q.KeyWords{"parent_id": artifact.ID}))
```
ywk253100: The `Artifact` model contains the references already, no need to list them again

src/api/artifact/controller.go
```diff
@@ -507,6 +512,48 @@ func (c *controller) RemoveLabel(ctx context.Context, artifactID int64, labelID
...
+	for queue.Len() > 0 {
+		elem := queue.Front()
+		queue.Remove(elem)
+
+		artifact := elem.Value.(*Artifact)
+		if err := walkFn(artifact); err != nil {
+			return err
+		}
+
+		if !artifact.HasChildren() {
```
ywk253100: Whether the artifact contains children or not can be determined by the length of references. The `HasChildren` isn't needed?

ywk253100: I'm not sure it's the correct way to determine whether the artifact contains children by checking the media type. But the checking the references must be

heww: I think we should check it by artifact media type, not the db query, and in artifact manager we should query references only when the media type is manifest index.

PR #10887, opened 2020-02-28
Filter artifact by type/tag  and change error setting

jwangyangls:


src/portal/src/lib/components/image-name-input/image-name-input.component.ts
```diff
@@ -42,7 +42,7 @@ export class ImageNameInputComponent implements OnInit, OnDestroy {
     ngOnInit(): void {
         this.proNameChecker
             .pipe(debounceTime(200))
-            .pipe(distinctUntilChanged(),
```
AllForNothing: why delete?

jwangyangls: Because we need to double check

PR #10892, opened 2020-02-28
feat(addition-link): only set vuls addition link when artifact scanable

heww:
1. Add Checker to check the scannable status of the artifact.
2. Only set vulnerabilities addition link when the artifact scanable in the
project.


src/api/scan/checker.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+
+		metadata = md
+		c.scannerMetadatas[projectID] = md
+	}
+
+	var scannable bool
+
+	walkFn := func(a *artifact.Artifact) error {
+		scannable = metadata.HasCapability(a.ManifestMediaType)
```
ywk253100: The type of artifact is identified by artifact.MediaType which is the media type of the config layer

src/api/artifact/controller.go
```diff
@@ -453,6 +459,10 @@ func (c *controller) Walk(ctx context.Context, root *Artifact, walkFn func(*Arti
 
 		artifact := elem.Value.(*Artifact)
 		if err := walkFn(artifact); err != nil {
+			if err == ErrBreak {
```
ywk253100: How about defining the walkFn as `func(*Artifact)(continue bool, err error)`, the first returning parameter indicates whether the walk should continue or not

src/api/scan/checker.go
heww: The capability of scanner use the manifest media type to check 

image

src/api/artifact/controller.go
heww: `func(*Artifact) err` is just ok for this, the `filepath.Walk` returns `filepath.SkipDir` to skip the remaining files in the containing directory.

src/api/scan/checker.go
```diff
@@ -0,0 +1,85 @@
...
+	scannerCtl       scanner.Controller
+	scannerMetadatas map[int64]*v1.ScannerAdapterMetadata
+}
+
+func (c *checker) IsScannable(ctx context.Context, art *artifact.Artifact) (bool, error) {
+	projectID := art.ProjectID
+
+	metadata, ok := c.scannerMetadatas[projectID]
+	if !ok {
+		registration, err := c.scannerCtl.GetRegistrationByProject(projectID, scanner.WithPing(false))
```
wy65701436: why not use WithPing(true)?

src/api/scan/checker.go
wy65701436: should we use the type as well? Otherwise, how can we know the Chart is un-scannable?

src/api/scan/checker.go
heww: `Ping` will be called in the next step, so disable ping for `GetRegistrationByProject`

PR #10899, opened 2020-03-01
fix[js]:fix ut case faulure

steven-zou:
- refactor default context creation to avoid data race
- refactor the timer interval in c_worker UT cases to avoid receiving signals at the same time

### DIFF
--- a/src/jobservice/worker/cworker/c_worker_test.go
+++ b/src/jobservice/worker/cworker/c_worker_test.go
@@ -17,10 +17,15 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync"
+	"testing"
+	"time"
+
 	common_dao "github.com/goharbor/harbor/src/common/dao"
 	"github.com/goharbor/harbor/src/jobservice/common/utils"
 	"github.com/goharbor/harbor/src/jobservice/env"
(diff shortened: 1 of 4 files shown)


src/jobservice/worker/cworker/c_worker_test.go
```diff
@@ -190,7 +191,7 @@ func (suite *CWorkerTestSuite) TestStopJob() {
 	t, err := suite.lcmCtl.New(genericJob)
 	require.NoError(suite.T(), err, "new job stats: nil error expected but got %s", err)
 
-	tk := time.NewTicker(500 * time.Millisecond)
+	tk := time.NewTicker(417 * time.Millisecond)
```
renmaosheng: just curious, why it is 417? this comment should not blocks your check -in.

steven-zou: An odd number to

`
refactor the timer interval in c_worker UT cases to avoid receiving signals at the same time
`

steven-zou: select has two channels, they may have signals coming in at the same time. Under this case, go will select one channel randomly to process, which may cause the time out failure to handle.

PR #10901, opened 2020-03-02
Fix xsrf error status bug when after beego update 1.10

jwangyangls:


AllForNothing:
Dulplicated with #10902

PR #10902, opened 2020-03-02
Fix xsrf error status bug when after beego update

jwangyangls:


src/portal/src/app/intercept-http.service.ts
```diff
@@ -14,7 +14,7 @@ export class InterceptHttpService implements HttpInterceptor {
   intercept(request: HttpRequest<any>, next: HttpHandler): Observable<any> {
 
     return next.handle(request).pipe(catchError(error => {
-      if (error.status === 403) {
+      if (error.status === 422) {
```
AllForNothing: better use a meaningful variable name

PR #10903, opened 2020-03-02
refactor(artifact): impl the copy artifact by walk of root artifact

heww:
1. Implement the copy artifact method by walk the src root artifact to
collect the copy artifacts.
2. This fix the issue that copy artifact not copy its children.

### DIFF
--- a/src/api/artifact/controller.go
+++ b/src/api/artifact/controller.go
@@ -347,67 +347,68 @@ func (c *controller) deleteDeeply(ctx context.Context, id int64, isRoot bool) er
 }
 
 func (c *controller) Copy(ctx context.Context, srcRepo, reference, dstRepo string) (int64, error) {
-	return c.copyDeeply(ctx, srcRepo, reference, dstRepo, true)
-}
-
-// as we call the docker registry APIs in the registry client directly,
-// this bypass our own logic(ensure, fire event, etc.) inside the registry handlers,
-// these logic must be covered explicitly here.
-// "copyDeeply" iterates the child artifacts and copy them first
-func (c *controller) copyDeeply(ctx context.Context, srcRepo, reference, dstRepo string, isRoot bool) (int64, error) {
-	var option *Option
-	// only get the tags of the root parent
-	if isRoot {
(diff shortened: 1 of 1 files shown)


src/api/artifact/controller.go
```diff
@@ -347,67 +347,68 @@ func (c *controller) deleteDeeply(ctx context.Context, id int64, isRoot bool) er
...
-	// the artifact doesn't exist under the destination repository, continue to copy
-	// copy child artifacts if contains any
-	for _, reference := range srcArt.References {
-		if _, err = c.copyDeeply(ctx, srcRepo, reference.ChildDigest, dstRepo, false); err != nil {
+	var rootID int64
+	for _, artifact := range copyArtifacts {
+		digest := artifact.Digest
+
+		// copy the artifact into the backend docker registry
+		if err := c.regCli.Copy(srcRepo, digest, dstRepo, digest, false); err != nil {
```
wy65701436: move it after Ensure success, otherwise, the ensure can be failed by immutable or signature checking.

src/api/artifact/controller.go
```diff
@@ -347,67 +347,68 @@ func (c *controller) deleteDeeply(ctx context.Context, id int64, isRoot bool) er
...
-	if err == nil {
-		// return conflict error if the root parent artifact already exists under the destination repository
-		if isRoot {
-			return 0, ierror.New(nil).WithCode(ierror.ConflictCode).
-				WithMessage("the artifact %s@%s already exists", dstRepo, digest)
+	var copyArtifacts []*Artifact
+	walkFn := func(a *Artifact) error {
+		digest := a.Digest
+
+		_, err := c.GetByReference(ctx, dstRepo, digest, nil)
```
wy65701436: what if the source artifact have tags but doesn't have in target? The tags are lost?

I think we can leverge Ensure to judge whether to create in the target.

src/api/artifact/controller.go
heww: Ensure will be failed when move this after the the `Ensure`, `Ensure` depend on the manifest in the registry.

PR #10907, opened 2020-03-02
feat(quota): add Request and Refresh middlewares for APIs

heww:
1. Introduce ReqquestMiddleware and RefereshMiddleware.
2. Add request middlware to copy artifact, mount blob, put blob upload,
put manifest, upload chart verson APIs.
3. Add refresh project middleware to delete manifest, delete artifact,
delete chart version, delete repository APIs.


src/api/blob/controller.go
```diff
@@ -188,6 +160,69 @@ func (c *controller) Exist(ctx context.Context, digest string, options ...Option
...
+	for _, blob := range blobs {
+		if !associated[blob.ID] {
+			results = append(results, blob)
+		}
+	}
+
+	return results, nil
+}
+
+func (c *controller) Get(ctx context.Context, digest string, options ...Option) (*blob.Blob, error) {
```
wy65701436: if I only pass the digest, what blob should I get?

src/server/middleware/blob/copy_artifact.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/goharbor/harbor/src/server/middleware"
+	"github.com/goharbor/harbor/src/server/middleware/util"
+)
+
+func isSuccess(statusCode int) bool {
+	return statusCode >= http.StatusOK && statusCode < http.StatusBadRequest
+}
+
+// CopyArtifactMiddleware middleware to sync the missing associations for the project
+func CopyArtifactMiddleware() func(http.Handler) http.Handler {
```
wy65701436: does the error code meet the OCI spec?

src/server/v2.0/handler/handler.go
```diff
@@ -31,6 +35,10 @@ func New() http.Handler {
 		log.Fatal(err)
 	}
 
+	api.RegisterMiddleware("CopyArtifact", middleware.Chain(quota.CopyArtifactMiddleware(), blob.CopyArtifactMiddleware()))
```
wy65701436: is it possible to merge these two middlewares? They share some queries, I worried about the performance impact.

heww: This are two things, quota can be disabled, blob can not be disabled , gc depent on blobs.

Performance problem can be solved by other methods.

src/server/middleware/blob/copy_artifact.go
heww: I think so, the `err` is processed by `SendError` in the `server/error` pkg.

src/api/blob/controller.go
heww: The blob of the digest itself.

src/server/middleware/quota/copy_artifact.go
```diff
@@ -0,0 +1,142 @@
...
+		log.Errorf("%s: parse from parameter failed, error: %v", logPrefix, err)
+		return nil, nil
+	}
+
+	ctx := r.Context()
+
+	art, err := artifactController.GetByReference(ctx, repository, reference, nil)
+	if ierror.IsNotFoundErr(err) {
+		// artifact not found, discontinue the API request
+		return nil, ierror.BadRequestError(nil).WithMessage("artifact %s not found", from)
```
ywk253100: Seems you can return error directly without creating a new one

src/pkg/blob/dao/dao.go
```diff
@@ -167,12 +172,34 @@ func (d *dao) UpdateBlob(ctx context.Context, blob *models.Blob) error {
...
-func (d *dao) ListBlobs(ctx context.Context, query *q.Query) ([]*models.Blob, error) {
-	qs, err := orm.QuerySetter(ctx, &models.Blob{}, query)
+func (d *dao) ListBlobs(ctx context.Context, params models.ListParams) ([]*models.Blob, error) {
+	qs, err := orm.QuerySetter(ctx, &models.Blob{}, nil)
 	if err != nil {
 		return nil, err
 	}
 
+	if len(params.BlobDigests) > 0 {
+		qs = qs.Filter("digest__in", params.BlobDigests)
```
ywk253100: Replace all `__in` and other Beego ORM operator if any with the new query model

src/server/middleware/quota/copy_artifact.go
heww: This change the response from 404 to 400.

src/pkg/blob/dao/dao.go
heww: I make `QuerySetter` support `xxx__in` key in `q.KeyWords`.

PR #10923, opened 2020-03-03
Make sure replication work with OCI artifacts(phase 1)

ywk253100:
This commit updates the definition of replicated resource(artifacts replace the vtags) and refactor the filter part


src/replication/adapter/aliacr/adapter.go
```diff
@@ -253,8 +253,7 @@ func (a *adapter) FetchImages(filters []*model.Filter) (resources []*model.Resou
 						Repository: &model.Repository{
 							Name: filepath.Join(repo.RepoNamespace, repo.RepoName),
 						},
-						Vtags:  filterTags,
-						Labels: []string{},
```
steven-zou: still not support label filter?

PR #10924, opened 2020-03-03
use unified version as tag name

### LINKED ISSUE #10915: use unified tagname for harbor docker images
Currently some harbor image's tags are using version, some are using version-version

image

This is quite complex, especially when using in operator.
We should use only version of harbor (eg. change from goharbor/notary-server-photon:v0.6.1-v1.10.1 to goharbor/notary-server-photon:v1.10.1), and we keep these version in a file, and in future we may provide a API to query this

bitsf:
fix #10915
use unified version as tag name
eg. change from goharbor/notary-server-photon:v0.6.1-v1.10.1 to goharbor/notary-server-photon:v1.10.1


reasonerjt:
Are there other references in the `.py` files that should be also cleaned up?

bitsf:
> the ./test/integration.sh should be updated, it pushs all dev images into docker hub for helm use.
> 
> The script is capturing the image tag and update it to `dev`, then to push them into docker hub.

no need change this

PR #10925, opened 2020-03-03
use unified version as tag name for 1.10.2

### LINKED ISSUE #10915: use unified tagname for harbor docker images

bitsf:
fix #10915
use unified version as tag name
eg. change from goharbor/notary-server-photon:v0.6.1-v1.10.1 to goharbor/notary-server-photon:v1.10.1


make/photon/prepare/utils/docker_compose.py
```diff
@@ -19,12 +19,12 @@ def prepare_docker_compose(configs, with_clair, with_notary, with_chartmuseum):
 
     rendering_variables = {
         'version': VERSION_TAG,
-        'reg_version': "{}-{}".format(REGISTRY_VERSION, VERSION_TAG),
+        'reg_version': VERSION_TAG,
```
reasonerjt: seems there are other references that can be removed in prepare scripts?

bitsf: 我觉得先留着吧，万一哪天希望在prepare时把这些版本信息打印出来或者输出到文件呢，还可以复用一下

PR #10937, opened 2020-03-04
Update CSRF mechanism

reasonerjt:
This commit replaces beego's CSRF mechanism with gorilla's csrf library.
The criteria for requests to skip the csrf check remain the same.


make/photon/prepare/templates/core/app.conf.jinja
```diff
@@ -4,6 +4,4 @@ enablegzip = true
 
 [dev]
 httpport = 8080
-EnableXSRF = true
 XSRFKey = {{xsrf_key}}
```
ywk253100: Is this still needed?

src/server/middleware/csrf/csrf.go
```diff
@@ -0,0 +1,90 @@
...
+			SameSite: http.SameSiteStrictMode,
+		})
+	} else {
+		log.Warningf("token not found in context, skip attaching")
+	}
+}
+
+func handleError(w http.ResponseWriter, r *http.Request) {
+	attachToken(w, r)
+	http.Error(w, fmt.Sprintf("%s - %s",
```
ywk253100: Use https://github.com/goharbor/harbor/blob/master/src/server/error/error.go#L49 instead, so that we can control the error format in one place

make/photon/prepare/templates/core/app.conf.jinja
reasonerjt: no, removed.

src/server/middleware/csrf/csrf.go
reasonerjt: Yes, definitely!

PR #10939, opened 2020-03-04
add audit logs API

wy65701436:
1, add API entry for get audit logs
2. add audit log manager to hanlder CRUD

Use the new format of audit log to cover differernt resource, artifact/tag/repostory/project

### DIFF
--- a/api/v2.0/swagger.yaml
+++ b/api/v2.0/swagger.yaml
@@ -499,6 +499,68 @@ paths:
           $ref: '#/responses/409'
         '500':
           $ref: '#/responses/500'
+  /audit-logs:
+    get:
+      summary: Get recent logs of the projects which the user is a member of
+      description: |
+        This endpoint let user see the recent operation logs of the projects which he is member of
+      tags:
+        - auditlog
+      operationId: listAuditLogs
+      parameters:
+        - $ref: '#/parameters/requestId'
+        - $ref: '#/parameters/page'
+        - $ref: '#/parameters/pageSize'
+        - name: username
+          in: query
+          type: string
+          required: false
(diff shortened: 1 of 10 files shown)


make/migrations/postgresql/0030_2.0.0_schema.up.sql
```diff
@@ -173,3 +173,14 @@ CREATE TABLE artifact_2
...
   CONSTRAINT unique_artifact_2 UNIQUE (project_id, repo, tag)
 );
+
+CREATE TABLE audit_log
+(
+ id             SERIAL PRIMARY KEY NOT NULL,
+ project_id     int NOT NULL,
+ operation      varchar(20) NOT NULL,
+ resource_type  varchar(255) NOT NULL,
+ resource       varchar(255) NOT NULL,
```
ywk253100: The length of repository+digest may exceeds  255

api/v2.0/swagger.yaml
```diff
@@ -499,6 +499,68 @@ paths:
           $ref: '#/responses/409'
         '500':
           $ref: '#/responses/500'
+  /auditlogs:
```
ywk253100: According to our internal REST API standard `URIs SHOULD use kebab-case`, `/audit-logs` is better

make/migrations/postgresql/0030_2.0.0_schema.up.sql
wy65701436: extend it to 1024

api/v2.0/swagger.yaml
wy65701436: updated, thx.

PR #10955, opened 2020-03-05
add sql for migrating access log

wy65701436:
1, loop each access log, change to resource/resource_type, and insert into audit log
2, loop each first push operation, change it to create repository and insert into audit log.

### DIFF
--- a/make/migrations/postgresql/0030_2.0.0_schema.up.sql
+++ b/make/migrations/postgresql/0030_2.0.0_schema.up.sql
@@ -184,3 +184,24 @@ CREATE TABLE audit_log
  username       varchar(255) NOT NULL,
  op_time        timestamp default CURRENT_TIMESTAMP
 );
+
+/*migrate access log to audit log*/
+/*TODO drop table access_log?*/
+DO $$
+DECLARE
+    access RECORD;
+BEGIN
+    FOR access IN SELECT * FROM access_log
+    LOOP
+      /*insert project create and delete*/
+      IF (access.operation = 'create' AND access.repo_tag = 'N/A') OR (access.operation = 'delete' AND access.repo_tag = 'N/A') THEN
+          INSERT INTO audit_log (project_id, operation, resource_type, resource, username, op_time) VALUES (access.project_id, access.operation, 'project', access.repo_name, access.username, access.op_time);
+      ELSIF access.operation = 'delete' AND access.repo_tag != 'N/A' THEN
+          INSERT INTO audit_log (project_id, operation, resource_type, resource, username, op_time) VALUES (access.project_id, 'delete', 'artifact', CONCAT(access.repo_name,':',access.repo_tag), access.username, access.op_time);
+      ELSIF access.operation = 'push' THEN
+          INSERT INTO audit_log (project_id, operation, resource_type, resource, username, op_time) VALUES (access.project_id, 'create', 'artifact', CONCAT(access.repo_name,':',access.repo_tag), access.username, access.op_time);
+      ELSIF access.operation = 'pull' THEN
+          INSERT INTO audit_log (project_id, operation, resource_type, resource, username, op_time) VALUES (access.project_id, 'pull', 'artifact', CONCAT(access.repo_name,':',access.repo_tag), access.username, access.op_time);
+      END IF;
+    END LOOP;
+END $$;


make/migrations/postgresql/0030_2.0.0_schema.up.sql
```diff
@@ -184,3 +184,43 @@ CREATE TABLE audit_log
...
+/*migrate access log to audit log*/
+/*TODO drop table access_log?*/
+DO $$
+DECLARE
+    access RECORD;
+BEGIN
+    FOR access IN SELECT * FROM access_log
+    LOOP
+      /*insert project create and delete*/
+      IF access.operation = 'create' OR access.operation = 'delete' THEN
```
ywk253100: The resources deleted can be project and image

make/migrations/postgresql/0030_2.0.0_schema.up.sql
```diff
@@ -184,3 +184,43 @@ CREATE TABLE audit_log
...
+          INSERT INTO audit_log (project_id, operation, resource_type, resource, username, op_time) VALUES (access.project_id, access.operation, 'project', access.repo_name, access.username, access.op_time);
+      END IF;
```
ywk253100: Using `if xxx else if xxx end` can reduce the comparing times

make/migrations/postgresql/0030_2.0.0_schema.up.sql
```diff
@@ -184,3 +184,43 @@ CREATE TABLE audit_log
...
+      /*insert push image -> create artifact*/
+      IF access.operation = 'push' THEN
+          INSERT INTO audit_log (project_id, operation, resource_type, resource, username, op_time) VALUES (access.project_id, 'create', 'artifact', CONCAT(access.repo_name,'/',access.repo_tag), access.username, access.op_time);
```
ywk253100: Let's use `push` as the operation as the commands are `push` and the operation times are also `push time`?

ywk253100: Should be `repo_name:repo_tag`

make/migrations/postgresql/0030_2.0.0_schema.up.sql
```diff
@@ -184,3 +184,43 @@ CREATE TABLE audit_log
...
+      END IF;
+      /*insert pull image -> pull tag*/
+      IF access.operation = 'pull' THEN
+          INSERT INTO audit_log (project_id, operation, resource_type, resource, username, op_time) VALUES (access.project_id, 'pull', 'tag', CONCAT(access.repo_name,'/',access.repo_tag), access.username, access.op_time);
```
ywk253100: The resource type should be `artifact`

make/migrations/postgresql/0030_2.0.0_schema.up.sql
```diff
@@ -184,3 +184,43 @@ CREATE TABLE audit_log
...
+      END IF;
+    END LOOP;
+END $$;
+
+/*migrate access log to audit log*/
+/*user the first push operation to generate the create repository log*/
+DO $$
```
ywk253100: If we generate the audit log for repository creation, how about the repository deletion? 
We cannot handle the case that the image is pushed, then deleted, then pushed again, then deleted, which means the repository is created, then deleted and created, deleted

make/migrations/postgresql/0030_2.0.0_schema.up.sql
wy65701436: updated.

make/migrations/postgresql/0030_2.0.0_schema.up.sql
wy65701436: that's why it just handles the first creation

make/migrations/postgresql/0030_2.0.0_schema.up.sql
wy65701436: the resource type is `artifact` when pull by digest? If pull by tag, the type shouldn't be `tag`?

make/migrations/postgresql/0030_2.0.0_schema.up.sql
wy65701436: in the new OCI framework, do we have `push` action? Isn't it create an artifact?

make/migrations/postgresql/0030_2.0.0_schema.up.sql
wy65701436: OK

make/migrations/postgresql/0030_2.0.0_schema.up.sql
wy65701436: sure

make/migrations/postgresql/0030_2.0.0_schema.up.sql
wy65701436: updated.

make/migrations/postgresql/0030_2.0.0_schema.up.sql
wy65701436: removed.

PR #10959, opened 2020-03-06
add ut for dragonfly driver

ansinlee:
test passed
dragonfly endpoint: 47.111.250.75:8002


chlins:
@ansinlee Looks great, but does this dragonfly endpoint always online, if the server done in the future, the unit test maybe fail?

steven-zou:
We have implemented P2P with a new code structure and UT cases for drivers have been added.

Let's close this PR

PR #10972, opened 2020-03-07
fix Conformance testing failure

wy65701436:
1, Return DIGEST_INVALID error in delete manifest instead of NOT_FOUND
2, Disable return 500 in immutable middleware
3, Return empty array in catalog and tags API instead of null

### DIFF
--- a/src/server/registry/manifest.go
+++ b/src/server/registry/manifest.go
@@ -19,6 +19,7 @@ import (
 	"github.com/goharbor/harbor/src/api/repository"
 	"github.com/goharbor/harbor/src/common/utils/log"
 	"github.com/goharbor/harbor/src/internal"
+	ierror "github.com/goharbor/harbor/src/internal/error"
 	serror "github.com/goharbor/harbor/src/server/error"
 	"github.com/goharbor/harbor/src/server/router"
 	"github.com/opencontainers/go-digest"
@@ -51,6 +52,18 @@ func getManifest(w http.ResponseWriter, req *http.Request) {
 func deleteManifest(w http.ResponseWriter, req *http.Request) {
 	repository := router.Param(req.Context(), ":splat")
 	reference := router.Param(req.Context(), ":reference")
+	// v2 doesn't support delete by tag
+	// add parse digest here is to return ErrDigestInvalidFormat before GetByReference throws an NOT_FOUND(404)
+	// Do not add the logic into GetByReference as it's a shared method for PUT/GET/DELETE/Internal call,
+	// and NOT_FOUND satisfy PUT/GET/Internal call.
+	if _, err := digest.Parse(reference); err != nil {
+		switch err {
+		case digest.ErrDigestInvalidFormat:
+			serror.SendError(w, ierror.New(nil).WithCode(ierror.DIGESTINVALID).
+				WithMessage(digest.ErrDigestInvalidFormat.Error()))
+			return
+		}
+	}
 	art, err := artifact.Ctl.GetByReference(req.Context(), repository, reference, nil)
 	if err != nil {
 		serror.SendError(w, err)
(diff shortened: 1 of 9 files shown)


wy65701436:
Summarizing 5 Failures:

[Fail] OCI Distribution Conformance Tests Blob Upload Monolithic [It] POST request with digest and blob should yield a 201 
/root/Conformance/distribution-spec/conformance/03_blob_upload_monolithic_test.go:31

[Fail] OCI Distribution Conformance Tests Blob Upload Monolithic [It] GET request to blob URL from prior request should yield 200 
/root/Conformance/distribution-spec/conformance/03_blob_upload_monolithic_test.go:38

[Fail] OCI Distribution Conformance Tests Blob Upload Chunked [It] Out-of-order blob upload should return 416 
/root/Conformance/distribution-spec/conformance/04_blob_upload_chunked_test.go:28

[Fail] OCI Distribution Conformance Tests Error codes [It] 400 response body should contain OCI-conforming JSON message 
/root/Conformance/distribution-spec/conformance/07_error_codes_test.go:20

[Fail] OCI Distribution Conformance Tests Manifest Delete [It] DELETE request to manifest should return 202, 400, or 405 
/root/Conformance/distribution-spec/conformance/08_manifest_delete_test.go:28

Ran 27 of 27 Specs in 1.353 seconds
FAIL! -- 22 Passed | 5 Failed | 0 Pending | 0 Skipped

src/server/registry/manifest.go
```diff
@@ -51,6 +52,17 @@ func getManifest(w http.ResponseWriter, req *http.Request) {
...
 	reference := router.Param(req.Context(), ":reference")
+	// v2 doesn't support delete by tag
+	// add parse digest here is to return ErrDigestInvalidFormat before GetByReference throws an NOT_FOUND(404)
+	// Do not add the logic into GetByReference as it's a shared method for PUT/GET/DELETE/Internal call,
+	// and NOT_FOUND satisfy PUT/GET/Internal call.
+	if _, err := digest.Parse(reference); err != nil {
+		switch err {
+		case digest.ErrDigestInvalidFormat:
+			serror.SendError(w, ierror.New(nil).WithCode(ierror.DIGESTINVALID).
+				WithMessage("provided digest did not match uploaded content"))
```
ywk253100: The error message is confused here

src/server/registry/manifest.go
```diff
@@ -51,6 +52,17 @@ func getManifest(w http.ResponseWriter, req *http.Request) {
...
 	repository := router.Param(req.Context(), ":splat")
 	reference := router.Param(req.Context(), ":reference")
+	// v2 doesn't support delete by tag
+	// add parse digest here is to return ErrDigestInvalidFormat before GetByReference throws an NOT_FOUND(404)
+	// Do not add the logic into GetByReference as it's a shared method for PUT/GET/DELETE/Internal call,
+	// and NOT_FOUND satisfy PUT/GET/Internal call.
+	if _, err := digest.Parse(reference); err != nil {
+		switch err {
+		case digest.ErrDigestInvalidFormat:
+			serror.SendError(w, ierror.New(nil).WithCode(ierror.DIGESTINVALID).
```
ywk253100: No return

src/server/router/router.go
```diff
@@ -23,7 +23,8 @@ import (
 	"path/filepath"
 )
 
-type contextKeyInput struct{}
+// ContextKeyInput ...
+type ContextKeyInput struct{}
```
ywk253100: Why exposes it out of the package?

wy65701436: will be used in the manifest_test UT

src/server/registry/manifest.go
wy65701436: yes, will add return.

src/server/registry/manifest.go
wy65701436: changed.

PR #10994, opened 2020-03-09
update common error

wy65701436:
1, New support construct with string or err
2, Add Wrap/Errorf method

### DIFF
--- a/src/internal/error/errors.go
+++ b/src/internal/error/errors.go
@@ -32,6 +32,12 @@ func (e *Error) WithCode(code string) *Error {
 	return e
 }
 
+// WithCause ...
+func (e *Error) WithCause(err error) *Error {
+	e.Cause = err
+	return e
+}
+
 // Unwrap ...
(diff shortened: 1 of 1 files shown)


src/internal/error/errors.go
```diff
@@ -101,18 +107,55 @@ const (
...
+	case error:
+		err = in
+	case *Error:
+		err = in.Cause
+	default:
+		err = fmt.Errorf("%v", in)
+	}
+
 	e := &Error{}
 	if err != nil {
```
steven-zou: `err` will Never be nil, right?

src/internal/error/errors.go
```diff
@@ -101,18 +107,55 @@ const (
...
 		e.Cause = err
 		e.Message = err.Error()
```
steven-zou: Will this cause duplicated messages, e.g: "error message: error message"?

src/internal/error/errors.go
wy65701436: sure, will remove the checking here.

PR #11001, opened 2020-03-10
feat(scan): support to scan image index

heww:
1. Add support to scan image index.
2. Get scan log in v2.0 API
3. Add black list to disable scanning for Helm chart.


src/api/scan/base_controller.go
```diff
@@ -110,118 +117,156 @@ func NewController() Controller {
 	}
 }
 
+func (bc *basicController) collectScanningArtifacts(ctx context.Context, r *scanner.Registration, artifact *ar.Artifact) ([]*ar.Artifact, bool, error) {
```
wy65701436: please add some comments

src/api/scan/base_controller.go
```diff
@@ -62,6 +65,8 @@ type jcGetter func() cj.Client
 type basicController struct {
 	// Manage the scan report records
 	manager report.Manager
```
wy65701436: rename it to `reportMrg`

src/api/scan/base_controller.go
```diff
@@ -110,118 +117,156 @@ func NewController() Controller {
...
-	matched := false
-	statusConflict := false
-	for _, ca := range meta.Capabilities {
-		for _, cm := range ca.ConsumesMimeTypes {
-			if cm == artifact.MimeType {
-				matched = true
-				break
-			}
-		}
+	type Param struct {
```
wy65701436: what does `Param` mean?

src/api/scan/base_controller.go
```diff
@@ -110,118 +117,156 @@ func NewController() Controller {
...
+
+		for _, pm := range r.GetProducesMimeTypes(art.ManifestMediaType) {
+			if err = create(ctx, art.Digest, r.UUID, pm, trackID, job.PendingStatus); err != nil {
+				return "", nil, err
+			}
+
+			producesMimes = append(producesMimes, pm)
+		}
+
+		if len(producesMimes) > 0 {
```
wy65701436: why return when `len(producesMimes) > 0`?

src/pkg/scan/vuln/summary.go
```diff
@@ -17,21 +17,76 @@ package vuln
...
+	r.ReportID = sum.ReportID
+	r.StartTime = minTime(sum.StartTime, another.StartTime)
+	r.EndTime = maxTime(sum.EndTime, another.EndTime)
+	r.Duration = r.EndTime.Unix() - r.StartTime.Unix()
+	r.Scanner = sum.Scanner
+	r.TotalCount = sum.TotalCount + another.TotalCount
+	r.CompleteCount = sum.CompleteCount + another.CompleteCount
+	r.CompletePercent = r.CompleteCount * 100 / r.TotalCount
+
+	if sum.Severity.String() != "" && another.Severity.String() != "" {
```
wy65701436: add a compare method in the model of `Severity`?

src/pkg/scan/vuln/summary.go
```diff
@@ -17,21 +17,76 @@ package vuln
...
+		} else {
+			r.Severity = another.Severity
+		}
+	} else if sum.Severity.String() != "" {
+		r.Severity = sum.Severity
+	} else {
+		r.Severity = another.Severity
+	}
+
+	if isRunningStatus(sum.ScanStatus) || isRunningStatus(another.ScanStatus) {
```
wy65701436: why running report can be merged? shouldn't return error?

src/api/scan/base_controller.go
```diff
@@ -110,118 +117,156 @@ func NewController() Controller {
...
 }
 
+func (bc *basicController) collectScanningArtifacts(ctx context.Context, r *scanner.Registration, artifact *ar.Artifact) ([]*ar.Artifact, bool, error) {
+	var (
+		scannable bool
+		artifacts []*ar.Artifact
+	)
+
+	walkFn := func(a *ar.Artifact) error {
+		hasCapability := HasCapability(r, a)
```
wy65701436: `HasCapability` is confusing. can we rename it as ArtScannable or sort of?

heww: There are two scenarios when artifact is scannable.

1. The scanner has capability for the artifact directly, eg the artifact is docker image.
2. The scanner does not has capability for the artifact, but the artifact is image index and the scanner has capability for any one of the artifacts referenced by the image index.

For these reasons,  rename `HasCapability` to `ArtScannable` is not suitable. And this function will be removed and use `HasCapability` of the scanner directly when this issue https://github.com/goharbor/pluggable-scanner-spec/issues/2 be fixed.

src/pkg/scan/vuln/summary.go
heww: This is for the summary overview of the image index.

When there is a running scan job of the artifacts which referenced by the image index, the status of summary overview for the image index is running even that others are all failed.

src/pkg/scan/vuln/summary.go
heww: There is already exist a method `Code` to compare the severity, but not work in this scenario.

When the severity is empty string, it's bigger than all others severity. So in this scenario we should first ensure that the severities are not empty string then compare thire values and chose the bigger one.

src/api/scan/base_controller.go
heww: That means the report placeholders are all created success, so return directly. Otherwise the artifact can not be scanned by the scanner then just create a error report placeholder for the artifact.

src/api/scan/base_controller.go
heww: Out of this PR's scope.

src/api/scan/base_controller.go
heww: No meanings, just a temp values to hold the track id and produces mimes types which will be used to submit scan job later.

src/server/v2.0/handler/scan.go
```diff
@@ -0,0 +1,78 @@
...
+
+	if err := s.scanCtl.Scan(ctx, artifact); err != nil {
+		return s.SendError(ctx, err)
+	}
+
+	return operation.NewScanArtifactAccepted()
+}
+
+func (s *scanAPI) GetReportLog(ctx context.Context, params operation.GetReportLogParams) middleware.Responder {
+	repository := fmt.Sprintf("%s/%s", params.ProjectName, params.RepositoryName)
```
ywk253100: Check permission?

src/pkg/notifier/handler/notification/scan_image_handler.go
```diff
@@ -1,12 +1,16 @@
 package notification
 
 import (
+	"context"
 	"time"
 
+	o "github.com/astaxie/beego/orm"
+	"github.com/goharbor/harbor/src/api/artifact"
```
ywk253100: Currently, the package pkg calls code of package api, this is against our program model. 
Let's move the event handler into `src/api/event` after the PR https://github.com/goharbor/harbor/pull/11010 merged

Maybe we should rename the package api to a better name

heww: Out of the scope this PR. I will submit another PR for this issue.

PR #11008, opened 2020-03-10
feat(ci) enhance govet check performance

bitsf:
enhance govet check performance by remote -L1 param, which will make the check cost from about 3 minutes to 10 seconds


reasonerjt:
Thanks for pointing this out!

PR #11014, opened 2020-03-10
Fixing broken link

techgnosis:
Found the correct link in the Create Projects page, which makes sense


bitsf:
use `git commit -s` to do signing, or the DCO check will fail

wy65701436:
close it as it's fixed in https://github.com/goharbor/website.

PR #11017, opened 2020-03-11
change some doc about harbor.yml.tmpl

bitsf:


reasonerjt:
Could you open an issue and wait for @xaleeks to confirm about how to update the doc?
@bitsf 
I don't think we should update the 1.10 doc folder if it's for master

bitsf:
open #11053 to track

PR #11037, opened 2020-03-12
Escapse the repository name in the link header returned in response

ywk253100:
Escapse the repository name in the link header returned in response

### DIFF
--- a/src/server/v2.0/handler/base.go
+++ b/src/server/v2.0/handler/base.go
@@ -121,26 +121,44 @@ func (b *BaseAPI) BuildQuery(ctx context.Context, query *string, pageNumber, pag
 
 // Links return Links based on the provided pagination information
 func (b *BaseAPI) Links(ctx context.Context, u *url.URL, total, pageNumber, pageSize int64) internal.Links {
-	url := *u
+	ul := *u
+	// try to unescape the repository name which contains escaped slashes
+	if escapedPath, err := url.PathUnescape(ul.Path); err == nil {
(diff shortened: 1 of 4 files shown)


src/server/v2.0/handler/base.go
```diff
@@ -121,26 +121,44 @@ func (b *BaseAPI) BuildQuery(ctx context.Context, query *string, pageNumber, pag
...
 	var links internal.Links
 	// prev
 	if pageNumber > 1 && (pageNumber-1)*pageSize < total {
-		q := url.Query()
+		q := ul.Query()
 		q.Set("page", strconv.FormatInt(pageNumber-1, 10))
-		url.RawQuery = q.Encode()
+		ul.RawQuery = q.Encode()
+		// try to unescape the query
+		if escapedQuery, err := url.QueryUnescape(ul.RawQuery); err == nil {
```
heww: Why do we need to unescape query string?

ywk253100: pkg/q/builder.Build() can only handle the unescaped characters

heww: I think the `q` is unescaped automic before pass it the `pkg/q/builder.Build`, is it right?

And we need to response the same values as they are in the query string, we only change the `page` paramater.

ywk253100: In line 136, we encode the query string, the special characters used in our query(~,[],etc) are all encoded, so here decode them first

PR #11038, opened 2020-03-12
Repair the count usage during the upgrading 

ywk253100:
As the count quota is against artifact rather than tag in 2.0, the count usage should be recalculated


heww:
Quota refreshing depent on the chartmuseum when helm chart enabled.

Issue #10064, opened 2019-12-02
Make sure the replication work as expected after introducing the OCI work


ywk253100:
- [ ] Replicate index/manifest list

PR #11046, opened 2020-03-12
Replicate tag deletion between Harbor instances

ywk253100:
This commit introduces the tag deletion as a new capability for registry adapters, and currently only Harbor supports it

### DIFF
--- a/src/replication/adapter/harbor/image_registry.go
+++ b/src/replication/adapter/harbor/image_registry.go
@@ -28,7 +28,7 @@ import (
 	"strings"
 )
 
-func (a *adapter) FetchImages(filters []*model.Filter) ([]*model.Resource, error) {
+func (a *adapter) FetchArtifacts(filters []*model.Filter) ([]*model.Resource, error) {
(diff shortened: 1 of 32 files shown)


src/replication/adapter/harbor/image_registry.go
```diff
@@ -168,3 +168,10 @@ func (a *adapter) listArtifacts(repository string, filters []*model.Filter) ([]*
 	}
 	return filter.DoFilterArtifacts(arts, filters)
 }
+
+func (a *adapter) DeleteTag(repository, tag string) error {
+	project, repository := utils.ParseRepository(repository)
+	url := fmt.Sprintf("%s/api/%s/projects/%s/repositories/%s/artifacts/%s/tags/%s",
```
steven-zou: This is not a good practice, it should be defined in a unified model linked with the API version.

Something like:
   `v2().Tag(t string).URL()` etc.

ywk253100:
Fixes https://github.com/goharbor/harbor/issues/10064

PR #11060, opened 2020-03-13
Set LDAP groupname when PopulateGroup

### LINKED ISSUE #10940: Groups which are auto-created have the DN name instead of the attribute value which we have provided in "LDAP Group GID"
**Expected behavior** 

We have set under the Harbor LDAP Configuration field "LDAP Group GID" to "cn"

Example:
DN: CN=MA_PKS-Admin,OU=MA,OU=Application,OU=Groups,OU=Global,OU=Production,DC=user,DC=com

Group name in harbor - MA_PKS-Admin( expected)

The description of "LDAP Group GID" says: "the group in Harbor is named with this attribute by default". And this is what would actually expect.

 **actual behavior:**

Plenty of groups which are not named like below are added under group in harbor.

 "CN=MA_PKS-Admin,OU=GA,OU=Application,OU=Groups,OU=Global,OU=Production,DC=User,DC=com". (basically group name == DN)

However the majority of groups that are around, got "autocreated" in the wrong way with "group name" == "DN", which is not as we configured.

**Versions:**

Harbor 1.9.3-build.2

**Additional context:**

If we add a group manually via "New Group" -> (Import LDAP Group) we can specify LDAP DN and Name. In this way we can set "DN" and "name" accordingly / correctly and everything is fine.

Harbor Authentication configured against LDAP (Opsman: LDAP - Login Harbor with LDAP Users).
Authentication is working fine generally.

stonezdj:
Search ldap group name with default ldap group attribute name
fixes #10940

### DIFF
--- a/src/common/utils/ldap/ldap.go
+++ b/src/common/utils/ldap/ldap.go
@@ -388,7 +388,7 @@ func (session *Session) searchGroup(baseDN, filter, groupName, groupNameAttribut
 		var group models.LdapGroup
 		group.GroupDN = ldapEntry.DN
 		for _, attr := range ldapEntry.Attributes {
-			// OpenLdap sometimes contain leading space in useranme
+			// OpenLdap sometimes contain leading space in username
 			val := strings.TrimSpace(attr.Values[0])
 			log.Debugf("Current ldap entry attr name: %s\n", attr.Name)
 			switch strings.ToLower(attr.Name) {
--- a/src/common/utils/ldap/ldap_test.go
+++ b/src/common/utils/ldap/ldap_test.go
@@ -318,6 +318,12 @@ func TestSession_SearchGroupByDN(t *testing.T) {
 		LdapGroupNameAttribute: "cn",
 		LdapGroupSearchScope:   2,
 	}
+	ldapGroupConfig2 := models.LdapGroupConf{
+		LdapGroupBaseDN:        "ou=group,dc=example,dc=com",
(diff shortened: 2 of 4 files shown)


src/common/utils/ldap/ldap.go
```diff
@@ -401,6 +401,21 @@ func (session *Session) searchGroup(baseDN, filter, groupName, groupNameAttribut
...
+		return groupDN, nil
+	}
+	groups, err := session.SearchGroupByDN(groupDN)
+	if err != nil {
+		return "", err
+	}
+	if len(groups) == 0 {
+		return "", fmt.Errorf("failed to retrieve group by DN: %v", groupDN)
+	}
+	return groups[0].GroupName, nil
```
reasonerjt, line 416: How does `LdapGroupNameAttribute` take effect here?

stonezdj: It is used in https://github.com/goharbor/harbor/blob/3c345f92fd49c0300f6a54b3b1bf3c99a5cd7bc6/src/common/utils/ldap/ldap.go#L368

reasonerjt: But in `SearchGroupByDN` you've already populated the groupname into each entry?  It feels like there are redundant funcs in this file.

stonezdj: Remove the SearchGroupNameByDN

PR #11072, opened 2020-03-13
add notification middleware

wy65701436:
the notification is for send out the event after DB transaction complete.
It's safe to send hook as this middleware is after transaction in the response path.

### DIFF
--- a/src/server/middleware/notification/notification.go
+++ b/src/server/middleware/notification/notification.go
@@ -1,0 +1,52 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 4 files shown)


src/server/middleware/notification/notification.go
```diff
@@ -0,0 +1,56 @@
...
+	for e := es.Front(); e != nil; e = e.Next() {
+		evt.BuildAndPublish(e.Value.(evt.Metadata))
+	}
+	return
+}
+
+// Middleware sends the notification after transaction success
+func Middleware(skippers ...middleware.Skipper) func(http.Handler) http.Handler {
+	return middleware.New(func(w http.ResponseWriter, r *http.Request, next http.Handler) {
+		res, ok := w.(*internal.ResponseBuffer)
```
ywk253100: Use internal.ResponseRecorder instead as no need to buffer the response

wy65701436: fixed.

PR #11084, opened 2020-03-16
feat(middleware,vulnerable): add image index checking for vulnerability prevention

heww:
1. Skip vulnerability prevention checking when artifact is not
scannable.
2. Skip vulnerability prevention checking when image index and its type is IMAGE or CNAB.
3. Skip vulnerability prevention checking when the artifact is pulling
by the scanner.
4. Change `HasCapability` from blacklist to whitelist.


renmaosheng:
is the 'skip' an option user to choose? I thought we are by default enforce the check of CVE, if users want to ignore, he needs to add the whitelist.

heww:
> is the 'skip' an option user to choose? I thought we are by default enforce the check of CVE, if users want to ignore, he needs to add the whitelist.

It is not an option user to choose. There are two scenarios to skip the vulnerability prevention checking.

1. We will skip the checking for the artifacts which are not supported by the scanner because there will never exist reports for these artifacts. Eg, the scanning of `CHART` artifacts is not supported by the `trivy`, when users pull helm chart, the checking will be skiped so that users can pull successfully.

2. We also will skip the checking for the artifacts which are image index and they are scannable but not supported directly by the scanner. Eg, there is an artifact named `library/redis:latest` which reference there artifacts: `library/redis:amd64`, `library/redis:s390x` and `library/redis:ppc64le`.  The severity of `library/redis:s390x`, `library/redis:ppc64le` is `Hight`, the severity of `library/redis:amd64` is 'Low`, so the severity of `library/redis:latest` will be `Hight`.  The severity for the project vulnerability prevention is `Hight`, then users pull `library/redis:latest` in the amd64 platform, the checking will be skipped for `library/redis:latest` so that users can pull real `library/redis:amd64` image.

The CVE whitelist will be used at the vulnerability prevention checking stage when the checking is not skipped.

src/api/scan/checker.go
```diff
@@ -86,9 +116,10 @@ func (c *checker) IsScannable(ctx context.Context, art *artifact.Artifact) (bool
 // HasCapability returns true when scanner has capability for the artifact
 // See https://github.com/goharbor/pluggable-scanner-spec/issues/2 to get more info
 func HasCapability(r *models.Registration, a *artifact.Artifact) bool {
-	if a.Type == "CHART" || a.Type == "UNKNOWN" {
-		return false
+	// use whitelist here because
```
reasonerjt: because what?

src/server/middleware/vulnerable/vulnerable.go
```diff
@@ -1,125 +1,138 @@
...
-						log.Infof("Vulnerable policy check: scannerPull CVE %s", cve)
-					}
-				}
+	return middleware.BeforeRequest(func(r *http.Request) error {
+		ctx := r.Context()
+
+		logger := log.G(ctx).WithFields(log.Fields{"middleware": "vulnerable"})
+
+		path := r.URL.EscapedPath()
+		art, err := artifactController.GetByReference(ctx, distribution.ParseName(path), distribution.ParseReference(path), nil)
```
reasonerjt: Can we use artifact Info as the single source of truth of a request?

src/api/scan/checker.go
```diff
@@ -24,6 +25,9 @@ import (
 
 // Checker checker which can check that the artifact is scannable
 type Checker interface {
+	// HasCapability returns true when the artifact supported by the scanner
```
reasonerjt: I find the capability .vs. scannable a little confusing.

src/server/middleware/vulnerable/vulnerable.go
```diff
@@ -1,125 +1,145 @@
...
+		summary, ok := rawSummary.(*vuln.NativeReportSummary)
+		if !ok {
+			return fmt.Errorf("report summary is invalid")
+		}
+
+		hasCapability, err := checker.HasCapability(ctx, art)
+		if err != nil {
+			return err // this will never happened because the artifact is scannable
+		}
+		if !hasCapability {
```
reasonerjt: I would rather use `art.HasChildren` to figure if it's an index.  Even if a scanner declares it has the capability of scanning an index, we should still treat the index the same way when handling it's scan result in the middleware.

reasonerjt:
Approved but with slightly different opinion:

Instead of using `Scannability` and `Capability` we should check if the artifact is an index, disregarding how scanner declares.

src/api/scan/checker.go
heww: Removed

src/server/middleware/vulnerable/vulnerable.go
heww: Changed to use artifact info

PR #11085, opened 2020-03-16
move notification handles and events metadata into api

wy65701436:
1, enable audit logs for notifications
2, move the handler and meatadata into API
3, use the notification middleware to send out notification


src/api/event/handler/chart/chart.go
```diff
@@ -1,28 +1,39 @@
-package notification
```
ywk253100: Seems all handlers under the package `src/notifier/handler/notification` are webhook related? If they are, how about moving all of then into a separated package?

wy65701436: only http and slack are left, all the others are move into ./api/event/handlers

src/api/event/handler/init.go
```diff
@@ -0,0 +1,51 @@
...
+	notifier.Subscribe(event.TopicDeleteArtifact, &auditlog.Handler{})
+	notifier.Subscribe(event.TopicCreateProject, &auditlog.Handler{})
+	notifier.Subscribe(event.TopicDeleteProject, &auditlog.Handler{})
+	notifier.Subscribe(event.TopicDeleteRepository, &auditlog.Handler{})
+	notifier.Subscribe(event.TopicCreateTag, &auditlog.Handler{})
+	notifier.Subscribe(event.TopicDeleteTag, &auditlog.Handler{})
+}
+
+// InitSupportedEvent ....
+func InitSupportedEvent() {
```
ywk253100: Move this into webhook related package

wy65701436: move it into api

PR #11091, opened 2020-03-16
Add create/delete tag API test case

ywk253100:
Add create/delete tag API test case

### DIFF
--- a/tests/apitests/python/test_create_delete_tag.py
+++ b/tests/apitests/python/test_create_delete_tag.py
@@ -1,0 +1,99 @@
+from __future__ import absolute_import
+
+
+import unittest
+
+import library.repository
+import library.docker_api
+from library.base import _assert_status_code
+from testutils import ADMIN_CLIENT
+from testutils import harbor_server
+
+from testutils import TEARDOWN
+from library.project import Project
+from library.user import User
+from library.repository import Repository
+from library.artifact import Artifact
+from library.repository import push_image_to_project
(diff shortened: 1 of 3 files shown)


tests/apitests/python/test_create_delete_tag.py
```diff
@@ -0,0 +1,99 @@
...
+        """
+        #1. Create a new user(UA);
+        TestProjects.user_id, user_name = self.user.create_user(user_password = self.user_password, **ADMIN_CLIENT)
+
+        TestProjects.USER_CLIENT=dict(with_tag = True, endpoint = self.url, username = user_name, password = self.user_password)
+
+        #2. Create a new project(PA) by user(UA);
+        TestProjects.project_id, TestProjects.project_name = self.project.create_project(metadata = {"public": "false"}, **TestProjects.USER_CLIENT)
+
+        #3. Push an image(IA) to Harbor by docker successfully;
```
renmaosheng: do we need to add such test for helm and cnab as well? thanks.

renmaosheng:
please link the issue# to be fixed. :)

ywk253100:
Fixes https://github.com/goharbor/harbor/issues/10764

tests/apitests/python/test_create_delete_tag.py
ywk253100: The way how the tag API works is same for all kinds of artifacts, if it works for image, it should work for other artifacts as well.

PR #11095, opened 2020-03-16
feat(log): track request id in the log message

heww:
Track request id in the log message.

### DIFF
--- a/src/common/utils/log/logger.go
+++ b/src/common/utils/log/logger.go
@@ -19,6 +19,7 @@ import (
 	"io"
 	"os"
 	"runtime"
+	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -31,18 +32,21 @@ const srcSeparator = "harbor" + string(os.PathSeparator) + "src"
 func init() {
 	lvl := os.Getenv("LOG_LEVEL")
 	if len(lvl) == 0 {
-		logger.SetLevel(InfoLevel)
+		logger.setLevel(InfoLevel)
 		return
 	}
 
 	level, err := parseLevel(lvl)
 	if err != nil {
-		logger.SetLevel(InfoLevel)
+		logger.setLevel(InfoLevel)
 		return
 	}
 
-	logger.SetLevel(level)
+	logger.setLevel(level)
 }
+
+// Fields type alias to map[string]interface{}
+type Fields = map[string]interface{}
 
 // Logger provides a struct with fields that describe the details of logger.
 type Logger struct {
@@ -51,7 +55,9 @@ type Logger struct {
 	lvl       Level
 	callDepth int
 	skipLine  bool
-	mu        sync.Mutex
+	fields    map[string]interface{}
+	fieldsStr string
+	mu        *sync.Mutex // ptr here to share one sync.Mutex for clone method
 }
 
 // New returns a customized Logger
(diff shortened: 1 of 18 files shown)


src/common/utils/log/logger.go
```diff
@@ -66,11 +76,20 @@ func New(out io.Writer, fmtter Formatter, lvl Level, options ...interface{}) *Lo
 			depth = d
 		}
 	}
+
+	mu, ok := writerMutexMap[out]
```
ywk253100: What is this used for?

src/common/utils/log/logger.go
```diff
@@ -79,6 +98,58 @@ func DefaultLogger() *Logger {
...
+		}
+		for key, value := range fields {
+			copyFields[key] = value
+		}
+
+		sortedKeys := make([]string, 0, len(copyFields))
+		for key := range copyFields {
+			sortedKeys = append(sortedKeys, key)
+		}
+		sort.Strings(sortedKeys)
```
ywk253100: After sorting the fields will not be the same sequence with the original one, is this expected?
FuncC(with field c)->funcB(with filed b)->FuncA(with filed a), the log will be: [a,b,c]xxx, right?

heww: 1. In Golang, iterations over maps are random, sorting here can make the fileds output is ordered in log

2. `FuncC(with field c)->funcB(with filed b)->FuncA(with filed a), the log will be: [a,b,c]xxx, right?` No, the log will be [c]xxx in `FuncC`, [b]xxx in `funcB` and [a]xxx in `FuncA`. `log.G(ctx)` returns the logger saved in the ctx (if not found, returns the default one), and `logger.WithFields` does not modify the logger itself, it returns new logger.

src/common/utils/log/logger.go
heww: This is used to lock the output writer, when multi logger use the same output writer.

En, this is may not needed for the is a root logger in the system (the default logger), and other loggers are clone from this. I will remove the `writerMutexMap`.

src/common/utils/log/logger.go
```diff
@@ -79,6 +88,58 @@ func DefaultLogger() *Logger {
...
+// WithDepth ...
+func (l *Logger) WithDepth(depth int) *Logger {
+	r := l.clone()
+	r.callDepth = depth
+
+	return r
+}
+
+// WithFields ...
+func (l *Logger) WithFields(fields Fields) *Logger {
```
reasonerjt: Can we make Withxx and Setxxx consistent?  Or protect Setxxx from being called by mistake?

src/server/middleware/log/log.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/goharbor/harbor/src/common/utils/log"
+	"github.com/goharbor/harbor/src/server/middleware"
+)
+
+// Middleware middleware which add logger to context
+func Middleware() func(http.Handler) http.Handler {
+	return middleware.New(func(w http.ResponseWriter, r *http.Request, next http.Handler) {
+		rid := r.Header.Get("X-Request-ID")
+		if rid != "" {
+			ctx := log.WithLogger(r.Context(), log.L.WithFields(log.Fields{"requestID": rid}))
```
reasonerjt: Would be helpful for debugging if we can log a message with request id and basic information of the request?

heww: debug info added

src/common/utils/log/logger.go
heww: SetXxx changed to private method to make logger is immutable.

PR #11097, opened 2020-03-16
update project logs api to v2.0

wy65701436:
use audit log api to get project logs

### DIFF
--- a/api/v2.0/swagger.yaml
+++ b/api/v2.0/swagger.yaml
@@ -615,6 +615,39 @@ paths:
           $ref: '#/responses/401'
         '403':
           $ref: '#/responses/403'
+  /projects/{project_name}/logs:
+    get:
+      summary: Get recent logs of the projects
+      description: Get recent logs of the projects
+      tags:
+        - project
+      operationId: getLogs
+      parameters:
+        - $ref: '#/parameters/projectName'
+        - $ref: '#/parameters/requestId'
+        - $ref: '#/parameters/query'
+        - $ref: '#/parameters/page'
+        - $ref: '#/parameters/pageSize'
+      responses:
+        '200':
+          description: Success
+          headers:
+            X-Total-Count:
+              description: The total count of auditlogs
+              type: integer
+            Link:
+              description: Link refers to the previous page and next page
+              type: string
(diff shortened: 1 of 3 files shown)


api/v2.0/swagger.yaml
```diff
@@ -615,6 +615,68 @@ paths:
...
+      description: Get recent logs of the projects
+      tags:
+        - project
+      operationId: getLogs
+      parameters:
+        - $ref: '#/parameters/projectName'
+        - $ref: '#/parameters/requestId'
+        - $ref: '#/parameters/page'
+        - $ref: '#/parameters/pageSize'
+        - name: username
```
ywk253100: Use the new query instead?

src/server/v2.0/handler/project.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"context"
+	"github.com/go-openapi/runtime/middleware"
+	api_project "github.com/goharbor/harbor/src/api/project"
+	"github.com/goharbor/harbor/src/common/rbac"
+	"github.com/goharbor/harbor/src/pkg/audit"
+	"github.com/goharbor/harbor/src/pkg/q"
+	"github.com/goharbor/harbor/src/server/v2.0/models"
+	"github.com/goharbor/harbor/src/server/v2.0/restapi/operations/project"
+	operation "github.com/goharbor/harbor/src/server/v2.0/restapi/operations/project"
```
heww: duplicated

api/v2.0/swagger.yaml
renmaosheng: it this like audit log? not just push, delete, also the pull event will be recorded? calling event would be more appropriate.

wy65701436: update to Q in query

src/server/v2.0/handler/project.go
wy65701436: updated.

api/v2.0/swagger.yaml
ywk253100: So the query strings username/resource/resource_type/etc can be removed?

Issue #11051, opened 2020-03-12
[Scanner] Internal server error in the project scanner manage page caused by an unhealthy scanner


qinshaoxuan:
**Expected behavior and actual behavior:**
I get internal server error in the project scanner manage page when the selected scanner is unhealthy 

**Expected behavior**:
The project admin should know the selected scanner is unhealthy
![image](https://user-images.githubusercontent.com/17852010/76513604-86091880-6491-11ea-9dfe-3926eb4d4229.png)

**Actual behavior**:
no scanner info
![image](https://user-images.githubusercontent.com/17852010/76513781-c5376980-6491-11ea-8e6a-cabca246d421.png)

Scanner unhealthy message added by #9807 is also not work  because `Core` returns `500 Internal Server Error`

**Steps to reproduce the problem:**
1. Harbor admin adds an external scanner
2. A project selects this scanner
3. Stop the external scanner 
4. Get internal server error when open the project scanner manage page

**Versions:**
- harbor version: 1.10.1

**Additional context:**

- **Log files:**
```
2020-03-09T16:30:07+08:00 [ERROR] [/pkg/scan/api/scanner/base_controller.go:229]: api controller: get project scanner: scanner controller: ping: v1 client: get metadata: Get http://<external scanner host>/api/v1/metadata: dial tcp <external scanner host>: connect: connection refused
2020-03-09T16:30:07+08:00 [ERROR] [/common/api/base.go:233]: scanner API: get project scanners: scanner controller: ping: v1 client: get metadata: Get http://<external scanner host>/api/v1/metadata: dial tcp <external scanner host>: connect: connection refused
2020-03-09T16:30:07+08:00 [ERROR] [/common/api/base.go:73]: GET /api/projects/17/scanner failed with error: {"code":500,"message":"Internal Server Error"}
2020/03/09 15:53:47.667 [D] [filter.go:59]  |  <client ip> | 500 |   10.76245ms|   match| GET      /api/projects/17/scanner   r:/api/projects/:pid([0-9]+)/scanner
```

It looks like that function `GetRegistrationByProject` should not return err when `Ping` return err.
https://github.com/goharbor/harbor/blob/6a25e6b2c67d460c3dc7f22b0d234a394341be96/src/api/scanner/base_controller.go#L227-L243

Can I create a PR to fix it?

Issue #9788, opened 2019-11-07
[Scanner] Unhealthy Scanner still be listed in the project scanner list.


wy65701436:
Get internal error when to switch to an unhealthy in the project.

Please give a reasonable message as the project admin could not know the scanner is unhealthy.

steven-zou:
@AllForNothing 

Before adding to the project, check the health of the selected scanner. If failed, print error message at top of the adding wizard

steven-zou:
@AllForNothing 

As there might be unhealthy ones in the candidate list, we can let the user choose and set it as the project level scanner. However, if the scanner is unhealthy, we should alert the user with a warning message.

PR #11106, opened 2020-03-17
Fix bug when scanner is unhealthy

qinshaoxuan:
The function GetRegistrationByProject should not return err when Ping
return err.  The return value 'registration' has 'Health' field which
shows the scanner health status.

Resolves: #11051
See also: #9788, #9807

Signed-off-by: qinshaoxuan <qinshaoxuan@baidu.com>


qinshaoxuan:
PTAL @danielpacak @steven-zou

PR #11113, opened 2020-03-17
Use new query model to get audit logs

wy65701436:
leverage the query builder to build query, remove the old style query string


src/server/v2.0/handler/auditlog.go
```diff
@@ -26,28 +25,9 @@ func (a *auditlogAPI) ListAuditLogs(ctx context.Context, params auditlog.ListAud
 	// if !a.HasPermission(ctx, rbac.ActionList, rbac.ResourceLog) {
```
ywk253100: Will we do the permission check in another PR?

wy65701436: yes  to resolve ToDo items

PR #11118, opened 2020-03-18
drop table access log in migration

wy65701436:
Use the audit log instead, the access log table should be dropped after migration


src/server/v2.0/route/legacy.go
```diff
@@ -62,8 +62,6 @@ func registerLegacyRoutes() {
 	beego.Router("/api/"+version+"/system/CVEWhitelist", &api.SysCVEWhitelistAPI{}, "get:Get;put:Put")
 	beego.Router("/api/"+version+"/system/oidc/ping", &api.OIDCAPI{}, "post:Ping")
 
-	beego.Router("/api/"+version+"/logs", &api.LogAPI{})
```
ywk253100: Please remove the router of getting project log as well

wy65701436: done

PR #11123, opened 2020-03-18
feat(quota,notification): notification for quota exceeded and warning

heww:
1. Send notification when request resources exceeded the quota limitation.
2. Send notification when quota usage exceeded 85% of the quota limitation.

### DIFF
--- a/src/pkg/quota/models/quota.go
+++ b/src/pkg/quota/models/quota.go
@@ -16,6 +16,7 @@ package models
 
 import (
 	"encoding/json"
+	"fmt"
 	"time"
 
 	"github.com/goharbor/harbor/src/pkg/quota/driver"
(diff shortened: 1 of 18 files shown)


src/pkg/quota/models/quota.go
```diff
@@ -86,3 +87,39 @@ func (q *Quota) SetUsed(used types.ResourceList) *Quota {
 
 	return q
 }
+
+// GetWarningResources returns resource names which exceeded the warning percent
+func (q *Quota) GetWarningResources(warningPercent int) ([]types.ResourceName, error) {
```
wy65701436: should only send out the current project quota message, excludes others.

heww: The `Quota` belongs the project, `GetWarningResources` is ok here.

Issue #10532, opened 2020-01-19
Rewrite the filters with middleware mechinism


ywk253100:
We introduce the middleware which can cover the filter part, so let's rewrite the filters with middleware

PR #11144, opened 2020-03-19
Rewrite the filters with middleware mechinism

ywk253100:
Fixes 10532,rewrite the filters with middleware mechinism


src/server/middleware/security/security.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+)
+
+// security context generator
+type generator interface {
+	Generate(req *http.Request) security.Context
+}
+
+// Middleware returns a security context middleware that populates the security context into the request context
+func Middleware() middleware.Middleware {
```
heww: Suggest to change the return type of `Middleware` func from `middleware.Middleware` to `func(http.Handler) http.Handler` so that we don't need do type conversion when run this middleware in beego.

src/core/middlewares/middlewares.go
```diff
@@ -74,6 +76,8 @@ func legacyAPISkipper(r *http.Request) bool {
 // MiddleWares returns global middlewares
 func MiddleWares() []beego.MiddleWare {
 	return []beego.MiddleWare{
+		beego.MiddleWare(session.Middleware()),
+		beego.MiddleWare(security.Middleware()),
 		csrf.Middleware(),
 		requestid.Middleware(),
 		log.Middleware(),
```
heww: Please help to move `requestid` and `log` in the front, thank you. 
The suggestion order is `requestid`, `log`, `session`, `csrf` and `security`.

src/core/auth/authenticator.go
```diff
@@ -212,18 +212,16 @@ func SearchGroup(groupKey string) (*models.UserGroup, error) {
 // SearchAndOnBoardUser ... Search user and OnBoard user, if user exist, return the ID of current user.
 func SearchAndOnBoardUser(username string) (int, error) {
 	user, err := SearchUser(username)
+	if err != nil {
```
reasonerjt: THANK YOU!

src/internal/context.go
```diff
@@ -78,3 +80,33 @@ func GetArtifactInfo(ctx context.Context) (art ArtifactInfo) {
...
 	return
 }
+
+// WithAuthMode returns a context with auth mode set
+func WithAuthMode(ctx context.Context, mode string) context.Context {
+	return setToContext(ctx, contextKeyAuthMode, mode)
+}
+
+// GetAuthMode gets the auth mode from the context
+func GetAuthMode(ctx context.Context) string {
```
reasonerjt: I just realized it will be slightly better if we add a type assertion in all `GetXXX` funcs in this file (not only the ones you added).

It would be better like 
```
mode, _ := value.(string)
```
such that there won't be panic.

src/server/middleware/security/auth_proxy.go
```diff
@@ -0,0 +1,104 @@
...
+
+type authProxy struct{}
+
+func (a *authProxy) Generate(req *http.Request) security.Context {
+	log := log.G(req.Context())
+	if internal.GetAuthMode(req.Context()) != common.HTTPAuth {
+		return nil
+	}
+	// only support docker login
+	if req.URL.Path != "/service/token" {
```
reasonerjt: Oh this should be updated in the `v2` -> basic auth refactor , I think it should check if it has /v2 prefix

src/core/middlewares/middlewares.go
```diff
@@ -74,9 +76,11 @@ func legacyAPISkipper(r *http.Request) bool {
 // MiddleWares returns global middlewares
 func MiddleWares() []beego.MiddleWare {
 	return []beego.MiddleWare{
-		csrf.Middleware(),
 		requestid.Middleware(),
 		log.Middleware(),
+		session.Middleware(),
+		csrf.Middleware(),
```
reasonerjt: csrf middleware can reuse the `internal.GetCarrySession`

reasonerjt:
some minor comments that can be addressed after this PR.

ywk253100:
Fixes https://github.com/goharbor/harbor/issues/10532

PR #11148, opened 2020-03-19
Unescape tags query when to list artifact

wy65701436:
The query string is encoded by UI, and we have to unescape the "=" in "q=tag=nil",
otherwise, the query doesn't work, and returns 400

### DIFF
--- a/src/pkg/q/builder.go
+++ b/src/pkg/q/builder.go
@@ -16,7 +16,9 @@ package q
 
 import (
 	"fmt"
+	"github.com/goharbor/harbor/src/common/utils/log"
 	ierror "github.com/goharbor/harbor/src/internal/error"
+	"net/url"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


src/server/v2.0/handler/base.go
```diff
@@ -101,13 +101,19 @@ func (b *BaseAPI) RequireProjectAccess(ctx context.Context, projectIDOrName inte
...
 func (b *BaseAPI) BuildQuery(ctx context.Context, query *string, pageNumber, pageSize *int64) (*q.Query, error) {
+
 	var (
 		qs string
 		pn int64
 		ps int64
 	)
 	if query != nil {
 		qs = *query
+		if escapedQuery, err := url.QueryUnescape(qs); err == nil {
```
ywk253100: How about put the logic into the function `pkg/q.Build()`, it's the entrypoint for parsing query string

heww: Realy very strange that before use the query , we needed to unescape it. Maybe there is something wrong.

wy65701436: Generally, the pattern "a=b" in the query is ok and do not need be escaped.
But we use "q=a=b" in the query, "a=b" is the value, thus UI will encode it. @heww

wy65701436: @ywk move it to `pkg/q.Build()`

src/pkg/q/builder.go
```diff
@@ -38,6 +40,12 @@ func Build(q string, pageNumber, pageSize int64) (*Query, error) {
 	if len(q) == 0 {
 		return query, nil
 	}
+	// try to escaped the 'q=tags%3Dnil' when to filter tags.
+	if escapedQuery, err := url.QueryUnescape(q); err == nil {
```
heww: `escapedQuery` => `unescapedQuery` ?

wy65701436: ok, will fix it.

PR #11154, opened 2020-03-20
Add permission check for audit logs API

wy65701436:
add a base method to require system admin permission

### DIFF
--- a/src/server/v2.0/handler/auditlog.go
+++ b/src/server/v2.0/handler/auditlog.go
@@ -2,8 +2,14 @@ package handler
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"github.com/go-openapi/runtime/middleware"
+	"github.com/goharbor/harbor/src/common/rbac"
+	"github.com/goharbor/harbor/src/common/security"
+	ierror "github.com/goharbor/harbor/src/internal/error"
 	"github.com/goharbor/harbor/src/pkg/audit"
+	"github.com/goharbor/harbor/src/pkg/q"
 	"github.com/goharbor/harbor/src/server/v2.0/models"
 	"github.com/goharbor/harbor/src/server/v2.0/restapi/operations/auditlog"
 	operation "github.com/goharbor/harbor/src/server/v2.0/restapi/operations/auditlog"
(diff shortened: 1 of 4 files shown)


src/server/v2.0/handler/auditlog.go
```diff
@@ -21,10 +21,9 @@ type auditlogAPI struct {
 }
 
 func (a *auditlogAPI) ListAuditLogs(ctx context.Context, params auditlog.ListAuditLogsParams) middleware.Responder {
-	// ToDo enable permission check
-	// if !a.HasPermission(ctx, rbac.ActionList, rbac.ResourceLog) {
-	//	return a.SendError(ctx, ierror.ForbiddenError(nil))
-	// }
+	if err := a.RequireSysAdmin(ctx); err != nil {
```
ywk253100: Non-admin users also need to list the related logs

wy65701436: why no-admin users can see the global logs? He can know the operation of other projects?
It's not make sense

ywk253100: Login as non-admin users, there is an item to show all logs related with the user on the left panel on portal

wy65701436: ok,I got it. So only verify the authN?

ywk253100: Please refer to the implement of 1.10

src/server/v2.0/handler/base.go
```diff
@@ -99,6 +99,18 @@ func (b *BaseAPI) RequireProjectAccess(ctx context.Context, projectIDOrName inte
...
 }
 
+// RequireSysAdmin checks the system admin permission according to the security context
+func (b *BaseAPI) RequireSysAdmin(ctx context.Context) error {
+	secCtx, ok := security.FromContext(ctx)
+	if !ok {
+		return ierror.UnauthorizedError(errors.New("security context not found"))
+	}
+	if !secCtx.IsSysAdmin() {
+		return ierror.ForbiddenError(nil).WithMessage(secCtx.GetUsername())
```
ywk253100: Check 401 error

src/server/v2.0/handler/auditlog.go
```diff
@@ -21,14 +26,34 @@ type auditlogAPI struct {
...
 	}
+
+	if !secCtx.IsSysAdmin() {
+		// ToDo remove the dependency on GetMyProjects()
+		projects, err := secCtx.GetMyProjects()
+		if err != nil {
+			return a.SendError(ctx, fmt.Errorf(
+				"failed to get projects of user %s: %v", secCtx.GetUsername(), err))
+		}
+		ids := []int64{}
```
ywk253100: We can use `q.OrList` directly here, so that the function `setPIDsQuery()` isn't needed anymore

src/server/v2.0/handler/auditlog.go
```diff
@@ -21,14 +26,34 @@ type auditlogAPI struct {
...
+		for _, project := range projects {
+			if a.HasProjectPermission(ctx, project.ProjectID, rbac.ActionList, rbac.ResourceLog) {
+				ids = append(ids, project.ProjectID)
+			}
+		}
+		query.Keywords["ProjectIDs"] = ids
```
ywk253100: Let's use `ProjectID` instead. The `ProjectIDs` isn't a property of `Log` model. The API consumers know nothing about it if we don't document it

src/server/v2.0/handler/base.go
wy65701436: done

src/server/v2.0/handler/base.go
```diff
@@ -94,11 +94,26 @@ func (b *BaseAPI) RequireProjectAccess(ctx context.Context, projectIDOrName inte
 		return ierror.UnauthorizedError(errors.New("security context not found"))
 	}
 	if !secCtx.IsAuthenticated() {
-		return ierror.UnauthorizedError(nil)
+		return ierror.UnauthorizedError(nil).WithMessage(secCtx.GetUsername())
```
ywk253100: If the request isn't authenticated, the `secCtx.GetUsername()` must return nil

src/server/v2.0/handler/base.go
```diff
@@ -94,11 +94,26 @@ func (b *BaseAPI) RequireProjectAccess(ctx context.Context, projectIDOrName inte
...
 }
 
+// RequireSysAdmin checks the system admin permission according to the security context
+func (b *BaseAPI) RequireSysAdmin(ctx context.Context) error {
+	secCtx, ok := security.FromContext(ctx)
+	if !ok {
+		return ierror.UnauthorizedError(errors.New("security context not found"))
+	}
+	if !secCtx.IsAuthenticated() {
+		return ierror.UnauthorizedError(errors.New("security context not found"))
```
ywk253100: The error message here confuses users

PR #11156, opened 2020-03-20
Migrate config to harbor2

ninjadq:
- move migrator to prepare
- add migrate script to harbor2

### DIFF
--- a/make/photon/prepare/commands/migrate.py
+++ b/make/photon/prepare/commands/migrate.py
@@ -3,8 +3,7 @@ import os, sys, importlib, shutil, glob
 import click
 
 from utils.misc import get_realpath
-from migration.utils import read_conf, search
-
+from utils.migration import read_conf, search
 
 @click.command()
 @click.option('-i', '--input', 'input_', default='', help="The path of original config file")
--- a/make/photon/prepare/commands/migrate.py
+++ b/make/photon/prepare/commands/migrate.py
@@ -9,13 +9,12 @@ from migration.utils import read_conf, to_module_path, search
 @click.command()
 @click.option('-i', '--input', 'input_', default='', help="The path of original config file")
 @click.option('-o', '--output', default='', help="the path of output config file")
-@click.option('-t', '--target', default='1.10.0', help="target version of input path")
+@click.option('-t', '--target', default='2.0.0', help="target version of input path")
 def migrate(input_, output, target):
+    if not output:
+        output = input_
     input_path = get_realpath(input_)
-    if output == '':
-        output_path = input_path
-    else:
-        output_path = get_realpath(output)
+    output_path = get_realpath(output)
 
     configs = read_conf(input_path)
 
@@ -29,11 +28,10 @@ def migrate(input_, output, target):
     for m in search(input_version, target):
         current_output_path = "harbor.yml.{}.tmp".format(m.revision)
         click.echo("migrating to version {}".format(m.revision))
-        print("migrating to version {}".format(m.revision))
         m.migrate(current_input_path, current_output_path)
         current_input_path = current_output_path
     shutil.copy(current_input_path, output_path)
-    print("Written new values to %s" % output_path)
+    click.echo("Written new values to {}".format(output))
     for tmp_f in glob.glob("harbor.yml.*.tmp"):
         os.remove(tmp_f)
 
--- a/make/photon/prepare/commands/migrate.py
+++ b/make/photon/prepare/commands/migrate.py
@@ -1,0 +1,39 @@
+import os, sys, importlib, shutil, glob
+
+import click
+
+from utils.misc import get_realpath
+from migration.utils import read_conf, to_module_path, search
+
+
+@click.command()
+@click.option('-i', '--input', 'input_', default='', help="The path of original config file")
+@click.option('-o', '--output', default='', help="the path of output config file")
+@click.option('-t', '--target', default='1.10.0', help="target version of input path")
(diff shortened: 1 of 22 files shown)


make/photon/prepare/commands/migrate.py
```diff
@@ -0,0 +1,37 @@
...
+    configs = read_conf(input_path)
+
+    input_version = configs.get('_version')
+    
+    if input_version == target:
+        click.echo("Version of input harbor.yml is identical to target {}, no need to upgrade".format(input_version))
+        sys.exit(0)
+
+    current_input_path = input_path
+    for m in search(input_version, target):
```
reasonerjt: does if fail if search returns empty list?

make/photon/prepare/migration/utils.py
```diff
@@ -0,0 +1,51 @@
...
+def to_module_path(ver):
+    return "migration.versions.{}".format(ver.replace(".","_"))
+
+def search(input_ver: str, target_ver: str) -> list:
+    """
+    Search accept a input version and the target version.
+    Returns the module of migrations in the upgrade path
+    """
+    def helper():
+        nonlocal basedir
```
reasonerjt: what's the benefit here to use `nonlocal` rather than pass these as parms?

the code outside `helper` does not rely on the change `helper` makes to these variables.

make/photon/prepare/migration/utils.py
```diff
@@ -0,0 +1,51 @@
...
+        nonlocal cur_target
+        while True:
+            module_path = to_module_path(cur_target)
+            if os.path.isdir(os.path.join(basedir, 'versions', cur_target.replace(".","_"))):
+                module = importlib.import_module(module_path)
+                yield module
+                if module.revision == input_ver:
+                    return
+                elif module.down_revision is not None:
```
reasonerjt: if the down_revision is set wrong there can be an infinite loop.

make/photon/prepare/migration/versions/2_0_0/harbor.yml.jinja
```diff
@@ -0,0 +1,387 @@
...
+  local:
+    # Log files are rotated log_rotate_count times before being removed. If count is 0, old versions are removed rather than rotated.
+    rotate_count: {{ log.local.rotate_count }}
+    # Log files are rotated only if they grow bigger than log_rotate_size bytes. If size is followed by k, the size is assumed to be in kilobytes.
+    # If the M is used, the size is in megabytes, and if G is used, the size is in gigabytes. So size 100, size 100k, size 100M and size 100G
+    # are all valid.
+    rotate_size: {{ log.local.rotate_size }}
+    # The directory on your host that store log
+    location: {{ log.local.location }}
+  {% if log.external_endpoint is defined%}
```
reasonerjt: space after `defined`?

make/photon/prepare/migration/versions/2_0_0/harbor.yml.jinja
```diff
@@ -0,0 +1,387 @@
...
+# https related config
+# https:
+#   # https port for harbor, default is 443
+#   port: 443
+#   # The path of cert and key files for nginx
+#   certificate: /your/certificate/path
+#   private_key: /your/private/key/path
+{% endif %}
+
+{% if internal_tls is defined %}
```
reasonerjt: is this chunk needed?

make/photon/prepare/commands/migrate.py
ninjadq: yes, I edited some code of `search`, it will return [] if empty

make/photon/prepare/migration/utils.py
ninjadq: This is a closure, the inner function can access the variables outside, in python3 if you wanna access outside variables you must declare it as `nonlocal`.

make/photon/prepare/migration/versions/2_0_0/harbor.yml.jinja
ninjadq: It's OK, I set `lstrip_blocks=True` in jinja env

ninjadq: yes, I'll add space after it for consistent. But space won't  affect the function

make/photon/prepare/migration/versions/2_0_0/harbor.yml.jinja
ninjadq: For convenience, you can just copy these codes for the next version.

reasonerjt: we should put this in next version! this is not needed in xx-> 2.0 migration right?

make/photon/prepare/utils/migration.py
```diff
@@ -0,0 +1,44 @@
...
+    upgrade_path, visited = deque(), set()
+    while True:
+        module_path = _to_module_path(target_ver)
+        visited.add(target_ver)  # mark current version for loop finding
+        if os.path.isdir(os.path.join(MIGRATION_BASE_DIR, 'version_{}'.format(target_ver.replace(".","_")))):
+            module = importlib.import_module(module_path)
+            if module.revision == input_ver:    # migration path found
+                break
+            elif module.down_revision is None:  # migration path not found
+                raise Exception('no migration path found')
```
reasonerjt: Not a blocker, but it would be helpful if you print more context in the errors.

For example what's the current migration path, and what 's the input/target version.

Same for other errors thrown in the `search` func.

PR #11184, opened 2020-03-23
feat(cicd) fix build base image

bitsf:
fix some cicd build base image problems

### DIFF
--- a/tests/resources/Harbor-Util.robot
+++ b/tests/resources/Harbor-Util.robot
@@ -56,7 +56,7 @@ Package Harbor Offline
     Log To Console  \nStart Docker Daemon
     Start Docker Daemon Locally
     Log To Console  \n\nmake package_offline BASEIMAGETAG=%{Harbor_Build_Base_Tag} NPM_REGISTRY=%{NPM_REGISTRY} VERSIONTAG=%{Harbor_Assets_Version} PKGVERSIONTAG=%{Harbor_Package_Version} NOTARYFLAG=${with_notary} CLAIRFLAG=${with_clair} MIGRATORFLAG=${with_migrator} CHARTFLAG=${with_chartmuseum} TRIVYFLAG=${with_trivy} HTTPPROXY=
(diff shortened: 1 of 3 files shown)


tests/resources/Harbor-Util.robot
```diff
@@ -93,7 +93,7 @@ Switch To LDAP
 Enable Notary Client
     ${rc}  ${output}=  Run And Return Rc And Output  rm -rf ~/.docker/
     Log  ${rc}
-    ${rc}  ${output}=  Run And Return Rc and Output  curl -o /notary_ca.crt -s -k -X GET --header 'Accept: application/json' -u 'admin:Harbor12345' 'https://${ip}/api/v2.0/systeminfo/getcert'
+    ${rc}  ${output}=  Run And Return Rc and Output  curl -o /notary_ca.crt -s -k -X GET --header 'Accept: application/json' -u 'admin:Harbor12345' 'https://${ip}/api/systeminfo/getcert'
```
reasonerjt: I don't think the url is right.

tests/resources/Harbor-Util.robot
```diff
@@ -52,21 +52,21 @@ Down Harbor
     Should Be Equal As Integers  ${rc}  0
 
 Package Harbor Offline
-    [Arguments]  ${with_notary}=true  ${with_clair}=true  ${with_migrator}=true  ${with_chartmuseum}=true  ${with_trivy}=true
+    [Arguments]  ${with_notary}=true  ${with_clair}=true  ${with_migrator}=true  ${with_chartmuseum}=true
```
reasonerjt: why remove trivy?

tests/resources/Harbor-Util.robot
wy65701436: we should use `v2.0` in the path, otherwise, it should be 404

bitsf: oh, my bad when fix conflict.

tests/resources/Harbor-Util.robot
bitsf: oh, my bad when fix conflict.

PR #11248, opened 2020-03-24
add clean untagged blobs in gc job

wy65701436:
Fix#11190, delete all of non-referenced blobs of each project in GC job, thun the quota
can be released.

### DIFF
--- a/src/jobservice/job/impl/gc/garbage_collection.go
+++ b/src/jobservice/job/impl/gc/garbage_collection.go
@@ -16,9 +16,12 @@ package gc
 
 import (
 	"fmt"
+	"github.com/goharbor/harbor/src/common/models"
 	"github.com/goharbor/harbor/src/controller/artifact"
+	"github.com/goharbor/harbor/src/controller/project"
 	"github.com/goharbor/harbor/src/lib/q"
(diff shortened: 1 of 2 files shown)


src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -240,3 +246,54 @@ func (gc *GarbageCollector) deleteCandidates(ctx job.Context) error {
...
+				params.Pagination.Page++
+			}
+
+		}()
+
+		return ch
+	}(50)
+
+	for project := range projects {
+		all, err := gc.blobMgr.List(ctx.SystemContext(), blob.ListParams{
```
heww: This may have performance problem when there are many many blobs in the project.

ywk253100: Is it possible to introduce a prune method in blob ctl/mgr to do the same work?

wy65701436: @heww compares with the performance of GC, it can be ignored?

wy65701436: @ywk253100 you mean combind these two methods?

Issue #11189, opened 2020-03-24
Adding/removing tags should NOT cause the whole detail page refreshed


steven-zou:
Only refresh the tag list is enough

Refreshing the whole is not a good experience.

AllForNothing:
For now, no axtra api to refresh the tag only,  need to refresh the whole artifact to refresh the tag

Issue #11218, opened 2020-03-24
Message tips for creating duplicated tags under the repository is not very clear


steven-zou:
<img width="1161" alt="Screen Shot 2020-03-24 at 14 46 00" src="https://user-images.githubusercontent.com/5753287/77396205-42da6e00-6dde-11ea-908c-2d771ad32cb8.png">

Repo number ID may be replaced by the name/current repository?

jwangyangls:
This message is returned from the backend，

ywk253100:
UI has added logic to check the existence of the name under the same repository

PR #11336, opened 2020-03-27
Add API to list tags under the specific repository

ywk253100:
Add API to list tags under the specific repository


ywk253100:
This API can be called by UI to fix https://github.com/goharbor/harbor/issues/11189 and https://github.com/goharbor/harbor/issues/11218

PR #11339, opened 2020-03-28
Add "order by" clause to avoid the duplicat rows

ywk253100:
Add "order by" clause to avoid the duplicat rows: https://www.postgresql.org/docs/9.6/queries-limit.html


bitsf:
UT case need fix too

PR #11352, opened 2020-03-30
[Fix] some harbor 2.0 UI bug

### LINKED ISSUE #9926: The header name "start time" is confusing.
The GC log is not sorted by id or time

image

version:1.9.1

### LINKED ISSUE #11034: No feedback when testing incorrect LDAP connection settings
I found this bug while watching a customer at the keyboard so I don't have screenshots.
We were following these steps
https://github.com/goharbor/harbor/blob/master/docs/1.10/administration/configure-authentication/ldap-auth.md

When you are setting up Harbor LDAP integration, there is a button that says "Test LDAP Server"

**Expected Behavior**: If the settings are wrong, the user is notified the settings are wrong. If the settings are correct, the user is notified that the settings are correct.

**Actual Behavior**: If the settings are wrong, nothing happens. If the settings are correct, the user is notified that the settings are correct.

This means the customer has no idea if they are doing it right until they do it right. With no feedback when the settings are incorrect, the customer might think the settings are correct and waste time trying to use the LDAP integration that is in fact incorrect.

Go to Administration -> Configuration -> Authentication
Select LDAP
Enter in your LDAP connection settings into the fields
Push "Test LDAP Server"

**Versions:**
- harbor version: 1.10.1
- docker engine version: unrelated
- docker-compose version: unrelated

**Additional context:**
This is purely a UX bug. We eventually got the LDAP settings correct which caused us to be informed by the UI that we got the settings correct. If we never got them correct we would have always been wondering if the settings were correct or not.

### LINKED ISSUE #11199: The data shown in the tag tooltip is not consistent with the tag view in the detail page
Screen Shot 2020-03-24 at 11 20 14

Suggest keeping consistent to avoid confusion

### LINKED ISSUE #11209: Add meaningful artifact kind prefix before the artifact digest in navigation bar/page title
To easily understand.

Screen Shot 2020-03-24 at 12 36 48

### LINKED ISSUE #11230: Some UX issues in the detail page of an artifact
1. There should be some element on UI to indicate the type of this artifact, especially for a manifest list:

Screen Shot 2020-03-24 at 4 27 38 PM

2. I find the term  `extra attributes` `additions` a little confusing.
3. I find the content of `extra attributes` less readable than `tags` and `additions`.  How about we put tags at top and `additions` below `tags` then `extra attributes`
Screen Shot 2020-03-24 at 4 31 27 PM

### LINKED ISSUE #11235: Partially data extracted from the CNAB bundle published by porter
Screen Shot 2020-03-24 at 16 46 06

This data should be recursively parsed.

### LINKED ISSUE #11312: harbor 2.0 UI 翻译问题，不含Tag
有的地方使用的是关键字 “不含Tag”，有的地方是“无Tag”，术语不一致容易造成混淆
image
image

建议使用 “无Tag”

另外术语Artifacts的大小写也比较随意，空格也不一致

jwangyangls:
1.fix #11312
2.fix #11235
3.fix #11230
4.fix #11209
5.fix #11199
6.fix #11034
7.fix #9926


src/portal/src/app/all-pipes/select-artifact-icon/select-artifact-icon.pipe.ts
```diff
@@ -0,0 +1,18 @@
...
+  name: 'selectArtifactIcon'
+})
+export class SelectArtifactIconPipe implements PipeTransform {
+
+  transform(value: string, ...args: any[]): any {
+
+      if (artifactImages.some(image => image === value)) {
+        return 'images/artifact-' + value.toLowerCase() + '.svg';
+      } else {
+        return 'images/artifact-default.svg';
```
AllForNothing: please use const for the values

src/portal/src/i18n/lang/zh-cn-lang.json
```diff
@@ -1024,7 +1024,7 @@
         "FILTER_FOR_ARTIFACTS": "Filter Artifacts",
         "ADDITIONS": "其他",
         "ANNOTATION": "注解",
-        "EXTRA_PROPERTIES": "额外属性",
+        "OVERVIEW": "总览属性",
```
AllForNothing: 属性总览？

PR #11387, opened 2020-04-01
feat(cicd) parameterize docker base image and external url

bitsf:
In order to make harbor build more flexible, also parameterize docker base image and external url

### DIFF
--- a/make/photon/Makefile
+++ b/make/photon/Makefile
@@ -13,12 +13,14 @@ SRCPATH=./src
 TOOLSPATH=$(CURDIR)/tools
 SEDCMD=$(shell which sed)
 WGET=$(shell which wget)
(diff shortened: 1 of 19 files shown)


make/photon/Makefile
```diff
@@ -29,6 +31,15 @@ JOBSERVICEBINARYNAME=harbor_jobservice
 # photon dockerfile
 DOCKERFILEPATH=$(MAKEPATH)/photon
 
+# dependency binaries
```
wy65701436, line 34: why define them twice?

bitsf: 因为两个 makefile 都有被单独调用的用法，如果只定义一个的话，会有报错

PR #11396, opened 2020-04-02
Add "_self" suffix after reposiotry name for all APIs that contain repository

ywk253100:
Add "_self" suffix for all APIs that contain repository to avoid path conflict


reasonerjt:
This looks horrible. 

Could you help me understand why there are still conflicts after introducing `_self` to repository api?

Issue #11271, opened 2020-03-25
[Public project]Re-push a deleted Chart via Helm V3 fails by 401


wy65701436:
- push one chart with helm v3
- delete it in harbor
- Re-push it

After that, I get a 401. 

```
The push refers to repository [10.193.21.1/library/dummy-chart]
ref:     10.193.21.1/library/dummy-chart:0.1.0
digest:  3e94f4f36ce87478ab22cbb459aa2f93f45b0f0635407cd81cb5262d7c509a3b
size:    3.2 KiB
name:    dummy-chart
version: 0.1.0
Error: failed commit on ref "manifest-sha256:a8c04b6060af0ac7bfa315203ba917371808adfc57630014563efe85f3bd56b2": unexpected status: 401 Unauthorized
```

The root cause is probably that Harbor returns 200 to helm for the manifest head request, and helm client doesn't take auth header in the following PUT request.

Same thing happens on `docker app push`

reasonerjt:
This is related to #11266 

Current response in `/v2` breaks some existing CLI tools when they push artifact to public projects.

reasonerjt:
Fixed in #11406

Verified in build ` v2.0.0-42801b76`

Issue #11266, opened 2020-03-25
Cannot pull public images using docker client without login


ywk253100:
```
yinw-a01:cnab-to-oci yinw$ docker logout harbor.10.160.195.133.xip.io
Removing login credentials for harbor.10.160.195.133.xip.io
WARNING: could not erase credentials: error erasing credentials - err: exit status 1, out: `error erasing credentials - err: exit status 1, out: `The specified item could not be found in the keychain.``
yinw-a01:cnab-to-oci yinw$ docker pull harbor.10.160.195.133.xip.io/library/hello-world:latest
Error response from daemon: Get https://harbor.10.160.195.133.xip.io/v2/library/hello-world/manifests/latest: no basic auth credentials
```

Seems that when docker client trying to pull the image, it sends a `/v2/` request first, Harbor responses with status code 401 and header `www-authenticate: Basic realm="harbor"`.
Docker client finds no basic auth credential in local host, it reports an error directly.

reasonerjt:
I'll double check docker's code.

Worst case if it always call `GET /v2` to decide whether to attach credentials to the subsequent requests, we'll have to return the token realm to the client, and decode the token in the middleware.

reasonerjt:
Fixed in #11406 
Verified in build ` v2.0.0-42801b76`

Issue #11188, opened 2020-03-24
Auth middleware should treat Head request as action Push not Pull


wy65701436:
We found that, during docker app to push a CNAB, it trys to head a untagged manifest. 

1,  Against Harbor, it returns a 200 for a public one, but 404 for a private one.
2, Against Docker-hub, it always returns a 404.

Docker client will define whether to use the auth header in the following request base on the response code.

reasonerjt:
If you you treat is as push you see problem when pulling the public artifacts

reasonerjt:
@wy65701436 @heww can we open an issue to `docker app` ?
Based on my understanding I don't think they are doing the right thing?

ywk253100:
Seems docker app useds cnab-to-oci as a library, cnab-to-oci has the same issue

wy65701436:
@reasonerjt I don't think it's an issue of docker app as we found that Helm v3 has the same issue. 
https://github.com/goharbor/harbor/issues/11271

reasonerjt:
Fixed by #11406 
Verified `docker app push` to a public project in build ` v2.0.0-42801b76`

PR #11406, opened 2020-04-02
Reenable token auth for cli

reasonerjt:
Docker CLI fails if it's not logged in upon seeing "basic" realm challenging while pinging the "/v2" endpoint. (#11266)
Some CLI will send HEAD to artifact endpoint before pushing (#11188)(#11271)

To fix such problems, this commit re-introduce the token auth flow to the CLIs.

For a request to "/v2/xxx" with no "Authoirzation" header, the v2_auth middleware populates the
"Www-Authenticate" header to redirect it to token endpoint with proper
requested scope.

It also adds security context to based on the content of the JWT which has the claims of the registry.
So a request from CLI carrying a token signed by the "/service/token" will have proper permissions.

### DIFF
--- a/src/server/middleware/v2auth/access.go
+++ b/src/server/middleware/v2auth/access.go
@@ -1,0 +1,101 @@
+package v2auth
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/goharbor/harbor/src/common/rbac"
+	"github.com/goharbor/harbor/src/lib"
+	"github.com/goharbor/harbor/src/lib/log"
+	"github.com/goharbor/harbor/src/server/middleware"
+)
+
+type target int
+
+const (
+	login target = iota
+	catalog
+	repository
+)
+
+func (t target) String() string {
+	return []string{"login", "catalog", "repository"}[t]
+}
+
+type access struct {
+	target target
+	name   string
+	action rbac.Action
+}
+
+func (a access) scopeStr(ctx context.Context) string {
+	logger := log.G(ctx)
+	if a.target != repository {
+		// Currently we do not support providing a token to list catalog
+		return ""
+	}
+	act := ""
+	if a.action == rbac.ActionPull {
+		act = "pull"
+	} else if a.action == rbac.ActionPush {
+		act = "pull,push"
+	} else if a.action == rbac.ActionDelete {
+		act = "delete"
+	} else {
+		logger.Warningf("Invalid action in access: %s, returning empty scope", a.action)
+		return ""
+	}
+	return fmt.Sprintf("repository:%s:%s", a.name, act)
+}
+
+func getAction(req *http.Request) rbac.Action {
+	actions := map[string]rbac.Action{
+		http.MethodPost:   rbac.ActionPush,
+		http.MethodPatch:  rbac.ActionPush,
+		http.MethodPut:    rbac.ActionPush,
+		http.MethodGet:    rbac.ActionPull,
+		http.MethodHead:   rbac.ActionPull,
+		http.MethodDelete: rbac.ActionDelete,
+	}
+	if action, ok := actions[req.Method]; ok {
+		return action
+	}
+	return ""
+}
+
+func accessList(req *http.Request) []access {
+	l := make([]access, 0, 4)
+	if req.URL.Path == "/v2/" {
+		l = append(l, access{
+			target: login,
+		})
+		return l
+	}
+	if len(middleware.V2CatalogURLRe.FindStringSubmatch(req.URL.Path)) == 1 {
+		l = append(l, access{
+			target: catalog,
+		})
+		return l
+	}
+	none := lib.ArtifactInfo{}
+	if a := lib.GetArtifactInfo(req.Context()); a != none {
+		action := getAction(req)
+		if action == "" {
+			return l
+		}
+		l = append(l, access{
+			target: repository,
+			name:   a.Repository,
+			action: action,
+		})
+		if req.Method == http.MethodPost && a.BlobMountRepository != "" { // need pull access for blob mount
+			l = append(l, access{
+				target: repository,
+				name:   a.BlobMountRepository,
+				action: rbac.ActionPull,
+			})
+		}
+	}
+	return l
+}
--- a/src/server/middleware/v2auth/auth.go
+++ b/src/server/middleware/v2auth/auth.go
@@ -17,59 +17,54 @@ package v2auth
 import (
 	"fmt"
 	"net/http"
+	"strings"
 	"sync"
 
 	"github.com/goharbor/harbor/src/common/rbac"
 	"github.com/goharbor/harbor/src/common/security"
 	"github.com/goharbor/harbor/src/core/config"
 	"github.com/goharbor/harbor/src/core/promgr"
-	"github.com/goharbor/harbor/src/lib"
+	"github.com/goharbor/harbor/src/core/service/token"
 	"github.com/goharbor/harbor/src/lib/errors"
 	"github.com/goharbor/harbor/src/lib/log"
 	serror "github.com/goharbor/harbor/src/server/error"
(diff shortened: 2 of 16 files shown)


src/common/security/internaltoken/context.go
```diff
@@ -0,0 +1,112 @@
...
+			logger.Debugf("Unable to get project name from resource %s, drop the access", ac.Name)
+			continue
+		}
+		actionMap := make(map[types.Action]struct{})
+		for _, a := range ac.Actions {
+			if a == "pull" || a == "*" {
+				actionMap[rbac.ActionPull] = struct{}{}
+			}
+			if a == "push" || a == "*" {
+				actionMap[rbac.ActionPush] = struct{}{}
```
ywk253100, line 101: Shouldn't we cover the action for deletion as well?

src/server/middleware/v2auth/access.go
```diff
@@ -0,0 +1,104 @@
...
+		logger.Warningf("Invalid action in access: %s, returning empty scope", a.action)
+		return ""
+	}
+	return fmt.Sprintf("repository:%s:%s", a.name, act)
+}
+
+func getAction(req *http.Request) rbac.Action {
+	pushActions := map[string]struct{}{
+		http.MethodPost:   {},
+		http.MethodDelete: {},
```
ywk253100, line 53: Is here any security issue?
If I'm a developer, I can get a bearer token with push permission, and then I can use the token to delete manifest/blobs?

src/common/security/internaltoken/context.go
```diff
@@ -0,0 +1,112 @@
+package internaltoken
```
heww, line 1: v2token ?

src/common/security/internaltoken/context.go
reasonerjt: I think not because it's for CLI only

src/common/security/internaltoken/context.go
reasonerjt: whatever makes you happy...

src/server/middleware/v2auth/access.go
reasonerjt: good point, I thought delete will be modified to "push" but it's not...

src/common/security/internaltoken/context.go
reasonerjt: I don't think we should do it now because the `service/token` does not sign token with `delete` scope and the `*` in `creator.go` equals push and pull

src/server/middleware/v2auth/access.go
```diff
@@ -0,0 +1,109 @@
...
+
+	if req.Method == http.MethodDelete {
+		return rbac.ActionDelete
+	}
+	if _, ok := pushActions[req.Method]; ok {
+		return rbac.ActionPush
+	}
+	if _, ok := pullActions[req.Method]; ok {
+		return rbac.ActionPull
+	}
```
heww, line 71:
```go
actions := map[string]rbac.Action{
	http.MethodPost:  rbac.ActionPush,
	http.MethodPatch: rbac.ActionPush,
	http.MethodPut:   rbac.ActionPush,
	http.MethodGet:   rbac.ActionPull,
	http.MethodHead:  rbac.ActionPull,
	http.MethodDelete:  rbac.ActionDelete,
}

if action, ok := actions[http.Method]; ok {
	return action
}

return ""
```

reasonerjt: definitely.

src/server/middleware/v2auth/auth.go
```diff
@@ -17,59 +17,54 @@ package v2auth
...
 		}
-		resource := rbac.NewProjectNamespace(pid).Resource(rbac.ResourceRepository)
-		if !securityCtx.Can(action, resource) {
-			return fmt.Errorf("unauthorized to access repository: %s, action: %s", a.Repository, action)
+		if a.target == catalog && !securityCtx.IsSysAdmin() {
+			return getChallenge(req, al), fmt.Errorf("unauthorized to list catalog")
 		}
-		if req.Method == http.MethodPost && a.BlobMountProjectName != "" { // check permission for the source of blob mount
-			pid, err := rc.projectID(a.BlobMountProjectName)
+		if a.target == repository && req.Header.Get(authHeader) == "" && req.Method == http.MethodHead { // make sure 401 is returned for CLI HEAD, see #11271
```
heww, line 53: This made the response for `Head` and `Get` inconformity, should always return 401 when `Authorization` header is empty.

src/server/middleware/v2auth/auth.go
```diff
@@ -17,59 +17,54 @@ package v2auth
...
+			pid, err := rc.projectID(pn)
 			if err != nil {
-				return err
+				return "", err
 			}
 			resource := rbac.NewProjectNamespace(pid).Resource(rbac.ResourceRepository)
-			if !securityCtx.Can(rbac.ActionPull, resource) {
-				return fmt.Errorf("unauthorized to access repository from which to mount blob: %s, action: %s", a.BlobMountRepository, rbac.ActionPull)
+			if !securityCtx.Can(a.action, resource) {
+				return getChallenge(req, al), fmt.Errorf("unauthorized to access repository: %s, action: %s", a.name, a.action)
```
heww, line 63: When `Authorization` header is provied, maybe 403 is appropriate here.

src/server/middleware/v2auth/auth.go
```diff
@@ -83,25 +78,31 @@ func (rc *reqChecker) projectID(name string) (int64, error) {
 	return p.ProjectID, nil
 }
 
-func getAction(req *http.Request) rbac.Action {
-	pushActions := map[string]struct{}{
-		http.MethodPost:   {},
-		http.MethodDelete: {},
-		http.MethodPatch:  {},
-		http.MethodPut:    {},
+func getChallenge(req *http.Request, accessList []access) string {
```
heww, line 81: `Basic` and  `Bearer` should returned together in the challenge so that cli will first try to use `Basic` auth, then `Bearer` auth.

I think most time the `Basic` auth will be ok, and the anonymous access will use the `Bearer` auth. (This needs to be confirmed)

reasonerjt: Return basic auth will fail #11266

src/server/middleware/v2auth/auth.go
reasonerjt: 401 is compatible for v2 registry, otherwise the automation test won't pass.

src/server/middleware/v2auth/auth.go
reasonerjt: This is intentional b/c I don't think `HEAD` a public resource should return 401, but I have no choice due to the implementation in CLI.

reasonerjt: I will follow up with helm and docker app to try to fix it, so they won't HEAD the resource to determine whether it should carry auth header for PUT

src/server/middleware/v2auth/auth.go
heww: 403 Forbidden is also compatible for spec, https://github.com/opencontainers/distribution-spec/blob/master/spec.md#on-failure-access-denied

reasonerjt: By v2 registry I mean the docker registry, it returns 401 when token has insufficient scope.  So, if you return 403, docker will not handle it as before, it will print less readable output and the automation test will fail.

src/server/middleware/v2auth/auth.go
heww: Agree. For client, eg docker cli, it use access `/v2` before push manifest list.

src/server/middleware/v2auth/auth.go
heww: I got it. It's terrible of the implemention of docker registry.

PR #11428, opened 2020-04-03
Fix container unhealth

ninjadq:
- Fix unhealthy status of harbor containers
- Add harbor_ca to jobservice trust store


reasonerjt:
I wish to suggest we avoid adding excessive ENV var that is merely for healthcheck ,

Is this `HEALTHCHECK` part of OCI standard  I think we may consider to remove it at some point.  B/C there are too many way for health check (internal healthcheck API, probes in k8s..)

For a quick fix will this work:
`HEALTHCHECK CMD curl --fail -s http://127.0.0.1:8080/api/v2.0/ping ||
curl  -k https://127.0.0.1:8443/api/v2.0/ping || exit 1`

PR #11445, opened 2020-04-06
deprecate quota count on artifact

### LINKED ISSUE #11241: harbor 2.0 quota count is hard to define now
when docker push an image, such as busybox:latest, because there're actually several sub artifacts in it in order to distinguish different OS/arch, so the quota will count this busybox:latest as N.
This may be difficult for user to define quota with count in future.

wy65701436:
Fixes #11241

1, remove count quota from quota manager
2, remove count in DB scheme
3, remove UI relates on quota
4, update UT, API test and UI UT.


wy65701436:
image

image

image

image

image

wy65701436:
@xaleeks please refer to the screenshot about UI change

Issue #11466, opened 2020-04-07
Parse cron spec with UTC option


steven-zou:
`Schedule.Next(Now().UTC())`

steven-zou:
Fixed in #11473

PR #11473, opened 2020-04-07
fix[jobservice]:enqueue job with UTC

steven-zou:
- schedule the periodical jobs following the UTC timezone
  - e.g: 5 10 10 * * * means run jobs at UTC time 10:10:05 everyday
- fix issue #11466


wy65701436:
please think about how to notify the cron change to the end-user.

PR #11479, opened 2020-04-07
Fix replication bug

ywk253100:
Remove the URL replacing logic temporarily to make replication work and will introduce a new solution for the hairpin issue


ywk253100:
Open issue https://github.com/goharbor/harbor/issues/11480 to track the following work

Issue #11492, opened 2020-04-08
New issue in introduced into the UT case of enqueuer of jobservice by previous bug fix


steven-zou:
The time of CRON spec created for the UT case may cause overflow problems.

```log
2020-04-07T10:58:22.8153087Z 10:58:22 go test -race -v -cover -coverprofile=profile.tmp -coverpkg github.com/goharbor/harbor/src/jobservice/period,github.com/goharbor/harbor/src/jobservice/common/rds,github.com/goharbor/harbor/src/jobservice/common/utils,github.com/goharbor/harbor/src/jobservice/job,github.com/goharbor/harbor/src/jobservice/lcm,github.com/goharbor/harbor/src/jobservice/logger github.com/goharbor/harbor/src/jobservice/period
2020-04-07T10:58:24.9757276Z 10:58:24 === RUN TestSchedulerTestSuite
2020-04-07T10:58:24.9861838Z 10:58:24 2020/04/07 10:58:24 Scheduler: periodic enqueuer is started
2020-04-07T10:58:24.9873468Z 10:58:24 === RUN TestSchedulerTestSuite/TestScheduler
2020-04-07T10:58:24.9903433Z 10:58:24 2020/04/07 10:58:24 Scheduled execution for periodic job DEMO:fake_policy at 1586257110
2020-04-07T10:58:24.9924169Z 10:58:24 2020/04/07 10:58:24 Scheduled execution for periodic job DEMO:fake_policy at 1586257130
2020-04-07T10:58:24.9939167Z 10:58:24 2020/04/07 10:58:24 Load 0 periodic job policies
2020-04-07T10:58:24.9950979Z 10:58:24 2020/04/07 10:58:24 Scheduler: periodic enqueuer is stopped
2020-04-07T10:58:24.9963448Z 10:58:24 --- PASS: TestSchedulerTestSuite (0.02s)
2020-04-07T10:58:24.9976646Z 10:58:24 --- PASS: TestSchedulerTestSuite/TestScheduler (0.01s)
2020-04-07T10:58:25.0009142Z 10:58:25 === RUN TestEnqueuerTestSuite
2020-04-07T10:58:25.0020796Z 10:58:25 2020/04/07 10:58:24 Scheduler: periodic enqueuer is started
2020-04-07T10:58:25.0032156Z 10:58:25 === RUN TestEnqueuerTestSuite/TestEnqueuer
[...]

PR #11493, opened 2020-04-08
fix(js UT cases):fix the cron spec bug

steven-zou:
- eliminate the time of cron spec  overflow issue in the UT case
- fix issue #11492 

### DIFF
--- a/src/jobservice/period/enqueuer_test.go
+++ b/src/jobservice/period/enqueuer_test.go
@@ -123,13 +123,14 @@ func (suite *EnqueuerTestSuite) prepare() {
 	now := time.Now()
 	minute := now.Minute()
 
-	coreSpec := fmt.Sprintf("0-59 %d-%d * * * *", minute, minute+2)
+	// Make sure we at least schedule jobs in the 3 minutes cycle
+	cronSpec := fmt.Sprintf("0-59 %d,%d,%d * * * *", minute, (minute+1)%60, (minute+2)%60)
 
 	// Prepare one
 	p := &Policy{
 		ID:       "fake_policy",
 		JobName:  job.SampleJob,
-		CronSpec: coreSpec,
+		CronSpec: cronSpec,
 	}
 	rawData, err := p.Serialize()
 	assert.Nil(suite.T(), err, "prepare data: nil error expected but got %s", err)


src/jobservice/period/enqueuer_test.go
```diff
@@ -120,16 +120,26 @@ func (suite *EnqueuerTestSuite) TestEnqueuer() {
...
-	coreSpec := fmt.Sprintf("0-59 %d-%d * * * *", minute, minute+2)
+	// Make sure we at least schedule jobs in the 3 minutes cycle
+	switch minute {
+	case 58:
+		cronSpec = "0-59 58,59,0 * * * *"
+	case 59:
+		cronSpec = "0-59 59,0,1 * * * *"
+	default:
+		cronSpec = fmt.Sprintf("0-59 %d-%d * * * *", minute, minute+2)
+	}
```
bitsf, line 136: I think we should avoid such random test, it will make the code coverage unstable.
It's better to use a for loop to verify each fake time minutes

src/jobservice/period/enqueuer_test.go
```diff
@@ -120,16 +120,26 @@ func (suite *EnqueuerTestSuite) TestEnqueuer() {
 }
 
 func (suite *EnqueuerTestSuite) prepare() {
+	var cronSpec string
+
 	now := time.Now()
 	minute := now.Minute()
 
-	coreSpec := fmt.Sprintf("0-59 %d-%d * * * *", minute, minute+2)
```
heww, line 126: `coreSpec := fmt.Sprintf("0-59 %d,%d,%d * * * *", minute, (minute+1) % 60, (minute+2) % 60)` will be better than `switch`.

src/jobservice/period/enqueuer_test.go
heww: Random testing is more realistic.

steven-zou: > om test, it will make the code coverage unstable.
> It's better to use a for loop to verify each fake time minutes

The case here is to make sure there are jobs enqueued based on the current minute. I don't think the `for` loop is suitable for this case.

Issue #11310, opened 2020-03-26
all scan info lose after upgrade from 1.10 to 2.0


bitsf:
build 3166
Because harbor 2.0 use Trivy as the default scanner, after upgrade to 2.0, it will force override the old clair scanner setting.
I can understand use Trivy as default when fresh installation.
But this is a bad user experience after upgrade case.
All old scan info is invalid, and even critical is that user can not pull images if enabled "Prevent vulnerable images from running"

reasonerjt:
We should make sure whatever scanner user chooses we should never update the default scanner.

xaleeks:
I think this is the experience we are looking for (apply to each individual project).  The already configured scanner always has highest priority, even if the upgrade specifies trivy with '--with-trivy'.  Trivy is higher priority than Clair.

The new install / upgrade experience should be like this
Brand New install
On brand new install, customer can pick 
     i. Nothing. In this case no scanner is installed
     ii. “with-trivy”. Trivy is installed and default
     iii. “with-clair”. Clair is installed and default
     iv. “with-clair” and “with-trivy”. Both are installed and Trivy is the default
 
Upgrades
Original Harbor had no scanner configured [this is the same case as for a brand new install]
     i. Nothing. In this case no scanner is installed
     ii. “with-trivy”. Trivy is installed and default
     iii. “with-clair”. Clair is installed  and default
     iv. “with-clair” and “with-trivy”. Both are installed. Trivy is the default

Original Harbor had a scanner X configured 
     i. Nothing. Error out since a scanner was configure pre-upgrade
     ii. “with-trivy”. Trivy is installed. X scanner is default
     iii. “with-clair”. Clair is installed . X scanner is default
     iv. “with-clair” and “with-trivy”.  both are installed. X scanner is default

steven-zou:
@xaleeks 

` Error out since a scanner was configured pre-upgrade`?

What does `Error out` mean? Throw an error when such case happen?

Issue #11549, opened 2020-04-09
Trivy should not be the default scanner when user already installed the clair 


stonezdj:
Install Harbor 1.9.0 with clair
Upgrade to Harbor tile 2.0 
The default scanner becomes trivvy.
It is OK if it is a fresh install, but current is an upgrade, it could not be the customer expected. they would like to take the clair as the default scanner.

<img width="1261" alt="Screen Shot 2020-04-09 at 5 25 28 PM" src="https://user-images.githubusercontent.com/2696760/78880170-95c34f00-7a87-11ea-93b6-ccfb4e3d9748.png">

steven-zou:
It was fixed by the PR https://github.com/goharbor/harbor/pull/11558. Close it now.

PR #11558, opened 2020-04-09
fix(scanner): Do not override the default scanner on init

danielpacak:
Resolves: #11310


renmaosheng:
it will also fixes https://github.com/goharbor/harbor/issues/11549

PR #11575, opened 2020-04-10
fix artifact trash filter issue

### LINKED ISSUE #11533: GC doesn't delete the useless manifest and it'sblob
`[Steps]`

1. push hello-world:latest to project A.
2. push hello-world:latest to project B.
3. delete tag of hello-world:latest from project A.
4. run GC with untagged, nothing removed. — artifact projectA/hello-world is removed, but do not call registry API to delete manifest as it’s referenced by project B.
5. delete tag of hello-world:latest from project B.
6. run GC with untagged, GC will call API to remove manifest projectB/hello-world, then do real GC. But, projectA/hello-world still in the back-end storage. GC will not remove any blobs.

`[Root Cause]`
GC is using artifact and artifact trash to decide which manifest should be removed. https://github.com/goharbor/harbor/blob/master/src/jobservice/job/impl/gc/garbage_collection.go#L249 

But, the Trash filter should use repo_name and digest to get the candidate. Currently, it's using digest only.

wy65701436:
fixes #11533

GC jobs will use the filter results to call registry API to delete manifest.

In the current imple, the filter function in some case does not return the deleted artifact as it's using digest as the filter condition.

Like: If one artifact is deleted, but there is another project/repo has a image with same digest with the deleted one, filter func will
not mark the deleted artifact as candidate. It results in, GC job does not call API to remove the manifest.

To fix it, update the filter to use both digest and repository name to filter candidate.

### DIFF
--- a/src/jobservice/job/impl/gc/garbage_collection.go
+++ b/src/jobservice/job/impl/gc/garbage_collection.go
@@ -215,10 +215,14 @@ func (gc *GarbageCollector) cleanCache() error {
 // 1, required part, the artifacts were removed from Harbor.
 // 2, optional part, the untagged artifacts.
 func (gc *GarbageCollector) deleteCandidates(ctx job.Context) error {
+	if os.Getenv("UTTEST") == "true" {
+		gc.logger = ctx.GetLogger()
+	}
 	// default is not to clean trash
 	flushTrash := false
 	defer func() {
 		if flushTrash {
+			gc.logger.Info("flush artifact trash")
 			if err := gc.artrashMgr.Flush(ctx.SystemContext()); err != nil {
 				gc.logger.Errorf("failed to flush artifact trash: %v", err)
 			}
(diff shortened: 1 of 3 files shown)


src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -235,28 +236,32 @@ func (gc *GarbageCollector) deleteCandidates(ctx job.Context) error {
...
 			return err
 		}
+		gc.logger.Info("start to delete untagged artifact.")
 		for _, art := range untagged {
-			gc.logger.Infof("delete the untagged artifact: ProjectID:(%d)-RepositoryName(%s)-MediaType:(%s)-Digest:(%s)",
-				art.ProjectID, art.RepositoryName, art.ManifestMediaType, art.Digest)
 			if err := gc.artCtl.Delete(ctx.SystemContext(), art.ID); err != nil {
 				// the failure ones can be GCed by the next execution
 				gc.logger.Errorf("failed to delete untagged:%d artifact in DB, error, %v", art.ID, err)
 			}
```
heww, line 244: `continue` needed here, otherwise line 245 will be logged.

wy65701436: fixed.

PR #11582, opened 2020-04-11
fix(blob): delete project blob with project_id

### LINKED ISSUE #11564: The quota of my project appeared to be 0.
I have an env, kept it running for 2 days, in this period of time, I did a few push/pull,  
triggered a few GC.

I occasionally found the quota usage of my project `test` is zero, while there are artifacts in the project.

I checked the db there's no data for that project in table `project_blob`

```
registry=# select project_id,repository_name,digest from artifact where project_id=3;
 project_id |   repository_name    |                                 digest
------------+----------------------+-------------------------------------------------------------------------
          3 | test/harbor-migrator | sha256:06d932402423c5b5fe6c358a48cec6551adaa8896c28cee1e9b899647220737f
          3 | test/hello-world     | sha256:a23d1a711f517cb15a38afebe1ea8efff9580a7ca5e58eb8bb7c4e743972db23
          3 | test/hello-world     | sha256:adaf641080c6999eda00329030755a684b03b1546d4a346a0254b149b66f1949
          3 | test/hello-world     | sha256:ba27d460cd1f22a1a4331bdf74f4fccbc025552357e8a3249c40ae216275de96
          3 | test/hello-world     | sha256:44a99eeeda49ab2907fbab83d726f62add7c9826f3dbddb375a647d8ab1ff1cb
(5 rows)

registry=# select * from project_blob where project_id=3;
 id | project_id | blob_id | creation_time
----+------------+---------+---------------
(0 rows)
```

heww:
Closes #11564

### DIFF
--- a/src/pkg/blob/dao/dao_test.go
+++ b/src/pkg/blob/dao/dao_test.go
@@ -267,13 +267,59 @@ func (suite *DaoTestSuite) TestExistProjectBlob() {
 func (suite *DaoTestSuite) TestDeleteProjectBlob() {
 	ctx := suite.Context()
 
-	projectID := int64(1)
-	blobID := int64(1000)
-
-	_, err := suite.dao.CreateProjectBlob(ctx, projectID, blobID)
+	digest := suite.DigestString()
+	blobID, err := suite.dao.CreateBlob(ctx, &models.Blob{Digest: digest})
 	suite.Nil(err)
 
-	suite.Nil(suite.dao.DeleteProjectBlob(ctx, projectID, blobID))
+	projectID1 := int64(1)
+	projectID2 := int64(2)
+	projectID3 := int64(3)
+
+	_, err = suite.dao.CreateProjectBlob(ctx, projectID1, blobID)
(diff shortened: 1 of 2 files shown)


src/pkg/blob/dao/dao_test.go
```diff
@@ -267,13 +267,44 @@ func (suite *DaoTestSuite) TestExistProjectBlob() {
 func (suite *DaoTestSuite) TestDeleteProjectBlob() {
 	ctx := suite.Context()
 
-	projectID := int64(1)
-	blobID := int64(1000)
+	digest := suite.DigestString()
+	blobID, err := suite.dao.CreateBlob(ctx, &models.Blob{Digest: digest})
+	suite.Nil(err)
```
reasonerjt, line 272: Ideally, you should also test the case when input project ID doesn't exist.

heww: Test case when input the projectID which doesn't have records in the `project_blob` table was added.

PR #11588, opened 2020-04-13
update trace level to error

wy65701436:
Use level error instead to log trace


src/server/error/error.go
```diff
@@ -51,7 +51,7 @@ func SendError(w http.ResponseWriter, err error) {
 	// the error detail is logged only, and will not be sent to the client to avoid leaking server information
 	if statusCode >= http.StatusInternalServerError {
 		log.Error(errPayload)
-		log.Debug(stackTrace)
+		log.Error(stackTrace)
```
ywk253100, line 54: Some other logs may be inserted between the error and its stack if use two `log.Error`, so suggest to use `log.Errorf("%s %s", errPayload, stackTrace)`

wy65701436: updated

PR #11620, opened 2020-04-15
feat(scanner): make Clair and Trivy as reserved name for scanners

### LINKED ISSUE #11524: Got four scanners after enabling TLS
image

heww:
Closes #11524

### DIFF
--- a/make/migrations/postgresql/0030_2.0.0_schema.up.sql
+++ b/make/migrations/postgresql/0030_2.0.0_schema.up.sql
@@ -200,3 +200,7 @@ ALTER TABLE replication_task ALTER COLUMN dst_resource TYPE varchar(512);
 /*remove count from quota hard and quota_usage used json*/
 UPDATE quota SET hard = hard - 'count';
 UPDATE quota_usage SET used = used - 'count';
+
+/* make Clair and Trivy as reserved name for scanners in-tree */
+UPDATE scanner_registration SET name = concat_ws('-', name, uuid) WHERE name IN ('Clair', 'Trivy') AND immutable = FALSE;
+UPDATE scanner_registration SET name = split_part(name, '-', 1) WHERE immutable = TRUE;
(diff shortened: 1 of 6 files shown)


src/pkg/scan/init.go
```diff
@@ -19,8 +19,6 @@ import (
 	"github.com/goharbor/harbor/src/lib/q"
 	"github.com/goharbor/harbor/src/pkg/scan/dao/scanner"
 	sc "github.com/goharbor/harbor/src/pkg/scan/scanner"
-	"github.com/goharbor/harbor/src/pkg/types"
-	"github.com/google/uuid"
 	"github.com/pkg/errors"
```
wy65701436, line 22: use lib/errors instead?

make/migrations/postgresql/0030_2.0.0_schema.up.sql
```diff
@@ -200,3 +200,7 @@ ALTER TABLE replication_task ALTER COLUMN dst_resource TYPE varchar(512);
 /*remove count from quota hard and quota_usage used json*/
 UPDATE quota SET hard = hard - 'count';
 UPDATE quota_usage SET used = used - 'count';
+
+/* make Clair and Trivy as reserved name for scanners in-tree */
+UPDATE scanner_registration SET name = concat_ws('-', name, uuid) WHERE name IN ('Clair', 'Trivy') AND immutable = FALSE;
```
wy65701436, line 205: we do need to document it as the name will be changed if user is using the external clair or trivy.

heww: This will be in docs, maybe also in release note.

renmaosheng:
what if pre-v2.0 version has Clair and Trivy installed? how is the upgrade behaving?

heww:
> what if pre-v2.0 version has Clair and Trivy installed? how is the upgrade behaving?

The scanner named Clair or Trivy which was installed manually will be renamed to `Clair-xxx-xxx` or `Trivy-xxx-xxx` (xxx-xxx is the uuid of the scanner).

The scanner named Clair which was installed by harbor will keep the same name as before.

make/migrations/postgresql/0030_2.0.0_schema.up.sql
steven-zou: @heww 

release notes may be more proper.

Issue #11281, opened 2020-03-25
harbor 2.0 slack webhook should consider API rate limit


bitsf:
And when using slack, we should consider the API rate limit, maybe add some cache to group the message help

This error maybe relative rate limit
![image](https://user-images.githubusercontent.com/188115/77526997-67aa1080-6ec6-11ea-9038-ed6efb564fcd.png)

But error code is not 409, according to the doc
https://api.slack.com/docs/rate-limits

reasonerjt:
Is there a way to augment the quota in Slack?

Another to mitigate the issue, could we explicitly set the job not to retry if it's sending notification in Slack?

mmpei:
i will check this scenario and try to make the webhook more graceful to slack.
and nowdays i think it is not an issue because harbor will retry after a backoff interval time when error occurs.

wy65701436:
Doc for reference: https://api.slack.com/docs/rate-limits

renmaosheng:
@mmpei @wy65701436 could we have some discussion in next community meeting?

steven-zou:
@mmpei 

For controlling the speed of sending messages to the slack, you can try to use the `Scheduled` job which is executed after the given time.

With this idea, you can use a base timestamp as a lock, a lock of the timestamp that indicates the triggering time of last job should be acquired before firing the next job. Then the jobs will be sequenced.

e.g.: 
t0=1000
t1=1002
t2=1003
...

wy65701436:
close it as https://github.com/goharbor/harbor/pull/11623 merged.

PR #11623, opened 2020-04-15
fix slack rate limit issue

mmpei:
https://github.com/goharbor/harbor/issues/11281

### DIFF
--- a/src/jobservice/job/impl/notification/http_helper.go
+++ b/src/jobservice/job/impl/notification/http_helper.go
@@ -1,0 +1,41 @@
+package notification
+
+import (
+	"net/http"
+
+	commonhttp "github.com/goharbor/harbor/src/common/http"
+)
+
+const (
+	secure   = "secure"
+	insecure = "insecure"
+)
+
+var (
+	httpHelper *HTTPHelper
+)
+
+// HTTPHelper in charge of sending notification messages to remote endpoint
+type HTTPHelper struct {
+	clients map[string]*http.Client
+}
+
+func init() {
+	httpHelper = &HTTPHelper{
+		clients: map[string]*http.Client{},
+	}
+	httpHelper.clients[secure] = &http.Client{
+		Transport: commonhttp.GetHTTPTransport(commonhttp.SecureTransport),
+	}
+	httpHelper.clients[insecure] = &http.Client{
+		Transport: commonhttp.GetHTTPTransport(commonhttp.InsecureTransport),
+	}
+}
+
+// GetHTTPInstance ...
+func GetHTTPInstance(insec bool) *http.Client {
+	if insec {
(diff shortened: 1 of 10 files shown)


src/jobservice/job/impl/notification/http_helper.go
```diff
@@ -0,0 +1,41 @@
...
+	secure   = "secure"
+	insecure = "insecure"
+)
+
+var (
+	httpHelper *HTTPHelper
+)
+
+// HTTPHelper in charge of sending notification messages to remote endpoint
+type HTTPHelper struct {
```
steven-zou, line 19: Do we really need this helper? Base on the scenario, it seems we just only need two singleton HTTP client functions.

```go
var insecHttpClient, secureHttpClient *http.Client

func getInsecuHttpClient() *http.Client{
  if insecHttpClient == nil {
      insecHttpClient = commonhttp.GetHTTPTransport(commonhttp.InsecureTransport)
  }

  return insecHttpClient
}

func getSecuHttpClient() *http.Client{
  if secureHttpClient == nil {
      secureHttpClient = commonhttp.GetHTTPTransport(commonhttp. SecureTransport)
  }

  return secureHttpClient
}
```

src/jobservice/job/impl/notification/slack_job.go
```diff
@@ -0,0 +1,104 @@
...
+	return 1
+}
+
+// ShouldRetry ...
+func (sj *SlackJob) ShouldRetry() bool {
+	return true
+}
+
+// Validate implements the interface in job/Interface
+func (sj *SlackJob) Validate(params job.Parameters) error {
```
steven-zou, line 48: I saw this job has parameters. I'd like to suggest validating the parameters here.

src/jobservice/job/impl/notification/http_helper.go
mmpei: we can define more httpclient of custom configurations. like connection pool settings, timeout settings and so on. and there are only two kinds now.

src/jobservice/job/impl/notification/slack_job.go
mmpei: ok. i will check.

mmpei: i checked that all jobs have parameters but leave the validate function null. so go through it in future?

mmpei: fixed

src/jobservice/job/impl/notification/http_helper.go
steven-zou: This is something like over-engineering because we don't have other requirements till now (especially different timeout cases). If you like, we can keep it as it is now.

PR #11627, opened 2020-04-15
fix notification policy ugrade issue

### LINKED ISSUE #11624: Core & jobservice containers can't be started after migrating from 1.9/1.10 with unreachable webhook to 2.0
Core & jobservice containers can't be started after migrating from 1.9/1.10 with unreachable webhook to 2.0.

Test steps:
1. Deploy harbor 1.9/1.10;
2. Create a project with webhook configured as https://2.2.2.2;
3. Migrate to 2.0;
4. Core & Jobservice container is in starting status all the time;

 Jobservice Error Log:
image

wy65701436:
Fixes #11624
All of the existing policies created v1.10 has no name, it fails the upgrade process.
When to set the unique constraint for policy name, the empty can be seen as duplicated key.

ERROR:  could not create unique index "notification_policy_name_key"
DETAIL:  Key (name)=() is duplicated.

### DIFF
--- a/make/migrations/postgresql/0030_2.0.0_schema.up.sql
+++ b/make/migrations/postgresql/0030_2.0.0_schema.up.sql
@@ -191,6 +191,9 @@ DROP TABLE IF EXISTS access_log;
 /*remove the constraint for project_id in table 'notification_policy'*/
 ALTER TABLE notification_policy DROP CONSTRAINT unique_project_id;
 
+/*the existing policy has no name, to make sure the unique constraint for name works, use the id as name*/
+/*if the name is set via API, it will be force to be changed with new pattern*/
+UPDATE notification_policy SET name=CONCAT('policy_', id);
 /*add the unique constraint for name in table 'notification_policy'*/
 ALTER TABLE notification_policy ADD UNIQUE (name);
 


make/migrations/postgresql/0030_2.0.0_schema.up.sql
```diff
@@ -191,6 +191,8 @@ DROP TABLE IF EXISTS access_log;
 /*remove the constraint for project_id in table 'notification_policy'*/
 ALTER TABLE notification_policy DROP CONSTRAINT unique_project_id;
 
+/*the existing policy has no name, to make sure the unique constraint for name works, use the id as name*/
+UPDATE notification_policy SET name=id;
```
heww, line 195: Will the name of the policy be displayed in the portal ? If so `ploicy-{id}` will be better than `{id}`.

wy65701436: ok

make/migrations/postgresql/0030_2.0.0_schema.up.sql
```diff
@@ -191,6 +191,8 @@ DROP TABLE IF EXISTS access_log;
 /*remove the constraint for project_id in table 'notification_policy'*/
 ALTER TABLE notification_policy DROP CONSTRAINT unique_project_id;
 
+/*the existing policy has no name, to make sure the unique constraint for name works, use the id as name*/
+UPDATE notification_policy SET name=CONCAT('policy_', id);
```
bitsf, line 195: 这个最好加个 where 判断一下name不是 policy_ 开头

wy65701436: we do have no name in the older version. @mmpei can you help to confirm?

PR #11641, opened 2020-04-16
fix(replication) fix replication hang seems when enable internal tls

bitsf:
It seems when enable internal tls, transfer some bigger blob, it will hang at 
https://github.com/goharbor/harbor/blob/master/src/pkg/registry/client.go#L390

the root reason is not clear now, it seems set req.ContentLength = -1 can workaround.


bitsf:
This fix will break Quota function that depend on content length

PR #11645, opened 2020-04-16
Fix: Default Transport HTTP2 related hang issue

ninjadq:
Create a transport the same as default except forceattempthttp2


renmaosheng:
could you please add a note in the code to explain the root-cause? thanks. @ninjadq

PR #11660, opened 2020-04-19
Add tls 12 min version for registry

ninjadq:


renmaosheng:
does this mean our usage is not correct or there might be a bug in go? thanks.

ninjadq:
@renmaosheng the library Default Transport has set the reasonable parameters for a Transport. But it sets (ForceAttemptHTTP2 to true)[https://github.com/golang/go/blob/dev.boringcrypto.go1.13/src/net/http/transport.go#L49] which will try http2 connecting. In this step, It probably hit this (bug)[https://github.com/golang/go/issues/20979]

PR #11666, opened 2020-04-20
fix catalog api issue

wy65701436:
The v2 catalog API needs filter out the empty repository and the repository which artifacts are all with no tags.

1，In v2.0.0, Harbor does not delete repository even there is no artifact, it's different with v1.10
2, Compare with docker distribution, it doesn't return the respository with untagged images.


src/server/registry/catalog.go
```diff
@@ -129,6 +140,34 @@ func (r *repositoryHandler) sendResponse(w http.ResponseWriter, req *http.Reques
...
+	}
+
+	for _, art := range arts {
+		if len(art.Tags) != 0 {
+			return true, nil
+		}
+	}
+
+	// if all of artifact are untagged, filter out
+	return false, nil
```
heww, line 168: How about using `Count` method? It's more lightweight than `List`

wy65701436: but, it needs two DB calls. The `list` cannot be avoid.

wy65701436: per offline discussion, use `list` to query artifact.

PR #11667, opened 2020-04-20
Clean up gitaction env and add tag for pytest

danfengliu:

### DIFF
--- a/tests/ci/api_run.sh
+++ b/tests/ci/api_run.sh
@@ -28,10 +28,13 @@ function uploader {
 set +e
 
 docker ps
-
+# Log storage space before running API_DB.robot
+df -H
 # run db auth api cases
 if [ "$1" = 'DB' ]; then
     pybot -X -v ip:$2 -v HARBOR_PASSWORD:Harbor12345 $DIR/../../tests/robot-cases/Group0-BAT/API_DB.robot
+    # Log storage space after running API_DB.robot
+    df -H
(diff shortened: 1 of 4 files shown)


tests/ci/api_run.sh
```diff
@@ -28,10 +28,10 @@ function uploader {
 set +e
 
 docker ps
-
+echo `df -H`
```
reasonerjt, line 31: Could you add more context or remove the `df` which is mainly for debug purpose?

tests/ci/api_run.sh
```diff
@@ -40,7 +40,9 @@ elif [ "$1" = 'LDAP' ]; then
                                   ldap_search_password=admin \
                                   ldap_base_dn=dc=example,dc=com \
                                   ldap_uid=cn
+    echo `df -H`
```
reasonerjt, line 43: I don't think this is necessary

tests/ci/api_run.sh
```diff
@@ -40,7 +40,9 @@ elif [ "$1" = 'LDAP' ]; then
...
     pybot -X -v ip:$2 -v HARBOR_PASSWORD:Harbor12345 $DIR/../../tests/robot-cases/Group0-BAT/API_LDAP.robot
+    echo `df -H`
```
reasonerjt, line 45: I don't think this is necessary

PR #11670, opened 2020-04-20
Tag retention webhook support

tedgxt:
Tag retention webhook support


src/controller/event/handler/webhook/artifact/retention.go
```diff
@@ -0,0 +1,144 @@
...
+		log.Debug("notification feature is not enabled")
+		return nil
+	}
+
+	trEvent, ok := value.(*event.RetentionEvent)
+	if !ok {
+		return errors.New("invalid tag retention event type")
+	}
+	if trEvent == nil {
+		return fmt.Errorf("nil tag retention event")
```
wy65701436, line 34: use the errors.New() instead?

src/controller/event/handler/webhook/artifact/retention.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	payload, dryRun, project, err := constructRetentionPayload(trEvent)
+	// if dry run, do not trigger webhook
+	if dryRun {
+		log.Debugf("retention task %v is dry run", trEvent.TaskID)
+		return nil
+	}
+	if err != nil {
```
wy65701436, line 43: please check the error first

src/controller/event/handler/webhook/artifact/retention.go
```diff
@@ -0,0 +1,144 @@
...
+		log.Errorf("failed to get tag retention policy %d: error: %v", execution.PolicyID, err)
+		return nil, false, 0, err
+	}
+	if md == nil {
+		return nil, false, 0, fmt.Errorf("policy %d not found with tag retention event", execution.PolicyID)
+	}
+
+	extURL, err := config.ExtURL()
+	if err != nil {
+		log.Errorf("Error while reading external endpoint URL: %v", err)
```
wy65701436, line 102: no return?

src/controller/event/handler/webhook/artifact/retention.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	for _, v := range event.Deleted {
+		target := v.Target
+		deletedArtifact := &model.ArtifactInfo{
+			Type:       target.Kind,
+			Status:     event.Status,
+			NameAndTag: target.Repository + ":" + target.Tags[0],
+		}
+		payload.EventData.Retention.DeletedArtifact = []*model.ArtifactInfo{deletedArtifact}
```
wy65701436, line 130: there is only one deleted artifact?

src/pkg/notifier/model/event.go
```diff
@@ -78,3 +80,27 @@ type ReplicationResource struct {
 	Provider     string `json:"provider,omitempty"`
 	Namespace    string `json:"namespace,omitempty"`
 }
+
+// Retention describes tag retention infos
```
wy65701436, line 84: the pkg/notifier should be a common package, it should not contain any specific model. 

We can define them into src/controller/event/model. OR use the `Custom` attribute of EventData.

BTW, please remove the replication model as well.

src/pkg/retention/faked_controller.go
```diff
@@ -0,0 +1,88 @@
+package retention
+
+import (
+	"github.com/goharbor/harbor/src/pkg/retention/policy"
+	"github.com/goharbor/harbor/src/pkg/retention/q"
+)
+
+// FakedRetentionController ...
+type FakedRetentionController struct {
```
wy65701436, line 9: can we define it into testing/pkg?

src/controller/event/handler/webhook/artifact/retention.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/goharbor/harbor/src/pkg/notifier/model"
+)
+
+// RetentionHandler preprocess tag retention event data
+type RetentionHandler struct {
+}
+
+// Handle ...
+func (r *RetentionHandler) Handle(value interface{}) error {
+	if !config.NotificationEnable() {
```
wy65701436, line 24: the logic is handled in the SendHookWithPolicies

src/controller/event/handler/webhook/artifact/retention.go
tedgxt: > there is only one deleted artifact?

right now, a notification of retention will received for one tag retention. So there is only one deleted artifact when receiving the notification.

src/controller/event/handler/webhook/artifact/retention.go
tedgxt: > the logic is handled in the SendHookWithPolicies

Fast return. while the value is false, we don't need to do the redundant parsing for event.

src/controller/event/handler/webhook/artifact/retention.go
tedgxt: > use the errors.New() instead?

will be fixed

src/controller/event/handler/webhook/artifact/retention.go
tedgxt: > please check the error first

will be fixed

src/controller/event/handler/webhook/artifact/retention.go
tedgxt: > no return?

I'm considering that the hostname get from the extURL is not essential info of webhook, so I will keep sending webhook when getting extURL failed.

wy65701436: But, can you make sure line 104 can be worked fine as there is an error raised before? It seems that the extURL is still be parsed, isn't it should be parsed only on success to get the extURL?

can we set a default value, like "" or whatever meaningful string, for hostname if an error happens?

src/controller/event/handler/webhook/artifact/retention.go
wy65701436: I agree with the return early. 

But, if we do want to do `fast return`, can we define it at one place at somewhere instead of at any of handler? In case someone who doesn't know this and introduces a new handler.

src/controller/event/handler/webhook/artifact/retention.go
tedgxt: 1. ExtURL() will always return nil error 
2. if extURL is empty, hostname will be empty("")

src/controller/event/handler/webhook/artifact/retention.go
bitsf: so if this is by design, you can add a break here.
And I suggest you add a totalNum / delNum

src/controller/event/handler/webhook/artifact/retention.go
```diff
@@ -0,0 +1,144 @@
...
+	payload := &model.Payload{
+		Type:     event.EventType,
+		OccurAt:  event.OccurAt.Unix(),
+		Operator: execution.Trigger,
+		EventData: &model.EventData{
+			Retention: &model.Retention{
+				Total:             task.Total,
+				Retained:          task.Retained,
+				HarborHostname:    hostname,
+				ProjectName:       event.Deleted[0].Target.Namespace,
```
bitsf, line 115: event.Deleted[0] may introduce index Exception
I suggest you don't send webhook if len(Deleted)==0

src/pkg/retention/faked_controller.go
tedgxt: > can we define it into testing/pkg?

It cannot be. Because the interface methods contains some struct in package `retention`, like `Execution` and `Task`

src/controller/event/handler/webhook/artifact/retention.go
tedgxt: now the model is  notification->event build->event publish->trigger webhook(handler sending webhook).   There are two parts: 1. event build&publish; 2. webhook sending.  The handler is the entrypoint to process webhook, so we have to judge webhook enable in the handler.

src/controller/event/handler/webhook/artifact/retention.go
tedgxt: We've already got `Total ` & `Retained ` in the webhook payload.

bitsf:
This PR seems need refactor based on PR https://github.com/goharbor/harbor/pull/12473

src/controller/event/handler/webhook/artifact/retention.go
tedgxt: > event.Deleted[0] may introduce index Exception
> I suggest you don't send webhook if len(Deleted)==0

fixed

src/pkg/notifier/model/event.go
tedgxt: moved to src/controller/event/model

tedgxt:
rebase error occurred

PR #11726, opened 2020-04-24
fix(retention) fix retention repository with slash

### LINKED ISSUE #11701: tag retention does not handle the extra "/" in repository name well.
Version: v2.0.0-rc

I have a repo called `test/ns/redis`

When I dry-run a retention rule there are error like this:
```
2020-04-23T09:13:44Z [INFO] [/pkg/retention/job.go:83]: Run retention process.
 Repository: test/ns/redis 
 Rule Algorithm: or 
 Dry Run: true
2020-04-23T09:13:44Z [ERROR] [/pkg/retention/job.go:235]: retention job: http error: code 404, message {"errors":[{"code":"NOT_FOUND","message":"path /api/v2.0/projects/test/repositories/ns/redis/artifacts was not found"}]}
```
probably the "/" should be encoded as "+" when calling the API

bitsf:
fix #11701
tag retention does not handle the extra "/" in repository, need encode the slash twice

### DIFF
--- a/src/pkg/clients/core/image.go
+++ b/src/pkg/clients/core/image.go
@@ -17,10 +17,12 @@ package core
 import (
 	"fmt"
 	modelsv2 "github.com/goharbor/harbor/src/controller/artifact"
+	"github.com/goharbor/harbor/src/lib/encode/repository"
 )
 
-func (c *client) ListAllArtifacts(project, repository string) ([]*modelsv2.Artifact, error) {
-	url := c.buildURL(fmt.Sprintf("/api/v2.0/projects/%s/repositories/%s/artifacts", project, repository)) // should query only tag
+func (c *client) ListAllArtifacts(project, repo string) ([]*modelsv2.Artifact, error) {
+	repo = repository.Encode(repo)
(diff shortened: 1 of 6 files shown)


src/pkg/clients/core/image.go
```diff
@@ -16,10 +16,13 @@ package core
 
 import (
 	"fmt"
+	"net/url"
+
```
steven-zou, line 20: One extra line here

src/pkg/clients/core/image.go
```diff
@@ -29,12 +32,17 @@ func (c *client) ListAllArtifacts(project, repository string) ([]*modelsv2.Artif
...
 	return c.httpclient.Delete(url)
 }
 
 func (c *client) DeleteArtifactRepository(project, repository string) error {
+	repository = encodeRepository(repository)
 	url := c.buildURL(fmt.Sprintf("/api/repositories/%s/%s", project, repository))
 	return c.httpclient.Delete(url)
 }
+
+func encodeRepository(repository string) string {
```
steven-zou, line 46: I remembered we had other places with the similar cases, did we define a utility encoding method before?

Is this repo encoding method align with the repo encoding way we discussed before?

PR #11751, opened 2020-04-26
fix return code on getting non exist manifest

wy65701436:
It's found by conformance test, it should be 404 instead of 500 when to get the non exist manifest

### DIFF
--- a/src/lib/context.go
+++ b/src/lib/context.go
@@ -14,7 +14,9 @@
 
 package lib
 
-import "context"
+import (
+	"context"
+)
 
 type contextKey string
 
--- a/src/server/middleware/contenttrust/contenttrust.go
+++ b/src/server/middleware/contenttrust/contenttrust.go
@@ -1,7 +1,6 @@
 package contenttrust
 
 import (
-	"fmt"
 	"net/http"
 
 	"github.com/goharbor/harbor/src/controller/artifact"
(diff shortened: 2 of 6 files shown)


src/lib/context.go
```diff
@@ -73,10 +76,12 @@ func WithArtifactInfo(ctx context.Context, art ArtifactInfo) context.Context {
 }
 
 // GetArtifactInfo gets the ArtifactInfo from the context
-func GetArtifactInfo(ctx context.Context) (art ArtifactInfo) {
+func GetArtifactInfo(ctx context.Context) (art ArtifactInfo, err error) {
 	value := getFromContext(ctx, contextKeyArtifactInfo)
 	if value != nil {
 		art, _ = value.(ArtifactInfo)
+	} else {
```
reasonerjt, line 83: you don't need the "else" here

src/lib/context.go
```diff
@@ -73,10 +76,12 @@ func WithArtifactInfo(ctx context.Context, art ArtifactInfo) context.Context {
 }
 
 // GetArtifactInfo gets the ArtifactInfo from the context
-func GetArtifactInfo(ctx context.Context) (art ArtifactInfo) {
+func GetArtifactInfo(ctx context.Context) (art ArtifactInfo, err error) {
```
reasonerjt, line 79: I don' t think an object does not exist in context should be an error.

wy65701436: you mean, we do check none in all of the callers, and define whether to return 404?

PR #11755, opened 2020-04-26
v2 auth middleware handles the ping request from internal

reasonerjt:
When scanner like trivy handles the auth flow to pull image, it pings
the /v2 and access the token service url in response body, by default it
will be external endpoint of Harbor.
There will be problem when Harbor is deployed on a single node with hairpinning not
supported.

This commit makes sure the address of token service in the challenge is
internal url of core component when the request is from internal.

### DIFF
--- a/src/server/middleware/v2auth/auth.go
+++ b/src/server/middleware/v2auth/auth.go
@@ -17,6 +17,7 @@ package v2auth
 import (
 	"fmt"
 	"net/http"
+	"net/url"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


src/server/middleware/v2auth/auth.go
```diff
@@ -105,6 +108,19 @@ func getChallenge(req *http.Request, accessList []access) string {
 	return challenge
 }
 
+func tokenSvcEndpoint(req *http.Request) (string, error) {
```
wy65701436, line 111: actually, no error returned in this func?

wy65701436, line 111: actually, there is no error returned in this func?

PR #11759, opened 2020-04-27
Add switch to the access log

xiaonancc77:
Add a switch to the log, because not all the time you want to record the log in the database.
Signed-off-by: elsanli <elsanli@tencent.com>


reasonerjt:
Sorry but I don't think we want this, we may consider implement a sweep policy to remove the access log older than certain time, or implement different receivers for the access "events" but not a simple switch to disable it.

PR #11765, opened 2020-04-27
feat(quota,webhook): send quota webhook for put and mount blob

### LINKED ISSUE #11712: Webhook is not triggered when failing to push image due to quotas exceed
Do you think a webhook message is needed when failing to push image like this:

```
93de22642fe0: Pushed 
dc842dc2a2ae: Pushed 
16542a8fc3be: Pushed 
6597da2e2e52: Pushed 
977183d4e999: Pushed 
c8be1b8f4d60: Pushing  65.58MB
denied: adding 76.3 MiB of storage resource, which when updated to current usage of 174.9 MiB will exceed the configured upper limit of 200.0 MiB.
```

heww:
Closes #11712

### DIFF
--- a/src/controller/event/metadata/quota.go
+++ b/src/controller/event/metadata/quota.go
@@ -34,17 +34,21 @@ func (q *QuotaMetaData) Resolve(evt *event.Event) error {
 		return errors.New("not supported quota status")
 	}
 
-	evt.Topic = topic
-	evt.Data = &event2.QuotaEvent{
+	data := &event2.QuotaEvent{
 		EventType: topic,
 		Project:   q.Project,
-		Resource: &event2.ImgResource{
+		OccurAt:   q.OccurAt,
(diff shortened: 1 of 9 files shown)


src/controller/event/metadata/quota.go
```diff
@@ -34,17 +34,21 @@ func (q *QuotaMetaData) Resolve(evt *event.Event) error {
...
-	evt.Data = &event2.QuotaEvent{
+	data := &event2.QuotaEvent{
 		EventType: topic,
 		Project:   q.Project,
-		Resource: &event2.ImgResource{
+		OccurAt:   q.OccurAt,
+		RepoName:  q.RepoName,
+		Msg:       q.Msg,
+	}
+	if q.Tag != "" || q.Digest != "" {
```
wy65701436, line 44: add UT for this?

PR #11783, opened 2020-04-28
Add internal https 4 portal

ninjadq:


make/photon/prepare/templates/core/env.jinja
```diff
@@ -18,6 +18,7 @@ POSTGRESQL_SSLMODE={{harbor_db_sslmode}}
 POSTGRESQL_MAX_IDLE_CONNS={{harbor_db_max_idle_conns}}
 POSTGRESQL_MAX_OPEN_CONNS={{harbor_db_max_open_conns}}
 REGISTRY_URL={{registry_url}}
+PORTAL_URL={{portal_url}}
```
ywk253100, line 21: Why core need to know the url of portal?

ninjadq: used for the health check on core

PR #11805, opened 2020-04-29
Improve i18n service

### LINKED ISSUE #11799: after harbor 2.0 upgrade, UI i18n json can't load automatically even refresh browser
upgrade from 1.10 to 2.0 build 3388
image

The i18n file is using fixed name, and no expired time.
image

User has to disable browser cache once to refresh the UI to workaround
image

AllForNothing:
Fix #11799 


AllForNothing:
Close as PR #11833 is better

PR #11924, opened 2020-05-13
fixing a link; "installation_guide.md" doesn't exist

squillace:


xaleeks:
@squillace  Thanks for catching this Ralph, please sign your commit so we can merge

squillace:
this is strange, as I made the change from the web ui I was under the impression it was signed by default. I'm cool if someone else makes the fix, as I am soooper busy and won't be able to get to this for a while.

bitsf:
use `git commit -s --amend` to sign commit again

PR #11978, opened 2020-05-20
Fix replication bugs #11974, #11939

### LINKED ISSUE #11939: Can not pull image from k8s.gcr.io when registry is docker-registry and rule contains brace
Version: harbor 2.0
OS: Ubuntu 1804

When replicating images from `k8s.gcr.io` through endpoint `docker-registry` and replications rule contain brace parenthesis. 
like this `kube-{apiserver,controller-manager,proxy,scheduler}`
The replication job will hang forever

### LINKED ISSUE #11974: Azure Replication not working after 2.0 Upgrade
Hi,

after the Harbor upgrade to 2.0 the replication to our Azure Container Registry does not work anymore. The registry is marked as unhealthy.
When I try to add the registry again I get the error message "Failed to ping endpoint".
In the debug logs, I see that the acr responds to the ping request with unauthorized.
Does harbor use the given credentials for its registry ping request?

ywk253100:
Fix replication bugs: fix #11974, fix #11939

### DIFF
--- a/src/pkg/registry/auth/authorizer.go
+++ b/src/pkg/registry/auth/authorizer.go
@@ -54,9 +54,9 @@ type authorizer struct {
 }
 
 func (a *authorizer) Modify(req *http.Request) error {
-	// Nil URL means this is the first time the authorizer is called
+	// Nil underlying authorizer means this is the first time the authorizer is called
 	// Try to connect to the registry and determine the auth scheme
-	if a.url == nil {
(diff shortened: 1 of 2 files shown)


fredgate:
When Harbor 2.0.1 will be released ? This fix is very important because 2.0.0 broke the replication to ACR, and we can no longer use it and are blocked :-(

Issue #11964, opened 2020-05-18
Found “unknown OS ” issue when doing some scanning with trivy


steven-zou:
i have a folder called contour. clair can scan all the artifacts ok, but trivy is failing with error

```
2020-05-15T01:37:14Z [INFO] [/pkg/scan/job.go:325]: registration:
2020-05-15T01:37:14Z [INFO] [/pkg/scan/job.go:336]: {
  "uuid": "b983bf84-94d6-11ea-80d9-0242ac1d0009",
  "name": "Trivy",
  "description": "The Trivy scanner adapter",
  "url": "http://trivy-adapter:8080",
  "disabled": false,
  "is_default": true,
  "health": "healthy",
  "auth": "",
  "skip_certVerify": false,
  "use_internal_addr": true,
  "adapter": "Trivy",
  "vendor": "Aqua Security",
  "version": "v0.6.0",
  "create_time": "2020-05-13T05:00:54.45531Z",
  "update_time": "2020-05-13T05:00:54.455313Z"
}
2020-05-15T01:37:14Z [INFO] [/pkg/scan/job.go:325]: scanRequest:
2020-05-15T01:37:14Z [INFO] [/pkg/scan/job.go:336]: {
  "registry": {
    "url": "http://core:8080",
    "authorization": "[HIDDEN]"
  },
  "artifact": {
    "namespace_id": 18,
    "repository": "contour/contour",
    "tag": "",
    "digest": "sha256:967b6fe66ab3b99d94234ff54a230bce8226e8b42fca8dcc5e89f5b17034f750",
    "mime_type": "application/vnd.docker.distribution.manifest.v2+json"
  }
}
2020-05-15T01:37:14Z [INFO] [/pkg/scan/job.go:156]: Report mime types: [application/vnd.scanner.adapter.vuln.report.harbor+json; version=1.0]
2020-05-15T01:37:14Z [INFO] [/pkg/scan/job.go:202]: Get report for mime type: application/vnd.scanner.adapter.vuln.report.harbor+json; version=1.0
2020-05-15T01:37:16Z [INFO] [/pkg/scan/job.go:219]: Report with mime type application/vnd.scanner.adapter.vuln.report.harbor+json; version=1.0 is not ready yet, retry after 5 seconds
[...]

PR #11983, opened 2020-05-20
fix(trivy): Handle gracefully scratch and slim images

danielpacak:
This commit bumps up Trivy to 0.7.0 and Trivy adapter service to 0.10.0
in order to handle scratch and slim images, for which we cannot detect
the underlying operating system.

Resolves: #11964


danielpacak:
I tested with projectcontour/contour:v0.15.3 which can be scanned with Trivy with no errors:

trivy_adapter

scan_contour

PR #11990, opened 2020-05-21
Added ip options for http, https and notary in docker-compose

spagno:
Signed-off-by: Spagno <aspagnolo@vmware.com>

### DIFF
--- a/make/photon/prepare/templates/docker_compose/docker-compose.yml.jinja
+++ b/make/photon/prepare/templates/docker_compose/docker-compose.yml.jinja
@@ -354,12 +354,12 @@ services:
 {% endif %}
     dns_search: .
     ports:
-      - {{http_port}}:8080
+      - {{http_ip}}:{{http_port}}:8080
 {% if protocol == 'https' %}
-      - {{https_port}}:8443
+      - {{https_ip}}:{{https_port}}:8443
 {% endif %}
 {% if with_notary %}
-      - 4443:4443
+      - {{notary_ip}}:4443:4443
 {% endif %}
     depends_on:
       - registry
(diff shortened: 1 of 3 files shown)


ninjadq:
you can set extra certs in a file and set the path of that file to `ca_bundle` item in `harbor.yml`

make/photon/prepare/templates/docker_compose/docker-compose.yml.jinja
```diff
@@ -354,12 +354,12 @@ services:
 {% endif %}
     dns_search: .
     ports:
-      - {{http_port}}:8080
+      - {{http_ip}}:{{http_port}}:8080
```
ninjadq, line 357: Could you please use one argument like `ip_port` here and explain them in comments and doc to reduce the number of config items?

make/photon/prepare/templates/docker_compose/docker-compose.yml.jinja
```diff
@@ -354,12 +354,12 @@ services:
...
 {% if protocol == 'https' %}
-      - {{https_port}}:8443
+      - {{https_ip}}:{{https_port}}:8443
```
ninjadq, line 359: the same as above

make/photon/prepare/templates/docker_compose/docker-compose.yml.jinja
```diff
@@ -354,12 +354,12 @@ services:
...
 {% endif %}
 {% if with_notary %}
-      - 4443:4443
+      - {{notary_ip}}:4443:4443
```
ninjadq, line 362: maybe here we can use an argument like `notary_ip_port`

Issue #11858, opened 2020-05-08
Replication doesn't support bintray


wy65701436:
Try to add bintray as a docker-registry endpoint, but when I try to replicate image from bintray, the replication job fails.

### [Logs]
```
2020-05-07T17:17:51Z [INFO] [/replication/transfer/image/transfer.go:255]: copying the blob sha256:92226bcf1537ac1108bd23b611d44ddf101b4e4ce55739cfb96ece100fcd163a...
2020-05-07T17:17:51Z [ERROR] [/replication/transfer/image/transfer.go:268]: failed to pulling the blob sha256:92226bcf1537ac1108bd23b611d44ddf101b4e4ce55739cfb96ece100fcd163a: strconv.ParseInt: parsing "": invalid syntax
2020-05-07T17:17:51Z [ERROR] [/replication/transfer/image/transfer.go:163]: strconv.ParseInt: parsing "": invalid syntax
```

It should be caused by the response header of pulling blob doesn't contain `Content-Length`, which is a **required** attribute in replication as Harbor needs it to acquire the project quota.

https://github.com/goharbor/harbor/blob/master/src/pkg/registry/client.go#L358

bitsf:
The main problem is pullblob method can't get the content-length which will passing to pushblob method.

Actually if we use curl to get the blob, we find after the redirect response it can return the content-length in header

After further testing I find that our http request include a "Accept-Encoding: gzip" header, and the response include a "Content-Encoding: gzip" too.
That means the response is actually compressed, so no real content-length returned.

A quick fix should be disable gzip by set `Transport.DisableCompression=true`
Or using the size in manifest maybe also work

bitsf:
should be same as issue https://github.com/goharbor/harbor/issues/9216

Issue #9216, opened 2019-09-23
Cannot replicate images from Red Hat registry (registry.access.redhat.com)


rfc1459:
Trying to replicate an image from the public Red Hat Container Registry (for instance `ubi7/ubi-minimal:7.7-10`) fails even though the same image can be pulled by a standard Docker daemon and Red Hat's own podman.

### Expected
Harbor should pull the image from the registry and add it to a local project.

### Actual
The replication job fails with `strconv.ParseInt: parsing "": invalid syntax` while trying to pull the first blob referenced by the manifest.

## Steps to reproduce the problem

1. Create a registry with name `redhat-registry`, type `docker-registry` and URL `https://registry.access.redhat.com`
2. Create a project (in my case: `ubi7`)
3. Create a replicaiton rule with the following settings:
  * Name: `redhat-ubi7`
  * Replication mode: Pull-based
  * Source registry: `redhat-registry`
  * Source resource filter:
    * Name: `ubi7/ubi-minimal`
    * Tag: 7.7-10
    * Resource: image
  * Destination namespace: `ubi7`
  * Trigger Mode: Manual
  * Override: Yes
  * Enable rule: Yes
4. Trigger the replication rule

## Versions

- harbor version: [1.9.0]
- docker engine version: [18.09.4]
- docker-compose version: [1.22.0]

## Additional context

Relevant excerpt from the job log:
```
2019-09-23T12:36:21Z [INFO] [/replication/transfer/image/transfer.go:95]: client for source registry [type: docker-registry, URL: https://registry.access.redhat.com, insecure: false] created
2019-09-23T12:36:21Z [INFO] [/replication/transfer/image/transfer.go:105]: client for destination registry [type: harbor, URL: http://core:8080, insecure: true] created
2019-09-23T12:36:21Z [INFO] [/replication/transfer/image/transfer.go:157]: copying ubi7/ubi-init:7.7-10(source registry) to ubi7/ubi-init:7.7-10(destination registry)...
2019-09-23T12:36:21Z [INFO] [/replication/transfer/image/transfer.go:261]: pulling the manifest of image ubi7/ubi-init:7.7-10 ...
2019-09-23T12:36:22Z [INFO] [/replication/transfer/image/transfer.go:272]: the manifest of image ubi7/ubi-init:7.7-10 pulled
2019-09-23T12:36:22Z [INFO] [/replication/transfer/image/transfer.go:294]: trying abstract a manifest from the manifest list...
[...]

PR #11998, opened 2020-05-21
fix(replication): ignore the problem of not getting Content-Length fr…

chlins:
…om the pull blob api

Fix: https://github.com/goharbor/harbor/issues/11858, https://github.com/goharbor/harbor/issues/9216

### DIFF
--- a/src/replication/transfer/image/transfer.go
+++ b/src/replication/transfer/image/transfer.go
@@ -16,12 +16,13 @@ package image
 
 import (
 	"errors"
+	"net/http"
+	"strings"
+
 	"github.com/docker/distribution/manifest/manifestlist"
 	"github.com/docker/distribution/manifest/schema1"
 	common_http "github.com/goharbor/harbor/src/common/http"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
-	"net/http"
-	"strings"
 
 	"github.com/docker/distribution"
 	"github.com/docker/distribution/manifest/schema2"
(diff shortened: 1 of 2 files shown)


reasonerjt:
We must make sure this doesn't impact the correctness for quota usage.

src/replication/transfer/image/transfer.go
```diff
@@ -269,6 +272,14 @@ func (t *transfer) copyBlob(srcRepo, dstRepo, digest string) error {
 		return err
 	}
 	defer data.Close()
+
+	t.logger.Debugf("the blob size from api is %d, from manifests is %d", size, sizeFromDescriptor)
```
bitsf, line 276: dup log as below

src/replication/transfer/image/transfer.go
```diff
@@ -269,6 +272,14 @@ func (t *transfer) copyBlob(srcRepo, dstRepo, digest string) error {
...
+	// get size 0 from PullBlob, use size from distribution.Descriptor instead.
+	if size == 0 {
+		size = sizeFromDescriptor
+		t.logger.Debugf("the blob size from api is 0, use size %d from manifests instead", size)
```
heww, line 280: `the blob size from api is 0, use size %d from manifests instead` to `the blob size from remote registry is 0, use size %d from manifests instead`

src/replication/transfer/image/transfer.go
```diff
@@ -243,12 +245,13 @@ func (t *transfer) copyContent(content distribution.Descriptor, srcRepo, dstRepo
 	// the media type of the layer or config can be "application/octet-stream",
 	// schema1.MediaTypeManifestLayer, schema2.MediaTypeLayer, schema2.MediaTypeImageConfig
 	default:
-		return t.copyBlob(srcRepo, dstRepo, digest)
+		return t.copyBlob(srcRepo, dstRepo, digest, size)
```
heww, line 248: `t.copyBlob(srcRepo, dstRepo, digest, size)` to `t.copyBlob(srcRepo, dstRepo, digest, conten.size)`

PR #12006, opened 2020-05-22
Add api to delete blob and manifest

wy65701436:
Enable the capability of registry controller to delete blob and manifest


src/registryctl/config/registry/config.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+// DefaultRegConf ...
+const DefaultRegConf = "/etc/registry/config.yml"
+
+// StorageDriver ...
+var StorageDriver storagedriver.StorageDriver
+
+// ResolveConfiguration ...
+func ResolveConfiguration(configPath ...string) (*configuration.Configuration, error) {
```
reasonerjt, line 17: I imagine the path of configuration file should be a parm to the binary? so the parm of this func is must-have.

and please try not to write comment like `...`

src/registryctl/main.go
```diff
@@ -58,7 +70,7 @@ func (s *RegistryCtl) Start() {
 
 func main() {
 
-	configPath := flag.String("c", "", "Specify the yaml config file path")
+	configPath := flag.String("c", "", "Specify the yaml rConf file path")
```
reasonerjt, line 73: Ideally there should be a config file of registryctl that points to the registry configuration file.

wy65701436:
https://github.com/goharbor/harbor/issues/12024

src/registryctl/config/registry/config.go
wy65701436: updated.

src/registryctl/main.go
wy65701436: updated.

src/go.mod
```diff
@@ -83,3 +87,5 @@ require (
 	k8s.io/client-go v0.17.3
 	k8s.io/helm v2.16.3+incompatible
 )
+
+replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.3+incompatible
```
reasonerjt, line 91: why we need the `replace` here?

src/registryctl/api/base.go
```diff
@@ -16,24 +16,36 @@ package api
 
 import (
 	"encoding/json"
+	"github.com/goharbor/harbor/src/lib/errors"
+	server_error "github.com/goharbor/harbor/src/server/error"
```
reasonerjt, line 20: I wish we can clarify the package `harbor/src/server` because that package seems to be for `core` only

Adding dependency from registryctl to sub-package under src/server makes the dependency confusing.

reasonerjt:
I find there are more files that need to be vendored than I expected.
Have you run `go mod tidy`?

src/registryctl/api/registry/blob/blob.go
```diff
@@ -0,0 +1,42 @@
...
+
+type handler struct{}
+
+// ServeHTTP ...
+func (h *handler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	switch req.Method {
+	case http.MethodDelete:
+		h.delete(w, req)
+	default:
+		api.HandleForbidden(w)
```
reasonerjt, line 25: Souldn't it be method not allowed?

src/registryctl/api/registry/blob/blob.go
```diff
@@ -0,0 +1,42 @@
...
+func (h *handler) delete(w http.ResponseWriter, r *http.Request) {
+	ref := mux.Vars(r)["reference"]
+	if ref == "" {
+		api.HandleBadRequest(w, errors.New("no reference specified"))
+		return
+	}
+	// don't parse the reference here as RemoveBlob does.
+	cleaner := storage.NewVacuum(r.Context(), regConf.StorageDriver)
+	if err := cleaner.RemoveBlob(ref); err != nil {
+		api.HandleError(w, err)
```
reasonerjt, line 39: Print some log here?

reasonerjt, line 39: Print some log here?

src/registryctl/api/registry/gc/gc.go
```diff
@@ -0,0 +1,62 @@
...
+	registryConf string
+}
+
+// ServeHTTP ...
+func (h *handler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	switch req.Method {
+	case http.MethodPost:
+		h.start(w, req)
+	default:
+		api.HandleForbidden(w)
```
reasonerjt, line 29: method not allowed?

src/registryctl/api/registry/gc/gc.go
```diff
@@ -0,0 +1,62 @@
...
+type Result struct {
+	Status    bool      `json:"status"`
+	Msg       string    `json:"msg"`
+	StartTime time.Time `json:"starttime"`
+	EndTime   time.Time `json:"endtime"`
+}
+
+// start ...
+func (h *handler) start(w http.ResponseWriter, r *http.Request) {
+	cmd := exec.Command("/bin/bash", "-c", "registry_DO_NOT_USE_GC garbage-collect --delete-untagged=false "+h.registryConf)
```
reasonerjt, line 43: `registry_DO_NOT...` should be a const? and would it be better if it's absolute path?  And are you sure we need `bash` here as the file is a binary??

reasonerjt: But this is less important as we plan to remove it, if everything goes well, right...

src/registryctl/api/registry/manifest/manifest.go
```diff
@@ -0,0 +1,58 @@
...
+
+type handler struct{}
+
+// ServeHTTP ...
+func (h *handler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	switch req.Method {
+	case http.MethodDelete:
+		h.delete(w, req)
+	default:
+		api.HandleForbidden(w)
```
reasonerjt, line 27: 405

src/registryctl/client/client.go
```diff
@@ -25,15 +26,24 @@ import (
 	"github.com/goharbor/harbor/src/common/http/modifier/auth"
 	"github.com/goharbor/harbor/src/common/utils"
 	"github.com/goharbor/harbor/src/lib/log"
-	"github.com/goharbor/harbor/src/registryctl/api"
+	"github.com/goharbor/harbor/src/registryctl/api/registry/gc"
+)
+
+// const definition
```
reasonerjt, line 32: This is the client core used to call registryctl's API right?
We should attach request ID and extract request ID on the server side.  
If you don't want to handle it in this PR could you open an issue?

src/registryctl/config/registry/config.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"fmt"
+	"github.com/docker/distribution/configuration"
+	storagedriver "github.com/docker/distribution/registry/storage/driver"
+	"os"
+)
+
+// StorageDriver ...
+var StorageDriver storagedriver.StorageDriver
```
reasonerjt, line 11: Why not make storagedriver part of the config of registryctl?

src/registryctl/api/registry/manifest/manifest.go
```diff
@@ -0,0 +1,58 @@
...
+		return
+	}
+	repoName := mux.Vars(r)["name"]
+	if repoName == "" {
+		api.HandleBadRequest(w, errors.New("no repository name specified"))
+		return
+	}
+	var tags []string
+	v := r.URL.Query()
+	queryTags := v.Get("tags")
```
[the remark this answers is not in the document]
heww: Do we need tags here? Since 2.0, tags may exist in harbor db but not exist in registry.

src/go.mod
wy65701436: to resolve the azure go-autotest versioning issue, similar things in k8s world:

https://github.com/kubernetes-sigs/external-dns/blob/master/go.mod#L68
https://github.com/kubernetes/kubernetes/blob/master/go.mod#L155

We can group the replace, but so far it's only one here.

src/registryctl/api/registry/gc/gc.go
wy65701436: updated

src/registryctl/api/registry/blob/blob.go
wy65701436: updated

src/registryctl/api/registry/manifest/manifest.go
wy65701436: updated

src/registryctl/client/client.go
wy65701436: I will raise an issue for it, and handle it in another PR.

src/registryctl/config/registry/config.go
wy65701436: The configuraitons of the driver are arleady in the registry configuration file, and the driver is inited on registryctl container launch. I think it's ok to use a varaible here.

src/registryctl/config/config.go
```diff
@@ -100,4 +101,9 @@ func (c *Configuration) loadEnvs() {
 		c.LogLevel = loggerLevel
 	}
 
+	registryConf := os.Getenv("REGISTRY_CONFIG")
```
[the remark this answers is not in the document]
wy65701436: easier for testing &debugging.

src/registryctl/api/registry/gc/gc.go
wy65701436: nothing changed on the command in this PR but refactor, it works fine and will be deprecated soon. so just keep it as it is?

src/registryctl/api/registry/blob/blob.go
wy65701436: added.

wy65701436: added

src/registryctl/api/base.go
```diff
@@ -16,24 +16,36 @@ package api
...
 
-// response status code will be written automatically if there is an error
-func writeJSON(w http.ResponseWriter, v interface{}) error {
+// HandleBadRequest ...
+func HandleBadRequest(w http.ResponseWriter, err error) {
+	HandleError(w, errors.BadRequestError(err))
+}
+
+// HandleError ...
+func HandleError(w http.ResponseWriter, err error) {
```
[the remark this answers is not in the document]
wy65701436: do not want to import the `server_error` as each api.go.

src/registryctl/api/base.go
wy65701436: It can be moved into `src/lib`, and let `core` and `registrtctl` importing it.

But, I won't do it in this PR.

src/registryctl/api/registry/blob/blob.go
wy65701436: updated.

src/registryctl/api/registry/blob/blob.go
wy65701436: updated.

src/registryctl/api/registry/manifest/manifest.go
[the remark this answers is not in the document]
wy65701436: the fundamental method needs this pararmeter, and it can be a null. Leave it at here will not impact GC deletes manifest via digest only.

Sometime, we may need to parse the tag into delete method as in Harbor v2.0.0, the first tag of artifact will be in the backend storage。 -- But, this brings another issue, how can GC know which is the first tag of the artifact in Harbor? otherwise, we have to leveage the code to enumerate repository tags.

wy65701436:
@reasonerjt @heww code is updated and comments added, can you help to review?

src/registryctl/api/base.go
reasonerjt: Sure, let's make sure it's tracked.

src/registryctl/api/registry/gc/gc.go
reasonerjt: ok

src/registryctl/client/client.go
reasonerjt: ok

src/registryctl/config/registry/config.go
reasonerjt: I strongly think putting in in config is a better design.
I'll let @heww chime in ,if he's OK with this I'm fine

src/registryctl/api/registry/manifest/manifest.go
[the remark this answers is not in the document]
reasonerjt: I feel this may reflect a misunderstanding here, I don't think the parm  "tags" makes sense for this API...

src/registryctl/config/registry/config.go
heww: If putting `StorageDriver` in the config of the registryctl, it will be clear in the main of the registryctl. 

In the `Load` method of the registryctl conf, we can resolve configuration of the distrubtion and create `StorageDriver` from the configurtion. The result is that is subpackage `config/registry` can be removed.

src/registryctl/api/registry/manifest/manifest.go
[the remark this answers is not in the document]
wy65701436: ok, I will remove it.

src/registryctl/config/registry/config.go
wy65701436: added.

src/registryctl/main.go
```diff
@@ -23,6 +23,16 @@ import (
...
+
+	_ "github.com/docker/distribution/registry/storage/driver/azure"
+	_ "github.com/docker/distribution/registry/storage/driver/filesystem"
+	_ "github.com/docker/distribution/registry/storage/driver/gcs"
+	_ "github.com/docker/distribution/registry/storage/driver/inmemory"
+	_ "github.com/docker/distribution/registry/storage/driver/middleware/cloudfront"
+	_ "github.com/docker/distribution/registry/storage/driver/middleware/redirect"
+	_ "github.com/docker/distribution/registry/storage/driver/oss"
+	_ "github.com/docker/distribution/registry/storage/driver/s3-aws"
+	_ "github.com/docker/distribution/registry/storage/driver/swift"
```
heww, line 35: These drivers can be move to config pkg as the storage driver is created there.

wy65701436: a blank import should be only in a main or test package

PR #12015, opened 2020-05-23
fix: remove dead link

bweston92:
Link is dead and I cannot find where the article has gone, not listed in the index either.

https://goharbor.io/docs/1.10/build-customize-contribute/


bweston92:
@jonasrosland does that mean it will go up eventually?

PR #12018, opened 2020-05-25
Chart server enumeration

ninjadq:


src/core/api/chart_repository.go
```diff
@@ -442,7 +442,7 @@ func (cra *ChartRepositoryAPI) requireNamespace(namespace string) bool {
 
 	// Not existing
 	if !existing {
-		cra.SendBadRequestError(fmt.Errorf("namespace %s is not existing", namespace))
+		cra.SendUnAuthorizedError(errors.New("UnAuthorized"))
```
reasonerjt, line 445: I think the error code should be related to the status of security context?

PR #12101, opened 2020-05-29
update blob controller & manager

wy65701436:
1, add two more attributes, update_time and status
2, add delete and fresh update time method in blob mgr & ctr.


make/migrations/postgresql/0035_2_1_0_schema.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE blob ADD COLUMN update_time timestamp default CURRENT_TIMESTAMP;
```
reasonerjt, line 1: Try to make it idempotent, such as "IF NOT EXIST"

src/controller/blob/controller.go
```diff
@@ -75,6 +75,9 @@ type Controller interface {
 
 	// GetAcceptedBlobSize returns the accepted size of stream upload blob.
 	GetAcceptedBlobSize(sessionID string) (int64, error)
+
+	// FreshUpdateTime updates the update time for the blob.
+	FreshUpdateTime(ctx context.Context, digest string, time time.Time) (err error)
```
reasonerjt, line 80: Refreh?

make/migrations/postgresql/0035_2_1_0_schema.up.sql
wy65701436: updated

src/controller/blob/controller.go
wy65701436: updated

src/controller/blob/controller.go
```diff
@@ -75,6 +75,9 @@ type Controller interface {
 
 	// GetAcceptedBlobSize returns the accepted size of stream upload blob.
 	GetAcceptedBlobSize(sessionID string) (int64, error)
+
+	// ReFreshUpdateTime updates the update time for the blob.
+	ReFreshUpdateTime(ctx context.Context, digest string, time time.Time) (err error)
```
reasonerjt, line 80: It should be `Refresh` not `ReFresh`, and if it's refresh is should always be `Now` correct?  Why it's a parm?

If we think about the flow `RefreshUpdateTime` is a side effect of the API call right?  Is that necessary to expose it on controller level?

wy65701436: it's ok to remove the param from method and use now in the method.

To expose it in the ctl is because that it should be called in the middleware. And in the current program model, middleware/API does only consume ctl.

src/pkg/blob/dao/dao.go
```diff
@@ -162,13 +177,46 @@ func (d *dao) GetBlobByDigest(ctx context.Context, digest string) (*models.Blob,
...
+	data["update_time"] = time.Now()
+	data["status"] = blob.Status
+
+	qt := o.QueryTable(&models.Blob{})
+	cond := beego_orm.NewCondition()
+	c := cond.And("version__gte", blob.Version)
+
+	/*
+		generated simple sql string.
+		UPDATE "blob" SET "version" = "version" + $1, "update_time" = $2, "status" = $3
```
heww, line 198: `"version" = "version" + $1` => `"version" = "version" + 1`

src/pkg/blob/dao/dao.go
```diff
@@ -162,13 +177,46 @@ func (d *dao) GetBlobByDigest(ctx context.Context, digest string) (*models.Blob,
...
+		WHERE "id" IN ( SELECT T0."id" FROM "blob" T0 WHERE T0."version" >= $4 AND T0."id" = $5 AND T0."status"  IN ('delete', 'deleting')  )
```
heww, line 199: UPDATE "blob" SET version=version+1, update_time=?, status=? WHERE id = ? AND status in (?, ?) AND version >= ?

src/server/middleware/blob/head_blob.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return err
+	}
+
+	switch bb.Status {
+	case blob_models.StatusNone, blob_models.StatusDelete:
+		bb.Status = blob_models.StatusNone
+		if err = blob.Ctl.Touch(ctx, bb); err != nil {
+			log.Errorf("failed to update blob: %s status to None, error:%v", digest, err)
+			return err
```
heww, line 49: Ensure that 404 returned here.

src/pkg/blob/dao/dao.go
```diff
@@ -49,6 +51,9 @@ type DAO interface {
 	// UpdateBlob update blob
 	UpdateBlob(ctx context.Context, blob *models.Blob) error
 
+	// UpdateBlob update blob status
+	UpdateBlobStatus(ctx context.Context, blob *models.Blob) error
```
heww, line 55: Can it return (int, error) for this method so that we chan deal with 404 response more clear in head blob middleware?

src/server/middleware/blob/head_blob.go
```diff
@@ -0,0 +1,58 @@
...
+	"net/http"
+)
+
+// HeadBlobMiddleware intercept the head blob request
+func HeadBlobMiddleware() func(http.Handler) http.Handler {
+	return middleware.New(func(rw http.ResponseWriter, req *http.Request, next http.Handler) {
+		if err := handleHead(req); err != nil {
+			serror.SendError(rw, err)
+			return
+		}
```
heww, line 22: The code of handleHead and handleBlob can move to here.

make/migrations/postgresql/0035_2_1_0_schema.up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE blob ADD COLUMN IF NOT EXISTS update_time timestamp default CURRENT_TIMESTAMP;
+ALTER TABLE blob ADD COLUMN IF NOT EXISTS status varchar(255);
+ALTER TABLE blob ADD COLUMN IF NOT EXISTS version BIGINT default 0;
```
heww, line 3: Please add index for the status and version columns.

src/server/middleware/blob/head_blob.go
wy65701436: updated.

src/pkg/blob/dao/dao.go
wy65701436: it's actually inside the method, but again to add a return value.

make/migrations/postgresql/0035_2_1_0_schema.up.sql
wy65701436: added

src/pkg/blob/dao/dao.go
wy65701436: designed by beego orm

src/controller/blob/controller.go
wy65701436: the code is refactored, just refer to the latest commit.

src/pkg/blob/dao/dao.go
```diff
@@ -162,13 +170,50 @@ func (d *dao) GetBlobByDigest(ctx context.Context, digest string) (*models.Blob,
...
+
+// UpdateBlob cannot handle the status change.
 func (d *dao) UpdateBlob(ctx context.Context, blob *models.Blob) error {
 	o, err := orm.FromContext(ctx)
 	if err != nil {
 		return err
 	}
-
-	_, err = o.Update(blob)
+	blob.Version = blob.Version + 1
```
reasonerjt, line 214: Is it thread-safe?

src/pkg/blob/dao/dao.go
```diff
@@ -162,13 +170,50 @@ func (d *dao) GetBlobByDigest(ctx context.Context, digest string) (*models.Blob,
...
+	}
+
+	/*
+		generated simple sql string.
+		UPDATE "blob" SET "version" = "version" + $1, "update_time" = $2, "status" = $3
+		WHERE "id" IN ( SELECT T0."id" FROM "blob" T0 WHERE T0."version" >= $4 AND T0."id" = $5 AND T0."status"  IN ('delete', 'deleting')  )
+	*/
+
+	return qt.SetCond(c).Filter("id", blob.ID).
+		Filter("status__in", models.StatusMap[blob.Status]).
```
reasonerjt, line 204: If the transition failed due to invalid state, nothing will be logged.

IMO it would be better to handle the transition at manager level, and no one should call the update blob bypassing the manager.

src/pkg/blob/dao/dao.go
```diff
@@ -162,13 +170,50 @@ func (d *dao) GetBlobByDigest(ctx context.Context, digest string) (*models.Blob,
...
+		Update(data)
+}
+
+// UpdateBlob cannot handle the status change.
```
reasonerjt, line 208: To avoid misuse, shall we explicitly check the status must not be updated using this func?

src/server/middleware/blob/head_blob.go
```diff
@@ -0,0 +1,56 @@
...
+		count, err := blob.Ctl.Touch(req.Context(), bb)
+		if err != nil {
+			log.Errorf("failed to update blob: %s status to None, error:%v", art.Digest, err)
+			return err
+		}
+		if count == 0 {
+			return errors.New("the asking blob is in GC, mark it as non existing").WithCode(errors.NotFoundCode)
+		}
+	case blob_models.StatusDeleting, blob_models.StatusDeleteFailed:
+		return errors.New("the asking blob is in GC, mark it as non existing").WithCode(errors.NotFoundCode)
```
reasonerjt, line 51: Please also print log with request ID here.

src/server/middleware/blob/head_blob.go
```diff
@@ -0,0 +1,56 @@
...
+
+// handleHead ...
+func handleHead(req *http.Request) error {
+	none := lib.ArtifactInfo{}
+	art := lib.GetArtifactInfo(req.Context())
+	if art == none {
+		return errors.New("cannot get the artifact information from request context").WithCode(errors.NotFoundCode)
+	}
+
+	bb, err := blob.Ctl.Get(req.Context(), art.Digest)
```
reasonerjt, line 34: What if the blob does not exist?

src/server/middleware/blob/head_blob.go
```diff
@@ -0,0 +1,56 @@
...
+			return
+		}
+		next.ServeHTTP(rw, req)
+	})
+}
+
+// handleHead ...
+func handleHead(req *http.Request) error {
+	none := lib.ArtifactInfo{}
+	art := lib.GetArtifactInfo(req.Context())
```
reasonerjt, line 29: rename the var or add some comment to explain this is in fact blob info?

src/controller/blob/controller.go
```diff
@@ -75,6 +73,15 @@ type Controller interface {
 
 	// GetAcceptedBlobSize returns the accepted size of stream upload blob.
 	GetAcceptedBlobSize(sessionID string) (int64, error)
+
+	// Touch updates the blob status and increase version every time.
+	Touch(ctx context.Context, blob *blob.Blob) (int64, error)
```
reasonerjt, line 78: Touch should only be called by the head middleware, to update the status to "None"?

src/pkg/blob/manager.go
```diff
@@ -112,10 +117,18 @@ func (m *manager) Update(ctx context.Context, blob *Blob) error {
 	return m.dao.UpdateBlob(ctx, blob)
 }
 
+func (m *manager) UpdateBlobStatus(ctx context.Context, blob *models.Blob) (int64, error) {
```
reasonerjt, line 120: if the status is not updated due to version or status mismatch ,it should fail early and explicitly?

src/server/middleware/blob/head_blob.go
wy65701436: `get` will raise an NotFoundError.

src/server/middleware/blob/head_blob.go
wy65701436: updated

src/pkg/blob/manager.go
wy65701436: the first return value can be used to judge update succcess status, as the head blob middleware does.

src/server/middleware/blob/head_blob.go
wy65701436: updated.

src/pkg/blob/dao/dao.go
wy65701436: the version is removed.

src/pkg/blob/dao/dao.go
wy65701436: log is added.

src/controller/blob/controller.go
wy65701436: updated.

src/pkg/blob/dao/dao.go
wy65701436: use UpdateBlobStatus to handle status change.

src/pkg/blob/dao/dao.go
```diff
@@ -162,13 +171,58 @@ func (d *dao) GetBlobByDigest(ctx context.Context, digest string) (*models.Blob,
...
+	*/
+
+	count, err := qt.SetCond(c).Filter("id", blob.ID).
+		Filter("status__in", models.StatusMap[blob.Status]).
+		Update(data)
+	if err != nil {
+		return count, err
+	}
+	if count == 0 {
+		log.Warningf("no blob is updated according to query condition, id: %d, status_in, %v", blob.ID, models.StatusMap[blob.Status])
```
reasonerjt, line 211: I think you should also log the version.

PR #12110, opened 2020-06-01
Introduce API v2.1

ywk253100:
This commit introduces API version 2.1, after this commit merged, Harbor will maintain two versions of API(2.0 and 2.1) in release 2.1


reasonerjt:
What's the reason we need bump up here?  Does it align with our principle?

ywk253100:
As we'll introduce new changes(such as proxy cache for project API) for API in Harbor 2.1, we should maintain two versions of API(v2.0 and v2.1) in Harbor 2.1 if we follow the principle in OCI-supporting proposal.
We can discuss more about this.

heww:
I think we don't need to introduce v2.1 when only add new APIs to the system. Adding new APIs not break the compatibility.

PR #12116, opened 2020-06-02
# Some images are not display

xytnba:
Modified the path of the picture


wy65701436:
thanks for the PR, but it has been fixed in https://github.com/goharbor/website.

PR #12128, opened 2020-06-03
fix(jobservice):mismatch status issue when stopping job

steven-zou:
CHERRY_PICK to 2.0

- returnning nil instead of error when trying to stop a job that has been in the final status(Error/Success/Stopped)
- do enhancements to the periodic job unschedule func
- fix a UT nil ptr issue


Larry0315:
Version 1.10.6 still has this problem that has not been fixed.

Is there any plan to fix this problem for version 1.10.x?

PR #12175, opened 2020-06-07
move send error to source lib

wy65701436:
Move the sendError into library in case the cycle dependency as regsitry and core are now the consumers.

### DIFF
--- a/src/lib/http/error.go
+++ b/src/lib/http/error.go
@@ -17,8 +17,6 @@ package http
 import (
 	"fmt"
 	openapi "github.com/go-openapi/errors"
-	"github.com/go-openapi/runtime"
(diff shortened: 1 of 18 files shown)


src/lib/http/error.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package error
+package http
```
ywk253100, line 15: Putting it under the package `lib/errors` is more reasonable as most of the logic is assembling errors?

ywk253100: And please move the definition of `ErrResponder` to `src/server/v2.0/handler/base.go`, it isn't a common model

PR #12200, opened 2020-06-10
fix: prepare script prompts SELinux relabeling of / is not allowed.

cobolbaby:
The following error would be reported during installation on CoreOS:

```
/run/torcx/bin/docker: Error response from daemon: error setting label on mount source '/': SELinux relabeling of / is not allowed.
```


ninjadq:
if `privileged` opinion is added. the `:z` label no longer needed. could you remove that and rebase ?

cobolbaby:
Yes, I will do it.

PR #12218, opened 2020-06-12
Update "registryctl" to invoke "registry garbage-collect ..." directly

tianon:
This avoid shelling out to "/bin/bash" unnecessarily.

### DIFF
--- a/src/registryctl/api/registry/gc/gc.go
+++ b/src/registryctl/api/registry/gc/gc.go
@@ -40,7 +40,7 @@ type Result struct {
 
 // start ...
 func (h *handler) start(w http.ResponseWriter, r *http.Request) {
-	cmd := exec.Command("/bin/bash", "-c", "registry_DO_NOT_USE_GC garbage-collect --delete-untagged=false "+h.registryConf)
+	cmd := exec.Command("registry_DO_NOT_USE_GC", "garbage-collect", "--delete-untagged=false", h.registryConf)
 	var outBuf, errBuf bytes.Buffer
 	cmd.Stdout = &outBuf
 	cmd.Stderr = &errBuf


wy65701436:
hi,thanks for the PR. But the native garbage collection is decreprated as non-blocking GC introduced. And it's just a plan B in the source code, and it works well for several releases, my suggestion is just keep the current implemention.

tianon:
> the native garbage collection is decreprated as non-blocking GC introduced. And it's just a plan B in the source code

Ok, that's fair, but even the fallback code should be robust and well-supported, right?  Does the fallback code still get tested regularly?

> it works well for several releases

I'm really confused by this sentiment.  By that logic, we shouldn't ever improve anything, right? :sweat_smile:

This isn't a hill I want to die on, but the reason I made this PR is that with the code as it's currently written, the image/environment where this service runs needs to have `/bin/bash` available.

At the time I opened this, I was considering deploying this service in an Alpine Linux environment where that is not true, and `/bin/bash` is only available if the `bash` package is explicitly installed, when it isn't actually necessary.

This would be slightly more versatile if it were adjusted from `/bin/bash` to `/bin/sh` (which is available in _more_ environments, and certainly any where `/bin/bash` is), but that still then requires a shell, whereas my suggested change requires _only_ the (already necessary) `registry_DO_NOT_USE_GC` binary, and on top of that will handle quoting on the `h.registryConf` properly (which the original version will not do, which is probably "fine" as it's unlikely to contain spaces, quotes, etc, but still seems better/safer to be defensive than relying on implicit assumptions).

wy65701436:
> > the native garbage collection is decreprated as non-blocking GC introduced. And it's just a plan B in the source code
> 
> Ok, that's fair, but even the fallback code should be robust and well-supported, right? Does the fallback code still get tested regularly?
> 
> > it works well for several releases
> 
> I'm really confused by this sentiment. By that logic, we shouldn't ever improve anything, right? 😅
> 
> This isn't a hill I want to die on, but the reason I made this PR is that with the code as it's currently written, the image/environment where this service runs needs to have `/bin/bash` available.
> 
> At the time I opened this, I was considering deploying this service in an Alpine Linux environment where that is not true, and `/bin/bash` is only available if the `bash` package is explicitly installed, when it isn't actually necessary.
> 
> This would be slightly more versatile if it were adjusted from `/bin/bash` to `/bin/sh` (which is available in _more_ environments, and certainly any where `/bin/bash` is), but that still then requires a shell, whereas my suggested change requires _only_ the (already necessary) `registry_DO_NOT_USE_GC` binary, and on top of that will handle quoting on the `h.registryConf` properly (which the original version will not do, which is probably "fine" as it's unlikely to contain spaces, quotes, etc, but still seems better/safer to be defensive than relying on implicit assumptions).

the fall back code will not be tested regularly(just have a validation before release), it just used as a plan B in the source code level in case any urgent bug introduced by non blocking GC. And the code with native GC should be removed in 2.2 or 2.3.

I agree that anything could be improved even it works well, and yes the `/bin/bash` is a actually requirement for the base system. Have you tried to test it on the original image and the Alpine you metioned?

PR #12219, opened 2020-06-12
Remove unused "sudo" package from most images

tianon:
Notably missing is the "log" image, which still uses sudo.

Relatedly, it might be useful to include `security_opt: [ 'no-new-privileges' ]` in the generated `docker-compose.yml` to avoid the ability of a setuid binary to elevate privileges in the containers.

### DIFF
--- a/make/photon/clair-adapter/Dockerfile.base
+++ b/make/photon/clair-adapter/Dockerfile.base
@@ -1,7 +1,5 @@
 FROM photon:2.0
 
-RUN tdnf install -y sudo >>/dev/null\
-    && tdnf clean all \
-    && mkdir /clair-adapter/ \
+RUN mkdir /clair-adapter/ \
     && groupadd -r -g 10000 clair-adapter \
     && useradd --no-log-init -m -r -g 10000 -u 10000 clair-adapter
--- a/make/photon/registry/Dockerfile.base
+++ b/make/photon/registry/Dockerfile.base
@@ -1,6 +1,4 @@
 FROM photon:2.0
 
-RUN tdnf install sudo -y >> /dev/null\
-    && tdnf clean all \
-    && mkdir -p /etc/registry \
+RUN mkdir -p /etc/registry \
     && groupadd -r -g 10000 harbor && useradd --no-log-init -m -g 10000 -u 10000 harbor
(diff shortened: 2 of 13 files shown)


tianon:
(Rebased on `master` in an attempt to get that confusing @codecov result to come back green.)

make/photon/clair-adapter/Dockerfile.base
```diff
@@ -1,7 +1,5 @@
 FROM photon:2.0
 
-RUN tdnf install -y sudo >>/dev/null\
-    && tdnf clean all \
```
ninjadq, line 4: why remove the clean all and mkdir?

make/photon/registry/Dockerfile.base
```diff
@@ -1,6 +1,4 @@
 FROM photon:2.0
 
-RUN tdnf install sudo -y >> /dev/null\
-    && tdnf clean all \
```
ninjadq, line 4: the same as above

make/photon/registryctl/Dockerfile.base
```diff
@@ -1,6 +1,4 @@
 FROM photon:2.0
 
-RUN tdnf install sudo -y >> /dev/null \
-    && tdnf clean all \
```
ninjadq, line 4: the same

make/photon/clair-adapter/Dockerfile.base
tianon: If you look down another line in the diff, it doesn't remove the `mkdir` - it just becomes the first command in the `RUN` statement, which is why it's a changed line.

The `tdnf clean all` gets removed because it's purpose is to clean up extra cache, etc files that `tdnf install` leaves behind.  If we don't invoke `tdnf install`, we don't need to invoke `tdnf clean all`.

ninjadq:
This PR is very helpful. Thank you @tianon 
could you please update the PR in the This Week? If you can we'll merge it to master

ninjadq:
@tianon it seems one check pending here. Could you rebase the master and merge it?

tianon:
> @tianon it seems one check pending here. Could you rebase the master and merge it?

Absolutely, I've pushed the rebase -- hopefully it succeeds. :crossed_fingers:

PR #12231, opened 2020-06-13
Fix img url.

70data:


steven-zou:
@70data 

All the documents have been moved to goharbor/website. The related broken img URLs are fixed then.

This PR will be closed and not merged. Thanks for the contributions. From now on, any document revision PR, please raised to the goharbor/website repo.

PR #12274, opened 2020-06-18
Add proxy cache feature

stonezdj:
Update route to add proxy related middleware
Add proxy controller

### DIFF
--- a/src/controller/proxy/controller.go
+++ b/src/controller/proxy/controller.go
@@ -1,0 +1,215 @@
+//  Copyright Project Harbor Authors
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package proxy
+
+import (
+	"context"
+	"github.com/opencontainers/go-digest"
+	"io"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/docker/distribution"
+	"github.com/docker/distribution/manifest/manifestlist"
+	"github.com/goharbor/harbor/src/common/models"
+	"github.com/goharbor/harbor/src/controller/artifact"
+	"github.com/goharbor/harbor/src/controller/blob"
+	"github.com/goharbor/harbor/src/lib"
+	"github.com/goharbor/harbor/src/lib/errors"
+	"github.com/goharbor/harbor/src/lib/log"
+	"github.com/goharbor/harbor/src/replication/registry"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+const (
+	// wait more time than manifest (maxManifestWait) because manifest list depends on manifest ready
+	maxManifestListWait = 20
+	maxManifestWait     = 10
+	sleepIntervalSec    = 20
+)
+
+var (
+	// Ctl is a global proxy controller instance
+	ctl  Controller
+	once sync.Once
+)
+
+// Controller defines the operations related with pull through proxy
+type Controller interface {
+	// UseLocalBlob check if the blob should use local copy
+	UseLocalBlob(ctx context.Context, art lib.ArtifactInfo) bool
+	// UseLocalManifest check manifest should use local copy
+	UseLocalManifest(ctx context.Context, art lib.ArtifactInfo) bool
+	// ProxyBlob proxy the blob request to the remote server, p is the proxy project
+	// art is the ArtifactInfo which includes the digest of the blob
+	ProxyBlob(ctx context.Context, p *models.Project, art lib.ArtifactInfo) (int64, io.ReadCloser, error)
+	// ProxyManifest proxy the manifest request to the remote server, p is the proxy project,
+	// art is the ArtifactInfo which includes the tag or digest of the manifest
+	ProxyManifest(ctx context.Context, p *models.Project, art lib.ArtifactInfo) (distribution.Manifest, error)
+}
+type controller struct {
+	blobCtl     blob.Controller
+	registryMgr registry.Manager
+	artifactCtl artifact.Controller
+	local       localInterface
+}
+
+// ControllerInstance -- Get the proxy controller instance
+func ControllerInstance() Controller {
+	// Lazy load the controller
+	// Because LocalHelper is not ready unless core startup completely
+	once.Do(func() {
+		ctl = &controller{
+			blobCtl:     blob.Ctl,
+			registryMgr: registry.NewDefaultManager(),
+			artifactCtl: artifact.Ctl,
+			local:       newLocalHelper(),
+		}
+	})
+
+	return ctl
+}
+
+func (c *controller) UseLocalBlob(ctx context.Context, art lib.ArtifactInfo) bool {
+	if len(art.Digest) == 0 {
+		return false
+	}
+	exist, err := c.local.BlobExist(ctx, art)
+	if err != nil {
+		return false
+	}
+	return exist
+}
+
+func (c *controller) UseLocalManifest(ctx context.Context, art lib.ArtifactInfo) bool {
+	if len(art.Digest) == 0 {
+		return false
+	}
+	return c.local.ManifestExist(ctx, art)
+}
+
+func (c *controller) ProxyManifest(ctx context.Context, p *models.Project, art lib.ArtifactInfo) (distribution.Manifest, error) {
+	var man distribution.Manifest
+	remoteRepo := getRemoteRepo(art)
+	r, err := newRemoteHelper(p.RegistryID)
+	if err != nil {
+		return man, err
+	}
+	ref := getReference(art)
+	man, err = r.Manifest(remoteRepo, ref)
+	if err != nil {
+		if errors.IsNotFoundErr(err) {
+			go func() {
+				c.local.DeleteManifest(remoteRepo, art.Tag)
+			}()
+		}
+		return man, err
+	}
+	ct, _, err := man.Payload()
+	if err != nil {
+		return man, err
+	}
+	// Push manifest in background
+	go func() {
+		c.waitAndPushManifest(ctx, remoteRepo, man, art, ct, r)
+	}()
+
+	return man, nil
+}
+
+func (c *controller) ProxyBlob(ctx context.Context, p *models.Project, art lib.ArtifactInfo) (int64, io.ReadCloser, error) {
+	remoteRepo := getRemoteRepo(art)
+	log.Debugf("The blob doesn't exist, proxy the request to the target server, url:%v", remoteRepo)
+	rHelper, err := newRemoteHelper(p.RegistryID)
+	if err != nil {
+		return 0, nil, err
+	}
+
+	size, bReader, err := rHelper.BlobReader(remoteRepo, art.Digest)
+	if err != nil {
+		log.Errorf("failed to pull blob, error %v", err)
+		return 0, nil, err
+	}
+	desc := distribution.Descriptor{Size: size, Digest: digest.Digest(art.Digest)}
+	go func() {
+		err := c.putBlobToLocal(remoteRepo, art.Repository, desc, rHelper)
+		if err != nil {
+			log.Errorf("error while putting blob to local repo, %v", err)
+		}
+	}()
+	return size, bReader, nil
+}
+
+func (c *controller) putBlobToLocal(remoteRepo string, localRepo string, desc distribution.Descriptor, r remoteInterface) error {
+	log.Debugf("Put blob to local registry!, sourceRepo:%v, localRepo:%v, digest: %v", remoteRepo, localRepo, desc.Digest)
+	_, bReader, err := r.BlobReader(remoteRepo, string(desc.Digest))
+	if err != nil {
+		log.Errorf("failed to create blob reader, error %v", err)
+		return err
+	}
+	defer bReader.Close()
+	err = c.local.PushBlob(localRepo, desc, bReader)
+	return err
+}
+
+func (c *controller) waitAndPushManifest(ctx context.Context, remoteRepo string, man distribution.Manifest, art lib.ArtifactInfo, contType string, r remoteInterface) {
+	if contType == manifestlist.MediaTypeManifestList || contType == v1.MediaTypeImageIndex {
+		err := c.local.PushManifestList(ctx, art.Repository, getReference(art), man)
+		if err != nil {
+			log.Errorf("error when push manifest list to local :%v", err)
+		}
+		return
+	}
+	var waitBlobs []distribution.Descriptor
+	for n := 0; n < maxManifestWait; n++ {
+		time.Sleep(sleepIntervalSec * time.Second)
+		waitBlobs = c.local.CheckDependencies(ctx, art.Repository, man)
+		if len(waitBlobs) == 0 {
+			break
+		}
+		log.Debugf("Current n=%v artifact: %v:%v", n, art.Repository, art.Tag)
+	}
+	if len(waitBlobs) > 0 {
+		// docker client will skip to pull layers exist in local
+		// these blobs is not exist in the proxy server
+		// it will cause the manifest dependency check always fail
+		// need to push these blobs before push manifest to avoid failure
+		log.Debug("Waiting blobs not empty, push it to local repo directly")
+		for _, desc := range waitBlobs {
+			err := c.putBlobToLocal(remoteRepo, art.Repository, desc, r)
+			if err != nil {
+				log.Errorf("Failed to push blob to local repo, error: %v", err)
+				return
+			}
+		}
+	}
+	err := c.local.PushManifest(art.Repository, getReference(art), man)
+	if err != nil {
+		log.Errorf("failed to push manifest, tag: %v, error %v", art.Tag, err)
+	}
+}
+
+// getRemoteRepo get the remote repository name, used in proxy cache
+func getRemoteRepo(art lib.ArtifactInfo) string {
+	return strings.TrimPrefix(art.Repository, art.ProjectName+"/")
+}
+
+func getReference(art lib.ArtifactInfo) string {
+	if len(art.Tag) > 0 {
+		return art.Tag
+	}
+	return art.Digest
+}
--- a/src/controller/proxy/inflight.go
+++ b/src/controller/proxy/inflight.go
@@ -1,0 +1,46 @@
+//  Copyright Project Harbor Authors
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package proxy
+
+import "sync"
+
+type inflightRequest struct {
+	mu     sync.Mutex
+	reqMap map[string]interface{}
+}
+
+var inflightChecker = &inflightRequest{
+	reqMap: make(map[string]interface{}),
+}
+
+// addRequest if the artifact already exist in the inflightRequest, return false
+// else return true
+func (in *inflightRequest) addRequest(artifact string) (suc bool) {
+	in.mu.Lock()
+	defer in.mu.Unlock()
+	_, ok := in.reqMap[artifact]
+	if ok {
+		// Skip some following operation if it is in reqMap
+		return false
+	}
+	in.reqMap[artifact] = 1
+	return true
+}
+
+func (in *inflightRequest) removeRequest(artifact string) {
+	in.mu.Lock()
+	defer in.mu.Unlock()
+	delete(in.reqMap, artifact)
+}
--- a/src/controller/proxy/local.go
+++ b/src/controller/proxy/local.go
@@ -1,0 +1,191 @@
+//  Copyright Project Harbor Authors
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package proxy
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"github.com/docker/distribution"
+	"github.com/docker/distribution/manifest/manifestlist"
+	comHttpAuth "github.com/goharbor/harbor/src/common/http/modifier/auth"
+	"github.com/goharbor/harbor/src/controller/artifact"
+	"github.com/goharbor/harbor/src/core/config"
+	"github.com/goharbor/harbor/src/lib"
+	"github.com/goharbor/harbor/src/lib/log"
+	"github.com/goharbor/harbor/src/pkg/registry"
+	"io"
+	"time"
+)
+
+// localInterface defines operations related to local repo under proxy mode
+type localInterface interface {
+	// BlobExist check if the blob exist in local repo
+	BlobExist(ctx context.Context, art lib.ArtifactInfo) (bool, error)
+	// Manifest check if the manifest exist in local repo
+	ManifestExist(ctx context.Context, art lib.ArtifactInfo) bool
+	// PushBlob push blob to local repo
+	PushBlob(localRepo string, desc distribution.Descriptor, bReader io.ReadCloser) error
+	// PushManifest push manifest to local repo, ref can be digest or tag
+	PushManifest(repo string, ref string, manifest distribution.Manifest) error
+	// PushManifestList push manifest list to local repo
+	PushManifestList(ctx context.Context, repo string, ref string, man distribution.Manifest) error
+	// CheckDependencies check if the manifest's dependency is ready
+	CheckDependencies(ctx context.Context, repo string, man distribution.Manifest) []distribution.Descriptor
+	// DeleteManifest cleanup delete tag from local cache
+	DeleteManifest(repo, ref string)
+}
+
+func (l *localHelper) ManifestExist(ctx context.Context, art lib.ArtifactInfo) bool {
+	a, err := l.artifactCtl.GetByReference(ctx, art.Repository, art.Digest, nil)
+	if err != nil {
+		log.Errorf("check manifest exist failed, error %v", err)
+		return false
+	}
+	return a != nil
+}
+
+// localHelper defines operations related to local repo under proxy mode
+type localHelper struct {
+	registry    registry.Client
+	artifactCtl artifactController
+}
+
+type artifactController interface {
+	GetByReference(ctx context.Context, repository, reference string, option *artifact.Option) (artifact *artifact.Artifact, err error)
+}
+
+// newLocalHelper create the localInterface
+func newLocalHelper() localInterface {
+	l := &localHelper{artifactCtl: artifact.Ctl}
+	l.init()
+	return l
+}
+
+func (l *localHelper) BlobExist(ctx context.Context, art lib.ArtifactInfo) (bool, error) {
+	return l.registry.BlobExist(art.Repository, art.Digest)
+}
+
+func (l *localHelper) init() {
+	if l.registry != nil {
+		return
+	}
+	log.Debugf("core url:%s, local core url: %v", config.GetCoreURL(), config.LocalCoreURL())
+	// the traffic is internal only
+	registryURL := config.LocalCoreURL()
+	authorizer := comHttpAuth.NewSecretAuthorizer(config.ProxyServiceSecret)
+	l.registry = registry.NewClientWithAuthorizer(registryURL, authorizer, true)
+}
+
+func (l *localHelper) PushBlob(localRepo string, desc distribution.Descriptor, bReader io.ReadCloser) error {
+	log.Debugf("Put blob to local registry, localRepo:%v, digest: %v", localRepo, desc.Digest)
+	ref := string(desc.Digest)
+	artName := localRepo + ":" + ref
+	// use inflight checker to avoid multiple requests to push blob to local in same time
+	if !inflightChecker.addRequest(artName) {
+		return nil
+	}
+	defer inflightChecker.removeRequest(artName)
+	err := l.registry.PushBlob(localRepo, ref, desc.Size, bReader)
+	return err
+}
+
+func (l *localHelper) PushManifest(repo string, ref string, manifest distribution.Manifest) error {
+	// Make sure there is only one go routing to push current artName to local repo
+	artName := repo + ":" + ref
+	// use inflight checker to avoid multiple requests to push manifest to local in same time
+	if !inflightChecker.addRequest(artName) {
+		return nil
+	}
+	defer inflightChecker.removeRequest(artName)
+
+	mediaType, payload, err := manifest.Payload()
+	if err != nil {
+		return err
+	}
+	_, err = l.registry.PushManifest(repo, ref, mediaType, payload)
+	return err
+}
+
+// DeleteManifest cleanup delete tag from local repo
+func (l *localHelper) DeleteManifest(repo, ref string) {
+	log.Debug("Remove tag from repo if it is exist")
+	if err := l.registry.DeleteManifest(repo, ref); err != nil {
+		// sometimes user pull a non-exist image
+		log.Warningf("failed to remove artifact, error %v", err)
+	}
+}
+
+// updateManifestList -- Trim the manifest list, make sure at least one depend manifests is ready
+func (l *localHelper) updateManifestList(ctx context.Context, repo string, manifest distribution.Manifest) (distribution.Manifest, error) {
+	switch v := manifest.(type) {
+	case *manifestlist.DeserializedManifestList:
+		existMans := make([]manifestlist.ManifestDescriptor, 0)
+		for _, m := range v.Manifests {
+			art := lib.ArtifactInfo{Repository: repo, Digest: string(m.Digest)}
+			if l.ManifestExist(ctx, art) {
+				existMans = append(existMans, m)
+			}
+		}
+		return manifestlist.FromDescriptors(existMans)
+	}
+	return nil, fmt.Errorf("current manifest list type is unknown, manifest type[%T], content [%+v]", manifest, manifest)
+}
+
+func (l *localHelper) PushManifestList(ctx context.Context, repo string, ref string, man distribution.Manifest) error {
+	// For manifest list, it might include some different manifest
+	// it will wait and check for 30 mins, if all depend manifests exist then push it
+	// if time exceed, only push the new updated manifest list which contains existing manifest
+	var newMan distribution.Manifest
+	var err error
+	for n := 0; n < maxManifestListWait; n++ {
+		log.Debugf("waiting for the manifest ready, repo %v, ref:%v", repo, ref)
+		time.Sleep(sleepIntervalSec * time.Second)
+		newMan, err = l.updateManifestList(ctx, repo, man)
+		if err != nil {
+			return err
+		}
+		if len(newMan.References()) == len(man.References()) {
+			break
+		}
+	}
+
+	if len(newMan.References()) == 0 {
+		return errors.New("manifest list doesn't contain any pushed manifest")
+	}
+	_, pl, err := newMan.Payload()
+	if err != nil {
+		log.Errorf("failed to get payload, error %v", err)
+		return err
+	}
+	log.Debugf("The manifest list payload: %v", string(pl))
+	return l.PushManifest(repo, ref, newMan)
+}
+
+func (l *localHelper) CheckDependencies(ctx context.Context, repo string, man distribution.Manifest) []distribution.Descriptor {
+	descriptors := man.References()
+	waitDesc := make([]distribution.Descriptor, 0)
+	for _, desc := range descriptors {
+		log.Debugf("checking the blob dependency: %v", desc.Digest)
+		art := lib.ArtifactInfo{Repository: repo, Digest: string(desc.Digest)}
+		exist, err := l.BlobExist(ctx, art)
+		if err != nil || !exist {
+			log.Debugf("Check dependency failed!")
+			waitDesc = append(waitDesc, desc)
+		}
+	}
+	log.Debugf("Check dependency result %v", waitDesc)
+	return waitDesc
+}
--- a/src/controller/proxy/remote.go
+++ b/src/controller/proxy/remote.go
@@ -1,0 +1,81 @@
+//  Copyright Project Harbor Authors
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package proxy
+
+import (
+	"fmt"
+	"github.com/docker/distribution"
+	"github.com/goharbor/harbor/src/lib/log"
+	"github.com/goharbor/harbor/src/replication/adapter"
+	"github.com/goharbor/harbor/src/replication/registry"
+	"io"
+)
+
+// remoteInterface defines operations related to remote repository under proxy
+type remoteInterface interface {
+	// BlobReader create a reader for remote blob
+	BlobReader(repo, dig string) (int64, io.ReadCloser, error)
+	// Manifest get manifest by reference
+	Manifest(repo string, ref string) (distribution.Manifest, error)
+}
+
+// remoteHelper defines operations related to remote repository under proxy
+type remoteHelper struct {
+	regID    int64
+	registry adapter.ArtifactRegistry
+}
+
+// newRemoteHelper create a remoteHelper interface
+func newRemoteHelper(regID int64) (remoteInterface, error) {
+	r := &remoteHelper{regID: regID}
+	if err := r.init(); err != nil {
+		log.Errorf("failed to create remoteHelper error %v", err)
+		return nil, err
+	}
+	return r, nil
+}
+
+func (r *remoteHelper) init() error {
+
+	if r.registry != nil {
+		return nil
+	}
+	reg, err := registry.NewDefaultManager().Get(r.regID)
+	if err != nil {
+		return err
+	}
+	if reg == nil {
+		return fmt.Errorf("failed to get registry, registryID: %v", r.regID)
+	}
+	factory, err := adapter.GetFactory(reg.Type)
+	if err != nil {
+		return err
+	}
+	adp, err := factory.Create(reg)
+	if err != nil {
+		return err
+	}
+	r.registry = adp.(adapter.ArtifactRegistry)
+	return nil
+}
+
+func (r *remoteHelper) BlobReader(repo, dig string) (int64, io.ReadCloser, error) {
+	return r.registry.PullBlob(repo, dig)
+}
+
+func (r *remoteHelper) Manifest(repo string, ref string) (distribution.Manifest, error) {
+	man, _, err := r.registry.PullManifest(repo, ref)
+	return man, err
+}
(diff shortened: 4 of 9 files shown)


src/server/registry/route.go
```diff
@@ -31,6 +32,7 @@ import (
 func RegisterRoutes() {
 	root := router.NewRoute().
 		Path("/v2").
+		Middleware(repoproxy.BlobGetMiddleware()).
```
ywk253100, line 35: Add it only to the blob path rather that all registry requests?

src/server/middleware/repoproxy/proxy.go
```diff
@@ -0,0 +1,86 @@
...
+
+		log.Debugf("getting blob with url: %v\n", urlStr)
+		ctx := r.Context()
+		pName := distribution.ParseProjectName(urlStr)
+		dig := utils.ParseDigest(urlStr)
+		repo := utils.ParseRepo(urlStr)
+		repo = utils.TrimProxyPrefix(pName, repo)
+		p, err := project.Ctl.GetByName(ctx, pName, project.Metadata(false))
+		if err != nil {
+			log.Errorf("failed to get project, error:%v", err)
```
ywk253100, line 47: Should the process continue when getting error? Please double check other error handling logic

src/controller/proxy/remote.go
```diff
@@ -0,0 +1,105 @@
...
+	BlobReader(orgRepo, dig string) (int64, io.ReadCloser, error)
+	// ManifestByDigest get manifest by digest
+	ManifestByDigest(repository string, dig string) (distribution.Manifest, error)
+	// ManifestByTag get manifest by tag
+	ManifestByTag(repository string, tag string) (distribution.Manifest, error)
+}
+
+// remote defines operations related to remote repository under proxy
+type remote struct {
+	adapter *native.Adapter
```
ywk253100, line 40: native.Adapter cannot handle all kinds of registries

src/server/middleware/repoproxy/proxy.go
```diff
@@ -0,0 +1,84 @@
...
+			return
+		}
+
+		repo := utils.TrimProxyPrefix(art.ProjectName, art.Repository)
+		log.Debugf("the digest is %v", string(art.Digest))
+		remote := proxy.CreateRemoteInterface(p.RegistryID)
+		err = proxy.Ctl.ProxyManifest(ctx, p, repo, art, w, remote)
+		if err != nil {
+			log.Errorf("failed to proxy the manifest, error:%v", err)
+			next.ServeHTTP(w, r)
```
ywk253100, line 80: We should interrupt the flow here and return 500 to the client directly?

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	return reg.Status == model.Healthy
+}
+
+func (c *controller) UseLocalManifest(ctx context.Context, p *models.Project, art lib.ArtifactInfo) bool {
+	if !c.isProxyReady(p) {
+		return true
+	}
+	if len(string(art.Digest)) > 0 {
+		exist, err := c.local.BlobExist(ctx, art.Digest)
```
ywk253100, line 87: Shouldn't check the existence of the manifest in the upstream registry?

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	return exist
+}
+
+func (c *controller) ProxyManifest(ctx context.Context, p *models.Project, repo string, art lib.ArtifactInfo, w http.ResponseWriter, r RemoteInterface) error {
+	var man distribution.Manifest
+	var err error
+	if len(string(art.Digest)) > 0 {
+		// pull by digest
+		log.Debugf("Getting manifest by digiest %v", art.Digest)
```
ywk253100, line 111: digest typo

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,230 @@
...
+	} else { // pull by tag
+		if len(art.Tag) == 0 {
+			art.Tag = "latest"
+		}
+		log.Debugf("Getting manifest by tag %v", art.Tag)
+		man, err = r.ManifestByTag(repo, string(art.Tag))
+	}
+
+	if err != nil {
+		if errors.IsNotFoundErr(err) && len(art.Tag) > 0 {
```
ywk253100, line 122: Why check the 404 error only for tag? How about the digest?

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,230 @@
...
+			go func() {
+				c.local.DeleteManifest(ctx, repo, string(art.Tag))
+			}()
+		}
+		serror.SendError(w, err)
```
ywk253100, line 127: If we write error to the client directly here, is necessary to return error for this function anymore?

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,230 @@
...
+	return err
+}
+
+func setHeaders(w http.ResponseWriter, size int64, mediaType string, dig string) {
+	w.Header().Set("Content-Length", fmt.Sprintf("%v", size))
+	if len(mediaType) > 0 {
+		w.Header().Set("Content-Type", mediaType)
+	}
+	w.Header().Set("Docker-Content-Digest", dig)
+	w.Header().Set("Etag", dig)
```
ywk253100, line 199: Why we need the `Etag` header?

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,230 @@
...
+	localRepo := art.ProjectName + "/" + repo
+	if contType == manifestlist.MediaTypeManifestList {
+		err := c.local.PushManifestList(ctx, p, localRepo, tag, art, man)
+		if err != nil {
+			log.Errorf("error when push manifest list to local:%v", err)
+		}
+		return
+	}
+
+	for n := 0; n < maxWait; n = n + 1 {
```
ywk253100, line 212: So we just wait for 200 seconds in total, is it correct?

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,230 @@
...
+		return err
+	}
+
+	ct, payload, err := man.Payload()
+	setHeaders(w, int64(len(payload)), ct, art.Digest)
+	w.Write(payload)
+
+	// Push manifest in background
+	go func() {
+		c.waitAndPushManifest(ctx, p, repo, string(art.Tag), man, art, ct, r)
```
ywk253100, line 137: Does `waitAndPushManifest` handle the request of pulling by digest?

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,230 @@
...
+	}()
+
+	return nil
+}
+
+func (c *controller) ProxyBlob(ctx context.Context, p *models.Project, repo string, dig string, w http.ResponseWriter, r RemoteInterface) error {
+	log.Debugf("The blob doesn't exist, proxy the request to the target server, url:%v", repo)
+	desc := distribution.Descriptor{}
+	size, bReader, err := r.BlobReader(repo, dig)
+	defer bReader.Close()
```
ywk253100, line 147: The `defer` should be put after the error checking logic

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,230 @@
...
+		inflight:    make(map[string]interface{}),
+		artifactCtl: artifact.Ctl,
+		local:       CreateLocalInterface(),
+	}
+}
+
+func (c *controller) putBlobToLocal(ctx context.Context, p *models.Project, orgRepo string, localRepo string, desc distribution.Descriptor, r RemoteInterface) error {
+	log.Debugf("Put blob to local registry!, sourceRepo:%v, localRepo:%v, digest: %v", orgRepo, localRepo, desc.Digest)
+	_, bReader, err := r.BlobReader(orgRepo, string(desc.Digest))
+	defer bReader.Close()
```
ywk253100, line 184: Put `defer` after the error checking logic

src/controller/proxy/local.go
```diff
@@ -0,0 +1,181 @@
...
+	PushManifestList(ctx context.Context, p *models.Project, repo string, tag string, art lib.ArtifactInfo, man distribution.Manifest) error
+	// CheckDependencies check if the manifest's dependency is ready
+	CheckDependencies(ctx context.Context, man distribution.Manifest, dig string, mediaType string) []distribution.Descriptor
+	// DeleteManifest cleanup delete tag from local cache
+	DeleteManifest(ctx context.Context, repo, ref string)
+}
+
+// local defines operations related to local repo under proxy mode
+type local struct {
+	adapter *base.Adapter
```
ywk253100, line 56: Use `github.com/goharbor/harbor/src/replication/adapter/harbor.Adapter` instead

src/controller/proxy/local.go
```diff
@@ -0,0 +1,181 @@
...
+// TODO: replace it with head request to local repo
+func (l *local) BlobExist(ctx context.Context, dig string) (bool, error) {
+	return blob.Ctl.Exist(ctx, dig)
+}
+
+func (l *local) init() error {
+	if l.adapter != nil {
+		return nil
+	}
+	registryURL := config.GetCoreURL()
```
ywk253100, line 73: As we discussed, we should use `127.0.0.1` here

src/controller/proxy/local.go
```diff
@@ -0,0 +1,181 @@
...
+		},
+	}
+	adapter, err := base.New(reg)
+	l.adapter = adapter
+	return err
+}
+
+func (l *local) PushBlob(ctx context.Context, p *models.Project, localRepo string, desc distribution.Descriptor, bReader io.ReadCloser) error {
+	log.Debugf("Put blob to local registry, localRepo:%v, digest: %v", localRepo, desc.Digest)
+	if err := l.init(); err != nil {
```
ywk253100, line 88: Why we need to call `init` for every function call?

src/common/utils/utils.go
```diff
@@ -295,3 +295,11 @@ func FindNamedMatches(regex *regexp.Regexp, str string) map[string]string {
 	}
 	return results
 }
+
+// TrimProxyPrefix remove proxy prefix from original original project name
+func TrimProxyPrefix(projectName, repo string) string {
+	if strings.HasPrefix(repo, projectName+"/") {
```
reasonerjt, line 301: this `if` is not needed TrimPrefix will handle it if the string does not have the prefix.  

This func also should not be part of common package, it's proxy specific.

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,230 @@
...
+	if len(string(art.Digest)) > 0 {
+		exist, err := c.local.BlobExist(ctx, art.Digest)
+		if err == nil && exist {
+			return true
+		}
+	}
+	return false
+}
+
+func (c *controller) UseLocalBlob(ctx context.Context, p *models.Project, digest string) bool {
```
reasonerjt, line 95: I don't see how `UseLocalManifest` and `UseLocalBlob` need to co-exist, they are doing the very same thing.

src/controller/proxy/local.go
```diff
@@ -0,0 +1,181 @@
...
+	"time"
+)
+
+var (
+	mu       sync.Mutex
+	inflight map[string]interface{} = make(map[string]interface{})
+)
+
+// LocalInterface defines operations related to local repo under proxy mode
+type LocalInterface interface {
```
reasonerjt, line 39: Do you envision the Local/Remote Interfaces be used by anything else by proxy controller?

src/controller/proxy/remote.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	"github.com/docker/distribution"
+	"github.com/goharbor/harbor/src/replication/adapter"
+	"github.com/goharbor/harbor/src/replication/registry"
+	"io"
+)
+
+// RemoteInterface defines operations related to remote repository under proxy
+type RemoteInterface interface {
```
reasonerjt, line 25: Does it need to be exposed?

src/controller/proxy/local.go
reasonerjt: +1 Why not move the code of init into Create.

src/controller/proxy/local.go
```diff
@@ -0,0 +1,181 @@
...
+	DeleteManifest(ctx context.Context, repo, ref string)
+}
+
+// local defines operations related to local repo under proxy mode
+type local struct {
+	adapter *base.Adapter
+}
+
+// CreateLocalInterface create the LocalInterface
+func CreateLocalInterface() LocalInterface {
```
reasonerjt, line 60: `CreatexxxxInterface` is a bad func name.

src/controller/proxy/remote.go
```diff
@@ -0,0 +1,84 @@
...
+	// BlobReader create a reader for remote blob
+	BlobReader(orgRepo, dig string) (int64, io.ReadCloser, error)
```
reasonerjt, line 27: what's the difference between `orgRepo` and `repository` in the other funcs of this interface??

src/controller/proxy/remote.go
reasonerjt: Looks like this interface can be replaced by a func to fetch registry adapter by ID.

src/controller/proxy/local.go
```diff
@@ -0,0 +1,181 @@
...
+		}
+		if len(existMans) > 0 {
+			// Avoid empty manifest in the manifest list
+			return manifestlist.FromDescriptors(existMans)
+		}
+	}
+	return manifest, nil
+}
+
+func (l *local) PushManifestList(ctx context.Context, p *models.Project, repo string, tag string, art lib.ArtifactInfo, man distribution.Manifest) error {
```
reasonerjt, line 158: art is not used.

src/controller/proxy/local.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// updateManifestList -- Trim the manifest list, make sure all depend manifests are ready
+func (l *local) updateManifestList(ctx context.Context, manifest distribution.Manifest) (distribution.Manifest, error) {
+	switch v := manifest.(type) {
+	case *manifestlist.DeserializedManifestList:
+		existMans := make([]manifestlist.ManifestDescriptor, 0)
+		for _, m := range v.Manifests {
+			exist, err := l.BlobExist(ctx, string(m.Digest))
+			if err != nil {
```
reasonerjt, line 143: why continue here?

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,230 @@
...
+	ProxyManifest(ctx context.Context, p *models.Project, repo string, art lib.ArtifactInfo, w http.ResponseWriter, r RemoteInterface) error
+}
+
+type controller struct {
+	blobCtl     blob.Controller
+	registryMgr registry.Manager
+	mu          sync.Mutex
+	inflight    map[string]interface{}
+	artifactCtl artifact.Controller
+	local       LocalInterface
```
reasonerjt, line 67: any reason you make local part of the struct but remote not part of it?

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,230 @@
...
+	// Ctl is a global proxy controller instance
+	Ctl = NewController()
+)
+
+// Controller defines the operations related with pull through proxy
+type Controller interface {
+	// UseLocalManifest check if the manifest should use local
+	UseLocalManifest(ctx context.Context, p *models.Project, art lib.ArtifactInfo) bool
+	// UseLocalBlob check if the blob should use local
+	UseLocalBlob(ctx context.Context, p *models.Project, digest string) bool
```
ywk253100, line 53: We should check the existence of the blob under the repository. Seems this function isn't according to the parameters

src/server/middleware/repoproxy/proxy.go
```diff
@@ -0,0 +1,84 @@
...
+		if err != nil {
+			log.Errorf("failed to get project, error:%v", err)
+		}
+		if proxy.Ctl.UseLocalBlob(ctx, p, art.Digest) {
+			next.ServeHTTP(w, r)
+			return
+		}
+		log.Debugf("the blob doesn't exist, proxy the request to the target server, url:%v", repo)
+		remote := proxy.CreateRemoteInterface(p.RegistryID)
+		err = proxy.Ctl.ProxyBlob(ctx, p, repo, art.Digest, w, remote)
```
ywk253100, line 49: How about including `proxy.CreateRemoteInterface(p.RegistryID)` into the implementation of `proxy.Ctl.ProxyBlob`? The `RemoteInterface` isn't needed by the middleware and it can be the internal implementation detail of proxy controller

src/server/middleware/repoproxy/proxy.go
```diff
@@ -0,0 +1,86 @@
...
+	return middleware.New(func(w http.ResponseWriter, r *http.Request, next http.Handler) {
+		log.Debugf("Request url is %v", r.URL)
+		urlStr := r.URL.String()
+		log.Debugf("getting blob with url: %v\n", urlStr)
+		ctx := r.Context()
+		art := lib.GetArtifactInfo(ctx)
+		repo := utils.TrimProxyPrefix(art.ProjectName, art.Repository)
+		p, err := project.Ctl.GetByName(ctx, art.ProjectName, project.Metadata(false))
+		if err != nil {
+			log.Errorf("failed to get project, error:%v", err)
```
ywk253100, line 40: The `log.Errorf` isn't needed, the error will be logged in the `serror.SendError(w, err)`

src/server/middleware/repoproxy/proxy.go
```diff
@@ -0,0 +1,86 @@
...
+			serror.SendError(w, err)
```
ywk253100, line 41: Return?

src/server/middleware/repoproxy/proxy.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+		if proxy.Ctl.UseLocalBlob(ctx, p, art.Digest) {
+			next.ServeHTTP(w, r)
+			return
+		}
+		log.Debugf("the blob doesn't exist, proxy the request to the target server, url:%v", repo)
+		remote := proxy.CreateRemoteInterface(p.RegistryID)
+		err = proxy.Ctl.ProxyBlob(ctx, p, repo, art.Digest, w, remote)
+		if err != nil {
+			log.Errorf("failed to proxy the request, error %v", err)
```
ywk253100, line 51: The log.Errorf isn't needed, the error will be logged in the serror.SendError(w, err)

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,224 @@
...
+	reg, err := c.registryMgr.Get(p.RegistryID)
+	if err != nil {
+		log.Errorf("failed to get registry, error:%v", err)
+		return false
+	}
+	return reg.Status == model.Healthy
+}
+
+func (c *controller) UseLocalManifest(ctx context.Context, p *models.Project, art lib.ArtifactInfo) bool {
+	if !c.isProxyReady(p) {
```
ywk253100, line 82: The checking is useless? We can just check the existence in local?

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,224 @@
...
+	if err != nil {
+		return false
+	}
+	return exist
+}
+
+func (c *controller) ProxyManifest(ctx context.Context, p *models.Project, repo string, art lib.ArtifactInfo, w http.ResponseWriter, r RemoteInterface) error {
+	var man distribution.Manifest
+	var err error
+	ref := string(art.Digest)
```
ywk253100, line 108: Why type converting?

src/controller/proxy/controller.go
ywk253100: Maybe these two functions can be defined as:
```go
UseLocalManifest(ctx context.Context, repository, reference string)
UseLocalBlob(ctx context.Context, repository, digest string)
```

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,224 @@
...
+	if len(ref) == 0 {
+		ref = art.Tag
+	}
+	if len(ref) == 0 {
+		ref = "latest"
```
ywk253100, line 113: The `ref` is either digest or tag, it cannot be null. So the logic here isn't needed and correct.

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+	man, err = r.Manifest(repo, ref)
+	if err != nil {
+		if errors.IsNotFoundErr(err) {
+			go func() {
+				c.local.DeleteManifest(ctx, repo, string(art.Tag))
```
ywk253100, line 119: This isn't needed as the `ProxyManifest` is called only when the manifest isn't in local

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,224 @@
...
+	if written != size {
+		e := errors.Errorf("The size mismatch, actual:%d, expected: %d", written, size)
+		return e
+	}
+	desc.Size = size
+	desc.Digest = digest.Digest(dig)
+
+	setHeaders(w, size, desc.MediaType, dig)
+	go func() {
+		err := c.putBlobToLocal(ctx, p, repo, p.Name+"/"+repo, desc, r)
```
ywk253100, line 156: Put this before `written, err := io.CopyN(w, bReader, size)`  so that the two process can run concurrently

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,224 @@
...
+	if len(mediaType) > 0 {
+		w.Header().Set("Content-Type", mediaType)
+	}
+	w.Header().Set("Docker-Content-Digest", dig)
+	w.Header().Set("Etag", dig)
+}
+
+func (c *controller) waitAndPushManifest(ctx context.Context, p *models.Project, repo, tag string, man distribution.Manifest, art lib.ArtifactInfo, contType string, r RemoteInterface) {
+	localRepo := art.ProjectName + "/" + repo
+	if contType == manifestlist.MediaTypeManifestList {
```
ywk253100, line 198: These are two kinds of media types for manifest list: docker registry manifest list and OCI index

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,224 @@
...
+		time.Sleep(sleepIntervalSec * time.Second)
+		waitBlobs := c.local.CheckDependencies(ctx, man, string(art.Digest), contType)
+		if len(waitBlobs) == 0 {
+			break
+		}
+		log.Debugf("Current n=%v", n)
+		if n+1 == maxWait && len(waitBlobs) > 0 {
+			log.Debug("Waiting blobs not empty, push it to local repo manually")
+			for _, desc := range waitBlobs {
+				c.putBlobToLocal(ctx, p, repo, localRepo, desc, r)
```
ywk253100, line 216: Handle error?

ywk253100, line 216: Add more comment to explain why

src/controller/proxy/controller.go
ywk253100: Should we consider that the index references another index?

src/controller/proxy/remote.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// remote defines operations related to remote repository under proxy
+type remote struct {
+	regID    int64
+	registry adapter.ArtifactRegistry
+}
+
+// CreateRemoteInterface create a remote interface
+func CreateRemoteInterface(regID int64) RemoteInterface {
```
ywk253100, line 39: CreateRemoteInterface-> Newxxx

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,224 @@
...
+	ProxyBlob(ctx context.Context, p *models.Project, repo string, dig string, w http.ResponseWriter, r RemoteInterface) error
+	// ProxyManifest proxy the manifest to the target server
+	ProxyManifest(ctx context.Context, p *models.Project, repo string, art lib.ArtifactInfo, w http.ResponseWriter, r RemoteInterface) error
+}
+
+type controller struct {
+	blobCtl     blob.Controller
+	registryMgr registry.Manager
+	mu          sync.Mutex
+	inflight    map[string]interface{}
```
ywk253100, line 64: Delete inflight and mu

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,225 @@
...
+	registryMgr registry.Manager
+	artifactCtl artifact.Controller
+	local       localInterface
+}
+
+// ControllerInstance -- Get the proxy controller instance
+func ControllerInstance() (Controller, error) {
+	// Lazy load the controller
+	// Because LocalHelper is not ready unless core startup completely
+	ctlLock.Lock()
```
ywk253100, line 69: Put the locking logic inside `if ctl == nil {}` and double check whether the `ctl` is nil. Your current implementation serializes all requests. 
```go
if ctl != nil{
    return ctl, nil
}
lock.Lock()
defer lock.Unlock()
if ctl == nil{
    xxx
}
return ctl, nil
```

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,225 @@
...
+var (
+	// Ctl is a global proxy controller instance
+	ctl     Controller
+	ctlLock sync.Mutex
+)
+
+// Controller defines the operations related with pull through proxy
+type Controller interface {
+	// UseLocal check if the manifest should use localHelper
+	UseLocal(ctx context.Context, art lib.ArtifactInfo) bool
```
ywk253100, line 52: Based on our discussion, we should put the logic of `UseLocal` into the middleware?

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,225 @@
...
+			artifactCtl: artifact.Ctl,
+			local:       helper,
+		}
+	}
+	return ctl, nil
+}
+
+func (c *controller) UseLocal(ctx context.Context, art lib.ArtifactInfo) bool {
+	if len(art.Digest) > 0 {
+		exist, err := c.local.BlobExist(ctx, art.Digest)
```
ywk253100, line 88: The existence checking logic for manifest and blob are differenet

src/controller/proxy/local.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	return l, nil
+}
+
+func (l *localHelper) BlobExist(ctx context.Context, dig string) (bool, error) {
+	// not using l.registry.BlobExist(repo, dig)
+	// sometimes the blob exist in storage,
+	// but it does not exist in the Harbor database.
+	// if push manifest to Harbor, it still get failed
+	return blob.Ctl.Exist(ctx, dig)
```
ywk253100, line 69: We cannot check the existence of blob by calling `blob.Ctl.Exist()`

src/lib/context.go
```diff
@@ -40,6 +41,11 @@ type ArtifactInfo struct {
 	BlobMountDigest      string
 }
 
+// ProxyCacheRemoteRepo get the remote repository name, used in proxy cache
+func (a *ArtifactInfo) ProxyCacheRemoteRepo() string {
```
ywk253100, line 45: All functions defined in this file are context related, but `ProxyCacheRemoteRepo` isn't

src/controller/proxy/local.go
```diff
@@ -0,0 +1,184 @@
...
+			AccessSecret: config.ProxyServiceSecret,
+		},
+		TokenServiceURL: registryURL + "/service/token",
+		Insecure:        true,
+	}
+	baseAdapter, err := base.New(reg)
+	if err != nil {
+		return err
+	}
+	adp := v2.New(baseAdapter)
```
ywk253100, line 94: Use the factory to create the adapter instead

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,225 @@
...
+	if written != size {
+		return errors.Errorf("The size mismatch, actual:%d, expected: %d", written, size)
+	}
+	desc.Size = size
+	desc.Digest = digest.Digest(art.Digest)
+
+	setHeaders(w, size, desc.MediaType, art.Digest)
+	// put blob to localHelper will start another connection to the remoteHelper,
+	// to reduce the impact of it, cache the blob after it send to the client
+	go func() {
```
ywk253100, line 159: Put it before writing response to client?

src/lib/context.go
reasonerjt: +1 let's add a package `/src/pkg/proxycache

src/controller/proxy/remote.go
stonezdj: change name to remoteRepo and localRepo

src/controller/proxy/controller.go

src/controller/proxy/controller.go
stonezdj: add comment here

src/controller/proxy/controller.go
stonezdj: There are different,  UseLocalBlob verify based on BlobExist and UseLocalManifest based on ManifestExist

src/server/middleware/repoproxy/proxy.go
stonezdj: Removed

src/controller/proxy/controller.go

src/controller/proxy/controller.go

src/controller/proxy/controller.go
stonezdj: Add more comments here

src/controller/proxy/local.go
stonezdj: change to registry.Client

src/controller/proxy/controller.go
stonezdj: Add check for OCI index

src/controller/proxy/controller.go
stonezdj: Move to the middleware

src/controller/proxy/controller.go
stonezdj: Will updated, keep it for debug quickly

stonezdj: Will updated it before release, keep it for debug quickly

src/lib/context.go
stonezdj: Move to controller proxy  package

src/controller/proxy/controller.go
stonezdj: > Should we consider that the index references another index?

Current we don't handle it, we can handle it case by case in future.

src/server/middleware/repoproxy/proxy.go
```diff
@@ -0,0 +1,117 @@
...
+			return
+		}
+	})
+}
+
+func canProxy(p *models.Project) bool {
+	if p.RegistryID < 1 {
+		return false
+	}
+	reg, err := registry.NewDefaultManager().Get(p.RegistryID)
```
reasonerjt, line 107: In this package it only needs one registry manager

src/server/middleware/repoproxy/proxy.go
```diff
@@ -0,0 +1,117 @@
...
+			return
+		}
+
+		proxyCtl, err := proxy.ControllerInstance()
+		if err != nil {
+			httpLib.SendError(w, err)
+			return
+		}
+
+		if canProxy(p) == false || proxyCtl.UseLocalBlob(ctx, art) {
```
reasonerjt, line 55: !canProxy(p)

src/server/middleware/repoproxy/proxy.go
```diff
@@ -0,0 +1,117 @@
...
+		err = proxyCtl.ProxyBlob(ctx, p, art, w)
+		if err != nil {
+			httpLib.SendError(w, err)
+			return
+		}
+		return
+	})
+}
+
+// ManifestGetMiddleware middleware handle request for get blob request
```
reasonerjt, line 68: wrong comment

src/server/middleware/repoproxy/proxy.go
```diff
@@ -0,0 +1,117 @@
...
+func canProxy(p *models.Project) bool {
+	if p.RegistryID < 1 {
+		return false
+	}
+	reg, err := registry.NewDefaultManager().Get(p.RegistryID)
+	if err != nil {
+		log.Errorf("failed to get registry, error:%v", err)
+		return false
+	}
+	return reg.Status == model.Healthy
```
reasonerjt, line 112: Can we print some log if the  reg is unhealthy

src/server/middleware/repoproxy/proxy.go
```diff
@@ -0,0 +1,117 @@
...
+			next.ServeHTTP(w, r)
+			return
+		}
+
+		if !validProxyRepo(art.Repository) {
+			httpLib.SendError(w, errors.NotFoundError(fmt.Errorf("repository %v not found", art.Repository)))
+			return
+		}
+
+		log.Debugf("the digest is %v", art.Digest)
```
reasonerjt, line 94: also log the tag?

src/server/middleware/repoproxy/proxy.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	reg, err := registry.NewDefaultManager().Get(p.RegistryID)
+	if err != nil {
+		log.Errorf("failed to get registry, error:%v", err)
+		return false
+	}
+	return reg.Status == model.Healthy
+}
+
+func validProxyRepo(repo string) bool {
```
reasonerjt, line 115: Is this necessary, what if a remote registry doesn't require namespace?

src/controller/proxy/remote.go
```diff
@@ -0,0 +1,80 @@
...
+type remoteHelper struct {
+	regID    int64
+	registry adapter.ArtifactRegistry
+}
+
+// newRemoteHelper create a remoteHelper interface
+func newRemoteHelper(regID int64) remoteInterface {
+	r := &remoteHelper{regID: regID}
+	if err := r.init(); err != nil {
+		log.Errorf("failed to create remoteHelper error %v", err)
```
reasonerjt, line 44: in this case r.registry is nil and there will be nil pointer if you call the func under this interface.

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,241 @@
...
+				artifactCtl: artifact.Ctl,
+				local:       newLocalHelper(),
+			}
+		})
+	}
+	return ctl, nil
+}
+
+func (c *controller) UseLocalBlob(ctx context.Context, art lib.ArtifactInfo) bool {
+	if len(art.Digest) > 0 {
```
reasonerjt, line 87: Can we deal with the duplicated code between `UseLocalBlob` and `UseLocalManifest` ?

src/server/middleware/repoproxy/proxy.go
```diff
@@ -0,0 +1,117 @@
...
+
+		if !validProxyRepo(art.Repository) {
+			httpLib.SendError(w, errors.NotFoundError(fmt.Errorf("repository %v not found", art.Repository)))
+			return
+		}
+
+		log.Debugf("the digest is %v", art.Digest)
+		err = proxyCtl.ProxyManifest(ctx, p, art, w)
+		if err != nil {
+			httpLib.SendError(w, err)
```
ywk253100, line 97: Writing data to response should be done on api/middleware level.
How about returning both manifest and error for `ProxyManifest()` so that we'll not pass `w` as the input?
The same to proxy blob

src/controller/proxy/inflight.go
```diff
@@ -0,0 +1,47 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package proxy
+
+import "sync"
+
+type inflightRequest struct {
```
reasonerjt, line 19: I don't think this is correct, you also need to lock the map when reading? So here you need an RWlock and to add a func to check existence of the key

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,241 @@
...
+	registryMgr registry.Manager
+	artifactCtl artifact.Controller
+	local       localInterface
+}
+
+// ControllerInstance -- Get the proxy controller instance
+func ControllerInstance() (Controller, error) {
+	// Lazy load the controller
+	// Because LocalHelper is not ready unless core startup completely
+	if ctl == nil {
```
reasonerjt, line 73: since it's once, you don't need to check if it's nil or not

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,241 @@
...
+	once sync.Once
+)
+
+// Controller defines the operations related with pull through proxy
+type Controller interface {
+	// UseLocalBlob check if the blob should use local copy
+	UseLocalBlob(ctx context.Context, art lib.ArtifactInfo) bool
+	// UseLocalManifest check manifest should use local copy
+	UseLocalManifest(ctx context.Context, art lib.ArtifactInfo) bool
+	// ProxyBlob proxy the blob request to the remote server
```
reasonerjt, line 57: Could you add more comment to explain the input parms, esp the project `p`

src/controller/proxy/local.go
```diff
@@ -0,0 +1,182 @@
...
+	// CheckDependencies check if the manifest's dependency is ready
+	CheckDependencies(ctx context.Context, repo string, man distribution.Manifest) []distribution.Descriptor
+	// DeleteManifest cleanup delete tag from local cache
+	DeleteManifest(repo, ref string)
+}
+
+func (l *localHelper) ManifestExist(ctx context.Context, art lib.ArtifactInfo) (bool, error) {
+	a, err := l.artifactCtl.GetByReference(ctx, art.Repository, art.Digest, nil)
+	if err != nil {
+		return false, nil
```
reasonerjt, line 52: log error

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,241 @@
...
+	r := newRemoteHelper(p.RegistryID)
+	ref := art.Digest
+	if len(ref) == 0 {
+		ref = art.Tag
+	}
+	man, err = r.Manifest(remoteRepo, ref)
+	if err != nil {
+		if errors.IsNotFoundErr(err) {
+			go func() {
+				c.local.DeleteManifest(remoteRepo, art.Tag)
```
reasonerjt, line 120: log errors？

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,241 @@
...
+		if n+1 == maxWait && len(waitBlobs) > 0 {
+			// docker client will skip to pull layers exist in localHelper
+			// these blobs is not exist in the proxy server
+			// it will cause the manifest dependency check always fail
+			// need to push these blobs before push manifest to avoid failure
+			log.Debug("Waiting blobs not empty, push it to localHelper remoteRepo directly")
+			for _, desc := range waitBlobs {
+				err := c.putBlobToLocal(remoteRepo, art.Repository, desc, r)
+				if err != nil {
+					log.Errorf("Failed to push blob to cache error: %v", err)
```
reasonerjt, line 226: cache?

src/controller/proxy/controller.go
```diff
@@ -0,0 +1,241 @@
...
+	desc.Size = size
+	desc.Digest = digest.Digest(art.Digest)
+
+	setHeaders(w, size, desc.MediaType, art.Digest)
+	// put blob to local will start another connection to the remote,
+	// to reduce the impact of it, cache the blob after it send to the client
+	go func() {
+		err := c.putBlobToLocal(remoteRepo, art.Repository, desc, r)
+		if err != nil {
+			log.Errorf("error while putting blob to localHelper, %v", err)
```
reasonerjt, line 173: is it local or localHelper?

src/controller/proxy/local.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/goharbor/harbor/src/core/config"
+	"github.com/goharbor/harbor/src/lib"
+	"github.com/goharbor/harbor/src/lib/log"
+	"github.com/goharbor/harbor/src/pkg/registry"
+	"io"
+	"time"
+)
+
+// localInterface defines operations related to local repo under proxy mode
+type localInterface interface {
```
reasonerjt, line 32: Why Pushblob, PushManifest and DeleteManifest does not have ctx as the parameter?

src/controller/proxy/local.go
reasonerjt: Could you explain the reason why?

I also noticed that l.ManifestExists will never return non-nil error....

src/controller/proxy/local.go
```diff
@@ -0,0 +1,182 @@
...
+			if exist {
+				existMans = append(existMans, m)
+			}
+		}
+		if len(existMans) > 0 {
+			// Avoid empty manifest in the manifest list
+			return manifestlist.FromDescriptors(existMans)
+		}
+	}
+	return manifest, nil
```
reasonerjt, line 145: If the `manifest` is a non-empty manifestlist and non of the referenced manifest exists, this func will return `manifest` with all referenced elements.

Is it the design?

src/controller/proxy/local.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+func (l *localHelper) PushManifestList(ctx context.Context, repo string, tag string, man distribution.Manifest) error {
```
reasonerjt, line 148: This may not block merging this PR, but the way to handle manifest list can be refined, could you open a new issue to track that work?

src/controller/proxy/local.go
```diff
@@ -0,0 +1,182 @@
...
+
+func (l *localHelper) PushBlob(localRepo string, desc distribution.Descriptor, bReader io.ReadCloser) error {
+	log.Debugf("Put blob to local registry, localRepo:%v, digest: %v", localRepo, desc.Digest)
+	err := l.registry.PushBlob(localRepo, string(desc.Digest), desc.Size, bReader)
+	return err
+}
+
+func (l *localHelper) PushManifest(repo string, tag string, manifest distribution.Manifest) error {
+	// Make sure there is only one go routing to push current artName to local repo
+	if len(tag) == 0 {
```
reasonerjt, line 97: what if request is to pull manifest by digest?

src/controller/proxy/local.go
stonezdj: If one of the depends manifest exist, then we can push the manifest list.

src/server/middleware/repoproxy/proxy.go
stonezdj: It requires to have a namespace/project

src/controller/proxy/local.go
stonezdj: it will be pushed and tagged with latest

src/controller/proxy/local.go
stonezdj: Update and return error

src/server/middleware/repoproxy/proxy.go
stonezdj: Update

src/server/middleware/repoproxy/proxy.go
stonezdj: upated

src/server/middleware/repoproxy/proxy.go
stonezdj: Updated

src/controller/proxy/local.go
stonezdj: Update to return error

src/controller/proxy/local.go
stonezdj: Usually the api called outside the pkg need to have a ctx, for internal method and it doesn't use the ctx, no need to add it.

src/controller/proxy/controller.go
stonezdj: Updated

src/controller/proxy/controller.go
stonezdj: Updated

src/controller/proxy/controller.go
stonezdj: the art.Tag will not be changed so it can be used in enclosure
error is already logged in DeleteManifest.

src/controller/proxy/local.go
stonezdj: Update

src/controller/proxy/controller.go
stonezdj: Updated

src/controller/proxy/controller.go
stonezdj: Updated

src/server/middleware/repoproxy/proxy.go
stonezdj: Updated

src/controller/proxy/remote.go
stonezdj: Add error in return param

src/server/middleware/repoproxy/proxy.go
stonezdj: Remove

src/server/middleware/repoproxy/proxy.go
stonezdj: Update

src/server/middleware/repoproxy/proxy.go
stonezdj: Update

src/controller/proxy/local.go
stonezdj: Updated to wait and check strategy

src/controller/proxy/local.go
```diff
@@ -0,0 +1,199 @@
...
+)
+
+var errManifestNotFound = errors.New("manifest list doesn't contain any pushed manifest")
+
+// localInterface defines operations related to local repo under proxy mode
+type localInterface interface {
+	// BlobExist check if the blob exist in local repo
+	BlobExist(ctx context.Context, art lib.ArtifactInfo) (bool, error)
+	// Manifest check if the manifest exist in local repo
+	ManifestExist(ctx context.Context, art lib.ArtifactInfo) (bool, error)
```
reasonerjt, line 39: It should not return error and the error should not be checked.

src/controller/proxy/local.go
```diff
@@ -0,0 +1,199 @@
...
+		if len(existMans) == 0 {
+			return nil, errManifestNotFound
+		}
+		return manifestlist.FromDescriptors(existMans)
+	}
+	log.Errorf("Current manifest list type is unknown, manifest type[%T], content [%+v]", manifest, manifest)
+	return manifest, nil
+}
+
+func (l *localHelper) PushManifestList(ctx context.Context, repo string, ref string, man distribution.Manifest) error {
```
reasonerjt, line 156: It should try best to wait till all manifests referenced by the manifest list is pulled.

stonezdj: Updated

src/controller/proxy/local.go
stonezdj: Updated

PR #12280, opened 2020-06-19
Implement task and execution manager

ywk253100:
Implement task and execution manager


src/core/config/config.go
```diff
@@ -231,7 +231,7 @@ func RegistryURL() (string, error) {
 
 // InternalJobServiceURL returns jobservice URL for internal communication between Harbor containers
 func InternalJobServiceURL() string {
-	return strings.TrimSuffix(cfgMgr.Get(common.JobServiceURL).GetString(), "/")
+	return os.Getenv("JOBSERVICE_URL")
```
bitsf, line 234: why not continue using common.JobServiceURL

src/pkg/task/execution.go
```diff
@@ -43,3 +53,211 @@ type ExecutionManager interface {
...
+			metrics.ErrorTaskCount = sc.Count
+		case job.PendingStatus.String():
+			metrics.PendingTaskCount = sc.Count
+		case job.RunningStatus.String():
+			metrics.RunningTaskCount = sc.Count
+		case job.ScheduledStatus.String():
+			metrics.ScheduledTaskCount = sc.Count
+		case job.StoppedStatus.String():
+			metrics.StoppedTaskCount = sc.Count
+		}
```
bitsf, line 222: how about add default clause and return error

src/core/middlewares/middlewares.go
```diff
@@ -82,9 +67,9 @@ func MiddleWares() []beego.MiddleWare {
 		csrf.Middleware(),
 		security.Middleware(),
 		readonly.Middleware(readonlySkippers...),
-		orm.Middleware(legacyAPISkipper),
+		orm.Middleware(),
```
bitsf, line 70: @heww Is this check not required now ?

heww: I think this skipper does not work for legacy APIs now.

src/pkg/task/execution.go
```diff
@@ -43,3 +53,211 @@ type ExecutionManager interface {
...
+		Keywords: map[string]interface{}{
+			"ExecutionID": id,
+		},
+	})
+	if err != nil {
+		return err
+	}
+
+	for _, task := range tasks {
+		if !job.Status(task.Status).Final() {
```
steven-zou, line 131: Is it possible to treat this operation as a default one? That means if the task is not completed, the deleting action will try to stop it first.

src/pkg/task/execution.go
steven-zou: default clause : just log error?

src/pkg/task/execution.go
```diff
@@ -43,3 +53,211 @@ type ExecutionManager interface {
...
+	metrics.TaskCount = metrics.SuccessTaskCount + metrics.ErrorTaskCount +
+		metrics.PendingTaskCount + metrics.RunningTaskCount +
+		metrics.ScheduledTaskCount + metrics.StoppedTaskCount
+	execution.Metrics = metrics
+}
+
+func (e *executionManager) populateExecutionStatus(execution *Execution) {
+	metrics := execution.Metrics
+	if metrics == nil {
+		return
```
steven-zou, line 233: What's the default status?

src/pkg/task/task.go
```diff
@@ -37,3 +51,168 @@ type Manager interface {
...
+		ExecutionID:  executionID,
+		Status:       job.PendingStatus.String(),
+		StatusCode:   job.PendingStatus.Code(),
+		ExtraAttrs:   string(data),
+		CreationTime: now,
+		UpdateTime:   now,
+	})
+}
+
+func (m *manager) submitJob(ctx context.Context, id int64, jb *Job) {
```
steven-zou, line 105: Maybe `submitJob` returns an error and status updating can be put into the `Create` method, which will be more clear.

src/common/job/client.go
```diff
@@ -18,10 +19,12 @@ import (
 
 var (
 	// GlobalClient is an instance of the default client that can be used globally
-	// Notes: the client needs to be initialized before can be used
-	GlobalClient             Client
-	statusBehindErrorPattern = "mismatch job status for stopping job: .*, job status (.*) is behind Running"
-	statusBehindErrorReg     = regexp.MustCompile(statusBehindErrorPattern)
+	GlobalClient             Client = NewDefaultClient(config.InternalJobServiceURL(), config.CoreSecret())
+	statusBehindErrorPattern        = "mismatch job status for stopping job: .*, job status (.*) is behind Running"
```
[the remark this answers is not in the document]
ywk253100: No change for this part in the PR, will remove it in a separated one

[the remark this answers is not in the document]
ywk253100: No change for this part in the PR, will remove it in a separated one

src/common/job/client.go
```diff
@@ -206,6 +204,9 @@ func (d *DefaultClient) PostAction(uuid, action string) error {
 				status: status,
 			}
 		}
+		if strings.Contains(err.Error(), "object is not found") {
```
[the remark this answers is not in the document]
ywk253100: Seems that the jobservice will not return 404 error?

src/core/config/config.go
ywk253100: The original implementation gets the URL from config manager which reads it from the environment variable`JOBSERVICE_URL` as well. The config manager depends on database when starting up, this is unnecessary for jobservice URL.

src/pkg/task/execution.go
ywk253100: A corner case: think about an execution which contains two tasks `a` and `b`, both of them are running. Now I try to delete the execution, stop `a` successfully, delete `a` from database, try to stop b but get error, as the operation is in a DB transaction, all database deletion rolls back, this causes the inconsistent between database and jobservice.

src/pkg/task/execution.go
ywk253100: No default status. If the execution contains any tasks, the status is calculated from the task statuses, or the status should be set by caller explicitly.

src/pkg/task/execution.go
ywk253100: Updated

src/common/job/client.go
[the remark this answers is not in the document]
steven-zou: Double check JS code, it will return 400 code if the error is NOT_FOUND

```go
if errs.IsObjectNotFoundError(err) {
			code = http.StatusNotFound
		} 
```

src/pkg/task/execution.go
steven-zou: ok, let's keep it with a simple way.

src/pkg/task/execution.go
steven-zou: By your design, there should not be an execution without any tasks, right? Shall we make sure this rule by adding validation logic?

src/common/job/client.go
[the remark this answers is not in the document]
steven-zou: Don't forget it.

src/pkg/task/execution.go
ywk253100: There are indeed executions contain no tasks. e.g. starting an execution for replication will create an execution record first, and after the initial/prepare/filter works, there maybe no artifact candidates to replicate, the execution contains no tasks in this case and the caller should set the status and end time for the execution record

PR #11829, opened 2020-05-05
upgrade aws-sdk-go to support OIDC feature

shivpathak:
**About this PR**

I wanted to use S3 backend for harbor registry storage.

In order to support k8s ServiceAccount feature with AWS IAM roles upgrading the aws-sdk-module.

PR #12281, opened 2020-06-19
Upgrade aws-sdk-go to 1.32.5 for OIDC support

pcfens:
This is the vendored version of #11829 (it looks to be in a weird state, so I opened a new one).

The primary motivation is the same - to support OIDC so that k8s service accounts can be given access to S3 and other AWS services as needed without providing keys.

Is there a process for selecting the version of a dependency? Rather than upgrading the library to the [minimum version](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-minimum-sdk.html) of 1.23.13 this PR upgrades aws-sdk-go to the latest version of 1.32.5.


reasonerjt:
@pcfens 

Thanks for this PR, let me clarify, this benefits the scenario where you deploy Harbor on EKS, so that you don't need to set the keys in configmap of registry correct?

Could you add more details in the commit message and I'll merge this PR.

Thanks for your patience.

pcfens:
Thanks @reasonerjt. I've added a little more detail to the commit message - let me know if you'd like any more.

PR #12314, opened 2020-06-23
add get GC candidate

wy65701436:

### DIFF
--- a/src/pkg/blob/dao/dao.go
+++ b/src/pkg/blob/dao/dao.go
@@ -400,7 +400,7 @@ func (d *dao) GetBlobsNotRefedByProjectBlob(ctx context.Context, timeWindowHours
 		return noneRefed, err
 	}
 
-	sql := fmt.Sprintf(`SELECT b.* FROM blob AS b LEFT JOIN project_blob pb ON b.id = pb.blob_id WHERE pb.id IS NULL AND b.update_time <= now()- interval '%d hours';`, timeWindowHours)
+	sql := fmt.Sprintf(`SELECT b.id, b.digest, b.content_type, b.status FROM blob AS b LEFT JOIN project_blob pb ON b.id = pb.blob_id WHERE pb.id IS NULL AND b.update_time <= now() - interval '%d hours';`, timeWindowHours)
 	_, err = ormer.Raw(sql).QueryRows(&noneRefed)
 	if err != nil {
 		return noneRefed, err
--- a/src/pkg/blob/dao/dao.go
+++ b/src/pkg/blob/dao/dao.go
@@ -75,6 +75,9 @@ type DAO interface {
 
 	// DeleteBlob delete blob
 	DeleteBlob(ctx context.Context, id int64) (err error)
+
+	// GetBlobsNotRefedByProjectBlob get the blobs that are not referenced by the table project_blob and also not in the reserve window(in hours)
+	GetBlobsNotRefedByProjectBlob(ctx context.Context, timeWindowHours int64) ([]*models.Blob, error)
 }
 
 // New returns an instance of the default DAO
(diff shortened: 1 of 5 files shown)


src/pkg/blob/dao/dao.go
```diff
@@ -389,3 +392,19 @@ func (d *dao) DeleteBlob(ctx context.Context, id int64) error {
...
 }
+
+func (d *dao) GetBlobsNotRefedByProjectBlob(ctx context.Context, timeWindow int64) ([]*models.Blob, error) {
+	var noneRefed []*models.Blob
+	ormer, err := orm.FromContext(ctx)
+	if err != nil {
+		return noneRefed, err
+	}
+
+	sql := fmt.Sprintf(`SELECT b.* FROM blob AS b LEFT JOIN project_blob pb ON b.id = pb.blob_id WHERE pb.id IS NULL AND b.update_time <= now()- interval '%d hours';`, timeWindow)
```
heww, line 403: Do we need pagination here because the result may very large?

src/pkg/blob/dao/dao.go
```diff
@@ -389,3 +392,19 @@ func (d *dao) DeleteBlob(ctx context.Context, id int64) error {
...
 }
+
+func (d *dao) GetBlobsNotRefedByProjectBlob(ctx context.Context, timeWindowHours int64) ([]*models.Blob, error) {
+	var noneRefed []*models.Blob
+	ormer, err := orm.FromContext(ctx)
+	if err != nil {
+		return noneRefed, err
+	}
+
+	sql := fmt.Sprintf(`SELECT b.* FROM blob AS b LEFT JOIN project_blob pb ON b.id = pb.blob_id WHERE pb.id IS NULL AND b.update_time <= now()- interval '%d hours';`, timeWindowHours)
```
reasonerjt, line 403: select *???

reasonerjt, line 403: should also be a space between `now()` and `-`

reasonerjt, line 403: Is "LEFT JOIN ...... WHERE xx IS NULL" a very common usage?  I would use a `NOT EXIST`...

src/pkg/blob/manager.go
```diff
@@ -129,6 +132,10 @@ func (m *manager) Delete(ctx context.Context, id int64) error {
 	return m.dao.DeleteBlob(ctx, id)
 }
 
+func (m *manager) UselessBlobs(ctx context.Context, timeWindowHours int64) ([]*models.Blob, error) {
+	return m.dao.GetBlobsNotRefedByProjectBlob(ctx, timeWindowHours)
```
reasonerjt, line 136: I think it would be a better design if timeWindowHousrs is handled at the manager level.
And let the func in `dao` pkg only cares about the timestamp where the update time should be no later than.

src/pkg/blob/dao/dao.go
wy65701436: ok

src/pkg/blob/manager.go
wy65701436: It needs to be set by GC job, and let GC job to give a value. And for debugging, it can be set to 0.

src/pkg/blob/dao/dao.go
wy65701436: image

wy65701436: do you mean pagination as Hww mentioned above?

src/pkg/blob/manager.go
wy65701436: and why not leveage Pg's buildin func `interval` to handle it

src/pkg/blob/dao/dao.go
wy65701436: add cloumn names.

src/pkg/blob/dao/dao.go
wy65701436: will file an issue, and consider to add it in the next sprint.

PR #12315, opened 2020-06-23
handle blob status chanage in put blob middlware

wy65701436:
After blob is uploaded success, the middleware will update the blob status accordingly.

### DIFF
--- a/src/server/middleware/blob/put_blob_upload.go
+++ b/src/server/middleware/blob/put_blob_upload.go
@@ -15,17 +15,54 @@
 package blob
 
 import (
-	"net/http"
-	"strconv"
-
+	"fmt"
+	"github.com/goharbor/harbor/src/controller/blob"
+	"github.com/goharbor/harbor/src/lib/errors"
 	"github.com/goharbor/harbor/src/lib/log"
+	blob_models "github.com/goharbor/harbor/src/pkg/blob/models"
 	"github.com/goharbor/harbor/src/pkg/distribution"
 	"github.com/goharbor/harbor/src/server/middleware"
+	"github.com/goharbor/harbor/src/server/middleware/requestid"
+	"net/http"
+	"strconv"
 )
 
 // PutBlobUploadMiddleware middleware to create Blob and ProjectBlob after PUT /v2/<name>/blobs/uploads/<session_id> success
 func PutBlobUploadMiddleware() func(http.Handler) http.Handler {
-	return middleware.AfterResponse(func(w http.ResponseWriter, r *http.Request, statusCode int) error {
+
+	before := middleware.BeforeRequest(func(r *http.Request) error {
+		v := r.URL.Query()
+		digest := v.Get("digest")
+
+		if digest == "" {
+			log.Warningf(fmt.Sprintf("the put blob request has no digest in query, %s", r.URL.String()))
+			return errors.New(nil).WithMessage(fmt.Sprintf("the put blob request has no digest in query, %s", r.URL.String()))
+		}
+
+		bb, err := blob.Ctl.Get(r.Context(), digest)
+		if err != nil {
+			if errors.IsNotFoundErr(err) {
+				return nil
+			}
+			return err
+		}
+
+		switch bb.Status {
+		case blob_models.StatusNone, blob_models.StatusDelete, blob_models.StatusDeleteFailed:
+			err := blob.Ctl.Touch(r.Context(), bb)
+			if err != nil {
+				log.Errorf("failed to update blob: %s status to StatusNone, error:%v", bb.Digest, err)
+				return errors.Wrapf(err, fmt.Sprintf("the request id is: %s", r.Header.Get(requestid.HeaderXRequestID)))
+			}
+		case blob_models.StatusDeleting:
+			return errors.New(nil).WithMessage(fmt.Sprintf("the asking blob is in GC, mark it as non existing, request id: %s", r.Header.Get(requestid.HeaderXRequestID))).WithCode(errors.NotFoundCode)
+		default:
+			return nil
+		}
+		return nil
+	})
+
+	after := middleware.AfterResponse(func(w http.ResponseWriter, r *http.Request, statusCode int) error {
 		if statusCode != http.StatusCreated {
 			return nil
 		}
@@ -63,4 +100,6 @@ func PutBlobUploadMiddleware() func(http.Handler) http.Handler {
 
 		return nil
 	})
+
+	return middleware.Chain(before, after)
 }
--- a/src/server/middleware/blob/put_blob_upload.go
+++ b/src/server/middleware/blob/put_blob_upload.go
@@ -27,18 +27,17 @@ import (
 	"strconv"
 )
 
-// PutBlobUploadMiddleware middleware to create Blob and ProjectBlob after PUT /v2/<name>/blobs/uploads/<session_id> success
+// PutBlobUploadMiddleware middleware to create Blob and ProjectBlob after PUT /v2/<name>/blobs/uploads/<session_id>?digest=<digest> success
 func PutBlobUploadMiddleware() func(http.Handler) http.Handler {
 
 	before := middleware.BeforeRequest(func(r *http.Request) error {
+		ctx := r.Context()
+		logger := log.G(ctx)
+
 		v := r.URL.Query()
 		digest := v.Get("digest")
 
-		if digest == "" {
-			log.Warningf(fmt.Sprintf("the put blob request has no digest in query, %s", r.URL.String()))
-			return errors.New(nil).WithMessage(fmt.Sprintf("the put blob request has no digest in query, %s", r.URL.String()))
-		}
-
+		// digest empty is handled by the blob controller GET method
 		bb, err := blob.Ctl.Get(r.Context(), digest)
 		if err != nil {
 			if errors.IsNotFoundErr(err) {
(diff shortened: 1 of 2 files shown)


src/server/middleware/blob/put_blob_upload.go
```diff
@@ -63,4 +101,8 @@ func PutBlobUploadMiddleware() func(http.Handler) http.Handler {
 
 		return nil
 	})
+
+	return func(next http.Handler) http.Handler {
+		return alice.New(before, after).Then(next)
+	}
```
heww, line 107: There is a function named `Chain` in the middleware pkg which does the same thing as `alice.New`.

src/server/middleware/blob/put_blob_upload.go
```diff
@@ -15,17 +15,55 @@
...
+
+		switch bb.Status {
+		case blob_models.StatusNone, blob_models.StatusDelete, blob_models.StatusDeleteFailed:
+			err := blob.Ctl.Touch(r.Context(), bb)
+			if err != nil {
+				log.Errorf("failed to update blob: %s status to StatusNone, error:%v", bb.Digest, err)
+				return errors.Wrapf(err, fmt.Sprintf("the request id is: %s", r.Header.Get(requestid.HeaderXRequestID)))
+			}
+		case blob_models.StatusDeleting:
+			return errors.New(nil).WithMessage(fmt.Sprintf("the asking blob is in GC, mark it as non existing, request id: %s", r.Header.Get(requestid.HeaderXRequestID))).WithCode(errors.NotFoundCode)
```
heww, line 59: Should we wait a moment and retry to check the status?

src/server/middleware/blob/put_blob_upload.go
wy65701436: updated.

src/server/middleware/blob/put_blob_upload.go
```diff
@@ -15,17 +15,54 @@
...
 	"github.com/goharbor/harbor/src/lib/log"
+	blob_models "github.com/goharbor/harbor/src/pkg/blob/models"
 	"github.com/goharbor/harbor/src/pkg/distribution"
 	"github.com/goharbor/harbor/src/server/middleware"
+	"github.com/goharbor/harbor/src/server/middleware/requestid"
+	"net/http"
+	"strconv"
 )
 
 // PutBlobUploadMiddleware middleware to create Blob and ProjectBlob after PUT /v2/<name>/blobs/uploads/<session_id> success
```
reasonerjt, line 30: The comment should be updated too

src/server/middleware/blob/put_blob_upload.go
```diff
@@ -15,17 +15,54 @@
...
 func PutBlobUploadMiddleware() func(http.Handler) http.Handler {
-	return middleware.AfterResponse(func(w http.ResponseWriter, r *http.Request, statusCode int) error {
+
+	before := middleware.BeforeRequest(func(r *http.Request) error {
+		v := r.URL.Query()
+		digest := v.Get("digest")
+
+		if digest == "" {
+			log.Warningf(fmt.Sprintf("the put blob request has no digest in query, %s", r.URL.String()))
+			return errors.New(nil).WithMessage(fmt.Sprintf("the put blob request has no digest in query, %s", r.URL.String()))
```
reasonerjt, line 39: Will this yield a 500?  It should be a 4xx

src/server/middleware/blob/put_blob_upload.go
```diff
@@ -15,17 +15,54 @@
...
+				return nil
+			}
+			return err
+		}
+
+		switch bb.Status {
+		case blob_models.StatusNone, blob_models.StatusDelete, blob_models.StatusDeleteFailed:
+			err := blob.Ctl.Touch(r.Context(), bb)
+			if err != nil {
+				log.Errorf("failed to update blob: %s status to StatusNone, error:%v", bb.Digest, err)
```
reasonerjt, line 54: You should use the logger with context so the message will also have the req ID.

wy65701436: updated, thanks.

src/server/middleware/blob/put_blob_upload.go
wy65701436: return 400 instead.

src/server/middleware/blob/put_blob_upload.go
wy65701436: updated.

src/server/middleware/blob/put_blob_upload.go
```diff
@@ -15,17 +15,53 @@
...
+	blob_models "github.com/goharbor/harbor/src/pkg/blob/models"
 	"github.com/goharbor/harbor/src/pkg/distribution"
 	"github.com/goharbor/harbor/src/server/middleware"
+	"github.com/goharbor/harbor/src/server/middleware/requestid"
+	"net/http"
+	"strconv"
 )
 
-// PutBlobUploadMiddleware middleware to create Blob and ProjectBlob after PUT /v2/<name>/blobs/uploads/<session_id> success
+// PutBlobUploadMiddleware middleware to create Blob and ProjectBlob after PUT /v2/<name>/blobs/uploads/<session_id>?digest=<digest> success
```
reasonerjt, line 30: The comment is still incorrect, this middleware is also handling the request before it's successful.

src/server/middleware/blob/put_blob_upload.go
```diff
@@ -15,17 +15,53 @@
...
+		v := r.URL.Query()
+		digest := v.Get("digest")
+
+		// digest empty is handled by the blob controller GET method
+		bb, err := blob.Ctl.Get(r.Context(), digest)
+		if err != nil {
+			if errors.IsNotFoundErr(err) {
+				return nil
+			}
+			return err
```
reasonerjt, line 46: Will such error yield a 4xx? or 5xx? if the request does not contain digest in query string the server should return a 4xx not 5xx

src/server/middleware/blob/put_blob_upload.go
```diff
@@ -15,17 +15,53 @@
...
+		}
+
+		switch bb.Status {
+		case blob_models.StatusNone, blob_models.StatusDelete, blob_models.StatusDeleteFailed:
+			err := blob.Ctl.Touch(r.Context(), bb)
```
reasonerjt, line 51: If its status is `None` you don't need to `Touch` it, correct?

src/server/middleware/blob/put_blob_upload.go
```diff
@@ -15,17 +15,53 @@
...
+			if err != nil {
+				logger.Errorf("failed to update blob: %s status to StatusNone, error:%v", bb.Digest, err)
+				return errors.Wrapf(err, fmt.Sprintf("the request id is: %s", r.Header.Get(requestid.HeaderXRequestID)))
+			}
+		case blob_models.StatusDeleting:
+			return errors.New(nil).WithMessage(fmt.Sprintf("the asking blob is in GC, mark it as non existing, request id: %s", r.Header.Get(requestid.HeaderXRequestID))).WithCode(errors.NotFoundCode)
```
reasonerjt, line 57: You should also log this situation.

src/server/middleware/blob/put_blob_upload.go
wy65701436: it can be discussed.

src/server/middleware/blob/put_blob_upload.go
wy65701436: updated.

src/server/middleware/blob/put_blob_upload.go
```diff
@@ -15,17 +15,53 @@
...
+
+	before := middleware.BeforeRequest(func(r *http.Request) error {
+		ctx := r.Context()
+		logger := log.G(ctx)
+
+		v := r.URL.Query()
+		digest := v.Get("digest")
+
+		// digest empty is handled by the blob controller GET method
+		bb, err := blob.Ctl.Get(r.Context(), digest)
```
reasonerjt, line 41: You should either use `blobController` or get rid of it from this package

src/server/middleware/blob/put_blob_upload.go
wy65701436: if you go into the code of blob controller get method, if the digest is empty, it will return a 400.

src/server/middleware/blob/put_blob_upload.go
wy65701436: log added.

src/server/middleware/blob/put_blob_upload.go
wy65701436: func (c *controller) Get(ctx context.Context, digest string, options ...Option) (*blob.Blob, error) {
	if digest == "" {
		return nil, errors.New(nil).WithCode(errors.BadRequestCode).WithMessage("require digest")
	}

	opts := newOptions(options...)

	params := blob.ListParams{
		ArtifactDigest: opts.ArtifactDigest,
		BlobDigests:    []string{digest},
		ProjectID:      opts.ProjectID,
	}

	blobs, err := c.blobMgr.List(ctx, params)
	if err != nil {
		return nil, err
	} else if len(blobs) == 0 {
		return nil, errors.NotFoundError(nil).WithMessage("blob %s not found", digest)
	}

	return blobs[0], nil
}

src/server/middleware/blob/put_blob_upload.go
wy65701436: updated.

src/server/middleware/blob/put_blob_upload.go
reasonerjt: ok

PR #12333, opened 2020-06-26
Format op_time JSON field in audit logs as RFC 3339

### LINKED ISSUE #11522: [Audit log]Should return stardard time string
What can we help you?
For both system level audit log api  and  project level   audit log api,

Current  time string  returned for example:
image

Expected:
image

nielsslot:
Fixes #11522 

Signed-off-by: Niels Slot <nielsslot@gmail.com>


wy65701436:
thanks for the PR, the root cause of the issue is the wrong format defined in the swagger, and it's fixed by https://github.com/goharbor/harbor/pull/12629. I'll close it and appreciate for the effort.

PR #12336, opened 2020-06-26
Require valid user to show version information on login page, move 'More info..' to About

JasonEverling:
A few minor cosmetic updates,

1. Moves the 'More Info...' link on the Login Form to the About-Dialog
2. Removes Version from About-Dialog, Requires valid user before being displayed


JasonEverling:
We want to use Harbor outside of our network and would like to expose it to the internet, this will at least make it harder for an attacker from knowing what version of Harbor running. I also moved the 'More Info' from the Login Form, I had users that kept clicking on it thinking it would give them different sign in options, it should be in About anyways because that is what it redirects to,

Thanks for considering!
Jason

PR #12337, opened 2020-06-27
feat: add p2p preheat policy model

chlins:

### DIFF
--- a/src/pkg/p2p/preheat/models/policy/policy.go
+++ b/src/pkg/p2p/preheat/models/policy/policy.go
@@ -1,0 +1,150 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


src/pkg/p2p/preheat/models/policy/policy.go
```diff
@@ -0,0 +1,143 @@
...
+	FilterTypeLabel         FilterType = "label"
+
+	// trigger types
+	TriggerTypeManual     TriggerType = "manual"
+	TriggerTypeScheduled  TriggerType = "scheduled"
+	TriggerTypeEventBased TriggerType = "event_based"
+)
+
+// Policy defines p2p preheat policy
+type Policy struct {
```
steven-zou, line 47: policy.Policy? Do we have a better name?

How about ` policy.Object` \ `policy.Definition` \ `policy.Schema`? / policy.?

chlins: I think policy.Schema is good!

kofj:
LGTM

PR #12338, opened 2020-06-27
feat(preheat):add preheat job

steven-zou:
- implement the preheat job
- add relevant UT case
- fix issue #12285

### DIFF
--- a/src/pkg/p2p/preheat/job.go
+++ b/src/pkg/p2p/preheat/job.go
@@ -1,0 +1,242 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package preheat
+
+import (
+	"time"
+
+	"github.com/goharbor/harbor/src/jobservice/job"
+	"github.com/goharbor/harbor/src/lib/errors"
+	"github.com/goharbor/harbor/src/pkg/p2p/preheat/models/provider"
+	pr "github.com/goharbor/harbor/src/pkg/p2p/preheat/provider"
+	"github.com/goharbor/harbor/src/pkg/p2p/preheat/provider/auth"
+)
+
+const (
+	// parameter keeps the preheating provider instance info.
+	preheatParamProvider = "provider"
+	// parameter keeps the preheating artifact (image) info.
+	preheatParamImage = "image"
+	// checkInterval indicates the interval of loop check.
+	checkInterval = 10 * time.Second
+	// checkTimeout indicates the overall timeout of the loop check.
+	checkTimeout = 1801 * time.Second
+)
+
+// Job preheats the given artifact(image) to the target preheat provider.
+type Job struct{}
+
+// MaxFails of preheat job. Don't need to retry.
+func (j *Job) MaxFails() uint {
+	return 1
+}
+
+// MaxCurrency indicates no limitation to the concurrency of preheat job.
+func (j *Job) MaxCurrency() uint {
+	return 0
+}
+
+// ShouldRetry indicates no need to retry preheat job as it's just for a cache purpose.
+func (j *Job) ShouldRetry() bool {
+	return false
+}
+
+// Validate the parameters of preheat job.
+func (j *Job) Validate(params job.Parameters) error {
+	_, err := parseParamProvider(params)
+	if err != nil {
+		return err
+	}
+
+	_, err = parseParamImage(params)
+
+	return err
+}
+
+// Run the preheat process.
+func (j *Job) Run(ctx job.Context, params job.Parameters) error {
+	// Get logger
+	myLogger := ctx.GetLogger()
+
+	// Parse parameters, ignore errors as they have been validated already
+	p, _ := parseParamProvider(params)
+	pi, _ := parseParamImage(params)
+
+	// Print related info to log first
+	myLogger.Infof(
+		"Preheating image '%s:%s' to the target preheat provider: %s %s:%s",
+		pi.ImageName,
+		pi.Tag,
+		p.Vendor,
+		p.Name,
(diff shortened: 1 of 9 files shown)


src/pkg/p2p/preheat/job.go
```diff
@@ -0,0 +1,225 @@
...
+
+const (
+	// parameter keeps the preheating provider instance info.
+	preheatParamProvider = "provider"
+	// parameter keeps the preheating artifact (image) info.
+	preheatParamImage = "image"
+	// checkInterval indicates the interval of loop check.
+	checkInterval = 10 * time.Second
+	// checkTimeout indicates the overall timeout of the loop check.
+	checkTimeout = 1801 * time.Second
```
chlins, line 35: I'm confused as to why checkTimeout use 1801, why add one seconds?

steven-zou: avoid the two time-signals coming at the same time. I remembered if they come at the same time, a random selection may happen. @chlins 

select{
  case <-tk.C:
  case <-tm.C:
}

src/pkg/p2p/preheat/job.go
```diff
@@ -0,0 +1,225 @@
...
+		return preheatJobRunningError(err)
+	}
+
+	myLogger.Infof("Get preheat provider driver: %s", p.Vendor)
+
+	// Start the preheat process
+	// First, check the health of the provider
+	h, err := d.GetHealth()
+	if err != nil {
+		return preheatJobRunningError(err)
```
mmpei, line 108: maybe we should record all error info in job log. so users can check the failure reason from UI.

mmpei, line 108: maybe we should record all error info in job log. so users can check the failure reason from UI.

src/pkg/p2p/preheat/job.go
```diff
@@ -0,0 +1,225 @@
...
+	return err
+}
+
+// Run the preheat process.
+func (j *Job) Run(ctx job.Context, params job.Parameters) error {
+	// Get logger
+	myLogger := ctx.GetLogger()
+
+	// Parse parameters, ignore errors as they have been validated already
+	p, _ := parseParamProvider(params)
```
mmpei, line 74: image that oneday somebody change the validate logic. this will be a risk.

src/pkg/p2p/preheat/job.go
```diff
@@ -0,0 +1,225 @@
...
+		select {
+		case <-tk.C:
+			s, err := d.CheckProgress(st.TaskID)
+			if err != nil {
+				myLogger.Error(err)
+				return preheatJobRunningError(err)
+			}
+
+			// Finished
+			if s.Status == provider.PreheatingStatusSuccess {
```
mmpei, line 150: Status will nerver be PreheatingStatusFail?

src/pkg/p2p/preheat/provider/preheat_image.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/pkg/errors"
+)
+
+const (
+	// SupportedType indicates the supported preheating type 'image'.
+	SupportedType = "image"
+)
+
+// PreheatImage contains related information which can help providers to get/pull the images.
+type PreheatImage struct {
```
mmpei, line 30: PreheatImage is still in use? not Artifact?

src/pkg/p2p/preheat/job.go
steven-zou: ? I don't understand your concern.

The job.Validate() is called before running the job by the jobservice, so it's safe to ignore the error because here sharing the same validation func with the job.Validate() method.

src/pkg/p2p/preheat/job.go
steven-zou: You should check your Kraken driver logic. This is a specific logic for Kraken driver.

In the Kraken driver comments, if preheat is failed, an error will be returned. Otherwise, it will definitely return success.

src/pkg/p2p/preheat/provider/preheat_image.go
steven-zou: I change my mind. `Artifact` is too heavy for such a case and we need to do lots of changes. Preheat image is enough for such a case and we can avoid more changes and then save efforts.

src/pkg/p2p/preheat/job.go
steven-zou: correct, missing the logs here

steven-zou: correct, missing the logs here

src/pkg/p2p/preheat/job.go
steven-zou: Anyway, enhance the logic for more flexible cases in the future

src/pkg/p2p/preheat/job.go
mmpei: i mean that we should keep the code independent. this will avoid the rist when the dependency changed.

mmpei: i mean that we should keep the code independent. this will avoid the rist when the dependency changed.

src/pkg/p2p/preheat/job.go
mmpei: it is not related to Kraken but the design. we have designed two final status PreheatingStatusSuccess and PreheatingStatusFail. so we should check these two status both here.

steven-zou: I totally understood. So far, this specific logic actually is only for Kraken case as Dragonfly will never return other statuses(it will return `pending`). Kraken will return error or non-error and Success.

Anyway, enhance the logic to support the original design.

steven-zou: Dragonfly: error or non-error & pending status

```go
bytes, err := client.GetHTTPClient(dd.instance.Insecure).Post(url, dd.getCred(), preheatingImage, nil)
	if err != nil {
		return nil, err
	}

	result := &dragonflyPreheatCreateResp{}
	if err := json.Unmarshal(bytes, result); err != nil {
		return nil, err
	}

	return &PreheatingStatus{
		TaskID: result.ID,
		Status: provider.PreheatingStatusPending, // default
	}, nil
```

Kraken: error & non-error & PreheatingStatusSuccess

```go
_, err = client.GetHTTPClient(kd.instance.Insecure).Post(url, kd.getCred(), payload, nil)
	if err != nil {
		return nil, err
	}

	return &PreheatingStatus{
		TaskID:     eventID,
		Status:     provider.PreheatingStatusSuccess,
		FinishTime: time.Now().String(),
	}, nil
```

PR #12341, opened 2020-06-29
Enhance: Support multi downversion in migration

ninjadq:
1. Change down version to list to accept multi verstion value
2. Update search function use BFS to find migration path
2. Add test case

### DIFF
--- a/make/photon/prepare/migrations/version_2_0_0/__init__.py
+++ b/make/photon/prepare/migrations/version_2_0_0/__init__.py
@@ -3,7 +3,7 @@ from jinja2 import Environment, FileSystemLoader, StrictUndefined
 from utils.migration import read_conf
 
 revision = '2.0.0'
-down_revision = '1.10.0'
+down_revisions = ['1.10.0']
 
 def migrate(input_cfg, output_cfg):
     config_dict = read_conf(input_cfg)
--- a/make/photon/prepare/utils/migration.py
+++ b/make/photon/prepare/utils/migration.py
@@ -4,7 +4,24 @@ import importlib
 import os
 from collections import deque
 
-from migrations import MIGRATION_BASE_DIR
+class MigratioNotFound(Exception): ...
+
+class MigrationVersion:
+    '''
+    The version used to migration
+
+    Arttribute:
+        name(str): version name like `1.0.0`
+        module: the python module object for a specific migration which contains migrate info, codes and templates
+        down_versions(list): previous versions that can migrated to this version
+    '''
+    def __init__(self, version: str):
+        self.name = version
+        self.module = importlib.import_module("migrations.version_{}".format(version.replace(".","_")))
+
+    @property
+    def down_versions(self):
+        return self.module.down_revisions
 
 def read_conf(path):
     with open(path) as f:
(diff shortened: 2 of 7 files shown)


make/photon/prepare/migrations/version_2_0_0/__init__.py
```diff
@@ -3,7 +3,7 @@
 from utils.migration import read_conf
 
 revision = '2.0.0'
-down_revision = '1.10.0'
+down_revision = ['1.10.0']
```
reasonerjt, line 6: Use plural `down_revisions`?

make/photon/prepare/utils/migration.py
```diff
@@ -15,29 +32,35 @@ def read_conf(path):
...
-            else:
-                upgrade_path.appendleft(module)
-                target_ver = module.down_revision
-                if target_ver in visited: # version visited before, loop found
-                    raise Exception('find a loop caused by {} on migration path'.format(target_ver))
-        else:
-            raise Exception('{} not dir'.format(os.path.join(MIGRATION_BASE_DIR, 'versions', target_ver.replace(".","_"))))
-    return upgrade_path
+    upgrade_path = deque()
+    parents, visited, q = dict(), set(), deque() # must ensure all version in parents are valid
```
reasonerjt, line 45: The var name parent is a map key is the old version and value is the new version right?

Why call it `parent`?

make/photon/prepare/utils/migration.py
```diff
@@ -15,29 +32,35 @@ def read_conf(path):
...
+    q.append(target_ver)
+    while q: # BFS to find a valid path
+        version = MigrationVersion(q.popleft())
+        if version.name in visited:
+            continue
+        visited.add(version.name)
+        if version.name == input_ver: 
+            upgrade_path.append(version)
```
reasonerjt, line 53: Why appending the input version in the path instead of using a simple flag like `found`?

make/photon/prepare/utils/migration.py
```diff
@@ -15,29 +32,35 @@ def read_conf(path):
...
+            break # break loop cause migration path found
+        for v in version.down_versions:
+            parents[v] = version.name
+            q.append(v)
+
+    if not upgrade_path:
+        raise MigratioNotFound('no migration path found to target version')
+
+    current_version = upgrade_path.popleft()
```
reasonerjt, line 62: I find it a little confusing you append the input_ver into upgrade_path and pop it and never put it back again.

If you use a flag like `found` in line 59 here you only need to assign `input_ver` to `current_version`, right?

make/photon/prepare/utils/migration.py
ninjadq: Because the upgrading path is a tree-style graph. When we find the version we expected we need to reverse the finding path. But there is no tree data structure here. So we need a parent array to record the parent.

make/photon/prepare/utils/migration.py
ninjadq: Because in this scenario the last element in `upgrade_path` can be used as a flag. Or you need an extra flag variable and initialize a MigrateVersion instance in another place.

PR #12356, opened 2020-06-30
add middlware for put manifest

wy65701436:
The middleware is to handle manifest(blob) status in different push manifest situation, similar with blob


src/server/middleware/blob/put_manifest.go
```diff
@@ -15,21 +15,60 @@
...
+			}
+			return err
+		}
+
+		switch bb.Status {
+		case blob_models.StatusNone, blob_models.StatusDelete, blob_models.StatusDeleteFailed:
+			err := blobController.Touch(r.Context(), bb)
+			if err != nil {
+				logger.Errorf("failed to update manifest: %s status to StatusNone, error:%v", bb.Digest, err)
+				return errors.Wrapf(err, fmt.Sprintf("the request id is: %s", r.Header.Get(requestid.HeaderXRequestID)))
```
heww, line 64: Do we need a request-id in the response? The request-id always will be in the header of the response and it will be in the log message when the message was logged by the logger from context.

wy65701436: @reasonerjt any comment?

reasonerjt: I would moderately agree,  B/C the response is for notifying the caller and the log is for further debugging.

src/server/middleware/blob/put_manifest.go
```diff
@@ -15,21 +15,60 @@
...
+
+		contentType := r.Header.Get("Content-Type")
+		_, descriptor, err := distribution.UnmarshalManifest(contentType, body)
+		if err != nil {
+			logger.Errorf("unmarshal manifest failed, error: %v", err)
+			return errors.Wrapf(err, "unmarshal manifest failed").WithCode(errors.MANIFESTINVALID)
+		}
+
+		// bb here is the actually a manifest, which is also stored as a blob in DB and storage.
+		bb, err := blobController.Get(r.Context(), descriptor.Digest.String())
```
reasonerjt, line 51: Does the code look the same in blob middleware?

Shall we reuse?

wy65701436: updated.

src/server/middleware/blob/util.go
```diff
@@ -0,0 +1,40 @@
+package blob
+
+import (
+	"fmt"
+	"github.com/goharbor/harbor/src/lib/errors"
+	"github.com/goharbor/harbor/src/lib/log"
+	blob_models "github.com/goharbor/harbor/src/pkg/blob/models"
```
heww, line 7: pkg rename not needed here

src/server/middleware/blob/util.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/goharbor/harbor/src/server/middleware/requestid"
+	"net/http"
+)
+
+// probeBlob handles config/layer and manifest status in the PUT Blob & Manifest middleware, and update the status before it passed into proxy(distribution).
+func probeBlob(r *http.Request, digest string) error {
+	ctx := r.Context()
+	logger := log.G(ctx)
```
heww, line 15: Delete `ctx := r.Context()` and use `log.G(r.Context())` otherwise use ctx direlcty in the following statements.

src/server/middleware/blob/util.go
wy65701436: updated.

src/server/middleware/blob/util.go
wy65701436: updated.

PR #12383, opened 2020-07-02
Suport filtering registries by type in listing registry API

ywk253100:
Suport filtering registries by type in listing registry API

### DIFF
--- a/src/core/config/config.go
+++ b/src/core/config/config.go
@@ -480,3 +480,12 @@ func QuotaSetting() (*models.QuotaSetting, error) {
 		StoragePerProject: cfgMgr.Get(common.StoragePerProject).GetInt64(),
 	}, nil
 }
+
+// GetPermittedRegistryTypesForProxyCache returns the permitted registry types for proxy cache
+func GetPermittedRegistryTypesForProxyCache() []string {
+	types := os.Getenv("PERMITTED_REGISTRY_TYPES_FOR_PROXY_CACHE")
+	if len(types) == 0 {
+		return []string{}
+	}
+	return strings.Split(types, ",")
+}
--- a/src/go.mod
+++ b/src/go.mod
@@ -57,6 +57,7 @@ require (
 	github.com/opencontainers/go-digest v1.0.0-rc1
 	github.com/opencontainers/image-spec v1.0.1
 	github.com/opentracing/opentracing-go v1.1.0 // indirect
+	github.com/pkg/errors v0.9.1
 	github.com/pquerna/cachecontrol v0.0.0-20180517163645-1555304b9b35 // indirect
 	github.com/robfig/cron v1.0.0
 	github.com/shiena/ansicolor v0.0.0-20151119151921-a422bbe96644 // indirect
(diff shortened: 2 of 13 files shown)


src/go.mod
```diff
@@ -57,6 +57,7 @@ require (
 	github.com/opencontainers/go-digest v1.0.0-rc1
 	github.com/opencontainers/image-spec v1.0.1
 	github.com/opentracing/opentracing-go v1.1.0 // indirect
+	github.com/pkg/errors v0.9.1
```
stonezdj, line 60: Wrong ?

src/core/config/config.go
```diff
@@ -480,3 +480,9 @@ func QuotaSetting() (*models.QuotaSetting, error) {
 		StoragePerProject: cfgMgr.Get(common.StoragePerProject).GetInt64(),
 	}, nil
 }
+
+// GetPermittedRegistryTypesForProxyCache returns the permitted registry types for proxy cache
+func GetPermittedRegistryTypesForProxyCache() []string {
+	types := os.Getenv("PERMITTED_REGISTRY_TYPES_FOR_PROXY_CACHE")
```
stonezdj, line 486: Setting in configure is better than env variable, because in tile env, if user need to modify the setting, change in the env cannot persistent, only configure in database can persistent in DB and can be migrated during upgrade.

reasonerjt: If we want to allow users to customize it, it will be exposed in harbor.yml, such that the env can be persisted

src/core/config/config.go
```diff
@@ -480,3 +480,9 @@ func QuotaSetting() (*models.QuotaSetting, error) {
...
+	return strings.Split(types, ",")
```
reasonerjt, line 487: If the env is not set this will return `[""]`, would it be better if it returns empty list?

src/replication/dao/registry.go
```diff
@@ -55,34 +49,37 @@ func GetRegistryByURL(url string) (*models.Registry, error) {
...
+	var countQuery *q.Query
+	if query != nil {
+		// ignore the page number and size
+		countQuery = &q.Query{
+			Keywords: query.Keywords,
+		}
 	}
-
-	total, err := q.Count()
+	countQs, err := liborm.QuerySetter(ctx, &models.Registry{}, countQuery)
```
reasonerjt, line 61: I may lack some background of the query but what's the different between countQs and qs? why not use the length of `registries` as the first return val?

src/core/config/config.go
ywk253100: The system configuration isn't persisted in the database

src/replication/dao/registry.go
ywk253100: The `countQs` contains no pagination information. 
And the `registries` may contain only the subset of all registry records if the pagination is set in query

src/go.mod
ywk253100: It's added automatically when run go build/test/etc.
It is used by https://github.com/goharbor/harbor/blob/master/src/pkg/p2p/preheat/provider/preheat_image.go#L21

src/core/config/config.go
ywk253100: Updated

PR #12385, opened 2020-07-02
support list blobs by update time

wy65701436:
Add support list blob with update time.
As introduces the time window in GC, it wants to list the blobs less than specific time.


src/pkg/blob/dao/dao.go
```diff
@@ -239,6 +239,10 @@ func (d *dao) ListBlobs(ctx context.Context, params models.ListParams) ([]*model
 		qs = qs.Filter("digest__in", params.BlobDigests)
 	}
 
+	if !params.UpdateTime.IsZero() {
```
reasonerjt, line 242: Is this necessary if it's zero, the lte is still true?

PR #12395, opened 2020-07-04
Feat/preheat controller policy part

chlins:
changes:

- preheat controller adds policy controller
- adds preheat policy related api's swagger documents
- implements preheat api policy handler

### DIFF
--- a/api/v2.0/swagger.yaml
+++ b/api/v2.0/swagger.yaml
@@ -814,6 +814,151 @@ paths:
           $ref: '#/responses/404'
         '500':
           $ref: '#/responses/500'
+  /projects/{project_name}/preheat/policies:
+    post:
+      summary: Create a preheat policy under a project
+      description: Create a preheat policy under a project
+      tags:
+        - preheat
+      operationId: CreatePolicy
+      parameters:
+        - $ref: '#/parameters/requestId'
+        - $ref: '#/parameters/projectName'
+        - name: policy
+          in: body
+          description: The policy schema info
+          required: true
+          schema:
+            $ref: '#/definitions/PreheatPolicy'
+      responses:
+        '201':
+          description: Create policy success
+          schema:
+            $ref: '#/responses/201'
+        '400':
+          $ref: '#/responses/400'
+        '401':
+          $ref: '#/responses/401'
+        '403':
+          $ref: '#/responses/403'
+        '409':
+          $ref: '#/responses/409'
+        '500':
+          $ref: '#/responses/500'
+    get:
+      summary: List preheat policies
+      description: List preheat policies
+      tags:
+        - preheat
+      operationId: ListPolicies
+      parameters:
+        - $ref: '#/parameters/requestId'
+        - $ref: '#/parameters/projectName'
+        - $ref: '#/parameters/page'
+        - $ref: '#/parameters/pageSize'
+        - $ref: '#/parameters/query'
+      responses:
+        '200':
+          description: List preheat policies success
+          headers:
+            X-Total-Count:
+              description: The total count of tags
+              type: integer
+            Link:
+              description: Link refers to the previous page and next page
+              type: string
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/PreheatPolicy'
+        '400':
+          $ref: '#/responses/400'
+        '401':
+          $ref: '#/responses/401'
+        '403':
+          $ref: '#/responses/403'
+        '500':
+          $ref: '#/responses/500'
+  /projects/{project_name}/preheat/policies/{preheat_policy_name}:
+    get:
+      summary: Get a preheat policy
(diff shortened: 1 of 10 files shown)


src/server/v2.0/handler/preheat.go
```diff
@@ -80,3 +86,128 @@ func convertProvidersToFrontend(backend []*provider.Metadata) (frontend []*model
...
+// ListPolicies is List preheat policies
+func (api *preheatAPI) ListPolicies(ctx context.Context, params operation.ListPoliciesParams) middleware.Responder {
+	query, err := api.BuildQuery(ctx, params.Q, params.Page, params.PageSize)
+	if err != nil {
+		return api.SendError(ctx, err)
+	}
+
+	if query != nil {
+		query.Keywords["project_id"] = params.ProjectID
+	}
```
kofj, line 152: query.Keywords["ProjectID"] = pro.ProjectID

steven-zou: We need to double-check the property key in the keywords should be the column name in the DB schema, right?

src/server/v2.0/handler/preheat.go
```diff
@@ -80,3 +86,128 @@ func convertProvidersToFrontend(backend []*provider.Metadata) (frontend []*model
 	}
 	return
 }
+
+// GetPolicy is Get a preheat policy
+func (api *preheatAPI) GetPolicy(ctx context.Context, params operation.GetPolicyParams) middleware.Responder {
```
steven-zou, line 91: Don't forget to add RBAC controlling later.

api/v2.0/swagger.yaml
```diff
@@ -814,6 +814,148 @@ paths:
...
+              $ref: '#/definitions/PreheatPolicy'
+        '400':
+          $ref: '#/responses/400'
+        '401':
+          $ref: '#/responses/401'
+        '403':
+          $ref: '#/responses/403'
+        '500':
+          $ref: '#/responses/500'
+  /projects/{project_id}/preheat/policies/{preheat_policy_id}:
```
steven-zou, line 879: If we keep the name unique, I think we should also use policy name here to keep consistent with other APIs.

src/server/v2.0/handler/preheat.go
kofj: I'll add RBAC after this merged.

src/server/v2.0/handler/preheat.go
```diff
@@ -80,3 +86,128 @@ func convertProvidersToFrontend(backend []*provider.Metadata) (frontend []*model
...
+
+	var payload []*models.PreheatPolicy
+	for _, policy := range policies {
+		p, err := convertPolicyToPayload(policy)
+		if err != nil {
+			return api.SendError(ctx, err)
+		}
+		payload = append(payload, p)
+	}
+	return operation.NewListPoliciesOK().WithPayload(payload).WithXTotalCount(total)
```
kofj, line 172: Pls add `WithLink(a.Links(ctx, params.HTTPRequest.URL, total, query.PageNumber, query.PageSize).String()).`.

src/server/v2.0/handler/preheat.go
chlins: the column name in db schema is 'project_id'

src/pkg/p2p/preheat/dao/policy/dao.go
```diff
@@ -122,6 +124,25 @@ func (d *dao) Get(ctx context.Context, id int64) (schema *policy.Schema, err err
 	return schema, nil
 }
 
+// GetByName gets a policy schema by name.
+func (d *dao) GetByName(ctx context.Context, projectID int64, name string) (schema *policy.Schema, err error) {
```
steven-zou, line 128: Actually it can be supported by the List(ctx,&q.Query{keywords:map[string]interfae{}{
  "project_id": 1,
  "name: "policy_name",
}}

src/server/v2.0/handler/preheat.go
```diff
@@ -80,3 +89,149 @@ func convertProvidersToFrontend(backend []*provider.Metadata) (frontend []*model
...
+
+	payload, err = convertPolicyToPayload(policy)
+	if err != nil {
+		return api.SendError(ctx, err)
+	}
+	return operation.NewGetPolicyOK().WithPayload(payload)
+}
+
+// CreatePolicy is Create a preheat policy under a project
+func (api *preheatAPI) CreatePolicy(ctx context.Context, params operation.CreatePolicyParams) middleware.Responder {
```
steven-zou, line 114: Two things we need to make sure here as well as the `UpdatePolicy` method:
  
- name uniqueness checking 
- the name should be limited to the specified format which should be aligned with the URL pattern ([\d_]+)?

PR #12400, opened 2020-07-06
add time windows support in artifact trash

wy65701436:
support with time window to filter and delete item in artifact trash

### DIFF
--- a/src/pkg/artifactrash/dao/dao.go
+++ b/src/pkg/artifactrash/dao/dao.go
@@ -15,10 +15,10 @@ type DAO interface {
 	Create(ctx context.Context, artifactrsh *model.ArtifactTrash) (id int64, err error)
 	// Delete the artifact trash specified by ID
 	Delete(ctx context.Context, id int64) (err error)
-	// Filter lists the artifact that needs to be cleaned
-	Filter(ctx context.Context) (arts []model.ArtifactTrash, err error)
-	// Flush clean the trash table
-	Flush(ctx context.Context) (err error)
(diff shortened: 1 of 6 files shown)


src/pkg/artifactrash/dao/dao.go
```diff
@@ -16,9 +16,9 @@ type DAO interface {
 	// Delete the artifact trash specified by ID
 	Delete(ctx context.Context, id int64) (err error)
 	// Filter lists the artifact that needs to be cleaned
-	Filter(ctx context.Context) (arts []model.ArtifactTrash, err error)
+	Filter(ctx context.Context, timeWindow time.Time) (arts []model.ArtifactTrash, err error)
```
reasonerjt, line 19: I think the var name is misleading, timewindow means a range of time.

src/pkg/artifactrash/manager.go
```diff
@@ -32,9 +33,9 @@ type Manager interface {
 	// Delete ...
 	Delete(ctx context.Context, id int64) (err error)
 	// Filter ...
-	Filter(ctx context.Context) (arts []model.ArtifactTrash, err error)
+	Filter(ctx context.Context, timeWindow int64) (arts []model.ArtifactTrash, err error)
```
reasonerjt, line 36: We should clarify the window here is hour

src/pkg/artifactrash/dao/dao.go
wy65701436: updated.

src/pkg/artifactrash/manager.go
wy65701436: updated.

PR #12410, opened 2020-07-07
Fix  project metadata for webhook

tedgxt:
Fix webhook bug for getting proejct metadata failed.

### DIFF
--- a/src/controller/event/handler/replication/replication.go
+++ b/src/controller/event/handler/replication/replication.go
@@ -15,13 +15,15 @@
 package replication
 
 import (
+	"strconv"
+
 	"github.com/goharbor/harbor/src/controller/event"
+	"github.com/goharbor/harbor/src/controller/project"
 	"github.com/goharbor/harbor/src/lib/log"
-	"github.com/goharbor/harbor/src/pkg/project"
+	"github.com/goharbor/harbor/src/lib/orm"
 	"github.com/goharbor/harbor/src/replication"
 	repevent "github.com/goharbor/harbor/src/replication/event"
 	"github.com/goharbor/harbor/src/replication/model"
-	"strconv"
 )
 
 // Handler ...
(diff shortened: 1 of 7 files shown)


src/controller/event/handler/replication/replication.go
```diff
@@ -57,13 +60,16 @@ func (r *Handler) IsStateful() bool {
...
-	} else {
-		log.Error(err)
+	project, err := project2.Ctl.Get(context.TODO(), art.ProjectID, project2.Metadata(true))
+	if err != nil {
+		log.Errorf("failed to get project: %d, error: %v", art.ProjectID, err)
+		return err
+	}
+	if project == nil {
+		return fmt.Errorf("project %d not found of replication event", art.ProjectID)
 	}
```
heww, line 70: This if statement is not needed because the `Get` method of the project controller will return `not found` error when the project not exists.

```go
func (c *controller) Get(ctx context.Context, projectID int64, options ...Option) (*models.Project, error) {
	p, err := c.projectMgr.Get(projectID)
	if err != nil {
		return nil, err
	}
	if p == nil {
		return nil, errors.NotFoundError(nil).WithMessage("project %d not found", projectID)
	}

	return c.assembleProject(ctx, p, newOptions(options...))
}
```

src/controller/event/handler/replication/replication.go
```diff
@@ -15,13 +15,16 @@
 package replication
 
 import (
+	"context"
+	"fmt"
+	"strconv"
+
 	"github.com/goharbor/harbor/src/controller/event"
+	project2 "github.com/goharbor/harbor/src/controller/project"
```
heww, line 23: Alias the pkg name to `project2` is not very good please keep it as `project', and we can make the var name `project` to `proj` or `pro' or `p` when needed.

src/controller/event/handler/replication/replication.go
```diff
@@ -57,13 +60,16 @@ func (r *Handler) IsStateful() bool {
 func (r *Handler) handlePushArtifact(event *event.PushArtifactEvent) error {
 	art := event.Artifact
 	public := false
-	project, err := project.Mgr.Get(art.ProjectID)
-	if err == nil && project != nil {
-		public = project.IsPublic()
-	} else {
-		log.Error(err)
+	project, err := project2.Ctl.Get(context.TODO(), art.ProjectID, project2.Metadata(true))
```
heww, line 63: `context.TODO()` does not work here because it `orm` instance is missing in `context.TODO()` please use `orm.Context()` in the src/lib pkg.

src/controller/event/handler/replication/replication.go
tedgxt: done

src/controller/event/handler/replication/replication.go
tedgxt: done

src/controller/event/handler/webhook/artifact/artifact.go
```diff
@@ -57,12 +57,13 @@ func (a *Handler) IsStateful() bool {
 
 func (a *Handler) handle(event *event.ArtifactEvent) error {
 	var err error
```
wy65701436, line 59: do we need to define err here?

tedgxt: > do we need to define err here?

err definition deleted

PR #12430, opened 2020-07-09
Preheat and healthcheck

kofj:
Add preheat APIs, handlers.

1. Manual preheat.
2. Instance health check.

### DIFF
--- a/api/v2.0/swagger.yaml
+++ b/api/v2.0/swagger.yaml
@@ -667,6 +667,33 @@ paths:
           $ref: '#/responses/404'
         '500':
           $ref: '#/responses/500'
+  /p2p/preheat/instances/ping:
+    post:
+      summary: Ping status of a instance.
+      description: |
+        This endpoint checks status of a instance, the instance can be given by ID or Endpoint URL (together with credential)
+      operationId: PingInstances
+      parameters:
+        - $ref: '#/parameters/requestId'
+        - name: instance
+          in: body
+          description: The JSON object of instance.
+          required: true
+          schema:
+            $ref: '#/definitions/Instance'
+      tags:
+        - preheat
+      responses:
+        '200':
+          $ref: '#/responses/200'
+        '400':
+          $ref: '#/responses/400'
+        '401':
+          $ref: '#/responses/401'
+        '404':
+          description: Instance not found (when instance is provided by ID).
+        '500':
+          $ref: '#/responses/500'
   /p2p/preheat/instances:
     get:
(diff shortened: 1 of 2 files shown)


api/v2.0/swagger.yaml
```diff
@@ -667,6 +667,33 @@ paths:
...
+          description: The JSON object of instance.
+          required: true
+          schema:
+            $ref: '#/definitions/Instance'
+      tags:
+        - preheat
+      responses:
+        '200':
+          description: Instance is healthy.
+        '400':
```
steven-zou, line 689: Why not use the ref `$ref: '#/responses/401'`?

src/server/v2.0/handler/preheat.go
```diff
@@ -248,6 +251,59 @@ func (api *preheatAPI) ListPolicies(ctx context.Context, params operation.ListPo
...
+	if err != nil {
+		return api.SendError(ctx, err)
+	}
+
+	policy, err := api.preheatCtl.GetPolicyByName(ctx, project.ProjectID, params.PreheatPolicyName)
+	if err != nil {
+		return api.SendError(ctx, err)
+	}
+
+	_, err = api.enforcer.EnforcePolicy(ctx, policy.ID)
```
steven-zou, line 266: I suggest accepting the returned execution id and return it to the front end with a proper format:

  - Option1: JSON data
  - Option2: add a location header in the response

src/server/v2.0/handler/preheat.go
```diff
@@ -22,6 +23,7 @@ func newPreheatAPI() *preheatAPI {
 	return &preheatAPI{
 		preheatCtl: preheatCtl.Ctl,
 		projectCtl: projectCtl.Ctl,
+		enforcer:   preheatCtl.NewEnforcer(),
```
chlins, line 26: How about add a global enforcer like other controllers？

api/v2.0/swagger.yaml
kofj: I'm missed it, fix later.

src/server/v2.0/handler/preheat.go
kofj: `mmpei` add a global enforcer at  #12428, I'll rebase and use it.

src/server/v2.0/handler/preheat.go
kofj: Great, I'll return JOSN data include ID.

PR #12437, opened 2020-07-09
chore(db): change max_connections of postgres to 1024

heww:
Change max_connections of Postgres to 1024 and the max_open_conns for the core to 1000.


make/photon/db/docker-entrypoint.sh
```diff
@@ -107,4 +107,14 @@ EOSQL
         echo
 fi
 
-postgres -D $PGDATA
+POSTGRES_PARAMETER=''
+
+file_env 'POSTGRES_MAX_CONNECTIONS' '100'
```
bitsf, line 112: 这个 file_env 是什么命令

make/photon/prepare/utils/db.py
```diff
@@ -15,4 +15,5 @@ def prepare_db(config_dict):
     render_jinja(
         db_env_template_path,
         db_conf_env,
-        harbor_db_password=config_dict['harbor_db_password'])
+        harbor_db_password=config_dict['harbor_db_password'],
+        harbor_db_max_connections=config_dict['harbor_db_max_open_conns'])
```
reasonerjt, line 19: Can we refrain from allowing user to set this in `harbor.yml`?

If it's set to a very large number I believe the `shared_buffer` must change accordingly.

I suggest for v2.1 and v2.0.2 we just hard code it in env of `harbor-db` with a relatively larger number, say 1000

make/photon/db/docker-entrypoint.sh
heww: file_env is a function defined in `docker-entrypoint.sh` which supports reading the value from env or file and also supports default value when the env or file not found.

make/photon/prepare/utils/db.py
heww: Changed to max_connections in Postgres to 1024 and max_open_conns to 1000 in the core.

PR #12439, opened 2020-07-09
revise gc job to align non blocking gc

wy65701436:
two phases:
1, mark, select the gc candidates bases on the DB and mark them as status delete.
2, sweep, select the candidate and mark it as status deleting and remove it from backend and database.

### DIFF
--- a/src/jobservice/job/impl/gc/garbage_collection.go
+++ b/src/jobservice/job/impl/gc/garbage_collection.go
@@ -15,8 +15,11 @@
 package gc
 
 import (
-	"fmt"
+	"github.com/goharbor/harbor/src/lib/errors"
+	"github.com/goharbor/harbor/src/pkg/artifactrash/model"
+	blob_models "github.com/goharbor/harbor/src/pkg/blob/models"
 	"os"
+	"strconv"
 	"time"
 
 	"github.com/goharbor/harbor/src/common/models"
@@ -27,8 +30,6 @@ import (
 	"github.com/goharbor/harbor/src/pkg/blob"
 
 	"github.com/garyburd/redigo/redis"
-	"github.com/goharbor/harbor/src/common"
-	"github.com/goharbor/harbor/src/common/config"
 	"github.com/goharbor/harbor/src/common/registryctl"
 	"github.com/goharbor/harbor/src/jobservice/job"
 	"github.com/goharbor/harbor/src/jobservice/logger"
@@ -37,21 +38,6 @@ import (
 
 var (
 	regCtlInit = registryctl.Init
-
-	getReadOnly = func(cfgMgr *config.CfgManager) (bool, error) {
-		if err := cfgMgr.Load(); err != nil {
-			return false, err
-		}
-		return cfgMgr.Get(common.ReadOnly).GetBool(), nil
-	}
-
-	setReadOnly = func(cfgMgr *config.CfgManager, switcher bool) error {
-		cfg := map[string]interface{}{
-			common.ReadOnly: switcher,
-		}
-		cfgMgr.UpdateConfig(cfg)
-		return cfgMgr.Save()
-	}
 )
 
 const (
@@ -71,10 +57,16 @@ type GarbageCollector struct {
 	projectCtl        project.Controller
 	registryCtlClient client.Client
 	logger            logger.Interface
-	cfgMgr            *config.CfgManager
-	CoreURL           string
 	redisURL          string
 	deleteUntagged    bool
+	dryRun            bool
+	// holds all of trashed artifacts' digest and repositories.
+	// The source data of trashedArts is the table ArtifactTrash and it's only used as a dictionary by sweep when to delete a manifest.
+	// As table blob has no repositories data, and the repositories are required when to delete a manifest, so use the table ArtifactTrash to capture them.
+	trashedArts map[string][]string
+	// hold all of GC candidates(non-referenced blobs), it's captured by mark and consumed by sweep.
+	deleteSet       []*blob_models.Blob
+	timeWindowHours int64
 }
 
 // MaxFails implements the interface in job/Interface
@@ -97,58 +89,11 @@ func (gc *GarbageCollector) Validate(params job.Parameters) error {
 	return nil
 }
 
-// Run implements the interface in job/Interface
-// The workflow of GC is:
-// 1, set harbor to readonly
-// 2, select the candidate artifacts from Harbor DB.
-// 3, call registry API(--delete-untagged=false) to delete manifest bases on the results of #2
-// 4, clean keys of redis DB of registry, clean artifact trash and untagged from DB.
-// 5, roll back readonly.
-// More details:
-// 1, why disable delete untagged when to call registry API
-//		Generally because that we introduce Harbor tag in v2.0, it's in database but no corresponding data in registry.
-//		Also one failure case example:
-// 			there are two parts for putting an manifest in Harbor: write database and write storage, but they're not in a transaction,
-//			which leads to the data mismatching in parallel pushing images with same tag but different digest. The valid artifact in
-//			harbor DB could be a untagged one in the storage. If we enable the delete untagged, the valid data could be removed from the storage.
-// 2, what to be cleaned
-//		> the deleted artifact, bases on table of artifact_trash and artifact
-//		> the untagged artifact(optional), bases on table of artifact.
-func (gc *GarbageCollector) Run(ctx job.Context, params job.Parameters) error {
-	if err := gc.init(ctx, params); err != nil {
-		return err
-	}
-	readOnlyCur, err := getReadOnly(gc.cfgMgr)
-	if err != nil {
-		return err
-	}
-	if readOnlyCur != true {
-		if err := setReadOnly(gc.cfgMgr, true); err != nil {
-			return err
-		}
-		defer setReadOnly(gc.cfgMgr, readOnlyCur)
-	}
-	gc.logger.Infof("start to run gc in job.")
-	if err := gc.deleteCandidates(ctx); err != nil {
-		gc.logger.Errorf("failed to delete GC candidates in gc job, with error: %v", err)
-	}
-	gcr, err := gc.registryCtlClient.StartGC()
-	if err != nil {
-		gc.logger.Errorf("failed to get gc result: %v", err)
-		return err
-	}
-	gc.removeUntaggedBlobs(ctx)
-	if err := gc.cleanCache(); err != nil {
-		return err
-	}
-	gc.logger.Infof("GC results: status: %t, message: %s, start: %s, end: %s.", gcr.Status, gcr.Msg, gcr.StartTime, gcr.EndTime)
-	gc.logger.Infof("success to run gc in job.")
-	return nil
-}
-
 func (gc *GarbageCollector) init(ctx job.Context, params job.Parameters) error {
 	regCtlInit()
 	gc.logger = ctx.GetLogger()
+	gc.deleteSet = make([]*blob_models.Blob, 0)
+	gc.trashedArts = make(map[string][]string, 0)
 	opCmd, flag := ctx.OPCommand()
 	if flag && opCmd.IsStop() {
 		gc.logger.Info("received the stop signal, quit GC job.")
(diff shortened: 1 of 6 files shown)


src/pkg/blob/models/blob.go
```diff
@@ -83,6 +85,11 @@ func (b *Blob) IsForeignLayer() bool {
 	return b.ContentType == schema2.MediaTypeForeignLayer
 }
 
+// IsManifest returns true if the blob is manifest layer
+func (b *Blob) IsManifest() bool {
```
reasonerjt, line 89: Also need to take into account the manifest of oci images?

src/pkg/artifactrash/dao/dao.go
```diff
@@ -71,9 +72,9 @@ func (d *dao) Filter(ctx context.Context, cutOff time.Time) (arts []model.Artifa
 		return deletedAfs, err
 	}
 
-	sql := `SELECT aft.* FROM artifact_trash AS aft LEFT JOIN artifact af ON (aft.repository_name=af.repository_name AND aft.digest=af.digest) WHERE (af.digest IS NULL AND af.repository_name IS NULL) AND aft.creation_time <= ?`
+	sql := fmt.Sprintf(`SELECT aft.* FROM artifact_trash AS aft LEFT JOIN artifact af ON (aft.repository_name=af.repository_name AND aft.digest=af.digest) WHERE (af.digest IS NULL AND af.repository_name IS NULL) AND aft.creation_time <= TO_TIMESTAMP('%f')`, float64(cutOff.UnixNano())/float64((time.Second)))
 
-	_, err = ormer.Raw(sql, cutOff).QueryRows(&deletedAfs)
+	_, err = ormer.Raw(sql).QueryRows(&deletedAfs)
```
reasonerjt, line 77: Is there a better way other than `Sprintf` the SQL?

Doesn't `orm` support comparing time?

src/jobservice/job/impl/gc/util.go
```diff
@@ -54,3 +55,13 @@ func deleteManifest(repository, digest string) error {
 	}
 	return nil
 }
+
+// ignoreNotFound ignores the NotFoundErr error
+func ignoreNotFound(f func() error) error {
+	if err := f(); err != nil {
+		if !errors.IsNotFoundErr(err) {
```
reasonerjt, line 62: I suppose you can use `&&` here?

src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -15,7 +15,9 @@
 package gc
 
 import (
-	"fmt"
+	"github.com/goharbor/harbor/src/lib/errors"
```
reasonerjt, line 18: I thought we discussed keeping the old gc job as backup plan?

src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -186,6 +114,162 @@ func (gc *GarbageCollector) init(ctx job.Context, params job.Parameters) error {
 			gc.deleteUntagged = untagged
 		}
 	}
+	// default is not execute dry run, and for dry run we can have button in the UI.
+	dryRun, exist := params["dry_run"]
+	if exist {
+		if dryRun, ok := dryRun.(bool); ok && !dryRun {
```
reasonerjt, line 120: `!dryRun` ???

src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -186,6 +114,162 @@ func (gc *GarbageCollector) init(ctx job.Context, params job.Parameters) error {
...
+
+	// update delete status for the candidates.
+	blobCt := 0
+	mfCt := 0
+	for _, blob := range blobs {
+		if !gc.dryRun {
+			blob.Status = blob_models.StatusDelete
+			_, err := gc.blobMgr.UpdateBlobStatus(ctx.SystemContext(), blob)
+			if err != nil {
+				gc.logger.Errorf("failed to mark gc candidate, skip it.: %s", blob.Digest)
```
reasonerjt, line 193: Also log the error?

src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -186,6 +114,162 @@ func (gc *GarbageCollector) init(ctx job.Context, params job.Parameters) error {
...
+				continue
+			}
+		}
+		gc.logger.Infof("blob eligible for deletion: %s", blob.Digest)
+		gc.deleteSet = append(gc.deleteSet, blob)
+		// as table blob has no repository name, here needs to use the ArtifactTrash to fill it in.
+		if blob.IsManifest() {
+			for _, at := range arts {
```
reasonerjt, line 201: if arts is a map there won't be nested loop

src/pkg/blob/models/blob.go
```diff
@@ -71,6 +72,7 @@ type Blob struct {
 	UpdateTime   time.Time `orm:"column(update_time);auto_now_add" json:"update_time"`
 	Version      int64     `orm:"column(version)" json:"version"`
 	CreationTime time.Time `orm:"column(creation_time);auto_now_add" json:"creation_time"`
+	Repositories []string  `orm:"-" json:"repositories"`
```
reasonerjt, line 75: Can we do it different in gc job?  

I don't think keeping this reference makes sense.

src/pkg/artifactrash/dao/dao.go
wy65701436: orm cannot handle comparing time in milliseconds level.

src/jobservice/job/impl/gc/garbage_collection.go
heww: `Warn` will be better than ``Error`

src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -186,6 +114,162 @@ func (gc *GarbageCollector) init(ctx job.Context, params job.Parameters) error {
...
+		// set the status firstly, if the blob is updated by any HEAD/PUT request, it should be fail and skip.
+		blob.Status = blob_models.StatusDeleting
+		_, err := gc.blobMgr.UpdateBlobStatus(ctx.SystemContext(), blob)
+		if err != nil {
+			gc.logger.Errorf("failed to mark gc candidate deleting, skip: %s, %s", blob.Digest, blob.Status)
+			continue
+		}
+
+		if blob.IsManifest() {
+			for _, repo := range blob.Repositories {
```
heww, line 227: Can we do DB query here to get the repositories from `ArtifactTrash` here?

If so, no `Repostiroies` field need to be added to `Blob`. And the results to filter ArtifactTrash by manifest digest is smaller than all ArtifactTrashs from the table.

src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -186,6 +114,162 @@ func (gc *GarbageCollector) init(ctx job.Context, params job.Parameters) error {
...
+	if err != nil {
+		gc.logger.Errorf("failed to delete GC candidates in gc job, with error: %v", err)
+		return err
+	}
+	// no need to execute GC as there is no removed artifacts.
+	// Do this is to handle if user trigger GC job several times, only one job should do the following logic as artifact trash table is flushed.
+	if len(arts) == 0 {
+		gc.logger.Info("no need to execute GC as there is no removed artifacts.")
+		return nil
+	}
```
heww, line 168: Upgrade from v1.9 and v1.10, the ArtifactTrash table is empty but some blobs may need to delete in GC.

src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -186,6 +114,162 @@ func (gc *GarbageCollector) init(ctx job.Context, params job.Parameters) error {
...
+					if err := gc.markDeleteFailed(ctx, blob); err != nil {
+						return err
+					}
+					return errors.Wrapf(err, "failed to delete manifest with v2 API: %s, %s", repo, blob.Digest)
+				}
+				// for manifest, it has to delete the revisions folder of each repository
+				gc.logger.Infof("delete manifest from storage: %s", blob.Digest)
+				if err := ignoreNotFound(func() error {
+					return gc.registryCtlClient.DeleteManifest(repo, blob.Digest)
+				}); err != nil {
```
heww, line 242: I think the following is more clear than the wrap of the function.
```go
if err := gc.registryCtlClient.DeleteManifest(repo, blob.Digest); ignoreNotFound(err) != nil {
    // ....
}
```

src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -186,6 +114,162 @@ func (gc *GarbageCollector) init(ctx job.Context, params job.Parameters) error {
...
+					if err := gc.markDeleteFailed(ctx, blob); err != nil {
+						return err
+					}
+					return errors.Wrapf(err, "failed to remove manifest from storage: %s, %s", repo, blob.Digest)
```
heww, line 246: Does this mean that we will break the GC job when deleting the manifest failed?

src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -186,6 +114,162 @@ func (gc *GarbageCollector) init(ctx job.Context, params job.Parameters) error {
...
+		}
+
+		// remove the blob record
+		if err := ignoreNotFound(func() error {
+			return gc.blobMgr.Delete(ctx.SystemContext(), blob.ID)
+		}); err != nil {
+			if err := gc.markDeleteFailed(ctx, blob); err != nil {
+				return err
+			}
+			return errors.Wrapf(err, "failed to delete blob from database: %s, %s", blob.Digest, blob.Status)
```
heww, line 270: Does this mean that we will break the GC job when deleting the blob in DB was failed?

src/jobservice/job/impl/gc/garbage_collection.go
wy65701436: I have considered about it, but if the user triggers the GC for the first time without delete any of artifact, GC will delete some blobs, it seemed a little strange to him. 

For the undeleted blobs in the older version, once user gets some artifat deleted, then the GC job will remove them.

src/jobservice/job/impl/gc/garbage_collection.go
wy65701436: for the current design, it's yes.

src/jobservice/job/impl/gc/garbage_collection.go
wy65701436: yes, do you mean that we can skip it?

src/jobservice/job/impl/gc/garbage_collection.go
wy65701436: @reasonerjt your idea, both are ok for me.

src/pkg/blob/models/blob.go
wy65701436: added

wy65701436: added.

src/jobservice/job/impl/gc/util.go
wy65701436: updated.

src/jobservice/job/impl/gc/garbage_collection.go
wy65701436: add log error with warning level.

src/jobservice/job/impl/gc/garbage_collection.go
wy65701436: @reasonerjt your idea, both are ok for me.

src/jobservice/job/impl/gc/garbage_collection.go
wy65701436: it should be `dryRun`, thanks.

src/jobservice/job/impl/gc/garbage_collection.go
wy65701436: do you mean that to add query Repostiroies for each blob?

src/jobservice/job/impl/gc/garbage_collection.go
heww: Yes, I think we can skip this blob when delete it failed and to handle the next one.

src/jobservice/job/impl/gc/garbage_collection.go
heww: GC happens not because users had deleted the artifacts it happens because there are blobs that need to delete.

heww: Oh, I know that when then ArtifactTrash is empty, we can't get the repository of the blob, it's so sad.

src/jobservice/job/impl/gc/garbage_collection.go
wy65701436: updated

src/pkg/blob/models/blob.go
wy65701436: removed.

src/jobservice/job/impl/gc/garbage_collection.go
heww: Query all artifact trash and make them to a map is also ok.

src/jobservice/job/impl/gc/garbage_collection.go
wy65701436: I'll create another PR to handle this, maybe can use a parameter to trigger differernt job.

src/pkg/artifactrash/dao/dao.go
reasonerjt: use prepared statements?

src/jobservice/job/impl/gc/garbage_collection.go
reasonerjt: both are ok IMO

Issue #11606, opened 2020-04-14
How to prevent user from executing GC in container of registry/registryctl 


wy65701436:
As from Harbor v2.0, `tag` is maintained in Harbor instead of back end registry, which leads to the valid images in Harbor with tags but without tags in the registry.

Therefore, Harbor disables its GC job to delete the untagged manifest when to call registry GC CLI.

But, how can we prevent a user from executing GC CLI with enabling delete untagged manifest option in the container of registry and regsitryctl. 

If the user executes GC with the option, it should be a catastrophic event as GC will remove most of the valid images from the backend storage and no way to restore.

reasonerjt:
We can rename the binary file to mitigate the risk, 
and we MUST clearly add release note and document

PR #12471, opened 2020-07-13
Rename registry_DO_NOT_USE_GC to registry

stefannica:
Harbor container images built for other OS distributions prefer installing
docker-registry from packages instead of building directly from source.
The registryctl code should use the more generic `registry` name instead
of the hardcoded `registry_DO_NOT_USE_GC` name, to enable such cases.

Signed-off-by: Stefan Nica <snica@suse.com>

### DIFF
--- a/make/photon/registry/Dockerfile
+++ b/make/photon/registry/Dockerfile
@@ -4,12 +4,12 @@ FROM ${harbor_base_namespace}/harbor-registry-base:${harbor_base_image_version}
 
 COPY ./make/photon/common/install_cert.sh /home/harbor
 COPY ./make/photon/registry/entrypoint.sh /home/harbor
-COPY ./make/photon/registry/binary/registry /usr/bin/registry_DO_NOT_USE_GC
+COPY ./make/photon/registry/binary/registry /usr/bin/registry
 
 RUN chown -R harbor:harbor /etc/pki/tls/certs \
     && chown harbor:harbor /home/harbor/entrypoint.sh && chmod u+x /home/harbor/entrypoint.sh \
     && chown harbor:harbor /home/harbor/install_cert.sh && chmod u+x /home/harbor/install_cert.sh \
-    && chown harbor:harbor /usr/bin/registry_DO_NOT_USE_GC && chmod u+x /usr/bin/registry_DO_NOT_USE_GC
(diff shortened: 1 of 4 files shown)


wy65701436:
Thanks for the PR.

You can refer to the issue(https://github.com/goharbor/harbor/issues/11606) to get more details. The name set to `registry_DO_NOT_USE_GC` is want to warning user to do not execute GC manually, at least in Harbor integration.

More background here, as Harbor v2.1 will introduce non-blocking GC, that means Harbor will deside which blob/manifest will be removed instead of distribution. So, Harbor needs this warning message.

For the docker distribution build you mentioned in the comments, so far docker distribution v2.7.1 cannot satisfy Harbor's requirment, so we patched two PRs into v2.7.1, that's why we use source code to build distribution.

The PR cannot be accepted, you can file issue if any further query.

stefannica:
> Thanks for the PR.
> 
> You can refer to the issue(#11606) to get more details. The name set to `registry_DO_NOT_USE_GC` is want to warning user to do not execute GC manually, at least in Harbor integration.
> 

Thanks for pointing that out @wy65701436. Was manually running the registry `garbage-collect` command inside the `harbor-registry` or `harbor-registryctl` containers ever an officially supported (and documented) method to do garbage collection ? I understand the intention is to warn off users from executing that command manually now, but I would say that it is _generally_ not recommended to manually run commands inside containers, because this would interfere with the registry's overall functionality, wouldn't you agree ?

> More background here, as Harbor v2.1 will introduce non-blocking GC, that means Harbor will deside which blob/manifest will be removed instead of distribution. So, Harbor needs this warning message.
> 
> For the docker distribution build you mentioned in the comments, so far docker distribution v2.7.1 cannot satisfy Harbor's requirment, so we patched two PRs into v2.7.1, that's why we use source code to build distribution.

This change was meant to allow images to be built from a generic docker-distribution OS package, which I should mention can also be patched with backported bugfixes, so that shouldn't be a problem. However, changing the name of binaries inside generic RPM packages is not possible.

> 
> The PR cannot be accepted, you can file issue if any further query.

Thanks, I understand. This is not something important enough to file an issue over, I can find a workaround that doesn't involve Harbor code changes. I was also going to suggest making the name of the registry binary configurable for `registryctl`, but it seems too much effort for something this small.

PR #12478, opened 2020-07-14
feat(p2p):enhance policy enforcer

### LINKED ISSUE #12443: Enhance preheat policy enforcer to support reading security settings (vulnerability scanning and signature) from the project configurations

steven-zou:
Read security settings from the project configurations and override the preheat policy settings if necessary.

Check the project security settings and override the related settings in the policy if necessary.
NOTES: if the security settings (relevant with signature and vulnerability) are set at the project configuration,
they will have the highest priority and override the related settings of the preheat policy.
 e.g (use signature as an example, similar case to vulnerability severity part):
   if policy.signature = false and project.config.signature = true; then policy.signature = true
   if policy.signature = true and project.config.signature = true; then policy.signature = true
   if policy.signature = true and project.config.signature = false; then policy.signature = true
   if policy.signature = false and project.config.signature = false; then policy.signature = false

fix #12443 

### DIFF
--- a/src/controller/p2p/preheat/enforcer.go
+++ b/src/controller/p2p/preheat/enforcer.go
@@ -28,8 +28,8 @@ import (
 	"github.com/goharbor/harbor/src/core/config"
 	"github.com/goharbor/harbor/src/core/service/token"
 	"github.com/goharbor/harbor/src/jobservice/job"
-	"github.com/goharbor/harbor/src/jobservice/logger"
 	"github.com/goharbor/harbor/src/lib/errors"
+	"github.com/goharbor/harbor/src/lib/log"
 	"github.com/goharbor/harbor/src/lib/q"
 	"github.com/goharbor/harbor/src/lib/selector"
 	"github.com/goharbor/harbor/src/pkg/p2p/preheat"
@@ -56,6 +56,9 @@ const (
 	resourcePullAction      = "pull"
 	manifestAPIPattern      = "%s/v2/%s/manifests/%s"
 	accessCredHeaderKey     = "Authorization"
+	proMetaKeyContentTrust  = "enable_content_trust"
+	proMetaKeyVulnerability = "prevent_vul"
+	proMetaKeySeverity      = "severity"
 )
 
 // Enforcer defines preheat policy enforcement operations.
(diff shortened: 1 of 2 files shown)


src/controller/p2p/preheat/enforcer.go
```diff
@@ -549,3 +570,55 @@ func checkProviderHealthy(inst *provider.Instance) error {
...
+					code := vuln.Severity(se).Code()
+
+					if sev, ok := fl.Value.(int); !ok || (ok && sev != code) {
+						// Record this is a override case
+						r2 := []interface{}{pro.Name, p.Name, fl.Type, fmt.Sprintf("%v:%d", fl.Value, sev), fmt.Sprintf("%s:%d", se, code)}
+						override = append(override, r2)
+					}
+
+					// Override: must be set align with project configuration setting
+					fl.Value = code
```
chlins, line 616: If the security requirements at the policy level are higher than at the project level, e.g. the project level is above blocking severity High, and the policy level is above blocking severity Low, will this also cover the project level? that means lowering the requirements at the policy level.

steven-zou: good point

steven-zou: @chlins 

Updated

src/controller/p2p/preheat/enforcer.go
```diff
@@ -257,20 +272,26 @@ func (de *defaultEnforcer) PreheatArtifact(ctx context.Context, art *artifact.Ar
 		// Get and check if the provider instance bound with the policy is healthy
 		inst, err := de.instMgr.Get(ctx, pl.ProviderID)
 		if err != nil {
-			logger.Errorf("Failed to get the preheat provider instance bound with the policy %d:%s with error: %s", pl.ID, pl.Name, err.Error())
+			log.Errorf("Failed to get the preheat provider instance bound with the policy %d:%s with error: %s", pl.ID, pl.Name, err.Error())
 			continue
 		}
 
 		// Skip unhealthy instance
 		if err := checkProviderHealthy(inst); err != nil {
```
kofj, line 280: Should we use the controller's `CheckHealth` method instead it?

PR #12483, opened 2020-07-15
Add configuration for anonymous access control

cafeliker:
This is the proposed change on the issue #10760

The administrator will be able to configure the Harbor instance to enable/disable the anonymous access.

Thanks
ye


wy65701436:
hi, this is more like a new feature. As for new feature, please follow this https://github.com/goharbor/harbor/blob/main/CONTRIBUTING.md#design-new-features to create an proposal firstly, and have it reivewed/approved by maintainers.

Let's clost it and move to the proposal part if you still like to introduce it.

PR #12496, opened 2020-07-16
Provide secret manager for proxy cache project

reasonerjt:
This commit provides the secret manager for proxy cache.
The secret is used for pushing blobs to local when it's proxied from
remote registry.
Each secret can be used only once and has a relatively short expiration
time.

### DIFF
--- a/src/pkg/proxy/secret/manager.go
+++ b/src/pkg/proxy/secret/manager.go
@@ -1,0 +1,70 @@
+package secret
+
+import (
+	"sync"
+	"time"
+
+	"github.com/goharbor/harbor/src/common/utils"
+)
+
+const defaultExpiration = 15 * time.Second
+
+type targetRepository struct {
+	name      string
+	expiresAt time.Time
+}
+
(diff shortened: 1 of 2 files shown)


src/pkg/proxy/secret/manager.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func (man *mgr) Verify(sec, rn string) bool {
+	v, ok := man.m.Load(sec)
+	if !ok {
+		return false
+	}
+	p, ok := v.(targetRepository)
+	if ok && p.name == rn {
+		defer man.m.Delete(sec)
```
stonezdj, line 46: What if Generate > Verify ? it will cause memory leak.

reasonerjt: Added GC mechanism in commit:
https://github.com/goharbor/harbor/pull/12496/commits/6b005727ddff8be47bc43d063d1ac6f7527c03a4

PR #12507, opened 2020-07-17
fix(preheat): fix preheat handler PingInstance and UpdateInstance

chlins:
Fix:
1. ping instance always ok
2. update instance not work
3. policy creation_time/update_time nil
4. policy swagger add provider name

### DIFF
--- a/src/portal/src/app/distribution/distribution-instances/distribution-instances.component.ts
+++ b/src/portal/src/app/distribution/distribution-instances/distribution-instances.component.ts
@@ -167,8 +167,10 @@ export class DistributionInstancesComponent implements OnInit, OnDestroy {
       operMessage.state = OperationState.progressing;
       operMessage.data.name = this.selectedRow[0].name;
       this.operationService.publishInfo(operMessage);
+      const instance: Instance = clone(this.selectedRow[0]);
+      instance.default = true;
       this.disService.UpdateInstance({
-          propertySet: {default: true},
+          instance: instance,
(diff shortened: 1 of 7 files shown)


src/portal/src/app/distribution/distribution-instances/distribution-instances.component.ts
```diff
@@ -167,8 +167,10 @@ export class DistributionInstancesComponent implements OnInit, OnDestroy {
       operMessage.state = OperationState.progressing;
       operMessage.data.name = this.selectedRow[0].name;
       this.operationService.publishInfo(operMessage);
+      const instance: Instance = clone(this.selectedRow[0]);
+      instance.default = true;
       this.disService.UpdateInstance({
-          propertySet: {default: true},
+          instance: instance,
```
steven-zou, line 173: as the instance has name property, when do we still need the second param `preheatInstanceName: this.selectedRow[0].name`?

src/server/v2.0/handler/preheat.go
```diff
@@ -350,7 +359,7 @@ func (api *preheatAPI) PingInstances(ctx context.Context, params operation.PingI
 			return operation.NewPingInstancesNotFound()
 		}
 		if err != nil {
-			api.SendError(ctx, err)
+			return api.SendError(ctx, err)
```
steven-zou, line 362: have you check other API handlers to make sure there are not same bugs existing？

chlins: yup

PR #12512, opened 2020-07-17
Include the registry info in the project summary API for proxy cache project

ywk253100:
Include the registry info in the project summary API for proxy cache project

### DIFF
--- a/src/common/models/project.go
+++ b/src/common/models/project.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	"github.com/goharbor/harbor/src/pkg/quota/types"
+	"github.com/goharbor/harbor/src/replication/model"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


src/common/models/project.go
```diff
@@ -211,5 +212,6 @@ type ProjectSummary struct {
 	GuestCount        int64 `json:"guest_count"`
 	LimitedGuestCount int64 `json:"limited_guest_count"`
 
-	Quota *QuotaSummary `json:"quota,omitempty"`
+	Quota    *QuotaSummary   `json:"quota,omitempty"`
+	Registry *model.Registry `json:"registry"`
```
reasonerjt, line 216: Would it expose too much information in the response?
I think all it needs is endpoint and may be ID?

ywk253100: The `URL` itself isn't enough to show the registry information as users can create two registries with the same `URL` but different `Credential`.  So the `ID`, `Name`, `Description`, `Type` and `URL` can describe the overall of a registry.
If so, hiding the other attributes(`Insecure`, `Status`, `CreationTime`, `UpdateTime`) isn't very meaningful?

PR #12518, opened 2020-07-17
Fix #12487: Proxy cache create duplicated operation log

### LINKED ISSUE #12487: Proxy cache create duplicated operation log
When pull a image already cached,  it generate duplicated operation log for creation
Screen Shot 2020-07-15 at 4 21 35 PM

stonezdj:
Change method UseLocalManifest to avoid pull manifest from remote repo frequently

### DIFF
--- a/src/controller/proxy/controller.go
+++ b/src/controller/proxy/controller.go
@@ -30,8 +30,7 @@ import (
 	"github.com/goharbor/harbor/src/lib"
 	"github.com/goharbor/harbor/src/lib/errors"
 	"github.com/goharbor/harbor/src/lib/log"
-	"github.com/goharbor/harbor/src/replication/registry"
-	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/opencontainers/image-spec/specs-go/v1"
 )
 
 const (
@@ -52,7 +51,7 @@ type Controller interface {
 	// UseLocalBlob check if the blob should use local copy
 	UseLocalBlob(ctx context.Context, art lib.ArtifactInfo) bool
 	// UseLocalManifest check manifest should use local copy
-	UseLocalManifest(ctx context.Context, art lib.ArtifactInfo) bool
+	UseLocalManifest(ctx context.Context, art lib.ArtifactInfo) (bool, error)
 	// ProxyBlob proxy the blob request to the remote server, p is the proxy project
 	// art is the ArtifactInfo which includes the digest of the blob
 	ProxyBlob(ctx context.Context, p *models.Project, art lib.ArtifactInfo) (int64, io.ReadCloser, error)
@@ -62,7 +61,6 @@ type Controller interface {
 }
 type controller struct {
 	blobCtl     blob.Controller
-	registryMgr registry.Manager
 	artifactCtl artifact.Controller
 	local       localInterface
 }
@@ -74,7 +72,6 @@ func ControllerInstance() Controller {
 	once.Do(func() {
 		ctl = &controller{
 			blobCtl:     blob.Ctl,
-			registryMgr: registry.NewDefaultManager(),
 			artifactCtl: artifact.Ctl,
 			local:       newLocalHelper(),
 		}
@@ -94,22 +91,23 @@ func (c *controller) UseLocalBlob(ctx context.Context, art lib.ArtifactInfo) boo
 	return exist
 }
 
-func (c *controller) UseLocalManifest(ctx context.Context, art lib.ArtifactInfo) bool {
+func (c *controller) UseLocalManifest(ctx context.Context, art lib.ArtifactInfo) (bool, error) {
 	if len(art.Digest) == 0 {
-		return false
+		return false, nil
 	}
-	return c.local.ManifestExist(ctx, art)
+	a, err := c.local.GetManifest(ctx, art)
+	return a != nil, err
 }
 
 func (c *controller) ProxyManifest(ctx context.Context, p *models.Project, art lib.ArtifactInfo) (distribution.Manifest, error) {
 	var man distribution.Manifest
 	remoteRepo := getRemoteRepo(art)
-	r, err := newRemoteHelper(p.RegistryID)
+	ref := getReference(art)
+	remote, err := newRemoteHelper(p.RegistryID)
 	if err != nil {
 		return man, err
 	}
-	ref := getReference(art)
-	man, err = r.Manifest(remoteRepo, ref)
+	man, dig, err := remote.Manifest(remoteRepo, ref)
 	if err != nil {
 		if errors.IsNotFoundErr(err) {
 			go func() {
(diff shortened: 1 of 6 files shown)


src/server/middleware/repoproxy/proxy.go
```diff
@@ -51,7 +51,7 @@ func handleBlob(w http.ResponseWriter, r *http.Request, next http.Handler) error
 	if err != nil {
 		return err
 	}
-	if !canProxy(p) || proxyCtl.UseLocalBlob(ctx, art) {
+	if proxyCtl.UseLocalBlob(ctx, art) || !canProxy(p) {
```
ywk253100, line 54: What's the difference of the changing? Is there any special reason for that?

ywk253100:
@stonezdj I'm wondering how this PR fix the duplicated operation logs?

src/server/middleware/repoproxy/proxy.go
```diff
@@ -94,12 +94,22 @@ func handleManifest(w http.ResponseWriter, r *http.Request, next http.Handler) e
...
+	if err != nil {
+		log.Errorf("failed to get remote registry, registry id: %v", p.RegistryID)
+		next.ServeHTTP(w, r)
+		return nil
+	}
+	useLocal, err := proxyCtl.UseLocalManifest(ctx, art, remote)
+	if err != nil {
+		return err
+	}
+	if !canProxy(p) || useLocal {
```
ywk253100, line 107: Put the `proxyCtl.UseLocalManifest(ctx, art, remote)` after the checking of `canProxy(p)` to avoid useless checking?

src/server/middleware/repoproxy/proxy.go
stonezdj: Updated, UseLocalBlob will start HTTP request to check the status, it is should always after the canProxy check.

src/server/middleware/repoproxy/proxy.go
stonezdj: yes, updated

stonezdj:
> @stonezdj I'm wondering how this PR fix the duplicated operation logs?

The previous implementation cause the duplicate create operation log, it always fetch the latest manifest and push to the local registry,  always generate a create operation in operation log.  To fix it,  we should compare the local manifest and the remote manifest, if their tags and digest are the same, then just pull it from local instead of pull from remote repo. most of the change is in the UseLocalManifest in src/controller/proxy/controller.go

src/controller/proxy/controller.go
```diff
@@ -94,22 +92,44 @@ func (c *controller) UseLocalBlob(ctx context.Context, art lib.ArtifactInfo) boo
 	return exist
 }
 
-func (c *controller) UseLocalManifest(ctx context.Context, art lib.ArtifactInfo) bool {
-	if len(art.Digest) == 0 {
-		return false
+func (c *controller) UseLocalManifest(ctx context.Context, art lib.ArtifactInfo, remote RemoteInterface) (bool, error) {
```
reasonerjt, line 95: Why in `ProxyManifest` the `remote` is created inside the func but in this case the `remote` is sent by parm

src/controller/proxy/controller.go
```diff
@@ -94,22 +92,44 @@ func (c *controller) UseLocalBlob(ctx context.Context, art lib.ArtifactInfo) boo
...
+	inLocal, lDig, err := c.local.ManifestExist(ctx, art)
```
reasonerjt, line 96: In `ProxyManifest` the manifest was read again, so would it be better if you check whether the digests are the same in `waitAndPushManifest` ?

stonezdj: Update

src/controller/proxy/controller.go
stonezdj: Updated, remove this param

src/controller/proxy/local.go
```diff
@@ -46,15 +49,21 @@ type localInterface interface {
 	CheckDependencies(ctx context.Context, repo string, man distribution.Manifest) []distribution.Descriptor
 	// DeleteManifest cleanup delete tag from local cache
 	DeleteManifest(repo, ref string)
+	// SameArtifact check if same artifact in local
+	SameArtifact(ctx context.Context, repo, tag, dig string) (bool, error)
```
reasonerjt, line 53: This func and `ManifestExists` can be merged into one?

src/controller/proxy/local.go
```diff
@@ -171,7 +197,34 @@ func (l *localHelper) PushManifestList(ctx context.Context, repo string, ref str
...
 	log.Debugf("The manifest list payload: %v", string(pl))
-	return l.PushManifest(repo, ref, newMan)
+	dig := digest.FromBytes(pl)
+	// Because the manifest list maybe updated, need to recheck if it is exist in local
+	inLocal, err := l.SameArtifact(ctx, repo, tag, string(dig))
+	if err != nil {
+		return err
+	}
+	if inLocal {
+		l.SendPullEvent(ctx, repo, tag)
```
reasonerjt, line 207: I don't think you should send a pull event in push func.

src/controller/proxy/local.go
```diff
@@ -63,6 +68,20 @@ type localHelper struct {
 	artifactCtl artifactController
 }
 
+func (l *localHelper) SameArtifact(ctx context.Context, repo, tag, dig string) (bool, error) {
```
reasonerjt, line 71: Why is it needed?

src/controller/proxy/controller.go
```diff
@@ -124,7 +122,24 @@ func (c *controller) ProxyManifest(ctx context.Context, p *models.Project, art l
...
+		a, err := c.local.GetManifest(ctx, art)
+		if err != nil {
+			log.Errorf("failed to check SameArtifact, error %v", err)
+			return
+		}
+		if a == nil {
+			c.waitAndPushManifest(ctx, remoteRepo, man, art, ct, remote)
+			return
+		}
+		if a.Digest == dig {
```
reasonerjt, line 140: Make sure the pull event is sent even the digest is not the same.

src/controller/proxy/controller.go
```diff
@@ -124,7 +122,24 @@ func (c *controller) ProxyManifest(ctx context.Context, p *models.Project, art l
...
+		if len(art.Tag) == 0 {
+			c.waitAndPushManifest(ctx, remoteRepo, man, art, ct, remote)
+			return
+		}
+		a, err := c.local.GetManifest(ctx, art)
+		if err != nil {
+			log.Errorf("failed to check SameArtifact, error %v", err)
+			return
+		}
+		if a == nil {
```
reasonerjt, line 136: what if the digest doesn't equal to the local one?

src/controller/proxy/local.go
stonezdj: Removed

src/controller/proxy/local.go
stonezdj: Removed

src/controller/proxy/controller.go
stonezdj: if the digest doesn't match the local, push to local

src/controller/proxy/local.go
stonezdj: Removed

src/controller/proxy/controller.go

src/controller/proxy/controller.go
```diff
@@ -124,7 +122,26 @@ func (c *controller) ProxyManifest(ctx context.Context, p *models.Project, art l
...
+		}
+
+		// Query artifact after push
+		if a == nil {
+			a, err = c.local.GetManifest(ctx, art)
+			if err != nil {
+				log.Errorf("failed to get manifest, error %v", err)
+			}
+		}
+		if a != nil {
```
reasonerjt, line 142: `else ...`

PR #12525, opened 2020-07-20
Feat: Enhance default processor to handle user-defined artifact like Machine Learning artifacts

hyy0322:
Proposal: https://github.com/goharbor/community/pull/143

Signed-off-by: Yiyang Huang huangyiyang.huangyy@bytedance.com


src/controller/artifact/processor/default.go
```diff
@@ -33,9 +43,10 @@ var (
 // currently, it only tries to parse the artifact type from media type
 type defaultProcessor struct {
 	mediaType string
```
ywk253100, line 45: The property `mediaType` isn't needed anymore as we passing `*artifact.Artifact` as the parameter of functions and it contains the `MediaType` already

src/controller/artifact/processor/default.go
```diff
@@ -44,16 +55,113 @@ func (d *defaultProcessor) GetArtifactType() string {
...
-func (d *defaultProcessor) ListAdditionTypes() []string {
-	return nil
+func (d *defaultProcessor) ListAdditionTypes(ctx context.Context, artifact *artifact.Artifact) []string {
+	artifactConfig := artifact.ExtraAttrs
+	schemaMap, ok := artifactConfig[schemaV1.XHarborAttributes]
+	if !ok {
+		return nil
+	}
+	schema := &schemaV1.Schema{}
+	mapstructure.Decode(schemaMap, schema)
```
ywk253100, line 65: Handle error here

src/controller/artifact/processor/default.go
```diff
@@ -44,16 +55,113 @@ func (d *defaultProcessor) GetArtifactType() string {
...
+	for _, addition := range schema.Additions {
+		additionTypes = append(additionTypes, addition.Name)
+	}
+	return additionTypes
 }
-func (d *defaultProcessor) AbstractMetadata(ctx context.Context, manifest []byte, artifact *artifact.Artifact) error {
-	// do nothing currently
-	// we can extend this function to abstract the metadata in the future if needed
+func (d *defaultProcessor) AbstractMetadata(ctx context.Context, artifact *artifact.Artifact, manifest []byte) error {
+	if artifact.ManifestMediaType != v1.MediaTypeImageManifest {
```
ywk253100, line 73: Add support for docker's manifest media type: https://github.com/goharbor/harbor/blob/master/src/controller/artifact/abstractor.go#L64

src/controller/artifact/processor/default.go
```diff
@@ -44,16 +55,113 @@ func (d *defaultProcessor) GetArtifactType() string {
...
+	// parse metadata from config layer
+	metadata := map[string]interface{}{}
+	if err := json.NewDecoder(blob).Decode(&metadata); err != nil {
+		log.Warningf("parse config layer error: %v", err)
+		return nil
+	}
+	// parse schema
+	schemaMap, ok := metadata[schemaV1.XHarborAttributes]
+	if !ok {
+		return nil
```
ywk253100, line 98: Per my understanding, what we proposed is that the default processor populates all content of the config layer into the extra attributes of artifact even thought there is no `XHarborAttributes` defined. 
Correct me if it not

src/controller/artifact/processor/default.go
```diff
@@ -44,16 +55,113 @@ func (d *defaultProcessor) GetArtifactType() string {
...
+	additionSchemas := make(map[string]schemaV1.AdditionSchema)
+	for _, addition := range schema.Additions {
+		additionSchemas[addition.Name] = addition
+	}
+	if _, ok := additionSchemas[addition]; !ok {
+		return nil, errors.New(nil).WithCode(errors.BadRequestCode).
+			WithMessage("addition %s isn't supported for %s", addition, artifact.Type)
+	}
+	contentType := additionSchemas[addition].ContentType
+	if contentType != "text/plain" {
```
ywk253100, line 148: We don't need to restrict the content type from backend, if UI cannot render it, UI can ignore it or show some error message

src/controller/artifact/processor/default.go
```diff
@@ -44,16 +55,113 @@ func (d *defaultProcessor) GetArtifactType() string {
...
+	}
+	content, err := ioutil.ReadAll(blob)
+	if err != nil {
+		return nil, err
+	}
+	blob.Close()
+
+	return &Addition{
+		Content:     content,
+		ContentType: "text/plain; charset=utf-8",
```
ywk253100, line 165: Use the content type read from the config

src/controller/artifact/processor/default.go
```diff
@@ -44,16 +55,113 @@ func (d *defaultProcessor) GetArtifactType() string {
...
+		log.Warningf("unsupported artifact config schema version %d", schema.SchemaVersion)
+		return nil
+	}
+	// check icon url
+	_, err = url.ParseRequestURI(schema.Icon)
+	if err != nil {
+		log.Warningf("icon url invalid %v", err)
+		schema.Icon = ""
+		m := make(map[string]interface{})
+		b, _ := json.Marshal(schema)
```
wy65701436, line 117: handle error

src/controller/artifact/processor/default.go
```diff
@@ -44,16 +55,113 @@ func (d *defaultProcessor) GetArtifactType() string {
...
-	// return error directly
-	return nil, errors.New(nil).WithCode(errors.BadRequestCode).
-		WithMessage("the processor for artifact %s not found, cannot get the addition", artifact.Type)
+	schemaMap, ok := artifact.ExtraAttrs[schemaV1.XHarborAttributes]
+	if !ok {
+		return nil, errors.New(nil).WithCode(errors.BadRequestCode).
+			WithMessage("the processor for artifact %s not found, cannot get the addition", artifact.Type)
+	}
+	schema := &schemaV1.Schema{}
+	mapstructure.Decode(schemaMap, schema)
```
wy65701436, line 138: handle error

src/controller/artifact/processor/default.go
```diff
@@ -44,16 +55,113 @@ func (d *defaultProcessor) GetArtifactType() string {
...
+	layerDigest := additionSchemas[addition].Digest
+	_, blob, err := d.RegCli.PullBlob(artifact.RepositoryName, layerDigest)
+	if err != nil {
+		return nil, err
+	}
+	content, err := ioutil.ReadAll(blob)
+	if err != nil {
+		return nil, err
+	}
+	blob.Close()
```
wy65701436, line 161: defer？

src/controller/artifact/schema/v1/schema.go
```diff
@@ -0,0 +1,33 @@
...
+const (
+	// XHarborAttributes is the specific key defined in artifact config layer
+	// for user-defined artifact to configure XHarborAttribute JSON schema
+	XHarborAttributes = "xHarborAttributes"
+	// SchemaVersionV1 1
+	SchemaVersionV1 = 1
+)
+
+// Schema define a JSON schema for default processor to extract artifact layer data
+// to artifact model in a specific way
```
wy65701436, line 12: give a example json in the comments?

src/controller/artifact/processor/processor.go
```diff
@@ -35,12 +37,12 @@ type Addition struct {
 // Processor processes specified artifact
 type Processor interface {
 	// GetArtifactType returns the type of one kind of artifact specified by media type
-	GetArtifactType() string
+	GetArtifactType(ctx context.Context, artifact *artifact.Artifact) string
 	// ListAdditionTypes returns the supported addition types of one kind of artifact specified by media type
-	ListAdditionTypes() []string
+	ListAdditionTypes(ctx context.Context, artifact *artifact.Artifact) []string
```
wy65701436, line 42: the implemention doesn't consume the `ctx`, do we really need it? Same as ListAdditionTypes

src/controller/artifact/processor/default.go
wy65701436: +1

src/controller/artifact/processor/default.go
```diff
@@ -44,16 +55,113 @@ func (d *defaultProcessor) GetArtifactType() string {
...
 }
-func (d *defaultProcessor) AbstractMetadata(ctx context.Context, manifest []byte, artifact *artifact.Artifact) error {
-	// do nothing currently
-	// we can extend this function to abstract the metadata in the future if needed
+func (d *defaultProcessor) AbstractMetadata(ctx context.Context, artifact *artifact.Artifact, manifest []byte) error {
+	if artifact.ManifestMediaType != v1.MediaTypeImageManifest {
+		return nil
+	}
+	// get manifest
+	mani := &v1.Manifest{}
```
wy65701436, line 77: shouldn't we use the `distribution.UnmarshalManifest()` to unmarshal manifest?

src/controller/artifact/processor/default.go
hyy0322: In ```ListAdditionTypes()```, we read artifact config from database, which these data was checked when ```AbstractMetadata()```, so I didn't handle error here. Is it neccessary?

src/controller/artifact/processor/default.go
hyy0322: Same as ```ListAdditionTypes()```, we read artifact config from database, which these data was checked when AbstractMetadata(), so I didn't handle error here. Is it neccessary?

src/controller/artifact/processor/default.go
hyy0322: ok

src/controller/artifact/schema/v1/schema.go
hyy0322: There are example json in unittest. I will add example in comments.

src/controller/artifact/processor/processor.go
hyy0322: In last discussion meeting, we agree to unify params for these four functions. Am I right? @steven-zou

src/controller/artifact/processor/default.go
hyy0322: Yes, you are right. I will correct it.

src/controller/artifact/processor/default.go
hyy0322: Ok, thanks

src/controller/artifact/processor/default.go
hyy0322: ok

src/controller/artifact/processor/default.go
hyy0322: No, just json umarshal is enough.

hyy0322:
@ywk253100 @wy65701436 Hi, I have modified code base on all comments.PTAL again.Thanks for your time.

src/controller/artifact/processor/processor.go
```diff
@@ -66,7 +68,9 @@ func Get(mediaType string) Processor {
 	// no registered processor found, use the default one
 	if processor == nil {
 		log.Debugf("the processor for media type %s not found, use the default one", mediaType)
-		processor = &defaultProcessor{mediaType: mediaType}
+		processor = &defaultProcessor{
```
ywk253100, line 71: As we remove the `mediaType` property from the default processor, the default processor can be a singleton and we don't need to create it every time

hyy0322: We have put default processor in processor registry. And we will get processor from it.So is it already singleton mode? I think it is.Do I need to modify code here? @ywk253100

hyy0322:
@wy65701436 ping

src/controller/artifact/processor/processor.go
hyy0322: I have modified this part, PTAL again. Thanks!

ywk253100:
LGTM

src/controller/artifact/processor/default.go
```diff
@@ -16,44 +16,178 @@ package processor
...
+		return nil
+	}
+	// check schema version
+	if schema.SchemaVersion != schemaV1.SchemaVersionV1 {
+		log.Warningf("unsupported artifact config schema version %d", schema.SchemaVersion)
+		return nil
+	}
+	// check icon url
+	_, err = url.ParseRequestURI(schema.Icon)
+	if err != nil {
```
steven-zou, line 126: The logic here is a bit weird.
Why do not copy the original schema text into a var before and use defer func to override the `metadata[schemaV1.XHarborAttributes] `?

src/controller/artifact/processor/default.go
```diff
@@ -16,44 +16,178 @@ package processor
...
+	}
+	// get manifest
+	mani := &v1.Manifest{}
+	if err := json.Unmarshal(manifest, mani); err != nil {
+		log.Warningf("unsupported manifest: %v", err)
+		return nil
+	}
+	// get config layer
+	_, blob, err := d.RegCli.PullBlob(artifact.RepositoryName, mani.Config.Digest.String())
+	if err != nil {
```
steven-zou, line 91: Not very understand return nil error even we get a real error and the return of the func is error type?

Maybe return an error and log all the error (as a warning) at the place of calling this interface method?

hyy0322: If we return real error, it will block artifact pushing to registry. I think whatever user defiend in artifact, we shouldn't block it pushing to registry.

src/controller/artifact/processor/default.go
hyy0322: ```metadata[schemaV1.XHarborAttributes]``` is defined as ```map[string]interface{}``` Because ```artifact.ExtraAttrs``` is defined as ```map[string]interface{}```. If icon is invalid, I need to set it as a empty string. And convert schema struct to ```map[string]interface{}```. What I do here is to covert data to ```map[string]interface{}```.

src/controller/artifact/processor/default.go
```diff
@@ -16,44 +16,178 @@ package processor
...
+	err := mapstructure.Decode(schemaMap, schema)
+	if err != nil {
+		return nil, errors.New(nil).WithCode(errors.BadRequestCode).
+			WithMessage("addition %s isn't supported for %s", addition, artifact.Type)
+	}
+
+	// gather all addition
+	additionSchemas := make(map[string]schemaV1.AdditionSchema)
+	for _, addition := range schema.Additions {
+		additionSchemas[addition.Name] = addition
```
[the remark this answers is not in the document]
hyy0322: Yes, I will correct it.

src/controller/artifact/processor/default.go
```diff
@@ -16,44 +16,178 @@ package processor
...
 }
 func (d *defaultProcessor) AbstractAddition(ctx context.Context, artifact *artifact.Artifact, addition string) (*Addition, error) {
-	// return error directly
-	return nil, errors.New(nil).WithCode(errors.BadRequestCode).
-		WithMessage("the processor for artifact %s not found, cannot get the addition", artifact.Type)
+	// no xHarborAttributes in artifact extraAttrs
+	schemaMap, ok := artifact.ExtraAttrs[schemaV1.XHarborAttributes]
+	if !ok {
+		return nil, errors.New(nil).WithCode(errors.BadRequestCode).
+			WithMessage("the processor for artifact %s not found, cannot get the addition", artifact.Type)
```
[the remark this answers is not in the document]
hyy0322: Agree.

hyy0322:
PTAL again. Thanks. @steven-zou

src/controller/artifact/processor/default.go
steven-zou: @ywk253100 Is that correct?

What I mean here is we can handle the error at the place where we call this func. If we think it should not block the pushing process, we can just log error and by pass it at the calling place.

```go
// ensure the artifact exists under the repository, create it if doesn't exist.
func (c *controller) ensureArtifact(ctx context.Context, repository, digest string) (bool, *artifact.Artifact, error) {
	art, err := c.artMgr.GetByDigest(ctx, repository, digest)
	// the artifact already exists under the repository, return directly
	if err == nil {
		return false, art, nil
	}

	// got other error when get the artifact, return the error
	if !errors.IsErr(err, errors.NotFoundCode) {
		return false, nil, err
	}

	// the artifact doesn't exist under the repository, create it first
	repo, err := c.repoMgr.GetByName(ctx, repository)
	if err != nil {
		return false, nil, err
	}
	artifact := &artifact.Artifact{
		ProjectID:      repo.ProjectID,
		RepositoryID:   repo.RepositoryID,
		RepositoryName: repository,
		Digest:         digest,
		PushTime:       time.Now(),
	}
	// abstract the metadata for the artifact
	if err = c.abstractor.AbstractMetadata(ctx, artifact); err != nil {
	// HERE WE LOG?
              log.warnning(error)
              // return false, nil, err
	}
     
        // ... ignore the left part
}
```

src/controller/artifact/processor/default.go
steven-zou: I think my question is why the invalid icon URL influences the whole data processing? We just need to ignore it (set it to empty str) and continue the processing.

src/controller/artifact/processor/default.go
ywk253100: I prefer to return the error and block the pushing operation.
If users define the customized configurations, what they expect is that the configuration items can be managed and shown in Harbor.
And fail early is helpful for users to realize the error and build their artifacts in Harbor standard way

src/controller/artifact/processor/default.go
hyy0322: What I do here is to set icon to empty str in metadata. Set icon empty in schema struct and convert it to metadata.If just continue to process, icon url still exist in original metadata.

src/controller/artifact/processor/default.go
hyy0322: > I prefer to return the error and block the pushing operation.
> If users define the customized configurations, what they expect is that the configuration items can be managed and shown in Harbor.
> And fail early is helpful for users to realize the error and build their artifacts in Harbor standard way

OK, you got me. I will return real error. Thanks.

src/controller/artifact/processor/default.go
hyy0322: Just ingore it in schema struct is useless, we also need to ignore it in metadata, which will be used in extra_attr.

hyy0322: Discussed with @steven-zou, we will not processor icon URL here. Maybe we will start another PR to process icon

src/controller/artifact/abstractor_test.go
```diff
@@ -198,6 +201,21 @@ var (
 }`
 )
 
+type fakeProcessor struct{}
```
gaocegege, line 204: Can we use gomock to generate such a fake processor?

src/controller/artifact/annotation/checker/v1alpha1.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+var V1Alpha1Checker = NewV1Alpha1Checker()
+
+func NewV1Alpha1Checker() *v1Alpha1Checker {
+	return &v1Alpha1Checker{Version: V1alpha1}
+}
+
+type v1Alpha1Checker struct {
+	Version string
```
gaocegege, line 19: Do we need such a version?

src/controller/artifact/annotation/controller.go
```diff
@@ -0,0 +1,44 @@
...
+	"fmt"
+
+	"github.com/goharbor/harbor/src/controller/artifact/annotation/checker"
+	"github.com/goharbor/harbor/src/pkg/artifact"
+
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+type AnnotationController interface {
+	ProcessAnnotations(artifact *artifact.Artifact, manifest *v1.Manifest, metadata map[string]interface{}) error
```
gaocegege, line 13: If it does not have multiple implementations, maybe we can define it as a struct?

src/controller/artifact/annotation/checker/v1alpha1.go
hyy0322: I think we don't need version anymore.

src/controller/artifact/abstractor_test.go
hyy0322: Tried to use mockery, but when I failed to mock a nil error. So here keep this way currently.

src/controller/artifact/annotation/controller.go
hyy0322: It doesn't matter. Use interface to separate different module is a good convention.

src/controller/artifact/annotation/checker/annotation.go
```diff
@@ -0,0 +1,86 @@
...
+		log.Errorf("Register v1alpha1 annotation Checker error: %v", err)
+	}
+
+	// if err := RegisterV1Alpha2Checker(); err != nil {
+	//	  log.Errorf("Register v1alpha2 annotation checker error: %v", err)
+	// }
+}
+
+// AnnotationChecker process specified annotation
+type AnnotationChecker interface {
```
ywk253100, line 65: Defining each annotation as a method isn't extensible. When we add a new annotation in the v1alpha2, we must add a new method to the interface, that means we must change the implementation of v1alpha1?

ywk253100: How about defining a parse of annotation as following?
```golang
type Parser interface{
    Parse(ctx context.Context, artifact *artifact.Artifact, manifest []byte)(err **error)**
}
```
One parser is responsible for processing all of the annotations belong to one scheme version:
- Parse the annotation from the manifest
- Process the metadata of the artifact according to the annotations
- Store the metadata into the provided artifact model

And there is another parser wrapper which also implements the `Parser` interface, it calls the registered parsers one by one to cover all versions of annotation

hyy0322: Yes, the extensibility is not so good. But it is a clear way for different version of annotation checker. And higher version checker will process annotation first.

src/controller/artifact/annotation/checker/v1alpha1.go
```diff
@@ -0,0 +1,87 @@
...
+	return nil
+}
+
+// ProcessIcon process icon annotation
+func (p *V1Alpha1Checker) ProcessIcon(artifact *artifact.Artifact, digest string) error {
+	// pull icon layer
+	_, icon, err := p.RegCli.PullBlob(artifact.RepositoryName, digest)
+
+	// check the size of the size <= 1MB
+	data, err := ioutil.ReadAll(io.LimitReader(icon, 1<<20))
```
wy65701436, line 68: can we use the image pkg to decode it? Like, img, imgType, err := image.Decode(icon)

hyy0322: You mean wrap this part to a function?

src/controller/artifact/annotation/checker/v1alpha1.go
```diff
@@ -0,0 +1,87 @@
...
+
+	artifact.ExtraAttrs = config
+
+	return nil
+}
+
+// ProcessIcon process icon annotation
+func (p *V1Alpha1Checker) ProcessIcon(artifact *artifact.Artifact, digest string) error {
+	// pull icon layer
+	_, icon, err := p.RegCli.PullBlob(artifact.RepositoryName, digest)
```
wy65701436, line 65: no error hanlding?

src/controller/artifact/annotation/checker/v1alpha1.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+		return err
+	}
+
+	// check the content type
+	contentType := http.DetectContentType(data)
+	switch contentType {
+	case XIcon, GIF, PNG, JPEG:
+	default:
+		return fmt.Errorf("unsupported content type: %s", contentType)
```
wy65701436, line 81: do we want to return a 415 here? If so, you should use the `lib/errors` package, otherwise, it will return a 500 eventually.

src/controller/artifact/annotation/checker/annotation.go
hyy0322: Thought about defined each annotation as interface. But different version annotation checker can not be a same type.

src/controller/artifact/annotation/checker/v1alpha1.go
wy65701436: i think we should leverage the go build-in image library to parse the io.reader to define whether it is a image, and its format.

src/controller/artifact/annotation/checker/v1alpha1.go
hyy0322: Didn't notice this part, I will check it.

src/controller/artifact/annotation/checker/v1alpha1.go
```diff
@@ -0,0 +1,87 @@
...
+
+// ProcessIcon process icon annotation
+func (p *V1Alpha1Checker) ProcessIcon(artifact *artifact.Artifact, digest string) error {
+	// pull icon layer
+	_, icon, err := p.RegCli.PullBlob(artifact.RepositoryName, digest)
+
+	// check the size of the size <= 1MB
+	data, err := ioutil.ReadAll(io.LimitReader(icon, 1<<20))
+	if err != nil {
+		if err == io.EOF {
```
wy65701436, line 70: errors.Is()?

src/controller/artifact/annotation/checker/v1alpha1.go
```diff
@@ -0,0 +1,87 @@
...
+			return fmt.Errorf("the maximum size of the icon is 1MB")
```
wy65701436, line 71: The 1MB is Harbor's limiation? If so, can we explain it at somewhere? Doc?

hyy0322: Process Icon part @ywk253100 helped code this part. 
Maybe it is a limitation we defined? @ywk253100

make/migrations/postgresql/0040_2.1.0_schema.up.sql
```diff
@@ -111,4 +111,6 @@ BEGIN
 END $$;
 
 ALTER TABLE schedule DROP COLUMN IF EXISTS job_id;
-ALTER TABLE schedule DROP COLUMN IF EXISTS status;
\ No newline at end of file
+ALTER TABLE schedule DROP COLUMN IF EXISTS status;
+
+ALTER TABLE artifact ADD COLUMN icon varchar(64);
```
wy65701436, line 116: 64 is enought to store a digtest string?

src/controller/artifact/abstractor_test.go
wy65701436: @heww could help on this.

make/migrations/postgresql/0040_2.1.0_schema.up.sql
hyy0322: ok.

src/controller/artifact/annotation/checker/annotation.go
```diff
@@ -0,0 +1,86 @@
...
+
+	// SortedAnnotationVersionList define the order of AnnotationChecker from high to low version.
+	// High version annotation checker will process annotation first.
+	// In the future we have v1alpha2 version, SortedAnnotationVersionList can be defined like
+	// SortedAnnotationVersionList = []string{V1alpha2, V1alpha1}
+	SortedAnnotationVersionList = []string{V1alpha1}
+)
+
+func init() {
+	if err := RegisterV1Alpha1Checker(); err != nil {
```
wy65701436, line 55: define init() in the RegisterV1Alpha1Checker and RegisterV1Alpha2Checker? and import with `_`?

src/controller/artifact/processor/default.go
```diff
@@ -16,43 +16,117 @@ package processor
...
+		if err := json.NewDecoder(blob).Decode(&metadata); err != nil {
+			return err
+		}
+	}
+
+	// Populate all metadata into the ExtraAttrs first.
+	artifact.ExtraAttrs = metadata
+
+	annotationController := annotation.NewController()
+	err = annotationController.ProcessAnnotations(artifact, mani, metadata)
```
wy65701436, line 120: how about do not return error here? It already has the configuration data at hand, in this case UI just doesn't render the icon. 

@steven-zou

hyy0322: This error has no relation with icon. Icon will be get in another API.

src/controller/artifact/annotation/checker/annotation.go
```diff
@@ -0,0 +1,86 @@
...
+
+var (
+	// Registry for registered annotation checkers
+	Registry = map[string]AnnotationChecker{}
+
+	// SortedAnnotationVersionList define the order of AnnotationChecker from high to low version.
+	// High version annotation checker will process annotation first.
+	// In the future we have v1alpha2 version, SortedAnnotationVersionList can be defined like
+	// SortedAnnotationVersionList = []string{V1alpha2, V1alpha1}
+	SortedAnnotationVersionList = []string{V1alpha1}
```
wy65701436, line 51: If we define `V1alpha3, V1alpha2, V1alpha1`, for the V1alpha1 artifact, it always needs to parsed three times to get icon? and each time will have a pull blob request, right?

src/controller/artifact/annotation/checker/annotation.go
hyy0322: > How about defining a parse of annotation as following?
> 
> ```go
> type Parser interface{
>     Parse(ctx context.Context, artifact *artifact.Artifact, manifest []byte)(err **error)**
> }
> ```
> 
> One parser is responsible for processing all of the annotations belong to one scheme version:
> 
> * Parse the annotation from the manifest
> * Process the metadata of the artifact according to the annotations
> * Store the metadata into the provided artifact model
> 
> And there is another parser wrapper which also implements the `Parser` interface, it calls the registered parsers one by one to cover all versions of annotation

I think I understand your idea, we can discussed it later about the detail.

src/controller/artifact/annotation/parser.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// RegisterAnnotationParser register annotation parser
+func RegisterAnnotationParser(parser Parser, version string) error {
+	_, exist := Registry[version]
+	if exist {
+		return fmt.Errorf("the annotation parser to parser artifact annotation version %s already exists", version)
+	}
+	Registry[version] = parser
+	SortedAnnotationVersionList = append(SortedAnnotationVersionList, version)
```
ywk253100, line 68: How could you guarantee the registered parsers are sorted in order?

hyy0322:
```
	// register v1alpha1 annotation parser
	_ "github.com/goharbor/harbor/src/controller/artifact/annotation/v1alpha1"

	"github.com/goharbor/harbor/src/controller/artifact/annotation"
	"github.com/goharbor/harbor/src/pkg/artifact"
	"github.com/goharbor/harbor/src/pkg/registry"
```
register it in order

src/controller/artifact/annotation/parser.go
```diff
@@ -0,0 +1,76 @@
...
+type Parser interface {
+	// Parse parses annotations in artifact manifest, abstracts data from artifact config layer into the artifact model
+	Parse(ctx context.Context, artifact *artifact.Artifact, manifest []byte) (err error)
+}
+
+type parser struct{}
+
+func (p *parser) Parse(ctx context.Context, artifact *artifact.Artifact, manifest []byte) (err error) {
+	for _, annotationVersion := range SortedAnnotationVersionList {
+		err = GetAnnotationParser(annotationVersion).Parse(ctx, artifact, manifest)
```
ywk253100, line 55: `if errr != nil`

src/controller/artifact/processor/default.go
```diff
@@ -16,43 +16,87 @@ package processor
...
+//           }
+//       },
+//       {
+//           "mediaType": "application/tar+gzip",
+//           "digest": "sha256:d923b93eadde0af5c639a972710a4d919066aba5d0dfbf4b9385099f70272da0",
+//           "size": 166015
+//       }
+//   ]
+// }
+func (d *defaultProcessor) AbstractMetadata(ctx context.Context, artifact *artifact.Artifact, manifest []byte) error {
```
ywk253100, line 89: How about abstracting the metadata from the config layer in the default processor level? This will avoid the config blob reading in every parsers

src/controller/artifact/processor/default.go
```diff
@@ -16,43 +16,87 @@ package processor
...
+	annotationParser := annotation.NewParser()
+	err := annotationParser.Parse(ctx, artifact, manifest)
+	if err != nil {
+		return err
+	}
```
ywk253100, line 94: Please set the icon as `sha256:da834479c923584f4cbcdecc0dac61f32bef1d51e8aae598cf16bd154efab49f` if the icon is empty after parsing.
`sha256:da834479c923584f4cbcdecc0dac61f32bef1d51e8aae598cf16bd154efab49f` is the digest of the default icon introducing by PR #12683

src/controller/artifact/annotation/parser.go
```diff
@@ -0,0 +1,76 @@
...
+	"fmt"
+
+	"github.com/goharbor/harbor/src/lib/log"
+
+	"github.com/goharbor/harbor/src/pkg/artifact"
+)
+
+const (
+	// XIcon is icon content type image/x-icon
+	XIcon = "image/x-icon"
```
ywk253100, line 14: Please remove the `x-icon` as the icon service introducing in #12683 only supports `gif`, `png` and `jpeg`

src/controller/artifact/annotation/parser.go
```diff
@@ -0,0 +1,76 @@
...
+func (p *parser) Parse(ctx context.Context, artifact *artifact.Artifact, manifest []byte) (err error) {
+	for _, annotationVersion := range SortedAnnotationVersionList {
+		err = GetAnnotationParser(annotationVersion).Parse(ctx, artifact, manifest)
+	}
+	log.Error(err)
+	return
+}
+
+// RegisterAnnotationParser register annotation parser
+func RegisterAnnotationParser(parser Parser, version string) error {
```
wy65701436, line 62: is the error necessery for registration？

src/controller/artifact/annotation/parser.go
```diff
@@ -0,0 +1,76 @@
...
+		return fmt.Errorf("the annotation parser to parser artifact annotation version %s already exists", version)
+	}
+	Registry[version] = parser
+	SortedAnnotationVersionList = append(SortedAnnotationVersionList, version)
+	log.Infof("the annotation parser to parser artifact annotation version %s registered", version)
+	return nil
+}
+
+// GetAnnotationParser register annotation parser
+func GetAnnotationParser(version string) Parser {
```
wy65701436, line 74: as we already have GetAnnotationParser to return a parser, should we private the Registry varaible?

src/controller/artifact/annotation/v1alpha1/v1alpha1.go
```diff
@@ -0,0 +1,114 @@
...
+	skipListAnnotationKey := fmt.Sprintf("%s.%s.%s", annotation.AnnotationPrefix, p.version, annotation.SkipList)
+	skipList, ok := mani.Config.Annotations[skipListAnnotationKey]
+	if ok {
+		skipKeyList := strings.Split(skipList, ",")
+		for _, skipKey := range skipKeyList {
+			delete(metadata, skipKey)
+		}
+		artifact.ExtraAttrs = metadata
+	}
+	// parse icon
```
wy65701436, line 78: from here to end, can you just give the default icon instead of returning error? @ywk253100

src/controller/artifact/annotation/parser.go
```diff
@@ -0,0 +1,76 @@
...
+
+// Parser parses annotations in artifact manifest
+type Parser interface {
+	// Parse parses annotations in artifact manifest, abstracts data from artifact config layer into the artifact model
+	Parse(ctx context.Context, artifact *artifact.Artifact, manifest []byte) (err error)
+}
+
+type parser struct{}
+
+func (p *parser) Parse(ctx context.Context, artifact *artifact.Artifact, manifest []byte) (err error) {
```
steven-zou, line 53: manifest will be the same one even there are multiple parsers are registered. Why not use an object ref as an argument and decode the bytes here only once?

src/controller/artifact/annotation/parser.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	Registry[version] = parser
+	SortedAnnotationVersionList = append(SortedAnnotationVersionList, version)
+	log.Infof("the annotation parser to parser artifact annotation version %s registered", version)
+	return nil
+}
+
+// GetAnnotationParser register annotation parser
+func GetAnnotationParser(version string) Parser {
+	return Registry[version]
```
steven-zou, line 75: I don't think we can make sure there will not be any race issue happened (`get` and `register` in different goroutines).

src/controller/artifact/annotation/parser.go
steven-zou: Is it reasonable to depend on the go importing to decide the order?

src/controller/artifact/annotation/v1alpha1/v1alpha1.go
```diff
@@ -0,0 +1,114 @@
...
+	metadata := map[string]interface{}{}
+	// Some artifact may not have empty config layer.
+	if mani.Config.Size != 0 {
+		if err := json.NewDecoder(blob).Decode(&metadata); err != nil {
+			return err
+		}
+	}
+	// Populate all metadata into the ExtraAttrs first.
+	artifact.ExtraAttrs = metadata
+	// parse skip-list
```
steven-zou, line 68:
```suggestion
	// parse skip-list
	// io.goharor.artifact.v1alpha1.skip-list
```

src/controller/artifact/annotation/v1alpha1/v1alpha1.go
steven-zou, line 78:
```suggestion
	// parse icon
	// io.goharor.artifact.v1alpha1.icon
```

src/controller/artifact/annotation/v1alpha1/v1alpha1.go
```diff
@@ -0,0 +1,114 @@
...
+		return
+	}
+}
+
+type v1alpha1Parser struct {
+	version string
+	regCli  registry.Client
+}
+
+func (p *v1alpha1Parser) Parse(ctx context.Context, artifact *artifact.Artifact, manifest []byte) error {
```
steven-zou, line 43: func is a bit long, I think we can use sub-functions(process extra attr, get layer process icon etc.) to complete the logic.

src/controller/artifact/annotation/parser.go
hyy0322: Register is invoked in init when service start, get can't be invoked after service start. I don't think race issues exists.

src/controller/artifact/annotation/parser.go
hyy0322: In the future may be we will support index. So here use []byte for extension. /cc @ywk253100

src/controller/artifact/annotation/parser.go
hyy0322: Register explicitly now in wrapper parser.

src/controller/artifact/annotation/parser.go
steven-zou: ok

PR #12528, opened 2020-07-20
add debugging env for GC time window

wy65701436:
For debugging, the tester/users wants to run GC to delete the removed artifact immediately instead of waitting for two hours, add the env(_GC_BLOB_TIME_WINDOW) to meet this.

### DIFF
--- a/src/common/const.go
+++ b/src/common/const.go
@@ -154,6 +154,6 @@ const (
 	CountPerProject       = "count_per_project"
 	StoragePerProject     = "storage_per_project"
 
-	// ForeignLayer
-	ForeignLayer = "application/vnd.docker.image.rootfs.foreign.diff.tar.gzip"
+	// DefaultGCBlobTimeWindow is the reserve blob time window used by GC, default is 2 hours
+	DefaultGCBlobTimeWindow = 2
 )
--- a/src/common/const.go
+++ b/src/common/const.go
@@ -154,6 +154,6 @@ const (
 	CountPerProject       = "count_per_project"
 	StoragePerProject     = "storage_per_project"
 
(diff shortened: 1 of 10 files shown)


src/common/const.go
```diff
@@ -154,6 +154,6 @@ const (
 	CountPerProject       = "count_per_project"
 	StoragePerProject     = "storage_per_project"
 
-	// ForeignLayer
-	ForeignLayer = "application/vnd.docker.image.rootfs.foreign.diff.tar.gzip"
+	// DefaultGCBlobTimeWindow is the reserve blob time window used by GC, default is 2 hours
+	DefaultGCBlobTimeWindow = 2
```
heww, line 158: DefaultGCBlobTimeWindowHours = int64(2)

src/core/config/config.go
```diff
@@ -478,3 +479,15 @@ func GetPermittedRegistryTypesForProxyCache() []string {
 	}
 	return strings.Split(types, ",")
 }
+
+// GetGCBlobTimeWindow returns the reserve time window of blob.
+func GetGCBlobTimeWindow() int64 {
+	// the env is for testing/debugging. For production, Do NOT set it.
+	if env, exist := os.LookupEnv("_GC_BLOB_TIME_WINDOW"); exist {
```
heww, line 486: `_GC_BLOB_TIME_WINDOW_HOURS`. Do we need `BLOB` in this env ? or `_GC_TIME_WINDOW_HOURS`?

reasonerjt: +1
and I don't think we need the first underscore

wy65701436: updated

src/common/const.go
wy65701436: updated.

PR #12538, opened 2020-07-21
perf(db): skip tx for get blob, patch/put blob upload apis

heww:
1. Skip database transaction middleware for GET Blob and PATCH Blob Upload APIs for there are no database writing operations in them.
2. Skip database transaction middleware for PUT Blob Upload API  and execute database transaction manually in the middlewares in it so that use some short time executions than one long time database transaction execution. Replication use PUT Blob Upload API to upload the whole blob which may spend a long time (minutes level or even more) taking up a database connection if only one database transaction execution in this API.

### DIFF
--- a/src/pkg/distribution/distribution.go
+++ b/src/pkg/distribution/distribution.go
@@ -47,7 +47,11 @@ var (
 var (
 	name      = fmt.Sprintf("(?P<name>%s)", ref.NameRegexp)
 	reference = fmt.Sprintf("(?P<reference>((%s)|(%s)))", ref.DigestRegexp, ref.TagRegexp)
+	dgt       = fmt.Sprintf("(?P<digest>%s)", ref.DigestRegexp)
 	sessionID = "(?P<session_id>[a-zA-Z0-9-_.=]+)"
+
+	// BlobURLRegexp regexp which match blob url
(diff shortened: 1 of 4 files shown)


src/pkg/distribution/distribution.go
```diff
@@ -47,8 +47,12 @@ var (
 var (
 	name      = fmt.Sprintf("(?P<name>%s)", ref.NameRegexp)
 	reference = fmt.Sprintf("(?P<reference>((%s)|(%s)))", ref.DigestRegexp, ref.TagRegexp)
+	dgt       = fmt.Sprintf("(?P<digest>%s)", ref.DigestRegexp)
 	sessionID = "(?P<session_id>[a-zA-Z0-9-_.=]+)"
 
+	// BlobURLRegexp regexp which match blob url
+	BlobURLRegexp = regexp.MustCompile(`^/v2/` + name + `/blobs/` + dgt)
```
wy65701436, line 54: duplicate with `V2BlobUploadURLRe` in ./ser/server/middlware/patterns.go?

heww: I think parse the pattern paramters from /v2 url need to put in this pkg not in src/server/middleware/patterns.go

I will create a issue to track this problem.

PR #12539, opened 2020-07-21
Fix: beego app config port hardcode

ninjadq:
the port should be flexible depend on the internal tls


make/photon/prepare/templates/core/app.conf.jinja
```diff
@@ -1,6 +1,10 @@
 appname = Harbor
-runmode = dev
+runmode = prod
 enablegzip = true
 
-[dev]
-httpport = 8080
+[prod]
+{% if internal_tls.enabled %}
+httpport = {{8080}}
```
bitsf, line 7: 是不是端口写反了

PR #12567, opened 2020-07-23
Schedule preheat policy.

kofj:

### DIFF
--- a/src/controller/p2p/preheat/controller.go
+++ b/src/controller/p2p/preheat/controller.go
@@ -11,6 +11,12 @@ import (
 	providerModels "github.com/goharbor/harbor/src/pkg/p2p/preheat/models/provider"
 	"github.com/goharbor/harbor/src/pkg/p2p/preheat/policy"
 	"github.com/goharbor/harbor/src/pkg/p2p/preheat/provider"
+	"github.com/goharbor/harbor/src/pkg/scheduler"
+)
+
+const (
+	// SchedulerCallback ...
+	SchedulerCallback = "P2PPreheatCallback"
 )
 
 var (
@@ -119,14 +125,16 @@ type controller struct {
 	// For instance
 	iManager instance.Manager
 	// For policy
-	pManager policy.Manager
+	pManager  policy.Manager
+	scheduler scheduler.Scheduler
 }
 
 // NewController is constructor of controller
 func NewController() Controller {
 	return &controller{
-		iManager: instance.Mgr,
-		pManager: policy.Mgr,
+		iManager:  instance.Mgr,
+		pManager:  policy.Mgr,
+		scheduler: scheduler.Sched,
 	}
 }
 
@@ -197,14 +205,45 @@ func (c *controller) CountPolicy(ctx context.Context, query *q.Query) (int64, er
 	return c.pManager.Count(ctx, query)
 }
 
+// TriggerParam ...
+type TriggerParam struct {
+	PolicyID int64
+}
+
 // CreatePolicy creates the policy.
-func (c *controller) CreatePolicy(ctx context.Context, schema *policyModels.Schema) (int64, error) {
+func (c *controller) CreatePolicy(ctx context.Context, schema *policyModels.Schema) (id int64, err error) {
 	if schema != nil {
 		now := time.Now()
 		schema.CreatedAt = now
 		schema.UpdatedTime = now
 	}
-	return c.pManager.Create(ctx, schema)
+
+	id, err = c.pManager.Create(ctx, schema)
+	if err != nil {
+		return
+	}
+
+	if schema.Trigger != nil &&
+		schema.Trigger.Type == policyModels.TriggerTypeScheduled &&
+		len(schema.Trigger.Settings.Cron) > 0 {
+		// schedule and update policy
+		schema.Trigger.Settings.JobID, err = c.scheduler.Schedule(ctx, schema.Trigger.Settings.Cron, SchedulerCallback, TriggerParam{PolicyID: id})
+		if err != nil {
+			return 0, err
+		}
+
+		schema.ID = id
+		err = c.pManager.Update(ctx, schema, "trigger")
+		if err != nil {
+			errUnsch := c.scheduler.UnSchedule(ctx, schema.Trigger.Settings.JobID)
+			if errUnsch != nil {
+				return 0, errUnsch
+			}
+			return 0, err
+		}
+	}
+
+	return
 }
 
 // GetPolicy gets the policy by id.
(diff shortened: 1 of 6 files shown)


src/controller/p2p/preheat/controller.go
```diff
@@ -197,13 +206,34 @@ func (c *controller) CountPolicy(ctx context.Context, query *q.Query) (int64, er
...
+
 // CreatePolicy creates the policy.
 func (c *controller) CreatePolicy(ctx context.Context, schema *policyModels.Schema) (int64, error) {
 	if schema != nil {
 		now := time.Now()
 		schema.CreatedAt = now
 		schema.UpdatedTime = now
 	}
+
+	if schema.Trigger != nil && schema.Trigger.Type == policyModels.TriggerTypeScheduled {
```
steven-zou, line 223: Do we need a rollback operation here?
If schedule is created successfully, but the scheme is failed to create, shall we remove the schedule here?

steven-zou: We can discuss more offline about this error handling logic.

kofj: OK, let's discuss offline

src/controller/p2p/preheat/controllor_test.go
```diff
@@ -295,3 +317,18 @@ func (s *preheatSuite) TestCheckHealth() {
 	err = s.controller.CheckHealth(s.ctx, instance)
 	s.NoError(err)
 }
+
+type fakePreheatScheduler struct {
```
steven-zou, line 321: There is a mock scheduler under `src/testing/pkg/scheduler`

steven-zou: Please use that one for mock cases

src/controller/p2p/preheat/controller.go
```diff
@@ -2,15 +2,23 @@ package preheat
...
 	policyModels "github.com/goharbor/harbor/src/pkg/p2p/preheat/models/policy"
 	providerModels "github.com/goharbor/harbor/src/pkg/p2p/preheat/models/provider"
 	"github.com/goharbor/harbor/src/pkg/p2p/preheat/policy"
 	"github.com/goharbor/harbor/src/pkg/p2p/preheat/provider"
+	"github.com/goharbor/harbor/src/pkg/scheduler"
+)
+
+const (
+	// SchedulerCallback ...
+	SchedulerCallback = "P2PPreheatCallback"
```
steven-zou, line 21: I did not find the callback func definition.

You should implement a callback func and register it with a name like `P2PPreheatCallback` to the scheduler.

In the callback func, you need to enforce the corresponding policy when the schedule signal is coming.

src/controller/p2p/preheat/controller.go
```diff
@@ -197,14 +207,42 @@ func (c *controller) CountPolicy(ctx context.Context, query *q.Query) (int64, er
...
-func (c *controller) CreatePolicy(ctx context.Context, schema *policyModels.Schema) (int64, error) {
+func (c *controller) CreatePolicy(ctx context.Context, schema *policyModels.Schema) (id int64, err error) {
 	if schema != nil {
 		now := time.Now()
 		schema.CreatedAt = now
 		schema.UpdatedTime = now
 	}
-	return c.pManager.Create(ctx, schema)
+
+	if err = orm.WithTransaction(func(ctx context.Context) error {
```
steven-zou, line 224: I don't think the DB transaction can roll back the already created schedule job at the job service side when the scheme is failed to persist.

Maybe we can double-check if the policy with the ID that is returned in the `TriggerParam ` is existing in the DB, if not, then unscheduled the corn job?

src/controller/p2p/preheat/controller.go
```diff
@@ -222,11 +252,87 @@ func (c *controller) UpdatePolicy(ctx context.Context, schema *policyModels.Sche
...
+					needSch = true
+				}
+			}
+		case policyModels.TriggerTypeEventBased:
+			needUn = true
+
+		case policyModels.TriggerTypeManual:
+			needUn = true
+
+		case "":
```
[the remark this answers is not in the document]
steven-zou: Why this comment is resolved. I did not see any answer here.

src/controller/p2p/preheat/controller.go
```diff
@@ -197,14 +207,38 @@ func (c *controller) CountPolicy(ctx context.Context, query *q.Query) (int64, er
...
 		schema.CreatedAt = now
 		schema.UpdatedTime = now
 	}
-	return c.pManager.Create(ctx, schema)
+
+	if err = orm.WithTransaction(func(ctx context.Context) error {
+		if schema.Trigger != nil && schema.Trigger.Type == policyModels.TriggerTypeScheduled {
+			var cron = schema.Trigger.Settings.Cron
+			if len(cron) > 0 {
+				jobid, err := c.scheduler.Schedule(ctx, cron, SchedulerCallback, TriggerParam{PolicyID: schema.ID})
```
steven-zou, line 227: It seems at this time there is no policy ID yet because it is not persisted yet

src/controller/p2p/preheat/controller.go
```diff
@@ -222,11 +256,77 @@ func (c *controller) UpdatePolicy(ctx context.Context, schema *policyModels.Sche
...
+				// schedule new
+				if len(cron) > 0 {
+					// valid cron
+					needSch = true
+				}
+			}
+
+		case "":
+		default:
+			return fmt.Errorf("not support Trigger %s", schema.Trigger.Type)
```
steven-zou, line 293:
```suggestion
			
```

src/controller/p2p/preheat/controller.go
```diff
@@ -222,11 +256,77 @@ func (c *controller) UpdatePolicy(ctx context.Context, schema *policyModels.Sche
...
+		}
+	}
+
+	// unschedule old
+	if needUn {
+		err = c.scheduler.UnSchedule(ctx, oldJobID)
+		if err != nil {
```
steven-zou, line 300: How can we handle the old jobID

src/core/api/base.go
```diff
@@ -202,7 +204,24 @@ func Init() error {
...
+	p2pPreheatCallbackFun := func(p interface{}) error {
+		str, ok := p.(string)
+		if !ok {
+			return fmt.Errorf("the type of param %v isn't string", p)
+		}
+		param := &preheat.TriggerParam{}
+		if err := json.Unmarshal([]byte(str), param); err != nil {
+			return fmt.Errorf("failed to unmarshal the param: %v", err)
+		}
+		_, err := preheat.Enf.EnforcePolicy(orm.Context(), param.PolicyID)
```
steven-zou, line 221: @heww 

There will be db queries in the EnforcePolicy method, is it ok to use `orm.Context()` here to create the context that contains the orm reference?

src/controller/p2p/preheat/controller.go
heww: `orm.WithTransaction` woks in job service side, it is a wrapper for `Begin`, `Commit` `Rollback`.

src/core/api/base.go
heww: `orm.Context` will work after the DB initialized, so it is ok here.

src/controller/p2p/preheat/controller.go
kofj: removed.

src/controller/p2p/preheat/controller.go
kofj: fixed.

src/controller/p2p/preheat/controllor_test.go
kofj: changed

src/controller/p2p/preheat/controller.go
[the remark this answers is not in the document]
kofj: removed

src/controller/p2p/preheat/controller.go
kofj: implemented P2PPreheatCallback func

src/controller/p2p/preheat/controller.go
kofj: if `UnSchedule` the old-one success but `Schedule` the new-one failed, user will always get error when retry update, because the old-one's `jobID` can not found. Can we return success if `jobID` not found when call `UnSchedule`?

PR #12569, opened 2020-07-23
validate references in put manifest

wy65701436:
validate all of references before the manifest request passed into distribution.

1. BadRequest, if any of blob in the manifest is marked as deleting.
2. BadRequest, if any of blob doesn't exist in DB.

After validation success, all of blobs are touched.

### DIFF
--- a/src/server/middleware/blob/put_manifest.go
+++ b/src/server/middleware/blob/put_manifest.go
@@ -38,10 +38,17 @@ func PutManifestMiddleware() func(http.Handler) http.Handler {
 		}
 
 		contentType := r.Header.Get("Content-Type")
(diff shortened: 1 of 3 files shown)


reasonerjt:
Honestly I'm not very sure if this is the right thing to do, most most likely this is not needed but you'll always read and write to DB for each blob the manifest references.

wy65701436:
per discussed with @reasonerjt and @heww, do not introduce this fix for the corner case. If user only call put manifest request, user has to fix it(re-push it) by himself if any blob belongs to the manifest is missing in the following use.

The fix is too heavy, as most of cases do not need to touch blobs twice.

PR #12571, opened 2020-07-24
Limit the permission of secret used by proxy cache service

### LINKED ISSUE #12257: The secret for the proxy to push layer should have minimal permission.
The PR #12214 makes the proxy secret can do anything, we should make sure the secret is short lived and it has minimal permission, i.e. push to a project that has proxy capability, and it expires quickly.

ywk253100:
Limit the permission of secret used by proxy cache service, fixes #12257

### DIFF
--- a/src/lib/patterns.go
+++ b/src/lib/patterns.go
@@ -1,4 +1,4 @@
-package middleware
+package lib
 
 import (
 	"fmt"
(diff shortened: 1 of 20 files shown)


src/pkg/proxy/secret/authorizer.go
```diff
@@ -0,0 +1,63 @@
...
+	return &authorizer{}
+}
+
+type authorizer struct{}
+
+func (s *authorizer) Modify(req *http.Request) error {
+	if req == nil {
+		return errors.New("the request is null")
+	}
+	repository, _, ok := lib.MatchManifestURLPattern(req.URL.Path)
```
stonezdj, line 41: The nested if condition can be refactored to 
```
	if _, _, ok := MatchFunc1(); ok {

	}else if _, _, ok := MatchFunc2(); ok {

	}else if _, _, ok := MatchFunc3(); ok {

	}else {
		//not handler
		return
	}
	//handle it
```

src/lib/patterns.go
```diff
@@ -29,3 +29,33 @@ var (
...
+	strs := V2BlobURLRe.FindStringSubmatch(path)
+	if len(strs) < 3 {
+		return "", "", false
+	}
+	return strs[1], strs[2], true
+}
+
+// MatchBlobUploadURLPattern checks whether the provided path matches the blob upload URL pattern,
+// if does, returns the repository as well
+func MatchBlobUploadURLPattern(path string) (repository string, match bool) {
```
stonezdj, line 55: No need to return the repo, because the logic is already implement in artifact_info.middleware

ywk253100: This is common utility function now.
And there is no artifact info when trying to push the artifact to local in the client side

src/pkg/proxy/secret/authorizer.go
ywk253100: This authorizer is used when trying to push the artifact, this is in the client side, no artifact info in the context

PR #12588, opened 2020-07-27
Enhance: Prolong the timeout of chartclient

ninjadq:
In some perspectives to reduce the too many charts performence issue

### DIFF
--- a/src/chartserver/client.go
+++ b/src/chartserver/client.go
@@ -14,7 +14,7 @@ import (
 )
 
 const (
-	clientTimeout         = 10 * time.Second
+	clientTimeout         = 30 * time.Second
 	maxIdleConnections    = 10
 	idleConnectionTimeout = 30 * time.Second
 )


src/chartserver/client.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 const (
-	clientTimeout         = 10 * time.Second
+	clientTimeout         = 30 * time.Second
```
rafal-prasal, line 17: this is not enough also ResponseHeaderTimeout must be set as in patch below

--- a/src/chartserver/client.go
+++ b/src/chartserver/client.go
@@ -14,9 +14,9 @@ import (
 )
 
 const (
-       clientTimeout         = 10 * time.Second
-       maxIdleConnections    = 10
-       idleConnectionTimeout = 30 * time.Second
+       clientTimeout         = 60 * time.Second
+       maxIdleConnections    = 60
+       idleConnectionTimeout = 60 * time.Second
 )
 
 // ChartClient is a http client to get the content from the external http server
@@ -36,6 +36,7 @@ func NewChartClient(credential *Credential) *ChartClient { // Create http client
                Transport: &http.Transport{
                        MaxIdleConns:    maxIdleConnections,
                        IdleConnTimeout: idleConnectionTimeout,
+                       ResponseHeaderTimeout: clientTimeout,
                },
        }

PR #12602, opened 2020-07-28
Modify column name for preheat task list UI

AllForNothing:


src/portal/src/app/project/p2p-provider/add-p2p-policy/add-p2p-policy.component.html
```diff
@@ -110,7 +110,7 @@ <h3 *ngIf="!isEdit" class="modal-title">{{'P2P_PROVIDER.ADD_POLICY' | translate}
                 <select id="severity" name="severity" class="clr-select"
                         #ngSeverity="ngModel"
                         [(ngModel)]="severity">
-                  <option class="display-none" value=""></option>
+                  <option>--</option>
```
selina-cui, line 113: Why use '--' here ?  I remember we use blank value in other similar scenario.

AllForNothing: They are different.   Severity here is not a required item.  add '--' , so user can select a 'null' value for this item

PR #12604, opened 2020-07-28
refactor(project): add more methods to project controller and manager

heww:


src/common/models/project.go
```diff
@@ -120,11 +124,71 @@ func (p *Project) AutoScan() bool {
 	return isTrue(auto)
 }
 
+// QueryPublic ...
+func (p *Project) QueryPublic(qs orm.QuerySeter, key string, value interface{}) orm.QuerySeter {
```
wy65701436, line 128: what is the `key` for？

src/common/models/project.go
```diff
@@ -146,6 +210,36 @@ type ProjectQueryParam struct {
...
+func (param *ProjectQueryParam) ToQuery() *q.Query {
+	kw := q.KeyWords{}
+	if param.Name != "" {
+		kw["name"] = q.FuzzyMatchValue{Value: param.Name}
+	}
+	if param.Owner != "" {
+		kw["owner"] = param.Owner
+	}
+	if param.Public != nil {
+		kw["public"] = *param.Public
```
wy65701436, line 223: *param.Public？

src/controller/project/controller.go
```diff
@@ -44,19 +49,48 @@ type Controller interface {
...
+		projectID, err = c.projectMgr.Create(ctx, project)
+		if err != nil {
+			return err
+		}
+
+		c.allowlistMgr.CreateEmpty(projectID)
+
+		if len(project.Metadata) > 0 {
+			if err = c.metaMgr.Add(ctx, projectID, project.Metadata); err != nil {
+				log.Errorf("failed to add metadata for project %s: %v", project.Name, err)
```
wy65701436, line 75: only log error?

src/lib/q/query.go
```diff
@@ -29,6 +29,20 @@ type Query struct {
 	Sorting string
 }
 
+// GetString get value of key in Keywords and convert it to string
+func (q *Query) GetString(key string) string {
```
wy65701436, line 33: what if the empty string meet the query condition？

src/pkg/project/dao/dao.go
```diff
@@ -0,0 +1,178 @@
...
+func New() DAO {
+	return &dao{}
+}
+
+type dao struct{}
+
+// Create create a project instance
+func (d *dao) Create(ctx context.Context, project *models.Project) (int64, error) {
+	if project.OwnerID <= 0 {
+		return 0, errors.BadRequestError(nil).WithMessage("owner_id invalid")
```
wy65701436, line 53: make the message readable, it may be rendered in UI.

src/pkg/project/dao/dao.go
```diff
@@ -0,0 +1,178 @@
...
+	sql := `select p.project_id, p.name, p.registry_id, u.username as owner_name, p.owner_id, p.creation_time, p.update_time
+	from project p left join harbor_user u on p.owner_id = u.user_id where p.deleted = false and p.name = ?`
+
+	projects := []*models.Project{}
+	count, err := o.Raw(sql, name).QueryRows(&projects)
+	if err != nil {
+		return nil, err
+	}
+
+	if count == 0 {
```
wy65701436, line 154: in the previous version, the `GetByName` doesn't raise NotFound error, please be care for any break change.

src/pkg/project/manager.go
```diff
@@ -30,37 +31,59 @@ var (
...
-	return &manager{}
+	return &manager{dao: dao.New()}
+}
+
+type manager struct {
+	dao dao.DAO
 }
 
-type manager struct{}
+func (m *manager) Create(ctx context.Context, project *models.Project) (int64, error) {
```
wy65701436, line 56: can we move the project.OwnerID <= 0 checking into manager level?

src/pkg/project/manager.go
```diff
@@ -30,37 +31,59 @@ var (
...
 	if ok {
-		return dao.GetProjectByID(id)
+		return m.dao.Get(ctx, id)
 	}
 	name, ok := idOrName.(string)
 	if ok {
-		return dao.GetProjectByName(name)
+		return m.dao.GetByName(ctx, name)
 	}
 	return nil, fmt.Errorf("invalid parameter: %v, should be ID(int64) or name(string)", idOrName)
```
wy65701436, line 88: user errors.Errorf() instead?

src/common/models/project.go
```diff
@@ -15,9 +15,13 @@
 package models
 
 import (
+	"fmt"
```
wy65701436, line 18: can we move this into pkg/project/model?

src/lib/orm/query.go
```diff
@@ -36,60 +41,152 @@ func QuerySetter(ctx context.Context, model interface{}, query *q.Query, ignored
...
-	// the program will panic when querying the columns that doesn't exist
-	// list the supported columns first to avoid the panic
-	cols := listQueriableCols(model, ignoredCols...)
+	ignored := map[string]bool{}
+	for _, col := range ignoredCols {
+		ignored[snake(col)] = true
+	}
+
+	columns := parseQueriableColumns(model)
+	methods := parseQueriableMethods(model)
```
wy65701436, line 50: add more UT for this.

src/lib/orm/query.go
```diff
@@ -20,10 +20,15 @@ import (
...
 
+// Params ...
+type Params = orm.Params
+
+// QuerySeter ...
+type QuerySeter = orm.QuerySeter
+
 // QuerySetter generates the query setter according to the query. "ignoredCols" is used to set the
 // columns that will not be queried
 func QuerySetter(ctx context.Context, model interface{}, query *q.Query, ignoredCols ...string) (orm.QuerySeter, error) {
```
wy65701436, line 34: generall, please add more comment for the newly added method, otherwise, it's hard to read.

src/common/models/pro_meta.go
```diff
@@ -36,5 +36,4 @@ type ProjectMetadata struct {
 	Value        string    `orm:"column(value)" json:"value"`
 	CreationTime time.Time `orm:"column(creation_time);auto_now_add" json:"creation_time"`
```
wy65701436, line 37: move the model to pkg as well?

src/common/models/project.go
wy65701436, line 128: PublicQuerySeter()?

src/common/models/project.go
```diff
@@ -120,11 +124,71 @@ func (p *Project) AutoScan() bool {
...
 
+// QueryPublic ...
+func (p *Project) QueryPublic(qs orm.QuerySeter, key string, value interface{}) orm.QuerySeter {
+	isPublic := false
+
+	switch v := value.(type) {
+	case bool:
+		isPublic = v
+	case string:
+		isPublic = isTrue(v)
```
wy65701436, line 135: move logic into `isTrue(value interfact{})`?

src/lib/q/query.go
heww: removed as it is not used

src/common/models/project.go
heww: `key` is the raw key in the `q.Query.KeyWords`

src/pkg/project/dao/dao.go
heww: In the beginning, the `Get` and `GetByName` methods of project controller will return `NotFound` error when the project was not found so it's ok for the DAO to return `NotFound` error here.

src/common/models/pro_meta.go
heww: The functions of the project in `common/dao` still use these models so these models will not be moved to pkg until the functions in `common/dao` were removed.

src/common/models/project.go
heww: Yes, convert the pointer to the value because the `QueryPublic` method of `Project` model only accepts `string` and `bool` value to do filter.

src/common/models/project.go
heww: The project-related models still used by functions in `common/dao`, these models will be moved to `pkg/project/models` when the functions about the project in the `common/dao` were removed.

src/common/models/project.go
heww: `FilterPublic` , `FilterMember`, `FilterOwner` or `FilterByPublic`, `FilterByMember`, `FilterByOwner` ?

`XxxQuerySeter` not very match the intention of the function.

src/lib/orm/query.go
```diff
@@ -16,17 +16,49 @@ package orm
...
+	}
+	return strings.Join(placeholders, ",")
+}
+
 // QuerySetter generates the query setter according to the query. "ignoredCols" is used to set the
 // columns that will not be queried
 func QuerySetter(ctx context.Context, model interface{}, query *q.Query, ignoredCols ...string) (orm.QuerySeter, error) {
+	val := reflect.ValueOf(model)
+	if val.Kind() != reflect.Ptr {
+		panic(fmt.Errorf("<orm.QuerySetter> cannot use non-ptr model struct `%s`", getFullName(reflect.Indirect(val).Type())))
```
wy65701436, line 59: don't use panic

src/lib/orm/query.go
```diff
@@ -129,34 +217,46 @@ func listQueriableCols(model interface{}, ignoredCols ...string) map[string]stru
...
 
-// Escape special characters
-func Escape(str string) string {
-	return dao.Escape(str)
+	methods := map[string]string{}
+	defer func() {
+		cache.Store(key, methods)
+	}()
+
+	prefix := "Query"
```
wy65701436, line 246: define a const varaible and gives it more description.

src/lib/orm/query.go
```diff
@@ -16,17 +16,49 @@ package orm
...
 )
 
+// Params ...
+type Params = orm.Params
+
+// QuerySeter ...
+type QuerySeter = orm.QuerySeter
+
+// Escape special characters
+func Escape(str string) string {
```
wy65701436, line 37: duplicate with `escape`?

src/common/models/project.go
heww: Changed to use `FilterByXxx`

src/lib/orm/query.go
ywk253100: +1, please add more comment to explain the mechanism of the query setter, this is helpful for consumers to use it and for the providers to know how to implement it's own filter logic

ywk253100: Such as how the method should be defined for the model that is queried

src/common/security/v2token/context.go
```diff
@@ -5,22 +5,22 @@ import (
...
 // Each request should have a different instance of tokenSecurityCtx
 type tokenSecurityCtx struct {
-	logger    *log.Logger
-	name      string
-	accessMap map[string]map[types.Action]struct{}
-	pm        project.Manager
+	logger     *log.Logger
+	name       string
+	accessMap  map[string]map[types.Action]struct{}
+	getProject func(int64) (*models.Project, error)
```
ywk253100, line 23: Why don't use the project manager directly?

heww: project manager needs ctx as its first argument but the `Can` method can't provide ctx argument so use a function here and the ctx argument from the ctx when create v2token security.

src/controller/project/controller.go
```diff
@@ -44,52 +53,100 @@ type Controller interface {
...
+
+func (c *controller) Create(ctx context.Context, project *models.Project) (int64, error) {
+	var projectID int64
+	h := func(ctx context.Context) (err error) {
+		projectID, err = c.projectMgr.Create(ctx, project)
+		if err != nil {
+			return err
+		}
+
+		c.allowlistMgr.CreateEmpty(projectID)
```
stonezdj, line 77: CreateEmpty will return error, should handle or log it

src/controller/project/controller.go
```diff
@@ -99,9 +156,28 @@ func (c *controller) List(ctx context.Context, query *models.ProjectQueryParam,
 	return projects, nil
 }
 
+func (c *controller) loadOwners(ctx context.Context, projects models.Projects) error {
+	owners, err := c.userMgr.List(ctx, q.New(q.KeyWords{"user_id__in": projects.OwnerIDs()}))
+	if err != nil {
+		return err
+	}
+	m := owners.MapByUserID()
+	for _, project := range projects {
```
stonezdj, line 165: Avoid to use project as var name, because project is imported package name

PR #12610, opened 2020-07-28
Do some refine for the scheduler

ywk253100:
1. Accept vendorType and vendorID when creating the schedule
2. Provide more methods in the scheduler interface to reduce the duplicated works of callers
3. Use a new ormer and transaction when creating the schedule


src/pkg/retention/controller.go
```diff
@@ -82,26 +83,23 @@ func (r *DefaultAPIController) GetRetention(id int64) (*policy.Metadata, error)
 
 // CreateRetention Create Retention
 func (r *DefaultAPIController) CreateRetention(p *policy.Metadata) (int64, error) {
+	id, err := r.manager.CreatePolicy(p)
```
bitsf, line 86: 此处改法和以前逻辑有一个区别，如果Schedule失败了，数据库无法回滚

ywk253100: This should be covered by the database transaction, opened a new issue for that https://github.com/goharbor/harbor/issues/12624

src/pkg/scheduler/scheduler.go
```diff
@@ -130,62 +181,126 @@ func (s *scheduler) Schedule(ctx context.Context, cron string, callbackFuncName
...
+		err = fmt.Errorf("failed to create the schedule: the task status is %s", job.ErrorStatus.String())
+		return 0, err
 	}
 
 	return id, nil
 }
 
-func (s *scheduler) UnSchedule(ctx context.Context, id int64) error {
-	schedule, err := s.dao.Get(ctx, id)
+func (s *scheduler) UnScheduleByID(ctx context.Context, id int64) error {
```
steven-zou, line 192: It's not by schedule id?

src/pkg/scheduler/scheduler.go
```diff
@@ -130,62 +181,126 @@ func (s *scheduler) Schedule(ctx context.Context, cron string, callbackFuncName
...
 			return err
 		}
-		if job.Status(exec.Status).Final() {
-			// delete schedule record
-			if err = s.dao.Delete(ctx, id); err != nil {
+		final := false
+		// after the stop called, the execution cannot be stopped immediately, and the execution
+		// cannot be deleted if it's status isn't in final status, so use the for loop to make
+		// sure the execution be in final status before deleting it
+		for t := 100 * time.Microsecond; t < 5*time.Second; t = t * 2 {
```
steven-zou, line 211: Do we have better solution to handle such cases?

src/pkg/scheduler/scheduler.go
ywk253100: The `id` here is the schedule ID

src/pkg/scheduler/scheduler.go
```diff
@@ -130,62 +181,126 @@ func (s *scheduler) Schedule(ctx context.Context, cron string, callbackFuncName
...
+		sched, err := s.convertSchedule(ctx, schedule)
+		if err != nil {
+			return nil, err
+		}
+		scheds = append(scheds, sched)
+	}
+	return scheds, nil
+}
+
+func (s *scheduler) convertSchedule(ctx context.Context, schedule *schedule) (*Schedule, error) {
```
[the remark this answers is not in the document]
ywk253100: One is the DAO model, the other is the business logic level model

PR #12618, opened 2020-07-29
Refresh the status of execution for every status changing of task

ywk253100:
Refresh the status of execution for every status changing of task to support filtering executions by status directly

### DIFF
--- a/src/pkg/task/dao/execution.go
+++ b/src/pkg/task/dao/execution.go
@@ -16,7 +16,10 @@ package dao
 
 import (
 	"context"
+	"fmt"
+	"github.com/goharbor/harbor/src/lib/log"
 
+	"github.com/goharbor/harbor/src/jobservice/job"
 	"github.com/goharbor/harbor/src/lib/errors"
 	"github.com/goharbor/harbor/src/lib/orm"
 	"github.com/goharbor/harbor/src/lib/q"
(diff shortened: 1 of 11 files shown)


src/server/handler/job_status_hook.go
```diff
@@ -56,7 +56,8 @@ func (j *jobStatusHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if err = j.handler.Handle(r.Context(), taskID, sc); err != nil {
 		// ignore the not found error to avoid the jobservice re-sending the hook
 		if errors.IsNotFoundErr(err) {
-			log.Warningf("got the status change hook for a non existing task %d", taskID)
+			log.Warningf("got not found error when handling the status change hook for task %d, ignore it: %v",
```
steven-zou, line 59:
```suggestion
			log.Errorf("Task %d does not exist, ignore the not_found error to avoid subsequent retrying webhooks from jobservice",
```

src/pkg/task/dao/execution.go
```diff
@@ -132,3 +137,63 @@ func (e *executionDAO) Delete(ctx context.Context, id int64) error {
...
+	if err != nil {
+		return err
+	}
+	// set a score for each status:
+	// 		pending, running, scheduled - 4
+	// 		error - 3
+	//		stopped - 2
+	//		success - 1
+	// and set the status of record with highest score as the status of execution
+	sql := `with status_score as (
```
steven-zou, line 152: SQL is a bit complicated and hard to maintain, can we split the logic as multiple steps/logics?

ywk253100: Use a single one sql to avoid the concurrent issue, will add more comment to explain the sql

steven-zou (approved):
LGTM

PR #12632, opened 2020-07-30
Update configure-https.md for Self-signed certificate instructions

lalithvaka:
V3.ext is updated to include SAN IP in case if any one trying with Self-signed certificate. Also updated the Docker Login instructions for self-signed certificate on Mac


wy65701436:
thanks for the contribution, the current document structure doesn't have mac os specific part, can you add a Q&A in the https://github.com/goharbor/harbor/wiki/Harbor-FAQs?

PR #12637, opened 2020-07-30
https certificate and key path update to include filename

lalithvaka:
Template is confusing to just include path instead of filenames along with path for both "certificate" and "private_key" attributes


ninjadq:
close because path also included the file name part.

lalithvaka:
> close because path also included the file name part.

@ninjadq  As I stated earlier this is confusing. path does not include the filename and if the file name is not included the install fails. Why can't we merge this to have a better documentation?

PR #12641, opened 2020-07-31
Improve p2p UI

AllForNothing:
Temp Env.: https://10.185.12.35:4200
Changed point:
image


selina-cui:
I think we need change the two tool-tips to the same background. The problem is in both dark and light mode.

Screen Shot 2020-08-03 at 11 19 01 AM
Screen Shot 2020-08-03 at 11 19 09 AM

AllForNothing:
> I think we need change the two tool-tips to the same background. The problem is in both dark and light mode.

Fixed as discussion

PR #12650, opened 2020-08-03
update GC job

wy65701436:
Remove the artifact trash record after manifest delete success.

Don't use the flush method is because that when all of records are removed, only GC job knows these informations, and they are in the memory. Once the jobservice is crashed and restarted at GC job execution phase, the trash records are lost, then these manifest are become orphan manifests, cannot be removed any more.

### DIFF
--- a/src/jobservice/job/impl/gc/garbage_collection.go
+++ b/src/jobservice/job/impl/gc/garbage_collection.go
@@ -62,7 +62,7 @@ type GarbageCollector struct {
 	// holds all of trashed artifacts' digest and repositories.
 	// The source data of trashedArts is the table ArtifactTrash and it's only used as a dictionary by sweep when to delete a manifest.
 	// As table blob has no repositories data, and the repositories are required when to delete a manifest, so use the table ArtifactTrash to capture them.
-	trashedArts map[string][]string
+	trashedArts map[string][]model.ArtifactTrash
 	// hold all of GC candidates(non-referenced blobs), it's captured by mark and consumed by sweep.
 	deleteSet       []*blob_models.Blob
 	timeWindowHours int64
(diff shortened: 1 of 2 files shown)


src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -350,24 +357,14 @@ func (gc *GarbageCollector) cleanCache() error {
 // deletedArt contains the two parts of artifact
 // 1, required part, the artifacts were removed from Harbor.
 // 2, optional part, the untagged artifacts.
-func (gc *GarbageCollector) deletedArt(ctx job.Context) (map[string][]string, error) {
+func (gc *GarbageCollector) deletedArt(ctx job.Context) (map[string][]model.ArtifactTrash, error) {
```
heww, line 360: `map[string][]model.ArtifactTrash` => `map[string][]*model.ArtifactTrash`, otherwise for every loop of []model.ArtifactTrash， struct copy will be happend.

src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -397,22 +394,19 @@ func (gc *GarbageCollector) deletedArt(ctx job.Context) (map[string][]string, er
...
 		_, exist := artMap[art.Digest]
 		if !exist {
-			artMap[art.Digest] = []string{art.RepositoryName}
+			artMap[art.Digest] = []model.ArtifactTrash{art}
 		} else {
 			repos := artMap[art.Digest]
-			repos = append(repos, art.RepositoryName)
+			repos = append(repos, art)
 			artMap[art.Digest] = repos
 		}
```
heww, line 406:
```go

trashSlice,  ok := artMap[art.Digest]
if !ok {
    trashSlice = []models.ArtifactTrash{}
}

artMap[art.Digest] = append(trashSlice, art)
```

src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -254,30 +254,37 @@ func (gc *GarbageCollector) sweep(ctx job.Context) error {
 
 		// remove tags and revisions of a manifest
 		if _, exist := gc.trashedArts[blob.Digest]; exist && blob.IsManifest() {
-			for _, repo := range gc.trashedArts[blob.Digest] {
+			for _, art := range gc.trashedArts[blob.Digest] {
```
heww, line 257: loop?

PR #12651, opened 2020-08-03
Add migration 2.1.0

ninjadq:
db_max_open_comms should be 1000 if its value between 100 and 1000

### DIFF
--- a/make/photon/prepare/migrations/version_2_1_0/__init__.py
+++ b/make/photon/prepare/migrations/version_2_1_0/__init__.py
@@ -1,0 +1,30 @@
+import os
+from jinja2 import Environment, FileSystemLoader, StrictUndefined
+from utils.migration import read_conf
+
+revision = '2.1.0'
+down_revisions = ['2.0.0']
+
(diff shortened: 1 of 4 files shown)


make/photon/prepare/migrations/version_2_1_0/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+revision = '2.1.0'
+down_revisions = ['2.0.0']
+
+def migrate(input_cfg, output_cfg):
+    def max_comm_need_update(db_conf):
+        if not db_conf:
+            return False
+        if db_conf.get('max_open_conns') and (db_conf['max_open_conns']<100 or db_conf['max_open_conns']>=1000) :
+                return False
+        return True
```
heww, line 14:
```python
max_open_conns = db_conf.get('max_open_conns', 100)
return 100 <= max_open_conns < 1000
```

ninjadq: If `max_open_conns` in the config file is `None`, this kind of code will cause Exception

PR #12652, opened 2020-08-03
Show the detail error message when failed to fetch the artifacts during replication

ywk253100:
Show the detail error message when failed to fetch the artifacts during replication


steven-zou:
@ywk253100 

It seems there are some issues in the codecov tool. The rate is decreased a lot.

PR #12654, opened 2020-08-03
fix：The project name is not allowed to be a pure numbe

coldy1992:
The project name can not be a pure number, otherwise,the image can not be uploaded to the project.


heww:
The project name can be a pure number in the harbor for compatibility, so this PR may not be accepted.

cc @reasonerjt

PR #12670, opened 2020-08-06
Upload the harbor log icon as it's referenced by the released Harbor chart

ywk253100:
Upload the harbor log icon as it's referenced by the released Harbor chart


wy65701436 (approved):
lgtm

PR #12683, opened 2020-08-07
Implement the icon API to get the icon of artifact

ywk253100:
Implement the icon API to get the icon of artifact


api/v2.0/swagger.yaml
```diff
@@ -1189,6 +1189,31 @@ paths:
...
+      responses:
+        '200':
+          description: Success
+          headers:
+            Cache-Control:
+              description: The cache control header to control the expiration time of the icon
+              type: string
+          schema:
+            $ref: '#/definitions/Icon'
+        '400':
```
wy65701436, line 1211: 401 and 403?

src/controller/icon/controller.go
```diff
@@ -0,0 +1,148 @@
...
+
+	"github.com/goharbor/harbor/src/lib/errors"
+	"github.com/goharbor/harbor/src/lib/q"
+	"github.com/goharbor/harbor/src/pkg/artifact"
+	"github.com/goharbor/harbor/src/pkg/registry"
+	"github.com/nfnt/resize"
+)
+
+// const definitions
+const (
```
wy65701436, line 39: where to store these icons? redis or distribution?

wy65701436: got it, them are read from files.

src/controller/icon/controller.go
```diff
@@ -0,0 +1,148 @@
...
+	DigestOfIconImage   = "sha256:0048162a053eef4d4ce3fe7518615bef084403614f8bca43b40ae2e762e11e06"
+	DigestOfIconChart   = "sha256:61cf3a178ff0f75bf08a25d96b75cf7355dc197749a9f128ed3ef34b0df05518"
+	DigestOfIconCNAB    = "sha256:089bdda265c14d8686111402c8ad629e8177a1ceb7dcd0f7f39b6480f623b3bd"
+	DigestOfIconDefault = "sha256:da834479c923584f4cbcdecc0dac61f32bef1d51e8aae598cf16bd154efab49f"
+)
+
+var (
+	icons = map[string]string{
```
wy65701436, line 47: buildinIcons？

src/controller/icon/controller.go
```diff
@@ -0,0 +1,148 @@
...
+				WithMessage("the icon %s not found", digest)
+		}
+		_, iconFile, err = c.regCli.PullBlob(artifacts[0].RepositoryName, digest)
+		if err != nil {
+			return nil, err
+		}
+		defer iconFile.Close()
+	}
+
+	img, _, err := image.Decode(iconFile)
```
wy65701436, line 122: why doesn't check the format here? any kind of image can be converted to png?

src/server/v2.0/handler/icon.go
```diff
@@ -0,0 +1,51 @@
...
+		ctl: icon.Ctl,
+	}
+}
+
+type iconAPI struct {
+	BaseAPI
+	ctl icon.Controller
+}
+
+func (i *iconAPI) GetIcon(ctx context.Context, params operation.GetIconParams) middleware.Responder {
```
wy65701436, line 41: if err here, how does UI  handler the page render? use the default icons or let the area as empty？

src/controller/icon/controller.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+}
+
+type controller struct {
+	artMgr artifact.Manager
+	regCli registry.Client
+	cache  sync.Map
+}
+
+func (c *controller) Get(ctx context.Context, digest string) (*Icon, error) {
```
wy65701436, line 84: I recall that the redis is the cache in the discussion?

api/v2.0/swagger.yaml
steven-zou: This is a public service that sharing the same icon (with the same digest) among all the projects.

api/v2.0/swagger.yaml
```diff
@@ -1434,6 +1465,9 @@ definitions:
         type: integer
         format: int64
         description: The size of the artifact
+      icon:
+        type: string
+        description: The digest of the icon
```
steven-zou, line 1470: Maybe the icon link is better

src/controller/icon/controller.go
steven-zou: Can we depend on Redis?

steven-zou: Redis has an expire machinery

src/controller/icon/controller.go
```diff
@@ -0,0 +1,144 @@
...
+	img = resize.Thumbnail(50, 50, img, resize.NearestNeighbor)
+
+	// encode the resized icon to png
+	buf := &bytes.Buffer{}
+	if err = png.Encode(buf, img); err != nil {
+		return nil, err
+	}
+
+	icon := &Icon{
+		ContentType: "image/png",
```
steven-zou, line 137: re-echo @wy65701436 's question here, how can we know the format of the icon? The icon annotation does contain the format info?

steven-zou: It seems that you'd like to convert all the icon formats to `png` format, correct?

src/controller/icon/controller.go
ywk253100: The format is checked when pushing the artifact, we only support png, gif and jpg, they can all be converted to png

src/controller/icon/controller.go
ywk253100: Yes, the controller resizes the icon to 50x50 and after that we need to encode the image into a specific format(use png here)

src/server/v2.0/handler/icon.go
ywk253100: I suggest to use a default icon

src/controller/icon/controller.go
ywk253100: I'm thinking after enabling the HTTP cache by `Cache-Control` header, maybe we can remove the memory or redis cache

reasonerjt: but `Cache-Control` only works for one client?  How about the scenario where multiple clients are requesting same icon?

api/v2.0/swagger.yaml
```diff
@@ -1919,3 +1949,12 @@ definitions:
         type: boolean
       default:
         type: boolean
+  Icon:
```
reasonerjt, line 1952: What's the reason you have to put it in a JSON?
Why not put the content-type in the header and return the content in response body directly?

ywk253100: It's requested by the front end guy. 
BTW, the content is based64 encoded, so if we put the content in the response body directly, the content-type header will not be image/xxx either.

src/server/v2.0/handler/icon.go
reasonerjt: Is it better to log the error and return the default icon?
Considering though this is an api, no one will really call it other than UI.  
Oh we should document that too...

ywk253100:
@reasonerjt I have updated the comment about the digest in the icon API

PR #12693, opened 2020-08-10
Fix: The version of prepare image to 2.0

ninjadq:


ninjadq:
close it as the doc no longer managed in harbor repo

PR #12698, opened 2020-08-10
[FIX] Delete executions of the P2P policy.

kofj:

### DIFF
--- a/src/controller/p2p/preheat/controller.go
+++ b/src/controller/p2p/preheat/controller.go
@@ -13,6 +13,7 @@ import (
 	"github.com/goharbor/harbor/src/pkg/p2p/preheat/policy"
 	"github.com/goharbor/harbor/src/pkg/p2p/preheat/provider"
 	"github.com/goharbor/harbor/src/pkg/scheduler"
+	"github.com/goharbor/harbor/src/pkg/task"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


src/controller/p2p/preheat/controller.go
```diff
@@ -108,8 +109,8 @@ type Controller interface {
 	GetPolicyByName(ctx context.Context, projectID int64, name string) (*policyModels.Schema, error)
 	// UpdatePolicy updates the policy.
 	UpdatePolicy(ctx context.Context, schema *policyModels.Schema, props ...string) error
-	// DeletePolicy deletes the policy by id.
-	DeletePolicy(ctx context.Context, id int64) error
+	// DeletePolicyAndExec deletes the policy by id.
+	DeletePolicyAndExec(ctx context.Context, id int64) error
```
steven-zou, line 113: I suggest keeping the deletion method as it was and defining an internal func called in the policy deletion method. API handler just calls the deletion method of CTL.

src/controller/p2p/preheat/controller.go
```diff
@@ -396,9 +393,33 @@ func (c *controller) DeletePolicy(ctx context.Context, id int64) error {
 		}
 	}
 
+	c.deleteExecs(ctx, id)
```
steven-zou, line 396: need to handle error.

steven-zou:
@kofj 

CI failed

PR #12729, opened 2020-08-11
Fix for duplicate webhook policy name

tedgxt:


make/migrations/postgresql/0031_2.0.1_schema.up.sql
```diff
@@ -0,0 +1,4 @@
+/*remove the constraint for name in table 'notification_policy'*/
+ALTER TABLE notification_policy DROP CONSTRAINT notification_policy_name_key;
```
ywk253100, line 2: You should add the sql into the sql file for 2.1: https://github.com/goharbor/harbor/blob/master/make/migrations/postgresql/0040_2.1.0_schema.up.sql

tedgxt: fixed

src/core/api/notification_policy.go
```diff
@@ -128,6 +129,11 @@ func (w *NotificationPolicyAPI) Post() {
 
 	id, err := notification.PolicyMgr.Create(policy)
 	if err != nil {
+		if e := orm.AsConflictError(err, "notification policy named %s already exists", policy.Name); e != nil {
```
wy65701436, line 132: we should put the logic in the dao level, like
https://github.com/goharbor/harbor/blob/master/src/pkg/label/dao.go#L79

tedgxt: considering the 409 / 500 response code should be sent in different circumstances

heww: You can use the `SendError` method which will return the response code depending on the err.

tedgxt: fixed

src/core/api/notification_policy.go
```diff
@@ -180,6 +181,11 @@ func (w *NotificationPolicyAPI) Put() {
 	policy.ProjectID = w.project.ProjectID
 
 	if err = notification.PolicyMgr.Update(policy); err != nil {
+		if e := orm.AsConflictError(err, "notification policy named %s already exists", policy.Name); e != nil {
```
ywk253100, line 184: Why don't we call `SendError()` directly here?

heww: I think he wants to return special error messages for different errors. But I found that the error message in this handler may be duplicated with the error message in dao. Eg, `failed to update the notification policy: failed to update the notification policy: balabala balabala`

tedgxt: already fixed

wy65701436 (approved):
lgtm

ywk253100 (approved):

PR #12736, opened 2020-08-11
fix gc foreign layer issue

### LINKED ISSUE #12717: The freed size by GC doesn't equal with quota
Replicate library/golang:1.14.5 from docker hub, the quota shows 2.46 GB, but after running GC, the log says freed size is 10195 MB.
Maybe it's related with the windows images

wy65701436:
fixes #12717


heww:
If we don't delete the foreign blobs we only need to exclude these blobs in the mark stage. Insert the following lines between 211 and 212.

```go
if blob.IsForeignLayer() {
    continue
}
```

PR #12749, opened 2020-08-12
Tag retention webhook support

tedgxt:


Purushotham233:
@tedgxt, could you add the reference of the issue this PR is fixing?

src/controller/event/handler/webhook/artifact/replication.go
```diff
@@ -8,14 +8,14 @@ import (
 	commonModels "github.com/goharbor/harbor/src/common/models"
 	"github.com/goharbor/harbor/src/controller/event"
 	"github.com/goharbor/harbor/src/controller/event/handler/util"
+	ctlModel "github.com/goharbor/harbor/src/controller/event/model"
```
wy65701436, line 11: evtModel?

src/controller/event/handler/webhook/artifact/retention.go
```diff
@@ -0,0 +1,148 @@
...
+// Handle ...
+func (r *RetentionHandler) Handle(value interface{}) error {
+	if !config.NotificationEnable() {
+		log.Debug("notification feature is not enabled")
+		return nil
+	}
+
+	trEvent, ok := value.(*event.RetentionEvent)
+	if !ok {
+		return errors.New("invalid tag retention event type")
```
wy65701436, line 31: use the `lib/errors` instead

src/controller/event/handler/webhook/artifact/retention.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	if trEvent == nil {
+		return errors.New("nil tag retention event")
+	}
+	if len(trEvent.Deleted) == 0 {
```
wy65701436, line 36: zero deletion is actually a event, why drop it?

src/controller/event/handler/webhook/artifact/retention.go
```diff
@@ -0,0 +1,148 @@
...
+			},
+		},
+	}
+
+	for _, v := range event.Deleted {
+		target := v.Target
+		deletedArtifact := &ctlModel.ArtifactInfo{
+			Type:       target.Kind,
+			Status:     event.Status,
+			NameAndTag: target.Repository + ":" + target.Tags[0],
```
wy65701436, line 132: be careful the `panic`(runtime error: index out of range) here.

src/core/service/notifications/jobs/handler.go
```diff
@@ -205,6 +206,23 @@ func (h *Handler) HandleRetentionTask() {
 			h.SendInternalServerError(err)
 			return
 		}
+
+		e := &event.Event{}
+		metaData := &metadata.RetentionMetaData{
+			Total:    retainObj.Total,
+			Retained: retainObj.Retained,
+			Deleted:  retainObj.Deleted,
+			Status:   "SUCCESS",
```
wy65701436, line 215: always success?

src/controller/event/handler/webhook/artifact/replication.go
tedgxt: fixed

src/controller/event/handler/webhook/artifact/retention.go
tedgxt: fixed

tedgxt: fixed

src/controller/event/handler/webhook/artifact/retention.go
tedgxt: > zero deletion is actually a event, why drop it?

project info can only be found in `Deleted` ， if there is no deletion，we cannot get the webhook policy of the project

src/controller/event/handler/webhook/artifact/retention.go
tedgxt: > be careful the `panic`(runtime error: index out of range) here.

fixed

src/core/service/notifications/jobs/handler.go
tedgxt: > always success?

Tag rentention info will be received only when task begin to  run.      when the task status with success received, there is no retention info with it.

PR #12750, opened 2020-08-12
build base image in CI

wy65701436:
In git action, use the local build base images instead of pulling from docker hub.


reasonerjt:
It makes the test 7mins longer, I think it's acceptable, and there are ways to improve it if we all think it's needed.

PR #12751, opened 2020-08-12
Use a separated database table to store the data version

### LINKED ISSUE #12747: The data migration runs for every installation
This is caused by that the data schema shares a same table `schema_migrations` with database schema, and when database migration runs, it empties the `schema_migrations` table and inserts a new record every time, so the data schema information lost.

ywk253100:
Use a separated database table to store the data version.
Fixes #12747

### DIFF
--- a/make/migrations/postgresql/0040_2.1.0_schema.up.sql
+++ b/make/migrations/postgresql/0040_2.1.0_schema.up.sql
@@ -116,3 +116,15 @@ ALTER TABLE schedule DROP COLUMN IF EXISTS status;
 UPDATE registry SET type = 'quay' WHERE type = 'quay-io';
 
 ALTER TABLE artifact ADD COLUMN icon varchar(255);
+
+CREATE TABLE IF NOT EXISTS data_migrations (
+    version int
+);
+INSERT INTO data_migrations (version) VALUES (
+    CASE
+        /*if the "extra_attrs" isn't null, it means that the deployment upgrades from v2.0*/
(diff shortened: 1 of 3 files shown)


src/migration/artifact.go
```diff
@@ -42,7 +42,7 @@ func abstractArtData(ctx context.Context) error {
 			continue
 		}
 		for _, repo := range repos {
-			log.Debugf("abstracting artifact metadata under repository %s ....", repo.Name)
+			log.Infof("abstracting artifact metadata under repository %s ....", repo.Name)
```
reasonerjt, line 45: Is it possible to add another log message like "listing artifacts for migration..."?
The list may take a long time.

make/migrations/postgresql/0040_2.1.0_schema.up.sql
```diff
@@ -116,3 +116,15 @@ ALTER TABLE schedule DROP COLUMN IF EXISTS status;
 UPDATE registry SET type = 'quay' WHERE type = 'quay-io';
 
 ALTER TABLE artifact ADD COLUMN icon varchar(255);
+
+CREATE TABLE IF NOT EXISTS data_migrations (
```
reasonerjt, line 120: Can we add constraint to make sure there's only one record?

src/migration/artifact.go
ywk253100: The artifact list is read from database

ywk253100:
Will submit another PR to do the update according to the comment

Issue #12745, opened 2020-08-11
Forgot password vanished for db_auth on sign-in page


monofone:
Hi,

we currently have db_auth running and need the ability to let users set their own password. I found the dialog still exists in the sign-in component, but the logic for the link to open the dialog is removed, is this on purpose ?

Regards monofone

ywk253100:
The ability of resetting password has been removed both in back-end and front-end, please refer to https://github.com/goharbor/harbor/issues/10712.

And the dialog in your environment should be the cache of browser

monofone:
Hey,

thanks for the answer, this was the issue I searched for but didn't found it. 
The location I mentioned is here https://github.com/goharbor/harbor/blob/master/src/portal/src/app/sign-in/sign-in.component.html#L47.

Regards monofone

Issue #10712, opened 2020-02-12
remove ability to reset pw 


xaleeks:
In it's current implementation, on a password reset email request, harbor will return different responses if an email exists or not in the database. This is not good practice as it would allow an attacker to check a list of users to see if they have access or not. Either we align the responses or get rid of it altogether.  

For most enterprise users not using local db for auth, this is an unnecessary feature and introduces risk and user can always contact admin for reset as a workaround.  We should vote on whether to deprecate this and make note of this break change in documentation and release notes.

or-hub:
Removing this takes away a very useful component from Harbor.
It is very sad that the "need for more security" cripples an anotherwise good product like Harbor !
Besides - It seems some parts are still present - so only the sign in option was removed ?

More security could have been achieved by leaving  the feature in and just not giving out any information whether an account was indeed present - Many many many other application out there offer this without excessively exposing them.

I d very much like to see this feature to return really - or allow for it to be turned on optionally !

harrishcris:
I agree with @or-hub, this is a very useful expected feature to have and to simply add a response that says "If you have an account then you will shortly recieve a password reset email" would have been a very simple solution to this.

PR #12762, opened 2020-08-12
remove reset password dialog from sign-in component

monofone:
As mentioned in #10712 forgot password should be removed. The
dialog was left in place.
See #12745


ywk253100:
@monofone Could you add your sign off to pass the DCO checking?

PR #12768, opened 2020-08-13
Use orm.Context instead request context in background go routing

### LINKED ISSUE #12741: [proxy cache] sql error in the log
I see such error in the log:
```
Aug 11 12:54:47 172.18.0.1 core[4975]: 2020-08-11T12:54:47Z [ERROR] [/controller/proxy/controller.go:127]: failed to get manifest, error sql: transaction has already been committed or rolled back
```
Is it because the sequence of middlewares caused all DB calls in the proxy controller failed with such error?

stonezdj:
Fixes #12741: error sql: transaction has already been committed or rolled back

### DIFF
--- a/src/controller/proxy/controller.go
+++ b/src/controller/proxy/controller.go
@@ -16,6 +16,8 @@ package proxy
 
 import (
 	"context"
+	"github.com/goharbor/harbor/src/common/security"
+	"github.com/goharbor/harbor/src/lib/orm"
 	"github.com/opencontainers/go-digest"
 	"io"
 	"strings"
@@ -120,9 +122,14 @@ func (c *controller) ProxyManifest(ctx context.Context, p *models.Project, art l
 	if err != nil {
 		return man, err
 	}
+	operator := ""
+	if secCtx, ok := security.FromContext(ctx); ok {
+		operator = secCtx.GetUsername()
+	}
 	// Push manifest in background
 	go func() {
-		a, err := c.local.GetManifest(ctx, art)
+		bCtx := orm.Context()
+		a, err := c.local.GetManifest(bCtx, art)
 		if err != nil {
 			log.Errorf("failed to get manifest, error %v", err)
 		}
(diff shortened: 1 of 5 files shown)


src/controller/proxy/controller.go
```diff
@@ -122,6 +123,7 @@ func (c *controller) ProxyManifest(ctx context.Context, p *models.Project, art l
 	}
 	// Push manifest in background
 	go func() {
+		ctx := orm.Context()
```
ywk253100, line 126: Use `func NewContext(ctx context.Context, o orm.Ormer) context.Context` instead? Or the other information in the context will be lost

heww: I don't think we need to use the `NewContext(ctx context.Context, o orm.Ormer)` here. What will happen when the HTTP goroutine was finished but this goroutine still not be scheduled by the runtime of Golang?

stonezdj: In current case, we need to get the user information from context, update to use the  NewContext(ctx context.Context, o orm.Ormer)

heww: I created an issue for the problem to get the operator of the event, https://github.com/goharbor/harbor/issues/12774.

If we don't get the operator in the `Resolve` of event metadata implicitly, I think we don't need the ctx from the HTTP request in this goroutine.

src/controller/proxy/controller.go
```diff
@@ -122,6 +124,7 @@ func (c *controller) ProxyManifest(ctx context.Context, p *models.Project, art l
 	}
 	// Push manifest in background
 	go func() {
+		ctx := orm.NewContext(ctx, borm.NewOrm())
```
reasonerjt, line 127: The first parm should be context.Background?

src/controller/proxy/controller.go
stonezdj: OK, I will add Operator field in PullArtifactEventMetadata

src/controller/proxy/controller.go
stonezdj: updated

PR #12799, opened 2020-08-18
fix gc job submit issue

wy65701436:
Return 409 on triggering a GC job when there is a job ongoing.


steven-zou:
@wy65701436 

CI failed

heww (approved):

PR #12808, opened 2020-08-18
Update Helm clients to recent editions of Helm v2 and Helm v3

dirkmueller:
This synchronizes the versions used in the e2e image. Also switch
to the 64 bit binaries rather than the legacy 32 bit binary (all
these require 64 bit anyway as they download 64bit binaries for
everything else). Last but not least, it updates helm to versions that
don't have fewer bugs.

Signed-off-by: Dirk Mueller <dirk@dmllr.de>
Signed-off-by: Dirk Mueller <dmueller@suse.com>


dirkmueller:
I don't understand the code coverage failure. do I have to care about this? I certainly didn't change any line of code, so I don't see how that would decrease coverage by 2 percent. it looks like we're including dynamically download or generated files in the code coverage analysis.

danfengliu (approved):
Thanks a lot!
LGTM

wy65701436 (approved):
lgtm

PR #12812, opened 2020-08-19
add transaction for DeletePolicyAndExec.

xiaonancc77:


heww:
Hi, @xiaonancc77  thanks for this PR. We had scheduled to refactor the code of retention feature, and then db operation will be in a transaction by default.  cc @bitsf

Issue #12818, opened 2020-08-19
Support tag deletion replication for Google GCR registry


bitsf:
Google GCR registry can do untag operation, just like harbor does, so we can enable untag operation when setup event-based replication

PR #12819, opened 2020-08-19
fix(replication) gcr deletion and tag deletion

### LINKED ISSUE #12568: The deletion replication of GCR is not work in harbor 2.x
When replicate a deletion operation to GCR, it will fail with error
```
failed to delete the manifest of artifact eminent-nation-87317/replication-ci/mybusybox:test on the destination registry: http status code: 400, body: {"errors":[{"code":"GOOGLE_MANIFEST_DANGLING_TAG","message":"Manifest is still referenced by tag: test"}]}
```

steps:
1. create a event-based replication rule with delete remote feature to Google GCR
2. delete the artifact, it will replicate fail

bitsf:
fix #12568 #12818

### DIFF
--- a/src/pkg/registry/client.go
+++ b/src/pkg/registry/client.go
@@ -93,6 +93,8 @@ type Client interface {
 	// is used to specify whether the destination artifact will be overridden if
 	// its name is same with source but digest isn't
 	Copy(srcRepository, srcReference, dstRepository, dstReference string, override bool) (err error)
+	// Do send generic HTTP requests to the target registry service
+	Do(req *http.Request) (*http.Response, error)
 }
 
 // NewClient creates a registry client with the default authorizer which determines the auth scheme
@@ -504,6 +506,10 @@ func (c *client) Copy(srcRepo, srcRef, dstRepo, dstRef string, override bool) er
 	}
 
 	return nil
+}
+
+func (c *client) Do(req *http.Request) (*http.Response, error) {
+	return c.do(req)
 }
 
 func (c *client) do(req *http.Request) (*http.Response, error) {
--- a/src/replication/adapter/googlegcr/adapter.go
+++ b/src/replication/adapter/googlegcr/adapter.go
@@ -15,10 +15,16 @@
 package googlegcr
 
 import (
+	"encoding/json"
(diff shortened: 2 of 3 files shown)


src/pkg/registry/client.go
```diff
@@ -93,6 +93,8 @@ type Client interface {
 	// is used to specify whether the destination artifact will be overridden if
 	// its name is same with source but digest isn't
 	Copy(srcRepository, srcReference, dstRepository, dstReference string, override bool) (err error)
+	// Do the request
```
ywk253100, line 96: Could you add more comment to explain why this method is needed?

src/pkg/registry/client.go
```diff
@@ -93,6 +93,8 @@ type Client interface {
...
+	Do(req *http.Request) (*http.Response, error)
```
steven-zou, line 97: It seems a common HTTP request handle function, is this the right place to add this func? 
It's very different with other functions in this interface.

steven-zou (commented):
commented

src/pkg/registry/client.go
bitsf: yes, I have to add this in order to call it in adapter

ywk253100 (approved):

src/pkg/registry/client.go
```diff
@@ -93,6 +93,8 @@ type Client interface {
 	// is used to specify whether the destination artifact will be overridden if
 	// its name is same with source but digest isn't
 	Copy(srcRepository, srcReference, dstRepository, dstReference string, override bool) (err error)
+	// Do this is just call the underlying method do(), usually this method should not be called directly.
```
steven-zou, line 96:
```suggestion
	// Do send generic HTTP requests to the target registry service
```

src/pkg/registry/client.go
steven-zou: @bitsf @ywk253100 

Let's raise a debt issue to track the refactor work for the future.

steven-zou (changes requested):
do changes

steven-zou (approved):
LGTM

PR #12826, opened 2020-08-20
CSRF token cookie -> header 

reasonerjt:
The current approach will prevent the effectiveness of `Cache-Control`
header and gorilla's library add `Vary:Cookie` header in all responses.

This PR sets the token in a header of response so the response can be
cached when needed.  It involves code change in both UI and backend middleware.

### DIFF
--- a/src/portal/src/app/services/intercept-http.service.ts
+++ b/src/portal/src/app/services/intercept-http.service.ts
@@ -1,28 +1,58 @@
 import { Injectable } from '@angular/core';
-import { HttpInterceptor, HttpRequest, HttpHandler, HttpEvent, HttpResponse } from '@angular/common/http';
+import { HttpInterceptor, HttpRequest, HttpHandler, HttpResponse } from '@angular/common/http';
 import { Observable, throwError } from 'rxjs';
-import { tap, catchError } from 'rxjs/operators';
-import { CookieService } from 'ngx-cookie';
+import { catchError, tap } from 'rxjs/operators';
+
+const SAFE_METHODS: string[] = ["GET", "HEAD", "OPTIONS", "TRACE"];
 
 @Injectable({
   providedIn: 'root'
 })
 export class InterceptHttpService implements HttpInterceptor {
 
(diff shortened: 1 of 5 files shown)


src/portal/src/app/services/intercept-http.service.ts
```diff
@@ -1,28 +1,56 @@
...
-          request = request.clone({ headers: request.headers.set('X-Harbor-CSRF-Token', Xsrftoken) });
-          return next.handle(request);
-        }
+    // Get the csrf token from localstorage
+    const token = localStorage.getItem("__csrf");
+    if (token) {
+      // Clone the request and replace the original headers with
+      // cloned headers, updated with the csrf token.
+      // not for 'get' request
+      if (request.method && request.method.toLocaleLowerCase() !== 'get') {
```
reasonerjt, line 20: The safe methods should  include:
"GET", "HEAD", "OPTIONS", "TRACE"

ywk253100 (approved):

src/portal/src/app/services/intercept-http.service.ts
AllForNothing: fixed

src/portal/src/app/services/intercept-http.service.ts
```diff
@@ -1,28 +1,58 @@
...
+           }
+         }
+       }))
+      .pipe(
+       catchError(error => {
+         if (error.status === 403) {
+           const csrfToken = localStorage.getItem("__csrf");
+           if (csrfToken) {
+             request = request.clone({ headers: request.headers.set('X-Harbor-CSRF-Token', csrfToken)});
+             return next.handle(request);
```
jwangyangls, line 51: If there is one who doesn't know ，the error results in a 403 all the time, and then there will always be requests

AllForNothing: This is the same as before。 if csrf token is timeout, back end will  return 403 error and  a valid csrf token,  so the next request will be successful

jwangyangls (approved):

PR #12832, opened 2020-08-20
Add id column to data_migration table

ywk253100:
Add id column to data_migration table and add logic to make sure there is only one data version record


wy65701436 (approved):
lgtm

PR #12809, opened 2020-08-18
Update to golang 1.14.7

dirkmueller:
We should use a golang that isn't having security issues.

This includes:
* go1.14.6 (released 2020/07/16) includes fixes to the go command, the
compiler, the linker, vet, and the database/sql, encoding/json,
net/http, reflect, and testing packages. See the Go 1.14.6 milestone on
our issue tracker for details.

* go1.14.7 (released 2020/08/06) includes security fixes to the
encoding/binary package. See the Go 1.14.7 milestone on our issue
tracker for details (CVE-2020-16845)

Signed-off-by: Dirk Mueller <dirk@dmllr.de>
Signed-off-by: Dirk Mueller <dmueller@suse.com>

PR #12835, opened 2020-08-20
Update all references of Golang to 1.14.7

dirkmueller:
In the previous change #12809 I missed a few places. This
should be complete now.


wy65701436 (approved):
lgtm

PR #12874, opened 2020-08-25
fix db migration issue

### LINKED ISSUE #12827: Fix negative repo ID in artifact table if the repo exists in repository table
In 1.x -> 2.0 migration the repo id of an artifact may be set to negative:

```
/*set the media type*/
UPDATE artifact AS art
    SET type='IMAGE', repository_id=repo.repository_id,
    manifest_media_type=blob.content_type,
    media_type=(
    CASE
        /*v2 manifest*/
        WHEN blob.content_type='application/vnd.docker.distribution.manifest.v2+json' THEN
            'application/vnd.docker.container.image.v1+json'
        /*manifest list*/
        WHEN blob.content_type='application/vnd.docker.distribution.manifest.list.v2+json' THEN
            'application/vnd.docker.distribution.manifest.list.v2+json'
        /*v1 manifest*/
        ELSE
            'application/vnd.docker.distribution.manifest.v1+prettyjws'
    END
    )
    FROM repository AS repo, blob AS blob
    WHERE art.repo=repo.name AND art.digest=blob.digest;
/*
It's a workaround for issue https://github.com/goharbor/harbor/issues/11754
The phenomenon is the repository data is gone, but artifacts belong to the repository are still there.
To set the repository_id to a negative, and cannot duplicate.
*/
UPDATE artifact SET repository_id = 0-artifact.id, type='IMAGE', media_type='UNKNOWN', manifest_media_type='UNKNOWN' WHERE repository_id IS NULL;
```

This assume the `repository_id IS NULL`  only happens when there's no corresponding record in the repository table, however, it may also happen when there are inconsistencies between blob table and artifact table.

We should reset the "repo_id" of these records to positive if the repository with the same name exist in repository table.

wy65701436:
fixes #12827
 After user migrates Harbor from v2.0.2, user got 404 when to pull specific images, and no work after push the same images again.

Fix:
1, If the issue is caused by missing repository data, this fix can revert the missing repository data and all things should be fine.
2, If the issue is caused by missing blob data, this fix can revert the missing repository data and still left the media type of artifact
as 'UNKNOWN', which leads the meta data and build history of the image cannot be shown in UI. User can delete and push the image again to
resolve it.


reasonerjt:
Wrong, you need a new SQL file.

reasonerjt:
Please update the sql file name.

ywk253100 (approved):

reasonerjt (approved):

PR #12880, opened 2020-08-25
Use exec in harbor database entrypoint

stefannica:
The harbor-db pod takes a long time to terminate. Using an `exec`
command in the entrypoint ensures that Unix signals reach the
postgres process [1].

[1] https://docs.docker.com/engine/reference/builder/#exec-form-entrypoint-example


heww (approved):

ninjadq (approved):
lgtm

PR #12894, opened 2020-08-27
 Fallback to local repo on errors when proxy to remote repo

### LINKED ISSUE #12853: failed to pull image by proxy-cache after change /etc/hosts
build 3833
1. docker pull busybox by proxy cache
2. change /etc/hosts of harbor-core
image
3. docker pull busybox by proxy cache again, it will fail with 500 Internal Server Error
image
4. log in harbor-core show timeout error
image

This does not meet the description of test case

stonezdj:
When the remote repo is offline or network issue or credential issue, fall back to local repo.
 Fixes #12853

### DIFF
--- a/src/server/middleware/repoproxy/proxy.go
+++ b/src/server/middleware/repoproxy/proxy.go
@@ -111,7 +111,19 @@ func handleManifest(w http.ResponseWriter, r *http.Request, next http.Handler) e
 		return nil
 	}
 	log.Debugf("the tag is %v, digest is %v", art.Tag, art.Digest)
-	man, err := proxyCtl.ProxyManifest(ctx, p, art)
+	err = proxyManifest(ctx, w, r, next, proxyCtl, p, art)
+	if err != nil {
+		if errors.IsNotFoundErr(err) {
+			return err
+		}
+		log.Warningf("Proxy to remote failed, fallback to local repo, error: %v", err)
+		next.ServeHTTP(w, r)
+	}
+	return nil
+}
+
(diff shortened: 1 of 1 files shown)


src/server/middleware/repoproxy/proxy.go
```diff
@@ -113,19 +113,27 @@ func handleManifest(w http.ResponseWriter, r *http.Request, next http.Handler) e
 	log.Debugf("the tag is %v, digest is %v", art.Tag, art.Digest)
 	man, err := proxyCtl.ProxyManifest(ctx, p, art)
 	if err != nil {
-		return err
+		fallbackToLocal(w, r, next, err)
```
ywk253100, line 116: Should we check whether the error is not found?

ywk253100:
Is this a reasonable solution to fallback to local when cannot connect to the upstream registry no matter what the error is? This will cause the pulled artifacts out of sync with the upstream ones.

The timeliness is important for the proxy cache? If it is not, users can just push the artifacts into harbor directly and pull the fixed artifacts from harbor directly

src/server/middleware/repoproxy/proxy.go
stonezdj: OK, return the NotFoundError

reasonerjt (approved):

reasonerjt:
> Is this a reasonable solution to fallback to local when cannot connect to the upstream registry no matter what the error is? This will cause the pulled artifacts out of sync with the upstream ones.
> 
> The timeliness is important for the proxy cache? If it is not, users can just push the artifacts into harbor directly and pull the fixed artifacts from harbor directly

IMO this is a expected behavior of a cache, since we call it cache if it failed to get latest data and it is not expired, it should be served to client

ywk253100 (approved):

PR #12906, opened 2020-08-28
fix(preheat):handle fail case of preheat in job

steven-zou:

### DIFF
--- a/src/pkg/p2p/preheat/job.go
+++ b/src/pkg/p2p/preheat/job.go
@@ -185,9 +185,15 @@ func (j *Job) Run(ctx job.Context, params job.Parameters) error {
 
 			myLogger.Infof("Check preheat progress: %s", s)
 
-			// Finished
-			if s.Status == provider.PreheatingStatusSuccess {
+			switch s.Status {
+			case provider.PreheatingStatusFail:
+				// Fail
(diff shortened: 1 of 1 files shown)


src/pkg/p2p/preheat/job.go
```diff
@@ -185,9 +185,15 @@ func (j *Job) Run(ctx job.Context, params job.Parameters) error {
 
 			myLogger.Infof("Check preheat progress: %s", s)
 
-			// Finished
-			if s.Status == provider.PreheatingStatusSuccess {
+			switch s.Status {
+			case provider.PreheatingStatusFail:
+				// Fail
+				return preheatJobRunningError(errors.Errorf("preheat failed: %s", s))
```
chlins, line 191: errors.Errorf("preheat failed: %s", s),  s is a struct? not a string, so maybe need change to  %v or s.Status.

chlins (approved):
LGTM

src/pkg/p2p/preheat/job.go
steven-zou: s has `String()` method. So it's ok.

```go
func (ps *PreheatingStatus) String() string
```

kofj (approved):
/LGTM

Issue #12891, opened 2020-08-27
Distribution can't be deleted if the related project was deleted instead of deleting p2p prehead policy first


danfengliu:
**Test Step:**
  1. Add a new distribution;
  2. Add a new project and add a new p2p preheat policy ror this project, policy referenced the new distribution;
  3. Delete this new project;

**Expectation Behavior and Actual Behavior:**
  distribution should be deleted successfully, but it's not actrually.

![image](https://user-images.githubusercontent.com/44018302/91376327-6901af80-e84f-11ea-956c-efbc37cb3b2e.png)

steven-zou:
@danfengliu 

It's by design. If some preheat policies have used the provider. It cannot be deleted. To delete it, you need to clean all the related preheat policies first.

danfengliu:
Project deletion should be including P2P preheat policy deletion, so distribution can be deleted.

PR #12917, opened 2020-08-31
[CHERRY-PICK]fix(p2p): delete all preheat policies before delete project

chlins:
Closes:  https://github.com/goharbor/harbor/issues/12891 

### DIFF
--- a/src/controller/p2p/preheat/controller.go
+++ b/src/controller/p2p/preheat/controller.go
@@ -117,6 +117,8 @@ type Controller interface {
 	ListPoliciesByProject(ctx context.Context, project int64, query *q.Query) ([]*policyModels.Schema, error)
 	// CheckHealth checks the instance health, for test connection
 	CheckHealth(ctx context.Context, instance *providerModels.Instance) error
+	// DeletePoliciesOfProject delete all policies under one project
+	DeletePoliciesOfProject(ctx context.Context, project int64) error
 }
 
 var _ Controller = (*controller)(nil)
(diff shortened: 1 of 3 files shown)


src/server/v2.0/handler/project.go
```diff
@@ -212,6 +215,11 @@ func (a *projectAPI) DeleteProject(ctx context.Context, params operation.DeleteP
 		}
 	}
 
+	// preheat policies under the project should be deleted before deleting the project
```
heww, line 218: `before` -> `after`

src/controller/p2p/preheat/controller.go
```diff
@@ -419,6 +421,21 @@ func (c *controller) DeletePolicy(ctx context.Context, id int64) error {
...
+		return err
+	}
+
+	for _, p := range policies {
+		if err = c.DeletePolicy(ctx, p.ID); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
heww, line 437: How many SQLs will be executed here? Can we do this by one SQL?

src/server/v2.0/handler/project.go
chlins: changed

src/controller/p2p/preheat/controller.go
chlins: @heww After review the code, i think we will not have too many policies under the project, and in DeletePolicy, we do extra clean work like unschedule scheduler and delete executions for one policy. So keep the logic.

heww (approved):

steven-zou (approved):
lgtm

PR #12949, opened 2020-09-02
Fix issue of missing swagger package for python API test

danfengliu:
1. Python API test can't import packages from swagger maker, the reason is that python install packages to site-packages instead of dist-packages,
   so the current solution is to copy pakages from site back to dist, I will try to find a way of using ENV variable to fix this;

2. Remove a python library no longer be used.


AllForNothing (approved):

wy65701436 (approved):
lgtm

PR #12973, opened 2020-09-03
Check the tag in isArtifactSigned func -- cherrypick to 2.1

reasonerjt:
This commit ensures that when CLI is pulling a tag, the content trust middleware check the data in notary to ensure the particular tag is signed, not only the digest.


wy65701436 (approved):
lgtm

PR #12984, opened 2020-09-06
Store User ID in session instead of the whole user model

### LINKED ISSUE #12934: Error message is confusing -  admin role is lost in session after upgrade
- install 1.0
- login harbor portal and open repository
- upgrade to v2.0
- navigate to project view (error message popped up
Screen Shot 2020-09-01 at 16 50 28
)

reasonerjt:
This commit makes a change so that the user id will be stored in sessoin
after user login instead of user model to avoid data inconsistency when
user model changes.

Fixes #12934


ywk253100 (approved):

wy65701436 (approved):
lgtm

PR #12986, opened 2020-09-06
fix: Location and Paging support in swagger api

hobti01:
* Add the Location Header to POST responses
  Without the header, you cannot know which object you just created.
* Add the paging Headers and parameters to GET responses for lists where the code already exists
  Without the header, you cannot retrieve the entire list of items.

### DIFF
--- a/api/v2.0/legacy_swagger.yaml
+++ b/api/v2.0/legacy_swagger.yaml
@@ -251,6 +251,10 @@ paths:
       responses:
         '201':
           description: Project member created successfully.
+          headers:
+            Location:
+              type: string
+              description: The URL of the created resource
         '400':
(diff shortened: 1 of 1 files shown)


hobti01:
Hi @ywk253100 , we noticed that these changes were necessary to build a useable client from the swagger API files.  The new and changed interfaces for v2 already have the Headers and parameters.  I've verified to only add the Headers and parameters where the underlying code already implements the functionality.  Please let me know if there are any changes you suggest.

ywk253100 (approved):

ywk253100:
@hobti01 Thanks for the PR.
@heww Please help to review

heww (approved):

PR #12997, opened 2020-09-07
Add ldap filter syntax validation when create search filter

### LINKED ISSUE #12613: LDAP Group Filter and Group Base DN have no affect
When a user authenticates against Harbor LDAP backend, one expects Harbor to lookup the associated LDAP groups and create entries for each, based on the filtering and naming configured through:

-   ldap_group_attribute_name
-   ldap_group_base_dn
-   ldap_group_search_filter

What actually happens is that Harbor ignores the `ldap_group_base_dn`; the `ldap_group_search_filter` and creates all of the groups associated with the user in the database. It also sets the group names to that of the DN attribute and not what is specified through `ldap_group_attribute_name`.

1.  Set LDAP Group configurations to:
```yaml
  "ldap_group_attribute_name": "cn",
  "ldap_group_base_dn": "OU=Containers,OU=Resources,DC=redacted,DC=com",
  "ldap_group_search_filter": "(&(objectclass=group)(cn=APR_CONTAINERS*))",
```

2. Authenticate a user

3. Review created group entries.
```shell
registry=# select * from user_group;
 id  |                                      group_name                                       | group_type |                                     ldap_group_dn                                     |       creation_time        |        update_time         
-----+---------------------------------------------------------------------------------------+------------+---------------------------------------------------------------------------------------+----------------------------+----------------------------
 540 | CN=APR_CyberArk_TSHypervisorServices,OU=CyberArk,OU=Resources,DC=redacted,DC=com          |          1 | cn=apr_cyberark_tshypervisorservices,ou=cyberark,ou=resources,dc=redacted,dc=com          | 2020-07-28 19:11:25.723819 | 2020-07-28 19:11:25.723819
 541 | CN=APR_CyberArk_TSHypervisorServicesApproval,OU=CyberArk,OU=Resources,DC=redacted,DC=com  |          1 | cn=apr_cyberark_tshypervisorservicesapproval,ou=cyberark,ou=resources,dc=redacted,dc=com  | 2020-07-28 19:11:25.725605 | 2020-07-28 19:11:25.725605
```

A. Based on search filter and base DN neither of the above groups should have been created. 
B. Had they matched the filter and base DN, they should have been created with the CN value, not the DN value.

**Versions:**
- harbor version: [2.0.1]

stonezdj:
Correct ldap search filter is enclosed with '(' and ')'
Fixes #12613 LDAP Group Filter and Group Base DN have no affect

### DIFF
--- a/src/common/utils/ldap/filter.go
+++ b/src/common/utils/ldap/filter.go
@@ -1,0 +1,75 @@
+//  Copyright Project Harbor Authors
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package ldap
+
+import (
+	ber "gopkg.in/asn1-ber.v1"
+	goldap "gopkg.in/ldap.v2"
+	"strings"
+)
+
+// FilterBuilder build filter for ldap search
+type FilterBuilder struct {
+	packet *ber.Packet
+}
+
+// Or ...
+func (f *FilterBuilder) Or(filterB *FilterBuilder) *FilterBuilder {
+	if f.packet == nil {
+		return filterB
+	}
+	if filterB.packet == nil {
+		return f
+	}
+	p := ber.Encode(ber.ClassContext, ber.TypeConstructed, goldap.FilterOr, nil, goldap.FilterMap[goldap.FilterOr])
+	p.AppendChild(f.packet)
+	p.AppendChild(filterB.packet)
+	return &FilterBuilder{packet: p}
+}
+
+// And ...
+func (f *FilterBuilder) And(filterB *FilterBuilder) *FilterBuilder {
+	if f.packet == nil {
+		return filterB
+	}
+	if filterB.packet == nil {
+		return f
+	}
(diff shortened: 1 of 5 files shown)


src/common/utils/ldap/filter.go
```diff
@@ -0,0 +1,75 @@
...
+// String ...
+func (f *FilterBuilder) String() (string, error) {
+	if f.packet == nil {
+		return "", nil
+	}
+	return goldap.DecompileFilter(f.packet)
+}
+
+// Parse parse FilterBuilder from string
+func Parse(filter string) (*FilterBuilder, error) {
```
reasonerjt, line 65: This is trivial but how about calling it `NewFilterBuilder(xxx)` otherwise it can be confusing from outside if a caller's code looks like `ldap.Parse(...)`

src/common/utils/ldap/ldap.go
```diff
@@ -401,40 +421,37 @@ func (session *Session) searchGroup(baseDN, filter, groupName, groupNameAttribut
...
+		log.Errorf("invalid ldap filter:%v", gFilterStr)
+		return "", err
 	}
-	return false
+	fb := origFilter.And(gFilter)
+	return fb.String()
 }
 
-// normalizeFilter - remove '(' and ')' in ldap filter
+// normalizeFilter - add '(' and ')' in ldap filter if it doesn't exist
```
reasonerjt, line 444: What if the input is `&(xxx)(xxx)`? after normalize it may look like `(&(xxx)(xxx)`

Can we have a UT for that if I am right?

stonezdj: Thanks, Add UT

src/common/utils/ldap/filter.go
stonezdj: Update

ywk253100 (approved):

reasonerjt (approved):

reasonerjt:
@stonezdj as discussed let's make sure this is cherry picked to 2.1.0 branch so it will be in `v2.1.0`

PR #13043, opened 2020-09-12
fix: Location Header in swagger api common response

hobti01:
* Add the Location Header to POST responses that use the common response
  Without the header, you cannot know which object you just created.


hobti01:
Hi @ywk253100, updating the common response was overlooked in #12986

ywk253100 (approved):

heww (approved):

PR #13085, opened 2020-09-16
Add disable/enable function to replication rules

### LINKED ISSUE #12780: Use separate button to trigger action of `enable`/`disable` a replication policy.
So far, disable/enable replication policy need to open the edit dialog. That is not convenient.

Doing this can also keep consistent with other components that have the similar scenarios.

AllForNothing:
Fixes #12780 
Temp Env: https://10.185.12.35


jwangyangls (approved):
lgtm

src/portal/src/lib/components/list-replication-rule/list-replication-rule.component.ts
```diff
@@ -158,6 +157,35 @@ export class ListReplicationRuleComponent implements OnInit, OnChanges {
         ) {
             this.deleteOpe(message.data);
         }
+        if ( message &&
```
selina-cui, line 160: else if ?

AllForNothing: should not use "else if", they are independent

selina-cui (approved):

PR #13091, opened 2020-09-16
fix(jobservice): batch blob deletions from the db during garbage colection

moshloop:
When running an online (v2.1.0-447c218d) GC that marks 100k+ blobs for deletion it can fail with the following error:

`[ERROR] [/jobservice/job/impl/gc/garbage_collection.go:435]: failed to clean untagged blobs of project, sql: expected 47762 arguments, got 113298`

This PR changes it to try and process only 1000 blobs at once


src/pkg/blob/manager.go
```diff
@@ -95,22 +99,36 @@ func (m *manager) CleanupAssociationsForArtifact(ctx context.Context, artifactDi
...
+		return nil
 	}
-
-	var blobIDs []int64
-	for _, blob := range shouldUnassociatedBlobs {
-		blobIDs = append(blobIDs, blob.ID)
+	for cursor := 0; cursor < len(batches); cursor += deleteBatchSize {
+
+		// create a batch of size deleteBatchSize or the remaining elements
+		blobs := batches[cursor:min(len(batches)-cursor, deleteBatchSize)]
```
heww, line 116: I think here should be `blobs := batches[cursor:min(cursor + deleteBatchSize, len(batches))]`

moshloop: Updated

moshloop:
replaced by https://github.com/goharbor/harbor/pull/14746

PR #13129, opened 2020-09-22
Tiny improvement for the task manager

ywk253100:
Add a new method "StopAndWait" which stops the execution and wait until the execution stopped or get an error


bitsf (approved):

steven-zou (approved):
The UT coverage downgrade too much, please take a look at it.

PR #13132, opened 2020-09-22
Fix internal tls config upgrade issue

ninjadq:
internal tls config upgrade is not included in template, this pr is to add it.


bitsf (approved):

wy65701436 (approved):
lgtm

PR #13139, opened 2020-09-23
Migrate gc to task manager

wy65701436:
1, remove the gc to new programming model
2, move api define to harbor v2 swagger
3, leverage task & execution manager to manage gc job schedule, trigger and log.


src/controller/gc/controller.go
```diff
@@ -0,0 +1,208 @@
...
+	"github.com/goharbor/harbor/src/lib/errors"
+	"github.com/goharbor/harbor/src/lib/log"
+	"github.com/goharbor/harbor/src/lib/q"
+	"github.com/goharbor/harbor/src/pkg/scheduler"
+	"github.com/goharbor/harbor/src/pkg/task"
+)
+
+var (
+	// GCCtl is a global garbage collection controller instance
+	GCCtl = NewController()
```
heww, line 16: Rename it to `Ctl` to keep the same as other controllers.

src/controller/gc/controller.go
```diff
@@ -0,0 +1,208 @@
...
+		return -1, err
+	}
+	defer func() {
+		if err == nil {
+			return
+		}
+		if err := c.taskMgr.Stop(ctx, taskID); err != nil {
+			log.Errorf("failed to stop the task %d: %v", taskID, err)
+		}
+	}()
```
heww, line 86: It seems that these are not needed.

src/controller/gc/controller.go
```diff
@@ -0,0 +1,208 @@
...
+	taskMgr      task.Manager
+	exeMgr       task.ExecutionManager
+	schedulerMgr scheduler.Scheduler
+}
+
+// Start starts the manual GC
+func (c *controller) Start(ctx context.Context, parameters map[string]interface{}) (int64, error) {
+	execID, err := c.exeMgr.Create(ctx, gcVendorType, -1, task.ExecutionTriggerManual, parameters)
+	if err != nil {
+		return -1, err
```
heww, line 67: Please return `0, err` here when there is no special reason.

src/controller/gc/controller.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+	taskID, err := c.taskMgr.Create(ctx, execID, &task.Job{
+		Name: job.ImageGC,
+		Metadata: &job.Metadata{
+			JobKind: job.KindGeneric,
+		},
+		Parameters: parameters,
+	})
+	if err != nil {
+		return -1, err
```
heww, line 77: Please return 0, err here when there is no special reason.

src/controller/gc/controller.go
```diff
@@ -0,0 +1,208 @@
...
+	if err != nil {
+		return nil, err
+	}
+	tasks, err := c.taskMgr.List(ctx, q.New(q.KeyWords{"ExecutionID": exe.ID}))
+	if err != nil {
+		return nil, err
+	}
+	if len(tasks) == 0 {
+		return nil, errors.New(nil).WithCode(errors.NotFoundCode).WithMessage("no gc task is found")
+	}
```
heww, line 125: Duplicate code, these codes also appeared in `GetLog`, can we provide a method `GetByExecutionID` in task manager so that we can use this method in the `GetLog` and `GetByID` methods of controller.

api/v2.0/swagger.yaml
```diff
@@ -1437,6 +1437,153 @@ paths:
...
           $ref: '#/responses/500'
+  /system/gc:
+    get:
+      summary: Get gc results.
+      description: This endpoint let user get gc execution history.
+      tags:
+        - gc
+      operationId: getGCHistory
+      parameters:
+        - $ref: '#/parameters/query'
```
ywk253100, line 1448: Can this API support the `query` directly? Seems the returning `GCHistory` model isn't the underlying database model

api/v2.0/legacy_swagger.yaml
```diff
@@ -1820,25 +1820,6 @@ paths:
...
-            type: array
-            items:
-              $ref: '#/definitions/GCResult'
-        '401':
-          description: User need to log in first.
-        '403':
-          description: User does not have permission of admin role.
-        '500':
-          description: Unexpected internal errors.
   '/system/gc/{id}':
```
ywk253100, line 1823: This should be removed as well?

src/core/api/base.go
```diff
@@ -197,6 +199,8 @@ func Init() error {
 
 	retentionController = retention.NewAPIController(retentionMgr, projectMgr, repository.Mgr, scheduler.Sched, retentionLauncher)
 
+	gcController = gc.NewController()
```
ywk253100, line 202: You can use the global gc controller defined in the package `controller/gc` so that we don't need to declare a new one?

src/controller/gc/controller.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+type controller struct {
+	taskMgr      task.Manager
+	exeMgr       task.ExecutionManager
+	schedulerMgr scheduler.Scheduler
+}
+
+// Start starts the manual GC
+func (c *controller) Start(ctx context.Context, parameters map[string]interface{}) (int64, error) {
```
ywk253100, line 64: Add `trigger` as a parameter of `Start` as `Start` aren't only called by starting GC API, it can be called by the scheduler, which means the trigger is schedule

ywk253100, line 64: `Start(ctx context.Context, parameters map[string]interface{}) (int64, error)`
As this is the interface method specific for GC, the `parameters map[string]interface{}` should be concrete rather than just a `map[string]interface{}`, it's hard for other callers to understand what should be passed

src/controller/gc/controller.go
```diff
@@ -0,0 +1,208 @@
...
+	execID, err := c.exeMgr.Create(ctx, gcVendorType, -1, task.ExecutionTriggerManual, parameters)
```
ywk253100, line 65: Split the `parameters` into separated ones so that the executions can be queries easily according to the customized properties

src/controller/gc/controller.go
```diff
@@ -0,0 +1,208 @@
...
+		}
+		if err := c.taskMgr.Stop(ctx, taskID); err != nil {
+			log.Errorf("failed to stop the task %d: %v", taskID, err)
+		}
+	}()
+	return execID, nil
+}
+
+// Stop ...
+func (c *controller) Stop(ctx context.Context, taskID int64) error {
```
ywk253100, line 91: Use the executionID as the input of the function as the ID returned by `Start` is the execution ID

src/controller/gc/controller.go
```diff
@@ -0,0 +1,208 @@
...
+	return &History{
+		ID:         exe.ID,
+		Name:       gcVendorType,
+		Kind:       exe.Trigger,
+		Parameters: string(extraAttrsString),
+		Deleted:    false,
+		Schedule: Schedule{Schedule: &ScheduleParam{
+			Type: exe.Trigger,
+		}},
+		Status:       tasks[0].Status,
```
ywk253100, line 140: Can we just use the properties of execution so that we don't need to search the task here?
Seems the execution doesn't contain `UpdateTime`, we can open an issue and add it

src/controller/gc/controller.go
wy65701436: updated.

src/controller/gc/controller.go
wy65701436: updated.

src/controller/gc/controller.go
wy65701436: agree, but it's not in the scope of this PR？

api/v2.0/legacy_swagger.yaml
wy65701436: removed.

src/core/api/base.go
wy65701436: updated.

ywk253100:
Is there another PR to cover the upgrade/migration part?

api/v2.0/swagger.yaml
```diff
@@ -2017,6 +2017,153 @@ paths:
...
+          $ref: '#/responses/403'
+        '404':
+          $ref: '#/responses/404'
+        '500':
+          $ref: '#/responses/500'
+  /system/gc/schedule:
+    get:
+      summary: Get gc's schedule.
+      description: This endpoint is for get schedule of gc job.
+      operationId: getSchedule
```
heww, line 2101: `getSchedule` => `getGCSchedule`?

api/v2.0/swagger.yaml
```diff
@@ -2017,6 +2017,153 @@ paths:
...
+          $ref: '#/responses/401'
+        '403':
+          $ref: '#/responses/403'
+        '500':
+          $ref: '#/responses/500'
+    post:
+      summary: Create a gc schedule.
+      description: |
+        This endpoint is for update gc schedule.
+      operationId: postSchedule
```
heww, line 2119: `postSchedule` => `createGCSchedule`?

api/v2.0/swagger.yaml
```diff
@@ -2017,6 +2017,153 @@ paths:
...
+          $ref: '#/responses/403'
+        '409':
+          $ref: '#/responses/409'
+        '500':
+          $ref: '#/responses/500'
+    put:
+      summary: Update gc's schedule.
+      description: |
+        This endpoint is for update gc schedule.
+      operationId: putSchedule
```
heww, line 2146: `putSchedule` => `updateGCSchedule`?

src/server/v2.0/handler/model/gc.go
```diff
@@ -0,0 +1,75 @@
...
+	ScheduleDaily = "Daily"
+	// ScheduleWeekly : 'Weekly'
+	ScheduleWeekly = "Weekly"
+	// ScheduleCustom : 'Custom'
+	ScheduleCustom = "Custom"
+	// ScheduleManual : 'Manual'
+	ScheduleManual = "Manual"
+	// ScheduleNone : 'None'
+	ScheduleNone = "None"
+)
```
heww, line 23: Move these to `util.go`/`const.go` file because scan all will use them too.

src/core/api/base.go
```diff
@@ -217,6 +223,16 @@ func Init() error {
 	}
 	err = scheduler.RegisterCallbackFunc(preheat.SchedulerCallback, p2pPreheatCallbackFun)
 
+	gcCallbackFun := func(ctx context.Context, p string) error {
```
ywk253100, line 226: Check error for the register of p2p callback function in line 224

ywk253100: And suggest to move the registering logic into the package of gc `controller/gc`. After the refactoring in the future, the current file should be removed

src/server/v2.0/handler/gc.go
```diff
@@ -0,0 +1,198 @@
...
+		policy := gc.Policy{
+			ExtraAttrs: parameters,
+		}
+		if dryRun, ok := parameters["dry_run"].(bool); ok {
+			policy.DryRun = dryRun
+		}
+		if deleteUntagged, ok := parameters["delete_untagged"].(bool); ok {
+			policy.DeleteUntagged = deleteUntagged
+		}
+		id, err = g.gcCtr.Start(ctx, policy, task.ExecutionTriggerManual)
```
ywk253100, line 71: check error

heww (approved):

src/server/v2.0/handler/gc.go
wy65701436: it returned at line 86

ywk253100: The error will be override by the code in line 73

wy65701436: it's in the differernt code path, cannot be overrided.

ywk253100 (approved):

PR #13162, opened 2020-09-25
Change the condition of LocalManifest

### LINKED ISSUE #13112: Avoid too many GET request to the dockerhub when checking the latest manifest
When pulling an image by proxy cache, if the image is already cached in local, it should check the latest manifest from remote by HEAD request instead of GET. 
https://www.docker.com/blog/scaling-docker-to-serve-millions-more-developers-network-egress/

stonezdj:
Compare the local digest and the remote digest when pull by tag
Use HEAD request (ManifestExist) instead of GET request (GetManifest)
to avoid been throttled
For manifest list, it can avoid GET request because cached manifest list maybe
differerent with the original manifest list
Make RemoteInterface public
Fixes #13112

### DIFF
--- a/src/controller/proxy/controller.go
+++ b/src/controller/proxy/controller.go
@@ -16,6 +16,7 @@ package proxy
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"strings"
 	"sync"
@@ -53,13 +54,13 @@ type Controller interface {
 	// UseLocalBlob check if the blob should use local copy
 	UseLocalBlob(ctx context.Context, art lib.ArtifactInfo) bool
 	// UseLocalManifest check manifest should use local copy
-	UseLocalManifest(ctx context.Context, art lib.ArtifactInfo) (bool, error)
+	UseLocalManifest(ctx context.Context, art lib.ArtifactInfo, remote RemoteInterface) (bool, error)
 	// ProxyBlob proxy the blob request to the remote server, p is the proxy project
 	// art is the ArtifactInfo which includes the digest of the blob
 	ProxyBlob(ctx context.Context, p *models.Project, art lib.ArtifactInfo) (int64, io.ReadCloser, error)
 	// ProxyManifest proxy the manifest request to the remote server, p is the proxy project,
 	// art is the ArtifactInfo which includes the tag or digest of the manifest
-	ProxyManifest(ctx context.Context, p *models.Project, art lib.ArtifactInfo) (distribution.Manifest, error)
+	ProxyManifest(ctx context.Context, p *models.Project, art lib.ArtifactInfo, remote RemoteInterface) (distribution.Manifest, error)
 }
 type controller struct {
 	blobCtl     blob.Controller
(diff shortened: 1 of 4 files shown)


src/controller/proxy/controller.go
```diff
@@ -93,22 +94,42 @@ func (c *controller) UseLocalBlob(ctx context.Context, art lib.ArtifactInfo) boo
...
-	return a != nil, err
+	// Pull by digest
+	if len(art.Digest) > 0 {
+		return true, nil
+	}
+	// Pull by tag
+	remoteRepo := getRemoteRepo(art)
+	exist, dig, err := remote.ManifestExist(remoteRepo, art.Tag) // HEAD
+	if err != nil {
+		if errors.IsNotFoundErr(err) {
```
ywk253100, line 113: The 404 error should already covered by the registry client, the `exist` is `false` when getting 404

src/controller/proxy/controller.go
```diff
@@ -93,22 +94,42 @@ func (c *controller) UseLocalBlob(ctx context.Context, art lib.ArtifactInfo) boo
...
+				c.local.DeleteManifest(remoteRepo, art.Tag)
+			}()
+		}
+		return true, err
+	}
+	if !exist {
+		go func() {
+			c.local.DeleteManifest(remoteRepo, art.Tag)
+		}()
+		return true, errors.NotFoundError(fmt.Errorf("repo %v, tag %v not found", art.Repository, art.Tag))
```
ywk253100, line 124: Return false?

src/controller/proxy/controller.go
```diff
@@ -53,13 +54,13 @@ type Controller interface {
 	// UseLocalBlob check if the blob should use local copy
 	UseLocalBlob(ctx context.Context, art lib.ArtifactInfo) bool
 	// UseLocalManifest check manifest should use local copy
-	UseLocalManifest(ctx context.Context, art lib.ArtifactInfo) (bool, error)
+	UseLocalManifest(ctx context.Context, art lib.ArtifactInfo, remote RemoteInterface) (bool, error)
```
ywk253100, line 57: I think the `RemoteInterface` can be a structure property of the `Controller` implementation rather than passing as a parameter of the function?

stonezdj: Because the controller is a singleton and initialize only once, it cannot handle the case when there more than 2 proxy servers.

src/controller/proxy/controller.go
stonezdj: Removed

src/controller/proxy/controller.go
stonezdj: handle it in case it is returned by error

src/controller/proxy/controller.go
ywk253100: The `RemoteInterface` is created inside the function implementation for `ProxyBlob`: https://github.com/goharbor/harbor/blob/master/src/controller/proxy/controller.go#L156

Can we keep all of them in a consistent way? 
Personally, I think creating it inside the function is friendly than passing as a parameter for callers

ywk253100 (approved):

src/controller/proxy/controller.go
stonezdj: keep the current impl to reduce the create of remoteInterface.

reasonerjt (approved):

Issue #13098, opened 2020-09-17
The tags API timeout when tags > 2000


stonezdj:
1. Push a image to Harbor and create 2000 tags in this repo.
2.Try to get the tags 
"https://<hostname>/api/repositories/library/nginx-photon/tags"
3. The request is timeout with 504 error 
This issue only ocurr in Harbor version <=1.10.x

It is because the https://github.com/goharbor/harbor/blob/d8336c45496f5139321b08d7a406271185312fa7/src/core/api/repository.go#L654-L653, it send lots of request to the registry and registry need to connect redis to get the tag information. for each request, it needs to open a file, if it reaches the max open file per process, then the redis can not be connected, with the following error:
"error connecting to redis instance redis:6379"

It is not recommended to create tags > 1000 in a single repo when Harbor version <= 1.10.x
This issue can not be reproduced in Harbor 2.x because it doesn't query tags by http request.

bitsf:
we can consider limit workers in the thread pool for this

PR #13168, opened 2020-09-25
feature(tag) add a thread pool size for tags list

bitsf:
fix https://github.com/goharbor/harbor/issues/13098
add a thread pool size for tags list, in case it takes up too much resources

### DIFF
--- a/src/core/api/repository.go
+++ b/src/core/api/repository.go
@@ -649,6 +649,7 @@ func assembleTagsInParallel(client *registry.Repository, projectID int64, reposi
 		}
 	}
 
+	limit := make(chan bool, 100)
 	c := make(chan *models.TagResp)
 	for _, tag := range tags {
 		go assembleTag(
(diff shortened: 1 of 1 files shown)


reasonerjt:
Could you also check the setting on redis to understand the limit? it may be a bottleneck for pulling images concurrently.

src/core/api/repository.go
```diff
@@ -649,6 +649,7 @@ func assembleTagsInParallel(client *registry.Repository, projectID int64, reposi
 		}
 	}
 
+	limit := make(chan bool, 100)
```
steven-zou, line 652: 100 is determined by the DB connection max number?

steven-zou (approved):
LGTM

src/core/api/repository.go
bitsf: just an experience value

heww (approved):

src/core/api/repository.go
```diff
@@ -659,6 +660,7 @@ func assembleTagsInParallel(client *registry.Repository, projectID int64, reposi
 			tag,
 			config.WithNotary(),
 			signatures,
+			limit,
```
ywk253100, line 663: Just a minor suggestion, is it better to get the `true` from `limit` before starting a goroutine? In this way only 100 goroutines will be created

ywk253100 (approved):

PR #13178, opened 2020-09-28
feat(tracing) Add sentry sdk to Portal

holyhope:
Enable it by specifying a sentryDSN in environment file.
Automatically detect production mode.
Disable user feedback ATM.

Signed-off-by: Pierre Péronnet <pierre.peronnet@ovhcloud.com>

### DIFF
--- a/src/portal/package-lock.json
+++ b/src/portal/package-lock.json
@@ -2673,6 +2673,120 @@
                 }
             }
         },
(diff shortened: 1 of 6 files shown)


AllForNothing:
@holyhope Thanks for the PR. But currently, the integration of third-party SDKS is not considered.

PR #13202, opened 2020-10-01
Upgrade the Photon OS version in harbor-db image

### LINKED ISSUE #12522: Clarify the support of different PostgreSQL  versions
AFAICS, Harbor only mentions the requirement for **High available PostgreSQL database** when installing via Helm chart and requiring external database.

When internal database is deployed, photon-based `harbor-db` image actually uses version 9.6.14.

So the question/AI is this: documentation should explain what version is required, if there's hard requirement.

Or, even better, official image could be bumped to use more recent version, or such goal could be present at some Roadmap.

FWIW, I've tested it myself with (SUSE based) postgresql12 image and everything seems to work fine.

jsuchome:
Using latest stable version of OS (photon:3.0) implies
using somewhat newer version of PostgreSQL: 10.10 instead of
postgresql 9.6 which is present in photon:2.0

Closes #12522

Signed-off-by: Jiří Suchomel <jiri.suchomel@suse.com>

### DIFF
--- a/make/photon/db/Dockerfile.base
+++ b/make/photon/db/Dockerfile.base
@@ -1,4 +1,4 @@
-FROM photon:2.0
+FROM photon:3.0
 
 ENV PGDATA /var/lib/postgresql/data
 


reasonerjt:
I don't think we wanna bump up to photon v3 right now.

jsuchome:
Thanks for comment, @reasonerjt . Could you please elaborate? Are there some specific concerns with newer version of Photon, or is it already on a roadmap for some later time?

PR #13254, opened 2020-10-13
Add pull time for proxy cache artifact in first pull

### LINKED ISSUE #13152: Newly pulled artifacts in a ProxyCache project have no Pull Time
When pulling an image from a ProxyCache project that doesn't exist in the cache, the image gets pulled from the upstream registry and served to the client, however the Pull Time for this image is not updated (Pull Time remains empty), if the image is not pulled again by another client it will be marked for deletion by the default retention rule since it meets the criteria "wasn't pulled during last 7 days".

I would expect to the Pull Time (and Pull Count ?) to be updated during the initial load of the artifact when the `proxy-cache-service` creates the artifact in the ProxyCache project.

1. Create a ProxyCache project (we used a Docker Hub registry endpoint)
2. pull only once an image that doesn't currently exist in the Project repos
3. make sure the image is pulled by the client and can be seen in the Project's repositories
4. notice that Pull Time has an empty value
5. run a DRY RUN for the project's default retention rule and check the log, this image should have a `DEL` retention result.

**Versions:**

- harbor version: v2.1.0
- docker engine version: 19.03.12
- docker-compose version: 1.27.2

**Additional context:**

N/A, happy to provide additional info as requested.

/cc @jonasrosland following our slack chat.

stonezdj:
Extract and move UpdatePullTime to util package
Fixes #13152

### DIFF
--- a/src/controller/proxy/controller.go
+++ b/src/controller/proxy/controller.go
@@ -142,7 +142,8 @@ func (c *controller) ProxyManifest(ctx context.Context, p *models.Project, art l
 			if err != nil {
 				log.Errorf("failed to get manifest, error %v", err)
 			}
-		} else {
+		}
+		if a != nil {
 			SendPullEvent(a, art.Tag, operator)
 		}
 	}(operator.FromContext(ctx))


src/controller/proxy/local.go
```diff
@@ -217,4 +218,5 @@ func SendPullEvent(a *artifact.Artifact, tag, operator string) {
 		Operator: operator,
 	}
 	event.BuildAndPublish(e)
+	util.UpdatePullTime(context.Background(), e.Artifact.ID, e.Tag)
```
ywk253100, line 221: The event handler will update the pull time automatically, why do you need to call the function to do the same work here?

stonezdj: Updated

reasonerjt (approved):

ywk253100 (approved):

fcrespofastly:
Hi @stonezdj! Thanks for working on this! 

Sorry I am not following how replacing the `else` by `if !=a` would fix this. What am I missing?

Issue #13244, opened 2020-10-12
helm hub adapter no longer works


reasonerjt:
https://helm.sh/blog/helm-hub-moving-to-artifact-hub/

As helm hub moves to artifact hub, the adapter needs rework to remain functioning.

We should consider to backport the fix to 2.1.x/2.0.x

mmpei:
artifact.hub provides apis for interaction which authorised by an api key from user profile.
`curl  GET 'https://artifacthub.io/api/v1/repositories/helm' --header 'X-API-KEY: XXXXXXXXXXXXXXXXX'
curl  GET 'https://artifacthub.io/api/v1/packages/search?facets=false&kind=0&verified_publisher=true&limit=50'`
api doc link https://artifacthub.github.io/hub/api/#/
we should parse from body of the repository for repo-url and fetch the index.yaml file. after that, we can get the target url path and then pull the package of the chart.

reasonerjt:
For v2.1.2 and v2.0.4 we'll remove the helm hub adapter.
Please open another issue to introduce "artifact hub adapter"

steven-zou:
@mmpei 

Please remove the helm hub adapter as it (helm hub) has not existed anymore.

steven-zou:
@AllForNothing 

This change may need some UI refactoring work.

AllForNothing:
![image](https://user-images.githubusercontent.com/30999793/97677538-603a8d80-1acd-11eb-8eaf-dc4701e274e1.png)
@xaleeks UI changes for existed Helm Hub registry

AllForNothing:
UI fixing for this issue:  PR #13410 to master(2.2), #13429 to 2.1  and  #13430 to 2.0

AllForNothing:
closing as all PRs merged

xaleeks:
@a-mccarthy  please call this out in the docs for the v2.0.4.  Please consult with @reasonerjt  if you have any questinoos

PR #13256, opened 2020-10-13
[Cherrypick]: Remove UT for helm hub adapter 

reasonerjt:
The helm hub adapter will not work due to the shift to artifact hub and
the API has changed. more details see #13244
Remove the UT for helm hub adapter to unblock CI.
Later we need to rework the adapter to replicate from artifact hub.


danfengliu (approved):

wy65701436 (approved):
lgtm

PR #13265, opened 2020-10-14
Extends the timeout in CI jobs

reasonerjt:
The jobs are taking more time to finish due to slowliness to pull docker
images from dockerhub (may due to rate limiting?)
This commit extends the timeout in some of the jobs to avoid CI failure.


reasonerjt:
Closing as the slowliness seems temporary.

PR #13267, opened 2020-10-14
Refine ldap searchGroup function 

### LINKED ISSUE #13156: ldap group filter
Hi,

until Harbor 2.0, in Harbor configuration, I never set the fields 'LDAP Group Base DN' and 'LDAP Group Filter' but I was still able to add and use LDAP groups in Harbor.
Using the same config in Harbor 2.1, I can't add a LDAP group and I have the following error in the logs:

Sep 24 13:48:08 172.18.0.1 core[1723]: 2020-09-24T18:48:08Z [ERROR] [/common/utils/ldap/ldap.go:302]: Wrong filter format, filter:()
Sep 24 13:48:08 172.18.0.1 core[1723]: 2020-09-24T18:48:08Z [WARNING] [/core/auth/ldap/ldap.go:208]: ldap search group fail: invalid filter syntax

I am able to login with a LDAP user so the user LDAP config seems OK

So is the LDAP group filter is now required or is that a bug?

Thanks,

Chris.

stonezdj:
Search LDAP group with groupDN+filter, then match baseDN
Create a default filter when ldap group filter is empty
Fixes #13156

### DIFF
--- a/src/common/utils/ldap/ldap.go
+++ b/src/common/utils/ldap/ldap.go
@@ -294,9 +294,9 @@ func (session *Session) SearchLdapAttribute(baseDN, filter string, attributes []
 	if err := session.Bind(session.ldapConfig.LdapSearchDn, session.ldapConfig.LdapSearchPassword); err != nil {
 		return nil, fmt.Errorf("Can not bind search dn, error: %v", err)
 	}
-	filter = strings.TrimSpace(filter)
-	if !(strings.HasPrefix(filter, "(") || strings.HasSuffix(filter, ")")) {
-		filter = "(" + filter + ")"
+	filter = normalizeFilter(filter)
+	if len(filter) == 0 {
+		return nil, ErrInvalidFilter
 	}
 	if _, err := goldap.CompileFilter(filter); err != nil {
 		log.Errorf("Wrong filter format, filter:%v", filter)
@@ -371,7 +371,11 @@ func (session *Session) SearchGroupByDN(groupDN string) ([]models.LdapGroup, err
 	if _, err := goldap.ParseDN(groupDN); err != nil {
 		return nil, ErrDNSyntax
 	}
-	groupList, err := session.searchGroup(groupDN, session.ldapGroupConfig.LdapGroupFilter, "", session.ldapGroupConfig.LdapGroupNameAttribute)
+	ldapFilter, err := createGroupSearchFilter(session.ldapGroupConfig.LdapGroupFilter, "", session.ldapGroupConfig.LdapGroupNameAttribute)
+	if err != nil {
+		return nil, err
+	}
+	groupList, err := session.searchGroup(groupDN, ldapFilter, "", session.ldapGroupConfig.LdapGroupNameAttribute)
 	if serverError, ok := err.(*goldap.Error); ok {
 		log.Debugf("resultCode:%v", serverError.ResultCode)
 	}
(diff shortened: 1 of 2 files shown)


src/common/utils/ldap/ldap.go
```diff
@@ -388,56 +392,93 @@ func (session *Session) groupBaseDN() string {
...
 }
 
-func createGroupSearchFilter(oldFilterStr, groupName, groupNameAttribute string) (string, error) {
-	origFilter, err := NewFilterBuilder(oldFilterStr)
+func groupName(entry *goldap.Entry, groupNameAttribute string) string {
+	for _, attr := range entry.Attributes {
+		// OpenLdap sometimes contain leading space in username
+		val := strings.TrimSpace(attr.Values[0])
+		log.Debugf("Current ldap entry attr name: %s\n", attr.Name)
+		switch strings.ToLower(attr.Name) {
```
reasonerjt, line 440: any reason not to use `if` ?

src/common/utils/ldap/ldap.go
```diff
@@ -388,56 +392,93 @@ func (session *Session) groupBaseDN() string {
...
+	}
+	child, err := goldap.ParseDN(childDN)
+	if err != nil {
+		return false, err
+	}
+	return base.AncestorOf(child) || base.Equal(child), nil
+}
+
+// createGroupSearchFilter - Create group search filter with base filter and group name filter condition
+func createGroupSearchFilter(baseFilter, groupName, groupNameAttr string) (string, error) {
```
reasonerjt, line 462: what if groupNameAttr is empty string?

src/common/utils/ldap/ldap.go
```diff
@@ -388,56 +392,93 @@ func (session *Session) groupBaseDN() string {
...
+	result, err := session.SearchLdapAttribute(groupDN, ldapFilter, []string{groupNameAttribute})
+	if err != nil {
+		return ldapGroups, err
+	}
+	if len(result.Entries) == 0 {
+		return ldapGroups, nil
+	}
+	group := models.LdapGroup{
+		GroupDN:   groupDN,
+		GroupName: groupName(result.Entries[0], session.ldapGroupConfig.LdapGroupNameAttribute),
```
reasonerjt, line 428: Let's double check the usage of the attribute array, if it's used for controlling what to be returned the function `groupName` is not needed.

src/common/utils/ldap/ldap.go
```diff
@@ -388,56 +392,93 @@ func (session *Session) groupBaseDN() string {
...
 	}
 	groupName = goldap.EscapeFilter(groupName)
 	gFilterStr := ""
-	if len(groupName) > 0 {
-		gFilterStr = fmt.Sprintf("(%v=%v)", goldap.EscapeFilter(groupNameAttribute), groupName)
+	// when groupName is empty, search all groups in current base DN
+	if len(groupName) == 0 {
+		groupName = "*"
 	}
+	gFilterStr = fmt.Sprintf("(%v=%v)", goldap.EscapeFilter(groupNameAttr), groupName)
```
reasonerjt, line 474: gFilterStr := xxxxx

src/common/utils/ldap/ldap.go
stonezdj: default to `cn`

src/common/utils/ldap/ldap.go
stonezdj: removed this function

src/common/utils/ldap/ldap.go
stonezdj: updated

src/common/utils/ldap/ldap.go
stonezdj: updated

reasonerjt (approved):

ywk253100 (approved):

PR #13270, opened 2020-10-14
Enable authentication for internal redis

stefannica:
Enable password authentication for the internal redis service.

Signed-off-by: Stefan Nica <snica@suse.com>

### DIFF
--- a/make/harbor.yml.tmpl
+++ b/make/harbor.yml.tmpl
@@ -43,6 +43,11 @@ database:
   # Note: the default number of connections is 1024 for postgres of harbor.
   max_open_conns: 1000
 
+# Harbor redis configuration
+redis:
+  # The password for the default user of Harbor redis. Change this before any production use.
+  password: redis123
+
 # The default data volume
 data_volume: /data
 
--- a/make/photon/prepare/templates/docker_compose/docker-compose.yml.jinja
+++ b/make/photon/prepare/templates/docker_compose/docker-compose.yml.jinja
@@ -324,6 +324,8 @@ services:
           - redis
   {% endif %}
     dns_search: .
+    env_file:
+      - ./common/config/redis/env
     depends_on:
       - log
     logging:
--- a/make/photon/prepare/templates/redis/env.jinja
+++ b/make/photon/prepare/templates/redis/env.jinja
@@ -1,0 +1,1 @@
+REDIS_PASSWORD={{harbor_redis_password}}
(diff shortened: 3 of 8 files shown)


mfriess2:
Will this PR come into Harbor any time, soon?
Because Harbor's unsecured redis server is reported by security tools.

Vad1mo:
hey @goharbor/maintainers, i think that would still make sense!

MinerYang:
> Will this PR come into Harbor any time soon? Because Harbor's unsecured redis server is reported by security tools.

We would like to know what these security tools are and what the security concern it is by not setting the internal redis password.
Based on our knowledge, this internal redis can not access externally. If you have permission to access the internal redis, it basically means you can get all the harbor configurations. Consider to this situation, if you intend to do something bad simply adding a password wouldn't be helping more.
And it only stores cache data like metadata from distribution, nothing sensitive as far as I know.

mfriess2:
The problem was reported by a Nessus Security Scanner (from Tenable) as a "critical" finding!
See details here: https://www.tenable.com/plugins/nessus/100634

The attack scenario can be:
* You have some container running in your cluster which has a security vulnerability. Let's assume it's some Ubuntu container.
* Someone achieves to exploit this vulnerability and gets shell access into that container in your cluster
* Now the attacker is in your cluster's network (note: **without** having kubectl control !!!)
* He can now scan the cluster for open ports and connect to your redis container without authorization (e.g. by just using "curl" in the exploited container)

Example:
image

Is it that much effort to put a password on redis and use the password in the harbor communication?

qnetter:
I agree that this is a significant issue. So much of the cloud-native universe has been designed with the assumption of a friendly world with properly configured network boundaries - a naïve assumption.

MinerYang:
True, thanks for your follow-up elaboration! 
We will have a discuss on this and evaluate the risk with my colleague soon.

MinerYang:
We have a discussion with security team. And they agree with the risk described above by @mfriess2. We probably consider to set this password configuration for internal Redis server as optional. Since current default setting of it is using a unauthenticated way (without password) and if we plan to change the default configuration, there would be an issue with upgrade/migration. So we need take care of this more specifically. 
And welcome to have more discussion and feedback from the community!

MinerYang:
we will  evaluate change behavior for downstream consumer like harbor-helm etc. And plan this as a feature for next release cycle (2.9.0)

PR #13275, opened 2020-10-15
Update the LDAP utils to support LDAP paging.

RageZBla:
#10940 & #11060 introduced a change where harbor would search for the name of LDAP group to populate user's group. In a situation where there is thousand of groups in LDAP the server would return result code 4 size limit exceed. This use paging to remove this edge case.


RageZBla:
Reference #13156 13156

Issue #12261, opened 2020-06-16
Core logs error "failed to verify token: oidc: malformed jwt: square/go-jose: compact JWS format must have three parts" after switch to S3 storage -- excessive warning messages.


msiebuhr:
**Expected behavior and actual behavior:**

On a Harbor installation (Kubernetes using harbor-helm), I'm seeing the following error for all requests to the core server:

```
2020-06-16T11:32:11Z [WARNING] [/server/middleware/security/idtoken.go:45][requestID="a4f074b4-2cd2-4976-a68c-5f378fbc4882"]: failed to verify token: oidc: malformed jwt: square/go-jose: compact JWS format must have three parts
2020/06/16 11:32:11.281 [D] [transaction.go:62]  |    10.217.5.96| 200 |    502.279µs|   match| GET      /api/v2.0/ping   r:/api/v2.0/ping
2020-06-16T11:32:14Z [WARNING] [/server/middleware/security/idtoken.go:45][requestID="29e92515-252d-4218-b16c-7530880cb5a6"]: failed to verify token: oidc: malformed jwt: square/go-jose: compact JWS format must have three parts
2020/06/16 11:32:14.062 [D] [transaction.go:62]  |    10.217.5.96| 200 |    294.838µs|   match| GET      /api/v2.0/ping   r:/api/v2.0/ping
```

If I switch the installation back to using PersistentVolumeClaims, the errors go away (and leave just the regular ping's).

**Steps to reproduce the problem:**

Seem to be related to using a S3 backend, likely in combination with us using oAuth for Single Sign-On.

**Versions:**
Please specify the versions of following systems.

- harbor version: 2.0.0
- harbor-helm 1.4.0
- kubernetes: 1.17.3
- oAuth server is dex (https://github.com/dexidp/dex)

**Additional context:**

(I was unable to locate any meaningful configuration file on the core server pod. Relevant logs reproduced above.)

reasonerjt:
Please elaborate on how you reproduce it, I don't think it's related to S3 or local disk.

It seems b/c you are using OIDC auth mode but the request is not carrying valid ID token.

msiebuhr:
It is deployed to Kubernetes using the neighboring *harbor-helm* chart. 

> lease elaborate on how you reproduce it, did you use a curl command to call Harbor's API?

The /ping requests are done by Kubernete's health-check. As Kubernetes does not restart the server, I guess it's returning something OK.

> I don't think it's related to S3 or local disk.

[...]

PR #13286, opened 2020-10-16
add whitelisting in idtoken for ping endpoint

sharkymcdongles:
Solves log spam in this issue:

https://github.com/goharbor/harbor/issues/12261

### DIFF
--- a/src/server/middleware/security/idtoken.go
+++ b/src/server/middleware/security/idtoken.go
@@ -40,6 +40,9 @@ func (i *idToken) Generate(req *http.Request) security.Context {
 	if !strings.HasPrefix(req.URL.Path, "/api") {
 		return nil
 	}
+	if strings.Contains(req.URL.Path, "ping") {
+		return nil
+        }
 	claims, err := oidc.VerifyToken(req.Context(), bearerToken(req))
 	if err != nil {
 		log.Warningf("failed to verify token: %v", err)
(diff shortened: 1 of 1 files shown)


src/server/middleware/security/idtoken.go
```diff
@@ -40,6 +40,9 @@ func (i *idToken) Generate(req *http.Request) security.Context {
 	if !strings.HasPrefix(req.URL.Path, "/api") {
 		return nil
 	}
+	if strings.Contains(req.URL.Path, "ping") {
+		return nil
+        }
```
msiebuhr, line 45: Won't this permit all requests with `ping` in the name to bypass OIDC auth? Say, I create an account/namespace/... called `ping` - then authentication's off.

(I know Harbor mostly uses numerical IDs in URLs, but that might change later on.)

```suggestion
      # Don't complain on healthchecks without OIDC login
  	if req.URL.path == "/api/v2.0/ping") {
		return nil
        }
```

sharkymcdongles: That's also fine. I added the ping contains in case the path ever changed, but I suppose this change can be updated then too.

heww:
Hi @sharkymcdongles Can you look at this PR https://github.com/goharbor/harbor/pull/13422 which will skip security middleware for ping endpoint?

Issue #13200, opened 2020-09-30
Quay.io pull fuzzy name not working


funzie19:
**Expected behavior and actual behavior:**
`company/**` sync up all repositories from Quay.io into Harbor

**Steps to reproduce the problem:**
- Setup a registry to Quay.io
- Setup a replication rule.
-- Use registry create before
-- Set as pull
-- Set name: `company/**`
-- Set destination
- Run replication
-- Replication will fail with:
```failed to fetch artifacts: http status code: 400, body: {"errors":[{"code":"INVALID_REQUEST","detail":{},"message":"Unable to decode repository and actions: registry:catalog:*"}]}```

**Versions:**
Please specify the versions of following systems.

- harbor version: 2.1.0

**Additional context:**
- Harbor is running in a GKE cluster
- Pulling `company/name` works as expected.

mfld:
good suggestion
+1

steven-zou:
@chlins 
Please take a look at this issue. Thanks!

chlins:
@funzie19 Can you try to configure the pattern of company/{repo1, repo2} and check if it can work well ?

mfld:
hello, just wanted to voice my concern.

Even though "company/{repo1,repo2}"  might work, it will be hard to maintain for companies with lots of repositories listed on quay. Please add support for fuzzy names.

chlins:
@mfld The pattern of "company/{repo1,repo2}" just help to find out the problem. Not mean we do not fix the fuzzy name like  company/**.  Our ultimate goal must be fuzzy pattern can work well(prerequisites:  with auth mode).

xaleeks:
@a-mccarthy please note this change in the docs

a-mccarthy:
@xaleeks i dont think this needs more docs than what is already on the site. let me know if i'm misunderstanding things

PR #13305, opened 2020-10-19
fix: fix quay adapter catalog api return error

chlins:
Close: https://github.com/goharbor/harbor/issues/13200

### DIFF
--- a/src/replication/adapter/quay/adapter.go
+++ b/src/replication/adapter/quay/adapter.go
@@ -1,3 +1,17 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


bitsf:
hi @chlins can you also upload a test result for case
https://github.com/goharbor/replication-verification/blob/master/master/cases/Quay.robot
you can change the `bitsf/{photon,alpine}` to `bitsf/{photon,alpine*}` in order to test the catalog api

chlins:
@bitsf Maybe we can also update the replication verification quay case to cover the scene. How do you think?

Before fix, if use fuzzy pattern which will call catlog cause below error:
图片
图片

After fix:
图片
图片

steven-zou:
> @bitsf Maybe we can also update the replication verification quay case to cover the scene. How do you think?
> 
> Before fix, if use fuzzy pattern which will call catlog cause below error:
> 图片
> 图片
> 
> After fix:
> 图片
> 图片

@bitsf I think it's a good idea

cc @chlins

steven-zou (approved):
LGTM

CC @bitsf

bitsf:
@chlins yes, I will fix the case

bitsf (approved):

steven-zou:
@chlins 

This fix should be cherry-picked to the 2.1 branch

PR #13306, opened 2020-10-19
refactor(security): use controller instead of promgr in security

heww:
Use `project.Controller` instead of `promgr.ProjectManager` in security
implementations because we will remove `promgr` package later.


wy65701436 (approved):
lgtm

src/pkg/project/dao/dao.go
```diff
@@ -39,6 +39,8 @@ type DAO interface {
 	GetByName(ctx context.Context, name string) (*models.Project, error)
 	// List list projects
 	List(ctx context.Context, query *q.Query) ([]*models.Project, error)
+	// Lists the roles of user for the specific project
+	ListRoles(ctx context.Context, projectID int64, userID int, groupIDs ...int) ([]int, error)
```
reasonerjt, line 43: It's not very clear the relationship between uid and gid?  it seems the uid MUST be a member of each GID provided?  If that's the case, would it be better if we combine the data in one data object. 

I recall models.User has the values but a lot of other irrelevant attributes.

src/pkg/project/manager.go
```diff
@@ -46,6 +46,9 @@ type Manager interface {
 
 	// List projects according to the query
 	List(ctx context.Context, query *q.Query) ([]*models.Project, error)
+
+	// ListRoles returns the roles of user for the specific project
+	ListRoles(ctx context.Context, projectID int64, userID int, groupIDs ...int) ([]int, error)
```
reasonerjt, line 51: See my comments in project/dao pkg...

src/pkg/project/dao/dao.go
heww: It's not very well to import `pkg/user/models` here. 

We are in agreement that one package in `pkg` doesn't import other packages in the `pkg` directory.

reasonerjt (approved):

reasonerjt:
I also find the way to organize project evaulator, user and builder a little confusing.  There may be a more straight forward way but it does not block merging this PR.

PR #13312, opened 2020-10-20
Add admin group support to OIDC auth mode

reasonerjt:
Add oidc_admin_group to configuration, and make sure a token with the
group name in group claim has the admin authority.

### DIFF
--- a/src/common/utils/oidc/helper.go
+++ b/src/common/utils/oidc/helper.go
@@ -126,7 +126,7 @@ type UserInfo struct {
 	Username         string   `json:"name"`
 	Email            string   `json:"email"`
 	Groups           []string `json:"groups"`
-	AdminGroupMember bool     `json:"AdminGroupMember"`
+	AdminGroupMember bool     `json:"admin_group_member"`
 	hasGroupClaim    bool
 }
 
@@ -382,8 +382,12 @@ func populateGroupsDB(groupNames []string) ([]int, error) {
 }
 
 // InjectGroupsToUser populates the group to DB and inject the group IDs to user model.
-// The third optional parm is for UT only, when using the func, the third
+// The third optional parm is for UT only.
 func InjectGroupsToUser(info *UserInfo, user *models.User, f ...populate) {
+	if info == nil || user == nil {
+		log.Warningf("user info or user model is nil, skip the func")
+		return
+	}
 	var populateGroups populate
 	if len(f) == 0 {
 		populateGroups = populateGroupsDB
--- a/src/common/utils/oidc/helper.go
+++ b/src/common/utils/oidc/helper.go
@@ -26,6 +26,8 @@ import (
 	"time"
 
 	gooidc "github.com/coreos/go-oidc"
+	"github.com/goharbor/harbor/src/common"
+	"github.com/goharbor/harbor/src/common/dao/group"
 	"github.com/goharbor/harbor/src/common/models"
 	"github.com/goharbor/harbor/src/core/config"
 	"github.com/goharbor/harbor/src/lib/log"
@@ -119,12 +121,13 @@ type Token struct {
 // UserInfo wraps the information that is extracted via token.  It will be transformed to data object that is persisted
 // in the DB
 type UserInfo struct {
-	Issuer        string   `json:"iss"`
-	Subject       string   `json:"sub"`
-	Username      string   `json:"name"`
-	Email         string   `json:"email"`
-	Groups        []string `json:"groups"`
-	hasGroupClaim bool
+	Issuer           string   `json:"iss"`
+	Subject          string   `json:"sub"`
+	Username         string   `json:"name"`
+	Email            string   `json:"email"`
+	Groups           []string `json:"groups"`
+	AdminGroupMember bool     `json:"AdminGroupMember"`
+	hasGroupClaim    bool
 }
 
 func getOauthConf() (*oauth2.Config, error) {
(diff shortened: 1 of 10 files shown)


src/common/utils/oidc/helper.go
```diff
@@ -119,12 +121,13 @@ type Token struct {
...
-	Username      string   `json:"name"`
-	Email         string   `json:"email"`
-	Groups        []string `json:"groups"`
-	hasGroupClaim bool
+	Issuer           string   `json:"iss"`
+	Subject          string   `json:"sub"`
+	Username         string   `json:"name"`
+	Email            string   `json:"email"`
+	Groups           []string `json:"groups"`
+	AdminGroupMember bool     `json:"AdminGroupMember"`
```
stonezdj, line 129: The AdminGroupMember should not put in the UserInfo, because when AdminGroup setting is changed, then previous admin user should lost the admin privilege at once, when the AdminGroupMember is persitent in token and saved in database, the user still have the admin privilege when the token is valid. because the groups is already contains the group information, we can use info to check it has admin role.

src/common/utils/oidc/helper.go
```diff
@@ -280,9 +283,11 @@ func mergeUserInfo(remote, local *UserInfo) *UserInfo {
 	}
```
stonezdj, line 283: Suggest to put the code these code logic  
https://github.com/reasonerjt/harbor/blob/f4ff369ed0f0d8118100ae54fbc6dc667588ec1e/src/common/utils/oidc/helper.go#L261-L272
into mergeUserInfo function

src/common/utils/oidc/helper.go
reasonerjt: Thanks for the review, but user info is tied to the id token, so each time Harbor gets a new id token it will extracts the admin group from setting and compare with token and populate to user info, then to user model.

It's like removing admin role from a user, this will take effect after user re-authenticate.

Moving the code to `mergeUserInfo` makes no improvement.

reasonerjt: And for clarification the adminGroup flag is not in the token or persisted. 
Each time the token is verified the flag will be regenerated based on the OIDC settings in Harbor.

src/common/utils/oidc/helper.go
reasonerjt: See my comment in the thread:
https://github.com/goharbor/harbor/pull/13312#discussion_r509008990

This is not necessary.

stonezdj (approved):

src/common/utils/oidc/helper.go
wy65701436, line 129: json:"AdminGroupMember" => json:"adminGroupMember" ?

src/server/middleware/security/idtoken.go
```diff
@@ -54,19 +54,16 @@ func (i *idToken) Generate(req *http.Request) security.Context {
...
-	if groupNames, ok := oidc.GroupsFromClaims(claims, settings.GroupsClaim); ok {
-		groups := models.UserGroupsFromName(groupNames, common.OIDCGroupType)
-		u.GroupIDs, err = group.PopulateGroup(groups)
-		if err != nil {
-			log.Errorf("failed to get group ID list for OIDC user %s: %v", u.Username, err)
-			return nil
-		}
+	info, err := oidc.UserInfoFromIDToken(ctx, &oidc.Token{RawIDToken: token}, *setting)
+	if err != nil {
+		log.Errorf("Failed to get user info from ID token: %v", err)
```
wy65701436, line 64: no return?

src/common/utils/oidc/helper.go
```diff
@@ -312,34 +318,41 @@ func userInfoFromIDToken(ctx context.Context, token *Token, setting models.OIDCS
 		return nil, err
 	}
 
-	return userInfoFromClaims(idt, setting.GroupsClaim, setting.UserClaim)
+	return userInfoFromClaims(idt, setting)
 }
 
-func userInfoFromClaims(c claimsProvider, g, u string) (*UserInfo, error) {
+func userInfoFromClaims(c claimsProvider, setting models.OIDCSetting) (*UserInfo, error) {
```
wy65701436, line 324: it's better to use a verb prefix in the name for this function.

src/common/utils/oidc/helper.go
```diff
@@ -362,6 +375,29 @@ func GroupsFromClaims(gp claimsProvider, k string) ([]string, bool) {
...
 }
 
+type populate func(groupNames []string) ([]int, error)
+
+func populateGroupsDB(groupNames []string) ([]int, error) {
+	return group.PopulateGroup(models.UserGroupsFromName(groupNames, common.OIDCGroupType))
+}
+
+// InjectGroupsToUser populates the group to DB and inject the group IDs to user model.
+// The third optional parm is for UT only, when using the func, the third
```
wy65701436, line 385: the third what? no need the specify the thire para besides UT?

src/common/utils/oidc/helper.go
```diff
@@ -362,6 +375,29 @@ func GroupsFromClaims(gp claimsProvider, k string) ([]string, bool) {
...
+func InjectGroupsToUser(info *UserInfo, user *models.User, f ...populate) {
```
wy65701436, line 386: is it necessary to add the nil checking for info and user?

src/server/middleware/security/idtoken.go
reasonerjt: thanks!

src/common/utils/oidc/helper.go
reasonerjt: ok..

src/common/utils/oidc/helper.go
reasonerjt: it's incomplete editing of comment...  I'll correct it.

src/common/utils/oidc/helper.go
reasonerjt: Yep, it's not likely to happen but I'll add the check thanks.

wy65701436 (approved):
lgtm

PR #13349, opened 2020-10-23
Don't fail when an error occurs during garbage collection runs

moshloop:
We periodically get errors during GC runs, the errors in of itself are the primary concern (some errors are to be expected when networks are involved) - The problem is that it neither retries, not continues to the next blob for garbage collection

### DIFF
--- a/src/jobservice/job/impl/gc/garbage_collection.go
+++ b/src/jobservice/job/impl/gc/garbage_collection.go
@@ -252,6 +252,8 @@ func (gc *GarbageCollector) sweep(ctx job.Context) error {
 			continue
 		}
 
+		errorCount := 0
+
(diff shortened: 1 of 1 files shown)


xaleeks:
@wy65701436  please review this

wy65701436:
If we'd like to introduce partial success of GC job, I think several things that need to be considered:

1,  what's the result status of the GC job? success or fail?
2,  the log should be friendly to the end user to show how many successs, fail or failure rate, and the root cause of each failure.
3, as harbor uses blob status to handle the concurence case, so any failure on the DB status update, it must be return.
4, the blob/manifest deletion failure maybe can passed away, but we have to reserve the DB record for it, then GC can choose it in the next execution.

PR #13411, opened 2020-11-02
Remove UT for helm hub adapter (#13255)

heww:
The helm hub adapter will not work due to the shift to artifact hub and
the API has changed. more details see #13244
Remove the UT for helm hub adapter to unblock CI.
Later we need to rework the adapter to replicate from artifact hub.


danfengliu (approved):
LGTM

bitsf (approved):

Issue #13155, opened 2020-09-24
client timeout calling core's api/v2.0/ping


xtreme-conor-nosal:
We're trying to upgrade from Harbor 2.0.1 to Harbor 2.1.0. One instance is healthy, but a second harbor is failing to respond to `api/v2.0/ping`.

In the healthy Harbor each ping generates 3 lines in the core debug logs:
```
2020-09-24T18:22:46Z [DEBUG] [/server/middleware/log/log.go:30]: attach request id fa91b1da-936a-4f72-9f4e-b30f7ced820a to the logger for the request GET /api/v2.0/ping
2020-09-24T18:22:46Z [DEBUG] [/server/middleware/artifactinfo/artifact_info.go:52]: In artifact info middleware, url: /api/v2.0/ping
2020-09-24T18:22:46Z [DEBUG] [/server/middleware/security/unauthorized.go:29][requestID="fa91b1da-936a-4f72-9f4e-b30f7ced820a"]: an unauthorized security context generated for request GET /api/v2.0/ping
```

In the unhealthy Harbor only the first line is logged for each ping
`2020-09-24T18:22:39Z [DEBUG] [/server/middleware/log/log.go:30]: attach request id 9cfc1ecc-9ebb-4d26-9196-6a678c454181 to the logger for the request GET /api/v2.0/ping`

**Versions:**
Please specify the versions of following systems.

- harbor version: [2.1.0]
- helm [1.5.0]

**Additional context:**
Looking at https://github.com/goharbor/harbor/blob/v2.1.0/src/core/middlewares/middlewares.go it appears that the first log entry is from the log middleware, and the second log entry from the artifact middleware. This suggests that in the unhealthy Harbor one of the middlewares in between might be hanging or consuming the request without writing a log?

Both harbors use the same base configuration with external DB and Redis, and deployed to the same GKE cluster. One has clair disabled and the healthy pod has an extra custom pod.

xtreme-conor-nosal:
core startup logs
```
Appending internal tls trust CA to ca-bundle ...
find: /etc/harbor/ssl: No such file or directory
Internal tls trust CA appending is Done.
2020-09-24T19:24:00Z [INFO] [/controller/artifact/annotation/parser.go:71]: the annotation parser to parser artifact annotation version v1alpha1 registered
2020-09-24T19:24:00Z [INFO] [/controller/artifact/processor/processor.go:59]: the processor to process media type application/vnd.cncf.helm.config.v1+json registered
[...]

PR #13413, opened 2020-11-02
Move security middleware ahead of transaction middleware

reasonerjt:
In v2.1 security middleware will query DB by creating new connection.
If it is put after transaction middleware there's a bigger chance of
deadlock if the concurrent open connections are set too low (#13155)

This commit mitigates that issue.  But we still need work to lower the
connections and better handle the case when http connection is closed.


reasonerjt:
Per discussion, this is not needed >=v2.2

PR #13421, opened 2020-11-03
Remove clair

ninjadq:


make/install.sh
```diff
@@ -30,8 +27,6 @@ while [ $# -gt 0 ]; do
             exit 0;;
             --with-notary)
             with_notary=true;;
-            --with-clair)
-            with_clair=true;;
```
reasonerjt, line 34: As was suggested by Alex, we should print a explicity warning to user letting him know that clair is removed and `--with-clair` is no longer accepted.

heww:
The clair database creation in harbor-db also need to be removed.

reasonerjt (approved):

heww (approved):

PR #13444, opened 2020-11-04
add robot mgr

wy65701436:
the robot account manager to handle the CRUD


src/pkg/robot2/dao/dao.go
```diff
@@ -0,0 +1,153 @@
...
+	if n == 0 {
+		return errors.NotFoundError(nil).WithMessage("robot account %d not found", id)
+	}
+	return nil
+}
+
+func (d *dao) List(ctx context.Context, query *q.Query) ([]*model.Robot, error) {
+	robots := []*model.Robot{}
+	query = q.MustClone(query)
+	query.Keywords["deleted"] = false
```
heww, line 127: `query.Keywords["deleted"] = false` => `query.Keywords["visible"] = true`

src/pkg/robot2/dao/dao.go
```diff
@@ -0,0 +1,153 @@
...
+	if n == 0 {
+		return errors.NotFoundError(nil).WithMessage("robot %d not found", r.ID)
+	}
+	return nil
+}
+
+func (d *dao) Get(ctx context.Context, id int64) (*model.Robot, error) {
+	r := &model.Robot{
+		ID:       id,
+		Disabled: false,
```
heww, line 78: Do we need `Disabled: false`?

src/pkg/robot2/dao/dao.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return 0, err
+	}
+	r.CreationTime = time.Now()
+	id, err := ormer.Insert(r)
+	if err != nil {
+		if e := orm.AsConflictError(err, "robot account %d:%s already exists",
+			r.ProjectID, r.Name); e != nil {
+			err = e
+		}
```
heww, line 54:
``` go
if err != nil {
    return 0, orm.WrapConflictError(err, "robot account %d:%s already exists", r.ProjectID, r.Name)
}
```

src/pkg/robot2/dao/dao.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	ormer, err := orm.FromContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if err := ormer.Read(r); err != nil {
+		if e := orm.AsNotFoundError(err, "robot %d not found", id); e != nil {
+			err = e
+		}
+		return nil, err
```
heww, line 88:
```go
if err := ormer.Read(r); err != nil {
    return nil, orm.WrapNotFoundError(err, "robot %d not found", id)
}
```

src/pkg/robot2/dao/dao.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	return r, nil
+}
+
+func (d *dao) Count(ctx context.Context, query *q.Query) (int64, error) {
+	query = q.MustClone(query)
+	query.Keywords["deleted"] = false
```
heww, line 95: "deleted" => "visible"

src/pkg/robot2/dao/dao.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	n, err := ormer.Delete(&model.Robot{
+		ID: id,
+	})
+	if err != nil {
+		return err
+	}
+	if n == 0 {
+		return errors.NotFoundError(nil).WithMessage("robot account %d not found", id)
+	}
```
heww, line 120: Not needed here because we can consider deletion is successful when the robot is not found in db.

wy65701436: discussed offline, keep it.

src/pkg/robot2/dao/dao.go
wy65701436: it should be deleted

wy65701436: no need, we should put it into the ctl

src/pkg/robot2/dao/dao.go
wy65701436: move it into ctl

src/pkg/robot2/dao/dao.go
wy65701436: move it into ctl

src/pkg/robot2/dao/dao.go
```diff
@@ -0,0 +1,144 @@
...
+	return nil
+}
+
+func (d *dao) List(ctx context.Context, query *q.Query) ([]*model.Robot, error) {
+	robots := []*model.Robot{}
+	query = q.MustClone(query)
```
heww, line 118: Not needed here as orm.QuerySetter can deal with query which is nil.

src/pkg/robot2/dao/dao.go
```diff
@@ -0,0 +1,144 @@
...
+			"project_id": projectID,
+		},
+	})
+	if err != nil {
+		return err
+	}
+	_, err = qs.Delete()
+	if err != nil {
+		return err
+	}
```
heww, line 142: These lines can be removed.

src/pkg/robot2/dao/dao.go
```diff
@@ -0,0 +1,144 @@
...
+	return id, err
+}
+
+func (d *dao) Update(ctx context.Context, r *model.Robot, props ...string) error {
+	ormer, err := orm.FromContext(ctx)
+	if err != nil {
+		return err
+	}
+	r.UpdateTime = time.Now()
+	n, err := ormer.Update(r, props...)
```
heww, line 62: `UpdateTime` needs to append to `props` when `props` is not empty otherwise `UpdateTime` will not be updated when props provided.

heww (approved):

src/pkg/robot2/dao/dao.go
wy65701436: remove it, and let the caller to set it

stonezdj (approved):

Issue #13074, opened 2020-09-16
Failed to add more webhook by API with name absent


danfengliu:
As you can see, no name in API paylod, but it will be success in the 1st record creation, after that there will be failure for posting a webhook without name in payload, please reference error log below. In UI operation, webhook name is an essential option, is webhook name an essential option by design?

**API with Payload:**
/api/v2.0/projects/$project_id/webhook/policies

{'targets': [{'type': 'http', 'address': 'https://2.2.2.2', 'skip_cert_verify': False, 'auth_header': 'bbb'}], 'event_types': ['DELETE_ARTIFACT', 'PULL_ARTIFACT', 'PUSH_ARTIFACT', 'DELETE_CHART', 'DOWNLOAD_CHART', 'UPLOAD_CHART', 'QUOTA_EXCEED', 'QUOTA_WARNING', 'REPLICATION', 'SCANNING_FAILED', 'SCANNING_COMPLETED'], 'enabled': False}

**Error Log:**
```
Sep 15 11:51:06 172.23.0.1 proxy[5980]: 10.158.214.76 - "POST /api/v2.0/projects/35/webhook/policies HTTP/1.1" 500 66 "-" "python-requests/2.18.4" 0.011 0.012 .
Sep 15 11:51:06 172.23.0.1 core[5980]: 2020-09-15T11:51:06Z [ERROR] [/server/error/error.go:53]: {"errors":[{"code":"INTERNAL_SERVER_ERROR","message":"failed to create the notification policy: pq: duplicate key value violates unique constraint \"notification_policy_name_key\""}]}
```

**Harbor Version:**
V2.0

mmpei:
policy name is unique in global. it is an identity of notification policy for user. so this action is in our design. and actully this is an optimization in v2.0.

reasonerjt:
@mmpei can you cherrypick the change in 2.1 to 2.0.x?

PR #13462, opened 2020-11-06
Fix for duplicate webhook policy name (#12729 #13074)

tedgxt:


make/migrations/postgresql/0031_2.0.3_schema.up.sql
```diff
@@ -49,4 +49,9 @@ UPDATE artifact AS art
...
    repository_id=art.repository_id
    FROM artifact as art
-   WHERE tag.artifact_id=art.id AND tag.repository_id!=art.repository_id;
\ No newline at end of file
+   WHERE tag.artifact_id=art.id AND tag.repository_id!=art.repository_id;
+
+/*remove the constraint for name in table 'notification_policy'*/
+ALTER TABLE notification_policy DROP CONSTRAINT notification_policy_name_key;
+/*add union unique constraint for name and project_id in table 'notification_policy'*/
+ALTER TABLE notification_policy ADD UNIQUE(name,project_id);
```
wy65701436, line 57: we should consider about the migration path: v2.0.0 - v2.0.4 - v2.1.0, these two lines will be executed twice.

The first line is idempotent, but the second line is not. It will generate two indexes like:


```
Indexes:
    “notification_policy_pkey” PRIMARY KEY, btree (id)
    “notification_policy_name_project_id_key” UNIQUE CONSTRAINT, btree (name, project_id)
    “notification_policy_name_project_id_key1” UNIQUE CONSTRAINT, btree (name, project_id)
```

reasonerjt:
Closing as we decided it's a won't fix in v2.0.x

PR #13472, opened 2020-11-09
add robot account version 2 controller

wy65701436:
the controller is for the enhanced robot account

### DIFF
--- a/src/controller/robot/controller.go
+++ b/src/controller/robot/controller.go
@@ -1,0 +1,343 @@
+package robot
+
+import (
+	"context"
+	"fmt"
+	rbac_common "github.com/goharbor/harbor/src/common/rbac"
+	"github.com/goharbor/harbor/src/common/utils"
+	"github.com/goharbor/harbor/src/core/config"
+	"github.com/goharbor/harbor/src/lib/errors"
+	"github.com/goharbor/harbor/src/lib/log"
+	"github.com/goharbor/harbor/src/lib/q"
+	"github.com/goharbor/harbor/src/pkg/permission/types"
+	"github.com/goharbor/harbor/src/pkg/project"
+	"github.com/goharbor/harbor/src/pkg/rbac"
+	rbac_model "github.com/goharbor/harbor/src/pkg/rbac/model"
+	robot "github.com/goharbor/harbor/src/pkg/robot2"
+	"github.com/goharbor/harbor/src/pkg/robot2/model"
+	"time"
+)
+
+var (
+	// Ctl is a global variable for the default robot account controller implementation
+	Ctl = NewController()
+)
+
+// Controller to handle the requests related with robot account
+type Controller interface {
+	// Get ...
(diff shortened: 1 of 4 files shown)


src/controller/robot/controller.go
```diff
@@ -0,0 +1,316 @@
...
+	"github.com/goharbor/harbor/src/pkg/rbac"
+	rbac_model "github.com/goharbor/harbor/src/pkg/rbac/model"
+	robot "github.com/goharbor/harbor/src/pkg/robot2"
+	"github.com/goharbor/harbor/src/pkg/robot2/model"
+	"time"
+)
+
+var (
+	// Ctr is a global variable for the default robot account controller implementation
+	Ctr = NewController()
```
heww, line 23: The default global controller name is usually `Ctl` here.

src/controller/robot/controller.go
```diff
@@ -0,0 +1,316 @@
...
+)
+
+// Controller to handle the requests related with robot account
+type Controller interface {
+	// Get ...
+	Get(ctx context.Context, id int64, option *Option) (*Robot, error)
+
+	// Create ...
+	Create(ctx context.Context, r *Robot) (*Robot, error)
```
heww, line 32: Can we only return the id of the new created robot account? When the API handler want to know the secret of the created robot account, it can call the `Get` method of the controller.

heww, line 32: Can we only return the id of the new created robot account? When the API handler wants to know the secret of the created robot account, it can call the `Get` method of the controller.

wy65701436: API will use the robot attribute to wrap a object and add it into the response body.

wy65701436: API will use the robot attribute to wrap a object and add it into the response body.

heww (approved):

src/controller/robot/controller.go
```diff
@@ -0,0 +1,343 @@
...
+			})
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func (d *controller) populate(ctx context.Context, r *model.Robot, option *Option) (*Robot, error) {
```
stonezdj, line 200: Fetch or attach is better than populate, on auth mod, the populate often involves data migrate, such as populateGroup, populateUser, current case only retrieves the information.

src/controller/robot/controller.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+	return robotID, nil
+}
+
+// Delete ...
+func (d *controller) Delete(ctx context.Context, id int64) error {
+	if err := d.robotMgr.Delete(ctx, id); err != nil {
+		return err
+	}
+	if err := d.rbacMgr.DeletePermissionsByRole(ctx, ROBOTTYPE, id); err != nil {
```
stonezdj, line 120: DeletePermissionsByRole is not a proper function name, it seems it need to take the role or role_id as the parameter, deletePermissions is OK

stonezdj (approved):

PR #13476, opened 2020-11-09
fix: ensure the role_id of role is correct

### LINKED ISSUE #13317: Unable to delete member of Project with role Limited Guest
I am not able to delete member with role LimitedGuest from any Project.
The Delete from UI or API only produces error:

```
{
  "errors": [
    {
      "code": "NOT_FOUND",
      "message": "project member 270 not found in project 1"
    }
  ]
}
```

This issue appears on instance that was initial deployed as version 1.7.4 and updated to new releases as they become available. I can give exact update path if required.
I was able to identify some discrepancy in Harbor database, see below in Additional context.

1. As user with required permissions, add new member to project with role LimitedGuest.
2. Try to delete the member

**Versions:**

- harbor version: 2.0.3
- docker engine version: 19.03.12
- docker-compose version: 1.18.0

**Additional context:**
It seems, the role limitedGuest has some strange ID in database. Content of table `role`:

```
registry=# select * from role;
 role_id | role_mask | role_code |     name
---------+-----------+-----------+--------------
       1 |         0 | MDRWS     | projectAdmin
       2 |         0 | RWS       | developer
       3 |         0 | RS        | guest
       4 |         0 | DRWS      | master
      34 |         0 | LRS       | limitedGuest
(5 rows)

registry=# select last_value from role_role_id_seq;
 last_value
------------
         34
(1 row)

```

But all users that are assigned with the limitedGuest role have non-existent role id set. In this case `5`:

```
registry=# select * from project_member where project_id=1;
 id  | project_id | entity_id | entity_type | role |       creation_time        |        update_time
-----+------------+-----------+-------------+------+----------------------------+----------------------------
   1 |          1 |         1 | u           |    1 | 2019-06-13 12:17:45.759919 | 2019-06-13 12:17:45.759919
 270 |          1 |         2 | g           |    5 | 2020-10-20 15:02:28.97263  | 2020-10-20 15:02:28.97263
(3 rows)
```

I have no idea at which point the role limitedGuest get created with `role_id` set to `34`.
I guess the `role_id` set to `5` is something that Harbor expects from fresh install.

As a workaround:
- I could manually update the table `role` in Harbor DB and set `role_id` to `5`. Please confirm this is safe.

As a fix:
- Harbor will get a patch where it will look for actual `role_id` instead of hardcoded values.

heww:
Closes #13317

### DIFF
--- a/make/migrations/postgresql/0050_2.2.0_schema.up.sql
+++ b/make/migrations/postgresql/0050_2.2.0_schema.up.sql
@@ -1,3 +1,10 @@
+/* 
+Fixes issue https://github.com/goharbor/harbor/issues/13317 
+  Ensure the role_id of maintainer is 4 and the role_id of limisted guest is 5
+*/
+UPDATE role SET role_id=4 WHERE name='maintainer' AND role_id!=4;
+UPDATE role SET role_id=5 WHERE name='limitedGuest' AND role_id!=5;
+
 ALTER TABLE schedule ADD COLUMN IF NOT EXISTS cron_type varchar(64);
 
 DO $$


ninjadq (approved):
lgtm

bitsf (approved):
The fix is a little confused, we should not use a magic id in code, but currently this seems be a easy way to fix.

reasonerjt:
if we add a column `id` and make it the sequence, we can use it role_id as an identifier of the role and hardcode it in the code of `core`?

heww:
> if we add a column `id` and make it the sequence, we can use it role_id as an identifier of the role and hardcode it in the code of `core`?

The `role_id` column is the primary key of the `role` table. Adding a column `id` and make it the sequence that makes the id as primary key and `role_id` as a normal key. The migration needs to add a new column, alter the `role_id` column,  update the `id` and `role_id` columns, and then check all the SQLs related `role` table, it's too expensive to fix issue #13317.

PR #13501, opened 2020-11-12
Support LDAP without memberof overlay

### LINKED ISSUE #9328: LDAP requirements incompatible with some LDAP deployments
**Description:**

Currently memberof overlay is a hard requirement when using groups with OpenLDAP.

The memberof overlay is incompatible with some deployments. From the documentation [0]:
''Note that slapo-memberOf is not compatible with syncrepl based replication, and should not be used in a replicated environment''.

This will be a deal-breaker for many deployments. To enable harbor LDAP in those environments we'd need support for matching (configurable) user entry attributes with (configurable) group attributes.

**Versions:**

- harbor version: 1.9.0

**Additional context:**

[0]  http://www.openldap.org/software/man.cgi?query=slapo-memberof&sektion=5&apropos=0&manpath=OpenLDAP+2.4-Release

thechristschn:
My harbor installation uses an openldap which doesn't have the memberof-overlay. Currently, I can't get the membership information for the users in harbor. Since the LDAP is not in my responsibility, I can''t just install the overlay.

This MR introduces a flag, which decides where to look for the membership attribute, at the user or at the group. In the default case, the flag is true and harbor would look for the memberof attribute at the user.
If the new flag is deactivated, harbor will search for the membership attribute at the groups. In most cases, this would be the `member` attribute. For each user that logs in, harbor will make a ldap search with the filter `(&(<default-group-filter>)(<membership-attribute>=<userDN>)`. The result are the groups where the user is a member of and created in harbor.

I would like to have some feedback about the way the problem would be handled. If it looks good, I can add a couple of tests. I'm also open to change the name of the flag (in Code and UI).

This might also resolve #9328, but I can't tell for sure, since it isn't my issue.

### DIFF
--- a/src/common/config/metadata/metadatalist.go
+++ b/src/common/config/metadata/metadatalist.go
@@ -99,6 +99,7 @@ var (
 		{Name: common.LDAPURL, Scope: UserScope, Group: LdapBasicGroup, EnvKey: "LDAP_URL", DefaultValue: "", ItemType: &NonEmptyStringType{}, Editable: false},
 		{Name: common.LDAPVerifyCert, Scope: UserScope, Group: LdapBasicGroup, EnvKey: "LDAP_VERIFY_CERT", DefaultValue: "true", ItemType: &BoolType{}, Editable: false},
 		{Name: common.LDAPGroupMembershipAttribute, Scope: UserScope, Group: LdapBasicGroup, EnvKey: "LDAP_GROUP_MEMBERSHIP_ATTRIBUTE", DefaultValue: "memberof", ItemType: &StringType{}, Editable: true},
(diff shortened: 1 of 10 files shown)


alexandrovas:
@thechristschn

There is another related problem. In some LDAP group attribute `member` can be contain another group (for nested group). For example in FreeIPA LDAP. So query `(&(<default-group-filter>)(<membership-attribute>=<userDN>)` will work only for cases if user is direct member of the group, but will not work for nested membership.

You need to rework the code to make an additional queries to the LDAP in such cases.

Current implementation of membership check logic in Harbor is very poorly implemented (especially in versions> = 2.1.0). I described this in the issue https://github.com/goharbor/harbor/issues/13780

PR #13525, opened 2020-11-17
feat: add cache library and enable it in config manager

heww:
1. Add two types of cache and initialize the Redis cache as the default cache.
2. Enable cache for the config manager when the default cache exists.


src/lib/retry.go
```diff
@@ -0,0 +1,36 @@
...
+
+package lib
+
+import (
+	"math/rand"
+	"time"
+)
+
+// Retry retry max attempts of f unit it's successful
+func Retry(attempts int, sleep time.Duration, f func() error) error {
```
reasonerjt, line 23: 1. This is opinionated but why not use a `bool` as the condition, when you wait for a condition it's not always  `err != nil`.  
2. Just curious, why you choose the exponential alg for the retry?  Normally it's for avoiding congestion.  We should also document this to avoid misuse.

src/lib/cache/cache.go
```diff
@@ -0,0 +1,126 @@
...
+
+// Initialize initialize the default cache from the addr
+func Initialize(addr string, opt ...Option) error {
+	c, err := New(addr, opt...)
+	if err != nil {
+		return err
+	}
+
+	if err := lib.Retry(5, time.Second, c.Ping); err != nil {
+		return fmt.Errorf("failed to ping the cache, error %v", err)
```
reasonerjt, line 115: If c.Ping failed due to different errors only the last one is logged.

src/lib/cache/context.go
```diff
@@ -0,0 +1,35 @@
...
+
+type cacheKey struct{}
+
+// FromContext returns cache from context
+func FromContext(ctx context.Context) (Cache, bool) {
+	c, ok := ctx.Value(cacheKey{}).(Cache)
+	return c, ok
+}
+
+// NewContext returns new context with cache
```
reasonerjt, line 29: Why config manager needs context?

src/lib/cache/codec.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"sync"
+
+	"github.com/goharbor/harbor/src/lib/log"
+	"github.com/vmihailenco/msgpack/v5"
+)
+
+// Codec codec interface for cache
+type Codec interface {
```
reasonerjt, line 25: I'm not sure if this is a must have for now.

Additionally, IMO the codec should be used by the reader/writer, for example, if we decide to store all objects into a cache, different objects may be encoded/decoded differently?

src/lib/cache/helper.go
```diff
@@ -0,0 +1,88 @@
...
+
+		// save the val to cache failed, copy it to the value directly
+		return simpleCopy(value, val)
+	}
+
+	return c.Fetch(key, value) // after the building, fetch value again
+}
+
+// OnceContext executes Once function when Cache found in ctx, otherwise builds and assigns the result to value
+func OnceContext(ctx context.Context, key string, value interface{}, builder func() (interface{}, error), expiration ...time.Duration) error {
```
reasonerjt, line 75: Not quite sure why this is needed.

src/lib/cache/helper.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/goharbor/harbor/src/lib/log"
+)
+
+var (
+	onceMu = keyMutex{m: &sync.Map{}}
+)
+
+// Once retrieve the cached key value, build and cache the value when the cached key value not found
+func Once(c Cache, key string, value interface{}, builder func() (interface{}, error), expiration ...time.Duration) error {
```
reasonerjt, line 32: The `Once` func is confusing, you call it more than once... also, having to add the first parm `c` means it should probably be moved as c's attribute.

It'd be better moved to the cache.Cache, introduce something like `loader` if the cache misses a key it always try to load it from the real source.  But then you need to introduce different loader for the cache b/c different key may map to different source.

Another simplified option I think is to move it to the caller(storage/driver) layer, so cache user need to worry about initializing the cache, if it misses the caller write the data to the cache.

src/lib/cache/helper.go
```diff
@@ -0,0 +1,88 @@
...
+	val, err := builder()
+	if err != nil {
+		return err
+	}
+
+	if err := c.Save(key, val, expiration...); err != nil {
+		log.Warningf("failed to save value to cache, error: %v", err)
+
+		// save the val to cache failed, copy it to the value directly
+		return simpleCopy(value, val)
```
reasonerjt, line 68: Here you imply the `builder` MUST return the same type of data as the input parm `value`, which may not be true in some caces.

src/lib/cache/util.go
```diff
@@ -0,0 +1,73 @@
...
+	m *sync.Map
+}
+
+func (km keyMutex) Lock(key interface{}) {
+	m := sync.Mutex{}
+	act, _ := km.m.LoadOrStore(key, &m)
+
+	mm := act.(*sync.Mutex)
+	mm.Lock()
+	if mm != &m {
```
reasonerjt, line 56: check the second ret value of `LoadOrStore` and you don't need this `if`?

src/lib/cache/util.go
```diff
@@ -0,0 +1,73 @@
...
+	return
+}
+
+func (km keyMutex) Unlock(key interface{}) {
+	act, exist := km.m.Load(key)
+	if !exist {
+		panic("unlock of unlocked mutex")
+	}
+	m := act.(*sync.Mutex)
+	km.m.Delete(key)
```
reasonerjt, line 71: I don't think you have to delete everytime, or you may use: https://godoc.org/sync#Map.LoadAndDelete

src/lib/cache/context.go
heww: Config manager does not need this, this is preparing for the DAO layers.

src/lib/cache/helper.go
heww: This is preparing for the DAOs. For example, adding a cache layer to the `Get` method of the DAO for the project may like this.


``` golang
func(d *dao) Get(ctx context.Context, id int64) (*models.Project, error) {
    key := fmt.Sprintf("project:%d", id)
    var project models.Project
    err := cache.OnceContext(ctx, key, &project, func() (interface{}, error) error {
        o, err := orm.FromContext(ctx)
        if err != nil {
            return nil, err
        }
	p := &models.Project{ProjectID: id, Deleted: false}
	if err = o.Read(p, "project_id", "deleted"); err != nil {
		return nil, orm.WrapNotFoundError(err, "project %d not found", id)
	}

        return p, nil
    })

   if err != nil {
       return nil, err
   }

   return &project, nil
}
```

src/lib/cache/util.go
heww: > LoadAndDelete deletes the value for a key, returning the previous value if any. The loaded result reports whether the key was present.

`LoadAndDelete` is not suitable here because it returns the `previous` value, the `Unlock` mutex must be the current one, not the `previous`.

Delete the mutex every time here to ensure that `act, _ := km.m.LoadOrStore(key, &m)` always let `act` be `m` when the `key` is not locked in the `Lock` method, otherwise, the `Lock` method will go to the `if mm != &m {` block which may drop in an infinite loop.

src/lib/cache/helper.go
heww: Rename this function name to `FetchOrSave`.

It's not very well to move this function to the `Cache` interface because this is a pattern to how to use cache to improve the performance of an expensive operation. This pattern is that try to fetch the value from the cache,  compute the value save it to the cache when it's not found in the cache so that next time we can fetch it from the cache directly.

This function does one more thing than the pattern is that it uses a lock to avoid computing and saving the value multi times in high concurrency.

The `intializer` is not part of the cache, it should be work together with a key. For example, we may only have one default Redis cache in the system. This cache can be used in the config manager, project DAO. I don't think `intializer` for the cache can match these scenarios.

src/lib/cache/helper.go
heww: Yes, they must be the same type. 

I don't think we will fetch a value which type is A from the cache,  and compute and return a value of type B when the key is not found in the cache.

src/lib/cache/util.go
heww: May because that `mm != &m` is clear to understand and will never be a mistake.

This if statement checks that the mutex load from the `sync.Map` must be the mutex created in the `Lock` function.  When it's not the mutex created in the `Lock` function, it means that another goroutine locked the key before this current goroutine, so the current goroutine needs to `Lock` the key again.

src/lib/cache/codec.go
heww: `Codec` is introduced to reduce the cost to use the cache library. We don't need to handle the details of how to save a struct to the cache, the fetched value from the cache is the type we wanted.

Without the codec implicitly related to the cache, we need to decode the value to the type the cache accepted before save it to the cache, usually, the accepted type is a string or bytes. And we also need to encode the value to the type we wanted after the value fetched from the cache.

The only problem for the codec is that it needs to deal with most types, luckily the msgpack works with primitives, arrays, maps, structs, time.Time and interface{}.

stonezdj:
Some of the code that use the redis pool should be changed to use cache library.

Blob Controller
https://github.com/goharbor/harbor/blob/e3fd83aad105c6f71fb2b3da0c9ff98a160c4cb9/src/controller/blob/controller.go#L299
Quota Controller
https://github.com/goharbor/harbor/blob/eb38180483a7001b4e66ec8f6ffc0f64b6f57306/src/controller/quota/controller.go#L170

stonezdj (approved):

src/lib/cache/cache.go
```diff
@@ -0,0 +1,139 @@
...
+		return err
+	}
+
+	cache = c
+
+	return nil
+}
+
+// Default returns the default cache
+func Default() Cache {
```
stonezdj, line 137: Maybe only one single cache type can not solve all our issue, we might need redis + memory cache together.  Is it possbile to get the cache by scope, for example GlobalCache, SessionCache, RequestCache etc. for example the RequestCache is only persist in request scope.

src/lib/cache/cache.go
```diff
@@ -0,0 +1,139 @@
...
+)
+
+// Initialize initialize the default cache from the addr
+func Initialize(addr string, opt ...Option) error {
+	c, err := New(addr, opt...)
+	if err != nil {
+		return err
+	}
+
+	u, _ := url.Parse(addr)
```
wy65701436, line 115: err handling?

src/lib/cache/cache.go
```diff
@@ -0,0 +1,139 @@
...
+	// ErrBadValue error returns when the value is not supported by the codec
+	ErrBadValue = errors.New("bad value")
+	// ErrNotFound error returns when the key value not found in the cache
+	ErrNotFound = errors.New("key not found")
+)
+
+// Cache cache interface
+type Cache interface {
+	// Contains returns true if key exists
+	Contains(key string) bool
```
wy65701436, line 38: should add error in the return, we cannot see the error condition as non-exist

src/lib/cache/util.go
```diff
@@ -0,0 +1,87 @@
...
+		return
+	}
+
+	return
+}
+
+func (km keyMutex) Unlock(key interface{}) {
+	act, exist := km.m.Load(key)
+	if !exist {
+		panic("unlock of unlocked mutex")
```
wy65701436, line 82: we should avoid using panic

heww: Unlock an unlocked mutex will throw a fatal error in stdlib https://golang.org/src/sync/mutex.go?s=5882:5906#L196.

Panic when unlocking an unlocked key here is ok.

src/lib/cache/cache.go
heww: No need to handle the err here because `New(addr, opt...)` success means there is not err happened when parse addr.

src/lib/cache/cache.go
heww: A chain cache is added.

reasonerjt: Sorry but I don't think a chain of cache is a good idea.  It will be problematic when a key exists in different caches and have different expiration time, different values.  
I don't think the scenario @stonezdj we need to worry about in midterm.  In the example he mentioned, the session/request cache would be handled by the framework but our code.

We should try to enforce that one key should only exist in one cache.  If this can be ensured, we can register multiple types of cache, use them separately without needing the chain.

src/lib/cache/chain.go
```diff
@@ -0,0 +1,85 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cache
+
+import (
+	"time"
+)
+
+type chain struct {
```
reasonerjt, line 21: Please see my comment above, can we refrain from introducing such advanced feature until we REALLY need?

src/lib/retry.go
```diff
@@ -0,0 +1,133 @@
...
+	// ErrRetryTimeout timeout error for retrying
+	ErrRetryTimeout = errors.New("retry timeout")
+)
+
+// RetryOptions options for the retry functions
+type RetryOptions struct {
+	InitialInterval time.Duration                        // the initial interval for retring after failure, default 100 milliseconds
+	MaxInterval     time.Duration                        // the max interval for retring after failure, default 1 second
+	Timeout         time.Duration                        // the total time before returning if something is wrong, default 1 minute
+	Callback        func(err error, sleep time.Duration) // the callback function for Retry when the f called failed
```
reasonerjt, line 38: This is trivial, but IMO 99% of cases of the callback are for printing a log message.
What you need here is a name of the retry session  so you can print a meaning message.

src/lib/cache/redis/redis.go
```diff
@@ -0,0 +1,132 @@
...
+	return conn.Do(commandName, args...)
+}
+
+// New returns redis cache
+func New(opts cache.Options) (cache.Cache, error) {
+	if opts.Address == "" {
+		opts.Address = "redis://localhost:6379/0"
+	}
+
+	name := fmt.Sprintf("%x", md5.Sum([]byte(opts.Address)))
```
reasonerjt, line 110: Is it for avoiding dup of pool name???

src/lib/cache/cache.go
```diff
@@ -0,0 +1,139 @@
...
+	if opts.Codec == nil {
+		opts.Codec = newCodec(u.Query().Get("codec"))
+	}
+
+	if opts.Prefix == "" {
+		opts.Prefix = "cache:"
+	}
+
+	for scheme, factory := range Factories() {
+		if scheme == u.Scheme {
```
reasonerjt, line 96: using the scheme to find a creator of the cache may be problematic, imagine there are two cloud based cache service which have different APIs but both have `https://` in the Address.

src/lib/cache/cache.go
```diff
@@ -0,0 +1,139 @@
...
+		return nil, err
+	}
+
+	opts := newOptions(opt...)
+
+	if opts.Address == "" {
+		opts.Address = addr
+	}
+
+	if opts.Codec == nil {
```
reasonerjt, line 87: if `?codec=xxx` is not a very well-known convention.  Can we just add `defaultCode` func and return a default one.

So user must specify the codec in his option if he really wants a customized codec.

src/lib/cache/redis/redis.go
```diff
@@ -0,0 +1,132 @@
...
+func (c *Cache) Ping() error {
+	_, err := c.do("PING")
+	return err
+}
+
+// Save cache the value by key
+func (c *Cache) Save(key string, value interface{}, expiration ...time.Duration) error {
+	data, err := c.opts.Codec.Encode(value)
+	if err != nil {
+		return cache.ErrBadValue
```
reasonerjt, line 77: Can we have more context, like bad value for key xxx, error: xxxx

src/lib/cache/util.go
reasonerjt: Current code works, but I believe the `previous` here means the `current` in your context.
It's the value that were mapped to the key before it's deleted.

src/lib/cache/util.go
```diff
@@ -0,0 +1,87 @@
...
+
+	mm := act.(*sync.Mutex)
+	mm.Lock()
+	if mm != &m {
+		mm.Unlock()
+		km.Lock(key)
+		return
+	}
+
+	return
```
reasonerjt, line 76: This is trivial, but what''s the purpose of these 2 return?

src/lib/cache/util.go
reasonerjt: I'm a little confused.... 
You did not remove `mm` from the map, so in the next `Lock` it's still `mm != &m` ?

src/lib/cache/util.go
```diff
@@ -0,0 +1,87 @@
...
+
+func indirect(reflectValue reflect.Value) reflect.Value {
+	for reflectValue.Kind() == reflect.Ptr {
+		reflectValue = reflectValue.Elem()
+	}
+	return reflectValue
+}
+
+func simpleCopy(dst interface{}, src interface{}) error {
+	from := indirect(reflect.ValueOf(src))
```
reasonerjt, line 45: is it correct to indirect src?

What if the dst is a pointer to pointer?  here src should stay unchanged?

src/lib/cache/helper.go
reasonerjt: So we should carefully document that the `value` in parm of `FetchOrSave` must be a pointer to the return value of `builder`?

Also see my comment in the simple copy, should we consider the case when the return value of `builder` is a pointer?  Shall we fail early if it's a pointer?

src/lib/cache/util.go
reasonerjt: I misunderstood, the point is that a go-routine needs to create a new mutex and manage the whole lifecycle of the mutex until the key is released.

src/lib/cache/redis/redis.go
heww: Yes, `redis.GetRedisPool` will return the pool directly when the name already exists in its cache.

src/lib/cache/cache.go
heww: Getting the codec from the query string of the address was removed. 

Users must provide the codec by `WithCodec` when they want to use a customized codec.

src/lib/cache/codec.go
```diff
@@ -0,0 +1,72 @@
...
+	// Encode returns the encoded byte array of v.
+	Encode(v interface{}) ([]byte, error)
+
+	// Decode analyzes the encoded data and stores
+	// the result into the pointer of v.
```
reasonerjt, line 30: result into v, not a pointer of v, because `v` itself is a pointer

src/lib/cache/codec.go
```diff
@@ -0,0 +1,72 @@
...
+func (*msgpackCodec) Encode(v interface{}) ([]byte, error) {
+	return msgpack.Marshal(v)
+}
+
+func (*msgpackCodec) Decode(data []byte, v interface{}) error {
+	return msgpack.Unmarshal(data, v)
+}
+
+func getCodec(typ string) Codec {
+	codecsMu.Lock()
```
reasonerjt, line 62: do you need to lock?

src/lib/cache/helper.go
```diff
@@ -0,0 +1,78 @@
...
+	val, err := builder()
+	if err != nil {
+		return err
+	}
+
+	if err := c.Save(key, val, expiration...); err != nil {
+		log.Warningf("failed to save value to cache, error: %v", err)
+
+		// save the val to cache failed, copy it to the value using the default codec
+		codec := getCodec("default")
```
reasonerjt, line 68: Non-blocking, but here we need a way to get codec of cache otherwise the design to make codec configurable per cache does not make sense.

src/lib/cache/redis/redis.go
reasonerjt: if this is the case how about adding a prefix like `cache`?

src/lib/cache/util.go
```diff
@@ -0,0 +1,61 @@
...
+	act, _ := km.m.LoadOrStore(key, &m)
+
+	mm := act.(*sync.Mutex)
+	mm.Lock()
+	if mm != &m {
+		mm.Unlock()
+		km.Lock(key)
+	}
+
+	return
```
reasonerjt, line 50: why need `return`?

reasonerjt (approved):
A few trivial comments.

src/lib/cache/helper.go
heww: All the cache changed to use the default codec now.

Issue #9149, opened 2019-09-19
systeminfo API may show less info when it's triggered by anonymous(unauthenticated) user


reasonerjt:
The example of response of GET /api/systeminfo
```
{
  "with_notary": false,
  "with_clair": false,
  "with_admiral": false,
  "admiral_endpoint": "",
  "auth_mode": "oidc_auth",
  "registry_url": "jt-dev.local.goharbor.io",
  "external_url": "https://jt-dev.local.goharbor.io",
  "project_creation_restriction": "everyone",
  "self_registration": false,
  "has_ca_root": false,
  "harbor_version": "v1.9.0-6dd2ae90",
  "registry_storage_provider_name": "filesystem",
  "read_only": false,
  "with_chartmuseum": false,
}
```

some of the data, such as `registry_storage_provider_name` `project_creation_restriction` is not necessary to be included when the requester is not authenticated.

So from security's perspective, we may consider return extra data only when user is authenitcated or have proper permission.

paulczar:
This unauthenticated endpoint exposes too much information about the harbor install, information that could help an attacker determine the attack service of harbor and its dependencies.

here's the output from my recent install.   An attacker would find out I have notary disabled, that I'm using db authentication, self registration is disabled, when my last clair vulnerability updates happened, and that I'm using s3 for storage. There's no reason an anonymous user should ever know this data.

```
{
  "with_notary": false,
  "with_clair": true,
  "with_admiral": false,
  "admiral_endpoint": "NA",
  "auth_mode": "db_auth",
  "registry_url": "harbor:8443",
  "external_url": "https://harbor:8443",
  "project_creation_restriction": "everyone",
  "self_registration": false,
  "has_ca_root": true,
  "harbor_version": "v1.9.3-c80751b8",
  "clair_vulnerability_status": {
    "overall_last_update": 1575578301,
    "details": [
      {
        "namespace": "amzn",
        "last_update": 1575578301
      },
      {
        "namespace": "debian",
        "last_update": 1575578301
      },
      {
        "namespace": "centos",
        "last_update": 1575578301
      },
      {
        "namespace": "ubuntu",
        "last_update": 1575578301
      },
      {
        "namespace": "oracle",
        "last_update": 1575578301
      },
      {
        "namespace": "alpine",
        "last_update": 1575578301
      }
    ]
  },
  "registry_storage_provider_name": "s3",
  "read_only": false,
  "with_chartmuseum": true,
  "notification_enable": true
}
```

[...]

PR #13539, opened 2020-11-18
Add exporter for harbor

ninjadq:
Add exporter component


bitsf (approved):

src/pkg/exporter/project_collector.go
```diff
@@ -0,0 +1,239 @@
...
+
+	"github.com/goharbor/harbor/src/common/dao"
+	"github.com/goharbor/harbor/src/lib/log"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// ProjectCollectorName ...
+const ProjectCollectorName = "ProjectCollector"
+
+var (
```
reasonerjt, line 15: Please make sure the SQLs are tested in the integration test otherwise a change in DB schema will break it.

src/cmd/exporter/main.go
```diff
@@ -0,0 +1,79 @@
+package main
+
+import (
+	"net/http"
+	_ "net/http/pprof"
```
reasonerjt, line 5: Do we really need this, afaik there's an overhead.

src/cmd/exporter/main.go
```diff
@@ -0,0 +1,79 @@
...
+			Port:         viper.GetInt("database.port"),
+			Username:     viper.GetString("database.username"),
+			Password:     viper.GetString("database.password"),
+			Database:     viper.GetString("database.dbname"),
+			SSLMode:      viper.GetString("database.sslmode"),
+			MaxIdleConns: viper.GetInt("database.max_idle_conns"),
+			MaxOpenConns: viper.GetInt("database.max_open_conns"),
+		},
+	}
+	if err := dao.InitDatabase(dbCfg); err != nil {
```
reasonerjt, line 38: Set the number of conns to relatively low, we don't expect the exportor to use many connections

make/photon/prepare/utils/misc.py
```diff
@@ -163,3 +163,9 @@ def get_realpath(path: str) -> Path:
     if os.path.isdir(host_root_dir):
         return host_root_dir.joinpath(path.lstrip('/'))
     return Path(path)
+
+def port_number_valid(port:int):
```
reasonerjt, line 167: This is trivial but why not

return port >0 and port <=65535

src/pkg/exporter/cache.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"sync"
+	"time"
+)
+
+var c *cache
+
+type cachedValue struct {
+	Value      interface{}
+	Expiration int64
```
reasonerjt, line 12: Why not use time.Time?

src/pkg/exporter/cache.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+type cache struct {
+	CacheTime int64
```
reasonerjt, line 16: Why this attribute is exported?

Also Cachetime is a confusing name.

src/pkg/exporter/cache.go
```diff
@@ -0,0 +1,88 @@
...
+	store     map[string]cachedValue
+	*sync.RWMutex
+}
+
+// CacheGet get a value from cache
+func CacheGet(key string) (value interface{}, ok bool) {
```
reasonerjt, line 22: This is not a blocker but ideally you should limit the accepted keys, otherwise the cache can grow endlessly by current design.

src/pkg/exporter/cache.go
```diff
@@ -0,0 +1,88 @@
...
+
+// CacheDelete delete a key from cache
+func CacheDelete(key string) {
+	c.Lock()
+	defer c.Unlock()
+	delete(c.store, key)
+}
+
+// StartCacheCleaner start a cache clean job
+func StartCacheCleaner() {
```
reasonerjt, line 56: why not call CacheClean... this func is a verb..

src/pkg/exporter/cache.go
reasonerjt, line 22: If this func is not intended to be called outside of the package, you should not export it.

The exporter should be pretty self contained, i.e. the pkg should expose relative minimum funcs.

Please go through all other funcs and objects, for example, it seems you don't need to export HarborClient either.

src/pkg/exporter/collector.go
```diff
@@ -0,0 +1,92 @@
...
+	scrapeSuccess = typedDesc{
+		desc:      newDescWithLables(subsystem, "collector_success", " Whether a collector succeeded.", "collector"),
+		valueType: prometheus.GaugeValue,
+	}
+)
+
+var hbrCli *HarborClient
+
+// HarborClient is client for request harbor
+type HarborClient struct {
```
reasonerjt, line 30: It seems so far all you need is http.Get and a URL constructor.

src/pkg/exporter/collector.go
```diff
@@ -0,0 +1,92 @@
...
+	hbrURL := hc.harborURL(p)
+	res, err := http.Get(hbrURL.String())
+	if err != nil {
+		return nil, err
+	}
+	return res, nil
+}
+
+// InitHarborClient initialize the harbor client
+func InitHarborClient(hc *HarborClient) {
```
reasonerjt, line 56: is it needed?

src/pkg/exporter/collector.go
```diff
@@ -0,0 +1,92 @@
...
+	hbrCli = hc
+}
+
+func newDesc(subsystem, name, help string) *prometheus.Desc {
```
reasonerjt, line 60: if `labels` is optional you don't need `newDesc` and `newDescWithLabels`

src/pkg/exporter/exporter.go
```diff
@@ -0,0 +1,97 @@
...
+// Opt is the config of Harbor exporter
+type Opt struct {
+	Port                   int
+	MetricsPath            string
+	ExporterMetricsEnabled bool
+	MaxRequests            int
+	TLSEnabled             bool
+	Certificate            string
+	Key                    string
+	CacheTime              int64
```
reasonerjt, line 24: see comments above regarding expiration.

src/pkg/exporter/exporter.go
```diff
@@ -0,0 +1,97 @@
...
+	"net/http"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+
+	"github.com/goharbor/harbor/src/lib/log"
+)
+
+// Opt is the config of Harbor exporter
+type Opt struct {
```
reasonerjt, line 16: a few attributes are not used.

the attribute `ExporterMetricsEnabled` seems redundant.  We just skip starting the exporter if we don't want to enable it.

src/pkg/exporter/cache.go
ninjadq: time.Time consumes more memory space and less efficient when comparing two values.

src/pkg/exporter/cache.go
ninjadq: CacheClean can not fully describe this function

src/pkg/exporter/cache.go
reasonerjt: So then explicitly add the unit of time please, and document everywhere so there's no misunderstanding.

src/pkg/exporter/cache.go
reasonerjt: It's not an async func, you are not starting any job or goroutine.  I really can't see why we have to add `Start`.

src/pkg/exporter/exporter.go
```diff
@@ -0,0 +1,97 @@
...
+	Port                   int
+	MetricsPath            string
+	ExporterMetricsEnabled bool
+	MaxRequests            int
+	TLSEnabled             bool
```
reasonerjt, line 21: Other attributes seem not used such as `MaxRequests` and `TLSEnabled `

src/pkg/exporter/exporter.go
```diff
@@ -0,0 +1,97 @@
...
+	r.MustRegister(exporter)
+	exporter.Server = newServer(opt, r)
+
+	return exporter
+}
+
+// Exporter is struct for Harbor which can used to connection Harbor and collecting data
+type Exporter struct {
+	*http.Server
+	Opt        *Opt
```
reasonerjt, line 50: Exposing Opt may be confusing, what if you intialized an Exporter and later the Opt of an Export is updated?  This will cause inconsistency.

src/pkg/exporter/exporter.go
```diff
@@ -0,0 +1,97 @@
...
+	ctx        context.Context
+	c          *cache
```
reasonerjt, line 52: `c` and `ctx` are not used?

src/pkg/exporter/exporter.go
```diff
@@ -0,0 +1,97 @@
...
+	Collectors map[string]prometheus.Collector
```
reasonerjt, line 53: Same issue, if it's exposed, there will be inconsistency if Collectors are written but not registered.

src/pkg/exporter/cache.go
```diff
@@ -0,0 +1,88 @@
...
+	defer c.Unlock()
+	for k, v := range c.store {
+		if v.Expiration < now {
+			delete(c.store, k)
+		}
+	}
+}
+
+// CacheEnabled returns if the cache in exporter enabled
+func CacheEnabled() bool {
```
reasonerjt, line 68: Can we assume cache is always enabled?

src/pkg/exporter/health_collector.go
```diff
@@ -0,0 +1,92 @@
...
+	for _, v := range healthResponse.Components {
+		result = append(result, harborComponentsHealth.MustNewConstMetric(healthy(v.Status), v.Name))
+	}
+	if CacheEnabled() {
+		CachePut(healthCollectorName, result)
+	}
+	return result
+}
+
+type responseHealth struct {
```
reasonerjt, line 77: We should reuse the data structure to be more resilient to change.

src/pkg/exporter/system_collector.go
```diff
@@ -0,0 +1,86 @@
...
+		sysInfoResponse.ExternalURL,
+		sysInfoResponse.HarborVersion,
+		sysInfoResponse.StorageProvider))
+	if CacheEnabled() {
+		CachePut(systemInfoCollectorName, result)
+	}
+	return result
+}
+
+type responseSysInfo struct {
```
reasonerjt, line 80: Reuse the data structure, this is defined in swagger and can be generated via `make gen_apis`

Also be careful that some attributes of the response will be missing the request is not authenticated:
https://github.com/goharbor/harbor/issues/9149

src/pkg/exporter/project_collector.go
```diff
@@ -0,0 +1,239 @@
...
+		c <- projectQuote.MustNewConstMetric(getQuotaValue(p.Quota), p.Name)
+		c <- projectMemberTotal.MustNewConstMetric(p.MemberTotal, p.Name)
+		c <- projectRepoTotal.MustNewConstMetric(p.RepoTotal, p.Name, getPublicValue(p.Public))
+		c <- artifactPullTotal.MustNewConstMetric(p.PullTotal, p.Name)
+		c <- projectArtifactTotal.MustNewConstMetric(p.Artifact.ArtifactTotal, p.Name, getPublicValue(p.Public), p.Artifact.ArtifactType)
+	}
+}
+
+type projectOverviewInfo struct {
+	projectTotals []projectCount
```
reasonerjt, line 108: would it make more sense to render public, non-public explicitly as the projectCount array will always have 2 entries?

src/pkg/exporter/cache.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+var c *cache
+
+type cachedValue struct {
+	Value      interface{}
+	Expiration int64
+}
+
+type cache struct {
```
reasonerjt, line 15: If you define an interface of cache it will help in UT.

src/pkg/exporter/project_collector.go
```diff
@@ -0,0 +1,239 @@
...
+		} else {
+			log.Errorf("%v, ID %d", errProjectNotFound, p.ProjectID)
+		}
+	}
+}
+
+func updateProjectArtifactInfo(projectMap map[int64]*projectInfo) {
+	aList := make([]artifactInfo, 0)
+	_, err := dao.GetOrmer().Raw(projectArtifactsSQL).QueryRows(&aList)
+	if err != nil {
```
reasonerjt, line 223: do not eat the err

src/pkg/exporter/project_collector.go
```diff
@@ -0,0 +1,239 @@
...
+	for _, a := range aList {
+		if _, ok := projectMap[a.ProjectID]; ok {
+			projectMap[a.ProjectID].Artifact = a
+		} else {
+			log.Errorf("%v, ID %d", errProjectNotFound, a.ProjectID)
+		}
+	}
+}
+
+func checkErr(err error, msg string) {
```
reasonerjt, line 235: This is not called.

src/pkg/exporter/cache.go
ninjadq: check  the `CacheInit` func

src/pkg/exporter/cache.go
ninjadq: actually, cache can be independent pkg but exporter case is simple. so all the component shared the same package name

src/pkg/exporter/collector.go
ninjadq: why not?

src/pkg/exporter/collector.go
ninjadq: Yep, But it's behavior is a harbor client

src/pkg/exporter/cache.go
ninjadq: OK, I can add comments on this. But there are numerous cache systems that use timestamps to represent expiration time. I think people will have a concept when they saw type `int64` as expiration.

src/pkg/exporter/exporter.go
ninjadq: `ExporterMetricsEnabled` is planned to use as a flag to decide if expose the metrics of exporter itself. Harbor info and exporter metrics are combined to expose at the moment. I can remove it now, and add it back in the future

reasonerjt (approved):

PR #13557, opened 2020-11-20
fix: remove auto re-readable body in BeforeRequest

### LINKED ISSUE #13556: Core component eat too much memory when execute replication
```
Both nodes running on v2.1.x Harbor versions consumes all the available RAM. Here are some observations:
5.1. This leads to a huge LA and, in most of the cases, to DoS.
5.2. harbor_core process also consumed those cases, when replication was able to successfully complete, the harbor_core process uses more than 80% of RAM and 15% of CPU. Overall, LA, at that moment, is 59.
5.3. Even after two hours after replication was able to successfully finish (a quite rear case), the harbor_core process was still consuming 71% of RAM, 1.5% of CPU. LA was 0.10.
5.4. The highest load for the v2.0.3 node was observed: harbor_core used 2.3% of RAM, 16.4% of CPU, and overall LA wasn't higher than 5. So I assume that starting from v2.1.0, harbor_core has a memory leak issue.
```
See https://github.com/goharbor/harbor/issues/13364#issuecomment-723136948 go get more info.

heww:
Remove nop close body in the BeforeRequest helper function.
Middleware must make the request body re-readable itself when it wants
read the body in the middleware.

Closes #13556

### DIFF
--- a/src/server/middleware/blob/put_manifest.go
+++ b/src/server/middleware/blob/put_manifest.go
@@ -15,10 +15,11 @@
 package blob
 
 import (
-	"github.com/goharbor/harbor/src/lib/errors"
 	"io/ioutil"
 	"net/http"
 
+	"github.com/goharbor/harbor/src/lib"
+	"github.com/goharbor/harbor/src/lib/errors"
 	"github.com/goharbor/harbor/src/lib/log"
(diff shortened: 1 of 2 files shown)


wy65701436 (approved):
lgtm

reasonerjt (approved):

PR #13600, opened 2020-11-25
fix(chartmuseum) compatible s3 cache fail

### LINKED ISSUE #13592: list chartmuseum 2 quite slowing when using DELL EMC s3 storage
When install harbor with chartmuseum and configure s3 on dell emc, it will always be slow to list projects with charts or list charts.
If the Api takes more than 30 seconds, it will directly failed to list charts

bitsf:
close #13592 
some compatible s3 like DELL EMC s3 has some strange behavior that make the chart index cache not working


ninjadq (approved):
lgtm

reasonerjt (approved):

PR #13615, opened 2020-11-26
Add limitation to artifact tags

### LINKED ISSUE #13099: Get artifact api should limit the count of tags returned
API: /projects/{project_name}/repositories/{repository_name}/artifacts/{reference}

When trying to reproduce the issue #13098 with Harbor 2.1, found this API doesn't limit the tags returned, when there are 5000 tags in same repo, it will return all tags and took long time to return (26-27 seconds)

It is better to limit the max count of tags returned.

https://github.com/goharbor/harbor/blob/91e2779822ae0c096ae97ea9de7c4cac20ab0a61/src/controller/artifact/controller.go#L599

### LINKED ISSUE #13488: Web interface display wrong info in the bubble
Normally when we hover the mouse on the tag of the image, Harbor WebUI will display a bubble telling some info.
I saw that, for signed image, the bubble display wrong info. It should be a screen stick instead of a red X for signed image.  

image

AllForNothing:
Fixes #13488
Fixes #13099
image


jwangyangls (approved):
lgtm

xuelichao (approved):

Issue #9006, opened 2019-09-06
Disabling admin user when using OIDC authentication


usualstuff:
Hi. I want to disable ability to login in any way except OIDC. That's includes default admin account.

It would be preferable to completely disable default login form but a way to disable admin account should suffice too.

reasonerjt:
Not a very valid requirement to me.

What's the purpose of this?  
Have you considered if the OIDC endpoint is down, or the dns name changed, how you gonna make such change if the admin is completely disabled?

reasonerjt:
@xaleeks let me know your thoughts.

usualstuff:
Well if my OIDC endpoint is down I will have a much bigger problem than just a broken registry :)

But I get your point. Maybe then this change could be implemented as a configuration flag so I can revert that in case of some disasters?

It's for a security reasons. I want to have a single source of truth about my users.

xaleeks:
I'm not sure I understand, can you give an example of where it introduces non single version of truth and how this prevents it @usualstuff

usualstuff:
Hi @xaleeks.

Basically right now I have two source of users to login: 
- harbor database which contains only root user 
- my OIDC database

I want harbor to only use OIDC so I don't have to manage separate credentials and monitor\audit another login endpoint.

If I'm able to disable or delete this user I will have only one source of harbor users.

For a reference you could look at gitlab. For them root user is just another user like everyone else and can be deleted or disabled. In case of some disaster it could be recreated from ruby cli.

svyotov:
@xaleeks the admin user is kind of vulnerability since it has admin privileges.
Any deployment will have to have to set up credentials management system, and rotation.
I am also guessing this does not have alerts for login or MFA, so if some attacker manages to guess using brute force the admin password, then Harbor will be compromised.

jeff-mccoy:
@reasonerjt has there been any more discussion on this?  This is a very valid security concern.

cafeliker:
Agree on this security concern, pls consider the enterprise level of cyber security requirements. A basic account should not be allowed for a production instance.

rochaporto:
Same here, it would be nice to have a config flag disabling basic auth.

MrD2:
Hello.
This came as a requirement for our use case also.

What can we do to speed this up? Should i take this? Are there any implementation thoughts on this?

xaleeks:
recently seeing variations of this more in govt. customers as well, we can discuss it for the next release

[...]

PR #13620, opened 2020-11-26
Add optional DBAuth

dioguerra:
fixes: https://github.com/goharbor/harbor/issues/9006

[portal]
* Only show username:password login fields for DBAuth if enabled (default: true)

[core]
* Add BASIC_AUTH_ENABLED to api/configurations to enable/disable user authentication using DBAuth (includes admin)

**HALP NEEDED**
I didn't manage to compile the core container after applying changes, and i don't know why it's not triggering

Also, my idea here is to enable/disable the registration of the DBAuth AuthN method so that authentication.login() would only work with other set-up methods. Do you think this is the correct way.
This actually depends on how the harbor-core reloads the configurations after the user submits these.

This is my first commit here, so i am a bit lost, still trying to understand the Arch. Some points would be nice

### DIFF
--- a/api/v2.0/legacy_swagger.yaml
+++ b/api/v2.0/legacy_swagger.yaml
@@ -3961,6 +3961,9 @@ definitions:
       external_url:
         type: string
         description: The external URL of Harbor, with protocol.
+      basic_auth_enabled:
+        type: boolean
+        description: 'Allow basicAuth login when auth_mode is not basicAuth'
       auth_mode:
         type: string
         description: The auth mode of current Harbor instance.
@@ -4075,6 +4078,9 @@ definitions:
   Configurations:
     type: object
     properties:
+      basic_auth_enabled:
+        type: boolean
+        description: 'Allow basicAuth login when auth_mode is not basicAuth'
       auth_mode:
         type: string
         description: 'The auth mode of current system, such as "db_auth", "ldap_auth"'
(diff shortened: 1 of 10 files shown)


MrD2:
I'm still having this error: '{"errors":[{"code":"BAD_REQUEST","message":"configure item is not defined in metadata, item name: basic_auth_enabled"}]}'

Which means that the parameter is not being correctly registered into swagger's API.

Any thoughts?

reasonerjt:
I don't disagree this may be useful in your scenario, but we want to balance the configurations that are exposed and options that are provide to user in order to control the complexity.

After review, we believe this is not a very widely required use case, hence, I'm closing this PR.

Please feel free to ping me on slack if you want to discuss.

rochaporto:
Is this really adding a lot of complexity? We still have users passing their user/pass in this form thinking they're logging in with their SSO credentials. We have an open issue with our security team regarding this, and might need a downstream patch if this doesn't make it upstream - meaning cloning the repo, rebuilding packages, releasing downstream, etc.

reasonerjt:
@rochaporto This PR alone does not adding A LOT complexity.

I'm discussing with Diogo on slack for a way to mitigate the security issue without having to add another configuration.  Hopefully, he will open another issue and provide a fix that everyone is OK with.

DaniD3v:
> @rochaporto This PR alone does not adding A LOT complexity.
> 
> I'm discussing with Diogo on slack for a way to mitigate the security issue without having to add another configuration. Hopefully, he will open another issue and provide a fix that everyone is OK with.

@reasonerjt 

What was the outcome of that discussion? #9006 has 17 upvotes.
This is pretty widely requested and poses an issue to every company with strict compliance needs.

PR #13623, opened 2020-11-27
updates on robot accounts

wy65701436:
1, add patch method to refresh secret of a robot
2, fix robot account update issue
3, add editable attribute to handle the version 1 robot account
4, add duration for robot account


stonezdj (approved):

src/controller/robot/controller.go
```diff
@@ -80,9 +80,14 @@ func (d *controller) Create(ctx context.Context, r *Robot) (int64, error) {
...
-		tokenDuration := time.Duration(config.RobotTokenDuration()) * time.Minute
-		r.ExpiresAt = time.Now().UTC().Add(tokenDuration).Unix()
+	var expiresAt int64
+	if r.Duration == -1 {
+		expiresAt = -1
+	} else if r.Duration == 0 {
+		// system default robot duration
+		expiresAt = time.Now().UTC().Add(time.Duration(config.RobotTokenDuration()) * 60 * 24 * time.Minute).Unix()
+	} else {
+		expiresAt = time.Now().UTC().Add(time.Duration(r.Duration) * 60 * 24 * time.Minute).Unix()
```
heww, line 90: `expiresAt = time.Now().AddDate(0, 0, r.Duration).Unix()` ?

No need to convert the time to UTC timezone because the timestamp is timezone independent.

src/controller/scan/base_controller.go
```diff
@@ -600,7 +600,7 @@ func (bc *basicController) makeRobotAccount(ctx context.Context, projectID int64
 			Name:        fmt.Sprintf("%s-%s", registration.Name, UUID),
 			Description: "for scan",
 			ProjectID:   projectID,
-			ExpiresAt:   -1,
+			Duration:    int64(config.RobotTokenDuration()),
```
heww, line 603: This line can be removed as the robot controller will set it to the value from the configuration when it's zero.

src/controller/robot/controller.go
wy65701436: updated.

src/controller/scan/base_controller.go
wy65701436: removed

heww (approved):

PR #13633, opened 2020-11-30
[CherryPick-2.1.0]Move security/readonly middleware ahead of transaction middleware

reasonerjt:
In v2.1 security/readonly middleware will query DB by creating new
connection.
If it is put after transaction middleware there's a bigger chance of
deadlock if the concurrent open connections are set too low (#13155)

This commit mitigates that issue.  But we still need work to lowerthe
connections and better handle the case when http connection is closed.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -27,7 +27,7 @@ jobs:
     runs-on:
       #- self-hosted
       - ubuntu-latest
-    timeout-minutes: 60
+    timeout-minutes: 180
     steps:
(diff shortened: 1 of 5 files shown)


wy65701436 (approved):
lgtm

heww (approved):

reasonerjt:
Currently solution will cause failure in proxy cache test case because the security context generator needs artifact info in the context. Closing this PR, and opened https://github.com/goharbor/harbor/pull/13686

PR #13640, opened 2020-11-30
Change image source in e2e pytest

danfengliu:
Change source of most of test image samples from docker-hub to local building ones, so it will cost less docker-hub pull requests.
And some of cases like push cnab, they have to use docker-hub, but image samples in cnab test will cost 17 quotas, in this PR, we
replace those samples, now cnab case will cost 6 quotas.


AllForNothing (approved):

wy65701436 (approved):
lgtm

PR #13641, opened 2020-11-30
Consider the default port when comparing the hosts

reasonerjt:
This commit cover the cases when the port is set in one of the Host of
request or the core URL, to make sure the comparison works as expected
when the default port (80, 443) is added in only one of them.

### DIFF
--- a/src/server/middleware/v2auth/auth_test.go
+++ b/src/server/middleware/v2auth/auth_test.go
@@ -266,3 +266,65 @@ func TestGetChallenge(t *testing.T) {
 	}
 
 }
+
+func TestMatch(t *testing.T) {
+	cases := []struct {
+		reqHost string
+		rawURL  string
+		expect  bool
+	}{
+		{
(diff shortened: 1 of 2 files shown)


src/server/middleware/v2auth/auth_test.go
```diff
@@ -266,3 +266,55 @@ func TestGetChallenge(t *testing.T) {
...
+		{
+			"abcd.com:443",
+			"https://abc.com",
+			false,
+		},
+		{
+			"abc.com:8443",
+			"https://abc.com:8443",
+			true,
+		},
```
bitsf, line 315: should also consider case
{
			"abc.com",
			"https://abc.com:443",
			true,
		},

reasonerjt: There is a case in line 301

bitsf (approved):

heww (approved):

Issue #1744, opened 2017-03-23
The "Remember me" doesn't seem to work.


reasonerjt:
Seems there's no effect whether or not I check the "Remember me" when logging in.
<img width="485" alt="screen shot 2017-03-23 at 5 28 07 pm" src="https://cloud.githubusercontent.com/assets/2390463/24241051/8e33378e-0fee-11e7-84cf-bf464d3d0b7a.png">

steven-zou:
To securely support this function , might need a backend api support.

steven-zou:
@lweitzman 

  About the 'remember me', just remember the username or remember the sigin session?

steven-zou:
@lweitzman 

  Can you give me the answer?

lweitzman:
just remember the username

steven-zou:
Support now.

reasonerjt:
Verified it remembers the username.

PR #13673, opened 2020-12-02
Changes "Remember me" to the more intuitive "Remember username"

gwkunze:
It took me finding #1744 to figure out why I kept having to log in to the portal every so often. Personally I've come to expect "Remember Me" to mean, "Don't make me have to log in for a while" (where "a while" is a period significantly longer than what happens when it is not selected). 

I also changed the INVALID_MSG's `"user name"` to `"username"` as that is more consistent with the placeholder text used for the field itself. According to the [dictionary](https://www.merriam-webster.com/dictionary/username) no space in username seems to be valid.

(last change is just my editor trimming trailing spaces)

Signed-off-by: Gijs Kunze <gwkunze@gmail.com>

### DIFF
--- a/src/portal/src/i18n/lang/en-us-lang.json
+++ b/src/portal/src/i18n/lang/en-us-lang.json
@@ -10,8 +10,8 @@
         "THEME_LIGHT_TEXT": "LIGHT"
     },
     "SIGN_IN": {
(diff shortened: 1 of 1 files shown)


gwkunze:
Not sure how me minimally editing a json file affects code coverage... Could this be a fluke?

PR #13684, opened 2020-12-03
refactor(scan): refactor scan/scan all job to task manager

heww:
1. Remove columns related to the scan job and change to fetch the status from the task.

2. Scan all move to one execution and multi-tasks mode.

3. Scan the artifact move to one execution and one task mode.


src/controller/scan/base_controller.go
```diff
@@ -233,78 +266,116 @@ func (bc *basicController) Scan(ctx context.Context, artifact *ar.Artifact, opti
...
+		return 0, err
 	}
 
-	// Parse options
-	ops, err := parseOptions(options...)
-	if err != nil {
-		return "", nil, errors.Wrap(err, "scan controller: scan")
+	if async {
+		go bc.startScanAll(bc.makeCtx(), executionID, async)
+	} else {
```
ywk253100, line 278: is it better putting the checking for `async` inside the function `startScanAll`?

src/controller/scan/base_controller.go
```diff
@@ -233,78 +266,116 @@ func (bc *basicController) Scan(ctx context.Context, artifact *ar.Artifact, opti
...
-		}
+	return executionID, nil
+}
 
-		_, e := bc.manager.Create(reportPlaceholder)
-		return e
+func (bc *basicController) startScanAll(ctx context.Context, executionID int64, async bool) error {
+	if async {
+		// if async, this method is running in another goroutine ensure the execution exists in db
+		err := lib.RetryUntil(func() error {
```
ywk253100, line 290: Why we need this? Shouldn't the execution record already be there as you have called `bc.execMgr.Create`?

src/controller/scan/base_controller.go
```diff
@@ -233,78 +266,116 @@ func (bc *basicController) Scan(ctx context.Context, artifact *ar.Artifact, opti
...
 		}
+	}
 
-		if len(producesMimes) > 0 {
-			return trackID, producesMimes, nil
+	count, err := bc.taskMgr.Count(ctx, q.New(q.KeyWords{"execution_id": executionID}))
+	if err != nil {
+		log.Errorf("failed to get the count of tasks for scan all, error %v", err)
+		return err
+	} else if count == 0 {
```
ywk253100, line 316: If all callings of `bc.Scan()` failed(no task record is created), the task count is 0, but this doesn't the case of `no task found`

src/pkg/task/model.go
```diff
@@ -52,6 +55,16 @@ type Execution struct {
 	EndTime    time.Time              `json:"end_time"`
 }
 
+// IsOnGoing returns true when the execution is pending/running
+func (exec *Execution) IsOnGoing() bool {
```
ywk253100, line 59: `Execution` has no `pending` status, so you can just determine whether an execution is ongoing by comparing the status with `running`

src/controller/scan/base_controller.go
heww: When startScanAll in another goroutine, the API may not be finished but the `startScanAll` be finished. In this case `bc.execMgr.MarkDone(ctx, executionID, "no task found")` will failed because the execution not found in db.

src/controller/artifact/helper.go
```diff
@@ -0,0 +1,56 @@
...
+		defer close(ch)
+
+		clone := q.MustClone(query)
+		clone.PageNumber = 1
+		clone.PageSize = int64(chunkSize)
+
+		for {
+			artifacts, err := Ctl.List(ctx, clone, option)
+			if err != nil {
+				log.Errorf("[scan all]: list artifacts failed, error: %v", err)
```
steven-zou, line 38: `[scan all]` is a little confusing with vuln scan all

src/controller/event/handler/init.go
```diff
@@ -31,7 +32,7 @@ func init() {
 	notifier.Subscribe(event.TopicQuotaWarning, &quota.Handler{})
 	notifier.Subscribe(event.TopicScanningFailed, &scan.Handler{})
 	notifier.Subscribe(event.TopicScanningCompleted, &scan.Handler{})
-	notifier.Subscribe(event.TopicDeleteArtifact, &scan.DelArtHandler{})
+	notifier.Subscribe(event.TopicDeleteArtifact, &scan.DelArtHandler{Context: orm.Context})
```
steven-zou, line 35: Why only this handler is a little different?  Require a `Context: orm.Context`?

src/controller/scan/callback.go
```diff
@@ -0,0 +1,110 @@
...
+	return err
+}
+
+func scanTaskStatusChange(ctx context.Context, taskID int64, status string) (err error) {
+	logger := log.G(ctx).WithFields(log.Fields{"task_id": taskID, "status": status})
+
+	js := job.Status(status)
+
+	if js.Final() {
+		t, err := task.Mgr.Get(ctx, taskID)
```
steven-zou, line 59: It seems here if the task manager can inject the task object itself, duplicated code can be reduced, right?

@ywk253100

make/migrations/postgresql/0050_2.2.0_schema.up.sql
```diff
@@ -268,3 +268,13 @@ BEGIN
   UPDATE scanner_registration SET is_default = TRUE WHERE name = 'Trivy' AND immutable = TRUE;
   END IF;
 END $$;
+
+/* Remove these columns in scan_report because execution-task pattern will handle them */
+ALTER TABLE scan_report DROP COLUMN IF EXISTS job_id;
+ALTER TABLE scan_report DROP COLUMN IF EXISTS track_id;
+ALTER TABLE scan_report DROP COLUMN IF EXISTS requester;
+ALTER TABLE scan_report DROP COLUMN IF EXISTS status;
```
steven-zou, line 276: It seems `status` is still using.

steven-zou (commented):
comments left

src/controller/event/handler/webhook/scan/delete.go
```diff
@@ -48,32 +48,20 @@ func (o *DelArtHandler) Handle(value interface{}) error {
...
-		// Passed for safe consideration
-	} else {
-		if len(l) == 0 {
-			digests = append(digests, evt.Artifact.Digest)
-			log.Debugf("prepare to remove the scan report linked with artifact: %s", evt.Artifact.Digest)
+	} else if count == 0 {
+		log.Debugf("prepare to remove the scan report linked with artifact: %s", evt.Artifact.Digest)
 
+		if err := scan.DefaultController.DeleteReports(ctx, evt.Artifact.Digest); err != nil {
+			return errors.Wrap(err, "delete image event handler")
```
wy65701436, line 61: delete aritfact

src/controller/scan/base_controller.go
```diff
@@ -687,23 +713,111 @@ func (bc *basicController) launchScanJob(ctx context.Context, trackID string, ar
...
+	}
+
+	return nil
+}
+
+func getArtifactID(extraAttrs map[string]interface{}) int64 {
+	var artifactID float64
+	if extraAttrs != nil {
+		if v, ok := extraAttrs[artifactIDKey]; ok {
+			artifactID, _ = v.(float64) // int64 Unmarshal to float64
```
wy65701436, line 787: err?

src/core/api/scan_all.go
```diff
@@ -68,9 +67,57 @@ func (sc *ScanAllAPI) Post() {
 		sc.SendBadRequestError(err)
 		return
 	}
-	ajr.Name = common_job.ImageScanAllJob
-	sc.submit(&ajr)
-	sc.Redirect(http.StatusCreated, strconv.FormatInt(ajr.ID, 10))
+
```
wy65701436, line 70: the api will migrate in the new program model as well?

src/core/service/notifications/admin/handler.go
```diff
@@ -108,10 +107,7 @@ func (h *Handler) HandleAdminJob() {
 		return
 	}
 
-	// For scan all job
-	if h.jobName == job.ImageScanAllJob && h.checkIn != "" {
-		go scan.HandleCheckIn(orm.NewContext(context.TODO(), o.NewOrm()), h.checkIn)
-	} else if h.jobName == job.ImageGC && h.status == models.JobFinished {
+	if h.jobName == job.ImageGC && h.status == models.JobFinished {
```
wy65701436, line 110: we decide to deprecate the admin job in 2.2, will you address another PR to remove the admin job dependency?

src/pkg/scan/report/manager.go
```diff
@@ -79,34 +74,134 @@ type Manager interface {
...
+	if _, err := bm.dao.Create(ctx, r); err != nil {
+		return "", err
+	}
+
+	return r.UUID, nil
+}
+
+func (bm *basicManager) Delete(ctx context.Context, uuid string) error {
+	query := q.Query{Keywords: q.KeyWords{"uuid": uuid}}
+	count, err := bm.dao.DeleteMany(ctx, query)
```
wy65701436, line 140: what about count >1?

src/controller/scan/base_controller.go
```diff
@@ -233,78 +266,116 @@ func (bc *basicController) Scan(ctx context.Context, artifact *ar.Artifact, opti
 	return nil
 }
 
-func (bc *basicController) makeReportPlaceholder(ctx context.Context, r *scanner.Registration, art *ar.Artifact, options ...Option) (string, []string, error) {
-	trackID, err := bc.uuid()
+func (bc *basicController) ScanAll(ctx context.Context, trigger string, async bool) (int64, error) {
+	extraAttrs := map[string]interface{}{}
```
wy65701436, line 270: the async mode can put into the extraAttrs for later use.

src/controller/scan/base_controller.go
```diff
@@ -233,78 +266,116 @@ func (bc *basicController) Scan(ctx context.Context, artifact *ar.Artifact, opti
...
+	}
 
-		if len(producesMimes) > 0 {
-			return trackID, producesMimes, nil
+	count, err := bc.taskMgr.Count(ctx, q.New(q.KeyWords{"execution_id": executionID}))
+	if err != nil {
+		log.Errorf("failed to get the count of tasks for scan all, error %v", err)
+		return err
+	} else if count == 0 {
+		if err := bc.execMgr.MarkDone(ctx, executionID, "no task found"); err != nil {
```
wy65701436, line 317: why does the controller need to make exection done in its controller directly?

src/controller/scan/base_controller.go
```diff
@@ -233,78 +266,116 @@ func (bc *basicController) Scan(ctx context.Context, artifact *ar.Artifact, opti
...
 	}
 
-	err = create(ctx, art.Digest, r.UUID, v1.MimeTypeNativeReport, trackID, job.ErrorStatus)
-	return "", nil, err
+	return nil
 }
 
-func (bc *basicController) scanArtifact(ctx context.Context, r *scanner.Registration, artifact *ar.Artifact, trackID string, producesMimes []string) error {
-	jobID, err := bc.launchScanJob(ctx, trackID, artifact, r, producesMimes)
+func (bc *basicController) makeReportPlaceholder(ctx context.Context, r *scanner.Registration, art *ar.Artifact) ([]*scan.Report, error) {
```
wy65701436, line 326: add more comments for this method? is it for report migration?

src/controller/scan/base_controller.go
```diff
@@ -687,23 +713,111 @@ func (bc *basicController) launchScanJob(ctx context.Context, trackID string, ar
...
+	}
+
+	arr := make([]string, len(reportUUIDs))
+	for i, reportUUID := range reportUUIDs {
+		arr[i] = "'" + reportUUID + "'"
+	}
+
+	// condition looks like (extra_attrs->'report_uuids')::jsonb ?| array['uuid1', 'uuid2', 'uuid3']
+	// that condition means to select the task record when any element in the report_uuids of the extra_attr is in the set ('uuid1', 'uuid2', 'uuid3')
+	condition := fmt.Sprintf("(extra_attrs->'%s')::jsonb ?| array[%s]", reportUUIDsKey, strings.Join(arr, ","))
```
wy65701436, line 747: the condition is used for sql or not? Or it's just an format defination.

src/controller/scan/callback.go
```diff
@@ -0,0 +1,110 @@
...
+		if err != nil {
+			return err
+		}
+
+		if js == job.SuccessStatus {
+			robotID := getRobotID(t.ExtraAttrs)
+			if robotID > 0 {
+				if err := robot.Ctl.Delete(ctx, robotID); err != nil {
+					// Should not block the main flow, just logged
+					logger.WithFields(log.Fields{"robot_id": robotID, "error": err}).Error("delete robot account failed")
```
wy65701436, line 69: how can we handle the dirty robot when the deletion is failed.

src/controller/scan/base_controller.go
heww: I don't think we need this. `async` here is only to let's test the `ScanAll` method easier.

src/controller/scan/base_controller.go
heww: Because there is no notification to change the status of the execution to be done when there is no task related to it so we need to mark the execution done ourselves.

src/controller/scan/base_controller.go
heww: This condition applies a raw filter to the querysetter of task which make the SQL looks like

```sql
select * from task where id in (select id from task where (extra_attrs->'report_uuids')::jsonb ?| array['uuid1', 'uuid2']
```

make/migrations/postgresql/0050_2.2.0_schema.up.sql
heww: `Status` is still kept in the model of `ScanReport` but it will be filled from the status of the task mode.

```golang
Status    string    `orm:"-"`
StartTime time.Time `orm:"-"`
EndTime   time.Time `orm:"-"`
```

src/controller/event/handler/init.go
heww: This is used let's test the `DelArthandler` with mocks. 

`chart.Handler`, `p2p.Handler` and `internal.Handler` use this method to write the testcase.

src/controller/event/handler/webhook/scan/delete.go
heww: When the artifact was deleted, the report for the digest of the artifact will be deleted when the digest does not exist in the artifact table.

src/core/api/scan_all.go
heww: This may in the next PR.

src/core/service/notifications/admin/handler.go
heww: I think`quota.RefreshForProjects(orm.NewContext(context.TODO(), o.NewOrm()))` needs to be changed in the refractor of gc.

And remove the admin job in another PR.

src/pkg/scan/report/manager.go
heww: `count > 1` will never happen because `uuid` is unique in the table.

src/pkg/task/model.go
```diff
@@ -76,6 +89,16 @@ type Task struct {
 	EndTime    time.Time `json:"end_time"`
 }
 
+// FilterByExtraAttrsElement returns orm.QuerySeter with extra_attrs element filter
+func (t *Task) FilterByExtraAttrsElement(ctx context.Context, qs orm.QuerySeter, key string, value interface{}) orm.QuerySeter {
```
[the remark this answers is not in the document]
heww: This method does not work in FilterRaw, will be removed.

src/controller/scan/base_controller.go
heww: This value must be the float64, so err does not need here.

src/controller/scan/base_controller.go
heww: It's not the report migration. For every artifact will be scanned, we will create a scan report for it before send the scan task to the jobservice.

src/controller/scan/base_controller.go
heww: Beego not support this way in the FilterRaw, this will be removed.

src/controller/scan/base_controller.go
heww: Mark the execution to be done here is ok.

The scan method returns the error only when the artifact is not supported by the scanner or the artifact is scanning by the scanner.

What is the status of the execution when 100 artifacts found and 99 of them scan failed and 1 success? According to the status change handler, is the status of the execution success?

src/controller/scan/base_controller.go
heww: `async` was removed from the parameters of the `startScanAll`

src/controller/scan/base_controller.go
ywk253100: If the execution contains failed tasks, it is marked as failed

ywk253100 (approved):

wy65701436 (approved):
lgtm

heww:
I will merge this PR, the status for the execution of the scan all will be in another PR when the update needed.

PR #13685, opened 2020-12-03
fix robot name conflicate issue

wy65701436:
add project name into project level robot account name

### DIFF
--- a/src/controller/robot/controller.go
+++ b/src/controller/robot/controller.go
@@ -76,7 +76,7 @@ func (d *controller) Count(ctx context.Context, query *q.Query) (int64, error) {
 
 // Create ...
 func (d *controller) Create(ctx context.Context, r *Robot) (int64, string, error) {
-	if err := d.setProjectID(ctx, r); err != nil {
+	if err := d.setProject(ctx, r); err != nil {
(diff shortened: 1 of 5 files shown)


heww (approved):

src/controller/robot/controller.go
```diff
@@ -273,24 +275,19 @@ func (d *controller) populatePermissions(ctx context.Context, r *Robot) error {
...
+func (d *controller) setProject(ctx context.Context, r *Robot) error {
 	if r == nil {
 		return nil
 	}
-	var projectID int64
-	switch r.Level {
-	case LEVELSYSTEM:
-		projectID = 0
-	case LEVELPROJECT:
+	if r.Level == LEVELPROJECT {
```
stonezdj, line 283: When the r.Level is wrong, there is no error. and the robot account can be created.  Is it an expected behaviour?

wy65701436: For the input checking, all of them are done by API handler. And this method is just to set the project information, if it's a system level, the project ID will be 0, it's a expected behavior.

stonezdj (approved):

PR #13690, opened 2020-12-04
Move proxy cache e2e test into a new CI job

danfengliu:
CI job timeout is 60 minutes, but proxy cache case timeout is 20 minutes, once
proxy cache case is timeout, it will reach CI job timeout, then cause CI job terminated
by timeout without running rest of test steps, expecially uploading useful harbor logs,
so move this test into a new clean job for saving time and debugging.


AllForNothing (approved):

wy65701436 (approved):
lgtm

PR #13709, opened 2020-12-09
Cache manifest list for proxy cache

### LINKED ISSUE #13566: Quota of dockerhub is still used in v2.1.1 after the image is cached.
Was reported by Remoco Breekveldt from slack channel:

1. Harbor Version v2.1.1-5f52168e with Proxy cache configured on example.com/docker-hub
2. Enter the Harbor Core pod and check the remaing rate limit with the following commands:
```
export HEADER='Authorization: Bearer'
export TOKEN=$(curl "https://auth.docker.io/token?service=registry.docker.io&scope=repository:ratelimitpreview/test:pull" | sed 's/{"token":"//g' | sed 's/","access_token":.*//g') 
curl --head -H "$HEADER $TOKEN" https://registry-1.docker.io/v2/ratelimitpreview/test/manifests/latest 2>&1
```
Remaining-limit is showing 250
3. From a local machine loop a pull for busybox using the following command (note: ratelimit is 250 today):
for i in {1..255}; do docker pull example.com/docker-hub/library/busybox; done
4. Enter the Harbor Core pod and check the remaing rate limit with the following commands:
```
export HEADER='Authorization: Bearer'
export TOKEN=$(curl "https://auth.docker.io/token?service=registry.docker.io&scope=repository:ratelimitpreview/test:pull" | sed 's/{"token":"//g' | sed 's/","access_token":.*//g') 
curl --head -H "$HEADER $TOKEN" https://registry-1.docker.io/v2/ratelimitpreview/test/manifests/latest 2>&1
```
5. We noticed that we consumed all our rate-limit and actually the remaining limit is 'minus' 5

stonezdj:
Fixes #13566: Quota of dockerhub is still used in v2.1.1 after the image is cached
Cache manifest list in redis cache

### DIFF
--- a/src/controller/proxy/controller.go
+++ b/src/controller/proxy/controller.go
@@ -29,6 +29,7 @@ import (
 	"github.com/goharbor/harbor/src/controller/blob"
 	"github.com/goharbor/harbor/src/controller/event/operator"
 	"github.com/goharbor/harbor/src/lib"
+	"github.com/goharbor/harbor/src/lib/cache"
 	"github.com/goharbor/harbor/src/lib/errors"
 	"github.com/goharbor/harbor/src/lib/log"
 	"github.com/goharbor/harbor/src/lib/orm"
@@ -41,6 +42,8 @@ const (
 	maxManifestListWait = 20
 	maxManifestWait     = 10
 	sleepIntervalSec    = 20
+	// keep manifest list in cache for one week
+	manifestListCacheIntervalSec = 7 * 24 * 60 * 60
 )
 
 var (
@@ -54,7 +57,7 @@ type Controller interface {
 	// UseLocalBlob check if the blob should use local copy
 	UseLocalBlob(ctx context.Context, art lib.ArtifactInfo) bool
 	// UseLocalManifest check manifest should use local copy
-	UseLocalManifest(ctx context.Context, art lib.ArtifactInfo, remote RemoteInterface) (bool, error)
+	UseLocalManifest(ctx context.Context, art lib.ArtifactInfo, remote RemoteInterface) (bool, *ManifestList, error)
 	// ProxyBlob proxy the blob request to the remote server, p is the proxy project
 	// art is the ArtifactInfo which includes the digest of the blob
 	ProxyBlob(ctx context.Context, p *models.Project, art lib.ArtifactInfo) (int64, io.ReadCloser, error)
@@ -68,6 +71,7 @@ type controller struct {
 	blobCtl     blob.Controller
 	artifactCtl artifact.Controller
 	local       localInterface
+	cache       cache.Cache
 }
 
 // ControllerInstance -- Get the proxy controller instance
@@ -79,6 +83,7 @@ func ControllerInstance() Controller {
 			blobCtl:     blob.Ctl,
 			artifactCtl: artifact.Ctl,
 			local:       newLocalHelper(),
+			cache:       cache.Default(),
 		}
 	})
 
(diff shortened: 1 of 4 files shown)


heww (approved):

src/controller/proxy/controller.go
```diff
@@ -209,6 +236,19 @@ func (c *controller) putBlobToLocal(remoteRepo string, localRepo string, desc di
 }
 
 func (c *controller) waitAndPushManifest(ctx context.Context, remoteRepo string, man distribution.Manifest, art lib.ArtifactInfo, contType string, r RemoteInterface) {
+	if contType == manifestlist.MediaTypeManifestList {
+		_, payload, err := man.Payload()
+		if err != nil {
+			log.Errorf("failed to get payload, error %v", err)
+		}
+		dig := digest.FromBytes([]byte(payload))
```
reasonerjt, line 244: This digest will be used to compare with the digest in the header in the response from remote.

It will be broken if the remote site, such as dockerhub use a different way to calculate the digest.

How about using FetchOrSave when fetching the cache, so the digest is the value from the response?

src/server/middleware/repoproxy/proxy.go
```diff
@@ -106,14 +106,25 @@ func handleManifest(w http.ResponseWriter, r *http.Request, next http.Handler) e
...
+
 	if err != nil {
 		return err
 	}
 	if useLocal {
+		if man != nil {
+			w.Header().Set("Content-Type", man.ContentType)
+			w.Header().Set("Docker-Content-Digest", man.Digest)
+			w.Header().Set("Etag", man.Digest)
+			w.Write(man.Content)
```
reasonerjt, line 119: You will return the content even if it's a HEAD request?

src/controller/proxy/controller.go
```diff
@@ -96,33 +101,55 @@ func (c *controller) UseLocalBlob(ctx context.Context, art lib.ArtifactInfo) boo
...
+
+	var content []byte
+	if c.cache != nil {
+		err = c.cache.Fetch(getManifestListKey(art.Repository, dig), &content)
+		if err == nil {
+			log.Debugf("Get the manifest list with key=cache:%v", getManifestListKey(art.Repository, dig))
+			return true, &ManifestList{content, dig, manifestlist.MediaTypeManifestList}, nil
+		}
+		if err == cache.ErrNotFound {
+			log.Debugf("Digest is not found in manifest list cache, key:cache:%v", getManifestListKey(art.Repository, dig))
```
reasonerjt, line 141: This is trivial but why key:cache:${key}?

src/controller/proxy/controller.go
```diff
@@ -96,33 +101,55 @@ func (c *controller) UseLocalBlob(ctx context.Context, art lib.ArtifactInfo) boo
...
 		}()
-		return false, errors.NotFoundError(fmt.Errorf("repo %v, tag %v not found", art.Repository, art.Tag))
+		return false, nil, errors.NotFoundError(fmt.Errorf("repo %v, tag %v not found", art.Repository, art.Tag))
+	}
+
+	var content []byte
+	if c.cache != nil {
+		err = c.cache.Fetch(getManifestListKey(art.Repository, dig), &content)
+		if err == nil {
+			log.Debugf("Get the manifest list with key=cache:%v", getManifestListKey(art.Repository, dig))
```
reasonerjt, line 137: This is trivial but why key=cache:${key} instead of key=${key}?

src/controller/proxy/controller.go
```diff
@@ -209,6 +236,19 @@ func (c *controller) putBlobToLocal(remoteRepo string, localRepo string, desc di
...
 
 func (c *controller) waitAndPushManifest(ctx context.Context, remoteRepo string, man distribution.Manifest, art lib.ArtifactInfo, contType string, r RemoteInterface) {
+	if contType == manifestlist.MediaTypeManifestList {
+		_, payload, err := man.Payload()
+		if err != nil {
+			log.Errorf("failed to get payload, error %v", err)
+		}
+		dig := digest.FromBytes([]byte(payload))
+		key := getManifestListKey(art.Repository, string(dig))
+		log.Debugf("Cache manifest list with key=cache:%v", key)
```
reasonerjt, line 246: see my comment above, the log message is confusing with what's the key in cache.

reasonerjt:
You should add more commit message because this change may look weird, it returns something that is not consistent with Harbor's storage.

You should also verify how this impacts the API, including the /v2/ API and Harbor's API.  And specify in the git message.

src/controller/proxy/controller.go
stonezdj: This is the actual key in the redis

src/server/middleware/repoproxy/proxy.go
stonezdj: Updated

src/controller/proxy/controller.go

src/controller/proxy/controller.go
```diff
@@ -209,6 +239,18 @@ func (c *controller) putBlobToLocal(remoteRepo string, localRepo string, desc di
 }
 
 func (c *controller) waitAndPushManifest(ctx context.Context, remoteRepo string, man distribution.Manifest, art lib.ArtifactInfo, contType string, r RemoteInterface) {
+	if contType == manifestlist.MediaTypeManifestList {
+		_, payload, err := man.Payload()
+		if err != nil {
+			log.Errorf("failed to get payload, error %v", err)
```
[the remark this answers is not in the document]
stonezdj: OK

reasonerjt (approved):

PR #13724, opened 2020-12-10
Support admin group in http authproxy

reasonerjt:
This commit adds admin_groups into the configuration of http_auth
settings, it's a string in the form of "group1, group2".  If the token
review result shows the user is in one of the groups in the setting he
will have the administrator role in Harbor.


wy65701436 (approved):
lgtm

ywk253100 (approved):

PR #13756, opened 2020-12-14
add permission checking for gc operation

wy65701436:

### DIFF
--- a/src/server/v2.0/handler/gc.go
+++ b/src/server/v2.0/handler/gc.go
@@ -28,6 +28,13 @@ func newGCAPI() *gcAPI {
 	}
 }
 
+func (g *gcAPI) Prepare(ctx context.Context, operation string, params interface{}) middleware.Responder {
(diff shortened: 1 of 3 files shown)


src/server/v2.0/handler/gc.go
```diff
@@ -28,6 +28,13 @@ func newGCAPI() *gcAPI {
 	}
 }
 
+func (g *gcAPI) Prepare(ctx context.Context, operation string, params interface{}) middleware.Responder {
+	if err := g.RequireSysAdmin(ctx); err != nil {
+		g.SendError(ctx, err)
```
heww, line 33: should be `return g.SendError(ctx, err)`

ywk253100 (approved):

src/server/v2.0/handler/gc.go
wy65701436: both are fine, but updated.

heww (approved):

PR #13790, opened 2020-12-17
fix gc log not found issue

wy65701436:
It needs to use the execution ID to get task firstly and then use the required task id to query GC log

### DIFF
--- a/src/server/v2.0/handler/gc.go
+++ b/src/server/v2.0/handler/gc.go
@@ -8,6 +8,7 @@ import (
 	"github.com/goharbor/harbor/src/controller/gc"
 	"github.com/goharbor/harbor/src/core/config"
 	"github.com/goharbor/harbor/src/lib/errors"
+	"github.com/goharbor/harbor/src/lib/q"
 	"github.com/goharbor/harbor/src/pkg/scheduler"
 	"github.com/goharbor/harbor/src/pkg/task"
 	"github.com/goharbor/harbor/src/server/v2.0/handler/model"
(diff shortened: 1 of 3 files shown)


ywk253100 (approved):

src/server/v2.0/handler/gc.go
```diff
@@ -197,7 +197,19 @@ func (g *gcAPI) GetGC(ctx context.Context, params operation.GetGCParams) middlew
...
+	}
+	query.Keywords["ExecutionID"] = params.GcID
+	tasks, err := g.gcCtr.ListTasks(ctx, query)
+	if err != nil {
+		return g.SendError(ctx, err)
+	}
+	if len(tasks) == 0 {
+		return g.SendError(ctx, errors.New(nil).WithCode(errors.NotFoundCode).WithMessage("garbage collection %d log is not found", params.GcID))
+	}
+	log, err := g.gcCtr.GetTaskLog(ctx, tasks[0].ID)
```
heww, line 212: How about removing GetTaskLog and ListTasks and implement `GetGCLog(ctx context.Context, executionID int64)` in the gc controller ?

wy65701436: as discussed with @ywk253100 , we should not provide specific method in the GC handler, so that it can be easy to migrate the new GC api in furture.

heww: But, I think GetTaskLog, is the method of the task controller/manager.

src/server/v2.0/handler/gc.go
```diff
@@ -197,7 +197,19 @@ func (g *gcAPI) GetGC(ctx context.Context, params operation.GetGCParams) middlew
 }
 
 func (g *gcAPI) GetGCLog(ctx context.Context, params operation.GetGCLogParams) middleware.Responder {
-	log, err := g.gcCtr.GetTaskLog(ctx, params.GcID)
+	query, err := g.BuildQuery(ctx, nil, nil, nil)
+	if err != nil {
+		return g.SendError(ctx, err)
+	}
+	query.Keywords["ExecutionID"] = params.GcID
```
heww, line 204: Please use `q.New()` to make the query.

heww (approved):

src/server/v2.0/handler/gc.go
wy65701436: resolved

Issue #13364, opened 2020-10-26
Broken replication of the "<registry URL>/<namespace>/<project>/<image>" docker images


blackprowler:
Hello Harbor team,

Starting from the Harbor v2.1.0 the replication of the docker images that have name `<registry URL>/<namespace>/<project>/<image>:tag` is broken. Such images can't be replicated.
It also seems like in such cases replication process drains all the available CPU and RAM resources leading to DoS.

Harbor v2.0.3 doesn't have this issue.

Images that have `<registry URL>/<namespace>/<project>:tag` name replicate without any issues.

It is also possible to push `<registry URL>/<namespace>/<project>/<image>:tag` image directly to Harbor without any issues.

### Expected behavior and actual behavior:
This issue has been faced wile setting up a replication from the GitLab Container Registry (GLCR) to the Harbor one.

Some of the docker images stored in the GLCR have name similar to the following:
`<registry URL>/<namespace>/<project>/<image>:tag`

E.g.,
Image to replicate (GLCR):`registry.example.com/group/project/image:latest`
This project needs to be replicated into the following Harbor project: `<Harbor registry URL>/<namespace>/<project>/<image>:tag`

#### Expected behavior
Harbor successfully does the pull-cased replication of the Docker image `<registry URL>/<namespace>/<project>/<image>:tag` to a `<Harbor registry URL>/<namespace>/<project>/<image>:tag` without any issues.

#### Actual behavior
Harbor fails the replication (Error) of the Docker image `<registry URL>/<namespace>/<project>/<image>:tag` to a `<Harbor registry URL>/<namespace>/<project>/<image>:tag`.

It is also looks like this causes DoS for the Harbor node as all the CPU and RAM resources are drained.

### Steps to reproduce the problem:
1. Create a GitLab Project in some Group and push into it a Docker image with the following name `<registry URL>/<namespace>/<project>/<image>:tag`. I assume might also be true for other Docker Registries.
2. Setup a replication in Harbor to replicate `<namespace>/**` from GitLab to a Harbor.
3. Run the replication.
4. Replication will fail without having `<registry URL>/<namespace>/<project>/<image>:tag`.
5. If many projects should be replicated, the replication of the `<registry URL>/<namespace>/<project>/<image>:tag` image will cause DoS for the Harbor.

### Versions:
Please specify the versions of following systems.

- harbor version: [v2.1.0]

#### Harbor error log
```
[...]

PR #13797, opened 2020-12-17
GitLab replication adapter: improve projects search

hlubek:
Adds a query parameter to search in ancestor namespaces which should
find projects in subgroups matching a name filter component.

### DIFF
--- a/src/replication/adapter/gitlab/client.go
+++ b/src/replication/adapter/gitlab/client.go
@@ -100,7 +100,7 @@ func (c *Client) getProjects() ([]*Project, error) {
 
 func (c *Client) getProjectsByName(name string) ([]*Project, error) {
 	var projects []*Project
-	urlAPI := fmt.Sprintf("%s/api/v4/projects?search=%s&membership=1&per_page=50", c.url, name)
+	urlAPI := fmt.Sprintf("%s/api/v4/projects?search=%s&search_namespaces=1&membership=1&per_page=50", c.url, name)
 	if err := c.GetAndIteratePagination(urlAPI, &projects); err != nil {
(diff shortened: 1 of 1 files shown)


hlubek:
Pull request created as mentioned in https://github.com/goharbor/harbor/issues/13364#issuecomment-746119483

bitsf (approved):
As I can't contact with the original developer for this gitlab adapter, so I suppose your PR is already tested for many cases and worked well

hlubek:
@bitsf I couldn't test all different scenarios with GitLab of course, since I only verified it against our instance (recent version) with lots of projects in different hierarchies. Reading the GitLab source code the `search_namespaces` parameter should only return more matches for projects (which are later filtered out by the adapter according to the name filter) - so it should not be worse than the current solution.

As written in #13364 the solution of using a (fuzzy) search for projects is not very robust and should be replaced by iterating over groups/projects via the GitLab API. I can have a look at that soon - but for now this fix would at least improve things.

bitsf:
that would be quite helpful if someone can figure out how gitlab API should work well.

bitsf:
hi @hlubek , I test this, we should not use `search_namespaces`, as it will also filter projects with namespace matched, not name matched
such as repo 'test/abc/repo1' and 'user/test/repo2', if filter with 'user/test/*', the two projects will be filtered both.

I think the issue that you faced is because param `membership=1` which 'Limit projects that the current user is a member of'
So you should add the user to all projects needed.

bitsf:
oh, I just understand your problem, I didn't notice gitlab can create multi level sub groups, I thought it can only has user as first level

PR #13813, opened 2020-12-20
refactor(api): move scan all apis to go-swagger

heww:
Move scan all APIs from beego to go-swagger.


danfengliu (approved):
LGTM

src/server/v2.0/handler/scan_all.go
```diff
@@ -0,0 +1,248 @@
...
+	query := q.New(q.KeyWords{"vendor_type": job.ImageScanAllJob})
+	schedules, err := s.scheduler.ListSchedules(ctx, query.First("-creation_time"))
+	if err != nil {
+		return nil, err
+	}
+
+	if len(schedules) > 1 {
+		return nil, fmt.Errorf("found more than one scheduled scan all job, please ensure that only one schedule left")
+	} else if len(schedules) == 0 {
+		return nil, nil
```
wy65701436, line 182: return ErrNotFound?

wy65701436 (approved):
lgtm

src/server/v2.0/handler/scan_all.go
heww: No need to return `ErrNotFound` here because get schedule API doesn't return the 404 when the schedule not found. It's really confusing.

ywk253100 (approved):

PR #13824, opened 2020-12-22
Fix docker20 policy check

wy65701436:
For docker client v20, it sends the head manifest rather than get manifest during pull image, to make sure the vul and ct middleware work,
it has to intercept the head manifest request.


reasonerjt:
I think we only need to block the `GET` request, after `HEAD` the client should issue a `GET` request against the digest, we should make sure that request is blocked as configured.

reasonerjt (changes requested):
I don't think this is right.

wy65701436:
after clean imags and restart docker & containerd, it works.

PR #13826, opened 2020-12-22
add system resource

wy65701436:
enable robot to have the system access scope


src/common/security/robot/context.go
```diff
@@ -32,16 +35,18 @@ type SecurityContext struct {
 	isSystemLevel bool
 	ctl           project.Controller
 	policy        []*types.Policy
-	evaluator     evaluator.Evaluator
+	proEvaluator  evaluator.Evaluator
+	sysEvaluator  evaluator.Evaluator
```
heww, line 39: No need these two evaluators here. evaluator.Evaluator is an interface, there is an evaluator.Evaluators which can include multi evaluators.

src/common/security/robot/context.go
```diff
@@ -78,13 +83,30 @@ func (s *SecurityContext) IsSolutionUser() bool {
...
-			s.evaluator = rbac.NewProjectEvaluator(s.ctl, rbac.NewBuilderForPolicies(s.GetUsername(), s.policy, filterRobotPolicies))
+			s.proEvaluator = rbac.NewProjectEvaluator(s.ctl, rbac.NewBuilderForPolicies(s.GetUsername(), s.policy, filterRobotPolicies))
 		}
 	})
 
-	return s.evaluator != nil && s.evaluator.HasPermission(ctx, resource, action)
+	if s.isSystemLevel {
+		return (s.sysEvaluator != nil && s.sysEvaluator.HasPermission(ctx, resource, action)) || (s.proEvaluator != nil && s.proEvaluator.HasPermission(ctx, resource, action))
+	}
+	return s.proEvaluator != nil && s.proEvaluator.HasPermission(ctx, resource, action)
```
heww, line 109:
```golang
if s.isSystemLevel {
    var evaluators evaluator.Evaluators
    if len(sysPolicies) > 0 {
        evaluators = evaluators.Add(system.NewEvaluator(s.GetUsername(), sysPolicies))
    }
    if len(proPolicies) > 0 {
        evaluators = evaluators.Add(rbac.NewProjectEvaluator(s.ctl, rbac.NewBuilderForPolicies(s.GetUsername(), proPolicies)))
    }

    s.evaluator = evaluators

    // or check the len of evaluators
    // if len(evaluators) > 1 {
    //   s.evaluator = evaluators
    // } else if len(evaluators) == 1 {
    //  s.evaluator = evaluators[0]    
    // }
} else {
    s.evaluator = rbac.NewProjectEvaluator(s.ctl, rbac.NewBuilderForPolicies(s.GetUsername(), s.policy, filterRobotPolicies))
} 
```

src/server/v2.0/handler/gc.go
```diff
@@ -30,13 +36,19 @@ func newGCAPI() *gcAPI {
...
 
 func (g *gcAPI) Prepare(ctx context.Context, operation string, params interface{}) middleware.Responder {
-	if err := g.RequireSysAdmin(ctx); err != nil {
-		return g.SendError(ctx, err)
+	secCtx, ok := security.FromContext(ctx)
+	if !ok {
+		return g.SendError(ctx, errors.UnauthorizedError(errors.New("security context not found")))
 	}
+	g.secCtx = secCtx
+	g.resource = system.NewNamespace().Resource(rbac.ResourceGarbageCollection)
```
heww, line 44: We can't assign the var to the gcAPI because the instance of gcAPI is shared for all API request.

heww: Suggest to add a method `RequireSystemAccess` in the BaseAPI to handle the system permission checking.

heww: Because now the operations of the GC use different permission actions so I think we need to remove the permission checking in the Prepare. Every operation uses the `RequireSystemAccess` to check permission.

src/common/security/robot/context.go
wy65701436: OK, do not know there is an evaluator.Evaluators, will use it.

src/common/security/robot/context.go
wy65701436: will use evaluator.Evaluators for instead

src/server/v2.0/handler/gc.go
wy65701436: do no know this either, ok will have a update.

src/server/v2.0/handler/auditlog.go
```diff
@@ -44,7 +43,7 @@ func (a *auditlogAPI) ListAuditLogs(ctx context.Context, params auditlog.ListAud
 		return a.SendError(ctx, err)
 	}
 
-	if !secCtx.IsSysAdmin() {
+	if err := a.RequireSysAdminAccess(ctx, rbac.ActionList, rbac.ResourceAuditLog); err != nil {
```
heww, line 46: Should only accept ForbiddenError, other errors need to return.

wy65701436: no logic change here, and this API requires authentication. 

Discussed offline, not a problem.

heww (approved):

src/server/v2.0/handler/base.go
```diff
@@ -101,8 +102,8 @@ func (b *BaseAPI) RequireProjectAccess(ctx context.Context, projectIDOrName inte
 	return errors.ForbiddenError(nil)
 }
 
-// RequireSysAdmin checks the system admin permission according to the security context
-func (b *BaseAPI) RequireSysAdmin(ctx context.Context) error {
+// RequireSysAdminAccess checks the system admin permission according to the security context
```
reasonerjt, line 105: I think to make the model more consistent we should treat sys admin equally.  i.e. check the permission regardless if the secCtx is sysadmin or not.

So the func should be `RequireAccess` and `if !secCtx.IsSysAdmin() {` will be redundant?

Could you please also consider to refactor the `HasPermission` and `HasProjectPermission` to make things more consistent and simplified.

reasonerjt (approved):

reasonerjt:
Approved with some minor comments, please also consider clean up the package `common/rbac` to move project based stuff to `common/rbac/project`

src/server/v2.0/handler/base.go
wy65701436: resolved.

PR #13846, opened 2020-12-25
Disallow developer role to delete artifacts via registry API

ywk253100:
Disallow developer role to delete artifacts via registry API


reasonerjt:
I checked the code I think `permToActions` should not be needed, and we should also refine the testcase to verify such problem.

Issue #13780, opened 2020-12-16
LDAP authentication delay


alexandrovas:
We are using LDAP auth with FreeIPA. While authentication I see this warnings in log file core.log:
```
Dec 16 14:48:25 172.29.0.1 core[1462]: 2020-12-16T11:48:25Z [WARNING] [/core/auth/ldap/ldap.go:121]: Can not get the ldap group name with DN cn=ipausers,cn=groups,cn=accounts,dc=mydomain,dc=local
Dec 16 14:48:25 172.29.0.1 core[1462]: 2020-12-16T11:48:25Z [WARNING] [/core/auth/ldap/ldap.go:121]: Can not get the ldap group name with DN cn=role_sysadmins,cn=groups,cn=accounts,dc=mydomain,dc=local
Dec 16 14:48:26 172.29.0.1 core[1462]: 2020-12-16T11:48:26Z [WARNING] [/core/auth/ldap/ldap.go:121]: Can not get the ldap group name with DN cn=admins,cn=groups,cn=accounts,dc=mydomain,dc=local
Dec 16 14:48:26 172.29.0.1 core[1462]: 2020-12-16T11:48:26Z [WARNING] [/core/auth/ldap/ldap.go:121]: Can not get the ldap group name with DN cn=Replication Administrators,cn=privileges,cn=pbac,dc=mydomain,dc=local
```
And authentication takes several seconds. This behavior sometimes leads to timeout in docker client.

This groups are not in filter scope and should not be queried.

Old version Harbor 2.0.1 with identical config authenticates instantly and no same warnings in log.

**Expected behavior and actual behavior:**
LDAP auth without delay and timeout in docker client.

**Steps to reproduce the problem:**
1. Install Harbor 2.1.2
2. Setup Ldap authentication
3. Login in Harbor via docker client (or Harbor GUI)
4. See log file core.log

**Versions:**

- harbor version: 2.1.2

**Additional context:**

My config in Harbor:
```json
{
  "auth_mode": {
    "value": "ldap_auth",
    "editable": false
  },
  "count_per_project": {
    "value": -1,
    "editable": true
  },
  "email_from": {
    "value": "admin <sample_admin@mydomain.com>",
    "editable": true
  },
  "email_host": {
    "value": "smtp.mydomain.com",
    "editable": true
  },
  "email_identity": {
    "value": "",
    "editable": true
  },
  "email_insecure": {
    "value": false,
    "editable": true
  },
  "email_port": {
    "value": 25,
[...]

PR #13851, opened 2020-12-25
Optimization of LDAP group membership processing for 2.1.0

alexandrovas:
Here is the fix for issue https://github.com/goharbor/harbor/issues/13780 that I have opened recently.
It optimizes LDAP group processing and reduces amount of LDAP queries thus reduces authentication delay.

### DIFF
--- a/src/common/utils/ldap/ldap.go
+++ b/src/common/utils/ldap/ldap.go
@@ -317,6 +317,7 @@ func (session *Session) SearchLdapAttribute(baseDN, filter string, attributes []
 
 	var result *goldap.SearchResult
 	var err error
+
 	if session.ldapConfig.LdapPageSize > 0 {
 		result, err = session.ldapConn.SearchWithPaging(searchRequest, uint32(session.ldapConfig.LdapPageSize))
 	} else {
--- a/src/common/utils/ldap/ldap.go
+++ b/src/common/utils/ldap/ldap.go
@@ -317,7 +317,6 @@ func (session *Session) SearchLdapAttribute(baseDN, filter string, attributes []
 
 	var result *goldap.SearchResult
 	var err error
-
 	if session.ldapConfig.LdapPageSize > 0 {
 		result, err = session.ldapConn.SearchWithPaging(searchRequest, uint32(session.ldapConfig.LdapPageSize))
 	} else {
--- a/src/common/utils/ldap/ldap.go
+++ b/src/common/utils/ldap/ldap.go
@@ -315,7 +315,7 @@ func (session *Session) SearchLdapAttribute(baseDN, filter string, attributes []
 		nil,
 	)
 
-	result, err := session.ldapConn.SearchWithPaging(searchRequest, uint32(session.ldapConfig.LDAPPageSize))
+	result, err := session.ldapConn.SearchWithPaging(searchRequest, uint32(session.ldapConfig.LdapPageSize))
 	if result != nil {
 		log.Debugf("Found entries:%v\n", len(result.Entries))
 	} else {
--- a/src/common/utils/ldap/ldap.go
+++ b/src/common/utils/ldap/ldap.go
@@ -315,7 +315,7 @@ func (session *Session) SearchLdapAttribute(baseDN, filter string, attributes []
 		nil,
 	)
 
-	result, err := session.ldapConn.Search(searchRequest)
+	result, err := session.ldapConn.SearchWithPaging(searchRequest, uint32(session.ldapConfig.LDAPPageSize))
 	if result != nil {
 		log.Debugf("Found entries:%v\n", len(result.Entries))
 	} else {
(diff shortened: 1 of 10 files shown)


alexandrovas:
This is my first PR. Please help to pass all the necessary checks.

sluetze:
> This is my first PR. Please help me to pass all the necessary checks.

Hi, you can follow the guidelines here: https://github.com/goharbor/harbor/blob/master/CONTRIBUTING.md

not one of the maintainers, but what needs to be done from my understanding:
- rebase the commits on the master branch and file a PR against master. (https://github.com/goharbor/harbor/blob/master/CONTRIBUTING.md#branch)
- squash your commits and sign them you can do so with interactive rebase (https://thoughtbot.com/blog/git-interactive-rebase-squash-amend-rewriting-history#squash-commits-together)

this should get DCO and CI/ UTTEST on green.  codecov must not be green

src/common/utils/ldap/ldap.go
```diff
@@ -432,6 +432,41 @@ func (session *Session) searchGroup(groupDN, filter, gName, groupNameAttribute s
...
+	// Search all groups with LDAP group filter condition
+	ldapFilter, err := createGroupSearchFilter(session.ldapGroupConfig.LdapGroupFilter, "", session.ldapGroupConfig.LdapGroupNameAttribute)
+	if err != nil {
+		log.Errorf("wrong filter format: filter:%v, groupNameAttribute:%v", session.ldapGroupConfig.LdapGroupFilter, session.ldapGroupConfig.LdapGroupNameAttribute)
+		return ldapGroups, err
+	}
+
+	// There maybe many groups under the LDAP group base DN
+	// If return all groups in LDAP group base DN, it might get "Size Limit Exceeded" error
+	result, err := session.SearchLdapAttribute(session.groupBaseDN(), ldapFilter, []string{session.ldapGroupConfig.LdapGroupNameAttribute})
```
stonezdj, line 448: Could you please handle the case when "Size Limit Exceeded" error?

alexandrovas: I think this case is not possible because the SizeLimit parameter in SearchRequest is 0:
https://github.com/goharbor/harbor/blob/5d9a01152acbbdb7c2d06a88b1f7f3161f07d72a/src/common/utils/ldap/ldap.go#L310

alexandrovas: Moreover, this error should theoretically be handled in this code:
https://github.com/goharbor/harbor/blob/5d9a01152acbbdb7c2d06a88b1f7f3161f07d72a/src/common/utils/ldap/ldap.go#L318

alexandrovas:
@stonezdj 
Any other comments on this PR?

src/common/utils/ldap/ldap.go
DASXCE: It is possible! The "Size Limit Exceeded" is an error returned by the LDAP Server. This was fixed in v2.1.1, i think, with https://github.com/goharbor/harbor/blob/a211b0c9d79b85368bf8c6c4c340e0ceb2cfc70e/src/common/utils/ldap/ldap.go#L411-L414

alexandrovas: I'm sorry. You are really right
For such a case, we must implement a request with pagination in LDAP. Page Size should be configured in Authentication page.

@DASXCE
Can you help me implement this?

alexandrovas: @DASXCE @stonezdj
I implemented pagination for LDAP queries. This should help in case of the "Size Limit Exceeded" error.

DASXCE:
@alexandrovas thanks for the pagination! 

I am not a maintainer but i would like to propose an additional optimization. In my environment the `SearchAllGroups`  would still take 10~15 seconds as there are ~45k groups.

As I understand the current code... groups are queried only to get the group name attribute which IMHO is not that necessary since we have the group DN.

Proposal:
If the `LDAP Group GID` field, from the UI configuration page, is blank (currently it defaults to `cn` ) then skip the group search completely and build the `models.LdapGroup` without the `GroupName` field OR default to extracting the CN value from the group DN. In case the `LDAP Group GID` is set then proceed with searching for the group attribute.

I would love to provide a commit for this but i am currently not proficient with Go. Looking forward for a feedback!

alexandrovas:
@DASXCE 

Hello!
You are not entirely right. You suggest taking group name from DN without additional queries in case when GID is CN (by default). But I'm not sure if this is always possible.
I don’t know all the behavior of various LDAP servers, but what if the value of the CN field is not contained in the group DN? In other words, what if in some LDAP schemes we cannot get the group name directly from CN? IN such case the code will crash?
By adding this kind of logic, we can break compatibility with some LDAP schemes.

To solve these problems, many applications add the ability to cache LDAP data, but this requires big code changes.

For your case, I suggest you use a filter to avoid queried all your 45k groups from LDAP.

DASXCE:
I suggest dropping the default GID=CN when GID is blank. I do not suggest if GID=CN then parse the DN. I find it also intuitive... if no GID is set do not search, if it is defined then search for the attribute... however I wouldn't debate whether it is intuitive or not :)

I am aware that the CN attribute does not have to match the CN from the DN string, but why care? I don't see an issue using whatever is inside the CN field from the DN (i assume every LDAP object has a CN in the DN). Also if parsing is not an option why not simply use the whole DN string? 

Regarding my group filter... you are right, but for my case it is easier said than done. I thought it would be nice to have that Option to skip the search...

stonezdj:
The master branch already handle this issue in another way

https://github.com/goharbor/harbor/blob/ec23ddabc3af15fcf88d2554f645a9624bc67ecf/src/pkg/ldap/ldap.go#L369-L371

Thanks for your contribution.

PR #13909, opened 2021-01-06
Provide a mechanism to sweep the execution/task records in task manager

### LINKED ISSUE #13888: Introduce a mechanism in task manager to clean up the tasks
Scan/scan all/gc/replication/retention have already moved to the task manager mechanism, more and more tasks will be created in the database as time goes on. 
We need to introduce a mechanism in task manager to clean up the outdated task/execution records

ywk253100:
Provide a mechanism to sweep the execution/task records in task manager
Fixes #13888

### DIFF
--- a/src/controller/scan/base_controller.go
+++ b/src/controller/scan/base_controller.go
@@ -60,6 +60,11 @@ const (
 	robotIDKey     = "robot_id"
 )
 
+func init() {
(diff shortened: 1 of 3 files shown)


heww (approved):

steven-zou (approved):
LGTM

PR #13967, opened 2021-01-12
fix(retention) retained/total empty when 0

bitsf:
fix retention nightly ui case failed when retained or total is empty


wy65701436 (approved):
lgtm

ywk253100 (approved):

PR #14002, opened 2021-01-14
Fix UI issues found in testing round 1

### LINKED ISSUE #13982: Issues with The name of project scope robotaccount
When creating a project level robot account the name will be `${prefix}${project_name}+${user_input}`
It is confusing to users because a non-admin do not know the prefix and the rules about adding project name.  He does not know what the name will look like when creating the robot account.

How about providing some hints in the creation dialog on UI, so a user can know what the robot name looks like when he is creating the robot account?

### LINKED ISSUE #13983: Editing the expiration time of a robot account is confusing.
The expiration time is represented as "# of days".

If I edit a robot account, update the expiration to change that number, seems the expiration time is calculated against the creation time.

So if I create a robot account, set the expiration as 3 days, then after 2 days I update the expiration to 4 days.  Does it mean the robot will expire in 2 days instead of 4 days?

This is confusing to users, can we use the date/time widget to edit the expiration time?

### LINKED ISSUE #13984: schedule is not translate correctly on GUI
build: v2.2.0-build.284
image

### LINKED ISSUE #13990: Display issue with robot account with system only permission
If I create a robot account with following request body:

```
{
	"name": "test-api-sys-only",
	"duration": 30,
	"description": null,
	"disable": false,
	"level": "system",
	"permissions": [{
		"kind": "system",
		"namespace": "/",
		"access": [{
			"resource": "catalog",
			"action": "read"
		}]
	}]
}
```

On UI the `PROJECTS` column looks weird

Screen Shot 2021-01-13 at 4 31 26 PM

If there's no project permission, we may display 0 projects and make the link not clickable?

AllForNothing:
Fixes #13990 
Fixes #13984
Fixes #13983
Fixes #13982

### DIFF
--- a/src/portal/src/app/project/repository/artifact-list-page/artifact-list/artifact-list-tab/artifact-list-tab.component.html
+++ b/src/portal/src/app/project/repository/artifact-list-page/artifact-list/artifact-list-tab/artifact-list-tab.component.html
@@ -93,8 +93,7 @@
                 <button id="scan-btn" [clrLoading]="scanBtnState" type="button" class="btn  btn-secondary scan-btn"
                     [disabled]="!(canScanNow() && selectedRowHasVul() && hasEnabledScanner && hasScanImagePermission )" (click)="scanNow()">
                     <clr-icon shape="shield-check" size="16"></clr-icon>&nbsp;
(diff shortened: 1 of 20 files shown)


jwangyangls (approved):
lgtm

xuelichao (approved):

PR #14010, opened 2021-01-14
Remove raw token from header in token review reuqest - 2.1.x

reasonerjt:
The server to handle token-review may have a limitation for the size of
the header.  When the token is huge the token review may fail.
This commit remove the necessary header to harden the flow.


wy65701436 (approved):
lgtm

ywk253100 (approved):

PR #14013, opened 2021-01-14
Harbor upgrading for 2.2

ninjadq:
1. if san not exists then remove that cert, prepare will regenerate one
2. Add upgrade script

### DIFF
--- a/make/photon/prepare/migrations/__init__.py
+++ b/make/photon/prepare/migrations/__init__.py
@@ -2,4 +2,4 @@ import os
 
 MIGRATION_BASE_DIR = os.path.dirname(__file__)
 
-accept_versions = {'1.9.0', '1.10.0', '2.0.0', '2.1.0'}
+accept_versions = {'1.9.0', '1.10.0', '2.0.0', '2.1.0', '2.2.0'}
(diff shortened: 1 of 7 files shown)


make/photon/prepare/migrations/__init__.py
```diff
@@ -2,4 +2,4 @@
 
 MIGRATION_BASE_DIR = os.path.dirname(__file__)
 
-accept_versions = {'1.9.0', '1.10.0', '2.0.0', '2.1.0'}
\ No newline at end of file
+accept_versions = {'1.9.0', '1.10.0', '2.0.0', '2.1.0', '2.2.0'}
```
reasonerjt, line 5: I don't think we need to keep adding versions as we only support n-2 upgrading path?

make/photon/prepare/migrations/version_2_2_0/harbor.yml.jinja
```diff
@@ -0,0 +1,435 @@
...
+  http_proxy:
+  https_proxy:
+  no_proxy:
+  components:
+    - core
+    - jobservice
+    - trivy
+{% endif %}
+
+{% if metric is defined %}
```
reasonerjt, line 425: will there be `metric` in 2.1.0?

wy65701436 (approved):
lgtm

make/photon/prepare/migrations/__init__.py
ninjadq: This is the input config's version list which our prepare script can handle. Even we only maintain n-2 versions, we shouldn't block earlier versions to upgrade.

reasonerjt: Fair enough.

reasonerjt (approved):

PR #14023, opened 2021-01-15
Fix the legacy scheduled job issue for GC/scan all

### LINKED ISSUE #13968: The existing scheduled GC/Scan all cannot be tracked by the execution of task manager after upgrading
After upgrading, the existing scheduled GC/Scan all job runs as expected, but they are not tracked by the execution of task manager, which means users cannot get the execution records from portal.

ywk253100:
Fix the legacy scheduled job issue for GC/scan all

Fixes #13968


heww (approved):

steven-zou (approved):
LGTM

PR #14038, opened 2021-01-19
Username from /userinfo

### LINKED ISSUE #14037: OIDC Onboarding fails, if username is not present in the ID Token

Harbor is creating an ID Token without extra claims, but expecting the username to be part of the ID Token. This is true for some OIDCs, but not defined in the standard.
Harbor fails early and is not able to receive the username from the /userinfo endpoint, as expected.

I tried to use [Satosa](https://github.com/IdentityPython/SATOSA) and this occurred.

OIDC config:
* OIDC Scope: openid,name,groups,profile
* Automatic onboarding: true
* Username Claim: "username"

**Versions:**
- harbor version: [2.1.3] and [master]
- docker engine version: [20.10.2]
- docker-compose version: [1.27.4]

shaardie:
This patch enabled Harbor to receive the username from the /userinfo endpoint
instead of only from the ID Token.

Closes #14037

### DIFF
--- a/src/common/utils/oidc/helper.go
+++ b/src/common/utils/oidc/helper.go
@@ -264,6 +264,10 @@ func UserInfoFromToken(ctx context.Context, token *Token) (*UserInfo, error) {
 	if err != nil {
 		log.Warningf("Failed to get userInfo by calling remote userinfo endpoint, error: %v ", err)
 	}
+
+	if setting.UserClaim != "" && local.Username == "" && remote.Username == "" {
+		return nil, fmt.Errorf("OIDC. Failed to recover Username from claim. Claim '%s' is invalid or not a string", setting.UserClaim)
+	}
 	if remote != nil && local != nil {
 		if remote.Subject != local.Subject {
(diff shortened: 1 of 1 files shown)


reasonerjt:
I don't think this is consistent with the design, currently, the "username Claim" in OIDC setting is for automatic onboarding a user.

In your change, if the  "username Claim" is set by mistake the username will be fetched via the `/userinfo` endpoint.

reasonerjt (approved):

wy65701436 (approved):
Lgtm

src/common/utils/oidc/helper.go
```diff
@@ -264,6 +264,10 @@ func UserInfoFromToken(ctx context.Context, token *Token) (*UserInfo, error) {
 	if err != nil {
 		log.Warningf("Failed to get userInfo by calling remote userinfo endpoint, error: %v ", err)
 	}
+
+	if setting.UserClaim != "" && local.Username == "" && remote.Username == "" {
```
reasonerjt, line 268: nil pointer?

PR #14046, opened 2021-01-20
Vulnerabilities metrics (exporter)

ErmakovDmitriy:
This pull request implements Artifact metrics.
The metrics are:

- Vulnerabilities
  - High
  - Medium
  - Low
  - FixedHigh
  - FixedMedium
  - FixedLow
- Artifact Size in bytes.

The exported metric names are:
- project_artifact_detected_vulnerabilities {project_id, repository_id, artifact_id, vulnerability_level}
- project_artifact_fixable_vulnerabilities {project_id, repository_id, artifact_id, vulnerability_level}
- project_artifact_size {project_id, repository_id, artifact_id}

--
Thank you in advance for review,
Dmitrii.


ninjadq:
@ErmakovDmitriy thanks for your PR, vulnerabilities is really important in production environment. But the number of artifacts will explosively increasing. So it isn't a good practice for prometheus

ErmakovDmitriy:
Hi @ninjadq,
I agree that it might not be a good idea to store metrics with growing cardinality.

On the other hand, there should be some other way to monitor the vulnerabilities in images.

Could we, for instance, make the images statistics disabled by default but with cli flag to enable it?
In such case people would be able to estimate tradeoffs and decide.

PR #14052, opened 2021-01-22
Fix UI issue for success rate calculation

### LINKED ISSUE #13861: Replication Failed Successfully
This isn't a big issue, but when the replication fails the **success rate** is **100%,** which is incorrect.

image

I think it would be better to rename the table header from **Success Rate** to **Completion Rate**. 
Completion Rate denotes that something was done but doesn't say if all went well. 

There are two solutions:

- A - Correct percentage of success rate if something fails. If nothing was transferred, set the percentage to 0%
- B - Rename table to **Completion Rate** 

I suspect B would be the simplest option:

Set up a replication rule that fails, pointing for example to a repo where the user doesn't have access.

**Versions:**

- harbor version: 2.1.2

### LINKED ISSUE #13951: OIDC scope filter too restrictive
When I fill in the set of scopes to request in the OIDC configuration webpage (in https://example.com/harbor/configs), it filters the list of scopes on the following regexp:

```
^([\w.]+,){0,}openid(,[\w.]+){0,}$
```

This is too restrictive; the IdP I'm using has a scope with the character `.`

The underlying [RFC](https://www.rfc-editor.org/rfc/rfc6749.txt) is fairly explicit here, valid scope characters are:

```
 %x21 / %x23-5B / %x5D-7E
```

In Python, the equivalent regex would be: 

```
re.compile("^([\x21\x23-\\\x5B\\\x5D-\x7E]+,){0,}openid(,[\x21\x23-\\\x5B\\\x5D-\x7E]+){0,}$")
```

but unfortunately I'm not so handy at javascript to do a PR on my own!

1.  Create a new harbor instance and setup an OIDC provider.
2. In the "OIDC Scopes" input box, input `openid,foo.bar`.  This is not accepted but `foo.bar` is a valid scope according to the RFC.

**Versions:**

- harbor version: 2.0
- docker engine version: unknown (using a provider)
- docker-compose version: using Kubernetes

AllForNothing:
Fixes #13951
Fixes #13861

### DIFF
--- a/src/portal/src/app/config/auth/config-auth.component.html
+++ b/src/portal/src/app/config/auth/config-auth.component.html
@@ -390,7 +390,7 @@
                 </label>
                 <input clrInput name="oidcScope" type="text" #oidcScopeInput="ngModel"
                 [(ngModel)]="currentConfig.oidc_scope.value" id="oidcScope" size="40" required
-                [disabled]="disabled(currentConfig.oidc_scope)" pattern="^(\w+,){0,}openid(,\w+){0,}$" />
+                [disabled]="disabled(currentConfig.oidc_scope)" pattern="^([\w.]+,){0,}openid(,[\w.]+){0,}$"/>
                 <clr-control-error>{{'TOOLTIP.SCOPE_REQUIRED' | translate}}</clr-control-error>
         </clr-input-container>
         <clr-checkbox-container>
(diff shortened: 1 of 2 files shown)


xuelichao (approved):
LGTM

jwangyangls (approved):

PR #14056, opened 2021-01-22
Aggregate  metrics

ninjadq:
1. Add operation id in ctx in baseapi before prepare
2. add operation id for registry proxy request
3. use url for other request


src/server/registry/util/util.go
```diff
@@ -7,6 +7,21 @@ import (
 	"strconv"
 )
 
+const (
```
reasonerjt, line 10: move to the metric pkg? As it will be used for metrics only.

src/server/middleware/metric/metric.go
```diff
@@ -28,9 +47,15 @@ func Middleware() func(http.Handler) http.Handler {
...
+	return transparentHandler
+}
+
+// InjectOpIDMiddleware returns a middleware used for injecting operations ID
+func InjectOpIDMiddleware(opID string) func(next http.Handler) http.Handler {
 	return func(next http.Handler) http.Handler {
-		return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
-			next.ServeHTTP(rw, req)
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			SetMetricOpID(r.Context(), util.CatalogOperationID)
```
reasonerjt, line 57: wrong operation ID?

bitsf:
根据上面的测试截图，有两个问题
1. 为什么 src 是GetPing 之类的，而不是CatalogOperationID 之类的
2. 为什么 src 有为 空 的

ninjadq:
> 根据上面的测试截图，有两个问题
> 
> 1. 为什么 src 是GetPing 之类的，而不是CatalogOperationID 之类的
> 2. 为什么 src 有为 空 的

heww:
I remember that the label name is `operation` not `src`.

src/server/middleware/metric/metric.go
```diff
@@ -1,25 +1,65 @@
...
+	ListTagOperationID = "v2_tags"
+	// ManifestOperationID ...
+	ManifestOperationID = "v2_manifest"
+	// BlobsOperationID ...
+	BlobsOperationID = "v2_blob"
+	// BlobsUploadOperationID ...
+	BlobsUploadOperationID = "v2_blob_upload"
+	// OthersOperationID ...
+	OthersOperationID = "v2_others"
+)
```
heww, line 31: These constants only are used in the `registry/route.go`, please move them to the `registry/route.go`.

bitsf:
对于 src 为空的，建议给个默认值，比如 Unknown 或 Legacy 什么的

src/server/middleware/metric/metric.go
reasonerjt: I was the one who suggested move them to metric pkg as they are for metrics only ...
https://github.com/goharbor/harbor/pull/14056#discussion_r562544486

But I'm fine either way.

src/server/middleware/metric/metric.go
```diff
@@ -1,25 +1,65 @@
...
-		rc := lib.NewResponseRecorder(w)
-		next.ServeHTTP(rc, r)
-		metric.TotalReqDurSummary.WithLabelValues(r.Method, url).Observe(time.Since(now).Seconds())
-		metric.TotalReqCnt.WithLabelValues(r.Method, strconv.Itoa(rc.StatusCode), url).Inc()
+		metric.TotalInFlightGauge.Inc()
+		defer metric.TotalInFlightGauge.Dec()
+		now, rc, src := time.Now(), lib.NewResponseRecorder(w), ""
+		if isChartMuseumURL(r.URL.Path) {
+			src = "chartmuseum"
+		}
```
heww, line 52: Please only do this checking when the `src` (may it named `operation`) is an empty string after the `next.ServeHTTP` called.

reasonerjt:
> 对于 src 为空的，建议给个默认值，比如 Unknown 或 Legacy 什么的

+1 for legacy

src/server/middleware/metric/metric.go
ninjadq: They are arguments of metric middleware, but at the same time they are only used for registry. 
So there are some overlap between these two categories, both of the categories are OK for me.

But when put them to registry, they shouldn't belong to route file  because they are not route. If let me choose one, I will choose `registry/util.go`

ninjadq:
> > 对于 src 为空的，建议给个默认值，比如 Unknown 或 Legacy 什么的
> 
> +1 for legacy

This value is no problem, I just feel it will confuse user because `legacy` is really a "developer" term. Most of admin didn't have sense of it.

And `unknown` is also not accurate to describe it. Because we know  what is the source of these

ninjadq:
> I remember that the label name is `operation` not `src`.

Because `operationID` only works for API and registry proxy request. And there are some legacy and chart related requests. There are value are still determined by other source like URL. So I use `src` as label

reasonerjt (approved):

reasonerjt:
The fix to the pip installation problem may need to Cherry picked to other release branches @ninjadq

heww (approved):

bitsf (approved):

PR #14059, opened 2021-01-23
feat: add health checker for trivy when it's enabled

### LINKED ISSUE #14055: trivy health is not shown in metrics
build: v2.2.0-build.284
Install harbor with metrics enabled and install trivy

curl http://10.161.187.91:9090/metrics|grep harbor_up
find that there's no trivy health
image

heww:
Closes #14055

### DIFF
--- a/src/core/api/health.go
+++ b/src/core/api/health.go
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"sort"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


src/core/api/health.go
```diff
@@ -77,6 +79,9 @@ func (h *HealthAPI) CheckHealth() {
 		}
 		components = append(components, componentStatus)
 	}
+
+	sort.Slice(components, func(i, j int) bool { return components[i].Name < components[j].Name })
```
ninjadq, line 83: why sort components here? I think the order is doesn't matter.

heww: Keep the components in order will make the response more readable when reading by the human.

And the stable response content is required when we want to enable cache-control ([ETag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag)) for the HTTP request. 

Of cause, the harbor doesn't enable the ETag currently. I sort the components only want to make it more readable when I check the health using curl.

ninjadq (approved):
lgtm

reasonerjt (approved):

PR #14078, opened 2021-01-25
Fix metric label related things

ninjadq:
Original type is bool, should convert it to string


wy65701436 (approved):
lgtm

bitsf (approved):

PR #14095, opened 2021-01-27
Tokens

jsoref:
While trying to debug some failing tests outside of Docker, certain things kept crashing. These changes make it easier to test.

One thing that I'm not doing here that would be helpful is if the tests actually set the env var to point to their key instead of expecting users to try to figure out how to set the environment variable in their IDE (which is not fun).


wy65701436 (approved):
lgtm

reasonerjt (approved):

PR #14103, opened 2021-01-27
fix catalog performance issue

wy65701436:

### DIFF
--- a/src/server/registry/catalog.go
+++ b/src/server/registry/catalog.go
@@ -15,35 +15,41 @@
 package registry
 
 import (
-	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/goharbor/harbor/src/controller/artifact"
+	"github.com/goharbor/harbor/src/common/models"
 	"github.com/goharbor/harbor/src/controller/repository"
+	"github.com/goharbor/harbor/src/controller/tag"
+	"github.com/goharbor/harbor/src/lib"
 	"github.com/goharbor/harbor/src/lib/errors"
 	lib_http "github.com/goharbor/harbor/src/lib/http"
+	"github.com/goharbor/harbor/src/lib/orm"
 	"github.com/goharbor/harbor/src/lib/q"
 	"github.com/goharbor/harbor/src/server/registry/util"
 	"net/http"
 	"sort"
 	"strconv"
+	"sync"
 )
 
 func newRepositoryHandler() http.Handler {
 	return &repositoryHandler{
 		repoCtl: repository.Ctl,
-		artCtl:  artifact.Ctl,
+		tagCtl:  tag.Ctl,
+		wp:      lib.NewWorkerPool(20),
 	}
 }
 
 type repositoryHandler struct {
 	repoCtl repository.Controller
-	artCtl  artifact.Controller
+	tagCtl  tag.Controller
+	wp      *lib.WorkerPool
 }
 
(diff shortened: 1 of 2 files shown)


src/server/registry/catalog.go
```diff
@@ -68,16 +75,50 @@ func (r *repositoryHandler) ServeHTTP(w http.ResponseWriter, req *http.Request)
...
+		r.wp.GetWorker()
+		go func(repo *models.RepoRecord) {
+			defer func() {
+				r.wp.ReleaseWorker()
+				wg.Done()
+			}()
+
+			valid, err := r.validateRepo(orm.Clone(req.Context()), repo.RepositoryID)
+			if err != nil {
+				lib_http.SendError(w, err)
```
reasonerjt, line 109: Use a channel to cancel the loop?

reasonerjt:
Once this is done, may consider to CP to 2.1.x?

src/server/registry/catalog.go
```diff
@@ -68,16 +75,50 @@ func (r *repositoryHandler) ServeHTTP(w http.ResponseWriter, req *http.Request)
...
+
+	for _, repo := range repoRecords {
+		r.wp.GetWorker()
+		go func(repo *models.RepoRecord) {
+			defer func() {
+				r.wp.ReleaseWorker()
+				wg.Done()
+			}()
+
+			valid, err := r.validateRepo(orm.Clone(req.Context()), repo.RepositoryID)
```
ywk253100, line 107: We can call the tag manager directly to check whether the repository contains tags or not without listing artifacts
Further more, how about to introduce a function in the repository manager to get the repositories with tags directly? We can join the table repository and tag in database level and get the desired result

heww: Agree to call the tag manager for checking. Not agree to do it in the repo manager using join between the repo table with the tag table.

Join is an expensive operation on large tables.

wy65701436: use the tagCtl for instead.

src/server/registry/catalog.go
```diff
@@ -68,16 +74,54 @@ func (r *repositoryHandler) ServeHTTP(w http.ResponseWriter, req *http.Request)
...
+
+	for _, repo := range repoRecords {
+		r.wp.GetWorker()
+		go func(repo *models.RepoRecord) {
+			defer func() {
+				r.wp.ReleaseWorker()
+				wg.Done()
+			}()
+
+			tags, err := r.tagCtl.List(orm.Clone(req.Context()), &q.Query{
```
ywk253100, line 106: Use `Count` instead

src/server/registry/catalog.go
```diff
@@ -68,16 +74,54 @@ func (r *repositoryHandler) ServeHTTP(w http.ResponseWriter, req *http.Request)
...
+				Keywords: map[string]interface{}{
+					"repository_id": repo.RepositoryID,
+				},
+			}, nil)
+			if err != nil {
+				lib_http.SendError(w, err)
```
ywk253100, line 112: Need to pass the error out of the goroutine and cancel all other goroutines

src/server/registry/catalog.go
wy65701436: updated

src/server/registry/catalog.go
wy65701436: updated

src/server/registry/catalog.go
wy65701436: defined a error chan

PR #14112, opened 2021-01-28
Add sen existed check for internal cert

ninjadq:
fali ealier when there is no san


bitsf (approved):

wy65701436 (approved):
lgtm

PR #14120, opened 2021-01-29
fix catalog performance issue

wy65701436:


ywk253100 (approved):

heww (approved):

skandragon:
Is it possible this broke the `/v2/_catalog` API call which will now return duplicate repositories?

PR #14124, opened 2021-01-29
Reschedule docker login policy in base image build process

danfengliu:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -251,8 +251,8 @@ HARBORPKG=harbor
 # pushimage
 PUSHSCRIPTPATH=$(MAKEPATH)
 PUSHSCRIPTNAME=pushimage.sh
-REGISTRYUSER=user
-REGISTRYPASSWORD=default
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -411,8 +411,10 @@ build_standalone_db_migrator: compile_standalone_db_migrator
 	make -f $(MAKEFILEPATH_PHOTON)/Makefile _build_standalone_db_migrator -e BASEIMAGETAG=$(BASEIMAGETAG) -e VERSIONTAG=$(VERSIONTAG)
 
 build_base_docker:
+	docker login -u $(REGISTRYUSER) -p $(REGISTRYPASSWORD)
```
heww, line 414: We should do the login only when the user and password are provided.

danfengliu: right, I should check it.

heww (approved):

AllForNothing (approved):

PR #14168, opened 2021-02-04
fix: remove default execution sweeper count

heww:
1. Remove the default execution sweeper count for execution vendor.
2. Set the execution sweeper count for gc, preheat, replication,
retention to 50.
3. Keep the executions for the scan job.


ywk253100 (approved):

wy65701436 (approved):
lgtm

PR #14195, opened 2021-02-06
update blob list query

### LINKED ISSUE #12269: Garbage Collection fails for certain Projects

We have a very huge Project (it mirros quite a few containers from docker hub) that doesnt get garbage collected anymore.

Have a huge project
**Versions:**

- harbor version: v1.10.3
- docker engine version: 19.03
- docker-compose version: 1.26.0

**Additional context:**

GC log:
```
2020-06-16T15:06:17Z [INFO] [/jobservice/job/impl/gc/job.go:87]: start to run gc in job.
2020-06-16T15:14:53Z [ERROR] [/jobservice/job/impl/gc/job.go:224]: cannot delete untagged blobs of project: 11, err: sql: expected 16349 arguments, got 147421, just skip it.
2020-06-16T15:14:53Z [INFO] [/jobservice/job/impl/gc/job.go:99]: GC results: status: true, message: a4/awx_rabbitmq```

### PR DESCRIPTION
**Pull request by wy65701436**
Deprecate blob list parameters, and use the query for instead.

Fixes #12269

### CODE REVIEW

File: src/controller/blob/controller.go [hunk 1]
```diff
@@ -216,13 +217,26 @@ func (c *controller) Get(ctx context.Context, digest string, options ...Option)
...
 
-	params := blob.ListParams{
-		ArtifactDigest: opts.ArtifactDigest,
-		BlobDigests:    []string{digest},
-		ProjectID:      opts.ProjectID,
+	keywords := make(map[string]interface{})
+	if digest != "" {
+		ol := q.OrList{
+			Values: []interface{}{
+				digest,
```
ywk253100, line 224: Seems there is only one value, use the exact match instead?

wy65701436: both are ok, but to make logic coherence, I perfer to keep it.

src/pkg/blob/models/blob.go
```diff
@@ -99,11 +103,37 @@ func (b *Blob) IsManifest() bool {
...
-	BlobDigests     []string  // list blobs which digest in the digests
-	ProjectID       int64     // list blobs which associated with the project
-	UpdateTime      time.Time // list blobs which update time less than updatetime
+// FilterByArtifactDigest returns orm.QuerySeter with artifact digest filter
+func (b *Blob) FilterByArtifactDigest(ctx context.Context, qs orm.QuerySeter, key string, value interface{}) orm.QuerySeter {
+	v, ok := value.(string)
+	if !ok {
+		return qs
+	}
+	sql := fmt.Sprintf("IN (SELECT digest_blob FROM artifact_blob WHERE digest_af IN (%s))", `'`+lib_orm.Escape(v)+`'`)
```
ywk253100, line 112: `Escapse()` should only be used on fuzzy match

src/pkg/blob/models/blob.go
```diff
@@ -99,11 +103,37 @@ func (b *Blob) IsManifest() bool {
...
+
+// FilterByArtifactDigests returns orm.QuerySeter with artifact digests filter
+func (b *Blob) FilterByArtifactDigests(ctx context.Context, qs orm.QuerySeter, key string, value interface{}) orm.QuerySeter {
+	artifactDigests, ok := value.([]string)
+	if !ok {
+		return qs
+	}
+	var afs []string
+	for _, v := range artifactDigests {
+		afs = append(afs, `'`+lib_orm.Escape(v)+`'`)
```
ywk253100, line 124: The same issue

ywk253100 (approved):

src/pkg/blob/models/blob.go
wy65701436: will remove it.

src/controller/blob/controller.go
heww: I prefer to change it to use exact match, the old method chose to use blob digests because there is no way to filter the blob by one digest.

src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -423,17 +423,33 @@ func (gc *GarbageCollector) removeUntaggedBlobs(ctx job.Context) {
...
-			continue
-		}
-		if err := gc.blobMgr.CleanupAssociationsForProject(ctx.SystemContext(), p.ProjectID, all); err != nil {
-			gc.logger.Errorf("failed to clean untagged blobs of project, %v", err)
-			continue
+
+		pn := int64(1)
+		for {
+			rg := q.Range{
+				Max: time.Now().Add(-time.Duration(gc.timeWindowHours) * time.Hour).Format(time.RFC3339),
```
heww, line 430: Should we move this line out of the for loop?  Otherwise, the max value will be changed in every loop, this may affect the stability of the results.

src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -423,17 +423,33 @@ func (gc *GarbageCollector) removeUntaggedBlobs(ctx job.Context) {
...
+			}
+			q := &q.Query{
+				Keywords: map[string]interface{}{
+					"update_time": &rg,
+					"projectID":   p.ProjectID,
+				},
+				PageNumber: pn,
+				PageSize:   10000,
```
heww, line 438: 10000 is so bigger.

src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -423,17 +423,33 @@ func (gc *GarbageCollector) removeUntaggedBlobs(ctx job.Context) {
...
+				break
+			}
+			if err := gc.blobMgr.CleanupAssociationsForProject(ctx.SystemContext(), p.ProjectID, all); err != nil {
+				gc.logger.Errorf("failed to clean untagged blobs of project, %v", err)
+				break
+			}
+			if len(all) < 10000 {
+				break
+			}
+			pn++
```
heww, line 452: There are blobs 1, 2, 3, 4, 5, 6, 7, 9, 10 in the project, we list the blobs with a page size of 2. And 1, 2, 3, 4 will be cleaned in GC.

Well, we will get 1, 2 when pg is 1, then we delete the blobs 1 and 2, pg increase to  2. The next time, we get the 5 and 6, blob 3 and 4 are skipped, this is wrong.

We need to ensure to get blob 3 and 4 even the blob 1 and 2 are deleted before.

Suggest to process like the following

```go

pageSize = 1000
lastBlobID := 0

q := &q.Query{
	Keywords: map[string]interface{}{
		"update_time": &rg,
		"projectID":   p.ProjectID,
		"id__gt" lastBlobID,
	},
	PageSize: int64(pageSize),
	PageNumber: 1,
	Sorting: "id"
}

for {
	blobs, err := gc.blobMgr.List(ctx.SystemContext(), q)
	if err != nil {
		gc.logger.Errorf("failed to get blobs of project, %v", err)
		break
	}
	if err := gc.blobMgr.CleanupAssociationsForProject(ctx.SystemContext(), p.ProjectID, blobs); err != nil {
		gc.logger.Errorf("failed to clean untagged blobs of project, %v", err)
		break
	}

	if len(blobs) < pageSize {
		break
	}

	lastBlobID = blobs[len(blobs) -1].ID
}

```

src/jobservice/job/impl/gcreadonly/garbage_collection.go
```diff
@@ -288,9 +288,8 @@ func (gc *GarbageCollector) removeUntaggedBlobs(ctx job.Context) {
 			continue
 		}
 		p := result.Data
-		all, err := gc.blobMgr.List(ctx.SystemContext(), blob.ListParams{
-			ProjectID: p.ProjectID,
-		})
+
+		all, err := gc.blobMgr.List(ctx.SystemContext(), q.New(q.KeyWords{"projectID": p.ProjectID}))
```
heww, line 292: Same processing as the GC task?

src/controller/artifact/abstractor.go
```diff
@@ -95,13 +96,13 @@ func (a *abstractor) abstractManifestV1Metadata(ctx context.Context, artifact *a
 		return err
 	}
 
-	digests := make([]string, len(manifest.FSLayers))
-	for i, fsLayer := range manifest.FSLayers {
-		digests[i] = fsLayer.BlobSum.String()
+	var ol q.OrList
+	for _, fsLayer := range manifest.FSLayers {
+		ol.Values = append(ol.Values, fsLayer.BlobSum.String())
```
heww, line 101: `q.OrList` here is very strange, suggest still use the digests here (change the type of element from string to interface{}), then make the value of the `digest` to `q.OrList` in the q.KeyWords.

src/controller/blob/controller.go
```diff
@@ -178,12 +179,12 @@ func (c *controller) FindMissingAssociationsForProject(ctx context.Context, proj
 		return nil, nil
 	}
 
-	var digests []string
+	var ol q.OrList
```
heww, line 182: q.OrList here is very strange here.

src/pkg/blob/dao/dao.go
```diff
@@ -218,38 +217,12 @@ func (d *dao) UpdateBlob(ctx context.Context, blob *models.Blob) error {
 	return err
 }
 
-func (d *dao) ListBlobs(ctx context.Context, params models.ListParams) ([]*models.Blob, error) {
-	qs, err := orm.QuerySetter(ctx, &models.Blob{}, nil)
+func (d *dao) ListBlobs(ctx context.Context, query *q.Query) ([]*models.Blob, error) {
+	qs, err := orm.QuerySetter(ctx, &models.Blob{}, query)
```
heww, line 221: Process the sorting in the query when want to fix the loop problem in the gc task.

heww (approved):

src/jobservice/job/impl/gcreadonly/garbage_collection.go
wy65701436: this will be deprecated, no need to update.

src/jobservice/job/impl/gc/garbage_collection.go
wy65701436: will do, but put it as here will not impace the results.

src/jobservice/job/impl/gc/garbage_collection.go
wy65701436: here maybe a misunderstanding, this step will not remove data from table blob but project_blob. So, the `all` is static within the for loop.

wy65701436: discussed offline, mark it as resolved.

src/jobservice/job/impl/gc/garbage_collection.go
wy65701436: update to 1000

src/jobservice/job/impl/gc/garbage_collection.go
wy65701436: good catch, buddy

src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -423,17 +423,40 @@ func (gc *GarbageCollector) removeUntaggedBlobs(ctx job.Context) {
...
+		blobRG := q.Range{
+			Min: lastBlobID,
+		}
+
+		for {
+			q := &q.Query{
+				Keywords: map[string]interface{}{
+					"update_time": &timeRG,
+					"projectID":   p.ProjectID,
+					"id":          &blobRG,
```
heww, line 441: blobRG need to move here.

reasonerjt (approved):

PR #14208, opened 2021-02-08
Fix sql issue for querying tasks

ywk253100:
Fix sql issue for querying tasks

### DIFF
--- a/src/pkg/task/dao/task.go
+++ b/src/pkg/task/dao/task.go
@@ -16,7 +16,6 @@ package dao
 
 import (
 	"context"
-	"fmt"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


wy65701436 (approved):
lgtm

src/pkg/task/dao/task.go
```diff
@@ -216,16 +217,43 @@ func (t *taskDAO) querySetter(ctx context.Context, query *q.Query) (orm.QuerySet
...
+		}
+		if len(ids) == 0 {
+			ids = append(ids, -1)
+		}
+
+		idStrs := []string{}
+		for _, id := range ids {
+			idStrs = append(idStrs, strconv.FormatInt(id, 10))
+		}
+		qs = qs.FilterRaw("id", fmt.Sprintf("in (%s)", strings.Join(idStrs, ",")))
```
heww, line 256: This result of `strings.Join(idStrs, ",")` may exceed the max values support by the postgres.

The gc task has the same problem which will be fixed by https://github.com/goharbor/harbor/pull/14195

ywk253100: It is in theory. 
But the real case it that there isn't many ids after the filtering and this PR imports no new issues.
We may open a new ticket to track the exceeding max value issue and provide a more general solution, it isn't for task only

heww (approved):

PR #14246, opened 2021-02-17
AWS public ECR Endpoint Add

folimy:
AWS public ECR Endpoint Add
https://docs.aws.amazon.com/general/latest/gr/ecr-public.html


folimy:
AWS public ECR Endpoint Add
https://docs.aws.amazon.com/general/latest/gr/ecr-public.html

PR #14094, opened 2021-01-26
Spelling

jsoref:
This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

The misspellings have been reported at https://github.com/jsoref/harbor/commit/2f42394794d4e0f02107b6b6d78b91cc30c11d88#commitcomment-46403790

The action reports that the changes in this PR would make it happy: https://github.com/goharbor/harbor/commit/5c439d696282b1977b5e95760fb26d5997350ead

Note: this PR does not include the action. If you're interested in running a spell check on every PR and push, that can be offered separately.

---

I'm aware that the commits are not labeled per this repo's style. I do not expect it to be merged as-is. Keeping the commits separate initially makes it easier for me to drop commits if they're objectionable, to split them off if there are reasons (e.g. wanting to do docs first or by module), rebase if/when requested, etc.

Note: there's one CI item that fails, but it fails even without my changes (LDAP). I hope there's an issue for it, but it's well beyond the scope of this PR.

I've left the [full set of changes on a branch](https://github.com/jsoref/harbor/tree/spelling-full) which I can rebase onto master and make further PRs once pieces from here are merged in.

PR #14263, opened 2021-02-19
Spelling

jsoref:
This is round 2 -- round 1 was #14094

This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

---

I'm aware that the commits are not labeled per this repo's style. I do not expect it to be merged as-is. Keeping the commits separate initially makes it easier for me to drop commits if they're objectionable, to split them off if there are reasons (e.g. wanting to do docs first or by module), rebase if/when requested, etc.

Note: there's one CI item that fails, but it fails even without my changes (LDAP). I hope there's an issue for it, but it's well beyond the scope of this PR.

Also, it appears that a number of CI tests are failing due to Docker Hub limits -- that's also well beyond the scope of this PR.

---

Output changes:
* Makefile (224c88f9d76d0c90f9959e868afdc6375e19d6af)

Bug fix?
* tpl (662e24bc5ba7d5622d88e5f45f7451bb2e455e44)

Rename (ad87da29de83ecf25b1471b75fcb465e5d831a64), Local (9505d121bf74819a5bb3ef521c933e82b20055bf), Identifier (e.g. 8c59f740eaf8b018925f4e3a3299cb67a4d4c220):
* scss

Generally suspicious:
* bak -- src/testing/apitests/apilib/harlogin.bak (I wonder if this file snuck in as part of 737eaa396c5a3dc5e52f359d6908b685ff2b9d11 and shouldn't be in the repo at all)

Not quite doc/comment (maybe requires a change to some other component?)
* index (b37cce65f87aa6db4fcc6dfdd45ec223404376f1, f32ae66f6a889fd065e89292ad30a31e7d4026d1)


AllForNothing:
@jsoref Thanks for the PR, but we decided not to merge this

PR #14271, opened 2021-02-19
add PR template to repo

a-mccarthy:
Signed-off-by: Abigail McCarthy <mabigail@vmware.com>

Opening this PR as a Draft to start a discussion on adding a PR template to the repo. I'm not sure if this has come up before, but I think these templates can be very helpful to make sure folks fill in some basic info about what the PR is for and remind folks to do the required steps to get the PR merged that can get missed, like the DCO.

I'd also like to make sure we call out updating docs now that the docs are kept in a separate repo.

Is there anything else we should call out here? Like for creating release notes or how to run/pass tests?

### DIFF
--- a/.github/PR_TEMPLATE.md
+++ b/.github/PR_TEMPLATE.md
@@ -10,4 +10,5 @@ Fixes #(issue)
 ### Please indicate you've done the following:
 
 - [ ] [Accepted the DCO](https://github.com/goharbor/harbor/blob/master/CONTRIBUTING.md#commit). Commits without the DCO will delay acceptance.
-- [ ] Updated the corresponding documentation in this repo or https://github.com/goharbor/website.
+- [ ] Made sure tests are passing and test coverage is added if needed.
+- [ ] Considered the docs impact and opened a new docs issue or PR with docs changes if needed.
--- a/.github/PR_TEMPLATE.md
+++ b/.github/PR_TEMPLATE.md
@@ -1,0 +1,13 @@
+Thank you for contributing to Harbor!
+
+### Summary of your change
+
+
+### Issue being fixed
+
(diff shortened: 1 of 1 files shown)


a-mccarthy:
cc: @xaleeks

.github/PR_TEMPLATE.md
```diff
@@ -0,0 +1,13 @@
...
+
+
+### Issue being fixed
+
+Fixes #(issue)
+
+### Please indicate you've done the following:
+
+- [ ] [Accepted the DCO](https://github.com/goharbor/harbor/blob/master/CONTRIBUTING.md#commit). Commits without the DCO will delay acceptance.
+- [ ] Updated the corresponding documentation in this repo or https://github.com/goharbor/website.
```
yanji09, line 13: Thanks Abbie for creating the template. Assuming this one is for those PRs need document change. If the PR doesn't need document change, how submitter make the choice. And it would be better to change the document after this PR is merged.
One suggestion is to make it as a note only.

a-mccarthy: Thats a good point. My first goal is to remind folks about updating the docs. I think my concern with making it a note, instead of a check mark is that there is no sign-off from a contributor.

What if we made it more open ended? something like,

```- [ ] Considered the docs impact and opened a new docs issue or PR with docs changes if needed```

My concern is that if we merge a PR without an option like this as a part of the checklist, then there will be no way to remember to open and docs issue or PR after the pr is merged. I'm interested to hear thoughts and feedback on this idea

yanji09: LGTM

reasonerjt (approved):

yanji09 (approved):

Issue #13589, opened 2020-11-24
Harbor Upgrade from v1.10.3 to v2.1.1 / Multiple Issues Seen


narayaar:
**Versions:**
- harbor version: 2.1.1
- docker engine version: 19.03.11
- docker-compose version: 1.24.1
- postgresql version: 11.5 (external)

**Issue:**
As part of our product upgrade cycle, we had 2 running instances of Harbor (1.10.3) that needed upgrading.  We decided to up
grade Harbor from 1.10.3 to 2.1.1, and followed the steps outlined in https://goharbor.io/docs/2.1.0/administration/upgrade/.
Both instances upgraded fine, but then on one of them, issues started cropping up.

The first one had to do with the Project Quotas page not coming up, with the following displayed in the core.log file;
```
[ERROR] [/lib/http/error.go:54]: {"errors":[{"code":"INTERNAL_SERVER_ERROR","message":"failed to query database for quotas, error: project 16 not found"}]}
```
Going by the steps outlined in https://github.com/goharbor/harbor/issues/13387, we were able to get the Project Quotas page working as expected.

The second one seen has to do with a date/time field out of range, with the following displayed in the core.log file;
```
[ERROR] [/core/service/notifications/jobs/handler.go:231]: failed to update the status of retention task 1041: pq: date/time field value out of range: "0000-12-31 19:03:58"
ERROR] [/lib/http/error.go:54]: {"errors":[{"code":"UNKNOWN","message":"unknown: pq: date/time field value out of range: \"0000-12-31 19:03:58\""}]}
```
The third one seen has to do with the jobservice's hook generating errors, with the following displayed in the jobservice.log file;
```
[ERROR] [/jobservice/hook/hook_agent.go:204]: Resend hook event error: outdated hook event: status change: job=ef1a6f86a335deda5e275d53, status=Scheduled
[WARNING] [/jobservice/hook/hook_agent.go:121]: Send hook event 'status change: job=ef1a6f86a335deda5e275d53, status=Scheduled' to 'http://core:8080/service/notifications/tasks/1' failed with error: {"errors":[{"code":"UNKNOWN","message":"internal server error"}]}
```
The fourth one seen has to do with the GC not working as expected, due to which untagged images/unwanted images are not getting deleted.  This was displayed in the jobservice.log file as the following;
```
[...]

Issue #12850, opened 2020-08-21
Fail to run GC job


stonezdj:
1. Install 2.1 build with local file system.
2. Replicate some images to projects 
3. Delete images in these projects
4. Change the GC_TIME_WINDOW_HOURS=0 and restart Harbor.
5. Run GC manually. the GC job failed with the following logs:  all image files such as blobs and repository remain in the /data/registry.
```
2020-08-21T10:24:04Z [INFO] [/jobservice/job/impl/gc/garbage_collection.go:257]: delete the manifest with registry v2 API: library/haproxy, application/vnd.docker.distribution.manifest.v2+json, sha256:8a0fdda4d13959c085d949fae822dae5fd2d384208a6a9600ebb46ea3baaff23
2020-08-21T10:24:04Z [INFO] [/jobservice/job/impl/gc/garbage_collection.go:269]: delete manifest from storage: sha256:8a0fdda4d13959c085d949fae822dae5fd2d384208a6a9600ebb46ea3baaff23
2020-08-21T10:24:04Z [INFO] [/jobservice/job/impl/gc/garbage_collection.go:281]: delete artifact trash record from database: 2, library/haproxy, sha256:8a0fdda4d13959c085d949fae822dae5fd2d384208a6a9600ebb46ea3baaff23
2020-08-21T10:24:04Z [INFO] [/jobservice/job/impl/gc/garbage_collection.go:293]: delete blob from storage: sha256:8a0fdda4d13959c085d949fae822dae5fd2d384208a6a9600ebb46ea3baaff23
2020-08-21T10:24:04Z [ERROR] [/jobservice/job/impl/gc/garbage_collection.go:163]: failed to execute GC job at sweep phase, error: failed to delete blob from storage: sha256:8a0fdda4d13959c085d949fae822dae5fd2d384208a6a9600ebb46ea3baaff23, deletefailed: http status code: 500, body: {"errors":[{"code":"UNKNOWN","message":"internal server error"}]}
```

reasonerjt:
The `.nfs0000` file was generated by NFS client:
https://askubuntu.com/questions/348135/nfs0000-files-keep-showing-up-and-disappearing-when-using-mercurial

Let's keep this  open to investigate why this file kept being generated on this env.

coolpalani:
We're also facing the same issue after upgraded from 1.10.3 to 2.1.1
Garbage collection completely broke.

linyinli:
[...]

PR #14287, opened 2021-02-23
add distribution patch 3309

wy65701436:
This patch can fix the GC failure in the NFS v3 env, see https://github.com/distribution/distribution/pull/3309#issuecomment-783606968


phin1x:
#13589 #12850

reasonerjt:
Let's hold this change for a while and push distribution to merge this change contain it in the next patch release.

PR #14300, opened 2021-02-24
refactor: generate quota APIs by go-swagger

heww:


src/server/v2.0/handler/quota.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	query := &q.Query{
+		Keywords: q.KeyWords{
+			"reference":    lib.StringValue(params.Reference),
+			"reference_id": lib.StringValue(params.ReferenceID),
+		},
+		PageNumber: *params.Page,
+		PageSize:   *params.PageSize,
+		Sorting:    lib.StringValue(params.Sort),
```
wy65701436, line 67: the sorting could be empty here, what about if it's a empty tring?

src/server/v2.0/handler/model/quota.go
```diff
@@ -0,0 +1,62 @@
...
+
+	hard, err := q.GetHard()
+	if err != nil {
+		fields := log.Fields{"quota_id": q.ID, "error": err}
+		log.G(ctx).WithFields(fields).Warningf("failed to get hard from quota")
+	}
+
+	used, err := q.GetUsed()
+	if err != nil {
+		fields := log.Fields{"quota_id": q.ID, "error": err}
```
wy65701436, line 45: What's the value of hard or used on error happen? What's the vaule in the response?

wy65701436 (approved):
lgtm

src/server/v2.0/handler/quota.go
heww: When `Sorting` is an empty string, the DAO of the quota will list by the creation_time in descending order.

src/server/v2.0/handler/model/quota.go
heww: Now, it's changed to return empty map when getting hard or used attribute failed.

ywk253100 (approved):

Issue #13835, opened 2020-12-23
refine the setting in nginx.conf


reasonerjt:
The header `X-Forwarded-Proto` may be set in a smarter way, see:
https://github.com/goharbor/harbor-helm/pull/760

reasonerjt:
Fixed in #14308

PR #14308, opened 2021-02-25
Refine the way to set X-Forwarded-Proto in nginx

reasonerjt:
Refine the way to set the header so user won't need to comment it if
Harbor is sitting behind a reverse proxy.


reasonerjt:
Fixes #13835

ninjadq (approved):
lgtm

ywk253100:
The doc here https://goharbor.io/docs/2.1.0/install-config/troubleshoot-installation/ (`Using nginx or Load Balancing` section) should also be updated?

ywk253100 (approved):

PR #14321, opened 2021-02-26
Fix quotas test issue in upgrade pipeline

danfengliu:
1. Change way for quotas verification in upgrade pipeline, prepare specific size of image, then it's an known value for verifcation;
2. Add notary key rotate test;
3. For issue brought by docker 20, clean containerd cache is the only effective way, so both dockerd and containerd should be cache cleard and restarted;
4. Upgrade E2E Dockerfile for importing readable file size package, and other issues;
5. Uncomment project level robot account test in nightly；
6. Get DNS from local setting, and set it into docker deamon config file.


wy65701436 (approved):
lgtm

wy65701436 (approved):
lgtm

AllForNothing (approved):

PR #14336, opened 2021-03-02
Add "*" to the claim set in the token for /v2 apis

reasonerjt:
The "*" is used by notary server for permission checking:
https://github.com/theupdateframework/notary/blob/84287fd8df4f172c9a8289641cdfa355fc86989d/server/server.go#L200
Hence, we need to add this into the JWT token such that actions like key
rotation can be executed.


wy65701436 (approved):
lgtm

ywk253100 (approved):

PR #14369, opened 2021-03-04
Introduce "sort" in query to provide a general solution for sorting

ywk253100:
Introduce "sort" in query to provide a general solution for sorting


src/pkg/project/dao/dao.go
```diff
@@ -94,13 +94,7 @@ func (d *dao) Create(ctx context.Context, project *models.Project) (int64, error
...
 // Count returns the total count of artifacts according to the query
 func (d *dao) Count(ctx context.Context, query *q.Query) (total int64, err error) {
-	query = q.MustClone(query)
-	query.Keywords["deleted"] = false
-	query.Sorting = ""
-	query.PageNumber = 0
-	query.PageSize = 0
-
-	qs, err := orm.QuerySetter(ctx, &models.Project{}, query)
+	qs, err := orm.QuerySetterForCount(ctx, &models.Project{}, query)
```
heww, line 97: `deleted=false` must in the keywords here.

src/pkg/blob/dao/dao.go
```diff
@@ -222,10 +222,6 @@ func (d *dao) ListBlobs(ctx context.Context, query *q.Query) ([]*models.Blob, er
 	if err != nil {
 		return nil, err
 	}
-
-	if query.Sorting != "" {
-		qs = qs.OrderBy(query.Sorting)
-	}
 	blobs := []*models.Blob{}
```
heww, line 225: The following in gc job should change to `Sorts` if `query.Sorting !=""` in this method was removed.
```golang
			blobRG := q.Range{
				Min: lastBlobID,
			}
			q := &q.Query{
				Keywords: map[string]interface{}{
					"update_time": &timeRG,
					"projectID":   p.ProjectID,
					"id":          &blobRG,
				},
				PageNumber: 1,
				PageSize:   int64(ps),
				Sorting:    "id",
			}
			blobs, err := gc.blobMgr.List(ctx.SystemContext(), q)
```

ywk253100: done

src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -442,7 +442,11 @@ func (gc *GarbageCollector) removeUntaggedBlobs(ctx job.Context) {
 				},
 				PageNumber: 1,
 				PageSize:   int64(ps),
-				Sorting:    "id",
+				Sorts: []*q.Sort{
+					{
+						Key: "id",
+					},
+				},
```
heww, line 449: How about making the `parseSorting` in the q pkg be public and change these to `Sorts: q.ParseSorting("id")`?

Maybe I prefer `q.NewSorts("id")` than `q.ParseSorting("id")`.

src/core/api/search.go
```diff
@@ -51,7 +51,11 @@ func (s *SearchAPI) Get() {
 	keyword := s.GetString("q")
 
 	query := q.New(q.KeyWords{})
-	query.Sorting = "name"
+	query.Sorts = []*q.Sort{
+		{
+			Key: "name",
+		},
+	}
```
heww, line 58: No need now because it's the default order by now.

src/pkg/repository/dao/dao.go
```diff
@@ -71,7 +66,6 @@ func (d *dao) List(ctx context.Context, query *q.Query) ([]*models.RepoRecord, e
 	if err != nil {
 		return nil, err
 	}
-	qs = qs.OrderBy("-CreationTime", "RepositoryID")
```
heww, line 74: What is the default order of the repositories when this line was removed?

src/pkg/tag/dao/dao.go
```diff
@@ -71,7 +65,6 @@ func (d *dao) List(ctx context.Context, query *q.Query) ([]*tag.Tag, error) {
 	if err != nil {
 		return nil, err
 	}
-	qs = qs.OrderBy("-PushTime", "ID")
```
heww, line 74: What is the default order for the tags when this line was removed?

heww (approved):

src/lib/orm/metadata.go
```diff
@@ -0,0 +1,217 @@
...
+
+	// get the metadata from cache first
+	fullName := getFullName(t)
+	cacheMetadata, exist := cache.Load(fullName)
+	if exist {
+		return cacheMetadata.(*metadata)
+	}
+
+	// pointer value
+	v := reflect.ValueOf(model)
```
wy65701436, line 73: why not use the `ptr` here?

wy65701436 (approved):
lgtm

src/lib/orm/metadata.go
ywk253100: `prt` is type, here is value

PR #14383, opened 2021-03-05
Use exec in all components' entrypoints

### LINKED ISSUE #14382: Entrypoints should use the exec Bash command
SIGTERM signals sent by kubernetes on a pod deletion are caught by the entrypoint script but not passed to the actual component binary.

To fix this issue, and as a general good practice, the entrypoint scripts should end with the exec Bash command, as explained in [Docker best pactices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/): 
> This script uses the exec Bash command so that the final running application becomes the container’s PID 1. This allows the application to receive any Unix signals sent to the container.

XavierDuthil:
Use the exec Bash command so that the final running application becomes
the container’s PID 1. This allows the application to receive any Unix
signals sent to the container, in accordance with
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#entrypoint

Currently, SIGTERM signals sent by kubernetes are not passed to the
executed binary.

Closes #14382

Signed-off-by: Xavier Duthil <xavier.duthil@ovhcloud.com>

### DIFF
--- a/make/photon/chartserver/docker-entrypoint.sh
+++ b/make/photon/chartserver/docker-entrypoint.sh
@@ -5,6 +5,4 @@ set -e
 /home/chart/install_cert.sh
 
 #Start the server process
-/home/chart/chartm
-
-set +e
+exec /home/chart/chartm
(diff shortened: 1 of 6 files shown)


holyhope (approved):

ninjadq (approved):
lgtm

PR #14387, opened 2021-03-08
Fix label color issue

### LINKED ISSUE #14377: The color selection in Label creation is unreadable in light mode
When creating labels in light mode, i noticed that the color selection is white on light gray background and is unreadable. This was on the nightly build (v2.3.0-3ecd60b8).

Screen Shot 2021-03-03 at 2 02 59 PM

AllForNothing:
Fixes #14377

light:
image

dark: 
image


jwangyangls (approved):
I think you should check the reason for this problem (whether there was a modification error). I checked the version V2.2.0 that I installed before. This problem did not appear, but why did the new version appear?

xuelichao (approved):

PR #14423, opened 2021-03-12
Update the execution status of retention schedule

ywk253100:
Update the execution status of retention schedule

### DIFF
--- a/make/migrations/postgresql/0041_2.1.4_schema.up.sql
+++ b/make/migrations/postgresql/0041_2.1.4_schema.up.sql
@@ -1,0 +1,9 @@
+/*
+When upgrading from 2.0 to 2.1, the status and revision of retention schedule execution isn't migrated, correct them here.
+As v2.2.0 isn't usable because of several serious bugs, we won't support upgrade from 2.1.4 to 2.2.0 anymore. After we add the
+sql file here, users will get error when upgrading from 2.1.4 to 2.2.0 because of this sql file doesn't exist on 2.2.0
+*/
+UPDATE execution
+SET revision=0, status=task.status
+FROM task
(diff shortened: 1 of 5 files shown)


reasonerjt (approved):

bitsf (approved):

make/migrations/postgresql/0041_2.1.4_schema.up.sql
```diff
@@ -0,0 +1,9 @@
+/*
+When upgrading from 2.0 to 2.1, the status and revision of retention schedule execution isn't migrated, correct them here.
+As v2.2.0 isn't usable because of several serious bugs, we won't support upgrade from 2.1.4 to 2.2.0 anymore. After we add the
+sql file here, users will get error when upgrading from 2.1.4 to 2.2.0 because of this sql file doesn't exist on 2.2.0
```
tianon, line 4: Is there somewhere I can read more about this / where this was discussed further?

If 2.2.0 is generally not usable / buggy, perhaps a note should be added to https://github.com/goharbor/harbor/releases/tag/v2.2.0 warning users of that fact? (at a minimum :grimacing:)

(I only even discovered this when trying to update an environment from 2.1.4 to 2.2.0 and ran into the apparently expected "error: no migration found for version 41" and happened to know where to dig into the code to help me find this PR :confused:)

tianon: I guess this is "fixed" for 2.3.x in #14447, but is there a plan to address this (and whatever underlying issues with 2.2.x are alluded to here) in the 2.2.x series?

tianon: Aha, found the backport in #14434 thanks to the just-tagged v2.2.1-rc1; sorry for the noise!

(Would still love to read more information about "v2.2.0 isn't usable because of several serious bugs" :smile:)

Issue #14414, opened 2021-03-11
Nginx 414 Request-URI Too Large when pushing large contents 


m4r1k:
When mirroring the OpenShift OperatorHub to Harbor 2.2, Nginx fails with the following error
```
<html><head><title>414 Request-URI Too Large</title></head><body><center><h1>414 Request-URI Too Large</h1></center><hr><center>nginx</center></body></html>
```

Follows a sample push command logged by Nginx
```
Mar 10 18:57:51 172.22.0.1 proxy[20934]: 192.168.222.1 - "GET /service/token?account=ocp4&scope=repository%3Aocp4-v4.6%2F3scale-amp2-3scale-rhel7-operator-metadata%3Apull%2Cpush&scope=reposi
tory%3Aocp4-v4.6%2F3scale-amp2-3scale-rhel7-operator%3Apull%2Cpush&scope=repository%3Aocp4-v4.6%2F3scale-amp2-apicast-gateway-rhel8%3Apull%2Cpush&scope=repository%3Aocp4-v4.6%2F3scale-amp2-a
picast-rhel7-operator-metadata%3Apull%2Cpush&scope=repository%3Aocp4-v4.6%2F3scale-amp2-apicast-rhel7-operator%3Apull%2Cpush&scope=repository%3Aocp4-v4.6%2F3scale-amp2-backend-rhel7%3Apull%2
Cpush&scope=repository%3Aocp4-v4.6%2F3scale-amp2-memcached-rhel7%3Apull%2Cpush&scope=repository%3Aocp4-v4.6%2F3scale-amp2-system-rhel7%3Apull%2Cpush&scope=repository%3Aocp4-v4.6%2F3scale-amp
2-zync-rhel7%3Apull%2Cpush&scope=repository%3Aocp4-v4.6%2F3scale-amp26-3scale-operator%3Apull%2Cpush&scope=repository%3Aocp4-v4.6%2Famq7-amq-broker-lts-operator-bundle%3Apull%2Cpush&scope=re
pository%3Aocp4-v4.6%2Famq7-amq-broker-lts-rhel7-operator%3Apull%2Cpush&scope=repository%3Aocp4-v4.6%2Famq7-amq-broker-lts-rhel7%3Apull%2Cpush&scope=repository%3Aocp4-v4.6%2Famq7-amq-broker-
[...]

PR #14429, opened 2021-03-12
[nginx][notary] increase buffers to avoid http 414

m4r1k:
PR opened due to issue https://github.com/goharbor/harbor/issues/14414


m4r1k:
Closing Harbor PR (and issue) given the problem was indeed in `oc`
https://github.com/openshift/oc/issues/789

PR #14445, opened 2021-03-15
Fix API python tests log missing issue

danfengliu:
After debugging for issue of missing some http message logs, we found out that swagger client configuration will be re-initiated by calling models in swagger client, so in API python tests, definition for models must be in front of swagger client definition.


heww (approved):

wy65701436 (approved):
lgtm

PR #14450, opened 2021-03-15
add 'distinct' to SQL

skandragon:
This fixes a bug introduced in the change to only list repos containing tagged artifacts.  The bug is related to the join, which can return the same repo more than once if there are multiple artifacts in the repo.

This adds "distinct" to the select, which should eliminate this.  The test is modified to also detect this, although I am not certain I was able to get the test to run properly.

This bug was introduced between 2.1.0 and 2.2.0, likely in PR #14120


rbadillo:
Thanks for fixing this. 

This was blocking the Docker Registry Integration on the Spinnaker side because it was causing a stack overflow.

skandragon:
@rbadillo I should also probably fix Spinnaker to be a little more defensive.  Do you have a stack trace you can slack to me?

rbadillo:
@skandragon Spinnaker was handling it gracefully (Clouddriver was not dying / No system degradation), so I don't think we need to modify anything on clouddriver, at least not right now.

skandragon:
Go home codecov/project, you're drunk.  This change can't both have a 0% change in codecov/patch and a -24.16% change to codecov/project.

wy65701436:
thanks @skandragon for your catch, as we're in the release pipeline today and the PR must be in. I have already raised an new one and cherry pick to release 2.2. https://github.com/goharbor/harbor/pull/14455

Close it as merged. Thanks.

PR #14504, opened 2021-03-24
Add grafana example for metrics

ninjadq:
add json file for grafana

### DIFF
--- a/contrib/grafana-dashborad/metrics-example.json
+++ b/contrib/grafana-dashborad/metrics-example.json
@@ -1,0 +1,4130 @@
+{
+  "__inputs": [
+    {
+      "name": "DS_PROMETHEUS",
+      "label": "Prometheus",
+      "description": "",
+      "type": "datasource",
+      "pluginId": "prometheus",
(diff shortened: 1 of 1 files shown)


bitsf (approved):

dioguerra:
Can you also add a configMap auto-loader, like so:

```
{{- if .Values.cern.dashboards.enabled -}}
{{- range $path, $_ :=  .Files.Glob "dashboards/*.json" }}
apiVersion: v1
kind: ConfigMap
metadata:
  name: dashboard-{{ trimSuffix ".json" (base $path) }}
  labels:
    {{- range $key, $val := $.Values.cern.dashboards.labels }}
    {{ $key }}: {{ $val | quote }}
    {{- end }}
data:
  {{ base $path }}: |-
    {{- $.Files.Get $path | nindent 4 }}
---
{{- end }}
{{- end -}}
```

ninjadq:
hi @a-mccarthy, yes it is the example for user to get started with. I think it can be added to doc also. Thanks for mention this

heww (approved):

PR #14533, opened 2021-03-29
✨feat(create project ): add global webhook

ruster-cn:
New function: when global webhook is enabled in the configuration file, the user will create an associated webhook when creating a new project.


reasonerjt:
@ruster-cn thanks for the PR but I don't think we are ready for this, if it helps for you env, please keep it in your repo.

If you want to contribute to upstream could you write a proposal and discuss it with other maintainers in the community metting?

Issue #13269, opened 2020-10-14
[OIDC] login with preferred_username fails


bsctl:
Logging with OIDC. We use `preferred_username` as claim for Username Claim without specify the email. Harbor logs first user but further login attempts with other usernames are denied with the following message: `"{"errors":[{"code":"CONFLICT","message":"Conflict, the user with same username or email has been onboarded."}]}"`.

**Expected behavior and actual behavior:**
It looks Harbor wants the `email` as unique user identifier for OIDC users. We expect to use the `preferred_username` only

**Steps to reproduce the problem:**
1. Provision OIDC login in Harbor

`curl -k -X PUT "https://harbor.example.com/api/v2.0/configurations" -H "accept: application/json" -H "Content-Type: application/json" -u "admin:Harbor12345" -d "{ \"auth_mode\": \"oidc_auth\", \"oidc_client_id\": \"harbor\", \"oidc_endpoint\": \"https://oidc.example.com\", \"oidc_groups_claim\": \"groups\", \"oidc_user_claim\": \"preferred_username\", \"oidc_name\": \"keycloak\", \"oidc_scope\": \"openid\", \"oidc_verify_cert\": \"false\", \"oidc_client_secret\": \"********\"}"`

2. Login with OIDC as user "alice": success
3. Logout
4. Login with OIDC as different user, e.g "joe": fail with the error above

An example of id_token released by the IdP is:

```
{
  "exp": 1602674724,
  "iat": 1602673824,
  "auth_time": 0,
  "jti": "*********",
  "iss": "oidc.example.com",
  "aud": "harbor",
  "sub": "***************",
  "typ": "ID",
  "azp": "harbor",
  "session_state": "***************",
  "at_hash": "**************",
  "preferred_username": "alice",
  "email_verified": false,
  "acr": "1",
  "groups": [
    "spectre"
  ],
  "client_id": "harbor",
  "username": "alice",
  "active": true
}
```

**Versions:**
Please specify the versions of following systems.

- harbor version: 2.1.0

**Additional context:**
Harbor deployed with Helm Chart 1.5.0

Morriz:
We have the same problem. Why is harbor feeling responsible to say something about the JWT payload? It is signed and verified, and should just accept the claims given. It might idempotently store them for internal reference if it needs to, but not bork.

So please guys, trust signed JWT and handle payload without failures.

Morriz:
We encounter this after a fresh install of keycloak, which might send something different. Let me inspect:

[...]

Issue #13068, opened 2020-09-16
[OIDC] Conflict in username, when change OIDC domain


yue-wen:
### Get this error, when change OIDC domain.
Error message:
"Conflict in username, the user with same username has been onboarded."

**Step to reproduce**:
* set up OIDC config with endpoint example: https://auth.oidc-old-domain.com/auth/realms/registry
* login some user via OIDC
* change oidc domain, example: https://auth.oidc-new-domain.com/auth/realms/registry
* update OIDC config with new endpoint
* re-login same user of step 2  

**Workaround to fix this case**
* Connect to harbor database `registry`
* find table `oidc_user` 
* change old OIDC endpoint of row `subiss` to new endpoint

_Originally posted by @yue-wen in https://github.com/goharbor/harbor/issues/9600#issuecomment-693119481_

reasonerjt:
@yue-wen unfortunately this is the current design, Harbor uses the `issuer` and `subject` to identify users from OIDC provider.  And your OIDC provider uses the domain as the `issuer`.  I wish to suggest you stick to the same issuer when you change the domain, because logically it is the same one.

I agree current approach does not have enough flexibility to cover the corner cases.  I'm thinking maybe we should introduce other settings so admin can customize the claim for identifying user.

There is another issue related to this limitation:
https://github.com/goharbor/harbor/issues/10797

yue-wen:
@reasonerjt 
Thanks for your reply.

Unfortunately we have to change the OIDC provider domain.

I have read the #10797 .
I have some thoughts instead changing design of "subiss".

If OIDC provider URL changed, which means it should be a "new" provider to harbor. 

Then old onboard oidc user data should removed.
Because base on current design, we can only use  one oidc provider to login at same time.

And the projects access can be reclaimed by oidc group mapping.

With the oidc "Automatic onboarding" feature, user can smoothly migrate to new oidc provider. 

I think the only disadvantage is, that oidc user need update they CLI secret after change OIDC provider url.

reasonerjt:
@yue-wen 
Even if you have to change the domain of OIDC provider I think it may be possible to update the configuration to make sure the `issuer` stays the same.

As for your suggestion, if we decide to support multiple OIDC provider, the groups with same name from different OIDC provider must be treated as different groups, same reason we can't assume `jack` from google and `jack` from facebook is the same person.  So I don't think there can be a smooth migration with such approach.

yue-wen:
@reasonerjt 
[...]

PR #14537, opened 2021-03-30
feat(oidc): support custom email claim

jeremy-verys:
Since `email` is required to be unique, this will enable other claims to be used in cases where the `email` claim from UserInfo endpoint is blank or does not exist.
Fixes https://github.com/goharbor/harbor/issues/13269

Signed-off-by: Jeremy Tan <jeremy.tan@verys.com>


reasonerjt:
@jeremy-verys 
Could you give us an example, which claim in your ID provider do you use to store user's Email?
If the concern is duplication, I think we may workaround the issue by updating the schema or using some preset values if not present.  Instead of letting user customize all claims.

dwa012-zmi:
bump

lindhe:
Maybe related to  #13068 and #16558

PR #14577, opened 2021-04-06
Refactor project member api to new programming model

stonezdj:
Add pkg/member/dao
    Add pkg/member/models
    Add pkg/member/manager
    Add controller/member
    Remove the old project member API


api/v2.0/swagger.yaml
```diff
@@ -2293,6 +2293,160 @@ paths:
           $ref: '#/responses/403'
         '500':
           $ref: '#/responses/500'
+  '/projects/{project_name_or_id}/members':
+    get:
+      summary: Get all project member information
+      description: Get all project member information
+      operationId: listProjectMembers
```
heww, line 2300: Add pagination support to this API?

src/server/v2.0/handler/project.go
```diff
@@ -712,14 +715,14 @@ func getProjectMemberSummary(ctx context.Context, p *project.Project, summary *m
 		wg.Add(1)
 		go func(role int, count *int64) {
 			defer wg.Done()
-
-			total, err := pro.GetTotalOfProjectMembers(p.ProjectID, role)
+			ctx = orm.Context()
+			total, err := a.memberMgr.GetTotalOfProjectMembers(ctx, p.ProjectID, role)
```
heww, line 719: `total, err := a.memberMgr.GetTotalOfProjectMembers(orm.Clone(ctx), p.ProjectID, role)`

orm.Clone(ctx) will insert a new ormer to the ctx so that the request id and other infos still exists in the new ctx.

stonezdj: Updated

api/v2.0/swagger.yaml
stonezdj: Updated

src/server/v2.0/handler/member.go
```diff
@@ -0,0 +1,168 @@
...
+
+func (m *memberAPI) ListProjectMembers(ctx context.Context, params operation.ListProjectMembersParams) middleware.Responder {
+	projectNameOrID := parseProjectNameOrID(params.ProjectNameOrID, params.XIsResourceName)
+	if err := m.RequireProjectAccess(ctx, projectNameOrID, rbac.ActionList, rbac.ResourceMember); err != nil {
+		return m.SendError(ctx, err)
+	}
+	entityName := ""
+	if params.Entityname != nil {
+		entityName = *params.Entityname
+	}
```
heww, line 112: `entityName := lib.StringValue(params.Entityname)`

src/server/v2.0/handler/member.go
```diff
@@ -0,0 +1,168 @@
...
+	query, err := m.BuildQuery(ctx, nil, nil, params.Page, params.PageSize)
+	if err != nil {
+		return m.SendError(ctx, err)
+	}
+	total, err := m.ctl.Count(ctx, projectNameOrID, query)
+	if err != nil {
+		return m.SendError(ctx, err)
+	}
```
heww, line 120: Return directly when the total is zero so that we can reduce a DB query.

heww (approved):

src/server/v2.0/handler/member.go
stonezdj: Updated

src/server/v2.0/handler/member.go
stonezdj: Updated

src/pkg/member/models/member.go
```diff
@@ -0,0 +1,33 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package models
+
+import (
+	"github.com/goharbor/harbor/src/pkg/user/models"
```
ywk253100, line 18: Let's avoid the importing for other pkg packages, this causes import cycle easily 

I check the use of `models.User`, seems only the `UserID` and `GroupID` are used, wrap them in a new model or use them directly

ywk253100 (approved):

src/pkg/member/models/member.go
stonezdj: Updated

PR #14578, opened 2021-04-06
Improve the performance of replication

ywk253100:
Improve the performance of replication by introducing a new API to check whether the blob can be mounted directly

### DIFF
--- a/api/v2.0/swagger.yaml
+++ b/api/v2.0/swagger.yaml
@@ -447,6 +447,37 @@ paths:
           $ref: '#/responses/404'
         '500':
           $ref: '#/responses/500'
+  /repositories:
+    get:
+      summary: List all authorized repositories
+      description: List all authorized repositories
+      tags:
+        - repository
+      operationId: listAllRepositories
+      parameters:
+        - $ref: '#/parameters/requestId'
+        - $ref: '#/parameters/query'
+        - $ref: '#/parameters/sort'
+        - $ref: '#/parameters/page'
+        - $ref: '#/parameters/pageSize'
+      responses:
+        '200':
+          description: Success
+          headers:
+            X-Total-Count:
+              description: The total count of repositories
+              type: integer
+            Link:
+              description: Link refers to the previous page and next page
+              type: string
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/Repository'
+        '400':
+          $ref: '#/responses/400'
+        '500':
+          $ref: '#/responses/500'
   /projects/{project_name}/repositories:
     get:
       summary: List repositories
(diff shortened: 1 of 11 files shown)


ywk253100:
Trying to replicate the same image goang:1.15.6(300M) from 50 different repositories and get the following test result:
|  | Jobservice Woker Count | Time Spent |
|--|--|--|
| Without Improvement | 10 | 5min32s |
| With Improvement | 10 | 42s |
| Without Improvement | 1 | 15min35s |
| With Improvement | 1 | 77s |

src/pkg/reg/adapter/harbor/v2/adapter.go
```diff
@@ -129,3 +129,14 @@ func (a *adapter) listArtifacts(repository string, filters []*model.Filter) ([]*
 	}
 	return filter.DoFilterArtifacts(artifacts, filters)
 }
+
+func (a *adapter) CanBeMount(digest string) (bool, string, error) {
+	repository, err := a.client.getRepositoryByBlobDigest(digest)
```
heww, line 134: Do we need to process the 404 here for the previous version of harbor?

api/v2.0/swagger.yaml
```diff
@@ -384,6 +384,43 @@ paths:
...
+              type: string
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/Repository'
+        '400':
+          $ref: '#/responses/400'
+        '401':
+          $ref: '#/responses/401'
+        '403':
```
reasonerjt, line 418: In what case it returns 403 or 404?

src/pkg/reg/adapter/harbor/v2/adapter.go
ywk253100: updated

src/common/models/repo.go
```diff
@@ -38,6 +45,48 @@ type RepoRecord struct {
 	UpdateTime   time.Time `orm:"column(update_time);auto_now" json:"update_time"`
 }
 
+// FilterByBlobDigest filters the repositories by the blob digest
+func (r *RepoRecord) FilterByBlobDigest(ctx context.Context, qs orm.QuerySeter, key string, value interface{}) orm.QuerySeter {
+	digest, ok := value.(string)
```
steven-zou, line 50: If you only accept a `string` as arg, why not directly declare the value to a `string` type? The caller needs to guarantee that the passing parameter should be a string value.

steven-zou (approved):
LGTM with a minor comment

src/common/models/repo.go
ywk253100: The method `FilterByXXX` is an interface defined for the ORM query, it isn't only for the repository

heww (approved):

PR #14604, opened 2021-04-09
 API for users to new model

reasonerjt:
This commit moves the legacy apis related to users to new model.
 Some funcs under common/dao are left b/c they are used by other module,
 which should also be shifted to leverage managers.
 We'll handle them separately.


src/pkg/user/dao/dao.go
```diff
@@ -17,24 +17,75 @@ package dao
...
+		return 0, errors.BadRequestError(nil).WithMessage("user ID is set when creating user: %d", user.UserID)
+	}
+	ormer, err := orm.FromContext(ctx)
+	if err != nil {
+		return 0, err
+	}
+	id, err := ormer.Insert(user)
+	if e := orm.AsConflictError(err, "user %s or email %s already exists", user.Username, user.Email); e != nil {
+		err = e
+	}
```
heww, line 70: `WrapConflictError` will be better than `AsConflictError`

```go
if err != nil {
    return 0, orm.WrapConflictError(err, "user %s or email %s already exists", user.Username, user.Email)
}
```

src/server/v2.0/handler/user.go
```diff
@@ -0,0 +1,485 @@
...
+func (u *usersAPI) SetCliSecret(ctx context.Context, params users.SetCliSecretParams) middleware.Responder {
+	uid := int(params.UserID)
+	if err := u.requireForCLISecret(ctx, uid); err != nil {
+		return u.SendError(ctx, err)
+	}
+	if err := requireValidSecret(params.Secret.Secret); err != nil {
+		return u.SendError(ctx, err)
+	}
+	if err := u.ctl.SetCliSecret(ctx, uid, params.Secret.Secret); err != nil {
+		logger.Errorf("Failed to set CLI secret, error: %v", err)
```
heww, line 62: `log.G(ctx).Errorf("Failed to set CLI secret, error: %v", err)`?

src/server/v2.0/handler/user.go
```diff
@@ -0,0 +1,485 @@
...
+		return u.SendError(ctx, err)
+	}
+	return users.NewSetCliSecretOK()
+}
+
+func (u *usersAPI) CreateUser(ctx context.Context, params users.CreateUserParams) middleware.Responder {
+	logger := log.GetLogger(ctx)
```
heww, line 69: No need to get a logger here, it's used only once in this method at line 88. Just using `log.G(ctx).Errorf("Failed to create user, error: %v", err)` is ok.

src/server/v2.0/handler/user.go
```diff
@@ -0,0 +1,485 @@
...
+	values := params.HTTPRequest.URL.Query()
+	for _, k := range []string{"username", "email"} {
+		if v := values.Get(k); v != "" {
+			query.Keywords[k] = &q.FuzzyMatchValue{Value: v}
+		}
+	}
+	total, err := u.ctl.Count(ctx, query)
+	if err != nil {
+		return u.SendError(ctx, err)
+	}
```
heww, line 113: May return directly here when the total is zero to reduce a DB query.

src/server/v2.0/handler/user.go
```diff
@@ -0,0 +1,485 @@
...
+	sctx, _ := security.FromContext(ctx)
+	if ns, ok := types.NamespaceFromResource(rbac.Resource(scope)); ok {
+		for _, policy := range ns.GetPolicies() {
+			if sctx.Can(ctx, policy.Action, policy.Resource) {
+				policies = append(policies, policy)
+			}
+		}
+	}
+	var res []*models.Permission
+	relative := params.Relative != nil && *params.Relative
```
heww, line 149: `relative := lib.BoolValue(params. Relative)` ?

src/server/v2.0/handler/user.go
```diff
@@ -0,0 +1,485 @@
...
+	}
+	return users.NewUpdateUserProfileOK()
+}
+
+func newUsersAPI() restapi.UsersAPI {
+	return &usersAPI{
+		ctl:     user.Ctl,
+		getAuth: config.AuthMode,
+	}
+}
```
heww, line 261: Move this func to the front or end?

src/server/v2.0/handler/user.go
reasonerjt: fixed

src/server/v2.0/handler/user.go
reasonerjt: fixed

src/server/v2.0/handler/user.go
reasonerjt: good point, fixed.

src/server/v2.0/handler/user.go
reasonerjt: Ah, thanks.  I'm not aware of this utility.

src/server/v2.0/handler/user.go
reasonerjt: Sure.

Vad1mo (approved):

wy65701436 (approved):
lgtm

heww (approved):

PR #14608, opened 2021-04-11
fix ldap_scope and ldap_search_dn swagger definition

hobti01:
The `ldap_scope` and `ldap_search_dn` swagger definitions use simple types, but the actual server responses use complex `StringConfigItem` and `IntegerConfigItem` types.  This change updates the swagger definition to match the implementation.

Signed-off-by: Tim Hobbs <timothy.hobbs@ic-consult.com>


stonezdj:
Thanks for your contribution to the Harbor! 
we have already fixed this issue in the master branch.
https://github.com/goharbor/harbor/blob/master/api/v2.0/swagger.yaml#L6922

PR #14619, opened 2021-04-12
Add --no-cache and --pull flag to image builds

PushkarJ:
This ensures latest security fixes are pulled from base image and package managers in case the cache is not clean on builder machines


wy65701436 (approved):
lgtm

reasonerjt (approved):

PR #14628, opened 2021-04-13
fix(js):job log not found issue

steven-zou:
- return 404 code when the job log is not found


wy65701436 (approved):
lgtm

heww (approved):

PR #14650, opened 2021-04-14
Api refactor label

wy65701436:


src/server/v2.0/handler/label.go
```diff
@@ -0,0 +1,218 @@
...
+		if err := lAPI.RequireProjectAccess(ctx, pid, rbac.ActionList, rbac.ResourceLabel); err != nil {
+			return lAPI.SendError(ctx, err)
+		}
+		query.Keywords["ProjectID"] = pid
+	}
+
+	total, err := lAPI.labelMgr.Count(ctx, query)
+	if err != nil {
+		return lAPI.SendError(ctx, err)
+	}
```
heww, line 123: Return directly when the total is zero to reduce a DB query.

src/server/v2.0/handler/label.go
```diff
@@ -0,0 +1,218 @@
...
+	query := q.New(q.KeyWords{"Name": label.Name, "Level": label.Level, "Scope": label.Scope, "ProjectID": label.ProjectID})
+	if label.Name != oldName {
+		labels, err := lAPI.labelMgr.List(ctx, query)
+		if err != nil {
+			return lAPI.SendError(ctx, err)
+		}
+		if len(labels) > 0 {
+			return lAPI.SendError(ctx, errors.New("conflict label").WithCode(errors.ConflictCode))
+		}
+	}
```
heww, line 174: Can we handle the conflicted error in the DAO layer?

src/server/v2.0/handler/label.go
```diff
@@ -0,0 +1,218 @@
...
+	label, err := lAPI.labelMgr.Get(ctx, params.LabelID)
+	if err != nil {
+		return lAPI.SendError(ctx, err)
+	}
+	if err := lAPI.requireAccess(ctx, label, rbac.ActionDelete); err != nil {
+		return lAPI.SendError(ctx, err)
+	}
+	id := label.ID
+	// TODO remove this step once chart-museum is removed.
+	if err := dao.DeleteResourceLabelByLabel(id); err != nil {
```
heww, line 193: I found that `DeleteResourceLabelByLabel` is just a function executing a DB query,  can we move this function to the manager/dao of the label?

src/server/v2.0/handler/label.go
```diff
@@ -0,0 +1,218 @@
...
+	case common.LabelScopeGlobal:
+		label.ProjectID = 0
+	case common.LabelScopeProject:
+		exist, err := lAPI.projectCtl.Exists(ctx, label.ProjectID)
+		if err != nil {
+			return lAPI.SendError(ctx, errors.Wrapf(err, "failed to check the existence of project %d", label.ProjectID))
+		}
+		if !exist {
+			return lAPI.SendError(ctx, errors.New(nil).WithMessage("project %d not found", label.ProjectID).WithCode(errors.BadRequestCode))
+		}
```
heww, line 51: The project will be checked in the `requireAccess` method, do we need these checking here?

src/server/v2.0/handler/label.go
```diff
@@ -0,0 +1,218 @@
...
+
+	query := q.New(q.KeyWords{"Name": label.Name, "Level": label.Level, "Scope": label.Scope, "ProjectID": label.ProjectID})
+
+	labels, err := lAPI.labelMgr.List(ctx, query)
+	if err != nil {
+		return lAPI.SendError(ctx, err)
+	}
+	if len(labels) > 0 {
+		return lAPI.SendError(ctx, errors.New("conflict label").WithCode(errors.ConflictCode))
+	}
```
heww, line 66: Can we handle this conflicted error in the DAO layer? Because the create also may meet conflicted error even this checking is passed.

src/pkg/label/manager.go
```diff
@@ -16,9 +16,10 @@ package label
 
 import (
 	"context"
-	"github.com/goharbor/harbor/src/common/models"
 	"github.com/goharbor/harbor/src/lib/errors"
 	"github.com/goharbor/harbor/src/lib/q"
+	dao2 "github.com/goharbor/harbor/src/pkg/label/dao"
```
heww, line 21: Remove the import alias?

src/server/v2.0/handler/label.go
wy65701436: see the TODO, it will be remove after chartmuseum is removed.

src/server/v2.0/handler/label.go
wy65701436: just refactor, no logic change.

src/server/v2.0/handler/label.go
wy65701436: just refactor, no logic change.

src/server/v2.0/handler/label.go
wy65701436: updated

src/server/v2.0/handler/label.go
wy65701436: remove the dup check

src/pkg/label/manager.go
wy65701436: updated

src/server/v2.0/handler/label.go
wy65701436: removed

src/core/label/manager.go
```diff
@@ -107,8 +112,8 @@ func (bm *BaseManager) GetLabelsOfResource(resourceType string, resourceIDOrName
 }
 
 // Exists is the implementation of same method in Manager interface.
-func (bm *BaseManager) Exists(labelID int64) (*models.Label, error) {
-	label, err := dao.GetLabel(labelID)
+func (bm *BaseManager) Exists(labelID int64) (*model.Label, error) {
+	label, err := bm.LabelMgr.Get(orm.Context(), labelID)
```
ywk253100, line 116: If the label doesn't exist, will a not found error be returned? If it is, should replace the logic:
```
if label == nil {
	return nil, NewErrLabelNotFound(labelID, "", nil)
}
```

src/server/v2.0/handler/label.go
```diff
@@ -0,0 +1,189 @@
...
+	case common.LabelScopeGlobal:
+		resource := system.NewNamespace().Resource(rbac.ResourceLabel)
+		return lAPI.RequireSystemAccess(ctx, action, resource)
+	case common.LabelScopeProject:
+		if len(subresources) == 0 {
+			subresources = append(subresources, rbac.ResourceLabel)
+		}
+		return lAPI.RequireProjectAccess(ctx, label.ProjectID, action, subresources...)
+	}
+	return errors.New("unsupported label scope").WithCode(errors.ForbiddenCode)
```
ywk253100, line 188: Should return 400 rather than 403?

wy65701436: OK

ywk253100 (approved):

src/core/label/manager.go
wy65701436: updated.

heww (approved):

PR #14652, opened 2021-04-14
[ 2.1.0 ] Add build base step to UT test in CI

danfengliu:
1. Add buid base step to UT test in CI, base image used by UI test
should be built before buiding harbor image;
2. In build package workflow, trigger build base image step in condition of changing both in
Dockerfile.base and VERSION
3. Add tag for setup nightly test


bitsf (approved):

wy65701436 (approved):
lgtm

PR #14677, opened 2021-04-16
fix(replication) aws adapter checkRepository

### LINKED ISSUE #14069: Replication to AWS ECR does not create target repository
## Description
I have a project named `tester`, and a repository named `foo`. I have setup replication from Harbor to destination ECR with Source resource filter: `tester/**`. Replication is set up with a blank destination namespace.

Pushing and pulling `tester/foo` from Harbor works. Replication to ECR fails when the repository does not exist in ECR. Replication succeeds when the repository is manually created in ECR.

Harbor version: v2.2.0-42559479.
Kubernetes version: AWS EKS 1.15

## Expected behavior
The repository `foo` does not exist in ECR. Expecting the ECR repository `<account>.dkr.ecr.<region>.amazonaws.com/tester/foo` to be created by Harbor before replication. Replication to succeed after repository is created.

## Actual behavior
The repository is not created in ECR. Replication fails with `failed to do the prepare work for pushing/uploading resources: RepositoryNotFoundException: The repository with name 'tester/foo' does not exist..`. Replication does not even start as the repository does not exist and was not created.

Screenshot 2021-01-25 at 15 26 36

Screenshot 2021-01-25 at 15 27 12

## Steps to reproduce the problem
1. Create a new project. Push an image into it.
2. Create a replication rule to ECR.
3. Do not create the target repository in ECR.
4. Manually trigger replication.
5. Replication fails.

## Additional context

The functionality to create the ECR repository is in `/src/replication/adapter/awsecr/adapter.go`. However, I suspect that the issue is with the `checkRepository` function at `https://github.com/goharbor/harbor/blob/master/src/replication/adapter/awsecr/adapter.go#L246`. An error is thrown which prevents `createRepository` from running.

bitsf:
fix #14069 
aws adapter catch error in checkRepository

### DIFF
--- a/src/pkg/reg/adapter/awsecr/adapter.go
+++ b/src/pkg/reg/adapter/awsecr/adapter.go
@@ -248,12 +248,14 @@ func (a *adapter) checkRepository(repository string) (exists bool, err error) {
 		RepositoryNames: []*string{&repository},
 	})
 	if err != nil {
+		if e, ok := err.(awserr.Error); ok {
+			if e.Code() == awsecrapi.ErrCodeRepositoryNotFoundException {
+				return false, nil
+			}
+		}
 		return false, err
 	}
-	if len(out.Repositories) > 0 {
-		return true, nil
-	}
-	return false, nil
+	return len(out.Repositories) > 0, nil
 }
 
 func (a *adapter) createRepository(repository string) error {


ywk253100 (approved):

steven-zou (approved):
LGTM

PR #14695, opened 2021-04-20
Refine the "not support" error when searching user

reasonerjt:
If authenticator does not support searching user/group, wraps it as a
not found error, such that the API will return 404 rather than confusing
500.


stonezdj (approved):

wy65701436 (approved):
lgtm

PR #14705, opened 2021-04-20
feat: add artifact count cache

moshloop:
The artifact count function is used heavily when browsing the repository pages (and I suspect during retention jobs as well).

This PR adds a 10-minute query result cache using the keywords passed to the query.

Tested projects with repositories with tags in the 5k+ range and works well

### DIFF
--- a/src/core/main.go
+++ b/src/core/main.go
@@ -45,8 +45,8 @@ import (
 	"github.com/goharbor/harbor/src/core/middlewares"
 	"github.com/goharbor/harbor/src/core/service/token"
 	"github.com/goharbor/harbor/src/lib/cache"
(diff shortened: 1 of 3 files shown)


heww:
Please use the `FetchOrSave` in the `src/lib/cache` to cache the count result.

Emm, the `cache.Default()` will always return a cache in the core component at the moment.

moshloop:
@heww done

Vad1mo (approved):

chlins:
@moshloop I think this way will pose the potential data inconsistency because if in the TTL interval, user create/delete artifact, that will lead to different count between cache and database, although it's a solution, but it's not a stringent solution, actually we have did a proposal about cache resources in harbor(https://github.com/goharbor/community/pull/185), could comment your idea if you are interested in it, thanks.

PR #14706, opened 2021-04-20
feat: add DISABLE_QUOTA option

moshloop:
Quotas are an expensive feature as `FOR UPDATE` locks are taken each time a image is written AND when tags are removed, 

On a large scale repo with per image tag counts in the 5k+ range, this reduces tag deletion by almost 80% from 10s to +- 2s


heww:
This is an env option to disable the quota feature, set `QUOTA_PER_PROJECT_ENABLE` env to `false` will disable the quota for the projects.

PR #14707, opened 2021-04-20
feat: add indexes on common fields

moshloop:


Vad1mo:
@chlins is this not related to your performance improvement efforts?

chlins:
@Vad1mo Yeah, it's about performance and yes we found some indexes can help to improve performance but we don't want to just to resolve issues by adding too many indexes. Also after our test we also create a wiki to give some useful indexes: https://github.com/goharbor/harbor/wiki/Harbor-FAQs#create-index-for-harbor-tables-to-improve-performance

stonezdj:
The script is run in migrate, which is wrapped in a transaction, the `CREATE INDEX CONCURRENTLY` is not allowed in a transaction.

stonezdj:
Because our recent PR #17022 include some of the changes, can I close this PR? Thanks for your contribution.

PR #14717, opened 2021-04-21
Add the ldap group config validation

stonezdj:
Add AssignByTag function to lib/config
When ldap_group_search_filter is set, check other required fields
When ldap_group_search_filter is empty, skip to attach group

### DIFF
--- a/src/lib/config/config.go
+++ b/src/lib/config/config.go
@@ -97,6 +97,7 @@ func Init() {
 	log.Info("init secret store")
 	// init secret store
 	initSecretStore()
+	DefaultCfgManager = common.DBCfgManager
 }
 
 // InitWithSettings init config with predefined configs, and optionally overwrite the keyprovider
--- a/src/pkg/config/validate/validate.go
+++ b/src/pkg/config/validate/validate.go
@@ -1,0 +1,30 @@
+//  Copyright Project Harbor Authors
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package validate
+
+import (
+	"context"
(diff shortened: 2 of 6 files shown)


src/lib/config/config.go
```diff
@@ -126,3 +133,109 @@ func Load(ctx context.Context) error {
 func Upload(cfg map[string]interface{}) error {
 	return defaultMgr().UpdateConfig(orm.Context(), cfg)
 }
+
+// AssignByTag set the value of config model by tags, if it exist in cfgs, it is updated, if it comes from cfgManager, updated is false
+// before using this function, the field of the model should have config tag or json tag, for example
+// type GroupConf struct {
+//	BaseDN              string `json:"ldap_group_base_dn,omitempty"`
+//	Filter              string `config:"ldap_group_search_filter" json:"ldap_group_filter,omitempty"`
```
reasonerjt, line 141: it seems the "config" and "validate" annotations are only needed when you want custom validation, and there's only one case in whole Harbor?

stonezdj: config, validate only required when need to use the validation lib

src/lib/config/config.go
```diff
@@ -126,3 +133,109 @@ func Load(ctx context.Context) error {
...
+//	AdminDN             string `json:"ldap_group_admin_dn,omitempty"`
+//	MembershipAttribute string `json:"ldap_group_membership_attribute,omitempty"`
+// }
+// config tag is the configure item key that you defined in metadatalist.go
+// json is the json attribute of the item
+// if the config tag item name is identical with the json attribute, just define the json tag
+func AssignByTag(ctx context.Context, cfgMgr Manager, cfgs map[string]interface{}, s interface{}) (bool, error) {
+	rv := reflect.ValueOf(s)
+	if rv.Kind() != reflect.Ptr || rv.IsNil() {
+		return false, libErrors.New("can not assign value to nil or pointer")
```
reasonerjt, line 153: should `s` be a pointer or not?

stonezdj: updated, it should be a pointer

heww (approved):

src/pkg/config/validate/validate.go
```diff
@@ -0,0 +1,30 @@
...
+//  limitations under the License.
+
+package validate
+
+import (
+	"context"
+	"github.com/goharbor/harbor/src/lib/config"
+)
+
+// Rule rules to validate the configure input
```
reasonerjt, line 22: Please clarify the input parm cfgs is partial.

reasonerjt (approved):

PR #14724, opened 2021-04-22
[Cherrypick to v2.2]: Add attribute admin username for authproxy

reasonerjt:
This commit adds the attribute "http_authproxy_admin_usernames", which
is string that contains usernames separated by comma, when a user logs
in and the username in the tokenreview status matches the setting of
this attribute, the user will have administrator permission.


reasonerjt:
Let's make it into v2.2.2 such that it can be picked up by operator release sooner.

ywk253100 (approved):

stonezdj (approved):

PR #14746, opened 2021-04-25
fix(jobservice): batch blob deletions during GC

moshloop:
replaces: https://github.com/goharbor/harbor/pull/13091


wy65701436:
Thanks for the PR, but this [PR](https://github.com/goharbor/harbor/pull/14195) has already fixed the blob deletion issue.

Issue #14512, opened 2021-03-25
Access to 'GET /projects/<name>' denied for robot account although permissions are set


defreng:
If you are reporting a problem, please make sure the following information are provided:

**Expected behavior and actual behavior:**
With a system-level robot account and the following permissions, I receive a 403 FORBIDDEN response when trying to access the `GET /projects/existing_name` API.

```
{
    'kind': 'system',
    'namespace': '/',
    'access': [
        {'resource': 'project', 'action': 'list'},
        {'resource': 'project', 'action': 'read'},
    ]
},
{
    'kind': 'project',
    'namespace': '*',
    'access': [
        {'resource': 'repository', 'action': 'list'},
        {'resource': 'artifact', 'action': 'list'},
        {'resource': 'artifact', 'action': 'read'},
        {'resource': 'artifact', 'action': 'create'},
    ]
}
```

On the other hand, listing projects using `GET /projects` works fine.

**Versions:**
Please specify the versions of following systems.

- harbor version: 2.2.0

defreng:
I just realized a better workaround, but even hackier by setting the following permission:

```
{
    'kind': 'project',
    'namespace': '*',
    'access': [
        # FIXME: Super hack.
        # The question mark will be interpreted as part of the regex in Harbor RBAC for resource matching and
        # negate the unnecessary '/'. Regex will become '/project/.*/?', which matches the requested resource
        # for reading a project, e.g. '/project/37'
        {'resource': '?', 'action': 'read'},
    ]
},
```

Notice that the permission must be set as one of kind `project`, not system as mentioned in the initial post.

wy65701436:
if you want the robot to have the list permission of all the projects, just try with:

```
{
    'kind': 'project',
    'namespace': '*',
    'access': [
        {'resource': '', 'action': 'list'}, --- let the resource be empty.
    ]
}

```
with a particular project, with:

```
{
    'kind': 'project',
    'namespace': 'project_name',
    'access': [
        {'resource': '', 'action': 'list'},
    ]
}
```

defreng:
I'm actually not trying to list all projects - but to read a specific one (to get the project ID from its name).

wy65701436:
@defreng the above comment helps?

poblahblahblah:
The only way I've found that I can use a robot user to list projects via the API is with the following permission set:

```
{
   "name":"service.test1",
   "duration":30,
   "description":"service.test2",
   "disable":false,
   "level":"system",
[...]

Issue #14145, opened 2021-02-02
Robot Account with Advanced Permissions


Vad1mo:
Now that Harbor 2.2 has support for more sophisticated robot accounts types with already 10 permission out of the box. 

It would be very beneficial to have an additional permission sets similar to normal users:

- System Admin
- Project Admin 
- Maintainer
- Developer
- Guest
- Limited Guest

**Why**
The primary reason behind the System Admin Robot account is that in an OIDC environment it is often impossible to use the REST api a normal user. See related issues #8033, #13093, #13683, #10597 #14236

**Expected**
System Admin with 3 permissions Read/Create/Delete

**API LEVEL**
On the REST API level 
- **READ** should match GET and CREATE 
-  **CREAE** should match PUT/PATCH/POST
- **DELETE** should match DELETE

wy65701436:
hi @Vad1mo, do you need the robot to have the system access? Can you just clarify the resources?

phin1x:
Robot Accounts with admin permissions would be very useful. As Vad1mo mentioned you cannot use the rest api programaticly because you need a oidc token. You can get a token with a application id and application secret (azure ad), but the token is useles because you need to onboard a user first and you cannot onboard a application. So we had to create the required database entries manually:

```sql
DO $$
DECLARE
        -- set application subject
        --
        -- howto get subject:
        -- url: https://login.microsoftonline.com/a8cea0e8-2d5f-4772-9440-2820c0ef44b9/oauth2/v2.0/token
        -- post body params:
        -- grant_type: client_credentials
        -- client_id: <broker application id>
        -- client_secret: <broker application secret>
        -- scope: <harbor app id>/.default
        --
        -- jwt in z.B. jwt.io decoden und subject entnehmen
        oidc_subject varchar := '';

        -- set issuer endpoint from web ui
        oidc_issuer varchar := '';

        userid harbor_user.user_id%TYPE;
BEGIN

INSERT INTO harbor_user (username, password, realname, sysadmin_flag) VALUES ('service-broker', '', 'Service Broker', true) RETURNING user_id INTO userid;

INSERT INTO oidc_user (user_id, secret, subiss) VALUES (userid, '', concat(oidc_subject, oidc_issuer));

END $$
```

If we have a robot account with admin permissions, we could avoid all this.

Vad1mo:
[...]

PR #14753, opened 2021-04-26
fix rbac project subject for robot accounts

phin1x:
as we began to use system level robot acocunt we could not list, delete or get projects from the api. we discoverd that the rbac subject for projects are wrong for robot accounts.

the the naming schema for subjects is:
```
/<system | project>/<resource id>/<subresource>
```

if a robot account has premissions to list all projects, the allowed subject looks like this:
```
/project/*/project
```

but harbor expects the following subject:
```
/project/*
```

These two subjects never match, so a robot account may never manage projects. For users, the subjects are created correctly. We checked this by changing the expected subject for projects. Now robot accounts were allowed to interact with projects, but users were not.

this pr fixes this issue by omitting the subresouce for project rbac subjects.

Related issues: #14145 #14512

### DIFF
--- a/src/server/middleware/security/robot2.go
+++ b/src/server/middleware/security/robot2.go
@@ -2,6 +2,7 @@ package security
 
 import (
 	"fmt"
+	"github.com/goharbor/harbor/src/common/rbac"
 	"github.com/goharbor/harbor/src/common/security"
 	robotCtx "github.com/goharbor/harbor/src/common/security/robot"
 	"github.com/goharbor/harbor/src/common/utils"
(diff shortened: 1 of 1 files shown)


phin1x:
@steven-zou what means "pending on voting"?

phin1x:
@heww @wy65701436 any updates on this pr?

sidewinder12s:
Any updates?

This issue mostly breaks Helm OCI integration as you can't list manifests on a helm chart without both project level permissions and docker catalog permissions. So, it'd probably be good to fix this before removing chart museum

wy65701436:
The [PR](https://github.com/goharbor/harbor/pull/15431) is for robot to list projects.

PR #14754, opened 2021-04-26
fix project creation with robot account

phin1x:
we tried to create projects using a robot account, but the api call always fails with the message "user robot$<name> not found". it turns out that harbor is trying to get the id of the current authorized user when that user is not a "solution user". since a robot account is not a regular user, the api call always aborts with the error "user not found".

this pr fixes the issue by checking if the current authorized user starts with the robot prefix. if thats the case the owner id is set to one.

Related issues: #14145

### DIFF
--- a/src/server/v2.0/handler/project.go
+++ b/src/server/v2.0/handler/project.go
@@ -158,10 +158,10 @@ func (a *projectAPI) CreateProject(ctx context.Context, params operation.CreateP
 	// set the owner as the system admin when the API being called by replication
 	// it's a solution to workaround the restriction of project creation API:
 	// only normal users can create projects
-	ownerName := secCtx.GetUsername()
-	if secCtx.IsSolutionUser() || strings.HasPrefix(ownerName, config.RobotPrefix(ctx)) {
+	if secCtx.IsSolutionUser() || secCtx.Name() == "robot" {
 		ownerID = 1
 	} else {
+		ownerName := secCtx.GetUsername()
 		user, err := a.userMgr.GetByName(ctx, ownerName)
 		if err != nil {
 			return a.SendError(ctx, err)
--- a/src/server/v2.0/handler/project.go
+++ b/src/server/v2.0/handler/project.go
@@ -159,7 +159,7 @@ func (a *projectAPI) CreateProject(ctx context.Context, params operation.CreateP
 	// it's a solution to workaround the restriction of project creation API:
 	// only normal users can create projects
 	ownerName := secCtx.GetUsername()
-	if secCtx.IsSolutionUser() || strings.HasPrefix(ownerName, config.RobotPrefix()) {
+	if secCtx.IsSolutionUser() || strings.HasPrefix(ownerName, config.RobotPrefix(ctx)) {
 		ownerID = 1
 	} else {
 		user, err := a.userMgr.GetByName(ctx, ownerName)
--- a/src/server/v2.0/handler/project.go
+++ b/src/server/v2.0/handler/project.go
@@ -101,9 +101,15 @@ func (a *projectAPI) CreateProject(ctx context.Context, params operation.CreateP
 	}
 
 	secCtx, _ := security.FromContext(ctx)
-	if onlyAdmin && !(a.isSysAdmin(ctx, rbac.ActionCreate) || secCtx.IsSolutionUser()) {
+	if onlyAdmin && secCtx.Name() != "robot" && !(a.isSysAdmin(ctx, rbac.ActionCreate) || secCtx.IsSolutionUser()) {
 		log.Errorf("Only sys admin can create project")
 		return a.SendError(ctx, errors.ForbiddenError(nil).WithMessage("Only system admin can create project"))
+	}
+
+	// always check for robots if they have permissions to create projects
+	if secCtx.Name() == "robot" && !a.isSysAdmin(ctx, rbac.ActionCreate) {
+		log.Errorf("Only sys admin can create project")
+                return a.SendError(ctx, errors.ForbiddenError(nil).WithMessage("Only system admin or users can create project"))
 	}
 
 	req := params.Project
--- a/src/server/v2.0/handler/project.go
+++ b/src/server/v2.0/handler/project.go
@@ -101,15 +101,14 @@ func (a *projectAPI) CreateProject(ctx context.Context, params operation.CreateP
 	}
 
 	secCtx, _ := security.FromContext(ctx)
-	if onlyAdmin && secCtx.Name() != "robot" && !(a.isSysAdmin(ctx, rbac.ActionCreate) || secCtx.IsSolutionUser()) {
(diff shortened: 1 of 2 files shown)


heww:
I don't think this is the correct way to assign the creation permission of the project to the robot account.

Please use (`/system/project`, `create`) to check the permission when the security context is not a solution user.

phin1x:
@heww this is not a problem with the actual permission. the rbac check already has passed. at this point the owner of the project is searched. if it is not a solution user, we search for the current authorized user in the user table! but a robot account is not a regular user so the api always repsonds with "user not found". also a robot account does not has a uid, so we have to assign a fix uid (i assigned the admin uid).

src/server/v2.0/handler/project.go
```diff
@@ -156,10 +156,10 @@ func (a *projectAPI) CreateProject(ctx context.Context, params operation.CreateP
 	// set the owner as the system admin when the API being called by replication
 	// it's a solution to workaround the restriction of project creation API:
 	// only normal users can create projects
-	if secCtx.IsSolutionUser() {
+	ownerName := secCtx.GetUsername()
+	if secCtx.IsSolutionUser() || strings.HasPrefix(ownerName, config.RobotPrefix(ctx)) {
```
heww, line 160: Please try to use the security context name to check that the request is from a robot account.

phin1x:
@heww i change the code using the security context name for checking if the current user is a robot acocunt

phin1x:
@heww if think is is already done via the rbac system. project level robot accounts have subjects with a "/project/" prefix, system level robot accounts have subjects with a "/system/" prefix. at the moment, you cannot mix project and system level permissions in a robot account, as we discused in other issues (https://github.com/goharbor/harbor/issues/14512#issuecomment-827062680). Since the rbac subject for creating projects is "/system/project", a project level robot account will never be able to create projects.

heww:
> @heww if think is is already done via the rbac system. project level robot accounts have subjects with a "/project/" prefix, system level robot accounts have subjects with a "/system/" prefix. at the moment, you cannot mix project and system level permissions in a robot account, as we discused in other issues ([#14512 (comment)](https://github.com/goharbor/harbor/issues/14512#issuecomment-827062680)). Since the rbac subject for creating projects is "/system/project", a project level robot account will never be able to create projects.

There is an option to allow everyone to create the project, everyone is the default value for this option in the harbor. When this option is everyone, a project-level robot can create a project.

image

phin1x:
i looked in the code and, yes, you are right. now harbor always checks the permissions when creating a project from a robot account.

wy65701436 (approved):
lgtm

src/server/v2.0/handler/project.go
```diff
@@ -99,11 +99,17 @@ func (a *projectAPI) CreateProject(ctx context.Context, params operation.CreateP
 	}
 
 	secCtx, _ := security.FromContext(ctx)
-	if onlyAdmin && !(a.isSysAdmin(ctx, rbac.ActionCreate) || secCtx.IsSolutionUser()) {
+	if onlyAdmin && secCtx.Name() != "robot" && !(a.isSysAdmin(ctx, rbac.ActionCreate) || secCtx.IsSolutionUser()) {
 		log.Errorf("Only sys admin can create project")
 		return a.SendError(ctx, errors.ForbiddenError(nil).WithMessage("Only system admin can create project"))
 	}
 
+	// always check if robots have permissions to create projects
```
wy65701436, line 107: if !onlyAdmin {
    if it's a project level robot
        return errors.ForbiddenError
}

phin1x: is there a method to check, if the robot acocunt scope is system or project level? i didn'd saw something like a getLevel method.

phin1x:
@wy65701436 i updated the check if the robot account is system level

src/server/v2.0/handler/project.go
```diff
@@ -156,7 +162,7 @@ func (a *projectAPI) CreateProject(ctx context.Context, params operation.CreateP
 	// set the owner as the system admin when the API being called by replication
 	// it's a solution to workaround the restriction of project creation API:
 	// only normal users can create projects
-	if secCtx.IsSolutionUser() {
+	if secCtx.IsSolutionUser() || secCtx.Name() == "robot" {
 		ownerID = 1
```
wy65701436, line 166: Per discussion with maintainers, we should give the owerID to 0 for robot to avoid any confusion for system. Like in the quota page, the `Owner` column will show the wrong message as robot is not a system admin, given it's a system level robot.

phin1x: does the user id 0 exists in the database?

wy65701436: no, it's not in the databsae. Set to 0 is a short term decision. For the long term, we should conside the usage of owner id the Harbor world wide. Whether to reserve it, or sepreate it into two column, reference and reference_type to support both harbor user and robot account.

wy65701436 (changes requested):
please reslove all the review comments before merging.

src/server/v2.0/handler/project.go
wy65701436: @heww is it better to use `Can` to define whether it's a system level robot?

wy65701436: Or here we can leverage the robot name to query robot, there is an attribute to indicate robot scope.

phin1x: the robot context have a isSystemLevel property, but it was not exported. i exported the property with a getter and used that getter.

wy65701436: Another trick way, as the robot name has a pattern, the project level name should be {projectname}+robotname, which is difference with system level robot. And `+` is illegal character when to issue a new robot, so here we can use whether contains `+` to judge.

src/server/v2.0/handler/project.go
phin1x: are we save to use a user id, which is not in the database. could we get any problems for example if harbor tries to resolve the owner name by the id?

src/server/v2.0/handler/project.go
phin1x: imho using a character to identify if a robot is system or project level is not a good idea. if you set the prefix for example to "robot+" the check will fail. the robot security context already has a property which identifies if a robot is system or project level so why not use it?

src/server/v2.0/handler/project.go
phin1x: updated the code, now the owner id will be 0 for robot accounts

phin1x: dco check fixed

phin1x:
@wy65701436 any updates?

wy65701436:
@phin1x can you rebase the PR to resolve the conflict?

phin1x:
@wy65701436 i rebased the pr

wy65701436:
thanks @phin1x , the robot to create project was covered by https://github.com/goharbor/harbor/pull/15461. Now, you can try the latest master, robot can support replication.

PR #14760, opened 2021-04-27
bump Clair to v2.1.7

burdzwastaken:
https://www.redhat.com/security/data/oval/com.redhat.rhsa-RHEL8.xml introduced a "criterion" "comment" entry that started with "Red Hat CoreOS " which violate the assumption that they start with "Red Hat Enterprise Linux ". This causes the pods to constantly restart why trying to update the vulnerability database. The new Clair release fixes this issue.

ref0: https://github.com/quay/clair/releases/tag/v2.1.7
ref1: https://cloud-native.slack.com/archives/CC1E09J6S/p1618994859226300


burdzwastaken:
@reasonerjt could we potentialy get this into 1.10.7? without this fix clair pods are restarting in 1.10.X due to an upstream issue

reasonerjt:
@burdzwastaken unfortunately we are not maintaining 1.10.x for such fixes.  The latest 1.10.x fix is only for fixing CVE in the core components (excl notary, clair...)

Issue #14718, opened 2021-04-21
Push hook event to the queue for retrying later


AyushCloud:
Hello,

I am trying to upgrade the Harbor from v2.1.2 to v2.2.0.
The upgrade is successful. Docker login/push/pull/replication are working fine but i could see the below error in jobservice.log.

Apr 21 09:01:40 ip-172-21-0-1.eu-west-1.compute.internal jobservice[4813]: 2021-04-21T09:01:40Z [WARNING] [/jobservice/hook/hook_agent.go:121]: Send hook event 'status change: job=6e2b37b869efbbf34557a310@1618876800, status=Success' to 'http://core:8080/service/notifications/jobs/adminjob/2' failed with error: {"errors":[{"code":"NOT_FOUND","message":"task with job ID 6e2b37b869efbbf34557a310@1618876800 not found"}]}

Harbor VM 2
Scalable VM 2
Database(Postgres) VM 1
Redis VM 3(Master-Slave-Slave)
Harbor Version v2.2.0
Docker version 19.0.0
docker-compose version 1.26

Could you please help me?

PR #14761, opened 2021-04-27
Read upstream job ID as ID when it isn't null

ywk253100:
Read upstream job ID as ID when it isn't null to fix the job not found error


wy65701436 (approved):
lgtm

bitsf (approved):

ywk253100:
Fixes https://github.com/goharbor/harbor/issues/14718 in v2.2

Issue #14711, opened 2021-04-21
Inconsistent status code for HEAD and GET when the request is sent to public artifact.


reasonerjt:
Such inconsistency was introduced to workaround issue in 
https://github.com/goharbor/harbor/issues/11271

This approach does not follow the HTTP spec and it bites us when Harbor is running behind a CDN, some user reported seeing inconsistent status code for HEAD, possibly due to CDN caching the response header for GET requests.

We should find a way to suit our usecase and not violating the http spec at the same time.
Some mentioned that dockerhub return 401 for GET request issued to public resources.  Harbor may do the same.

We should make sure the change will not break any exisitng use cases.

reasonerjt:
The minimal change in #14768 may fix the problem but this does introduce a break change in the case 
client sends GET request without `Authorization` Header to fetch public artifact, and this should be a release note.

PR #14768, opened 2021-04-28
Return 401 for GET request to /v2 API for public artifacts.

reasonerjt:
This commits make sure when the request does not carry authorization
headers, the HEAD and GET will get the same response code.  This change
should be made due to #14711

### DIFF
--- a/src/server/middleware/v2auth/auth.go
+++ b/src/server/middleware/v2auth/auth.go
@@ -60,7 +60,8 @@ func (rc *reqChecker) check(req *http.Request) (string, error) {
 				return getChallenge(req, al), fmt.Errorf("unauthorized to list catalog")
 			}
 		}
-		if a.target == repository && req.Header.Get(authHeader) == "" && req.Method == http.MethodHead { // make sure 401 is returned for CLI HEAD, see #11271
+		if a.target == repository && req.Header.Get(authHeader) == "" &&
+			(req.Method == http.MethodHead || req.Method == http.MethodGet) { // make sure 401 is returned for CLI HEAD, see #11271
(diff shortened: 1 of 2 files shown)


ywk253100 (approved):

heww (approved):

src/server/middleware/v2auth/auth.go
```diff
@@ -60,7 +60,8 @@ func (rc *reqChecker) check(req *http.Request) (string, error) {
 				return getChallenge(req, al), fmt.Errorf("unauthorized to list catalog")
 			}
 		}
-		if a.target == repository && req.Header.Get(authHeader) == "" && req.Method == http.MethodHead { // make sure 401 is returned for CLI HEAD, see #11271
+		if a.target == repository && req.Header.Get(authHeader) == "" &&
+			(req.Method == http.MethodHead || req.Method == http.MethodGet) { // make sure 401 is returned for CLI HEAD, see #11271
 			return getChallenge(req, al), fmt.Errorf("authorize header needed to send HEAD to repository")
```
karuppiah7890, line 65: The error message here says `authorize header needed to send HEAD to repository` but this is true for `GET` too (with this PR), right? With `(req.Method == http.MethodHead || req.Method == http.MethodGet)`

cc @reasonerjt

karuppiah7890: Should it says `HEAD or GET` ? Or maybe interpolate the `req.Method` in the string for accuracy? Something like this maybe? -

```go
fmt.Errorf("authorize header needed to send %s to repository", req.Method)
```

PR #14662, opened 2021-04-15
Refactor the replication policy destination namespace logic

ywk253100:
Support specifying what part of the repository will be replaced by the provided namespace

Signed-off-by: Wenkai Yin <yinw@vmware.com>

PR #14818, opened 2021-05-07
UI enhancement for replication namespace

### LINKED ISSUE #14741: Support specifying the destination namespace replace count for replication on UI

AllForNothing:
Template Env.: https://10.185.12.35

Changed points:
image

image

Fixes #14741

Related back-end PR #14662


Vad1mo:
We have to work on the UX here!

It took me 1 hour and also looking into code to understand what will actually happen if I change in the UI from -1 to 0. If didn't create a fix that addressed the same problem I wouldn't be able to understand what this functionality is doing.

Vad1mo (changes requested):
Can the default be **1** instead of **-1**? Nobody knows what the legacy mode is.

A user expects that the given copy input produce the same output so for the input **A/B/C** with the argument **N** the outcome would be *N/B/C*

A good analogy is the linux `cp` command `cp /a/* /n/` will behave as we set replication count to 1. 

So it is easier to grasp for the user what happens.

Vad1mo (changes requested):
## Naming

**Replace Count** - raises more question than answers. Replace what? Count what "replaces" ?
**Legacy Mode** - How exactly does the legacy mode behaves? The majority of existing users don't know how the legacy mode behaves.   

GitLab and ACR are referring to "this" as **nested** names. 
GitLab -> Nested Image names
ACR -> Nested namespaces
https://docs.gitlab.com/ee/user/packages/container_registry/#image-naming-convention

In Harbor, it is not clear as it sometimes referred as namespaces and sometimes as projects.

My suggestion is to replace vague term **Replace Count** with a more precise expression.


## Wording

Main objective is to use a wording known and understandable to the users and not describe the implementation detail. 

What actually happening during image copy is not replacing (the image). I assume here the word replacing refers here to the actual feature implementation of string replacement. This is not something the user should know.

From the consumer perspective the images are repositioned or copied and reassembled into a different hierarchy.
The term flattening or tree flattening is often used in this context.  


## Proposed Change

Make it transparent to the user what is about to happen.

From the data entry point of view I don't see the point of having two inputs for one action hence I would only leave the dropdown omitting the input. The reason for not needing the input is that there is no practical sense to have a deep nested repository. (Azure and GitLab for example allow 4 levels only). I guess there is also a string limitation on OCI level. Also, should be become necessary to add additional levels it would be easy to add an option later on. 


- Change wording from **Replace Count** to **Repository Nesting** or **Repository Flattening**
- Just one dropdown with the following values:
  (If **Repository Nesting** is used as label)
    - All Levels // this would represent the int value of 0
    - Omit 1 Level
    - Omit 2 Levels
    - Omit 3 Levels
    - No Nesting (This would equal the legacy mode)
- (if **Repository Flattening** is used as label)
    - No Flatting // this would represent the int value of 0
    - Flatten 1 Level
    - Flatten 2 Levels
    - Flatten 3 Levels
    - Flatten all Levels (This would equal the legacy mode)
    

## Translation

For Nesting
```json
"NAMESPACE": "Namespace",
"NESTING_LEVEL_0": "All Levels",
"NESTING_LEVEL_1": "Omit 1 Level",
"NESTING_LEVEL_2": "Omit 2 Level",
"NESTING_LEVEL_3": "Omit 3 Level",
"NESTING_LEVEL_LEGACY": "No Nesting",
"NESTING_LEVEL_TIP": "Reduce the nested repository structure when copying images. Default 'Omit 1 Level' will replace the left most pepository component with the destination namespace a/b/c/img -> ns/b/c/img. No Nesting - is the legacy option used prior v2.3 resulting in a/b/c/img -> ns/img.",
```

steven-zou:
> ## Naming
> **Replace Count** - raises more question than answers. Replace what? Count what "replaces" ?
> **Legacy Mode** - How exactly does the legacy mode behaves? The majority of existing users don't know how the legacy mode behaves.
> 
> GitLab and ACR are referring to "this" as **nested** names.
> GitLab -> Nested Image names
> ACR -> Nested namespaces
> https://docs.gitlab.com/ee/user/packages/container_registry/#image-naming-convention
> 
> In Harbor, it is not clear as it sometimes referred as namespaces and sometimes as projects.
> 
> My suggestion is to replace vague term **Replace Count** with a more precise expression.
> 
> ## Wording
> Main objective is to use a wording known and understandable to the users and not describe the implementation detail.
> 
> What actually happening during image copy is not replacing (the image). I assume here the word replacing refers here to the actual feature implementation of string replacement. This is not something the user should know.
> 
> From the consumer perspective the images are repositioned or copied and reassembled into a different hierarchy.
> The term flattening or tree flattening is often used in this context.
> 
> ## Proposed Change
> Make it transparent to the user what is about to happen.
> 
> From the data entry point of view I don't see the point of having two inputs for one action hence I would only leave the dropdown omitting the input. The reason for not needing the input is that there is no practical sense to have a deep nested repository. (Azure and GitLab for example allow 4 levels only). I guess there is also a string limitation on OCI level. Also, should be become necessary to add additional levels it would be easy to add an option later on.
> 
> * Change wording from **Replace Count** to **Repository Nesting** or **Repository Flattening**
> * Just one dropdown with the following values:
>   (If **Repository Nesting** is used as label)
>   
>   * All Levels // this would represent the int value of 0
>   * Omit 1 Level
>   * Omit 2 Levels
>   * Omit 3 Levels
>   * No Nesting (This would equal the legacy mode)
> * (if **Repository Flattening** is used as label)
>   
>   * No Flatting // this would represent the int value of 0
>   * Flatten 1 Level
>   * Flatten 2 Levels
>   * Flatten 3 Levels
>   * Flatten all Levels (This would equal the legacy mode)
> 
> ## Translation
> This is a translation example for the case if **nesting** is used as the term
> 
> ```json
> "NAMESPACE": "Namespace",
> "NESTING_LEVEL_0": "All Levels",
> "NESTING_LEVEL_1": "Omit 1 Level",
> "NESTING_LEVEL_2": "Omit 2 Level",
> "NESTING_LEVEL_3": "Omit 3 Level",
> "NESTING_LEVEL_LEGACY": "No Nesting",
> "NESTING_LEVEL_TIP": "Reduce the nested repository structure when copying images. Default 'Omit 1 Level' will replace the left most repository component with the destination namespace a/b/c/img -> ns/b/c/img. No Nesting - is the legacy option used prior v2.3 resulting in a/b/c/img -> ns/img.",
> ```

@Vad1mo 

Thanks for the improvement proposal. It looks good. We'll discuss and follow up later.

AllForNothing:
@Vad1mo After discussion, I've updated the UI according to your advice, thanks a lot!!

Vad1mo:
Splendid work.

AllForNothing:
> is this planned to be shipped with 2.3?

Yes, the related issue is labeled with "target 2.3"

xuelichao (approved):

jwangyangls (approved):

PR #14829, opened 2021-05-09
Move user related funcs from common/dao

reasonerjt:
This commit moves more user related funcs, such as ChangePassword,
Login, ChangeUserProfile from common/dao to rely on /pkg/user and
pkg/oidc.
It also removes the code for resetting user's password as it's disabled.

### DIFF
--- a/src/pkg/user/manager.go
+++ b/src/pkg/user/manager.go
@@ -35,7 +35,7 @@ var (
 type Manager interface {
 	// Get get user by user id
 	Get(ctx context.Context, id int) (*models.User, error)
-	// GetByName get user by username
+	// GetByName get user by username, it will return an error if the user does not exist
 	GetByName(ctx context.Context, username string) (*models.User, error)
 	// List users according to the query
 	List(ctx context.Context, query *q.Query) (models.Users, error)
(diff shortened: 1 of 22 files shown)


src/pkg/user/manager.go
```diff
@@ -75,20 +76,33 @@ func (m *manager) Delete(ctx context.Context, id int) error {
...
+		return nil, err
 	}
-	return utils.Encrypt(password, u.Salt, u.PasswordVersion) == u.Password, nil
+	l = append(l, u)
+	for _, entry := range l {
+		if utils.Encrypt(password, entry.Salt, entry.PasswordVersion) == entry.Password {
+			return entry, nil
+		}
+	}
+	return nil, nil
```
heww, line 94: There is a special char `@` in the email, it is invalid in the username.

wy65701436 (approved):
lgtm

src/pkg/user/manager.go
reasonerjt: But this is for querying?
I don't recall how the decision was made, but this is for compatibility with `LoginByDb` in `common/dao` pkg:
https://github.com/goharbor/harbor/blob/release-2.2.0/src/common/dao/user.go#L81

heww: There are two db queries in this method `MatchLocalPassword `, only one db query in `LoginByDb`.

And when `usernameOrEmail` is a email, `u, err := m.GetByName(ctx, usernameOrEmail)` will must get a not found error.

heww (approved):

src/pkg/user/manager.go
reasonerjt: I have added a `FilterBy...` func to user model.

PR #14837, opened 2021-05-10
perf: skip db tx for get, head and options api requests

### LINKED ISSUE #14109: Avoid to use orm transaction in GET request
After Harbor 2.x, almost every request there is an orm transaction, it consumes too many db connection resources, for all GET request, because it is idempotent, there is no need to start a transaction. it could improve the Harbor performance.

heww:
Closes #14109

### DIFF
--- a/src/core/middlewares/middlewares.go
+++ b/src/core/middlewares/middlewares.go
@@ -15,7 +15,6 @@
 package middlewares
 
 import (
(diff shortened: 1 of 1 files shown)


stonezdj (approved):

wy65701436 (approved):
lgtm

PR #14846, opened 2021-05-11
support pg upgrade

wy65701436:
1, use the pg source and photon spec to build postgres 9.6
2, install 9.6 on the photon 4.0
3, then leverage pg_upgrade to handle the pg major version migration

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -429,9 +429,12 @@ build_base_docker:
 	@for name in $(BUILDBASETARGET); do \
 		echo $$name ; \
 		sleep 30 ; \
+		if [ $$name == "db" ] ; then \
+			cd $(MAKEFILEPATH_PHOTON)/$$name && $(MAKEFILEPATH_PHOTON)/$$name/rpm_builder.sh && cd - ; \
+		fi; \
 		$(DOCKERBUILD) --pull --no-cache -f $(MAKEFILEPATH_PHOTON)/$$name/Dockerfile.base -t $(BASEIMAGENAMESPACE)/harbor-$$name-base:$(BASEIMAGETAG) --label base-build-date=$(date +"%Y%m%d") . && \
 		if [ -n "$(PUSHBASEIMAGE)" ] ; then \
-			$(PUSHSCRIPTPATH)/$(PUSHSCRIPTNAME) $(BASEIMAGENAMESPACE)/harbor-$$name-base:$(BASEIMAGETAG) $(REGISTRYUSER) $(REGISTRYPASSWORD) || exit 1; \
+		  	$(PUSHSCRIPTPATH)/$(PUSHSCRIPTNAME) $(BASEIMAGENAMESPACE)/harbor-$$name-base:$(BASEIMAGETAG) $(REGISTRYUSER) $(REGISTRYPASSWORD) || exit 1; \
 		fi ; \
 	done
 
--- a/make/photon/db/Dockerfile
+++ b/make/photon/db/Dockerfile
@@ -5,6 +5,8 @@ FROM ${harbor_base_namespace}/harbor-db-base:${harbor_base_image_version}
 VOLUME /var/lib/postgresql/data
 
 COPY ./make/photon/db/docker-entrypoint.sh /docker-entrypoint.sh
+COPY ./make/photon/db/initdb.sh /initdb.sh
+COPY ./make/photon/db/upgrade.sh /upgrade.sh
 COPY ./make/photon/db/docker-healthcheck.sh /docker-healthcheck.sh
 COPY ./make/photon/db/initial-notaryserver.sql /docker-entrypoint-initdb.d/
 COPY ./make/photon/db/initial-notarysigner.sql /docker-entrypoint-initdb.d/
@@ -12,7 +14,7 @@ COPY ./make/photon/db/initial-registry.sql /docker-entrypoint-initdb.d/
 RUN chown -R postgres:postgres /docker-entrypoint.sh /docker-healthcheck.sh /docker-entrypoint-initdb.d \
     && chmod u+x /docker-entrypoint.sh /docker-healthcheck.sh
 
-ENTRYPOINT ["/docker-entrypoint.sh"]
+ENTRYPOINT ["/docker-entrypoint.sh", "96", "13"]
 HEALTHCHECK CMD ["/docker-healthcheck.sh"]
 
 USER postgres
--- a/make/photon/db/docker-entrypoint.sh
+++ b/make/photon/db/docker-entrypoint.sh
@@ -1,114 +1,48 @@
 #!/bin/bash
 set -e
 
-# usage: file_env VAR [DEFAULT]
-#    ie: file_env 'XYZ_DB_PASSWORD' 'example'
-# (will allow for "$XYZ_DB_PASSWORD_FILE" to fill in the value of
-#  "$XYZ_DB_PASSWORD" from a file, especially for Docker's secrets feature)
-file_env() {
-        local var="$1"
-        local fileVar="${var}_FILE"
-        local def="${2:-}"
-        if [ "${!var:-}" ] && [ "${!fileVar:-}" ]; then
-                echo >&2 "error: both $var and $fileVar are set (but are exclusive)"
-                exit 1
-        fi
-        local val="$def"
-        if [ "${!var:-}" ]; then
-                val="${!var}"
-        elif [ "${!fileVar:-}" ]; then
-                val="$(< "${!fileVar}")"
-        fi
-        export "$var"="$val"
-        unset "$fileVar"
-}
+source $PWD/initdb.sh
 
-# look specifically for PG_VERSION, as it is expected in the DB dir
-if [ ! -s "$PGDATA/PG_VERSION" ]; then
-        file_env 'POSTGRES_INITDB_ARGS'
-        if [ "$POSTGRES_INITDB_XLOGDIR" ]; then
-                export POSTGRES_INITDB_ARGS="$POSTGRES_INITDB_ARGS --xlogdir $POSTGRES_INITDB_XLOGDIR"
-        fi
-        initdb -D $PGDATA  -U postgres -E UTF-8 --lc-collate=en_US.UTF-8 --lc-ctype=en_US.UTF-8 $POSTGRES_INITDB_ARGS
-        # check password first so we can output the warning before postgres
-        # messes it up
-        file_env 'POSTGRES_PASSWORD'
-        if [ "$POSTGRES_PASSWORD" ]; then
-                pass="PASSWORD '$POSTGRES_PASSWORD'"
-                authMethod=md5
+CUR=$PWD
+PG_VERSION_OLD=$1
+PG_VERSION_NEW=$2
+
+PGBINOLD="/usr/local/pg${PG_VERSION_OLD}/bin"
+
+PGDATAOLD=${PGDATA}/pg${PG_VERSION_OLD}
+PGDATANEW=${PGDATA}/pg${PG_VERSION_NEW}
+# to handle the PG 9.6 only
+if [ -s $PGDATA/PG_VERSION ]; then
+         PGDATAOLD=$PGDATA
+fi
+
+#
+# Init DB: $PGDATA is empty.
+# Upgrade DB: 1, has $PGDATA\PG_VERSION. 2, has pg old version directory with PG_VERSION inside.
+#
+if [ "$(ls -A $PGDATA)" ]; then
+        if [ ! -d $PGDATANEW ]; then
+                if [ ! -d $PGDATAOLD ] || [ ! -s $PGDATAOLD/PG_VERSION ]; then
+                        echo "incorrect data: $PGDATAOLD, make sure $PGDATAOLD is not empty and with PG_VERSION inside."
+                        exit 1
+                fi
+                initPG $PGDATANEW false
+                # because of the 'set -e', the upgrade failure will break execution.
+                ./$CUR/upgrade.sh --old-bindir $PGBINOLD --old-datadir $PGDATAOLD --new-datadir $PGDATANEW
+                echo "remove the $PGDATAOLD after upgrade success."
+                if [ "$PGDATAOLD" = "$PGDATA" ]; then
+                        find $PGDATA/* -prune ! -name pg${PG_VERSION_NEW} -exec rm -rf {} \;
+                else
+                        rm -rf $PGDATAOLD
+                fi
         else
-                # The - option suppresses leading tabs but *not* spaces. :)
-                cat >&2 <<-EOF
-                        ****************************************************
-                        WARNING: No password has been set for the database.
-                                        This will allow anyone with access to the
-                                        Postgres port to access your database. In
-                                        Docker's default configuration, this is
-                                        effectively any other container on the same
-                                        system.
-                                        Use "-e POSTGRES_PASSWORD=password" to set
-                                        it in "docker run".
-                        ****************************************************
-EOF
-
-                pass=
-                authMethod=trust
+                echo "no need to upgrade postgres, launch it."
         fi
-
-        {
-                echo
-                echo "host all all all $authMethod"
-        } >> "$PGDATA/pg_hba.conf"
-        echo `whoami`
-        # internal start of server in order to allow set-up using psql-client
-        # does not listen on external TCP/IP and waits until start finishes
-        pg_ctl -D "$PGDATA" -o "-c listen_addresses=''" -w start
-
-        file_env 'POSTGRES_USER' 'postgres'
-        file_env 'POSTGRES_DB' "$POSTGRES_USER"
-
-        psql=( psql -v ON_ERROR_STOP=1 )
-
-        if [ "$POSTGRES_DB" != 'postgres' ]; then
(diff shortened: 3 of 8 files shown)


make/photon/db/npm_builder.sh
```diff
@@ -0,0 +1,27 @@
+set -e
```
reasonerjt, line 1: Incorrect filename

wy65701436: updated.

make/photon/db/rpm_builder.sh
```diff
@@ -0,0 +1,27 @@
...
+cur=$PWD
+
+workDir='pg_builder'
+rm -rf $workDir && mkdir -p $workDir && cd ./$workDir
+
+# step 1: get source code of pg 9.6, and rename the code directory from postgres to postgres96
+wget http://ftp.postgresql.org/pub/source/v$version/$name-$version.tar.bz2
+bzip2 -d ./$name-$version.tar.bz2 && tar -xvf ./$name-$version.tar
+mkdir -p ${name}96-$version && cd ./$name-$version && cp -rf ./* ../${name}96-$version && cd .. && rm -rf ./$name-$version
+tar -cvjSf ${name}96-$version.tar.bz2 ${name}96-$version
```
reasonerjt, line 15: 1)  just `cp -r ./$name-$version/* ./${name}96-$version/` and  you don't need `cd ./$name-$version` and `cd ..`
2) why sometimes use `./` sometime does not?

make/photon/db/rpm_builder.sh
```diff
@@ -0,0 +1,27 @@
+set -e
+
+name='postgresql'
+version='9.6.21'
+
+cur=$PWD
+
+workDir='pg_builder'
```
reasonerjt, line 8: we don't use camel case in `.sh`

make/photon/db/rpm_builder.sh
```diff
@@ -0,0 +1,27 @@
...
+rm -rf $workDir && mkdir -p $workDir && cd ./$workDir
```
reasonerjt, line 9: There's risk the directory on user's host is removed.

make/photon/db/Dockerfile
```diff
@@ -4,6 +4,12 @@ FROM ${harbor_base_namespace}/harbor-db-base:${harbor_base_image_version}
 
 VOLUME /var/lib/postgresql/data
 
+ENV PGBINOLD /usr/local/pg96/bin/
```
reasonerjt, line 7: Do we need these envs set at the container/image level?

make/photon/db/Dockerfile
```diff
@@ -4,6 +4,12 @@ FROM ${harbor_base_namespace}/harbor-db-base:${harbor_base_image_version}
...
+ENV PGBINNEW /usr/bin
+
+ENV PGDATAOLD /var/lib/postgresql/data/9.6
```
reasonerjt, line 10: how about  `/tmp/pg/v9.6` and `/tmp/pg/v13`?
And things may work with `mv` instead of having to use `-prune ! name` parms.

make/photon/db/rpm_builder.sh
```diff
@@ -0,0 +1,27 @@
+set -e
```
reasonerjt, line 1: To simplify the verification.  It should require bash?

Makefile
```diff
@@ -415,9 +415,12 @@ build_base_docker:
 	@for name in $(BUILDBASETARGET); do \
 		echo $$name ; \
 		sleep 30 ; \
+		if [ $$name == "db" ] ; then \
+			cd $(MAKEFILEPATH_PHOTON)/$$name && $(MAKEFILEPATH_PHOTON)/$$name/rpm_builder.sh && cd - ; \
+		fi; \
```
heww, line 420: Why not create a target for the rpm building and let the `build_base_docker` target depend on the target?

make/photon/db/docker-entrypoint.sh
```diff
@@ -1,6 +1,18 @@
 #!/bin/bash
 set -e
 
+PG_VERSION_NEW=$1
```
heww, line 4: `PG_VERSION_NEW` => `PG_VERSION` ? From the point of the `harbor-db` , 13 is the current version of the `harbor-db` not the`new` version.

make/photon/db/docker-entrypoint.sh
```diff
@@ -89,26 +101,61 @@ EOSQL
...
+                  --new-options '-c config_file=$PGDATANEW/postgresql.conf'
+                mv $PGDATAOLD/pg_hba.conf $PGDATANEW/pg_hba.conf
+
+                echo 'PostgreSQL start to move $PGDATANEW to $PGDATA.'
+                mv -f $PGDATANEW/* $PGDATA
+                rm -rf $PGDATAOLD;
+                rm -rf $PGDATANEW;
+        fi
+else
+        initPG $PGDATA true
```
heww, line 155: Could add some comments for arguments of the initPG function? Especially the second argument.

make/photon/db/docker-entrypoint.sh
```diff
@@ -89,26 +101,61 @@ EOSQL
...
+}
+
+# look specifically for PG_VERSION, if exist and needs upgrade, use pg_upgrade to handle it. Otherwise just init db.
+if [ -s $PGDATA/PG_VERSION ]; then
+        PG_VERSION_OLD=$(cat $PGDATA/PG_VERSION);
+        if [ "`echo "${PG_VERSION_OLD} < $PG_VERSION_NEW" | bc`" -eq 1 ]; then
+                echo 'PostgreSQL start to move $PGDATA to $PGDATAOLD.'
+                mkdir -p $PGDATAOLD
+                chown -R postgres "$PGDATAOLD"
+                mv -f $PGDATA/* $PGDATAOLD
```
heww, line 133: What will happen using mv here when the upgrade process failed?

make/photon/db/docker-entrypoint.sh
```diff
@@ -1,6 +1,18 @@
 #!/bin/bash
 set -e
 
+PG_VERSION_NEW=$1
+
+PGBINOLD="/usr/local/pg96/bin/"
+PGBINNEW="/usr/bin"
+PGDATAOLD="/tmp/pg/v9.6"
+PGDATANEW="/tmp/pg/v13"
```
heww, line 9: Copy database files from volume mounted from the host directory to the directory of the container may take the disk during the container is running even the files in `/tmp/pg` are deleted after the upgrade.

make/photon/db/docker-entrypoint.sh
reasonerjt: how about cp to the `/tmp` directory, and let the migration happen in the `/tmp` directory, and after the migration is successful, replace the content in `PGDATA` with the migrated data?

make/photon/db/docker-entrypoint.sh
```diff
@@ -1,120 +1,59 @@
...
+
+#
+# Init DB: $PGDATA is empty.
+# Upgrade DB: 1, has $PGDATA\PG_VERSION. 2, has pg version directory with PG_VERSION inside.
+# Error: $PGDATA is not empty and has files or invalid directory.
+#
+if [ "$(ls -A $PGDATA)" ]; then
+        if [ ! -d $PGDATANEW ]; then
+                TO_UPGRADE=true
+                if [ -s $PGDATA/PG_VERSION ]; then
```
reasonerjt, line 21: I think you can search for the file `PG_VERSION` under $PGDATA, and use the `dirname` of the file as `$PGDATAOLD`, such that you don't need the `if...else...` in line 21~32

make/photon/db/docker-entrypoint.sh
```diff
@@ -1,120 +1,59 @@
...
+        initPG $PGDATANEW false
+        bash "$CUR/upgrade.sh" --old-datadir $PGDATAOLD --new-datadir $PGDATANEW
+        if [ $? -ne 0 ]; then
+                exit 1
         fi
-
-        if [ "$POSTGRES_USER" = 'postgres' ]; then
-                op='ALTER'
+        if [ "$PGDATAOLD" = "$PGDATA" ]; then
+                find $PGDATA/* -prune ! -name pg${PG_VERSION_NEW} -exec rm -rf {} \;
```
reasonerjt, line 45: `! -name $PGDATANEW` ?

make/photon/db/docker-entrypoint.sh
```diff
@@ -1,120 +1,59 @@
...
+if [ $TO_UPGRADE == true ]; then
+        initPG $PGDATANEW false
+        bash "$CUR/upgrade.sh" --old-datadir $PGDATAOLD --new-datadir $PGDATANEW
+        if [ $? -ne 0 ]; then
+                exit 1
         fi
-
-        if [ "$POSTGRES_USER" = 'postgres' ]; then
-                op='ALTER'
+        if [ "$PGDATAOLD" = "$PGDATA" ]; then
```
reasonerjt, line 44: print a log for removing the `$PGDATAOLD`?

make/photon/db/upgrade.sh
```diff
@@ -0,0 +1,50 @@
...
+        exit 1
+fi
+
+export PGDATAOLD=$PGDATAOLD
+export PGDATANEW=$PGDATANEW
+export PGBINNEW=$PGBINNEW
+export PGBINOLD=$PGBINOLD
+
+echo 'start to upgrade.'
+cd /tmp
```
reasonerjt, line 34: is this needed?

make/photon/db/upgrade.sh
```diff
@@ -0,0 +1,50 @@
...
+  --old-options '-c config_file=$PGDATAOLD/postgresql.conf' \
+  --new-options '-c config_file=$PGDATANEW/postgresql.conf'
+
+if [ $? -ne 0 ]; then
+        echo 'fail to upgrade.'
+        cat /tmp/pg_upgrade_internal.log
+        exit 1
+fi
+
+mv $PGDATAOLD/pg_hba.conf $PGDATANEW/pg_hba.conf
```
reasonerjt, line 49: it should be `cp` as we discussed we should persist old data until upgrade is successful?

make/photon/db/upgrade.sh
```diff
@@ -0,0 +1,50 @@
...
+                -B|--new-datadir) PGDATANEW="$2"; shift ;;
+                -d|--old-bindir) PGBINOLD="$2"; shift ;;
+                -D|--new-bindir) PGBINNEW="$2"; shift ;;
+                *) echo "Unknown parameter passed: $1"; exit 1 ;;
+        esac
+        shift
+done
+
+if [ "$PGDATAOLD" = "" ] || [ "$PGDATANEW" = "" ]; then
+        echo "Required parameter is missing: $PGDATAOLD, $PGDATANEW"
```
reasonerjt, line 24: wrong message?

make/photon/db/initdb.sh
```diff
@@ -0,0 +1,112 @@
...
+        if [ "${!var:-}" ]; then
+                val="${!var}"
+        elif [ "${!fileVar:-}" ]; then
+                val="$(< "${!fileVar}")"
+        fi
+        export "$var"="$val"
+        unset "$fileVar"
+}
+
+# usage: initPG $Dir
```
reasonerjt, line 26: Add comment about the 2nd parm?

make/photon/db/upgrade.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+set -e
+
+source $PWD/initdb.sh
+
+PGDATA="/var/lib/postgresql/data"
```
reasonerjt, line 6: is this `PGDATA` needed?

make/photon/db/upgrade.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+set -e
```
reasonerjt, line 2: if you use `set -e`

There may not be a chance to see the log messages in `/tmp` directory?

make/photon/db/docker-entrypoint.sh
```diff
@@ -1,120 +1,59 @@
...
-
-        if [ "$POSTGRES_DB" != 'postgres' ]; then
-                "${psql[@]}" --username postgres <<-EOSQL
-                        CREATE DATABASE "$POSTGRES_DB" ;
-EOSQL
-                echo
+if [ $TO_UPGRADE == true ]; then
+        initPG $PGDATANEW false
+        bash "$CUR/upgrade.sh" --old-datadir $PGDATAOLD --new-datadir $PGDATANEW
+        if [ $? -ne 0 ]; then
```
reasonerjt, line 41: with `set -e` this is not needed?

make/photon/db/upgrade.sh
wy65701436: yes, this should be removed.

make/photon/db/docker-entrypoint.sh
wy65701436: added

make/photon/db/docker-entrypoint.sh
wy65701436: added

make/photon/db/upgrade.sh
wy65701436: the pg_upgrade log will be generated in the tmp folder, only failure will be printed.

make/photon/db/upgrade.sh
wy65701436: sure

make/photon/db/initdb.sh
wy65701436: added

make/photon/db/upgrade.sh
wy65701436: removed

make/photon/db/docker-entrypoint.sh
wy65701436: remove the `set -e` from upgrade.sh

heww (approved):

make/photon/db/docker-entrypoint.sh
wy65701436: updated.

make/photon/db/initdb.sh
```diff
@@ -0,0 +1,114 @@
...
+        elif [ "${!fileVar:-}" ]; then
+                val="$(< "${!fileVar}")"
+        fi
+        export "$var"="$val"
+        unset "$fileVar"
+}
+
+# usage: initPG $Dir $initSql
+#   Use $Dir to index where to init the postgres db
+#   Use $initSql to index whether to execute the sql under docker-entrypoint-initdb.d, default is not.
```
reasonerjt, line 28: what do you mean by `index`?

make/photon/db/docker-entrypoint.sh
```diff
@@ -1,120 +1,54 @@
...
-        {
-                echo
-                echo "host all all all $authMethod"
-        } >> "$PGDATA/pg_hba.conf"
-        echo `whoami`
-        # internal start of server in order to allow set-up using psql-client
-        # does not listen on external TCP/IP and waits until start finishes
-        pg_ctl -D "$PGDATA" -o "-c listen_addresses=''" -w start
+CUR=$PWD
+TO_UPGRADE=false
```
reasonerjt, line 7: TO_UPGRADE is not used.

make/photon/db/Dockerfile
```diff
@@ -5,14 +5,16 @@ FROM ${harbor_base_namespace}/harbor-db-base:${harbor_base_image_version}
...
+COPY ./make/photon/db/upgrade.sh /upgrade.sh
 COPY ./make/photon/db/docker-healthcheck.sh /docker-healthcheck.sh
 COPY ./make/photon/db/initial-notaryserver.sql /docker-entrypoint-initdb.d/
 COPY ./make/photon/db/initial-notarysigner.sql /docker-entrypoint-initdb.d/
 COPY ./make/photon/db/initial-registry.sql /docker-entrypoint-initdb.d/
 RUN chown -R postgres:postgres /docker-entrypoint.sh /docker-healthcheck.sh /docker-entrypoint-initdb.d \
     && chmod u+x /docker-entrypoint.sh /docker-healthcheck.sh
 
-ENTRYPOINT ["/docker-entrypoint.sh"]
+ENTRYPOINT ["/docker-entrypoint.sh", "96", "13"]
```
reasonerjt, line 17: Will it still need 2 parms in next version when there's no upgrade needed?

make/photon/db/docker-entrypoint.sh
```diff
@@ -1,120 +1,54 @@
...
+#
+if [ "$(ls -A $PGDATA)" ]; then
+        if [ ! -d $PGDATANEW ]; then
+                if [ ! -d $PGDATAOLD ] || [ ! -s $PGDATAOLD/PG_VERSION ]; then
+                        echo "incorrect data: $PGDATAOLD, make sure $PGDATAOLD is not empty and with PG_VERSION inside."
+                        exit 1
+                fi
+                initPG $PGDATANEW false
+                ./$CUR/upgrade.sh --old-bindir $PGBINOLD --old-datadir $PGDATAOLD --new-datadir $PGDATANEW
+                if [ $? -ne 0 ]; then
```
reasonerjt, line 32: This is not needed as there's `set -e`

reasonerjt (approved):

PR #14860, opened 2021-05-12
[cherry-pick]Fix duplicate execution record issue

ywk253100:
When the core service cannot response the checkin request in time, duplicated execution records may be created, this commit introduces the revision column to make sure there is only one record for one schedule trigger


steven-zou (approved):
LGTM

bitsf (approved):

PR #14875, opened 2021-05-13
fix: using new ctx to scan artifact when scanAll

heww:
DB transaction failure may break the loop query of the artifacts, the
result is that not all artifacts are scanned in one scan all job. Using
a new DB connection to call the Scan method of the controller to avoid
this problem.


wy65701436 (approved):
lgtm

reasonerjt (approved):

PR #14896, opened 2021-05-17
Add IPv6 support

ninjadq:
* 127.0.0.1 to localhost
* listening net addr add ipv6 format
* add ipv6 to test cert generating script

### DIFF
--- a/make/photon/prepare/templates/docker_compose/docker-compose.yml.jinja
+++ b/make/photon/prepare/templates/docker_compose/docker-compose.yml.jinja
@@ -22,6 +22,7 @@ services:
         target: /etc/rsyslog.d/rsyslog_docker.conf
     ports:
       - 127.0.0.1:1514:10514
+      - ::1:1514:10514
     networks:
       - harbor
   registry:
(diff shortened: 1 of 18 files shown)


make/photon/prepare/templates/docker_compose/docker-compose.yml.jinja
```diff
@@ -22,6 +22,7 @@ services:
         target: /etc/rsyslog.d/rsyslog_docker.conf
     ports:
       - 127.0.0.1:1514:10514
+      - ::1:1514:10514
```
bitsf, line 25: is it possible to use [::1] here

bitsf (approved):

ywk253100:
@ninjadq I think there are still some `127.0.0.1` in the go files. Please make sure all the `127.0.0.1` are replaced
e.g. 
https://github.com/goharbor/harbor/blob/master/src/pkg/reg/adapter/harbor/base/client.go#L147
https://github.com/goharbor/harbor/blob/master/src/lib/config/metadata/metadatalist.go#L73

Issue #14814, opened 2021-05-06
Make the P95 of list-artifacts to be 23s in v2.3


heww:
In the small size of user data, the P95 of list-artifacts API is about 56s  in 300 concurrencies, we need to make it to be 23s.

| API               | Avg    | Min    | Med    | Max    | P(90)  | P(95)  | Success Rate | Iterations Rate |
| ----------------- | ------ | ------ | ------ | ------ | ------ | ------ | ------------ | ---------------- |
| list-artifacts         | 32.57s | 0.04ms | 30.51s  |  60.1s  | 51.36s  | 56.28s  | 99.83%      | 8.74/s         |

heww:
Closed by #14925

Issue #14813, opened 2021-05-06
Make the success rate of list-repositories API be 100% in v2.3


heww:
In the small size of user data, the success rate of list-repositories API is almost zero in 300 concurrencies, we need to make the success rate to 100%.

| API               | Avg    | Min    | Med    | Max    | P(90)  | P(95)  | Success Rate | Iterations Rate |
| ----------------- | ------ | ------ | ------ | ------ | ------ | ------ | ------------ | ---------------- |
| list-repositories | 60.45s | 0.06ms | 60.34s | 61.41s | 61.36s | 61.36s | 0.33%        | 4.92/s           |

heww:
Closed by #14925

PR #14899, opened 2021-05-17
perf: improve the perf of count and list artifacts dao

### LINKED ISSUE #13890: Postgresql has many query connections and causes high load of node
I found so many query connections in  pg container, It is more than 300.  and pg consumes 7 cpu . It causes hight load of node . The load value is  almost 200+.

**Versions:**

- harbor version: 2.0.1
- docker engine version: server 19.03.9
- k8s version: v1.16.6

**Additional context:**
pg container stats:
```
CONTAINER ID        NAME                                                                                  CPU %               MEM USAGE / LIMIT   MEM %               NET I/O             BLOCK I/O           PIDS
dd409f20aaf0        k8s_database_harbor-harbor-database-0_harbor_8bcbba28-05a2-4368-af8e-b98d80503cc9_0   588.74%             98.19MiB / 4GiB     2.40%               0B / 0B             0B / 0B             54
```
row count in  pg table: 
`repository`:  1500+
`tag`:  50k
`blob`: 300k
`artifact `: 50k

current job in harbor:
- docker pull  & docker push from docker client on node
- no replication 
- no scan 
  
current query in postgress:
[harbor-pg-query.sql.txt](https://github.com/goharbor/harbor/files/5768449/harbor-pg-query.sql.txt)
[harbor-query.txt](https://github.com/goharbor/harbor/files/5768461/harbor-query.txt)

heww:
Closes #13890 #14813 #14814


codergs:
@heww do you have the new perf stats from before and after?

heww:
> @heww do you have the new perf stats from before and after?

Before
|API                 |Avg     |Min     |Med     |Max     |P(90)   |P(95)   |Success Rate|Iterations Rate|
| -- | -- | -- | -- | -- | -- | -- | -- | -- |
|list-artifact-tags  |1.97s   |0.04ms  |1.84s   |4s      |3.04s   |3.44s   |100%        |136.70/s       |
|list-artifacts      |25.25s  |0.06ms  |23.99s  |39.31s  |34.54s  |36.01s  |100%        |10.77/s        |
|list-repositories   |59.95s  |0.06ms  |60.32s  |60.75s  |60.71s  |60.72s  |2%          |4.95/s         |

After
|API                 |Avg     |Min    |Med     |Max     |P(90)   |P(95)   |Success Rate|Iterations Rate|
| -- | -- | -- | -- | -- | -- | -- | -- | -- |
|list-artifact-tags  |822.15ms|0.09ms |731.18ms|1.71s   |1.48s   |1.61s   |100%        |334.54/s       |
|list-artifacts      |2.36s   |0.04ms |2.38s   |3.95s   |3.17s   |3.37s   |100%        |119.27/s       |
|list-repositories   |464.37ms|0.06ms |372.82ms|1.82s   |898.26ms|948.49ms|100%        |324.88/s       |

codergs:
@heww awesome stuff. great for documentation + performance gains are really awesome. thanks for your hard work :)

src/pkg/artifact/dao/dao.go
```diff
@@ -251,131 +241,89 @@ func (d *dao) DeleteReferences(ctx context.Context, parentID int64) error {
...
-	if !ok || b != "*" {
-		return qs, errors.New(nil).WithCode(errors.BadRequestCode).
-			WithMessage(`the value of "base" query can only be exact match value with "*"`)
+func filterBase(query *q.Query) selector.Option {
+	return func(builder *selector.Builder, cond *selector.Cond) error {
+		base, exist := query.GetValue("base")
+		if !exist {
+			builder.Where(
+				cond.Or(
+					cond.Exists("SELECT 1 FROM tag WHERE tag.artifact_id = T0.id"),
```
ywk253100, line 254: Is possible to hide the `T0` which is the implementation detail of the `selector`?

src/pkg/artifact/dao/dao.go
```diff
@@ -79,29 +59,39 @@ func New() DAO {
...
-	qs, err := querySetter(ctx, query)
-	if err != nil {
-		return nil, err
-	}
-	if _, err = qs.All(&artifacts); err != nil {
-		return nil, err
+	model := &Artifact{}
+
+	options := []selector.Option{
+		selector.From(model),
```
ywk253100, line 79: Seems every function call for `selector` we need to pass the `model` as a parameter, is it possible to avoid this?

src/pkg/artifact/dao/dao.go
```diff
@@ -251,131 +241,89 @@ func (d *dao) DeleteReferences(ctx context.Context, parentID int64) error {
...
+		// "*" for listing tagged artifacts
+		// "nil" for listing untagged artifacts
+		// others for get the artifact with the specified tag
+		if s, ok := tags.(string); ok {
+			if s == "*" {
+				builder.Where(cond.Exists("SELECT 1 FROM tag WHERE tag.artifact_id = T0.id"))
+			} else if s == "nil" {
+				builder.Where(cond.NotExists("SELECT 1 FROM tag WHERE tag.artifact_id = T0.id"))
+			} else {
+				builder.Join("tag AS t", "t.artifact_id = T0.id", cond.Equal("t.name", s))
```
ywk253100, line 296: Is the sql injection handled by the framework?

src/lib/orm/metadata.go
```diff
@@ -30,33 +30,74 @@ var (
...
 	Filterable bool
 	FilterFunc func(context.Context, orm.QuerySeter, string, interface{}) orm.QuerySeter
 	Sortable   bool
 }
 
-type metadata struct {
-	Keys         map[string]*key
-	DefaultSorts []*q.Sort
+// Column the columns of the model
+type Column struct {
```
ywk253100, line 42: Seems the `Column` is dup of `Key`?

src/lib/orm/metadata.go
```diff
@@ -30,33 +30,74 @@ var (
...
+	Sortable   bool   // True when the column is orderable
 }
 
-func (m *metadata) Filterable(key string) (*key, bool) {
+// Metadata metadata of model
+type Metadata struct {
+	Keys map[string]*Key
+
+	columnIndexes map[string]*Column
+	Columns       []*Column // ordered columns
```
ywk253100, line 53: What's the different of `Keys`, `columnIndexes` and `Columns`?

src/pkg/artifact/dao/dao.go
heww: No, it's not possible to hide the `T0`. When applying `JOIN tag` and filter by `repository_id` to the select, the SQL doesn't know where the repository_id comes from.

src/pkg/artifact/dao/dao.go
heww: `cond.Equal("t.name", s)` will generate string `t.name = ?`

src/lib/orm/metadata.go
heww: I will remove the Keys in the metadata.

`Columns` keeps the order for columns, the selector will use this to generate a stable SQL statement (means using the same model and filters will generate the same SQL).  `map` doesn't work for this case because iterations of the map in golang are random seq when using `range`.

`columnIndexes` is used to get the Column by the db column name or model FieldName. And the items in the `Columns` are all tracked by the db column name, on model FieldName in it.

While, there FieldName, db column name, filter by method in the Keys is not very clear.

src/pkg/artifact/dao/dao.go
heww: Although we only use `selector` in the dao of artifacts that using the `selector.From`, but it also can work with raw SQL. For example,

```go
func undeleted (b, c) nil {
     b.Where( c.IsNull("deleted_at",))
}

selector.New(ctx, func(b, c) nil {
   b.From("tablea").Where("col1", c.E("c1"))
},
undeleted,
)
```

selector.From is just a helper to apply `FROM tablename` for the model.

src/lib/orm/metadata.go
heww: `selector` wants to know the all columns in the table for the model, it's not possible by the Keys because there are `FieldName`, `FilterByXxx` in the Keys.

heww: `Key` was removed, now the `Column` only be the fields info from the model.

src/lib/orm/metadata.go
heww: Now, `Columns` only be the info parsed from the fields of the model and the filterFuncs be the info parsed from the filter methods and filter fields of the model.

heww:
Closed as another PR #14925 will fix the issues.

PR #14900, opened 2021-05-17
Remove GetUser and Onboard from common/dao

reasonerjt:
Replaced by funcs in src/pkg/user and src/controller/user


src/pkg/user/manager.go
```diff
@@ -65,6 +69,27 @@ type manager struct {
 	dao dao.DAO
 }
 
+func (m *manager) Onboard(ctx context.Context, user *models.User) error {
+	id, err := m.Create(ctx, user)
```
heww, line 73: This operation needs to be wrapped with transactions otherwise the creation failure will make the top transaction commit to fail.

reasonerjt: Per discussion, I'll Read the user first and if there's  notfound error (not from DB level) insert the record.

heww (approved):

wy65701436 (approved):
lgtm

Issue #8723, opened 2019-08-16
Robot accounts should be allowed to access the full Harbor api


clausa:
**Expected behavior and actual behavior:**
I would like to use a robot account to programatically obtain vulnerability data for any repo:tag as well as other functions exposed by the api.

**Expected:**
Robot accounts should be allowed to access /api/ like regular user accounts.

**Actual:**
Robot accounts are only allowed to push/pull images and charts. 
Trying to access /api always returns 401

**Steps to reproduce the problem:**
- (Enable OIDC)
- Create project.
- Add robot account with full push and pull permissions.
- Use robot to access: '/api/repositories/{repo}/tags/{tag}/vulnerability/details' 
`$ curl -X GET "https://harbor/api/repositories/myrepo/tags/latest/vulnerability/details" -H "accept: application/json" -H "authorization: Bearer eyJhbG.....robot-account.token"`
- Get a 401 `{"code":401,"message":"Unauthorized"}`

**Versions**:

harbor version: 1.8.1

**Additional context:**
For now, creating robot accounts via the API should be sufficient - UI might come later.

Something like:
```
{
  "name": "scanner",
  "description": "robot account to fetch vulnerability reports from any project",
  "access": [
    {
      "resource": "/api/repositories/*/tags/*/vulnerability/details",
      "action": "get"
    }
  ]
}
```

fentas:
remove stale? :cat2:

wy65701436:
@xaleeks idea?

kawaiipantsu:
I think it would be a good idea, many are using OIDC for auth and that can cause problems with api also would be nice to use a more "neutral" user in your scripts etc than a actual employee/username.

troy256:
We would really like this feature too.  Specifically to pull vulnerability data for specific images/tags.

rubin55:
We ran into this issue when trying to use a robot account to DELETE tags (we want to use a cleanup script we previously used with GitLab registry that is failing with Harbor due to permissions):

```
%> curl -XGET https://myharbor/api/v2.0/projects/foo/repositories/example/artifacts
<lots of JSON, this works fine>

%> curl --basic -u 'somerobot:somepass' -XDELETE https://myharbor/api/v2.0/projects/foo/repositories/example/artifacts/null/tags/200616-057eb77a
{"errors":[{"code":"UNAUTHORIZED","message":"unauthorized"}]}
```

[...]

PR #14901, opened 2021-05-17
Allow using robot accounts for Image replication

### LINKED ISSUE #13384: Robot users should be able to sync private projects (Harbor -> Harbor)
I am trying to replicate a private repository from Harbor to Harbor (both using version 2.1.1) by using pull-based replication.

As you can see, my source-project contains images on multiple levels:

grafik

On  the target side, I configure this registry by using "Provider: Harbor" and using the credentials of the robot user:

grafik

Then, still on the target side, I configure the pull-based replication:

grafik

But no matter what I put into the "Source resource filter", Harbor won't even replicate a single image. There is no error when running the replication, just a false success message that appears almost immediately:

grafik

When using a regular user instead of a robot user, everything works as expected. Is this a known limitation of robot users?

**Versions:**

- harbor version: 2.1.1

### LINKED ISSUE #13795: Unable to replicate image to another Harbor instance with robot account
I'd like to be able to replicate image from one Harbor instance used by my team to another managed by another team on a different continent. I attempted to do this via a robot account, reasoning that all I need is docker push/pull access and then discovered the user requires the ability to create projects. As robots don't get REST access, this obviously fails, but if the system is configured to only allow admins to create projects, then this requires you to use an administrator account to just replicate an image.

My project exists, so I don't really know why Harbor decides to perform a project create first. Looking at:

https://github.com/goharbor/harbor/blob/df1f52dba4e5e6613a14645e9ab23d5fe6f6c595/src/replication/adapter/harbor/base/adapter.go#L171-L181

It looks like the project create is attempted regardless of it existing. I wouldn't object to a 403 being a warning in the logs if the project create failed because the push will inevitably fail down the line if the project is missing.

- Configure one instance of harbor with permissions that limit project creation to admins.
- Create a project on said Harbor instance and create a robot within that with write access.
- Configure another harbor instance to replicate using the robot account.
- Trigger the replication and observe the failure.

**Versions:**

- harbor version: 2.0.2 (issue appears to be present on master as well)
- docker engine version: 19.03.13
- docker-compose version: 1.26.2

**Additional context:**

- **Harbor config files:** N/A (not much of use in them for this)
- **Log files:** Not much in here, apart from:

```
Dec 17 14:34:25 172.18.0.1 core[15531]: 2020-12-17T14:34:25Z [ERROR] [/replication/operation/controller.go:108]: the execution 31550 failed: failed to do the prepare work for pushing/uploading resources: http error: code 403, mess
age {"errors":[{"code":"FORBIDDEN","message":"Only system admin can create project"}]}
```

Looks similar to #13268, so I wouldn't be surprised if the other adapters have a similar workflow that expects create permissions.

### LINKED ISSUE #14640: Pull-based replication from Harbor doesn't work with project-level robot accounts
**Expected behavior**
In the following setup consisting of HarborA (master) and HarborB (replica), I'm trying to setup pull-based replication inside HarborB.

As only a single project from HarborA should be replicated, I created a project-level robot account inside HarborA and used this one to configure the pull replication in HarborB, which is expected to discover the artifacts in that project in HarborA and copy them.

**Actual**
The replication jobs all report status SUCCESS, but don't copy any artifacts (no images, no charts). The jobservice logs are empty and just report SUCCESS

I'm assuming the issue comes from the following callchain, which tries to find the origin project by calling the ListProjects API in Harbor, which is inaccessible for project-level robot accounts:

1. https://github.com/goharbor/harbor/blob/v2.2.0/src/replication/adapter/harbor/v2/adapter.go#L54
1. https://github.com/goharbor/harbor/blob/v2.2.0/src/replication/adapter/harbor/base/adapter.go#L200
1. https://github.com/goharbor/harbor/blob/v2.2.0/src/replication/adapter/harbor/base/client.go#L117

**Versions:**
- harbor version: 2.2.0

Vad1mo:
This patch makes it possible to use robot accounts for image replication. 

The patch only changes `getProject` to call the API directly instead of filtering the list of `getProjects`. 
The reason is that `getProject` doesn't need system level permission and hance allows robot account to do replication.  

resolves #14640, resolves #13384, resolves #13795
related #8723

To do replication with robot accounts the robot accounts needs to have some more permissions 

Permissions needed for robot accounts.
```js
 [
        {
            'action': 'read',
            'resource': '?'
        },
        {
            'action': 'list',
            'resource': 'repository'
        },
        {
            'action': 'pull',
            'resource': 'repository'
        },
        {
            'action': 'list',
            'resource': 'artifact'
        },
        {
            'action': 'read',
            'resource': 'artifact'
        },
        {
            'action': 'list',
            'resource': 'helm-chart'
        },
        {
            'action': 'read',
            'resource': 'helm-chart'
        },
        {
            'action': 'list',
            'resource': 'helm-chart-version'
        },
        {
            'action': 'read',
            'resource': 'helm-chart-version'
        },
    ]
```

### DIFF
--- a/src/portal/src/app/account/sign-in/sign-in.component.spec.ts
+++ b/src/portal/src/app/account/sign-in/sign-in.component.spec.ts
@@ -1,18 +1,15 @@
 import { waitForAsync, ComponentFixture, TestBed } from '@angular/core/testing';
 import { SignInComponent } from './sign-in.component';
-import { TranslateModule, TranslateService } from '@ngx-translate/core';
-import { RouterTestingModule } from '@angular/router/testing';
 import { AppConfigService } from '../../services/app-config.service';
 import { SessionService } from '../../shared/services/session.service';
 import { CookieService } from 'ngx-cookie';
 import { SkinableConfig } from "../../services/skinable-config.service";
 import { CUSTOM_ELEMENTS_SCHEMA, NO_ERRORS_SCHEMA } from '@angular/core';
-import { ClarityModule } from "@clr/angular";
-import { FormsModule, ReactiveFormsModule } from '@angular/forms';
 import { of } from "rxjs";
 import { throwError as observableThrowError } from 'rxjs/internal/observable/throwError';
 import { HttpErrorResponse } from '@angular/common/http';
-import { HttpClientTestingModule } from '@angular/common/http/testing';
+import { SharedTestingModule } from "../../shared/shared.module";
+import { UserPermissionService } from "../../shared/services";
 
 describe('SignInComponent', () => {
     let component: SignInComponent;
@@ -20,21 +17,23 @@ describe('SignInComponent', () => {
     const mockedSessionService = {
         signIn() {
             return of(true);
+        },
+        getCurrentUser() {
+            return {};
+        }
+    };
+    const mockedUserPermissionService = {
+        clearPermissionCache() {
         }
     };
     beforeEach(waitForAsync(() => {
         TestBed.configureTestingModule({
             imports: [
-                TranslateModule.forRoot(),
-                RouterTestingModule,
-                ClarityModule,
-                FormsModule,
-                ReactiveFormsModule,
-                HttpClientTestingModule
+                SharedTestingModule
             ],
             declarations: [SignInComponent],
             providers: [
-                TranslateService,
+                { provide: UserPermissionService, useValue: mockedUserPermissionService},
                 { provide: SessionService, useValue: mockedSessionService},
                 {
                     provide: AppConfigService, useValue: {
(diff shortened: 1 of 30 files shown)


wsams:
@Vad1mo , I know this is a bit old, but since this issue is still around I wanted to confirm that these fixes require everyone to be able to create projects. If it's set to admins we still get a 403 when replicating local projects with a robot account. (The last section on this page https://goharbor.io/docs/2.2.0/administration/general-settings/) Is that the understanding here?

olifur:
Hi there,

as @wsams stated in the comment above, the problem is still there in v2.4+. We run v2.5.0-98e1b82f and if we set the 
"Project Creation" option to "Everyone" it works that our robot account (with full permission) can replicate our registry whereas if we set it to "Admin only" it does not work, we get 403 Forbidden, only "system admin can create project". Setting it to "Everyone" is not an option for us in a enterprise environment.
Best, Oli

Vad1mo:
yes, this is correct. We need to address that in another issue/PR
.

PR #14905, opened 2021-05-17
Allows using robot accounts for Image replication

### LINKED ISSUE #13384: Robot users should be able to sync private projects (Harbor -> Harbor)

### LINKED ISSUE #13795: Unable to replicate image to another Harbor instance with robot account

### LINKED ISSUE #14640: Pull-based replication from Harbor doesn't work with project-level robot accounts

Vad1mo:
This patch (more of a bug fix) will allow using robot accounts for image replication. 

Currently, it is not possible to create a robot account that can replicate images from one Harbor registry to another. (You need a System Admin account, big no go in corp environments)

The patch only changes `getProject` to call the API directly, instead of filtering the list of `getProjects` because `getProjects` needs admin level permission and hence is not possible to be used with robot accounts. 

## Side Effects

The change has no or littel side effects: 
image

its only called by `listProjects` and is always providing and expect a single result, 

This patch only works for Harbor => 2.2.0 and resolves a bunch of open issues.

resolves #14640, resolves #13384, resolves #13795
related #8723

## How to creaet a Robot Accounts for Replication

In order to a replication with robot accounts, a robot account with those permissions needs to be created.

Permissions needed for robot accounts.

```js
 [
        {
            'action': 'read',
            'resource': '?'
        },
        {
            'action': 'list',
            'resource': 'repository'
        },
        {
            'action': 'pull',
            'resource': 'repository'
        },
        {
            'action': 'list',
            'resource': 'artifact'
        },
        {
            'action': 'read',
            'resource': 'artifact'
        },
        {
            'action': 'list',
            'resource': 'helm-chart'
        },
        {
            'action': 'read',
            'resource': 'helm-chart'
        },
        {
            'action': 'list',
            'resource': 'helm-chart-version'
        },
        {
            'action': 'read',
            'resource': 'helm-chart-version'
        },
    ]
```

Signed-off-by: Vadim Bauer <vb@container-registry.com>

### DIFF
--- a/src/pkg/reg/adapter/harbor/base/chart_registry_test.go
+++ b/src/pkg/reg/adapter/harbor/base/chart_registry_test.go
@@ -29,6 +29,17 @@ func TestFetchCharts(t *testing.T) {
 	server := test.NewServer([]*test.RequestHandlerMapping{
 		{
 			Method:  http.MethodGet,
+			Pattern: "/api/projects/library",
+			Handler: func(w http.ResponseWriter, r *http.Request) {
+				data := `{
+					"name": "library",
+					"metadata": {"public":true}
+				}`
+				w.Write([]byte(data))
+			},
+		},
+		{
+			Method:  http.MethodGet,
 			Pattern: "/api/projects",
 			Handler: func(w http.ResponseWriter, r *http.Request) {
 				data := `[{
--- a/src/pkg/reg/adapter/harbor/base/client.go
+++ b/src/pkg/reg/adapter/harbor/base/client.go
@@ -114,16 +114,12 @@ func (c *Client) ListProjects(name string) ([]*Project, error) {
 
 // GetProject gets the specific project
 func (c *Client) GetProject(name string) (*Project, error) {
-	projects, err := c.ListProjects(name)
-	if err != nil {
+	project := &Project{}
+	url := fmt.Sprintf("%s/projects/%s", c.BasePath(), name)
+	if err := c.C.Get(url, project); err != nil {
 		return nil, err
 	}
-	for _, project := range projects {
-		if project.Name == name {
-			return project, nil
-		}
-	}
-	return nil, nil
+	return project, nil
 }
 
 // BasePath returns the API base path that contains version part


a-mccarthy:
@Vad1mo,  I think adding docs around this scenario will be helpful for users. can you write up some more context here about this use case? Or create an issue so we can track adding this into the docs? thanks!

OrlinVasilev:
@reasonerjt can you check that please?

defreng:
Hi @wy65701436 

I would be super interested in the issue with merging this bugfix? To me this seems to be a fairly obvious improvement of code quality and stability of Harbor.

The current code which first retrieves the unfiltered list of all projects and then gets the wanted one by looping over them in memory to me seems to be a clear antipattern. It will cause sluggishness - especially when a Harbor instance is managing more and more projects.

wy65701436:
To make robot account can do replicaiton will in the next release, I'll have a review on the currently design and find out gaps.
There are some details that needs to be discussed.

Vad1mo:
@wy65701436 this PR makes Replication possible. We currently run replication like this. Let me know if I can be of any help

defreng:
@wy65701436 even though the title suggests otherwise, from my perspective this is really just a code quality improvement / bugfix.

The fact that it also fixes replication using robot accounts (**which does work**! We're using a version of Harbor with this patch in production), is merely a nice side-effect.

phin1x:
@Vad1mo do you allow every authenticated user to create projects?

Vad1mo:
I don't know, probably yes as it is the default.

phin1x:
harbor wants to create the project in the prepare method of the replication adpater, regadless if the project exists (it just ignores the error in this case). but if only admins are allowed to create projects, this api call will never succeed (old problem, mixing system and project level permissions in robot acocunts). we also had to patch out the code for creating the project in the prepare method and the replication succeeded (#14982).

wy65701436:
@Vad1mo this PR is for robot account to call list project & list repositories API. Thanks for the contributing, we can keep the client here.

Vad1mo:
ok, thank you. 

Would this PR not still make sense? Because instead of iterating over `ListProjects` and filtering the one project, why not just call directly getProjects API?

Vad1mo:
thank you @wy65701436,

Would this PR not still make sense?  Instead of iterating over `ListProjects` and filtering the one project the is needed,  it would make more sense to just call directly getProject API?

PR #14906, opened 2021-05-17
Provide detailed output about missing RBAC permissions in debug mode.

Vad1mo:
This change displays missing RBAC permissions in the log when the debug mode is enabled. 
The functionality well help the normal user to craft permissions for robot accounts. This is especially useful if you want to perform API operations with robot accounts.

Signed-off-by: Vadim Bauer <vb@container-registry.com>


heww (approved):

wy65701436 (approved):
lgtm

PR #14912, opened 2021-05-18
Truncate the string when deletinng the resources when the length exceeds the limit

ywk253100:
Truncate the string when deletinng the resources when the length exceeds the limit


wy65701436 (approved):
lgtm

heww (approved):

Issue #14913, opened 2021-05-18
Failed to preheat image when execute the preheat task second time.


yunkunrao:
**Expected behavior and actual behavior:**
Harbor: v2.2.1
Dragonfly: 1.0.6

In my environment, I have configured Harbor to create a dragonfly preheat Provider Instances. Then created a Harbor project and p2p preheat policy. When I tried to trigger the preheat policy for the first time. It's OK, but for the second time, the task failed always.

Expected behavior: preheat policy can be trigger for any times successfully.
Actual behavior: preheat policy can only be trigger for the first time successfully.

**Additional context:**

- **Log files:**

job service log

[root@harbor1 dragonfly]# kubectl logs -n demo h1-harbor-jobservice-db6665495-78rjp | grep -i "error"
2021-05-18T08:59:41Z [ERROR] [/pkg/p2p/preheat/provider/client/http_client.go:75]: Get http://192.168.0.2:8002/preheats/108b66ecf0cd867b4c367a2e32925405ac9beac4566807973526d9eb4166fef2 with cred=&{NONE map[]}, params=map[], options=map[]: GET 'http://192.168.0.2:8002/preheats/108b66ecf0cd867b4c367a2e32925405ac9beac4566807973526d9eb4166fef2' error: 404 Not Found {"code":404,"message":"108b66ecf0cd867b4c367a2e32925405ac9beac4566807973526d9eb4166fef2 doesn't exists"}
2021-05-18T08:59:41Z [ERROR] [/jobservice/runner/redis.go:111]: Job 'P2P_PREHEAT:6144b2f0358735c9a3cc745c' exit with error: run error: preheat job running error: GET 'http://192.168.0.2:8002/preheats/108b66ecf0cd867b4c367a2e32925405ac9beac4566807973526d9eb4166fef2' error: 404 Not Found {"code":404,"message":"108b66ecf0cd867b4c367a2e32925405ac9beac4566807973526d9eb4166fef2 doesn't exists"}
[...]

PR #14919, opened 2021-05-19
Fix dragonfly preheat error.

yunkunrao:
Signed-off-by: yunkunrao <yunkunrao@gmail.com>

This PR fix issue #14913

### DIFF
--- a/src/pkg/p2p/preheat/provider/dragonfly.go
+++ b/src/pkg/p2p/preheat/provider/dragonfly.go
@@ -4,8 +4,10 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/http"
 	"strings"
 
+	common_http "github.com/goharbor/harbor/src/common/http"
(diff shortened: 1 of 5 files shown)


src/pkg/p2p/preheat/provider/dragonfly.go
```diff
@@ -87,9 +87,17 @@ func (dd *DragonflyDriver) Preheat(preheatingImage *PreheatImage) (*PreheatingSt
 		return nil, err
 	}
 
+	// Currently if resource was preheated before at least once, Dragonfly will return StatusAlreadyReported.
```
steven-zou, line 90: @yunkunrao 

Why don't directly check if the status == StatusAlreadyReported? I'm thinking the `result.ID == ""` may also happen in other cases.

steven-zou (commented):
@yunkunrao 

Please check my comments.

bitsf (approved):

steven-zou (approved):
LGTM

PR #14939, opened 2021-05-21
remove common project code

wy65701436:


heww (approved):

src/pkg/project/models/project.go
```diff
@@ -15,14 +15,192 @@
...
+func (p *Project) AutoScan() bool {
+	auto, exist := p.GetMetadata(ProMetaAutoScan)
+	if !exist {
+		return false
+	}
+	return isTrue(auto)
+}
+
+// FilterByPublic returns orm.QuerySeter with public filter
+func (p *Project) FilterByPublic(ctx context.Context, qs orm.QuerySeter, key string, value interface{}) orm.QuerySeter {
```
reasonerjt, line 137: Is this being used? It doesn't look correct under the structure we agreed on.

wy65701436: cc @heww

reasonerjt: I didn't know this was based on convention.  
I'm fine with this approach.

src/pkg/project/dao/dao.go
```diff
@@ -70,7 +70,7 @@ func (d *dao) Create(ctx context.Context, project *models.Project) (int64, error
 			return orm.WrapConflictError(err, "The project named %s already exists", project.Name)
 		}
 
-		member := &Member{
+		member := &models.Member{
```
ywk253100, line 73: There is another `Member` definition under package `pkg/member`, seems they are duplicated?

wy65701436: will raise another PR to handle this.

PR #14941, opened 2021-05-21
Set null list when or list is nil in query

ywk253100:
Set null list when or list is nil in query


wy65701436 (approved):
lgtm

heww (approved):

Issue #14407, opened 2021-03-10
The ChartMuseum is scheduled to be deprecated in a future v2.4.0 release.


eugenesotnikov:
I read features of relese 2.2.0. And i was very surprised by this
"The ChartMuseum is scheduled to be deprecated in a future v2.4.0 release."
What does it mean? I didn't find anything about it on site.
We have deploy based on helm+harbor. And i want to be ready for this.

Vad1mo:
Helm 3 uses OCI for package distribution. It ships with two new subcommands to help with this: registry and chart. The registry sub-command handles authentication with OCI registries while chart takes care of save, push and pull charts to the registry.
The fetch and pull sub-commands used for interacting with the chart repositories, chartmuseum, (in earlier versions) are still retained for backward compatibility.

**If you use helm 3 then OCI is the future, there will be no need for ChartMuseum, so it's safe to deprecate it.**

MaxRink:
>If you use helm 3 then OCI is the future, there is no need for ChartMuseum, so it's safe to deprecate it.

This is not true, yet. The default helm behaviour still doesnt support OCI and OCI suppoort is still experimental

sagikazarmark:
Quoting the [official website](https://helm.sh/docs/topics/registries/):

> Currently OCI support is considered experimental.

This seems to be quite a major change...major enough not to hide it in a changelog.

I also think that most of the ecosystem still uses the "old" helm repository format, so talking about ChartMuseum deprecation seems to be a bit premature to me.

hojerst:
Considering that Helm Registry support was a major reason why we acutally chose Harbor, I'd also like to see the Chartmuseum supported a bit longer. As mentioned before, the OCI support in helm is still experimental and even needs to be turned on with a env feature flag. I'd really like to see chartmuseum support in Harbor at least until Helm officially supports OCI for production use. Just my two cents.

Vad1mo:
hey all, 

We should close this issue and continue our discussion on https://github.com/goharbor/harbor/discussions/15057

steven-zou:
Agree with @Vad1mo. If you have any comments on the `chartmuseum` deprecation work, move to the discussions on #15057.

cc @hojerst / @sagikazarmark  / @MaxRink / @eugenesotnikov / @alxbse

PR #14944, opened 2021-05-21
Add GUI support for chart label permissions

SimonAlling:
If a `create` or `delete` permission for the `helm-chart-version-label`
resource is [added to a robot account using the REST API][1], it shows
up in the GUI as just "Create" or "Delete", respectively. So the
permissions for the robot account may look like this:

  * Push Artifact
  * Pull Artifact
  * Delete Artifact
  * Read Helm Chart
  * Create Helm Chart Version
  * Delete Helm Chart Version
  * Create Tag
  * Delete Tag
  * Create Artifact label
  * Create Scan
  * Create                    👈
  * Delete                    👈

This PR adds GUI support for said permissions, so that they are shown as
"Create Helm Chart Version label" and "Delete Helm Chart Version label",
respectively. The names are derived from "Create Helm Chart Version"
(title-case "Helm Chart Version") and "Create Artifact label"
(lower-case "label", except in German, which has "Artefakt Label").

[1]: https://github.com/goharbor/harbor/issues/8723#issuecomment-845719044

Signed-off-by: Simon Alling <alling.simon@gmail.com>

### DIFF
--- a/src/portal/src/app/base/left-side-nav/system-robot-accounts/system-robot-util.ts
+++ b/src/portal/src/app/base/left-side-nav/system-robot-accounts/system-robot-util.ts
@@ -107,6 +107,7 @@ export const ACTION_RESOURCE_I18N_MAP = {
     'helm-chart-version': 'SYSTEM_ROBOT.HELM_VERSION',
     'tag': 'REPLICATION.TAG',
     'artifact-label': 'SYSTEM_ROBOT.ARTIFACT_LABEL',
+    'helm-chart-version-label': 'SYSTEM_ROBOT.HELM_CHART_VERSION_LABEL',
     'scan': 'SYSTEM_ROBOT.SCAN',
     'scanner-pull': 'SYSTEM_ROBOT.SCANNER_PULL'
 };
--- a/src/portal/src/i18n/lang/de-de-lang.json
+++ b/src/portal/src/i18n/lang/de-de-lang.json
@@ -1675,6 +1675,7 @@
         "ENABLE_NEW_SECRET": "Aktiviere diese Option um das neue Secret selbst zu definieren",
         "DELETE": "Löschen",
         "ARTIFACT_LABEL": "Artefakt Label",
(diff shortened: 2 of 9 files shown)


stonezdj:
We are going to deprecate chartmusem in future release, so we don't want to add this feature.

SimonAlling:
> We are going to deprecate chartmusem in future release, so we don't want to add this feature.

What is chartmuseum? How can I tell whether I'm using it? (The best clue I could find was "chartmuseum is not enabled by default" in `make/install.sh`.)

Also, is this really a feature? It feels more like a bug fix to me. Please note that when I visit https://harbor.example.com/harbor/robot-accounts, a robot account that has permissions to create and delete charts confusingly appears to have the permissions "Create" and "Delete" (among others), as shown above.

lindhe:
> @stonezdj We are going to deprecate chartmusem in future release, so we don't want to add this feature.

Are there discussions regarding this that I can read somewhere? What will replace chartmuseum, and how will that effect the available permissions for labeling Helm charts?

SimonAlling:
> > @stonezdj We are going to deprecate chartmusem in future release, so we don't want to add this feature.
> 
> Are there discussions regarding this that I can read somewhere? What will replace chartmuseum, and how will that effect the available permissions for labeling Helm charts?

For the record:

* #14407
* #15057

AllForNothing:
@SimonAlling Closing it, duplicated PR with #15815

SimonAlling:
> @SimonAlling Closing it, duplicated PR with #15815

What? This PR was created in May, the other one in October. Also:

> We are going to deprecate chartmusem in future release, so we don't want to add this feature.

I don't understand. Oh well.

AllForNothing:
@SimonAlling 
We should have merged this earlier PR, but #15815 is more complete. Thanks again for your contribution.

As for the chartmusem, after discussion, we decided to synchronize the chart label permissions in the front and back ends first, and then delete them together after they are deprecated.

PR #14949, opened 2021-05-22
Fall back to local registry when upstream registry is not working

### LINKED ISSUE #14822: Proxy cache does not provide images when target registry is not working
The Proxy cache is one of the most important features of harbor to ensure that the clients will get the most recent available images without having to connect to the upstream registry. But it does not work as expected, if the upstream registry has a problem, for example login fails at hub.docker.com, as seen in #14802 

*Expected:*
If connecting to a _previously working_ target registry fails,  Harbor should
* serve the most recent cached image if it has one (even if it cannot ensure it is the most recent one on the upstream server)
* fail with an error if it has not cached the requested image yet

*Actual:*
* Harbor fails with an error even if it has cached that image in the past

* Install a fresh version of Harbor
* Create a proxy cache project which points to Dockerhub hub.docker.com with working credentials
* Pull an image (for example `busybox`) through the proxy
* If necessary, repeat the step above until the image appears in the Repositories list
* Invalidate the access_key by editing the values in the `registry` table in the database
* Pull the same image again

Result: an error occurs even if the image is in the cache already

**Versions:**
- harbor version: v2.2.1
- reports from #14802 indicate the same problem with version v2.1.3

**Additional context:**
If many users and pipelines are using Harbor as a proxy cache, it is better to get a (possibly outdated) image instead of no image at all.

stonezdj:
Fixes #14822
When the upstream registry not working, but status might stay healthy because the health check interval is 5 minutes, if a pull request comes before registry status turns to unhealthy, the proxy cache middleware might proxy the request to the upstream registry and get a 401 error and this 401 error might translate to a http 500 error to the client eventually.

To solve this issue, it fall back all error to local registry when proxying manifest except the NotFoundError from the local regist

### DIFF
--- a/src/server/middleware/repoproxy/proxy.go
+++ b/src/server/middleware/repoproxy/proxy.go
@@ -95,7 +95,12 @@ func preCheck(ctx context.Context) (art lib.ArtifactInfo, p *models.Project, ctl
 func ManifestMiddleware() func(http.Handler) http.Handler {
 	return middleware.New(func(w http.ResponseWriter, r *http.Request, next http.Handler) {
 		if err := handleManifest(w, r, next); err != nil {
-			httpLib.SendError(w, err)
+			if errors.IsNotFoundErr(err) {
+				httpLib.SendError(w, err)
+				return
+			}
+			log.Errorf("failed to proxy manifest, fallback to local, request uri: %v, error: %v", r.RequestURI, err)
+			next.ServeHTTP(w, r)
 		}
 	})
 }


src/controller/proxy/remote.go
```diff
@@ -73,6 +74,16 @@ func (r *remoteHelper) init() error {
 	}
 	adp, err := factory.Create(reg)
 	if err != nil {
+		if reg.Status == model.Healthy {
+			defer func(reg model.Registry) {
```
ywk253100, line 78: Why use `defer` here?

src/controller/proxy/remote.go
```diff
@@ -73,6 +74,16 @@ func (r *remoteHelper) init() error {
 	}
 	adp, err := factory.Create(reg)
 	if err != nil {
+		if reg.Status == model.Healthy {
```
reasonerjt, line 77: The factory has different implementations so it returns error for various reasons, marking the registry unhealthy for whatever error doesn't seem correct to me.
How it will impact if we just leave it as is?
@ywk253100  thoughts?

src/controller/proxy/remote.go
stonezdj: sorry, should be go func

src/controller/proxy/remote.go
stonezdj: remove it

ywk253100 (approved):

src/server/middleware/repoproxy/proxy.go
```diff
@@ -95,7 +95,12 @@ func preCheck(ctx context.Context) (art lib.ArtifactInfo, p *models.Project, ctl
 func ManifestMiddleware() func(http.Handler) http.Handler {
 	return middleware.New(func(w http.ResponseWriter, r *http.Request, next http.Handler) {
 		if err := handleManifest(w, r, next); err != nil {
-			httpLib.SendError(w, err)
+			if errors.IsNotFoundErr(err) {
+				httpLib.SendError(w, err)
+				return
+			}
+			log.Errorf("failed to proxy to manifest with the error %v, fallback to local", err)
```
reasonerjt, line 102: I think the grammar of this message is incorrect.  But I'll approve

reasonerjt (approved):

PR #14951, opened 2021-05-23
fix: delete robots by project id using raw sql

### LINKED ISSUE #14678: Cannot delete project - PostgreSQL only supports 65535 parameters
When attempting to delete a seemingly empty project in harbor's UI an 'Internal Server Error' message is displayed and the harbor core pod logs show;

> 2021-04-16T15:11:31Z [ERROR] [/lib/http/error.go:54]: {"errors":[{"code":"UNKNOWN","message":"unknown: pq: got 98119 parameters but PostgreSQL only supports 65535 parameters"}]}

We are using harbor 2.2.0 deployed on a Kubernetes cluster, with an external Postgres 9.6.3.

The s3 backend and database has many projects which have been deleted from the UI but remain in the DB/Storage and may be causing some issues, but without knowing what the query is that fails I can't debug any further.

heww:
QuerySeter.Delete will call two queies in the orm of beego. First it
will query all primary keys of the model by the QuerySeter, and then the
orm will delete the models using the IN statement of these primary keys.
This will be failed in the postgres when the records more than 65535.
This commit changes to use the raw query to delete the robots of the
project.

Closes #14678

### DIFF
--- a/src/pkg/robot/dao/dao.go
+++ b/src/pkg/robot/dao/dao.go
@@ -2,11 +2,12 @@ package dao
 
 import (
 	"context"
+	"time"
+
 	"github.com/goharbor/harbor/src/lib/errors"
 	"github.com/goharbor/harbor/src/lib/orm"
 	"github.com/goharbor/harbor/src/lib/q"
 	"github.com/goharbor/harbor/src/pkg/robot/model"
-	"time"
 )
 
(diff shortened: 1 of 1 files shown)


wy65701436 (approved):
Lgtm

ninjadq (approved):
lgtm

PR #14985, opened 2021-05-26
Fix bug of migration sql for replication policy

ywk253100:
Fix bug of migration sql for replication policy


wy65701436 (approved):
lgtm

PR #14995, opened 2021-05-27
add build arch parameter in Makefile

wy65701436:
Add parameter BUILDARCH for make file. DB base builds pg96 for x86_64 only

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,6 +74,7 @@ PORTAL_PATH=$(BUILDPATH)/src/portal
 CHECKENVCMD=checkenv.sh
 
 # parameters
+BUILDARCH=x86_64 # default is x86_64
 REGISTRYSERVER=
 REGISTRYPROJECTNAME=goharbor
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -74,6 +74,7 @@ PORTAL_PATH=$(BUILDPATH)/src/portal
 CHECKENVCMD=checkenv.sh
 
 # parameters
+BUILDARCH=x86_64 # default is x86_64
```
reasonerjt, line 77: Adding BUILDARCH may cause a result that more and more arch specific path adding to the upstream project, which is something we want to avoid.

Can we introduce a flag specifically like `BUILD_PG96` ?

wy65701436: eventually, we will have a Arch parameter and ship harbor images index, can we just start it from this PR? 

Fo

reasonerjt (approved):

heww (approved):

PR #15037, opened 2021-06-01
update the log level of robot v1 context

wy65701436:
The robotv1 context uses the robot$ as a hardcoded prefix to identify robot account, it will raise error
for a valid robotv2 account with this prefix.

Update the log level to avoid the redundant logs for the default installation.

### DIFF
--- a/src/server/middleware/security/robot.go
+++ b/src/server/middleware/security/robot.go
@@ -46,7 +46,7 @@ func (r *robot) Generate(req *http.Request) security.Context {
 	}
 	rtk, err := pkg_token.Parse(defaultOpt, robotTk, rClaims)
 	if err != nil {
-		log.Errorf("failed to decrypt robot token: %v", err)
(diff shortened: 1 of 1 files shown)


src/server/middleware/security/robot.go
```diff
@@ -46,7 +46,7 @@ func (r *robot) Generate(req *http.Request) security.Context {
 	}
 	rtk, err := pkg_token.Parse(defaultOpt, robotTk, rClaims)
 	if err != nil {
-		log.Errorf("failed to decrypt robot token: %v", err)
+		log.Debugf("failed to decrypt robot token: %v", err)
```
reasonerjt, line 49: You should include the robot name in the msg for better clarity.

wy65701436: added

ywk253100 (approved):

PR #15047, opened 2021-06-02
fixes 13976

### LINKED ISSUE #13976: dry_run for gc does not count all blobs to be deleted
1. prepare an artifacts with two layers, 10MB and 50MB
2. limit the project with quota 50MB
3. push the artifacts, and it is blocked as desired, left one garbage 10MB layer
4. wait 2 hours or adjust the env GC_TIME_WINDOW_HOURS=0 in harbor-core
5. run GC job with `dry_run` option, found no blobs to be deleted
image
6. run GC job, found 1 blob to be deleted
image

`dry_run` GV job is expected to reflect the real operation

wy65701436:
Fixes #13976 for the quota exceed case, gc will print the untagged blobs for dry-run

### DIFF
--- a/src/jobservice/job/impl/gc/garbage_collection.go
+++ b/src/jobservice/job/impl/gc/garbage_collection.go
@@ -189,13 +189,14 @@ func (gc *GarbageCollector) mark(ctx job.Context) error {
 
 	// get gc candidates, and set the repositories.
 	// AS the reference count is calculated by joining table project_blob and blob, here needs to call removeUntaggedBlobs to remove these non-used blobs from table project_blob firstly.
-	if !gc.dryRun {
-		gc.removeUntaggedBlobs(ctx)
-	}
+	untaggedBlobs := gc.markOrSweepUntaggedBlobs(ctx)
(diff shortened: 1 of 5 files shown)


src/jobservice/job/impl/gc/garbage_collection.go
```diff
@@ -415,8 +416,11 @@ func (gc *GarbageCollector) deletedArt(ctx job.Context) (map[string][]model.Arti
 	return artMap, nil
 }
 
-// clean the untagged blobs in each project, these blobs are not referenced by any manifest and will be cleaned by GC
-func (gc *GarbageCollector) removeUntaggedBlobs(ctx job.Context) {
+// mark or sweep the untagged blobs in each project, these blobs are not referenced by any manifest and will be cleaned by GC
+// * dry-run, find and return the untagged blobs
+// * non dry-run, remove the untagged blobs
```
reasonerjt, line 421: This is not accurate cause just remove the reference between blob and project, it's not removal.

reasonerjt (approved):

heww (approved):

PR #15055, opened 2021-06-03
Add IPv6 support

ninjadq:
* 127.0.0.1 to localhost


bitsf (approved):

reasonerjt:
Considering the `$IP` can be complicated when it's ipv6, how'bout hard code some temp file name instead of using it as part of the filename?

reasonerjt (approved):

Issue #14736, opened 2021-04-23
Does harbor have a plan to support replicate action which trigger by docker pull using pull mode?


yunkunrao:
**Backgroud
Currently when one harbor cluster wants to sync with other harbor. In our case we create a replication rule with pull mode enabled. Admin had to sync image manually by clicking on replicate button. And it's a tedious task.

**Describe the solution you'd like**
When user executed cmd "docker pull XXX", and harbor didn't sync the image yet or the images is not latest, it will automatically pull from the other harbor. And it behaves much like proxy cache anyway.

wy65701436:
Can you clarify the scenario? It seems that proxy cache can handle it?

yunkunrao:
Yeah, currently proxy cache seems to be able to handle this scenario, but the prerequisite is docker client must change image name according. For example, original image name is demo-pro/redis:latest and user must add the prefix "proxy/" i.e. "proxy/demo-pro/redis:latest". In other words, we want to a solution which is transparent to docker client or k8s.

PR #15060, opened 2021-06-04
Support project auto sync feature by pull-based replication mode

yunkunrao:
Signed-off-by: yunkunrao <yunkunrao@gmail.com>

### Summary
This PR adds event-based trigger mode for pull-based replication mode. This PR is related with #14736 .

### Background
Briefly, current replication policy doesn't support event-based trigger mode for pull-based replication mode.
And alternative solution proxy cache seems to have a similar function, but it has it's own limitation that docker image namespace has to be changed with a new proxy cache project as the prefix.

### Solution
Harbor admin can create a project with the same project name as the upsteam Harbor's project name with specified upstream registry_id and enable auto sync option.

When client run the docker pull cmd, local harbor will synchronize the docker image automically in the project. In other words it's totally transparent to docker client without any prefix in docker image full name.

For example, if we want to pull the latest image from downstream Harbor
- For proxy cache, we can run `docker pull <downstream_harbor:port>/proxy/<your ns>/<your image>:latest`
- For pull-based replication mode, we can run `docker run <downsteam_harbor:port>/<your ns>/<your image>:latest`. Only  prerequisites are project name should be the same as upstream's project name and project enable auto-synced option.

### DIFF
--- a/api/v2.0/swagger.yaml
+++ b/api/v2.0/swagger.yaml
@@ -6213,6 +6213,9 @@ definitions:
       cve_allowlist:
         description: The CVE allowlist of this project.
         $ref: '#/definitions/CVEAllowlist'
+      auto_synced:
+        type: boolean
+        description: Whether the project can be auto-synced with pull-based replication mode.
   ProjectDeletable:
     type: object
     properties:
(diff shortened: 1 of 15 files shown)


yunkunrao:
Proposal was submitted, please refer to https://github.com/goharbor/community/pull/169

PR #15084, opened 2021-06-08
Flattening repository path components excepting the last one

### LINKED ISSUE #15072: replication flattening 2 replicate different images into same repo
Source registry
source

Create a replication rule with flatten2 option
config

Then the replication failed with the following msg
error

Then change the destination to flattern2_test2/mytest, then it can pass
config1

Then manually trigger the replication, found that harbor-core and openldap image are replicate to the same repo flattern2_test2/mytest
dest

Most users are expecting to keep the image name when the path is flattenned.
```
flattern2_test2/mytest/harbor-core:dev
flattern2_test2/mytest/openldap:dev
```

ywk253100:
Flattening repository path components excepting the last one in replication

Fixes #15072

### DIFF
--- a/src/controller/replication/flow/stage.go
+++ b/src/controller/replication/flow/stage.go
@@ -17,13 +17,13 @@ package flow
 import (
 	"fmt"
 	"github.com/goharbor/harbor/src/lib/errors"
+	"path"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


bitsf (approved):

steven-zou (approved):
LGTM

Issue #14299, opened 2021-02-24
Includues latest N tags of the artifact in the payload of SCANNING_COMPLETED webhook


heww:
From v2.0, we scan the artifact by the digest, the tag attribute in the payload of SCANNING_COMPLETED webhook is always empty (see #13464 #14293), the tag attribute was removed by PR #14057.

We can include the latest N tags of the artifact in the payload of the SCANNING_COMPLETED webhook looks like we will display N tags of the artifact in the portal. Users can use this tags attribute when they want to process the webhook by tag.

PR #15090, opened 2021-06-08
fix: supply the latest tag of artifact to scan request

### LINKED ISSUE #14416: Interrogation Services API doesn't supply `tag` value on POST /scan
When Harbor sends its POST request to our interrogation service on endpoint `/scan` it always contains empty string for `tag` field. As written in the API spec, it should contain the actual tag of the image. 

Please note, that the tag is displayed correctly in the Harbor UI, it's just not send to the interrogation service.

Just inspect the POST request coming from Harbor with image that is not based on `:latest` tag. By the way, `latest` is never contained in  the `tag` field as well, but I think this is by convention.

**Versions:**
Harbor 2.2

heww:
Closes #14416 #14299


steven-zou (approved):
LGTM

ywk253100 (approved):

PR #15095, opened 2021-06-09
Set "x-omitempty" as "false" for the pull count and artifact count of…

ywk253100:
… repository

Set "x-omitempty" as "false" for the pull count and artifact count of repository to avoid the missing of these properties in API response


heww (approved):

ninjadq (approved):
lgtm

PR #15133, opened 2021-06-13
Fix docker access deny issue for build base workflow

danfengliu:
It would be a docker access deny failure if no logout, so add logout back to push script.


AllForNothing (approved):

ninjadq (approved):
lgtm

PR #15196, opened 2021-06-24
Add SearchDN error, mark field as required in UI

sixeIa:
…ection

Fix #15196


stonezdj (commented):
The original logic is to verify the SearchDn and the error message is incorrect, could you please add a new error for it.

```
 // ErrEmptySearchDN ...
var ErrEmptySearchDN = errors.New("empty search dn")
```

sixeIa:
SearchDN isn't a mandatory field (according to the UI) so why should we check if it is empty or not ?

stonezdj:
> SearchDN isn't a mandatory field (according to the UI) so why should we check if it is empty or not ?
 searchDN is mandatory in backend

sixeIa:
@stonezdj I've done the required changes and edited UI to mark is as required for user. Tell me if there are additional code to review/change.

stonezdj (approved):

reasonerjt:
Please correct the commit message.

sixeIa:
@reasonerjt modified commit history.

reasonerjt (approved):

PR #15201, opened 2021-06-25
Add TiB support for display, fix unit to correct one

### LINKED ISSUE #15200: Storage used displayed in Bytes, and not in GiB/TiB
Hi !

I'm running the latest version of harbor, and I have a repository with more than a TiB of data.

This size displayed on the top right corner is displayed in Bytes, even if the value is `1372900780740Byte`. I feel like it'd be better to have it in GiB/TiB. Because of the length, it's not fully displayed, and causes horizontal scrolling.

bytes

Let me know if you need more information about this

sixeIa:
Should fix #15200.

Based on the same pattern that GB/KB.

In the same time, I fixed the data unit used as it's not (kilo/giga/tera)byte but kibibit, gibibit (according to unit tests expecting value).


AllForNothing (approved):
LGTM

wy65701436 (approved):
lgtm

Issue #14856, opened 2021-05-11
Harbor can't recover when the connection to the database drops.


Vad1mo:
When the connection to the database is lost (unreliable connectivity), Harbor can't recover itself and the health check doesn't notice that there is a problem. Restarting the pod solves the issues immediately until the next connectivity problem.

This problem occurs multiple times a day. 
I am quite confident that this isn't a DB issues as other Harbor instances in the same network keep on working fine and the DB is basically idling.

```sh

2021-05-11 23:21:23 | 2021-05-11T21:21:23Z [WARNING] [/common/utils/oidc/helper.go:72]: Failed to refresh configuration, error: failed to load OIDC setting: write tcp 100.1.1.1:49918->51.59.7.7:5432: write: broken pipe
-- | --
  |   |   | 2021-05-11 23:21:20 | 2021-05-11T21:21:20Z [WARNING] [/common/utils/oidc/helper.go:72]: Failed to refresh configuration, error: failed to load OIDC setting: write tcp 100.1.1.1::40860->>51.59.7.7:5432: write: connection reset by peer
  |   |   | 2021-05-11 23:21:17 | 2021-05-11T21:21:17Z [WARNING] [/common/utils/oidc/helper.go:72]: Failed to refresh configuration, error: failed to load OIDC setting: write tcp 100.1.1.1::49912->>51.59.7.7:5432: write: broken pipe
  |   |   | 2021-05-11 23:21:15 | 2021-05-11T21:21:15Z [ERROR] [/replication/registry/healthcheck.go:65]: Health check error: write tcp 100.1.1.1::40860->51.59.7.7:5432: write: connection reset by peer


```

I think the health check should 
Harbor 2.2.1 - Don't know if that is also happening on older versions. (This is a new setup)

Vad1mo:
In addition to that the job service does also not recover once temporarily the connection to the DB is dropped

```
2021-05-05T20:01:30Z [ERROR] [/jobservice/logger/sweeper_controller.go:40]: sweep logs error in *sweeper.DBSweeper at 1620244890: sweep logs in DB failed before 2021-04-21 20:01:30.5578332 +0000 UTC m=+1209600.049316437 with error: write tcp 100.11.12.13:43386->51.59.7.7:5432: write: broken pipe
[...]

PR #15210, opened 2021-06-26
Fix/14856

Vad1mo:
This should resolve https://github.com/goharbor/harbor/issues/14856 or make it very unlikely to appear same was also done in the pasts with https://github.com/goharbor/harbor/pull/4719


Vad1mo:
resolved  by https://github.com/goharbor/harbor/pull/15211

Issue #15048, opened 2021-06-02
harbor-core sporadically throws HTTP.5xx - Database performance related 


phoinixgrr:
Running Harbor `v2.2.1,` and using an AWS RDS as database.
Some specs for our Harbor deployment: 
- 300 Repositories 
- 800 artifacts per Repository (approximately) 

We are observing high database usage generated from harbor. We've enabled PostgreSQL `slow_log,` to gather some insights on the database usage. The following queries are reported with duration over 10seconds: 

```
| 1622621735000 | 2021-06-02 08:15:35 UTC:10.1.15.1(1331):harbor@core_db:[21602]:LOG:  duration: 13985.656 ms  execute <unnamed>: SELECT "id", "reference", "reference_id", "hard", "creation_time", "update_time" FROM "quota" WHERE "reference" = $1 AND "reference_id" = $2 FOR UPDATE    | mgmt-harbor-serverless-prod.1 |
| 1622621735000 | 2021-06-02 08:15:35 UTC:10.1.17.254(33891):harbor@core_db:[21663]:LOG:  duration: 14040.258 ms  execute <unnamed>: SELECT "id", "reference", "reference_id", "hard", "creation_time", "update_time" FROM "quota" WHERE "reference" = $1 AND "reference_id" = $2 FOR UPDATE | mgmt-harbor-serverless-prod.1 |
| 1622621735000 | 2021-06-02 08:15:35 UTC:10.1.11.177(40627):harbor@core_db:[21719]:LOG:  duration: 13736.308 ms  execute <unnamed>: SELECT "id", "reference", "reference_id", "hard", "creation_time", "update_time" FROM "quota" WHERE "reference" = $1 AND "reference_id" = $2 FOR UPDATE | mgmt-harbor-serverless-prod.1 |
```

These are `FOR UPDATE` queries in which [causes the rows retrieved by the SELECT statement to be locked as though for update](https://www.postgresql.org/docs/9.0/sql-select.html). 
  
Database Locking graph confirms the DB locking issue, showing 400+ locked tables.  
![Screen Shot 2021-06-02 at 12 23 43 PM](https://user-images.githubusercontent.com/36594156/120455386-62cfcf00-c39d-11eb-94d4-569b85714c50.png)

When this happens, we are sporadically getting `HTTP.5xx` errors thrown by harbor on `PUT` Requests, thus image upload fails for clients: 
```
[...]

PR #15211, opened 2021-06-26
Fixes the broken pipe connection error with postgres.

Vad1mo:
This should resolve https://github.com/goharbor/harbor/issues/14856 or make it very unlikely to appear.  We also did the same in the past with https://github.com/goharbor/harbor/pull/4719

related to #15048


heww (approved):

wy65701436 (approved):
lgtm

Issue #8243, opened 2019-07-09
no way to config multi-hostname for a single harbor instance


rbhuang:
Our scenario is we have deployed Harbor in our data center and we have internal and external hostnames for harbor instance. We are trying to use CI to push images to Harbor's internal hostname. Other systems may fetch the images from Habor through either internal or external hostnames. I found Harbor only support one hostname so I would like to seek your help to check how to config Harbor to support two hostname scenario.
thanks
rb

reasonerjt:
I don't think of good resolution because before read/write to registry you need to fetch a token via the endpoint: https://github.com/goharbor/harbor/blob/master/make/photon/prepare/templates/registry/config.yml.jinja#L31

My very immature thought, maybe you can try to modify the config.yml of registry after installation of Harbor according to
https://docs.docker.com/registry/configuration/
to enable `htpasswd`, and use `htpasswd` for internal.

reasonerjt:
@xaleeks I suggest mark it as `won't fix`

neo502721:
we have save scenario, we scp  images to the harbor host first， then push to harbor

holinnn:
We also have a similar scenario, we need to expose Harbor both internally and externally.
Internally for obvious reasons to push/pull images from clusters and with VPN.
Externally because we use other cloud providers, and this external endpoint will be restricted by IPs.
So we need 2 hostnames (one per LoadBalancer) but it seems that we hit the issue described by @reasonerjt (when doing `docker login` with the external domain it redirects to the internal one)

What would be the best way to achieve this ?
I thought about creating another `core` deployment and service, pointing to the existing one, and exposing this deployement with the external ingress.
But I'm not sure if it will create issues, like race conditions or other problems.

Timoses:
Facing the same issue: We have an internal network (for the clusters to fetch the images) and an external network attached.

Using the Web UI from either network/domain works fine.

However, if the hostname is set to the domain leading to the internal network, then pushing replication from another external harbor instance fails:
```
2020-06-11T10:49:48Z [INFO] [/replication/transfer/image/transfer.go:125]: client for destination registry [type: harbor, URL: https://harbor.ext, insecure: true] created
 2020-06-11T10:49:48Z [INFO] [/replication/transfer/image/transfer.go:158]: copying ourproj/templateservice:[0.0.3](source registry) to destproj/templateservice:[0.0.3](destination registry)...
[...]

PR #15212, opened 2021-06-27
Support for specifing token service url by TOKEN_SERVICE_URL

DanielQujun:
When an external client makes a  pull or push request to Harbor, Harbor will respond extEndpoint as service token authorization URL.
https://github.com/goharbor/harbor/blob/b789674ada2ed21fb5277daeb55657c20788de6b/src/server/middleware/v2auth/auth.go#L120-L125

This is great for most situation; however, there're still some limits.
1. In our container platform, we deploy harbor as container registry service within it. The platform requests extEndpoint of harbor service. And there're multiple platforms installed by the one installer package, and To keep the platform's configuration simple, we set a fixed FQDN to harbor's extEndpoint.So If I  try to push the image from one client to serveral platform's harbor, I was always struck by the token service since all these harbor return the same FQDN token service URL even though the request's HOST was set to different IP.
2. If I try to set harbor's token service Url to another standalone service, there's no entry to doing this.

IMHO, we could define a configurable item for service token URL. 

Luckily, I found there is TOKEN_SERVICE_URL defined in config but no reference on it, so I take it for granted to bypass the limits.
https://github.com/goharbor/harbor/blob/b789674ada2ed21fb5277daeb55657c20788de6b/src/lib/config/metadata/metadatalist.go#L125

Wish I do it in the right direction~

Signed-off-by: 屈骏 <qujun@tiduyun.com>


DanielQujun:
@reasonerjt Hi, would you pls take a look~

ChristianCiach:
This is roughly related to the multi-domain issue: https://github.com/goharbor/harbor/issues/8243#issuecomment-884258765

Whatever will be done here, please make sure that you don't make it harder to support multiple hostnames (or just using the hostname of the current request) in the future.

PR #15214, opened 2021-06-28
Improve login page for OIDC

### LINKED ISSUE #14968: [portal] enhance harbor login form when OIDC auth is configured
This feature request tries to solve the Harbor landing page when the user is not authenticated when Harbor is configured to use other Auth Services (in our case OIDC) 

As discussed with @reasonerjt he proposed to add:
image
Which i also agree, at least makes the user think. This is important as so the user does not expose user:passwd from the organization into the Harbor service.

We where also thinking on adding configurable Text button for the OIDC as some users are not technical. So, having something like "CompanyX SSO". This goes against @reasonerjt wishes of not introducing configuration complexity but i also would like you to consider it :) 

I have no opinion on this

The solution proposed by Daniel its mainly front-end form editing. For the rest is more complex and if accepted we can split into another ticket

**Additional context**
Harbor Login

AllForNothing:
new UI:
image
image

UI  design:
https://www.figma.com/proto/cPY9Lh8LbIQawAJSGmi1KA/Harbor-v2.0?page-id=0%3A1&node-id=2419%3A79553&viewport=4205%2C10144%2C1&scaling=min-zoom

Fixes #14968

### DIFF
--- a/src/portal/src/app/account/sign-in/sign-in.component.html
+++ b/src/portal/src/app/account/sign-in/sign-in.component.html
@@ -2,55 +2,80 @@
     <global-message [isAppLevel]="true"></global-message>
     <navigator (showDialogModalAction)="openModal($event)"></navigator>
     <search-result></search-result>
-    <div class="login-wrapper" [ngStyle]="{'background-image': customLoginBgImg? 'url(/images/' + customLoginBgImg + ')': ''}">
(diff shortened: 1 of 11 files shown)


xuelichao (approved):

jwangyangls (approved):
lgtm

PR #15223, opened 2021-06-29
roll back go for notary binary

### LINKED ISSUE #14932: Harbor Version Upgrade from 2 to 2.2 Causing Key Issue
We ran into a key issue when upgrading from Harbor 2 to 2.2. We were unable to sign / push to Harbor. This was the case for both existing images as well as new images. After reviewing the logs of the notary-signer, we noticed the following error:

kubectl logs -l component=notary-signer -n harbor -f
2021/05/11 15:55:14 Updating database.
2021/05/11 15:55:14 schema_migrations table does not require update, skip.
no change
notarysigner database migrated to latest version
{"level":"info","msg":"Version: 0.6.1, Git commit: d6e1431f","time":"2021-05-11T15:55:14Z"}
{"go.version":"go1.15.6","level":"error","msg":"GetKeyInfo: key fb3da5e64d325bc7088d561b43fcd86f8161cc76830460742d48fb0efb039d70 not found","time":"2021-05-14T17:45:20Z"}

I have been unable to determine what key this is referencing. Furthermore, we had tried to reload all of the signer roles and attempt to sign and push again. This was to no avail.  The error when attempting to sign and push also is extremely vague - unable to reach trust server at this time: 500.

This is as if the keys are in a corrupted state.  We have to remove all the keys and roles, then reload all the roles / keys for each image then resign all the images from scratch in order to remediate. Is there a different way?

wy65701436:
Fixes #14932

Harbor recompiles the notary v0.6.1 with go 1.15 from v2.2.0, which introduces an break change that leads to notary key not found after migration.

[Root cause]
Notary v0.6.1 consumed an old version dvsekhvalnov/jose2, which is not compatible with go 1.15.

[References]
https://github.com/dvsekhvalnov/jose2go/issues/26
https://github.com/golang/go/issues/41089

[Resolve]
To resolve this issue, we have to roll back go vesrion to v1.14 for notary v0.6.1 binary and keep it until upstream have a patch release to support go 1.15 or above.

[Break change]
If you pushed and signed image using Harbor v2.2.0 ~ v2.2.2 and created new repository key in notary,  you will encouter the same issue after migrate to v2.2.3(or above) or v2.3.1(or above) because of the go version downgrade. We will have a FAQ to help you to resovle this particular scenario.

The influence path of the particular case:
Harbor v2.1.0(or lower) --> [v2.2.0 ~ v2.2.2] --> v2.2.3(or above)
Harbor v2.1.0(or lower) --> v2.3.0 --> v2.3.1(or above)

The non influence path of the paticular case:
Harbor v2.1.0(or lower) --> v2.2.3(or above)
Harbor v2.1.0(or lower) --> v2.3.1(or above)

[Fix in Version]
Harbor v2.2.3 or above
Harbor v2.3.1 or above

[Note]
If you're a heavy user of notary, avoid using v2.2.0, v2.2.1, v2.2.2 and v2.3.0, and use the fixed version for instead.


reasonerjt:
Thanks

To emphasize this only impacts if new key is added, shall we rephrase this

```
If you're already on Harbor v2.2.0 ~ v2.2.2 and have signed images with notary
```
to
```
If you pushed and signed image using Harbor v2.2.0 ~ v2.2.2 and created new repository key in this process...
```

heww (approved):

reasonerjt (approved):

PR #15242, opened 2021-07-01
Fix refreshing oidc access token breaks users subiss in database

### LINKED ISSUE #15241: refreshing oidc access token breaks users subiss in database
Refreshing an access token, will lead to an empty `subiss` field in the database. Which then forces harbor to recreate a user that already exists.

Actual:
Refreshing an access token, will lead to an empty `subiss` field in the database. Which then forces harbor to recreate a user that already exists.

Expected:
Refreshing an access token should not empty  `subiss` field in database.

- login to harbor via OIDC
- wait until access token expires
- visit harbor again, so that harbor is forced to refresh the token

**Versions:**
- harbor version: v2.3.0-047b122c

**Additional context:**
The following line updates the oidc_user table in the database:
https://github.com/goharbor/harbor/blob/29ccdff766541aadc0d4eeae6e18676a955ccc23/src/pkg/oidc/secret.go#L115

But when we have a look at how the `oidcUser` object is fetched from the database:
https://github.com/goharbor/harbor/blob/29ccdff766541aadc0d4eeae6e18676a955ccc23/src/pkg/oidc/dao/meta.go#L46

We can see that the `subiss` field is not part of the selected fields and therefore is empty.

I am providng a pull request soon.

kschu91:
### Summary of your change
Ensures that `subiss` field is fetched from the database. Otherwise this field is empty and will cause several issues when working with the user object. Eg. when writing it back to the database it resets the `subiss` in the database.

### Issue being fixed

Fixes #15241

### Please indicate you've done the following:

- [x] [Accepted the DCO](https://github.com/goharbor/harbor/blob/master/CONTRIBUTING.md#commit). Commits without the DCO will delay acceptance.
- [x] Considered the docs impact and opened a new docs issue or PR with docs changes if needed.

### DIFF
--- a/src/pkg/oidc/dao/meta.go
+++ b/src/pkg/oidc/dao/meta.go
@@ -43,7 +43,7 @@ func NewMetaDao() MetaDAO {
 type metaDAO struct{}
 
 func (md *metaDAO) GetByUsername(ctx context.Context, username string) (*models.OIDCUser, error) {
-	sql := `SELECT oidc_user.id, oidc_user.user_id, oidc_user.secret, oidc_user.token,
+	sql := `SELECT oidc_user.id, oidc_user.user_id, oidc_user.secret, oidc_user.subiss, oidc_user.token,
 				oidc_user.creation_time, oidc_user.update_time FROM oidc_user
 			JOIN harbor_user ON oidc_user.user_id = harbor_user.user_id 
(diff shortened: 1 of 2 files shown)


reasonerjt (approved):

wy65701436 (approved):
Lgtm

PR #15263, opened 2021-07-05
Fix grammar

lindhe:
I recently wanted to create a new replication rule and found some translation issues in the tooltip. Since these grammatical changes actually makes a difference I thought it would be best to try and fix them.

* Replace "lower case characters" with "lower case letters" (only letters can be lower or upper case)
    * There's also a comment in `src/vendor/github.com/docker/distribution/reference/regexp.go` referring to "lower case characters" but I skipped changing that to avoid mixing translation and code changes.
* Clarify what characters strings must begin with (since numbers are also characters, so "characters and numbers" makes little sense).

Tooltip for Name in the dialog for "New Replication Rule"

This change is best viewed using:

    git diff --color-words='\w+'


Vad1mo:
@lindhe thank you for the PR. Your DCO is missing and there is a minor conflict with the PT-BR translation.

lindhe:
I merged in master and fixed the conflict and DCO.

My Portuguese is limited, but it seemed like the translation was correct so I didn't change anything there. :)

wy65701436:
hi, please clean the PR so that we can have a review, thanks.

lindhe:
Clean?

lindhe:
Do you mean I should merge in master and fix any merge conflicts? Or what do you mean?

wy65701436:
@lindhe do you still work on this? Could you please squash the commit, and rebase it? It's hard to have a review the changes.

lindhe:
No, with months between responses there's no point of me working on this.

PR #15291, opened 2021-07-08
if username is not available in remote, fall back to username from token

### LINKED ISSUE #15290: Inconsistent username claim behavior
Harbor tries to fetch username and groups from the ID Token and from the USERINFO endpoint:
```go
func mergeUserInfo(remote, local *UserInfo) *UserInfo {
	res := &UserInfo{
		// data only contained in ID token
		Subject: local.Subject,
		Issuer:  local.Issuer,
		// Used data from userinfo
		Username: remote.Username,
		Email:    remote.Email,
	}
	if remote.hasGroupClaim {
		res.Groups = remote.Groups
		res.AdminGroupMember = remote.AdminGroupMember
		res.hasGroupClaim = true
	} else if local.hasGroupClaim {
		res.Groups = local.Groups
		res.AdminGroupMember = local.AdminGroupMember
		res.hasGroupClaim = true
	} else {
		res.Groups = []string{}
	}
	return res
}
```
It assumes that the Username is always fetchted from the remote.

Now consider the following use case:
An IDP that has claims for username and groups in the token. Eg. `cognito:username` and `cognito:groups`.
But in the userinfo endpoint of the IDP the username field is just named `username` and groups does not exist at all in the remote.

In the OIDC settings of harbor I am now forced to specify the `user_claim` as `username`, even though the actual claim is `cognito:username`.
This leads to misunderstanding of the setting, as it is not clear that the settings value is not the actual claim in the token instead it is the field in the userinfo endpoint. When adding the actual claim value `cognito:username`, onboarding of a user fails with:
```
Failed to recover Username from claim. Claim 'cognito:username' is invalid or not a string
```

Expected:
Fetch username **and** groups from the remote first. If not available in the remote, fall back to information in the token.

- setup harbor oidc with aws cognito idp
- `cognito:username` as user claim
- `cognito:groups` as groups claim

**Versions:**

- harbor version: v2.3.0-047b122c

kschu91:
### Summary of your change
Ensures that username is used from token if not present in remote. Otherwise username is always fetched from userinfo endpoint even though the configured username claim does not exist in userinfo endpoint (more details in #15290).

### Issue being fixed

Fixes #15290

### Please indicate you've done the following:

- [x] [Accepted the DCO](https://github.com/goharbor/harbor/blob/master/CONTRIBUTING.md#commit). Commits without the DCO will delay acceptance.
- [x] Considered the docs impact and opened a new docs issue or PR with docs changes if needed.

### DIFF
--- a/src/pkg/oidc/helper.go
+++ b/src/pkg/oidc/helper.go
@@ -350,7 +350,6 @@ func userInfoFromClaims(c claimsProvider, setting cfgModels.OIDCSetting) (*UserI
 		if username, ok := allClaims[setting.UserClaim].(string); ok {
 			res.Username = username
 		} else {
-			log.Debugf("OIDC. Failed to recover Username from claims: %+v", allClaims)
 			log.Warningf("OIDC. Failed to recover Username from claim. Claim '%s' is invalid or not a string", setting.UserClaim)
 		}
 	}
--- a/src/pkg/oidc/helper.go
+++ b/src/pkg/oidc/helper.go
@@ -289,8 +289,12 @@ func mergeUserInfo(remote, local *UserInfo) *UserInfo {
 		Subject: local.Subject,
 		Issuer:  local.Issuer,
 		// Used data from userinfo
-		Username: remote.Username,
-		Email:    remote.Email,
+		Email: remote.Email,
+	}
+	if remote.Username != "" {
+		res.Username = remote.Username
+	} else {
+		res.Username = local.Username
 	}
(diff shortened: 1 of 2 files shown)


reasonerjt (approved):

reasonerjt:
Adding the label to remind me cherrypicking it to v2.3.x

reasonerjt:
For some unknown reason the git actions were not triggered, closing and reopening it to see if it helps...

heww (approved):

PR #15295, opened 2021-07-09
Update helper.go

wutongjie23hao:
issue 校验去除


reasonerjt:
We should not skip issuer checking, it's probably your env issue.
Please keep this change to your setup.

PR #15296, opened 2021-07-09
deprecate version 1 robot account

wy65701436:
1, deprecate support for version 1 robot support, the robotv1 cannot be used anymore.
2, reserve the /project/{id_or_name}/robots api.

After the PR, user cannot use the robotv1 to login, and do any interaction with Harbor,
but still can view & delete them with UI or API.


heww:
Why do we remove the robot APIs in the project scope? 

In my opinion, deprecating the v1 robot account means the v1 robots are not accepted by the system. 

The robots created by the APIs in the project scope are version 2 so we don't need to remove the APIs.

heww (approved):

src/server/middleware/security/robot.go
```diff
@@ -1,74 +1,76 @@
...
-// limitations under the License.
-
 package security
 
 import (
-	"net/http"
-	"strings"
-
-	"github.com/goharbor/harbor/src/common"
+	"fmt"
```
reasonerjt, line 4: why remove the comment section? and the import should be re-ordered.

src/server/middleware/security/robot.go
```diff
@@ -1,74 +1,76 @@
...
-	defaultOpt := pkg_token.DefaultTokenOptions()
-	if defaultOpt == nil {
-		log.Error("failed to get default token options")
+	if !strings.HasPrefix(name, config.RobotPrefix(req.Context())) {
 		return nil
 	}
-	rtk, err := pkg_token.Parse(defaultOpt, robotTk, rClaims)
+	// The robot name can be used as the unique identifier to locate robot as it contains the project name.
+	robots, err := robot_ctl.Ctl.List(req.Context(), q.New(q.KeyWords{
+		"name": strings.TrimPrefix(name, config.RobotPrefix(req.Context())),
```
reasonerjt, line 33: not sure how the config.RobotPrefix is implemented, but you don't need to call it twice in this func.

src/server/middleware/security/robot.go
```diff
@@ -1,74 +1,76 @@
...
-		log.Debugf("failed to decrypt robot token of v1 robot: %s, as: %v", robotName, err)
+		log.Errorf("failed to list robots: %v", err)
 		return nil
 	}
-	// Do authn for robot account, as Harbor only stores the token ID, just validate the ID and disable.
-	ctr := ctl_robot.Ctl
-	robot, err := ctr.Get(req.Context(), rtk.Claims.(*robot_claim.Claim).TokenID, nil)
-	if err != nil {
-		log.Errorf("failed to get robot %s: %v", robotName, err)
+	if len(robots) == 0 {
```
reasonerjt, line 41: print a debug log to indicate the specific robot name is not found?

reasonerjt (approved):

PR #15303, opened 2021-07-12
Remediate the empty subiss problem

reasonerjt:
This commit include 2 changes to mitigate and remediate the problem
described in #15241

1. When the token is to be updated in the "oidc_user" table, make sure
   only the column "token" will be udpated.
2. Restore the subiss column for the record that has this column cleared
   by mistake, by decoding the persisted token.

### DIFF
--- a/src/core/main.go
+++ b/src/core/main.go
@@ -27,6 +27,7 @@ import (
 	"time"
 
 	configCtl "github.com/goharbor/harbor/src/controller/config"
+	"github.com/goharbor/harbor/src/pkg/oidc"
 
 	"github.com/astaxie/beego"
 	_ "github.com/astaxie/beego/session/redis"
@@ -234,6 +235,9 @@ func main() {
 	}
 
 	log.Infof("Version: %s, Git commit: %s", version.ReleaseVersion, version.GitCommit)
+
+	log.Info("Fix empty subiss for meta info data.")
+	oidc.FixEmptySubIss(orm.Context())
 	beego.RunWithMiddleWares("", middlewares.MiddleWares()...)
 }
 
--- a/src/pkg/oidc/fix.go
+++ b/src/pkg/oidc/fix.go
@@ -1,0 +1,70 @@
+package oidc
+
+import (
+	"context"
+	"encoding/base64"
+	"encoding/json"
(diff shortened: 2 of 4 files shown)


src/core/main.go
```diff
@@ -234,6 +235,9 @@ func main() {
 	}
 
 	log.Infof("Version: %s, Git commit: %s", version.ReleaseVersion, version.GitCommit)
+
+	log.Info("Fix empty subiss for meta info data.")
+	oidc.FixEmptySubIss(orm.Context())
```
wy65701436, line 240: if auth_mode = "OIDC", then fix?

src/pkg/oidc/fix.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/goharbor/harbor/src/pkg/oidc/dao"
+)
+
+// FixEmptySubIss remediates the issue https://github.com/goharbor/harbor/issues/15241
+// by restoring the subiss via the persisted token
+func FixEmptySubIss(ctx context.Context) error {
+	metaMgr := NewMetaMgr()
+	meta, err := metaMgr.GetBySubIss(ctx, "", "")
+	if errors.IsNotFoundErr(err) {
+		log.Infof("Not found any records with empty subiss, good to go.")
```
wy65701436, line 23: log.Info?

src/pkg/oidc/fix.go
```diff
@@ -0,0 +1,70 @@
+package oidc
```
wy65701436, line 1: add UT?

wy65701436 (approved):
lgtm

src/core/main.go
reasonerjt: Though this problem impact oidc auth mode only but we should try to fix data inconsistency any way regardless the auth mode.

src/pkg/oidc/fix.go
reasonerjt: ok

src/pkg/oidc/fix.go
reasonerjt: added

heww (approved):

PR #15306, opened 2021-07-12
🎈 perf(migrations/postgresql): index of quota table

ruster-cn:
Because the values of the reference field in the quota table are basically the same, the reference field is not the same_ ID is more distinguishable. Index query follows forward matching rule. So the reference_ If Id is put in front of index, the query efficiency will be higher.


wy65701436:
hi @ruster-cn according to our testing results, this PR cannot have improvment for performance. Please give more infor, either some tech materials or testing result from your side, then reopen it.

PR #15326, opened 2021-07-16
drop adminjob router

wy65701436:
As the admin job has already droped from v2.3, remove the request handling from router.

### DIFF
--- a/tests/apitests/python/library/system.py
+++ b/tests/apitests/python/library/system.py
@@ -71,52 +71,6 @@ class System(base.Base):
         base._assert_status_code(expect_status_code, status_code)
         return data
 
-    def set_gc_schedule(self, schedule_type = 'None', cron = None, expect_status_code = 200, expect_response_body = None, **kwargs):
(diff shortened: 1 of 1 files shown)


heww (approved):

src/server/route.go
```diff
@@ -50,7 +50,6 @@ func registerRoutes() {
 	beego.Router("/api/internal/syncquota", &api.InternalAPI{}, "post:SyncQuota")
 
 	beego.Router("/service/notifications/jobs/webhook/:id([0-9]+)", &jobs.Handler{}, "post:HandleNotificationJob")
-	router.NewRoute().Method(http.MethodPost).Path("/service/notifications/jobs/adminjob/:id([0-9]+)").Handler(handler.NewJobStatusHandler())         // legacy job status hook endpoint for adminjob
```
ywk253100, line 53: It is possible that there are still some legacy admin jobs running inside Jobservice, so we should not remove these routes?

ywk253100 (approved):

PR #15329, opened 2021-07-16
[CP-v2.3]:Remediate the empty subiss problem

reasonerjt:
This commit include 2 changes to mitigate and remediate the problem
described in #15241

1. When the token is to be updated in the "oidc_user" table, make sure
   only the column "token" will be udpated.
2. Restore the subiss column for the record that has this column cleared
   by mistake, by decoding the persisted token.


heww (approved):

wy65701436 (approved):
lgtm

PR #15330, opened 2021-07-16
[CP - v2.3]subiss needs to be part of selected fields, otherwise this attribute …

reasonerjt:
…would be empty and will cause issues when working with the user object later on.

Signed-off-by: Kevin Schu <kevin.schu@aoe.com>


wy65701436 (approved):
lgtm

heww (approved):

PR #15384, opened 2021-08-03
add option to not automatically start Harbor in the install-script

msc-xdev:
I felt like there needed to be an option to not start Harbor, if you need to edit the docker-compose.yml for example.


ninjadq:
you can just run prepare script which will do all the prepare tasks without start harbor

msc-xdev:
> you can just run prepare script which will do all the prepare tasks without start harbor

i want the checks of the install.sh, prepare doesn't have those.
for example `check_docker` or the loading of Harbor Images

i just want the ability to edit the `docker-compose.yml` with all the benefits of the install-script but without having to wait for the containers to be stated and then stopped again.

PR #15393, opened 2021-08-06
Adjust "flattening" tooltip quotes to be consistent

tianon:
This is a minor consistency issue, but it also turns into a minor rendering issue where something is converting "backtick + a" into "à"


AllForNothing (approved):
LGTM

danfengliu (approved):

PR #15399, opened 2021-08-09
refactor: implement a lock free quota

heww:
Using optimistic lock to implement the quota.

### DIFF
--- a/make/migrations/postgresql/0070_2.4.0_schema.up.sql
+++ b/make/migrations/postgresql/0070_2.4.0_schema.up.sql
@@ -2,3 +2,7 @@
 DELETE FROM project_member pm WHERE pm.entity_type = 'u' AND EXISTS (SELECT NULL FROM harbor_user u WHERE pm.entity_id = u.user_id AND u.deleted = true );
 
 ALTER TABLE replication_policy ADD COLUMN IF NOT EXISTS speed_kb int;
+
+/* add version fields for lock free quota */
+ALTER TABLE quota ADD COLUMN IF NOT EXISTS version bigint DEFAULT 0;
+ALTER TABLE quota_usage ADD COLUMN IF NOT EXISTS version bigint DEFAULT 0;
--- a/src/controller/quota/controller.go
+++ b/src/controller/quota/controller.go
@@ -23,19 +23,17 @@ import (
 	"github.com/goharbor/harbor/src/lib/log"
 	"github.com/goharbor/harbor/src/lib/orm"
 	"github.com/goharbor/harbor/src/lib/q"
-	redislib "github.com/goharbor/harbor/src/lib/redis"
+	"github.com/goharbor/harbor/src/lib/retry"
 	"github.com/goharbor/harbor/src/pkg/quota"
 	"github.com/goharbor/harbor/src/pkg/quota/driver"
 	"github.com/goharbor/harbor/src/pkg/quota/types"
-	"github.com/gomodule/redigo/redis"
 
 	// quota driver
 	_ "github.com/goharbor/harbor/src/controller/quota/driver"
 )
 
 var (
-	// expire reserved resources when no actions on the key of the reserved resources in redis during 1 hour
-	defaultReservedExpiration = time.Hour
+	defaultRetryTimeout = time.Minute * 5
 )
 
 var (
@@ -82,8 +80,7 @@ type Controller interface {
 // NewController creates an instance of the default quota controller
 func NewController() Controller {
 	return &controller{
-		reservedExpiration: defaultReservedExpiration,
-		quotaMgr:           quota.Mgr,
+		quotaMgr: quota.Mgr,
 	}
 }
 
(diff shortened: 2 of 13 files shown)


make/migrations/postgresql/0070_2.4.0_schema.up.sql
```diff
@@ -1,2 +1,7 @@
 /* cleanup deleted user project members */
-DELETE FROM project_member pm WHERE pm.entity_type = 'u' AND EXISTS (SELECT NULL FROM harbor_user u WHERE pm.entity_id = u.user_id AND u.deleted = true )
\ No newline at end of file
+DELETE FROM project_member pm WHERE pm.entity_type = 'u' AND EXISTS (SELECT NULL FROM harbor_user u WHERE pm.entity_id = u.user_id AND u.deleted = true );
+
+ALTER TABLE quota ADD COLUMN IF NOT EXISTS version bigint;
+ALTER TABLE quota_usage ADD COLUMN IF NOT EXISTS version bigint;
```
wy65701436, line 5: cannot the two sqls be merged into one?

src/pkg/quota/dao/dao.go
```diff
@@ -171,47 +170,57 @@ func (d *dao) GetByRef(ctx context.Context, reference, referenceID string) (*mod
...
-}
+	var (
+		sql    string
+		params []interface{}
+	)
 
-func (d *dao) Update(ctx context.Context, quota *models.Quota) error {
-	o, err := orm.FromContext(ctx)
+	if quota.UsedChanged {
+		sql = "UPDATE quota_usage SET used = ?, update_time = ?, version = ? WHERE id = ? AND version = ?"
```
wy65701436, line 193: In case someone use the updated version in the following logic. Should not we refresh the it for model quota after it's updated?

src/controller/quota/controller.go
```diff
@@ -167,107 +164,46 @@ func (c *controller) List(ctx context.Context, query *q.Query, options ...Option
...
+		lib.RetryCallback(func(err error, sleep time.Duration) {
+			log.G(ctx).Debugf("failed to update the quota usage, error: %v", err)
+		}),
 	}
-
-	return orm.WithTransaction(unreserve)(ctx)
+	return lib.RetryUntil(f, options...)
 }
 
 func (c *controller) Refresh(ctx context.Context, reference, referenceID string, options ...Option) error {
```
wy65701436, line 209: Since we add retry for the Refresh, it may cause the performance issue for API on DeleteArtifact & DeleteRepository in the failure case, is it expected?

src/controller/quota/controller.go
```diff
@@ -167,107 +164,46 @@ func (c *controller) List(ctx context.Context, query *q.Query, options ...Option
...
-			log.G(ctx).Errorf("failed to set reserved resources for %s %s, error: %v", reference, referenceID, err)
-			return err
-		}
-
-		return nil
+	options := []lib.RetryOption{
+		lib.RetryTimeout(defaultRetryTimeout),
+		lib.RetryBackoff(false),
+		lib.RetryCallback(func(err error, sleep time.Duration) {
+			log.G(ctx).Debugf("failed to update the quota usage, error: %v", err)
```
wy65701436, line 203: can the reference and referenceID be logged?

src/controller/quota/controller.go
```diff
@@ -19,23 +19,21 @@ import (
...
-	"github.com/gomodule/redigo/redis"
 
 	// quota driver
 	_ "github.com/goharbor/harbor/src/controller/quota/driver"
 )
 
 var (
-	// expire reserved resources when no actions on the key of the reserved resources in redis during 1 hour
-	defaultReservedExpiration = time.Hour
+	defaultRetryTimeout = time.Minute * 5
```
wy65701436, line 36: It may cause some APIs failed by http time out, but the quota still be updated in the backend?

src/controller/quota/controller.go
```diff
@@ -167,107 +164,46 @@ func (c *controller) List(ctx context.Context, query *q.Query, options ...Option
...
-		return fmt.Errorf("bad reply value")
-	}
-
-	return nil
-}
-
-func (c *controller) reserveResources(ctx context.Context, reference, referenceID string, resources types.ResourceList) error {
-	reserve := func(ctx context.Context) error {
-		q, err := c.quotaMgr.GetByRefForUpdate(ctx, reference, referenceID)
+func (c *controller) updateUsage(ctx context.Context, reference, referenceID string, op func(hardLimits, used types.ResourceList) (types.ResourceList, error)) error {
```
wy65701436, line 167: name it as `retryUpdateUsage` or `updateUsageWithRetry`? Then the caller could know this func with retry.

src/controller/quota/controller.go
heww: Using the `select for update` has the same problem.

I think this is not a problem if the Nginx closed the TCP conn when the timeout happened. In this case, writing the data to the response writer will fail, and the tx will be abort.

src/controller/quota/controller.go
heww: Using the `select for update` has the same problem.

src/pkg/quota/dao/dao.go
heww: The callers need to get the quota again themself when they want to update the quota again after they updated the quota variable, otherwise, the second updating will always fail.

ywk253100 (approved):

wy65701436 (approved):
lgtm

PR #15431, opened 2021-08-15
support robot to list project

wy65701436:
1, add permission check for API of List Project
2, use the self defined query to handle both names and public query

### DIFF
--- a/src/pkg/project/models/project.go
+++ b/src/pkg/project/models/project.go
@@ -56,6 +56,12 @@ type Project struct {
 	RegistryID   int64                  `orm:"column(registry_id)" json:"registry_id"`
 }
 
+// NamesQuery ...
+type NamesQuery struct {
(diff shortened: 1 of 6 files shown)


src/pkg/project/models/project.go
```diff
@@ -182,6 +188,30 @@ func (p *Project) FilterByMember(ctx context.Context, qs orm.QuerySeter, key str
...
+		return qs
+	}
+
+	if len(query.Names) == 0 {
+		return qs
+	}
+
+	var names []string
+	for _, v := range query.Names {
+		names = append(names, `'`+v+`'`)
```
heww, line 204: suggest using the lib/pg.QuoteLiteral here

heww (approved):

ywk253100 (approved):

PR #15438, opened 2021-08-16
Helm Chart should not be translated.

tan9:
And the existing translation is also not accurate to the meaning of Helm.


AllForNothing:
@tan9 Thanks for your PR
Can sign off your PR by run "git commit -s --amend",  this is required by DCO check

tan9:
@AllForNothing done the signing :+1:

danfengliu (approved):

PR #15448, opened 2021-08-18
Disable add group button for LDAP project admin

### LINKED ISSUE #15432: Projects admins can't add groups to projects when using LDAP
We are using Harbor 2.3.0 together with LDAP for authentication. We are finding that project admins can't add groups to projects and only Harbor admins can add groups. 
When a project admin clicks the group button only a spinning circle appears as if there is no response or time out. When performing the same procedure as a harbor admin the popup works as expected.

Screenshot 2021-08-16 at 09 31 44

Looking at the browsers developer console we can see that a call to /usergroup fails with a 403.
Manually calling the usergroup function from the API-helper results in the same 403.
 
We are not seeing any errors in the Harbor logs, or any kind of logging related to the action.
Is this by design or is it a possible misconfiguration on our side?

AllForNothing:
Fixes #15432


ninjadq (approved):
lgtm

stonezdj (approved):

PR #15505, opened 2021-08-28
Add distributed tracing

ninjadq:


src/lib/trace/config.go
```diff
@@ -0,0 +1,97 @@
...
+	C.Otel.Endpoint = viper.GetString("otel_endpoint")
+	C.Otel.URLPath = viper.GetString("otel_url_path")
+	C.Otel.Compression = viper.GetBool("otel_compression")
+	C.Otel.Insecure = viper.GetBool("otel_insecure")
+	C.Otel.Timeout = viper.GetInt("otel_timeout")
+	var jsonExample = []byte(viper.GetString("attributes"))
+	viper.ReadConfig(bytes.NewBuffer(jsonExample))
+	fmt.Println(viper.GetStringMapString("attributes"))
+	C.Attributes = viper.GetStringMapString("attributes")
+	log.Infof("ns: %s attr %+v", C.Namespace, C.Attributes)
```
stonezdj, line 57: All these configure items should be put into the system config.
Refer https://github.com/goharbor/harbor/blob/107e468b6017c74ca899ef90414789d7a0d90b01/src/lib/config/metadata/metadatalist.go#L66

src/lib/orm/orm.go
```diff
@@ -84,27 +88,34 @@ func Clone(ctx context.Context) context.Context {
 // WithTransaction a decorator which make f run in transaction
 func WithTransaction(f func(ctx context.Context) error) func(ctx context.Context) error {
 	return func(ctx context.Context) error {
+		_, span := tracelib.StartTrace(ctx, tracerName, "start-transaction")
```
stonezdj, line 91: It is better to pass the operation name by context and display the operation name instead of "start-transaction"

src/lib/orm/orm.go
```diff
@@ -84,27 +88,34 @@ func Clone(ctx context.Context) context.Context {
...
 		}
 
 		tx := ormerTx{Ormer: o}
 		if err := tx.Begin(); err != nil {
+			tracelib.RecordError(span, err, "begin transaction failed")
 			log.Errorf("begin transaction failed: %v", err)
 			return err
 		}
 
 		if err := f(ctx); err != nil {
```
stonezdj, line 106: The context created by the tracelib.StartTrace() should propagate to the children span

src/common/http/transport.go
```diff
@@ -0,0 +1,136 @@
...
+		ExpectContinueTimeout: 1 * time.Second,
+	}
+}
+
+// WithInternalTLSConfig returns a TransportOption that configures the transport to use the internal TLS configuration
+func WithInternalTLSConfig() func(*http.Transport) {
+	return func(tr *http.Transport) {
+		tlsConfig, err := GetInternalTLSConfig()
+		if err != nil {
+			panic(err)
```
wy65701436, line 75: do we need panic here?

src/jobservice/main.go
```diff
@@ -57,6 +60,9 @@ func main() {
 		panic(err)
 	}
 
+	cfgLib.InitTraceConfig(ctx)
+	defer tracelib.InitGlobalTracer(context.Background()).Shutdown()
```
wy65701436, line 64: defer should move to the entry of this func.

src/jobservice/runner/redis.go
```diff
@@ -94,7 +103,7 @@ func (rj *RedisJob) Run(j *work.Job) (err error) {
 
 		// Log error and exit
 		logger.Errorf("Job '%s:%s' exit with error: failed to get job tracker: %s", j.Name, j.ID, err)
-
+		tracelib.RecordError(span, err, "failed to get job tracker")
```
wy65701436, line 106: do we need to track the id, name... ?

src/jobservice/runner/redis.go
```diff
@@ -137,7 +149,9 @@ func (rj *RedisJob) Run(j *work.Job) (err error) {
 		metric.JobservieTaskProcessTimeSummary.WithLabelValues(j.Name, "success").Observe(time.Since(now).Seconds())
 		// Mark job status to success.
 		logger.Infof("Job '%s:%s' exit with success", j.Name, j.ID)
+		span.SetStatus(codes.Ok, "job exit with success")
 		if er := tracker.Succeed(); er != nil {
+			tracelib.RecordError(span, err, "failed to mark job success")
```
wy65701436, line 154: ditto

src/lib/config/metadata/value.go
```diff
@@ -95,6 +95,21 @@ func (c *ConfigureValue) GetInt64() int64 {
 	return 0
 }
 
+func (c *ConfigureValue) GetFloat64() float64 {
```
wy65701436, line 98: add UT?

src/lib/config/trace.go
```diff
@@ -0,0 +1,37 @@
...
+	"context"
+
+	"github.com/goharbor/harbor/src/common"
+	commonhttp "github.com/goharbor/harbor/src/common/http"
+	"github.com/goharbor/harbor/src/lib/log"
+	tracelib "github.com/goharbor/harbor/src/lib/trace"
+)
+
+func InitTraceConfig(ctx context.Context) {
+	cfgMgr, err := GetManager(common.InMemoryCfgManager)
```
wy65701436, line 13: Is it right to use the memory rather than restcfgManager? @stonezdj correct me.

ninjadq: rest Manager fetch data through the core component, but some data are special for jobservice like service name, so it need to parse its own env

src/jobservice/runner/redis.go
ninjadq: Yes, if no job name, user will not know what kind of job they are tracing. And job id used to identify job, if user wanna know the details of job, they can use these id to check job log

src/core/main.go
```diff
@@ -86,17 +88,27 @@ func updateInitPassword(ctx context.Context, userID int, password string) error
...
 	signals := make(chan os.Signal, 1)
 	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
 	log.Infof("capture system signal %s, to close \"closing\" channel", <-signals)
 	close(closing)
-	select {
-	case <-done:
+	shutdownChan := make(chan struct{}, 1)
+	go func() {
+		for _, s := range shutdowns {
+			s()
```
wy65701436, line 99: do we need to log the shutdown error?

src/jobservice/runner/redis.go
```diff
@@ -28,10 +29,14 @@ import (
...
 	"github.com/goharbor/harbor/src/lib/errors"
 	"github.com/goharbor/harbor/src/lib/metric"
+	tracelib "github.com/goharbor/harbor/src/lib/trace"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
 )
 
 const (
 	maxTrackRetries = 6
+	tracerName      = "goharbor/harbor/src/jobservice/runner/redis"
```
wy65701436, line 39: can we add a lib func to get the currently package path?

src/lib/orm/orm.go
wy65701436: we need to check the nil of span that returned by StartTrace. Or we can do it in the StartTrace generally.

make/harbor.yml.tmpl
```diff
@@ -199,3 +199,30 @@ proxy:
...
+# trace:
+#   enabled: true
+#   # set sample_rate to 1 if you wanna sampling 100% of trace data; set 0.5 if you wanna sampling 50% of trace data, and so forth
+#   sample_rate: 1
+#   # # namespace used to diferenciate different harbor services
+#   # namespace:
+#   # # attributes is a key value dict contains user defined attributes used to initialize trace provider
+#   # attributes:
+#   #   application: harbor
+#   # jaeger:
```
wy65701436, line 217: When we parse these settings, the logic is either endpoint or agent, but uses the length as the judgment. 

Most likely is, user uncomment jadge, but let the endpoint be the default value(http://hostname:14268/api/traces). But, the code goes into the endpoint path, right?

src/lib/trace/trace.go
```diff
@@ -0,0 +1,139 @@
...
+
+	"github.com/goharbor/harbor/src/lib/log"
+	"github.com/goharbor/harbor/src/pkg/version"
+)
+
+func initExporter(ctx context.Context) (tracesdk.SpanExporter, error) {
+	var err error
+	var exp tracesdk.SpanExporter
+	cfg := GetGlobalConfig()
+	if len(cfg.Jaeger.Endpoint) != 0 {
```
wy65701436, line 39: When we parse these settings, the logic is either endpoint or agent, but uses the length as the judgment. 

Most likely, user uncomments the jadge part, but let the endpoint be the default value(http://hostname:14268/api/traces). But, the code goes into the endpoint path, right?

We need add more logs here.

src/lib/trace/trace.go
```diff
@@ -0,0 +1,139 @@
...
+			opts = append(opts, otlptracehttp.WithInsecure())
+		}
+		exp, err = otlptracehttp.New(ctx, opts...)
+	} else {
+		log.Fatalf("Trace is enabled but no tracer provider is specified")
+	}
+	return exp, err
+}
+
+func initProvider(exp tracesdk.SpanExporter) (*tracesdk.TracerProvider, error) {
```
wy65701436, line 75: There is literally no error will be raised?

src/registryctl/api/registry/blob/blob.go
```diff
@@ -33,14 +40,19 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
 	if ref == "" {
-		api.HandleBadRequest(w, errors.New("no reference specified"))
+		err := errors.New("no reference specified")
+		tracelib.RecordError(span, err, "no reference specified")
+		api.HandleBadRequest(w, err)
 		return
 	}
 	// don't parse the reference here as RemoveBlob does.
-	cleaner := storage.NewVacuum(r.Context(), h.storageDriver)
+	cleaner := storage.NewVacuum(ctx, h.storageDriver)
```
wy65701436, line 53: is the ctx same as r.Context()?

src/registryctl/api/registry/manifest/manifest.go
```diff
@@ -1,16 +1,23 @@
...
+	storagedriver "github.com/docker/distribution/registry/storage/driver"
+	tracelib "github.com/goharbor/harbor/src/lib/trace"
 	"github.com/gorilla/mux"
 	"github.com/opencontainers/go-digest"
-	"net/http"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
 )
 
+const tracerName = "goharbor/harbor/src/registryctl/api/registry/blob"
```
wy65701436, line 19: goharbor/harbor/src/registryctl/api/registry/manifest ?

src/registryctl/api/registry/manifest/manifest.go
```diff
@@ -34,25 +41,37 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 
 // delete deletes manifest ...
 func (h *handler) delete(w http.ResponseWriter, r *http.Request) {
+	var span trace.Span
+	ctx := r.Context()
 	ref := mux.Vars(r)["reference"]
+	if tracelib.Enabled() {
+		ctx, span = tracelib.StartTrace(ctx, tracerName, "delete-manifest", trace.WithAttributes(attribute.Key("method").String(r.Method)))
```
wy65701436, line 48: ctx is overwrited here, is it expected?

src/server/middleware/log/log.go
```diff
@@ -30,6 +34,9 @@ func Middleware() func(http.Handler) http.Handler {
 			logger.Debugf("attach request id %s to the logger for the request %s %s", rid, r.Method, r.URL.Path)
 
 			ctx := log.WithLogger(r.Context(), logger.WithFields(log.Fields{"requestID": rid}))
+			if tracelib.Enabled() {
```
wy65701436, line 37: why not do it in the request id middleware?

src/lib/trace/trace.go
```diff
@@ -0,0 +1,139 @@
...
+
+// Shutdown shutdown the trace provider
+func (s ShutdownFunc) Shutdown() {
+	s()
+}
+
+// Init initializes the trace provider
+func InitGlobalTracer(ctx context.Context) ShutdownFunc {
+	if !Enabled() {
+		otel.SetTracerProvider(oteltrace.NewNoopTracerProvider())
```
wy65701436, line 120: what about if we do not set the NoopTracerProvider when it's disabled? Seems no impact?

src/lib/trace/trace.go
```diff
@@ -0,0 +1,139 @@
...
+		attriSlice = append(attriSlice, semconv.ServiceNamespaceKey.String(cfg.Namespace))
+	}
+	if cfg.Attributes != nil {
+		for i, a := range cfg.Attributes {
+			attriSlice = append(attriSlice, attribute.String(i, a))
+		}
+	}
+
+	// prepare tp options
+	ops := make([]tracesdk.TracerProviderOption, 0, 4)
```
wy65701436, line 95: I am not quite sure that the interval and message size can be set when to init a provider? and it's not exposed to the end-user. 

We can discuss it, but this will not impact this PR.

src/lib/trace/trace.go
ninjadq: yes, but we need to add tremendous code like `if trace.Enabled ...` to the place where you wanna add tracing

src/jobservice/main.go
ninjadq: place it here because of reuse the ctx

src/core/main.go
ninjadq: I think error should be handled by shutdown function, the golang HTTP server library has a similar design, its registered shutdown functions do not need to return the error for it.

src/jobservice/runner/redis.go
ninjadq: if implement a get current path function, it must be returned this value in runtime. because the app is running with binary, it's hard to get source code file name info. It will introduce a lot of troubles.

src/jobservice/runner/redis.go
ninjadq: do you mean, we need to add job id and name in this line?
no, we didn't, because the line previous code 
```
	span.SetAttributes(attribute.Key("jobID").String(jID), attribute.Key("jobName").String(j.Name))
```
already set those attribute

src/registryctl/api/registry/blob/blob.go
ninjadq: yes, it's source ctx is `r.Context()`,  and `tracelib.StartTrace(r.Context()...` added some more info in it

src/registryctl/api/registry/manifest/manifest.go
ninjadq: yes, it added trace info in ctx,  if related code in docker distribution has to trace logic, it will link its span to harbor's span inside one trace. And it's not harmful if docker distribution don't need such trace info.

src/lib/orm/orm.go
ninjadq: nil span will create a new span from scratch

src/jobservice/runner/redis.go
ninjadq: the same as above

src/lib/trace/trace.go
ninjadq: I think we can discuss it later. I think we've already introduced a lot of new configs. if a lot of user needs it when they deployed it in production we can add them in the future.

wy65701436 (approved):
lgtm

src/lib/config/trace.go
ninjadq: But in the case of tracing, the env var like `TRACE_SERVICE_NAME` has different values for harbor-core and harbor-job service. But if we use the `RestConfigManager` the value will be the same as harbor-core which is not our intention. As a result, either way, we need an `InMemoryCfgManager`.

stonezdj (approved):

PR #15520, opened 2021-08-31
missing syslog home directory breaks logrotate

### LINKED ISSUE #15468: log rotation is not working in recent release
1. test with harbor 2.2.0 and harbor 2.3.0
2. change rotate size in `/home/harbor/harbor/common/config/log/logrotate.conf`
3. `docker exec -ti harbor-log bash` and check logs in /var/log/docker
4. script in /etc/cron.hourly/logrotate seems not run hourly
5. if run logrotate manually, logs can rotate

sevendials:
Logrotate is run with sudo as the syslog user by cron.hourly 
Without the syslog home directory, the job fails. Currently the following stderr is being thrown away by the cron script:
```
error: cannot open current directory: Permission denied 
```

Fixes #15468

### DIFF
--- a/make/photon/log/Dockerfile.base
+++ b/make/photon/log/Dockerfile.base
@@ -2,6 +2,6 @@ FROM photon:4.0
 
 RUN tdnf install -y cronie rsyslog logrotate shadow tar gzip sudo >> /dev/null\
     && mkdir /var/spool/rsyslog \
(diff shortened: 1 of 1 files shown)


sevendials:
I'm closing this as it doesn't solve the problem on its own. I created a newer PR https://github.com/goharbor/harbor/pull/15527 which addresses the issue.

PR #15523, opened 2021-09-01
fix: use ctx from http request for middlewares

### LINKED ISSUE #15474: Concurrency block while creating/deleting objects
currently i'm trying to run harbor/perf in a cluster i have setup using harbor/helm.
The core pods are configured to be 3 and the connections are limited to 20/5 (max open/max idle)

I notice if running VU >> (pods * max connect) the core server stalls, blocking everything. If i check the DB i have many pending transaction connections in a blocked state. 
The DB server is not under stress, so this seems to be a problem with the DB factories.

To be honest, i still could not make the perf test work 1 single time. The script blocks always in **prepare** step  3 **scripts/data/03-project-member.js**.

Any pointers?

heww:
1. Use ctx from http request for the readonly middleware.
2. Refactor the AuthenticateHelper to let it get orm from ctx of the http request.
3. Change to use ctx from http request for oidc and authproxy http handlers.

Closes #15474 

### DIFF
--- a/src/controller/gc/callback.go
+++ b/src/controller/gc/callback.go
@@ -1,13 +1,27 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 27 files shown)


wy65701436 (approved):
lgtm

Issue #15338, opened 2021-07-19
How to stop scan_all Job ?


judexzhu:
scan_all job stuck for days, how can I stop it

![image](https://user-images.githubusercontent.com/17834225/126207070-1774e4f3-fdc3-4e94-a2df-497c5b161f9c.png)

Check the db, the job stauts is running
 ```
registry=# select * from execution where vendor_type='SCAN_ALL';
  id  | vendor_type | vendor_id | status  | status_message | trigger  | extra_attrs |         start_time         |      end_time       | revision |     update_time
------+-------------+-----------+---------+----------------+----------+-------------+----------------------------+---------------------+----------+---------------------
 1410 | SCAN_ALL    |         0 | Running |                | MANUAL   | {}          | 2021-07-15 22:49:50.037368 |                     |     3723 | 2021-07-15 23:22:43
 1408 | SCAN_ALL    |         0 | Error   |                | MANUAL   | {}          | 2021-07-15 21:20:55.04013  | 2021-07-15 22:06:29 |     4217 | 2021-07-15 22:17:05
  673 | SCAN_ALL    |         0 | Running |                | SCHEDULE | {}          | 2021-06-26 00:00:04.216228 |                     |    54594 | 2021-07-01 00:50:47
 1409 | SCAN_ALL    |         0 | Success |                | MANUAL   | {}          | 2021-07-15 22:36:51.53495  | 2021-07-15 22:48:46 |     2020 | 2021-07-15 22:48:46
  986 | SCAN_ALL    |         0 | Running |                | SCHEDULE | {}          | 2021-07-04 08:00:00.793581 |                     |    45421 | 2021-07-05 11:55:27
 1398 | SCAN_ALL    |         0 | Success |                | MANUAL   | {}          | 2021-07-15 15:55:48.666091 | 2021-07-15 19:07:39 |    11840 | 2021-07-15 22:17:49
 1404 | SCAN_ALL    |         0 | Success |                | MANUAL   | {}          | 2021-07-15 19:31:22.934841 | 2021-07-15 21:10:07 |    11047 | 2021-07-15 22:17:49
(7 rows)
```
```
registry=# select * from task where vendor_type = 'SCAN_ALL' and status = 'Running';
   id   | execution_id |          job_id          | status  | status_code | status_revision | status_message | run_count |                                                                                                                extra_attrs
                                          |       creation_time        |     start_time      |     update_time     |      end_time       | vendor_type
[...]

Issue #14831, opened 2021-05-10
there is  no button to stopping scanning images in web page


svenlu:
Hi everyone,
how to stop harbor  scanning images in  web page ? scanning images cause redis memory raise suddenly。
kind regardrs

steven-zou:
@svenlu 

Please provide more background context such as how many repos are being scanned etc.?

svenlu:
> @svenlu
> 
> Please provide more background context such as how many repos are being scanned etc.?

![bc4fbf80-fa9f-40b2-ab92-59e0e44406a0](https://user-images.githubusercontent.com/5123378/117603130-ecb4bf80-b184-11eb-8f52-3624b9b858d8.png)

![企业微信截图_1620469471628](https://user-images.githubusercontent.com/5123378/117603147-f8a08180-b184-11eb-92e2-2a27392f4634.png)

steven-zou:
@heww 

Can we consider adding a stop function in the next release?

PR #15548, opened 2021-09-06
Add stop scan a particular artifact & stop scan all

zyyw:
Closes: #14831
Closes: #15338

### DIFF
--- a/src/controller/scan/base_controller.go
+++ b/src/controller/scan/base_controller.go
@@ -313,6 +313,24 @@ func (bc *basicController) Scan(ctx context.Context, artifact *ar.Artifact, opti
 	return nil
 }
 
+// Stop scan job of a given artifact
+func (bc *basicController) Stop(ctx context.Context, artifact *ar.Artifact) error {
+	if artifact == nil {
+		return errors.New("nil artifact to stop scan")
+	}
+	query := q.New(q.KeyWords{"extra_attrs.artifact.digest": artifact.Digest})
+	executions, err := bc.execMgr.List(ctx, query)
+	if err != nil {
+		return err
+	}
+	if len(executions) == 0 {
+		message := fmt.Sprintf("no scan job for artifact digest=%v", artifact.Digest)
+		return errors.BadRequestError(nil).WithMessage(message)
+	}
+	execution := executions[0]
+	return bc.execMgr.Stop(ctx, execution.ID)
+}
+
 func (bc *basicController) ScanAll(ctx context.Context, trigger string, async bool) (int64, error) {
 	executionID, err := bc.execMgr.Create(ctx, VendorTypeScanAll, 0, trigger)
 	if err != nil {
--- a/src/pkg/scan/job.go
+++ b/src/pkg/scan/job.go
@@ -146,6 +146,16 @@ func (j *Job) Run(ctx job.Context, params job.Parameters) error {
 	// Get logger
 	myLogger := ctx.GetLogger()
 
+	// shouldStop checks if the job should be stopped
+	shouldStop := func() bool {
+		if cmd, ok := ctx.OPCommand(); ok && cmd == job.StopCommand {
+			myLogger.Info("scan job being stopped")
+			return true
+		}
+
+		return false
+	}
+
 	// Ignore errors as they have been validated already
 	r, _ := extractRegistration(params)
 	req, _ := ExtractScanReq(params)
@@ -155,6 +165,10 @@ func (j *Job) Run(ctx job.Context, params job.Parameters) error {
 	printJSONParameter(JobParamRegistration, params[JobParamRegistration].(string), myLogger)
 	printJSONParameter(JobParameterRequest, removeAuthInfo(req), myLogger)
 	myLogger.Infof("Report mime types: %v\n", mimeTypes)
+
+	if shouldStop() {
+		return nil
+	}
 
 	// Submit scan request to the scanner adapter
 	client, err := r.Client(v1.DefaultClientPool)
@@ -180,6 +194,10 @@ func (j *Job) Run(ctx job.Context, params job.Parameters) error {
 	}
 	if err != nil {
 		logAndWrapError(myLogger, err, "scan job: make authorization")
+	}
+
+	if shouldStop() {
+		return nil
 	}
 
 	req.Registry.Authorization = authorization
@@ -210,6 +228,10 @@ func (j *Job) Run(ctx job.Context, params job.Parameters) error {
 			for {
 				select {
 				case t := <-tm.C:
+					if shouldStop() {
+						return
+					}
+
 					myLogger.Debugf("check scan report for mime %s at %s", m, t.Format("2006/01/02 15:04:05"))
 
 					rawReport, err := client.GetScanReport(resp.ID, m)
@@ -249,6 +271,10 @@ func (j *Job) Run(ctx job.Context, params job.Parameters) error {
 
 	// Wait for all the retrieving routines are completed
 	wg.Wait()
+
+	if shouldStop() {
+		return nil
+	}
 
 	// Merge errors
 	for _, e := range errs {
--- a/src/pkg/task/dao/execution.go
+++ b/src/pkg/task/dao/execution.go
@@ -329,11 +329,25 @@ func (e *executionDAO) querySetter(ctx context.Context, query *q.Query) (orm.Que
 				break
 			}
 		}
-		if len(keyPrefix) == 0 {
+		if len(keyPrefix) == 0 || keyPrefix == key {
 			return qs, nil
 		}
-		inClause, err := orm.CreateInClause(ctx, "select id from execution where extra_attrs->>?=?",
-			strings.TrimPrefix(key, keyPrefix), value)
+
+		// key with keyPrefix supports multi-level query operator on PostgreSQL JSON data
+		// examples:
+		// key = extra_attrs.id,
+		//  ==> sql = "select id from execution where extra_attrs->>?=?", args = {id, value}
+		// key = extra_attrs.artifact.digest
+		//  ==> sql = "select id from execution where extra_attrs->?->>?=?", args = {artifact, id, value}
+		// key = extra_attrs.a.b.c
+		//  ==> sql = "select id from execution where extra_attrs->?->?->>?=?", args = {a, b, c, value}
+		keys := strings.Split(strings.TrimPrefix(key, keyPrefix), ".")
+		var args []interface{}
+		for _, item := range keys {
+			args = append(args, item)
+		}
+		args = append(args, value)
+		inClause, err := orm.CreateInClause(ctx, buildInClauseSqlForExtraAttrs(keys), args...)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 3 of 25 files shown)


src/controller/scan/callback.go
```diff
@@ -64,6 +64,7 @@ func scanTaskStatusChange(ctx context.Context, taskID int64, status string) (err
 	logger := log.G(ctx).WithFields(log.Fields{"task_id": taskID, "status": status})
 
 	js := job.Status(status)
+	logger.Infof("scan task status changed, js: %+v", js)
```
heww, line 67: Not needed

zyyw: > Not needed

will remove it when pushing next commit

src/pkg/task/dao/execution.go
```diff
@@ -342,3 +349,22 @@ func (e *executionDAO) querySetter(ctx context.Context, query *q.Query) (orm.Que
 
 	return qs, nil
 }
+
+// Param keys is strings.Split() after trim "extra_attrs."/"ExtraAttrs." prefix
+func buildInClauseSqlForExtraAttrs(keys []string) string {
+	switch len(keys) {
+	case 0:
+		// won't fall into this case, as the if condition on "keyPrefix == key"
+		return fmt.Sprintf("select id from execution where extra_attrs=?")
```
heww, line 358: Remove these because it will never happen.

zyyw: > Remove these because it will never happen.
@heww 
we should make sure "default" case equivalent to "len(keys) >= 2",  therefore, this `case 0` should probably be kept.

heww: But the return string for case 0 is a bad result.

zyyw: > But the return string for case 0 is a bad result or unusable result.

so shall we `return ""` for `case 0:`

zyyw: because anyway `default` should be equivalent to `len(keys) >= 2`, I guess `case 0` should be listed/kept as a place holder.

heww: returns empty string is ok for me

src/controller/scan/base_controller.go
```diff
@@ -313,6 +313,24 @@ func (bc *basicController) Scan(ctx context.Context, artifact *ar.Artifact, opti
...
+	executions, err := bc.execMgr.List(ctx, query)
+	if err != nil {
+		return err
+	}
+	if len(executions) == 0 {
+		message := fmt.Sprintf("no scan job for artifact digest=%v", artifact.Digest)
+		return errors.BadRequestError(nil).WithMessage(message)
+	}
+	execution := executions[0]
+	return bc.execMgr.Stop(ctx, execution.ID)
```
wy65701436, line 331: the execMgr.stop actually will not raise the error on cancelling a task, we need to think about how to report the errors to end user.

src/server/v2.0/handler/scan.go
```diff
@@ -48,6 +48,24 @@ func (s *scanAPI) Prepare(ctx context.Context, operation string, params interfac
 	return nil
 }
 
+func (s *scanAPI) StopScanArtifact(ctx context.Context, params operation.StopScanArtifactParams) middleware.Responder {
+	if err := s.RequireProjectAccess(ctx, params.ProjectName, rbac.ActionDelete, rbac.ResourceScan); err != nil {
```
wy65701436, line 52: we define POST method in the API doc, but use `ActionDelete` to judeg permission?

src/server/v2.0/handler/scan_all.go
```diff
@@ -54,6 +54,28 @@ func (s *scanAllAPI) Prepare(ctx context.Context, operation string, params inter
 	return nil
 }
 
+// StopScanAll stops the execution of scan all artifacts.
+func (s *scanAllAPI) StopScanAll(ctx context.Context, params operation.StopScanAllParams) middleware.Responder {
+	if err := s.requireAccess(ctx, rbac.ActionDelete); err != nil {
```
wy65701436, line 59: ditto

src/pkg/task/dao/execution.go
```diff
@@ -329,11 +329,18 @@ func (e *executionDAO) querySetter(ctx context.Context, query *q.Query) (orm.Que
...
 		}
-		if len(keyPrefix) == 0 {
+		if len(keyPrefix) == 0 || keyPrefix == key {
 			return qs, nil
 		}
-		inClause, err := orm.CreateInClause(ctx, "select id from execution where extra_attrs->>?=?",
-			strings.TrimPrefix(key, keyPrefix), value)
+
+		// key with keyPrefix supports multi-level query operator on PostgreSQL JSON data
+		keys := strings.Split(strings.TrimPrefix(key, keyPrefix), ".")
```
wy65701436, line 337: give an example in the comments to better understand.

src/pkg/task/dao/execution.go
wy65701436: what's the sql result on the case 0 path? The executions with the extra_attrs is empty?

If you'd like to handle case 0 path, just add a if condition 

```
keys := strings.Split(strings.TrimPrefix(key, keyPrefix), ".")
if len(keys) == 0 {
    return qs, nil
}
```

src/server/v2.0/handler/scan_all.go
```diff
@@ -54,6 +54,28 @@ func (s *scanAllAPI) Prepare(ctx context.Context, operation string, params inter
...
+
+	execution, err := s.getLatestScanAllExecution(ctx)
+	if err != nil {
+		return s.SendError(ctx, err)
+	}
+	if execution == nil {
+		message := fmt.Sprintf("no scan all job is found currently")
+		return s.SendError(ctx, errors.BadRequestError(nil).WithMessage(message))
+	}
+	err = s.execMgr.Stop(ctx, execution.ID)
```
wy65701436, line 71: did you test the large amount artifacts scenario? If there are too many taskes to be cancelled, the API can easily time out.

If the http is timeout, the handler will be break? What about the un-stopped task, art they continue scanning in the backend? Is it possible to continue stop them?

src/pkg/scan/job.go
```diff
@@ -146,6 +146,15 @@ func (j *Job) Run(ctx job.Context, params job.Parameters) error {
 	// Get logger
 	myLogger := ctx.GetLogger()
 
+	// shouldStop checks if the job should be stopped
+	shouldStop := func() bool {
+		if cmd, ok := ctx.OPCommand(); ok && cmd == job.StopCommand {
+			return true
```
wy65701436, line 152: add a log to mention the job is cancelled?

src/pkg/scan/job.go
```diff
@@ -268,6 +289,10 @@ func (j *Job) Run(ctx job.Context, params job.Parameters) error {
 		return err
 	}
 
+	if shouldStop() {
```
wy65701436, line 292: we can move the check to line 252, after wg.wait()

wy65701436:
@zyyw thanks, wiil we add a API test for it?

cc @YangJiao0817

src/controller/scan/base_controller.go
zyyw: > the execMgr.stop actually will not raise the error on cancelling a task, we need to think about how to report the errors to end user.

Harbor currently raises an error when cancelling a task only if cancelling the last task (out of a list of tasks managed by execMgr) errors out.

src/server/v2.0/handler/scan.go
zyyw: > we define a POST method in the API doc, but use `ActionDelete` to judge permission?

I was initially using `ActionUpdate`, but later on reaching an agreement with @heww to use `ActionDelete` for permission checking. But we are open to further discussion on this.

src/server/v2.0/handler/scan_all.go
zyyw: > ditto

same as above

src/pkg/task/dao/execution.go
zyyw: > give an example in the comments for better understanding.

will add more examples for this.

src/pkg/scan/job.go
zyyw: > add a log to mention the job is cancelled?

we may not want to overwhelm the core.log with "stop scan job" message in the large scale scenario.

zyyw:
> @zyyw thanks, wiil we add a API test for it? And please squash before merging.
> 
> cc @YangJiao0817

sure, `git squash` will be done before merge the PR. 
And btw, API tests have already been added within this PR.

src/controller/scan/base_controller.go
wy65701436: if you trace the code, it eventlly can task.Stop() to stop scan. But task manager is just log the error instead of raising it, how does user know the failures from the API response?

```
for _, task := range tasks {
		if err = e.taskMgr.Stop(ctx, task.ID); err != nil {
			log.Errorf("failed to stop task %d: %v", task.ID, err)
			continue
		}
}
```

wy65701436:
> > @zyyw thanks, wiil we add a API test for it? And please squash before merging.
> > cc @YangJiao0817
> 
> sure, `git squash` will be done before merge the PR.
> And btw, API tests have already been added within this PR.

for API test case, I mean the python tests, see https://github.com/goharbor/harbor/tree/master/tests/apitests/python

zyyw:
> https://github.com/goharbor/harbor/tree/master/tests/apitests/python

aha, I thought it was UT like these two files submitted in this PR: 
* src/server/v2.0/handler/scan_test.go
* src/server/v2.0/handler/scan_all_test.go

src/controller/scan/base_controller.go
zyyw: > if you trace the code, it eventlly calles task.Stop() to stop scan. But task manager is just log the error instead of raising it, how does user know the failures from the API response?
> 
> ```
> for _, task := range tasks {
> 		if err = e.taskMgr.Stop(ctx, task.ID); err != nil {
> 			log.Errorf("failed to stop task %d: %v", task.ID, err)
> 			continue
> 		}
> }
> ```
 if it errors out when cancelling the last task, it will report this error,  there is a `return err` after this for loop. for those error out in the middle of a list of tasks, it will only log the error. whether reports error depends on cancelling the last task.

src/pkg/task/dao/execution.go
zyyw: the case 0 of `buildInClauseSqlForExtraAttrs()` is just a place holder to ensure `default` case equivalent to "len(keys) >= 2". it won't fall into this `case 0` because there is a if check  condition`keyPrefix == key`.

wy65701436 (approved):
lgtm

src/server/v2.0/handler/scan_all.go
```diff
@@ -54,6 +56,31 @@ func (s *scanAllAPI) Prepare(ctx context.Context, operation string, params inter
...
+	}
+	if execution == nil {
+		message := fmt.Sprintf("no scan all job is found currently")
+		return s.SendError(ctx, errors.BadRequestError(nil).WithMessage(message))
+	}
+	go func(ctx context.Context, eid int64) {
+		err = s.execMgr.Stop(ctx, eid)
+		if err != nil {
+			log.Errorf("failed to stop the execution of executionID=%+v", execution.ID)
+			return
```
heww, line 77: not needed

src/server/v2.0/handler/scan_all.go
```diff
@@ -54,6 +56,31 @@ func (s *scanAllAPI) Prepare(ctx context.Context, operation string, params inter
...
+	execution, err := s.getLatestScanAllExecution(ctx)
+	if err != nil {
+		return s.SendError(ctx, err)
+	}
+	if execution == nil {
+		message := fmt.Sprintf("no scan all job is found currently")
+		return s.SendError(ctx, errors.BadRequestError(nil).WithMessage(message))
+	}
+	go func(ctx context.Context, eid int64) {
+		err = s.execMgr.Stop(ctx, eid)
```
heww, line 74: `err = s.execMgr.Stop(ctx, eid)` => `err := s.execMgr.Stop(ctx, eid)` please don't overwrite the err out of the scope.

zyyw: nice catch. thanks

heww (approved):

PR #15626, opened 2021-09-21
[Cherrypick - v2.3] Make sure auto-onboard username claim has higher priority

### LINKED ISSUE #15504: Defined Username claim mapping broken
We are hosting a Harbor instance which is connected over OIDC to MS Azure AD.
Since the 2.3.2 we encountered an unexpected behavior with automatic on boarded users and their usernames.
As Username claims we're using the claim `preferred_username`. But instead of the value from the preferred_username claimin the ID Token, the name from the user info endpoint is set as the username.

In our case, only the local (ID Token) contains the defined claim `preferred_username` and is set correctly as the username. The remote info (UserInfo endpoint) doesn't contains this claim. Instead the user info response contains a field named `name`, which is used as username and won't be overwritten due the missing preferred_username claim
Afterwards the method mergeUserInfo merges the remote with local user info, but the remote info takes precedence. ([see here](https://github.com/goharbor/harbor/blob/master/src/pkg/oidc/helper.go#L286-L311)) 

Expected Behavior:
The local info parsed from the ID token should take precedence over the remote user info. Or at least if the defined claim exists in the ID token but is missing in the response from the user info endpoint, the username should not be overwritten.  

1. Configure Harbor auhentication with Azure AD
2. Set `preferred_username` as Username Claim
3. Login with non onboarded User
4. The username displayed in Harbor doesn't match with the preferred_username claim in the ID token

**Versions:**
- harbor version: 2.3.2

Here the commit which broke the previous/expected behavior.
https://github.com/goharbor/harbor/commit/0679f4701e67e0b76dfb39031f167e104bdbe1a4#diff-e300eb848c9bca554e794da0ec59f102d1cd730eaf1dbb3ccb7155807dcead2d

reasonerjt:
This commit adjust the priority when consolidating data from userinfo
endpoint and id token, making sure the auto-onboard username claim from
ID token has highest priority.

fixes #15504


heww (approved):

wy65701436 (approved):
lgtm

PR #15675, opened 2021-09-27
Allow empty email attribute for ldap/oidc user

### LINKED ISSUE #10400: Locked on with duplicate key value violates unique constraint "harbor_user_email_key"
Hi all,

May be I missed something, but I think I got a logical issue with Harbor/LDAP integration.

When removing an existing user in an ldap server, there is no way to remove its counterpart in Harbor.
Beside having 'zombi' users, this prevent to re-use its email address, due to unicity constraint.

- Have an Harbor instance with properly configured ldap connection.
- Create a user in ldap: For example uid=jclark, email=jim.clark@lotus.com
- Log in harbor interface with this user
- This user is now displayed in the user pane. Which is readonly, as authentication is configured with ldap.
- Remove this user in ldap server.
- Create another one with same email. For example: uid=clarkj, email=jim.clark@lotus.com
- Unable to log using this new user. Digging in core.log, we can see:

Jan  5 16:21:16 172.18.0.1 core[6020]: 2020-01-05T16:21:16Z [ERROR] [/core/controllers/base.go:109]: Error occurred in UserLogin: pq: duplicate key value violates unique constraint "harbor_user_email_key"

Same error when trying to use this new user as a project member.

And without a way to delete this old user (Excepting of course hacking the DB), this email adresse is unusable anymore..

Beside the ability to cleanup Zombi user, I am wondering why there is this unicity constraint on email value ? It is quite frequent to have several account using the same email (As they generally refer to the same humain being).

**Versions:**

- harbor version: 1.10.0
- docker-compose version:  1.18.0,

**Additional context:**

Tested against a Freeipa LDAP server. But I don't think this has something to do with ldap server type.

Thanks.

stonezdj:
Define user.Email as sql.NullString
  Fixes #10400

### DIFF
--- a/src/common/models/user.go
+++ b/src/common/models/user.go
@@ -15,58 +15,39 @@
 package models
 
 import (
-	"context"
 	"time"
-
-	"github.com/astaxie/beego/orm"
 )
-
-// UserTable is the name of table in DB that holds the user object
-const UserTable = "harbor_user"
 
 // User holds the details of a user.
 type User struct {
-	UserID          int    `orm:"pk;auto;column(user_id)" json:"user_id"`
-	Username        string `orm:"column(username)" json:"username" sort:"default"`
-	Email           string `orm:"column(email)" json:"email"`
-	Password        string `orm:"column(password)" json:"password"`
-	PasswordVersion string `orm:"column(password_version)" json:"password_version"`
-	Realname        string `orm:"column(realname)" json:"realname"`
-	Comment         string `orm:"column(comment)" json:"comment"`
-	Deleted         bool   `orm:"column(deleted)" json:"deleted"`
-	Rolename        string `orm:"-" json:"role_name"`
-	// if this field is named as "RoleID", beego orm can not map role_id
-	// to it.
-	Role         int  `orm:"-" json:"role_id"`
-	SysAdminFlag bool `orm:"column(sysadmin_flag)" json:"sysadmin_flag"`
+	UserID          int    `json:"user_id"`
+	Username        string `json:"username" sort:"default"`
+	Email           string `json:"email"`
+	Password        string `json:"password"`
+	PasswordVersion string `json:"password_version"`
+	Realname        string `json:"realname"`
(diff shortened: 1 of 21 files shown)


src/common/models/user.go
```diff
@@ -15,58 +15,39 @@
 package models
```
wy65701436, line 15: we should not keep model in common.

src/common/models/user.go
```diff
@@ -15,58 +15,39 @@
...
-
-// FilterByUsernameOrEmail generates the query setter to match username or email column to the same value
-func (u *User) FilterByUsernameOrEmail(ctx context.Context, qs orm.QuerySeter, key string, value interface{}) orm.QuerySeter {
-	usernameOrEmail, ok := value.(string)
-	if !ok {
-		return qs
-	}
-	subCond := orm.NewCondition()
-	subCond = subCond.Or("Username", usernameOrEmail).Or("Email", usernameOrEmail)
+type Users []*User
```
wy65701436, line 44: why the FilterByUsernameOrEmail is removed? It's a kind of reflection in orm model, it could be used at somewhere.

src/pkg/user/models/user.go
```diff
@@ -17,23 +17,81 @@ package models
...
+	return user
+}
 
-	return m
+// ToCommonUser ...
+func ToCommonUser(u *User) *commonmodels.User {
+	user := &commonmodels.User{}
+	user.UserID = u.UserID
+	user.Username = u.Username
+	user.Email = u.Email.String
```
wy65701436, line 81: do we need u.Emai.Valid()?

wy65701436 (approved):
lgtm

reasonerjt:
1.  Please add more details in the commit message to explain why this change has to be made
2. Please try to make the new `User` struct internal to `user/dao` package, if it's not possible, please add comments in the `User` struct in pkg/user/models/ to explcitly point out this is only used internally in package user/dao,

reasonerjt (approved):

Issue #15679, opened 2021-09-27
[PORTAL] Wrong filename at Helm Chart download [FIX SUGGESTION]


antbreton:
Hi, (please read this till the end)

we're facing a really anoying issue while downloading helm charts from **helm-chart versions view**.

If I click download on a selected version, the resulting filename is a generated ID instead of correct filename.

![image](https://user-images.githubusercontent.com/16175249/134907741-e00b4003-28e4-43aa-b2e8-1a78003288e3.png)

_NOTE: this is **not** reproduced on harbor **helm-chart list charts view** as you can see on the following screenshot._
![image](https://user-images.githubusercontent.com/16175249/134908586-8298ec03-335a-4f05-86e5-b718efb9d98e.png)

This issue occured on v1.10.7 but IMO the root cause has not changed since this version.

The bug is related to this [line](https://github.com/goharbor/harbor/blob/38e0910fc775a9c81e8846355bc722457d48357c/src/portal/src/app/base/project/helm-chart/helm-chart-detail/helm-chart.service.ts#L181)

Let me explain:
- sometimes filename can be `charts/${selectedChart.name}-${selectedChart.latest_version}.tgz` ==> relative URI (Working)
- sometimes filename comes from the backend [let filename = this.chartDetail.metadata.urls[0];](https://github.com/goharbor/harbor/blob/38e0910fc775a9c81e8846355bc722457d48357c/src/portal/src/app/base/project/helm-chart/helm-chart-detail/chart-detail/chart-detail.component.ts#L86) and contains full URL such as `https://<domain>/chartrepo/<project>/charts/<chart_name>-<version>.tgz`

This leads to unexpected behavious because filename.split('/')[1] **returns empty string** in the second case.
![image](https://user-images.githubusercontent.com/16175249/134913235-6dde1de5-78b5-4806-8553-aadde219878e.png)
Indeed, `split("\")[1]` extract the second part so it DOES NOT work with "https://...." 

I suggest you extract the last item of the split array. This will returns the filename whatever if you provide URI or full URL:
![image](https://user-images.githubusercontent.com/16175249/134914011-939fb9df-bbe3-4101-804e-4bede96fede8.png)

I would be glad to fix this by myself but I'm not part of the harbor dev community yet.

This is really anoying for our customers since they have to rename the downloaded file in order to extract it.

This is strongly related to #11901 and the proposal will fix it

PR #15680, opened 2021-09-27
fix filename processing to support both URI and URL

antbreton:
Fix proposal for #15679 cf https://github.com/goharbor/harbor/issues/15679#issue-1008136974


AllForNothing (approved):
LGTM

Issue #15623, opened 2021-09-21
Interrogation Service - Vuln response v1.0 overrides v1.1


skrech:
Hello,

I've noticed that if one Interrogation Service reports support for both of these mime types from its `/metadata` endpoint:
`application/vnd.scanner.adapter.vuln.report.harbor+json; version=1.0`
`application/vnd.security.vulnerability.report; version=1.1`

Then Harbor makes two requests - one for v1.0 and another one for 1.1. First, it seems redundant to request for v1.0 data when v1.1 data is a superset of the v1.0 data. Second, it seems that v1.0 takes precedence over v1.1 data, which means that even though CVSS scores are returned from the Interrogation Service, they don't appear in the UI.

**Versions:**
- Harbor version: 2.3
- Running in Kubernetes, default config

bitsf:
@danielpacak what do you suggest about this ?

danielpacak:
We allowed scanners to return multiple mime types for backward compatibility, especially to install newer scanners with older Harbor deployments. However, this is problematic as pointed out by @skrech. At the same time we wanted to introduce canonical names for the vulnerability report mime type, which is somehow independent of Harbor data model.

To fix this problem we could:
1. Require scanner adapters to return only one mime type that corresponds to vulnerability report displayed in the UI. Release notes should indicate that it breaks compatibility with previous versions of Harbor.
2. Patch Harbor service to skip fetching report with legacy mime type `application/vnd.scanner.adapter.vuln.report.harbor+json; version=1.0` if scanner declared also `application/vnd.security.vulnerability.report; version=1.1`.

To quick fix this problem for the default Trivy scanner I'd release new version of the scanner adapter that only declares `application/vnd.security.vulnerability.report; version=1.1` mime type and bump it up in Harbor Makefile. WDYT? @heww @prahaladdarkin @steven-zou

prahaladdarkin:
@danielpacak  IMHO, we should patch the Harbor implementation to pull only the latest  vulnerability MIME type if the scanner returns both the vulnerability MIME types (Option 2).
This is the only way to address any backward compatibility issues. @steven-zou  @heww  your pointers or alternate approaches are welcome.

heww:
Hi @skrech, if you want the report of version 1.1, please request the API with `application/vnd.security.vulnerability.report; version=1.1` for `X-Accept-Vulnerabilities` header.

heww:
```
[...]

PR #15701, opened 2021-09-28
fix(scanners): Vulnerability report v1.0 might override v1.1

danielpacak:
When both application/vnd.scanner.adapter.vuln.report.harbor+json; version=1.0
and application/vnd.security.vulnerability.report; version=1.1 mime types
are returned by a scanner we may end up with data inconsistency.
The root cause is that two reports pulled down by Harbor may override records
stored in the vulnerability_record table.

Resolves: #15623

### DIFF
--- a/api/v2.0/swagger.yaml
+++ b/api/v2.0/swagger.yaml
@@ -5389,7 +5389,7 @@ parameters:
     name: X-Accept-Vulnerabilities
     in: header
     type: string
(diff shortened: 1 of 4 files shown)


heww:
This PR is not needed because the trivy shipped in harbor 2.4 only supports one scan report version. 

And it's by design that harbor will generate and save all the scan reports supported by the scanner.

PR #15734, opened 2021-10-07
（cherry-pick) Allow empty email attribute for ldap/oidc user

### LINKED ISSUE #10400: Locked on with duplicate key value violates unique constraint "harbor_user_email_key"

stonezdj:
Define user.Email as sql.NullString to avoid unique constraint when email is empty in LDAP/OIDC
  Separate the common/models/User with the pkg/user/dao/User
  Fixes #10400


bitsf (approved):

wy65701436 (approved):
lgtm

PR #15747, opened 2021-10-10
Add function to avoid writing creds in jobservice logs, switch to debug instead of info

### LINKED ISSUE #15745: access_credentials field leaks in jobservice logs when trivy scan is launched by an user
Hello,

Using 2.3.3 (and 2.3.0 more specifically), it appears that the user who started a scan can have their credentials leaking in "access_credentials" field whereas AuthKey is marked as "[HIDDEN]".

It has an impact when using LDAP access and where Harbor does not manage access locally.

sixeela:
Closes #15745.

I added a function to handle log writing in jobservices and avoid retrieving password for auth such as Basic or more (used by LDAP for example).
I also switched logs to debug as it adds a lot more verbosity that intended in INFO.

Please keep me informed if modifications need to be done as I am not that familiar with Golang.


steven-zou:
@sixeIa 

Please make sure the pipeline is passed.

PR #15884, opened 2021-10-27
CSV Export Functionality for Scan data

prahaladdarkin:
Scope of change:
* Introduced a set of APIs to trigger CSV export of vulnerability data, track execution status and download scan data
* Ability to store the generated CSV reports as an OCI artifact
* Associated tests for the above functionality


heww:
Closed as a new PR #15998 is opened.

PR #15913, opened 2021-11-01
Cache content type for manifest list and image index in perspective

### LINKED ISSUE #15837: Cannot download images with oci media types via pull-through
Hello,

We are experiencing problems with pulling images via harbor pull-through proxy with `docker`.  
From the observations it does look like issues are occuring for images with these media types present in manifest list:
- `application/vnd.oci.image.manifest.v1+json`
- `application/vnd.oci.image.index.v1+json`  

Only as an example for describing situation image `eks-distro/kubernetes-csi/node-driver-registrar:v2.0.1-eks-1-18-1` stored in `public.ecr.aws` is used (similar behavior we observed with `ghcr.io`).  

We created project `publicecraws` in our local Harbor for ECR pull-through proxy.  
It is possible to pull image only once via Harbor ECR pull-through proxy - when it was not yet cached in Harbor.

Next pull and subsequent pulls are failing with such message:
```console
docker pull MY-LOCAL-HARBOR/publicecraws/eks-distro/kubernetes-csi/node-driver-registrar:v2.0.1-eks-1-18-1
Error response from daemon: mediaType in manifest list should be 'application/vnd.docker.distribution.manifest.list.v2+json' not 'application/vnd.oci.image.index.v1+json'
```

No problems were observed when pulling mentioned image directly from upstream.

With image `public.ecr.aws/cloudwatch-agent/cloudwatch-agent:1.247349.0b251399` there are no issues downloading it via Harbor ECR pull-through proxy.  
This image has these media types present in manifest list:
- `application/vnd.docker.distribution.manifest.list.v2+json`
- `application/vnd.docker.distribution.manifest.v2+json`  

We are using Harbor version: `v2.3.2`

Do you have any suggestions why problems are observed when pulling mentioned image via Harbor ECR pull-through proxy ?

stonezdj:
manifest list: application/vnd.docker.distribution.manifest.list.v2+json
  image index: application/vnd.oci.image.index.v1+json
  fixes #15837

### DIFF
--- a/src/controller/proxy/controller.go
+++ b/src/controller/proxy/controller.go
@@ -23,7 +23,6 @@ import (
 	"time"
 
 	"github.com/docker/distribution"
-	"github.com/docker/distribution/manifest/manifestlist"
 	"github.com/goharbor/harbor/src/controller/artifact"
 	"github.com/goharbor/harbor/src/controller/blob"
 	"github.com/goharbor/harbor/src/controller/event/operator"
@@ -138,6 +137,10 @@ type ManifestList struct {
 	ContentType string
 }
 
+// UseLocalManifest check if these manifest could be found in local registry,
+// the return error should be nil when it is not found in local and need to delegate to remote registry
+// the return error should be NotFoundError when it is not found in remote registry
+// the error will be captured by framework and return 404 to client
 func (c *controller) UseLocalManifest(ctx context.Context, art lib.ArtifactInfo, remote RemoteInterface) (bool, *ManifestList, error) {
 	a, err := c.local.GetManifest(ctx, art)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


src/controller/proxy/controller.go
```diff
@@ -161,26 +164,38 @@ func (c *controller) UseLocalManifest(ctx context.Context, art lib.ArtifactInfo,
...
-		} else {
-			log.Errorf("Failed to get manifest list from cache, error: %v", err)
-		}
+	var contentType string
+	if c.cache == nil {
+		return a != nil && string(desc.Digest) == a.Digest, nil, nil // digest matches
+	}
+
+	err = c.cache.Fetch(manifestListKey(art.Repository, string(desc.Digest)), &content)
+	if err == cache.ErrNotFound {
```
ywk253100, line 173: In the original logic, when `err == cache.ErrNotFound`, no error is returned, but now it returns an error because it also matches `if err != nil` code block

ywk253100 (approved):

wy65701436 (approved):
lgtm

src/controller/proxy/controller.go
```diff
@@ -161,26 +164,39 @@ func (c *controller) UseLocalManifest(ctx context.Context, art lib.ArtifactInfo,
...
+			log.Debugf("Digest is not found in manifest list cache, key=cache:%v", manifestListKey(art.Repository, string(desc.Digest)))
 		} else {
 			log.Errorf("Failed to get manifest list from cache, error: %v", err)
 		}
+		return a != nil && string(desc.Digest) == a.Digest, nil, nil
 	}
-	return a != nil && string(desc.Digest) == a.Digest, nil, nil // digest matches
+	err = c.cache.Fetch(manifestListContentTypeKey(art.Repository, string(desc.Digest)), &contentType)
+	if err != nil {
+		log.Debugf("failed to get the manifest list content type, not use local.")
```
wy65701436, line 183: log err here?

src/controller/proxy/controller.go
```diff
@@ -161,26 +164,39 @@ func (c *controller) UseLocalManifest(ctx context.Context, art lib.ArtifactInfo,
...
+
+	err = c.cache.Fetch(manifestListKey(art.Repository, string(desc.Digest)), &content)
+	if err != nil {
 		if err == cache.ErrNotFound {
-			log.Debugf("Digest is not found in manifest list cache, key=cache:%v", getManifestListKey(art.Repository, string(desc.Digest)))
+			log.Debugf("Digest is not found in manifest list cache, key=cache:%v", manifestListKey(art.Repository, string(desc.Digest)))
 		} else {
 			log.Errorf("Failed to get manifest list from cache, error: %v", err)
 		}
+		return a != nil && string(desc.Digest) == a.Digest, nil, nil
```
wy65701436, line 179: why not reture err?

src/controller/proxy/controller.go
wy65701436: and why not reture err?

stonezdj: return error will cause the error to return a non 200 http code to the client-side and pull fail, current situation, the error to access cache can be logged and ignored.

src/controller/proxy/controller.go
stonezdj: updated

Issue #15894, opened 2021-10-28
duplicate key value violates unique constraint "project_name_key" 


xammep4:
We use 2 harbor registries version 2.2.3.
Replication by label is configured, see the screenshot between them.

When performing replication, the database of the final registry fixes errors.

Analyzing the request for which the database registers errors, during replication, an attempt is made to create a replicated project in the final registry, but if there is already one, an error occurs:

duplicate key value violates unique constraint "project_name_key"

The request can also be seen in the screenshot

<img width="578" alt="Снимок экрана 2021-10-28 в 09 57 28" src="https://user-images.githubusercontent.com/10144377/139204200-eba6d03b-b2b1-4fbd-be08-521dfbdb3d3f.png">
<img width="1194" alt="Снимок экрана 2021-10-28 в 10 04 47" src="https://user-images.githubusercontent.com/10144377/139204238-497e0885-d831-4405-a9f4-813104f1588d.png">

PR #15934, opened 2021-11-03
fix(replication): list projects before replicate

chlins:
Fix: https://github.com/goharbor/harbor/issues/15894

### DIFF
--- a/src/pkg/reg/adapter/harbor/base/adapter.go
+++ b/src/pkg/reg/adapter/harbor/base/adapter.go
@@ -15,7 +15,7 @@
 package base
 
 import (
-	"errors"
+	"fmt"
 	"net/http"
 	"os"
 	"strconv"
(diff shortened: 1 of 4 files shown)


src/pkg/reg/adapter/harbor/base/adapter.go
```diff
@@ -169,7 +179,28 @@ func (a *Adapter) PrepareForPush(resources []*model.Resource) error {
 			Metadata: metadata,
 		}
 	}
-	for _, project := range projects {
+
+	// get exist projects
+	q := buildProjectsQuery(projects)
+	queryProjects, err := a.Client.ListProjectsWithQuery(q, false)
+	if err != nil {
+		return fmt.Errorf("list projects with query %s error: %w", q, err)
```
wy65701436, line 187: please use lib/errors for instead

src/pkg/reg/adapter/harbor/base/adapter.go
```diff
@@ -139,6 +140,15 @@ func (a *Adapter) Info() (*model.RegistryInfo, error) {
 	return info, nil
 }
 
+func buildProjectsQuery(projects map[string]*Project) string {
```
wy65701436, line 143: if it‘s not a util func, suggest using nested function.

src/pkg/reg/adapter/harbor/base/adapter.go
chlins: ok

src/pkg/reg/adapter/harbor/base/adapter.go
chlins: moved.

ywk253100 (approved):

wy65701436 (approved):
lgtm

PR #15935, opened 2021-11-03
Support right click to open a link in a new tab

### LINKED ISSUE #15893: Support right click to open a link in a new tab
Original issue #15436
Put it into target2.5.0

AllForNothing:
Fixes #15893

1. Enable  right-click to open in a new tab for "project links" and "repository links" (other links have extra logic, so can not enable it)

### DIFF
--- a/src/portal/src/index.html
+++ b/src/portal/src/index.html
@@ -4,7 +4,7 @@
 <head>
     <meta charset="utf-8">
     <title>Harbor</title>
-    <base href="/">
+    <base href="/test">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <link rel="icon" type="image/x-icon" href="favicon.ico?v=2">
     <link rel="preload" as="style" href="./light-theme.css">
--- a/src/portal/src/app/base/left-side-nav/projects/list-project/list-project.component.html
+++ b/src/portal/src/app/base/left-side-nav/projects/list-project/list-project.component.html
@@ -15,7 +15,7 @@
     <clr-dg-column [clrDgSortBy]="'creation_time'">{{'PROJECT.CREATION_TIME' | translate}}</clr-dg-column>
     <clr-dg-row *ngFor="let p of projects" [clrDgItem]="p">
         <clr-dg-cell>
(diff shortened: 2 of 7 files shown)


YangJiao0817 (approved):

wy65701436 (approved):
lgtm

src/portal/src/index.html
```diff
@@ -4,7 +4,7 @@
 <head>
     <meta charset="utf-8">
     <title>Harbor</title>
-    <base href="/">
+    <base href="/test">
```
lindhe, line 7: Why is this set to `/test` now?

AllForNothing: @lindhe Thanks for the info. I forgot to change it back。 It should be "/"

lindhe: That's alright, it's easy to make mistakes. Even the code review from @wy65701436 didn't catch it. Please refer to this PR when you submit the fix, so it's easy to find for people looking at this comment thread. :)

Pro tip: Instead of using `git add .` to add all changes, you can use `git add -p` which lets you inspect each hunk of patch individually before adding.

lindhe: And maybe there should be a regression test implemented to catch this in the future?

AllForNothing: @lindhe Of course, Considering adding a new nightly(e2e) case to cover this

lindhe: Thank you for fixing this so quickly. That's nice.

I've noticed that PR descriptions are often very short in this project. I think that if you spent some more time writing a description of what this change is supposed to do, and perhaps mention any surprising work-arounds made in the code, it would be easier for the code reviewer to know if the code looks like it should. Just a thought! :)

Again, thank you for fixing this. I appreciate seeing that Harbor is actively worked on.

lindhe:
@AllForNothing Which version of Harbor will ship with this fix? :) I'm really looking forward to it.

AllForNothing:
@lindhe The upcoming 2.5 release already contains  this fix

PR #15949, opened 2021-11-04
Enable db config cache explicitly

stonezdj:
previous config is not cached because it is initialized when cache is not ready


heww (approved):

wy65701436 (approved):
lgtm

Issue #15973, opened 2021-11-08
Replication rule user interaction should be more friendly


wy65701436:
I got 400 after I improper configurate the replication rule, but the error msg doesn't help.
![image](https://user-images.githubusercontent.com/2841473/140769217-5806354e-ead9-4cec-a375-770ea6733e17.png)

The root cause is that there is a white space between the repository, it works after I remove it.
![image](https://user-images.githubusercontent.com/2841473/140769576-80e4ec9a-2c45-45ab-a09c-efcf24c04f76.png)

And, I didn't get useful error msg from both source & target Harbor core/proxy log. 
It should be enhanced.

PR #15985, opened 2021-11-10
fix(replication): enhance the replication rule validation

chlins:
Fix: https://github.com/goharbor/harbor/issues/15973


src/pkg/reg/adapter/harbor/base/adapter.go
```diff
@@ -224,6 +224,10 @@ func (a *Adapter) ListProjects(filters []*model.Filter) ([]*Project, error) {
 		names, ok := util.IsSpecificPathComponent(projectPattern)
 		if ok {
 			for _, name := range names {
+				if strings.Contains(name, " ") {
```
wy65701436, line 227: or can we use strings.TrimSpace() to ensure name? And let user to config white space in UI? Just a suggestion.

ywk253100 (approved):

wy65701436 (approved):
lgtm

Issue #15988, opened 2021-11-10
Default robot permissions should be all false


dioguerra:
Security wise, it is best to have the robot permissions
all set to false and if the users wish to enable a
certain permission to expressely do it. The other way
around loses the permission policy enforcement from
lazy/unknowing users.

AllForNothing:
@xaleeks It makes sense, but it doesn’t seem to be very important， what do you think?

ChristianCiach:
I second that request, but for ergonomical reasons.

Almost all of our robot-accounts should only be allowed to `PULL` - and nothing else. Currently I have to remove *a lot* of permissions when creating new robot accounts, which is very prone to error.

AllForNothing:
@ChristianCiach For your use case，It's surely more comfortable for you if all the permissions are unselected default.
But we are not sure if this is a very common case.
In version 2.5, we have added "select all" and "unselect all" operations for robot permissions， see #15774.
It can meet the requirement to a certain extent

dioguerra:
What is the decision here? This does not make sense?

PR #15962, opened 2021-11-08
Add select-all and unselect-all for robot permissions

AllForNothing:
Fixes #15774
image
image

Signed-off-by: AllForNothing <sshijun@vmware.com>

PR #15989, opened 2021-11-10
set default permissions to false

dioguerra:
Security wise, it is best to have the robot permissions
all set to false and if the users wish to enable a
certain permission to expressely do it. The other way
around loses the permission policy enforcement from
lazy/unknowing users.

Fixes: #15988

Signed-off-by: Diogo Guerra <diogo.filipe.tomas.guerra@cern.ch>

### DIFF
--- a/src/portal/src/app/base/left-side-nav/system-robot-accounts/system-robot-util.ts
+++ b/src/portal/src/app/base/left-side-nav/system-robot-accounts/system-robot-util.ts
@@ -50,82 +50,82 @@ export const INITIAL_ACCESSES: FrontAccess[] = [
     {
         "resource": "repository",
         "action": "push",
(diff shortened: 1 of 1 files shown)


dioguerra:
@xaleeks can we add this to 2.5?

AllForNothing:
@dioguerra Thanks for your contribution.
We decided to close it. By #15962, now we have "select all" and "unselect all" for the permissions

dioguerra:
I would still like to reiterate that the select/unselect all does not address my concern.

Also, having this button can be used the other way. But instead the user would be consiously activating all the permissions...

wy65701436:
Would you like to draft an discussion? We could collect more thoughts.

PR #15998, opened 2021-11-12
Vulnerability scan data export functionality

prahaladdarkin:
Scope of changes:
* Implement mechanism to export vulnerability scan data to CSV formatted file.
* Implement a Job - SCAN_DATA_EXPORT_JOB that would perform the export in an asynchronous manner.
* Implement mechanism to push the generated CSV artifact to the registry.
* periodic clean up logic to delete older CSV files.
* Expose set of APIs to:
   * Trigger an export job
   * Get the status of an export job execution
   * Download the CSV artifact.
 
Proposal Link : https://github.com/goharbor/community/pull/174


api/v2.0/swagger.yaml
```diff
@@ -8657,3 +8756,71 @@ definitions:
...
+
+  ScanDataExportJob:
+    type: object
+    description: The metadata associated with the scan data export job
+    properties:
+      id:
+        type: integer
+        format: int64
+        description: The id of the scan data export job
+  ScanDataExportExecution:
```
AllForNothing, line 8799: Need to  add  a "name" property to this object, for a better UX

api/v2.0/swagger.yaml
```diff
@@ -8657,3 +8756,71 @@ definitions:
         format: int64
         description: The total storage consumption of blobs, only be seen by the system admin
         x-omitempty: false
+  ScanDataExportCriteria:
```
AllForNothing, line 8759: Need to add a "name" property to this object, for a better UX

api/v2.0/swagger.yaml
```diff
@@ -8657,3 +8756,71 @@ definitions:
...
+        type: array
+        items:
+          type: integer
+          format: int64
+        description: A list of repositories for which to export the scan data, defaults to all if empty
+      cveIds:
+        type: array
+        items:
+          type: string
+      tags:
```
AllForNothing, line 8785: Should  use  double star rule  for "repos" and "tags".
And for "cveIds", need to discuss

api/v2.0/swagger.yaml
```diff
@@ -5334,6 +5334,99 @@ paths:
...
+          $ref: '#/responses/403'
+        '404':
+          $ref: '#/responses/404'
+        '405':
+          $ref: '#/responses/405'
+        '409':
+          $ref: '#/responses/409'
+        '500':
+          $ref: '#/responses/500'
+  /export/scan/execution/{execution_id}:
```
AllForNothing, line 5372: Should change it  to a  list API, "/export/scan/executions", to list the executions for the current user

AllForNothing: And  the response of this API needs to carry the job name info(need to discuss how to implement)

dioguerra:
Any chance i can help review this?

api/v2.0/swagger.yaml
```diff
@@ -9016,3 +9144,89 @@ definitions:
         type: string
         format: date-time
         description: The creation time of the accessory
+
+  ScanDataExportRequest:
+    type: object
+    description: The criteria to select the scan data to export.
+    properties:
+      job_name:
```
wy65701436, line 9152: is it necessary? User might not know it.

src/server/v2.0/handler/scanexport.go
```diff
@@ -0,0 +1,202 @@
...
+		return operation.NewExportScanDataBadRequest().WithPayload(errors)
+	}
+
+	// loop through the list of projects and validate that scan privilege and create privilege
+	// is available for all projects
+	// TODO : Should we just ignore projects that do not have the required level of access?
+
+	projects := params.Criteria.Projects
+	for _, project := range projects {
+		if err := se.RequireProjectAccess(ctx, project, rbac.ActionCreate, rbac.ResourceScan); err != nil {
```
wy65701436, line 60: is it a system level API? If so, only system access needs to be verified.

api/v2.0/swagger.yaml
```diff
@@ -5575,6 +5575,122 @@ paths:
...
+        '400':
+          $ref: '#/responses/400'
+        '401':
+          $ref: '#/responses/401'
+        '403':
+          $ref: '#/responses/403'
+        '404':
+          $ref: '#/responses/404'
+        '405':
+          $ref: '#/responses/405'
```
wy65701436, line 5608: can you point where the 405 will be raised?

src/server/v2.0/handler/scanexport.go
```diff
@@ -0,0 +1,202 @@
...
+			return se.SendError(ctx, err)
+		}
+	}
+
+	scanDataExportJob := new(models.ScanDataExportJob)
+
+	// vendor id associated with the execution can be set to -1
+	// is not associated with the state of any other scan report entity
+	userContext := context.WithValue(ctx, "vendorId", -1)
```
wy65701436, line 69: hardcode it in the scanDataExportCtl?

src/controller/scandataexport/execution.go
```diff
@@ -0,0 +1,191 @@
...
+	makeCtx        func() context.Context
+	sysArtifactMgr systemartifact.Manager
+}
+
+func (c *controller) GetExecutions(ctx context.Context, userName string) ([]*export.Execution, error) {
+	keywords := make(map[string]interface{})
+	keywords["VendorType"] = job.ScanDataExport
+	keywords[fmt.Sprintf("ExtraAttrs.%s", export.UserNameAttribute)] = userName
+
+	q := new(q2.Query)
```
wy65701436, line 55: use the q2.New()

api/v2.0/swagger.yaml
AllForNothing: This is a pure UI requirement that helps users to distinguish between exporting jobs

src/controller/scandataexport/execution.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	execs := make([]*export.Execution, 0)
+	for _, execForUser := range execsForUser {
+		execs = append(execs, c.convertToExportExecStatus(ctx, execForUser))
+	}
+	return execs, nil
+}
+
+func (c *controller) GetTask(ctx context.Context, executionId int64) (*task.Task, error) {
+	query := new(q2.Query)
```
wy65701436, line 69: ditto

src/controller/scandataexport/execution.go
```diff
@@ -0,0 +1,191 @@
...
+	tasks, err := c.taskMgr.List(ctx, query)
+	if err != nil {
+		return nil, err
+	}
+	if len(tasks) == 0 {
+		return nil, errors.Errorf("No task found for execution Id : %d", executionId)
+	}
+	// for the export JOB there would be a single instance of the task corresponding to the execution
+	// we will hence return the latest instance of the task associated with this execution
+	logger.Infof("Returning task instance with ID : %v", tasks[0].ID)
```
chlins, line 88: %d

src/jobservice/runtime/bootstrap.go
```diff
@@ -17,6 +17,7 @@ package runtime
 import (
 	"context"
 	"fmt"
+	"github.com/goharbor/harbor/src/jobservice/job/impl/scandataexport"
```
chlins, line 20: pls format imports

src/pkg/scan/export/filter_processor.go
```diff
@@ -0,0 +1,127 @@
...
+	repoMgr    repository.Manager
+	tagMgr     tag.Manager
+	usrMgr     user.Manager
+	projectMgr project.Manager
+}
+
+// NewFilterProcessor constructs an instance of a FilterProcessor
+func NewFilterProcessor() FilterProcessor {
+	return &DefaultFilterProcessor{
+		repoMgr:    repository.New(),
```
chlins, line 34: pls use `pkg.RepositoryMgr`

src/pkg/scan/export/filter_processor.go
```diff
@@ -0,0 +1,127 @@
...
+		tagMgr:     tag.Mgr,
+		usrMgr:     user.Mgr,
+		projectMgr: project.New(),
```
chlins, line 37: `pkg.ProjectMgr`

src/testing/controller/scan/export/controller.go
```diff
@@ -0,0 +1,78 @@
+package export
```
chlins, line 1: This file seems not generated by mockery?

src/testing/controller/scan/export/registryclient.go
```diff
@@ -0,0 +1,71 @@
+package export
```
chlins, line 1: ditto

src/jobservice/job/impl/scandataexport/scan_data_export.go
```diff
@@ -0,0 +1,345 @@
...
+
+func (sde *ScanDataExport) MaxFails() uint {
+	return 1
+}
+
+// MaxCurrency of the job. Unlike the WorkerPool concurrency, it controls the limit on the number jobs of that type
+// that can be active at one time by within a single redis instance.
+// The default value is 0, which means "no limit on job concurrency".
+func (sde *ScanDataExport) MaxCurrency() uint {
+	return 0
```
wy65701436, line 44: we should set it to 1

src/go.mod
```diff
@@ -28,6 +28,7 @@ require (
 	github.com/go-openapi/validate v0.19.10
 	github.com/go-redis/redis/v8 v8.11.4
 	github.com/go-sql-driver/mysql v1.5.0
+	github.com/gocarina/gocsv v0.0.0-20210516172204-ca9e8a8ddea8
```
wy65701436, line 31: why don't we choose the "encoding/csv"?

api/v2.0/swagger.yaml
```diff
@@ -9016,3 +9144,89 @@ definitions:
...
+          type: integer
+          format: int64
+        description: A list of one or more labels for which to export the scan data, defaults to all if empty
+      repositories:
+        type: string
+        description: A list of repositories for which to export the scan data, defaults to all if empty
+      cveIds:
+        type: string
+        description: CVE-IDs for which to export data. Multiple CVE-IDs can be specified by separating using ',' and enclosed between '{}'. Defaults to all if empty
+      tags:
```
wy65701436, line 9173: do we need the tags here? Since the scan result is associated with the digest, rather than tag. Repository is enough, IMO.

src/jobservice/job/impl/scandataexport/scan_data_export.go
```diff
@@ -0,0 +1,345 @@
...
+	csvFile, err := os.OpenFile(fileName, os.O_RDONLY, os.ModePerm)
+	if err != nil {
+		logger.Errorf(
+			"Export Job Id = %v. Error when moving report file %s to persistent storage: %v", params["JobId"], fileName, err)
+		return err
+	}
+	baseFileName := filepath.Base(fileName)
+	repositoryName := strings.TrimSuffix(baseFileName, filepath.Ext(baseFileName))
+	logger.Infof("Creating repository for CSV file with blob : %s", repositoryName)
+	stat, err := os.Stat(fileName)
```
wy65701436, line 106: what about if the size is 0?

src/server/v2.0/handler/scanexport.go
```diff
@@ -0,0 +1,202 @@
...
+		JobName:     execution.JobName,
+		UserName:    execution.UserName,
+		FilePresent: execution.FilePresent,
+	}
+
+	return operation.NewGetScanDataExportExecutionOK().WithPayload(&sdeExec)
+}
+
+func (se *scanDataExportAPI) DownloadScanData(ctx context.Context, params operation.DownloadScanDataParams) middleware.Responder {
+	err := se.RequireAuthenticated(ctx)
```
wy65701436, line 109: does that mean anyone can download CVE data of any exeuction?

src/server/v2.0/handler/scanexport.go
```diff
@@ -0,0 +1,202 @@
...
+
+	repositoryName := fmt.Sprintf("scandata_export_%v", params.ExecutionID)
+	file, err := se.sysArtifactMgr.Read(ctx, strings.ToLower(export.Vendor), repositoryName, execution.ExportDataDigest)
+	if err != nil {
+		return se.SendError(ctx, err)
+	}
+	logger.Infof("reading data from file : %s", repositoryName)
+
+	return middleware.ResponderFunc(func(writer http.ResponseWriter, producer runtime.Producer) {
+		defer se.cleanUpArtifact(ctx, repositoryName, execution.ExportDataDigest, params.ExecutionID)
```
wy65701436, line 126: combind these two defer?

src/server/v2.0/handler/scanexport.go
```diff
@@ -0,0 +1,202 @@
...
+		defer file.Close()
+		writer.Header().Set("Content-Type", "text/csv")
+		writer.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=%q", fmt.Sprintf("%s.csv", repositoryName)))
+		nbytes, err := io.Copy(writer, file)
+		if err != nil {
+			logger.Errorf("Encountered error while copying data: %v", err)
+		} else {
+			logger.Infof("Copied %v bytes from file to client", nbytes)
```
wy65701436, line 134: use debugf to avoid noise?

src/controller/scandataexport/execution.go
```diff
@@ -0,0 +1,191 @@
...
+	if err := c.execMgr.StopAndWait(ctx, executionID, 10*time.Second); err != nil {
+		logger.Errorf("failed to stop the execution %d: %v", executionID, err)
+	}
+	if err := c.execMgr.MarkError(ctx, executionID, err.Error()); err != nil {
+		logger.Errorf("failed to mark error for the execution %d: %v", executionID, err)
+	}
+}
+
+func (c *controller) convertToExportExecStatus(ctx context.Context, exec *task.Execution) *export.Execution {
+	execStatus := &export.Execution{
```
wy65701436, line 161: check nil firstly

src/pkg/scan/export/manager.go
```diff
@@ -0,0 +1,230 @@
...
+	"github.com/goharbor/harbor/src/jobservice/logger"
+	"github.com/goharbor/harbor/src/lib/orm"
+	q2 "github.com/goharbor/harbor/src/lib/q"
+	"strconv"
+	"strings"
+)
+
+const (
+	VulnScanReportView  = "vuln_scan_report"
+	VulnScanReportQuery = `select row_number() over() as result_row_id, project.project_id as project_id, project."name" as project_name, harbor_user.user_id as user_id, harbor_user.username as project_owner, repository.repository_id, repository.name as repository_name, artifact.digest as artifact_digest, scanner_registration.id as scanner_id, scanner_registration."name" as scanner_name, vulnerability_record.cve_id, vulnerability_record.package, vulnerability_record.severity,
```
wy65701436, line 18: as for the CSV column, 

- the repository name is enough.

- no need the project owner.

src/server/v2.0/handler/scanexport.go
```diff
@@ -0,0 +1,202 @@
...
+		UserID:      execution.UserID,
+		JobName:     execution.JobName,
+		UserName:    execution.UserName,
+		FilePresent: execution.FilePresent,
+	}
+
+	return operation.NewGetScanDataExportExecutionOK().WithPayload(&sdeExec)
+}
+
+func (se *scanDataExportAPI) DownloadScanData(ctx context.Context, params operation.DownloadScanDataParams) middleware.Responder {
```
wy65701436, line 108: can someone else download my exported data?

src/controller/scandataexport/execution.go
```diff
@@ -0,0 +1,191 @@
...
+	task.SetExecutionSweeperCount(job.ScanDataExport, 50)
+}
+
+var Ctl = NewController()
+
+type Controller interface {
+	Start(ctx context.Context, criteria export.Request) (executionId int64, err error)
+	GetExecution(ctx context.Context, executionId int64) (*export.Execution, error)
+	GetExecutions(ctx context.Context, userName string) ([]*export.Execution, error)
+	GetTask(ctx context.Context, executionId int64) (*task.Task, error)
```
chlins, line 29: one execution only for one task? for other resource, it often use ListTasks().

api/v2.0/swagger.yaml
```diff
@@ -5575,6 +5575,122 @@ paths:
         '500':
           $ref: '#/responses/500'
 
+  /export/scan:
```
wy65701436, line 5578: `/export/cve?`

src/controller/scandataexport/execution.go
```diff
@@ -0,0 +1,191 @@
...
+		return nil, err
+	}
+
+	return c.convertToExportExecStatus(ctx, exec), nil
+}
+
+func (c *controller) DeleteExecution(ctx context.Context, executionId int64) error {
+	err := c.execMgr.Delete(ctx, executionId)
+	if err != nil {
+		logger.Errorf("Error when deleting execution  for ExecutionId: %d error : %v", executionId, err)
```
chlins, line 105: "Error when deleting execution for ExecutionId: %d, error: %v"

src/jobservice/job/impl/scandataexport/scan_data_export.go
```diff
@@ -0,0 +1,345 @@
...
+	if err != nil {
+		logger.Errorf("Export Job Id = %v. Error when fetching execution record for update : %v", params["JobId"], err)
+		return err
+	}
+	attrsToUpdate := make(map[string]interface{})
+	for k, v := range exec.ExtraAttrs {
+		attrsToUpdate[k] = v
+	}
+	attrsToUpdate[export.DigestKey] = hash.String()
+	err = sde.execMgr.UpdateExtraAttrs(ctx.SystemContext(), execId, attrsToUpdate)
```
chlins, line 143: return sde.execMgr.UpdateExtraAttrs(ctx.SystemContext(), execId, attrsToUpdate)

src/jobservice/job/impl/scandataexport/scan_data_export.go
```diff
@@ -0,0 +1,345 @@
...
+	return err
+}
+
+func (sde *ScanDataExport) writeCsvFile(ctx job.Context, params job.Parameters, fileName string) error {
+	csvFile, err := os.OpenFile(fileName, os.O_WRONLY|os.O_CREATE|os.O_APPEND, os.ModePerm)
+	systemContext := ctx.SystemContext()
+	defer csvFile.Close()
```
chlins, line 150: csvFile whether be nil if open file error?

src/jobservice/job/impl/scandataexport/scan_data_export.go
```diff
@@ -0,0 +1,345 @@
...
+		}
+
+		exportParams = export.Params{
+			Projects:     filterCriteria.Projects,
+			Repositories: repoIds,
+			CVEIds:       filterCriteria.CVEIds,
+			Labels:       filterCriteria.Labels,
+		}
+	} else {
+		exportParams = export.Params{
```
chlins, line 209: do we need else?

src/pkg/scan/export/filter_processor.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	logger.Infof("Selected %d projects administered by user %s ", len(projects), userName)
+	projectIds := make([]int64, 0)
+	for _, proj := range projects {
+		projectIds = append(projectIds, proj.ProjectID)
+	}
+
+	// check if the project ids specified in the filter are present in the list
+	// of projects of which the current user is a project admin
+	if projectIdsToFilter == nil || len(projectIdsToFilter) == 0 {
```
chlins, line 67: `len(projectIdsToFilter) == 0` is enough.

src/pkg/scan/export/filter_processor.go
```diff
@@ -0,0 +1,127 @@
...
+		return projectIds, nil
+	}
+	m := make(map[int64]bool)
+	for _, projectId := range projectIds {
+		m[projectId] = true
+	}
+	filtered := make([]int64, 0)
+
+	for _, filteredProjId := range projectIdsToFilter {
+		if _, ok := m[filteredProjId]; ok {
```
chlins, line 77: as the map value is boolean type, so here can just use `if m[filteredProjId] {...}`

src/pkg/scan/export/filter_processor.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	return filtered, nil
+}
+
+func (dfp *DefaultFilterProcessor) ProcessRepositoryFilter(ctx context.Context, filter string, projectIds []int64) ([]*selector.Candidate, error) {
+	sel := doublestar.New(doublestar.RepoMatches, filter, "")
+	candidates := make([]*selector.Candidate, 0)
+
+	for _, projectId := range projectIds {
+		query := q.Query{Keywords: map[string]interface{}{"ProjectID": projectId}}
```
chlins, line 89: q.New()

src/pkg/scan/export/filter_processor.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	return sel.Select(candidates)
+}
+
+func (dfp *DefaultFilterProcessor) ProcessTagFilter(ctx context.Context, filter string, repositoryIds []int64) ([]*selector.Candidate, error) {
+	sel := doublestar.New(doublestar.Matches, filter, "")
+	candidates := make([]*selector.Candidate, 0)
+
+	for _, repoId := range repositoryIds {
+		query := q.Query{Keywords: map[string]interface{}{"RepositoryID": repoId}}
```
chlins, line 111: q.New()

src/server/v2.0/handler/scanexport.go
chlins: If the vendorId is -1, the execution will only be retained as 50 for all users, so we could use userId as the vendorId to let every user can keep self 50 records. Ref: https://github.com/goharbor/harbor/blob/dc1a03fde25cc05f82705e13106ba9cfae8fc918/src/pkg/task/execution.go#L146

src/server/v2.0/handler/scanexport.go
prahaladdarkin: As discussed, there is no change required here. Skipping this comment.

src/server/v2.0/handler/scanexport.go
prahaladdarkin: @chlins  Based on my earlier discussions, we had decided to keep the vendor id as -1. Is there a library method to get the user id for the current requesting user irrespective of the auth mechanism?

src/controller/scandataexport/execution.go
prahaladdarkin: Done

src/server/v2.0/handler/scanexport.go
prahaladdarkin: @wy65701436  I will hold on to this comment. If we want to pass the user id as the vendor id then we will have to do it within this location itself. I wanted to understand the possible implications of using  -1 other than the fact that executions would be retained across all users as opposed to specific user ids.

make/photon/prepare/templates/docker_compose/docker-compose.yml.jinja
```diff
@@ -21,6 +21,7 @@ services:
         target: /etc/rsyslog.d/rsyslog_docker.conf
     ports:
       - 127.0.0.1:1514:10514
+      - ::1:1514:10514
```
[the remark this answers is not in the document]
prahaladdarkin: Removed. This line was also the reason why my dev setups were not coming up correctly.

src/controller/scandataexport/execution.go
prahaladdarkin: Done.

src/pkg/scan/export/filter_processor.go
prahaladdarkin: Done

src/pkg/scan/export/filter_processor.go
prahaladdarkin: Done

src/testing/controller/scan/export/controller.go
prahaladdarkin: Done. Cleaned up

src/testing/controller/scan/export/registryclient.go
prahaladdarkin: Done. cleaned up

src/jobservice/job/impl/scandataexport/scan_data_export.go
prahaladdarkin: Done

src/server/v2.0/handler/scanexport.go
wy65701436: I think why @chlins purpose to use the userid as the vendor id is to avoid the execute sweeper to sweep the valid exeuctions.

Some backgound: 
To avoid having too many execution records in the database, Harbor introduced an execution sweeper to clean up what we consider to be useless execution records. When a user triggers a new exeuction, the sweeper starts sweeping the execution records. The key point here is that the sweeper uses the vendor ID and vendor type as query criteria to query the exeuction records and keep the latest 50. This means that if we use -1 as the vendor ID for CSV export execution, the sweeper will keep the latest 50 reports for all users. However, if we have hunderds of users and at least one CVE report per person, then 50 entries are not enough here.

src/go.mod
prahaladdarkin: Deliberately chose the current `gocarina/gocsv` library over the `encoding/csv` since it sis supports binding CSV fields to the struct fields during definition phase itself. For reference please take a look at how the `export.Data` struct is defined with the CSV annotations. This is not possible in `encoding/csv`  resulting in more verbose code

src/jobservice/job/impl/scandataexport/scan_data_export.go
prahaladdarkin: Currently, an empty file gets generated. However in my opinion this is ok. An empty CSV indicates to the user that his/her search criteria did not yield any results. This is also the behavior for many reporting tools that allow for offline report generation where-in the user triggering the report would download an empty CSV file if the report criteria is not satisfied.

src/server/v2.0/handler/scanexport.go
prahaladdarkin: Good catch. I missed this in my earlier implementation. Perhaps because the userID is set to -1, the above scenario pointed out is feasible. 
One fix would be to filter the executions based on the name of the user triggering the request and return only those execution records or download data that belong to the user
The other alternative is what requires your input - send the user id as the vendor id in which case, we return the executions filtered by the user id.

src/server/v2.0/handler/scanexport.go
prahaladdarkin: Done

src/server/v2.0/handler/scanexport.go
prahaladdarkin: Done

src/pkg/scan/export/manager.go
prahaladdarkin: The reason that the project owner was included was because when an admin runs a report, they should be able to immediately filter the obtained CSV based on certain fields of which project owner would be an important one. For e.g. the current implementation facilitates a workflow where the admin downloads the CSV for all projects and then performs targeted analysis in Excel for  specific set of project owners.

src/server/v2.0/handler/scanexport.go
prahaladdarkin: Currently yes. I will fix this post the discussion on the approach to do it.

src/controller/scandataexport/execution.go
prahaladdarkin: When we had originally discussed the proposal, as far as I remember correctly it was required to have only one execution per task since the CSV API and Job framework was supposed to behave that way. 
Also did you imply to rename the method `GetExecutions` to `ListExecutions`?

prahaladdarkin: When we had originally discussed the proposal, as far as I remember correctly it was required to have only one execution per task since the CSV API and Job framework was supposed to behave that way. 
Also did you imply to rename the method `GetExecutions` to `ListExecutions`?

wy65701436:
@daixiang0 feel free to add your comments.

daixiang0:
> @daixiang0 feel free to add your comments.

@wy65701436 unexpected ping?

wy65701436:
@dioguerra feel free to add your comments. 

My bad, sorry @daixiang0

src/pkg/scan/export/manager.go
prahaladdarkin: As discussed in a meeting, I will be currently keeping the project owner column as is and then we can decide the exact columns to be placed in the CSV based on the performance benchmarks.

src/jobservice/job/impl/scandataexport/scan_data_export.go
prahaladdarkin: As discussed in the meeting, I will keep the behavior as is. Also, the presence or absence of data within the file does not really add much load to the system since the user anyways has to download the file to examine the data. and hence does not really save anything in terms of I/O.

src/server/v2.0/handler/scanexport.go
prahaladdarkin: Done. I have implemented a check against the name of the user initiating the download versus the name of the user who owns the task execution and return a 401 if they do not match

src/controller/scandataexport/execution.go
prahaladdarkin: Done

src/jobservice/job/impl/scandataexport/scan_data_export.go
prahaladdarkin: Done

src/pkg/scan/export/filter_processor.go
prahaladdarkin: Done

src/pkg/scan/export/filter_processor.go
prahaladdarkin: Done

src/jobservice/job/impl/scandataexport/scan_data_export.go
prahaladdarkin: Done. removed

src/pkg/scan/export/filter_processor.go
prahaladdarkin: Done

src/server/v2.0/handler/scanexport.go
prahaladdarkin: This has been fixed. A check has been incorporated that compares the name of the logged in user with the owner of the export job

src/controller/scandataexport/execution.go
prahaladdarkin: Currently keep the names as is since these are internal APIs and not directly exposed to the end user via the Rest API.

src/server/v2.0/handler/scanexport.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	// check if the execution being downloaded is owned by the current user
+	secContext, err := se.GetSecurityContext(ctx)
+	if err != nil {
+		return se.SendError(ctx, err)
+	}
+
+	if secContext.GetUsername() != execution.UserName {
+		return middleware.ResponderFunc(func(writer http.ResponseWriter, producer runtime.Producer) {
```
wy65701436, line 145: here we can just return an forbidden error?

wy65701436 (approved):
lgtm

src/controller/scandataexport/execution.go
```diff
@@ -0,0 +1,194 @@
...
+	artifactExists := c.isCsvArtifactPresent(ctx, exec.ID, execStatus.ExportDataDigest)
+	execStatus.FilePresent = artifactExists
+	return execStatus
+}
+
+func (c *controller) isCsvArtifactPresent(ctx context.Context, execID int64, digest string) bool {
+	repositoryName := fmt.Sprintf("scandata_export_%v", execID)
+	exists, err := c.sysArtifactMgr.Exists(ctx, strings.ToLower(export.Vendor), repositoryName, digest)
+	if err != nil {
+		exists = false
```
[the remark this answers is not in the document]
prahaladdarkin: Ack. Will take this in a subsequent CL

src/server/v2.0/handler/scanexport.go
prahaladdarkin: Ack. Will pick this in a subsequent CL

src/pkg/scan/export/filter_processor.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func (dfp *DefaultFilterProcessor) getProjectQueryFilter(user *commonmodels.User) *q.Query {
+	secContext := local.NewSecurityContext(user)
+	if secContext.IsSysAdmin() {
+		logger.Infof("User %v is sys admin. Selecting all projects for export.", user.Username)
+		return q.New(q.KeyWords{})
+	}
+	logger.Infof("User %v is not sys admin. Selecting projects with admin roles for export.", user.Username)
+	return q.New(q.KeyWords{"member": &models.MemberQuery{UserID: user.UserID, Role: common.RoleProjectAdmin}})
```
chlins, line 141: This behavior means if the user is not system admin, he can only list projects for which he is the project administrator? What happened If the user is the Developer role?

prahaladdarkin: Yes currently that is the implementation. I am facing a technical limitation in using the `user.MemberQuery` struct - it only allows one role at a time. I would not be in favor of writing another raw SQL for getting users with specific set of roles. Is there a role I can use which is a combination of both `Project Admin` and `Project Developer`?
cc @wy65701436

wy65701436: let's merge this PR firstly, then use other PR to fix.

chlins (approved):
lgtm

src/server/v2.0/handler/scanexport.go
prahaladdarkin: Done

PR #16045, opened 2021-11-23
add accessory dao service

wy65701436:


chlins (approved):
/lgtm

make/migrations/postgresql/0080_2.5.0_schema.up.sql
```diff
@@ -0,0 +1,22 @@
+/* create table of accessory */
+CREATE TABLE IF NOT EXISTS artifact_accessory (
+    id SERIAL PRIMARY KEY NOT NULL,
+    /*
+       the artifact id of the accessory itself.
+    */
+    artifact_id int,
```
heww, line 7: bigint?

make/migrations/postgresql/0080_2.5.0_schema.up.sql
```diff
@@ -0,0 +1,22 @@
...
+    /*
+     the subject artifact id of the accessory.
+    */
+    subject_artifact_id int,
```
heww, line 11: bigint?

src/pkg/accessory/model/accessory.go
```diff
@@ -0,0 +1,104 @@
...
+	RefProvider
+	RefIdentifier
+	GetID() int64
+	GetArtifactID() int64
+	GetSubjectArtID() int64
+	GetType() string
+	GetSize() int64
+	GetDigest() string
+	GetCreateTime() time.Time
+	GetIcon() string
```
heww, line 75: It's crazy for these methods in this interface here.

```go
type struct AccessoryData {
    ID int64
    ArtifactID int64 
   ....
}

type Accessory interface {
        RefProvider
	RefIdentifier
        GetAccessoryData() AccessoryData
}
```

src/pkg/accessory/model/options.go
```diff
@@ -0,0 +1,82 @@
...
+
+package model
+
+import "time"
+
+// Option function to set the options of the cache
+type Option func(*Options)
+
+// Options options of the cache
+type Options struct {
```
heww, line 23: If there is a method `GetAccessoryData() AccessoryData` in the interface, this `Options` is not needed.

src/pkg/accessory/model/accessory.go
heww: BTW, if there is only one implementation of this interface at the moment, I don't think this interface is needed.

wy65701436: for futher support, it will have SBom, Nydus.

heww (approved):

Issue #15963, opened 2021-11-08
DB lock under 1k concurrency pull


chlins:
What can we help you?

Under the 1k concurrency, pull one artifact, discover some db locks under wait state.

sql: `select wait_event, state, query from pg_stat_activity where wait_event_type = 'Lock'`

![image](https://user-images.githubusercontent.com/31262637/140696806-0c4eeb0b-2141-448f-a559-1d9901c07bfd.png)

PR #16107, opened 2021-12-10
feat: async update artifact pull time and repository pull count

chlins:
Fixes: https://github.com/goharbor/harbor/issues/15963


src/controller/event/handler/internal/artifact.go
```diff
@@ -16,19 +16,63 @@ package internal
...
 	"github.com/goharbor/harbor/src/controller/tag"
+	"github.com/goharbor/harbor/src/lib/config"
 	"github.com/goharbor/harbor/src/lib/log"
+	"github.com/goharbor/harbor/src/lib/orm"
 	"github.com/goharbor/harbor/src/lib/q"
 )
 
+const (
+	// defaultAsyncFlushDuration is the default flush interval.
+	defaultAsyncFlushDuration = 10 * time.Second
```
wy65701436, line 38: can it be configurated?

src/controller/event/handler/internal/artifact.go
```diff
@@ -55,43 +99,137 @@ func (a *Handler) IsStateful() bool {
...
 
 func (a *Handler) onPull(ctx context.Context, event *event.ArtifactEvent) error {
+	// if duration is equal to 0 or negative, keep original sync mode.
+	if asyncFlushDuration <= 0 {
+		var tagName string
+		if len(event.Tags) > 0 {
+			tagName = event.Tags[0]
+		}
+
+		if !config.PullTimeUpdateDisable(ctx) {
```
wy65701436, line 109: do we still have to keep PullTimeUpdateDisable & PullCountUpdateDisable?

cc @stonezdj

src/controller/event/handler/internal/artifact.go
```diff
@@ -55,43 +99,137 @@ func (a *Handler) IsStateful() bool {
...
+	defer a.pullCountLock.Unlock()
+
+	a.pullCountStore[event.Artifact.RepositoryID] = a.pullCountStore[event.Artifact.RepositoryID] + 1
+}
+
+func (a *Handler) syncFlushPullTime(ctx context.Context, artifactID int64, tagName string, time time.Time) {
+	var tagID int64
+
+	if tagName != "" {
 		tags, err := tag.Ctl.List(ctx, &q.Query{
```
wy65701436, line 169: q.New()?

src/controller/event/handler/internal/artifact.go
```diff
@@ -55,43 +99,137 @@ func (a *Handler) IsStateful() bool {
...
 		} else {
 			if len(tags) != 0 {
 				tagID = tags[0].ID
 			}
 		}
 	}
-	if err := artifact.Ctl.UpdatePullTime(ctx, event.Artifact.ID, tagID, time.Now()); err != nil {
-		log.Debugf("failed to update pull time form artifact %d, %v", event.Artifact.ID, err)
+
+	if err := artifact.Ctl.UpdatePullTime(ctx, artifactID, tagID, time); err != nil {
```
wy65701436, line 184: It seems that artifact.Ctl.UpdatePullTime() doesn't handle well the situation of tagID equals 0, please update it to handle the ErrorNotFound.

src/controller/event/handler/internal/artifact.go
```diff
@@ -55,43 +99,137 @@ func (a *Handler) IsStateful() bool {
...
+		log.Warningf("failed to update pull time form artifact %d, %v", artifactID, err)
```
wy65701436, line 185: form ==> for?

src/controller/event/handler/internal/artifact.go
```diff
@@ -55,43 +99,137 @@ func (a *Handler) IsStateful() bool {
...
-	if err := repository.Ctl.AddPullCount(ctx, event.Artifact.RepositoryID); err != nil {
-		log.Debugf("failed to add pull count repository %d, %v", event.Artifact.RepositoryID, err)
+func (a *Handler) asyncFlushPullTime(ctx context.Context) {
+	for {
+		<-time.After(asyncFlushDuration)
+		a.pullTimeLock.Lock()
+
+		for key, time := range a.pullTimeStore {
+			keys := strings.Split(key, ":")
+			artifactID, err := strconv.ParseInt(keys[0], 10, 64)
```
wy65701436, line 202: check keys length before using?

src/controller/event/handler/internal/artifact.go
```diff
@@ -55,43 +99,137 @@ func (a *Handler) IsStateful() bool {
...
 
-func (a *Handler) addPullCount(ctx context.Context, event *event.ArtifactEvent) {
-	if err := repository.Ctl.AddPullCount(ctx, event.Artifact.RepositoryID); err != nil {
-		log.Debugf("failed to add pull count repository %d, %v", event.Artifact.RepositoryID, err)
+func (a *Handler) asyncFlushPullTime(ctx context.Context) {
+	for {
+		<-time.After(asyncFlushDuration)
+		a.pullTimeLock.Lock()
+
+		for key, time := range a.pullTimeStore {
```
wy65701436, line 200: can we consider batch update?

src/controller/event/handler/internal/artifact.go
```diff
@@ -16,19 +16,63 @@ package internal
...
+)
+
+var (
+	asyncFlushDuration time.Duration
+)
+
+func init() {
+	// get async flush duration from env, if not provide,
+	// use default value: 10*time.second
+	envDuration := os.Getenv("ARTIFACT_PULL_TIME_ASYNC_FLUSH_DURATION")
```
wy65701436, line 48: change another name as it will handle pull count as well?

wy65701436:
@chlins please add the situation description that we discussed offline.

src/controller/event/handler/internal/artifact.go
reasonerjt: I hope not.

src/controller/event/handler/internal/artifact.go
```diff
@@ -55,43 +99,137 @@ func (a *Handler) IsStateful() bool {
 }
 
 func (a *Handler) onPull(ctx context.Context, event *event.ArtifactEvent) error {
+	// if duration is equal to 0 or negative, keep original sync mode.
```
reasonerjt, line 102: Does this func always return nil?

src/controller/event/handler/internal/artifact.go
reasonerjt: seems we should handle that in the artifact controller?

src/pkg/artifact/dao/dao.go
```diff
@@ -180,7 +180,27 @@ func (d *dao) Update(ctx context.Context, artifact *Artifact, props ...string) e
...
+			if artifact.PullTime.Before(old.PullTime) {
+				artifact.PullTime = old.PullTime
+				continue
+			}
+		}
+
+		newProps = append(newProps, prop)
+	}
+
+	n, err := ormer.Update(artifact, newProps...)
```
heww, line 203: I prefer to add a new method to only update the pull time in the DAO.

src/controller/event/handler/internal/artifact.go
chlins: ARTIFACT_ASYNC_FLUSH_DURATION

src/controller/event/handler/internal/artifact.go
chlins: > Does this func always return nil?

Yes, it always returns nil and only print error logs if meet error.

src/controller/event/handler/internal/artifact.go
chlins: ok

src/controller/event/handler/internal/artifact.go
chlins: Let's use another pr to fix the situation.

src/pkg/artifact/dao/dao.go
chlins: good point

src/controller/event/handler/internal/artifact.go
chlins: keep for more investigation.

wy65701436 (approved):
lgtm

src/controller/event/handler/internal/artifact.go
chlins: use other pr to remove it if we need.

src/pkg/artifact/dao/dao.go
```diff
@@ -189,6 +193,27 @@ func (d *dao) Update(ctx context.Context, artifact *Artifact, props ...string) e
...
+func (d *dao) UpdatePullTime(ctx context.Context, id int64, pullTime time.Time) error {
+	ormer, err := orm.FromContext(ctx)
+	if err != nil {
+		return err
+	}
+
+	// can only be retained to the second if not format
+	formatPullTime := pullTime.Format("2006-01-02 15:04:05.999999")
+	// update db only if pull_time is null or pull_time < (in-coming)pullTime
+	sql := "UPDATE artifact SET pull_time = ? WHERE id = ? AND pull_time IS NULL OR pull_time < ?"
```
heww, line 206: `"UPDATE artifact SET pull_time = ? WHERE id = ? AND pull_time IS NULL OR pull_time < ?"` => 

`"UPDATE artifact SET pull_time = ? WHERE id = ? AND (pull_time IS NULL OR pull_time < ?)"`

src/controller/event/handler/internal/artifact.go
```diff
@@ -16,19 +16,63 @@ package internal
...
+	} else {
+		duration, err := strconv.ParseInt(envDuration, 10, 64)
+		if err != nil {
+			log.Warningf("error to parse ARTIFACT_PULL_ASYNC_FLUSH_DURATION: %v, will use default value: %v", err, defaultAsyncFlushDuration)
+			asyncFlushDuration = defaultAsyncFlushDuration
+		} else {
+			asyncFlushDuration = time.Duration(duration) * time.Second
+		}
+	}
+}
```
heww, line 61: Should this value be configurable in the system configuration? @wy65701436

src/controller/event/handler/internal/artifact.go
```diff
@@ -55,43 +99,136 @@ func (a *Handler) IsStateful() bool {
...
+	if err := repository.Ctl.AddPullCount(ctx, repositoryID, count); err != nil {
+		log.Warningf("failed to add pull count repository %d, %v", repositoryID, err)
 	}
-	return
 }
 
-func (a *Handler) addPullCount(ctx context.Context, event *event.ArtifactEvent) {
-	if err := repository.Ctl.AddPullCount(ctx, event.Artifact.RepositoryID); err != nil {
-		log.Debugf("failed to add pull count repository %d, %v", event.Artifact.RepositoryID, err)
+func (a *Handler) asyncFlushPullTime(ctx context.Context) {
```
wy65701436, line 194: generally, we should consider the listen os shutdown signal in the goroutine to ensure update?

src/controller/event/handler/internal/artifact.go
chlins: Use another pr to expose this to system configuration if we need.

src/controller/event/handler/internal/artifact.go
stonezdj: We should keep the setting of PullAuditLogDisable

stonezdj (approved):
LGTM

PR #16150, opened 2021-12-21
Cosign artifact api

wy65701436:
1. update artifact list & delete api to support accessory
2. add list accesories api


src/controller/artifact/controller.go
```diff
@@ -229,11 +232,23 @@ func (c *controller) List(ctx context.Context, query *q.Query, option *Option) (
...
+	var artifacts []*artifact.Artifact
+	// Only the displayed accessory will in the artifact list
 	for _, art := range arts {
-		artifacts = append(artifacts, c.assembleArtifact(ctx, art, option))
+		accs, err := c.accessoryMgr.List(ctx, q.New(q.KeyWords{"ArtifactID": art.ID, "digest": art.Digest}))
+		if err != nil {
+			return nil, err
+		}
+		if len(accs) == 0 || (len(accs) > 0 && accs[0].Display()) {
+			artifacts = append(artifacts, art)
```
chlins, line 243: can we use `artifacts = append(artifacts,  c.assembleArtifact(ctx, art, option))` directly to avoid below additional iterate and assemble ?

src/controller/artifact/controller.go
```diff
@@ -229,11 +232,23 @@ func (c *controller) List(ctx context.Context, query *q.Query, option *Option) (
...
-	var artifacts []*Artifact
+	var artifacts []*artifact.Artifact
+	// Only the displayed accessory will in the artifact list
 	for _, art := range arts {
-		artifacts = append(artifacts, c.assembleArtifact(ctx, art, option))
+		accs, err := c.accessoryMgr.List(ctx, q.New(q.KeyWords{"ArtifactID": art.ID, "digest": art.Digest}))
+		if err != nil {
+			return nil, err
+		}
+		if len(accs) == 0 || (len(accs) > 0 && accs[0].Display()) {
```
chlins, line 242: the Display() is hard code?

wy65701436: base model under the accessory set it to false, as well as cosign.

src/controller/artifact/controller.go
wy65701436: updated.

chlins (approved):
lgtm

src/common/rbac/const.go
```diff
@@ -55,6 +55,7 @@ const (
 	ResourceScanner               = Resource("scanner")
 	ResourceArtifact              = Resource("artifact")
 	ResourceTag                   = Resource("tag")
+	ResourceAccessory             = Resource("accessory")
```
heww, line 58: We need to update the permissions map of the role.

wy65701436: added, thanks.

heww (approved):

PR #16178, opened 2021-12-29
Extend install.sh with switch to specify offline installer package

chrifey:
The `install.sh` script uses `./harbor*.tar.gz` as path to the offline install package (to load the container images). 

It would be nice to have this configurable as an option. This pullrequest adds this switch to `install.sh`.

Example:
```
./install.sh --with-trivy --with-chartmuseum --offline-installer-file /opt/harbor.v2.4.1.tar.gz
```
If nothing is specified, it just defaults to `/harbor*.tar.gz`.


bitsf:
the install.sh and harbor.yaml.tmpl may also changed in new version, you should not just use the image tar.

chrifey:
Hi @bitsf,

I think I do not fully understand your answer.

Usually I download the offline Installer bundle and use this file to install or upgrade harbor instances (within airgapped environments). The main reason for this is, to be able to "docker load" the images (versus pulling them directly from dockerhub or a local mirror registry).

Here is my actual problem, since the path so the `tar.gz` is hardcoded at the moment.

Can you elaborate on why I should not use the offline bundle and why it does not make sense to have this configurable?

Kind regards
Christian

PR #16187, opened 2022-01-05
Cosign policy checker

wy65701436:
Enable policy checker for cosign, when it's enabled, user cannot pull artifact without cosign.

### DIFF
--- a/src/server/middleware/util/util.go
+++ b/src/server/middleware/util/util.go
@@ -15,7 +15,6 @@
 package util
 
 import (
-	"context"
 	"fmt"
 	"github.com/goharbor/harbor/src/common/rbac/project"
 	"net/http"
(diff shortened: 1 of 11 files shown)


chlins (approved):
lgtm

src/server/middleware/contenttrust/notaryv1.go
```diff
@@ -28,8 +28,8 @@ var (
 	}
 )
 
-// Middleware handle docker pull content trust check
-func Middleware() func(http.Handler) http.Handler {
+// NotaryV1 handle docker pull content trust check
+func NotaryV1() func(http.Handler) http.Handler {
```
heww, line 32: is there version 2 of the notary? If not I think it's enough to name it as `Notary` here.

wy65701436: yes, there is. https://github.com/notaryproject/notaryproject

src/server/middleware/util/util.go
```diff
@@ -57,13 +56,17 @@ func ParseProjectName(r *http.Request) string {
...
-	// only scanner pull access can bypass.
-	if ok && secCtx.Name() == "v2token" &&
-		secCtx.Can(ctx, rbac.ActionScannerPull, project.NewNamespace(projectID).Resource(rbac.ResourceRepository)) {
-		return true
+	// 1, scanner pull access can bypass.
+	// 2, cosign pull can bypass, it needs to pull the manifest before pushing the signature.
+	if ok && secCtx.Name() == "v2token" {
+		if secCtx.Can(r.Context(), rbac.ActionScannerPull, project.NewNamespace(projectID).Resource(rbac.ResourceRepository)) ||
+			(secCtx.Can(r.Context(), rbac.ActionPush, project.NewNamespace(projectID).Resource(rbac.ResourceRepository)) &&
+				strings.Contains(r.UserAgent(), "cosign")) {
```
heww, line 67: Is there an escape here?

I think we only can skip checking the sig image, not all the images.

src/server/middleware/contenttrust/notaryv1.go
wy65701436: it's ok to use notary.

heww (approved):

src/server/middleware/util/util.go
wy65701436: yes, the request from cosign client with push permission can bypass the check.

PR #16193, opened 2022-01-07
[cherry-pick]Fix: using traditional `PKCS#1` format RSA key

YangJiao0817:
The openssl 3.0.0 using newer `PKCS#8` format.
But it's not compatitable with harbor core
So using tradictional format instead


wy65701436 (approved):
lgtm

chlins (approved):
lgtm

Issue #16200, opened 2022-01-10
Need current time property for systeminfo  API


AllForNothing:
We should add the current time property to the response of the systeminfo  API, which can help the UI to calculate some important Infos, like the time remaining for the robot token.

PR #16205, opened 2022-01-11
feat: add current_time to the response of systeminfo api when user logged in

zyyw:
Closes: #16200 

### DIFF
--- a/src/controller/systeminfo/controller.go
+++ b/src/controller/systeminfo/controller.go
@@ -17,14 +17,15 @@ package systeminfo
 import (
 	"context"
 	"fmt"
-	"github.com/goharbor/harbor/src/lib/config"
-	"github.com/goharbor/harbor/src/lib/config/models"
 	"io"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/goharbor/harbor/src/common"
 	"github.com/goharbor/harbor/src/common/utils"
+	"github.com/goharbor/harbor/src/lib/config"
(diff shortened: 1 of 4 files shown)


src/controller/systeminfo/controller.go
```diff
@@ -50,6 +51,7 @@ type Data struct {
 }
 
 type protectedData struct {
+	CurrentTime                 time.Time
```
wy65701436, line 54: why it is an protected data?

zyyw: because the current_time is only added to the response of systeminfo API when user is logged in.

src/server/v2.0/handler/systeminfo.go
```diff
@@ -90,6 +91,8 @@ func (s *sysInfoAPI) convertInfo(d *si.Data) *models.GeneralInfo {
 		res.ReadOnly = &d.Protected.ReadOnly
 		res.RegistryStorageProviderName = &d.Protected.RegistryStorageProviderName
 		res.NotificationEnable = &d.Protected.NotificationEnable
+		currentTime := strfmt.DateTime(d.Protected.CurrentTime)
+		res.CurrentTime = &currentTime
```
heww, line 95: If this field is always returned, we don't need to set it in the protected.

wy65701436 (approved):
lgtm

AllForNothing (approved):

PR #16210, opened 2022-01-12
Support repo scope in webhook

mmpei:
Signed-off-by: mmpei <peimingming@corp.netease.com>


Vad1mo:
@mmpei can you elaborate a bit more on your PR? 

We don't won't that your effort is wasted, help us understand the issue.

PR #16228, opened 2022-01-14
fix cosign conflict error on landing data

wy65701436:
Cosign client will generate the same signature to the same manifest, ignore the conflict error in middleware

### DIFF
--- a/src/server/middleware/cosign/cosign.go
+++ b/src/server/middleware/cosign/cosign.go
@@ -1,12 +1,14 @@
 package cosign
 
 import (
+	"context"
 	"fmt"
 	"github.com/docker/distribution/reference"
 	"github.com/goharbor/harbor/src/controller/artifact"
 	"github.com/goharbor/harbor/src/lib"
 	"github.com/goharbor/harbor/src/lib/errors"
 	"github.com/goharbor/harbor/src/lib/log"
+	"github.com/goharbor/harbor/src/lib/orm"
 	"github.com/goharbor/harbor/src/pkg/accessory"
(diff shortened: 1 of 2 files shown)


chlins (approved):
lgtm

src/server/middleware/cosign/cosign.go
```diff
@@ -111,8 +111,8 @@ func CosignSignatureMiddleware() func(http.Handler) http.Handler {
 				Digest:        desc.Digest.String(),
 				Type:          model.TypeCosignSignature,
 			})
-			if err != nil {
-				logger.Errorf("failed to get cosign signature artifact: %s, error: %v", desc.Digest.String(), err)
+			if err != nil && !errors.IsConflictErr(err) {
+				logger.Errorf("failed to create cosign signature artifact: %s, error: %v", desc.Digest.String(), err)
 				return err
 			}
 		}
```
heww, line 118: Please wrap these lines in the tx, otherwise, the conflict error will break the DB conn if the conn is from a tx.

```
			_, err = accessory.Mgr.Create(ctx, model.AccessoryData{
				ArtifactID:    art.ID,
				SubArtifactID: subjectArt.ID,
				Size:          desc.Size,
				Digest:        desc.Digest.String(),
				Type:          model.TypeCosignSignature,
			})
			if err != nil && !errors.IsConflictErr(err) {
				logger.Errorf("failed to create cosign signature artifact: %s, error: %v", desc.Digest.String(), err)
				return err
			}
```

heww (approved):

PR #16235, opened 2022-01-17
fix: Clean old version scan reports of trivy

### LINKED ISSUE #16221: Clean old version scan reports of trivy
From harbor 2.4, the trivy does not support scan report v1.0, we need to remove the away from the db.

```sql
DELETE FROM scan_report WHERE mime_type='application/vnd.scanner.adapter.vuln.report.harbor+json; version=1.0' AND registration_uuid IN (SELECT uuid FROM scanner_registration WHERE name='Trivy' AND immutable='true')
```

heww:
From harbor 2.4, the trivy does not support scan report v1.0, we need to remove the old version scan reports from the DB.

Closes #16221


zyyw (approved):

wy65701436 (approved):
lgtm

Issue #16145, opened 2021-12-20
Trivy is not fully disconnected usable with skip_update


sizowie:
When using the Trivy scanner with Harbor and the configuration setting `skip_update: true`, my expectations is that scans can be done fully disconnected from the internet. OS scanning works as expected with offline db patterns. But at least with Harbor 2.4 Trivy runs with ENV `SCANNER_TRIVY_VULN_TYPE=os,library` meaning that libraries/files types with JAR/WAR/EAR will be scanned dynamically through search.maven.org with the checksum of the corespoding files.

That results in really slow scans when having no internet connectivity, as the scanner will retry to reach search.maven.org.

Either the vuln_type should be configurable in harbor.yml or a dependency between `skip_update:true` and the SCANNER_TRIVY_VULN_TYPE (meaning = OS only) should be set, as long as trivy doesn't have a fix for that.

See also https://github.com/aquasecurity/trivy/issues/1185

Harbor: 2.4.0

heww:
Hi, @danielpacak seems that this issue was resolved in trivy. Do we need to bump up a new release of trivy in harbor 2.4.x to resolve this problem on the harbor side?

danielpacak:
> Hi, @danielpacak seems that this issue was resolved in trivy. Do we need to bump up a new release of trivy in harbor 2.4.x to resolve this problem on the harbor side?

This would require bumping up Trivy to 0.22.0 in harbor-scanner-trivy and add support for the --offline-scan flag that was added to resolve exactly this problem.

PR #16246, opened 2022-01-19
feat: bump TRIVYVERSION to v0.22.0 and bump TRIVYADAPTERVERSION to v0.25.0

zyyw:
Closes: #16145


make/harbor.yml.tmpl
```diff
@@ -91,6 +91,7 @@ trivy:
   # https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line
   #
   # github_token: xxx
+  offline_scan: false
```
danielpacak, line 94: We should describe this new param and how it relates to `skip_update`. You can copy the explanation from https://github.com/aquasecurity/trivy/discussions/1518

danielpacak (changes requested):
LGTM. Let's just add description to the new param before we merge this PR.

make/harbor.yml.tmpl
danielpacak: NIT: Add it as close to `skip_update` property. In air-gapped environment both properties must be set to true so it may help grouping related configuration options.

zyyw: thank you for the comment. the PR is updated now.

heww (approved):

danielpacak (approved):

PR #16260, opened 2022-01-22
Group members lose access to push or see projects on Harbor

### LINKED ISSUE #16220: Group members lose access to push or see projects on Harbor - LDAP authentication
Under ldap_auth mode, some LDAP user can login, but they cannot access the project which they are group members.
During the user login, the core log output the error.
```
pq: duplicate key value violates unique constraint "unique_group_name"
```
Root cause:
When the LDAP user login, Harbor will retrieve the LDAP group which the current user is a member of.
and search the user_group to find out if existing user with the same ldap_group_dn exist, if not present, then create a user group , the group name is the CN of the group. but the CN might be duplicated in LDAP, it result in the duplicate user group name error.

stonezdj:
Handle the case if there is duplicate user group name when onboard ldap user group
   Continue to attach groups when it fail on one item
   Fixes #16220

### DIFF
--- a/src/pkg/usergroup/manager.go
+++ b/src/pkg/usergroup/manager.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"github.com/goharbor/harbor/src/common"
 	"github.com/goharbor/harbor/src/common/utils"
+	"github.com/goharbor/harbor/src/lib/log"
 	"github.com/goharbor/harbor/src/lib/q"
 	"github.com/goharbor/harbor/src/pkg/usergroup/dao"
 	"github.com/goharbor/harbor/src/pkg/usergroup/model"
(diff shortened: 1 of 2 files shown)


src/pkg/usergroup/manager.go
```diff
@@ -102,10 +105,35 @@ func (m *manager) UpdateName(ctx context.Context, id int, groupName string) erro
...
+			g.ID = ug[0].ID
+			return nil
+		}
+		// if duplicated with name, fall back to ldap group dn
+		if len(g.LdapGroupDN) <= 255 {
+			g.GroupName = g.LdapGroupDN
+		} else {
+			g.GroupName = g.LdapGroupDN[:254]
+		}
+		log.Warningf("existing duplicate user group with the same name, name the current user group with ldap group DN %v", g.GroupName)
```
wy65701436, line 132: move this line to 122, this will recore all the duplicate cases.

wy65701436 (approved):
lgtm

chlins (approved):
lgtm

PR #16282, opened 2022-01-25
add cosign support in replication

wy65701436:
For the case Harbor-to-Harbor, the accessory can be replicated from source or to target.


chlins (approved):
lgtm

heww (approved):

PR #16285, opened 2022-01-25
Add options to the user.Count method

### LINKED ISSUE #16269: User count is incorrect
Expected: The number of users is counted correctly
Actual: The number of users is incorrectly counted

Go to users page
WechatIMG79

**Versions:**
- harbor version: v2.5.0-70155172

stonezdj:
Exclude the admin account by default
  Fixes #16269


src/pkg/user/manager.go
```diff
@@ -116,7 +116,11 @@ func (m *manager) MatchLocalPassword(ctx context.Context, usernameOrEmail, passw
 	return nil, nil
 }
 
-func (m *manager) Count(ctx context.Context, query *q.Query) (int64, error) {
+func (m *manager) Count(ctx context.Context, query *q.Query, options ...models.Option) (int64, error) {
```
wy65701436, line 119: should we keep the list & count as same logic?

stonezdj: update

src/pkg/user/manager.go
```diff
@@ -116,7 +116,11 @@ func (m *manager) MatchLocalPassword(ctx context.Context, usernameOrEmail, passw
...
+	opts := models.NewOptions(options...)
+	if !opts.IncludeDefaultAdmin && query != nil && query.Keywords != nil {
```
ywk253100, line 121: If the opts.IncludeDefaultAdmin is false and the query is nil, seems this doesn't work as expected.
How about changing it as following?
```go
if !opts.IncludeDefaultAdmin {
  if query == nil {
    // new a new query
 }
if query.Keywords == nil {
  // new keywords
}
}
```

stonezdj: ok, updated

ywk253100 (approved):

wy65701436 (approved):
lgtm

Issue #15155, opened 2021-06-17
Remove replication of proxy project


govindkailas:
We enable replication from our Harbor primary site to DR site. Something that we notice is whenever it reaches the proxy projects the replications are failing with the below error, 
` http status code: 403, body: {"errors":[{"code":"DENIED","message":"denied: can not push artifact to a proxy project: dockerhub-proxy-cache"}]}`

If this is by design, please consider skipping **proxy project** from replication.

stonezdj:
It is possible to exclude the proxy cache project when pulling images to a remote, we could check the current project is proxy cache project.

stonezdj:
Create proxy cache project dockhub_proxy in server A and B. and setup an event-based replication, it seems that this error can not reproduce now.
When caching an image in `A/dockerhub_proxy/library/hello-world:latest `, it trigger a replication job which replicate the image to `A/dockerhub_proxy/library/hello-world:latest`  to `B/dockerhub_proxy/library/hello-world:latest`.
Before replication, server A check if the artifact exists in the remote registry, it sends a HEAD request the server B. check if the `B/dockerhub_proxy/library/hello-world:latest` exist, because B/dockerhub_proxy is a proxy-cache project, it proxy the request to dockerhub and HTTP 200 to the server A eventually, then server A consider this image already exists in server B. and mark this replication job succeed.

pjastrzabek:
Before 2.5.x versions we had a working helm charts replication into proxy project.

That harbor project had 2 functions:
- it was a proxy project for docker images
- and was holding replicated helm charts 

After update to 2.5.x helm chart replication (explicitly selected in replication rule) does not work (All created replication tasks are skipped with a log "skip create replication task because of dest limitation")

PR #16286, opened 2022-01-25
fix: skip replication to proxy cache project

chlins:
Fixes: https://github.com/goharbor/harbor/issues/15155

### DIFF
--- a/src/controller/replication/flow/copy.go
+++ b/src/controller/replication/flow/copy.go
@@ -20,6 +20,7 @@ import (
 
 	repctlmodel "github.com/goharbor/harbor/src/controller/replication/model"
 	"github.com/goharbor/harbor/src/jobservice/job"
+	"github.com/goharbor/harbor/src/jobservice/logger"
 	"github.com/goharbor/harbor/src/lib/log"
 	"github.com/goharbor/harbor/src/pkg/reg/model"
 	"github.com/goharbor/harbor/src/pkg/task"
@@ -103,12 +104,29 @@ func (c *copyFlow) isExecutionStopped(ctx context.Context) (bool, error) {
 }
 
 func (c *copyFlow) createTasks(ctx context.Context, srcResources, dstResources []*model.Resource, speed int32) error {
-	for i, resource := range srcResources {
-		src, err := json.Marshal(resource)
+	var taskCnt int
+	defer func() {
+		// if no task be created, mark execution done.
+		if taskCnt == 0 {
+			if err := c.executionMgr.MarkDone(ctx, c.executionID, "no resources need to be replicated"); err != nil {
+				logger.Errorf("failed to mark done for the execution %d: %v", c.executionID, err)
+			}
+		}
+	}()
+
+	for i, srcResource := range srcResources {
+		dstResource := dstResources[i]
+		// if dest resource should be skipped, ignore replicate.
+		if dstResource.Skip {
+			log.Warningf("skip create replication task because of dest limitation, src: %s, dst: %s", srcResource.Metadata, dstResource.Metadata)
+			continue
+		}
+
+		src, err := json.Marshal(srcResource)
 		if err != nil {
 			return err
 		}
-		dest, err := json.Marshal(dstResources[i])
+		dest, err := json.Marshal(dstResource)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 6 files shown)


src/pkg/reg/adapter/harbor/base/adapter.go
```diff
@@ -205,6 +210,17 @@ func (a *Adapter) PrepareForPush(resources []*model.Resource) error {
 		}
 		log.Debugf("project %s created", project.Name)
 	}
+
+	// do filter for proxy cache projects.
+	for _, res := range resources {
+		paths := strings.Split(res.Metadata.Repository.Name, "/")
```
wy65701436, line 216: use util.ParseRepository()

src/pkg/reg/adapter/harbor/base/adapter.go
```diff
@@ -182,9 +182,14 @@ func (a *Adapter) PrepareForPush(resources []*model.Resource) error {
 		return errors.Wrapf(err, "list projects with query %s", q)
 	}
 
+	proxyCacheProjects := make(map[string]*Project)
```
wy65701436, line 185: make(map[string]bool) is enough?

src/pkg/reg/model/resource.go
```diff
@@ -33,6 +38,9 @@ type Resource struct {
 	IsDeleteTag bool `json:"is_delete_tag"`
 	// indicate whether the resource can be overridden
 	Override bool `json:"override"`
+	// Skip is a flag for resource which satisfies replication rules but should
+	// be skipped because of other limits like when dest project's type is proxy cache.
+	Skip bool `json:"-"`
```
wy65701436, line 43: it's better to add an Skip() method since we may add more attribute to control the project filter, no just for proxy cached project.

src/controller/replication/flow/copy.go
```diff
@@ -127,11 +145,13 @@ func (c *copyFlow) createTasks(ctx context.Context, srcResources, dstResources [
...
-			"resource_type":        string(resource.Type),
-			"source_resource":      getResourceName(resource),
-			"destination_resource": getResourceName(dstResources[i])}); err != nil {
+			"resource_type":        string(srcResource.Type),
+			"source_resource":      getResourceName(srcResource),
+			"destination_resource": getResourceName(dstResource)}); err != nil {
 			return err
 		}
+
+		taskCnt++
```
wy65701436, line 154: Is it possible that even the task creation returns error, but the task still to be created? And in this situation, what will happen after calls MarkDone()?

src/pkg/reg/adapter/harbor/base/adapter.go
```diff
@@ -182,9 +182,14 @@ func (a *Adapter) PrepareForPush(resources []*model.Resource) error {
...
 	existProjects := make(map[string]*Project)
 	for _, p := range queryProjects {
 		existProjects[p.Name] = p
+		// if project with registry_id, that means this is a proxy cache project.
+		if p.RegistryID > 0 {
```
wy65701436, line 190: we don't have any other attribution to judge this but the registry id?

chlins: Discuss with @stonezdj @AllForNothing, no other attribution can identify proxy cache project, same judgement with UI.

src/pkg/reg/adapter/harbor/base/adapter.go
chlins: `util.ParseRepository()` will return `a/b` when repository is `a/b/c`, the actual project name is `a`.

src/pkg/reg/adapter/harbor/base/adapter.go
chlins: updated

src/controller/replication/flow/copy.go
chlins: no, if create task error, task manager will ensure no task in database and jobservice.

src/pkg/reg/model/resource.go
chlins: if in struct have a filed `Skip`, we can not add method name same with field.

wy65701436 (approved):
lgtm

stonezdj (approved):

src/pkg/reg/adapter/harbor/base/adapter.go
stonezdj: It seems that the util.ParseRepository() is an unused function and can be removed

PR #16301, opened 2022-01-27
(Cherry-pick) fix: vulnerability_record & report_vulnerability_record table "integer out of range" error

zyyw:


zyyw:
This will not roll out in 2.4.2

PR #16304, opened 2022-01-27
fix accessory lost tags

wy65701436:
It needs to copy tags for the accessories.


heww (approved):

chlins (approved):
lgtm

PR #16190, opened 2022-01-06
Upgrade distribution to 2.8

darend:
## what

Upgrade the [distribution package to 2.8](https://github.com/distribution/distribution/releases/tag/v2.8.0)

## why

Harbor is currently on 2.7.1 which was released in 2019. Full release notes available here https://github.com/distribution/distribution/releases/tag/v2.8.0

## EDIT: IRSA support not included in this release

See https://github.com/distribution/distribution/pull/3552#issuecomment-1018761583

~The 2.7 release uses an out of date AWS SDK which does not support the AWS AssumeRoleWithWebIdentity required for deployments under Kubernetes that use IRSA. This was fixed in the 2.8 release, see https://github.com/distribution/distribution/issues/3097. Support for IRSA is of growing importance. It is the recommended solution when running Kubernetes on AWS~

~Unfortunately while distribution has a 2.8 branch they have not made a release. I pinned the upgrade to the last commit of the 2.8 release branch~

~IRSA Documentation:~
- ~Overview https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/~
- ~Required AWS SDK versions https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-minimum-sdk.html~

~This fixes https://github.com/goharbor/harbor-helm/issues/725~

PR #16322, opened 2022-02-02
fix: add patch for registry layers larger than 10G with S3 backend

### LINKED ISSUE #15719: Uploading image with large layers fails after push
### How can we help you?
We have an issue pushing a 15GB+ container with a maximum single layer size of 13.5GB.

We are running an instance of Harbor on ec2 via a `t3a.small` instance type:
 * Harbor: Version v2.1.2-fcc6751d

The system works fine with all containers to-date, and we also have several that are large than 15GB in total size.  Today we tried to push an image with a maximum single-layer size of 13.5GB, and it failed repeatedly.  On the client side we only see:
```
Sep 29 22:29:19 backend-ci dockerd[934]: time="2021-09-29T22:29:19.628869277+02:00" level=error msg="Upload failed, retrying: blob upload unknown"
Sep 29 22:38:28 backend-ci dockerd[934]: time="2021-09-29T22:38:28.463305244+02:00" level=error msg="Upload failed, retrying: blob upload unknown"
Sep 29 22:47:49 backend-ci dockerd[934]: time="2021-09-29T22:47:49.120969007+02:00" level=error msg="Upload failed, retrying: blob upload unknown"
```

We are not running any additional proxy, and we have the harbor nginx config set as recommended:
```
client_max_body_size 0;
proxy_send_timeout 900;
proxy_read_timeout 900;
```

We are running an instance of `clair`, for which I know you are removing support, and it seems to be quitting intermittently, but it does not seem to be an issue for any other push/pull activities.

NOTE: again the `push` activity actually completes, but the client side error only appears afterwards.  I don't see any errors in the docker service when it fails.

Any thoughts or advice will be very welcomed; perhaps there are other experiencing similar issues with very-large-layers.

franznemeth:
This PR fixes an issue if you use S3 as a storage backend, where you cannot push layers larger than 10GB in size.

The registry fails with the following message:
```
harbor-registry-7fd5cd687b-xz4kc registry time="2022-02-01T10:11:34.289971246Z" level=error msg="upload resumed at wrong offest: 5242880000 != 7615513624" auth.user.name="harbor_registry_user" go.version=go1.15.6 http.request.host=registry.example.org http.request.id=<removed> http.request.method=PUT http.request.remoteaddr=<removed> http.request.uri="/v2/<removed>" http.request.useragent="docker/19.03.5 go/go1.12.12 git-commit/2ee0c57608 os/windows arch/amd64 UpstreamClient(Docker-Client/19.03.5 \(windows\))" vars.name="<removed>" vars.uuid=<removed>
```

This problem was fixed upstream in the distribution/distribution repository through https://github.com/distribution/distribution/pull/2815 but never included in a release which is why I am adding it as a patch to the registry version that harbor is building.

If you want me to adapt https://github.com/goharbor/harbor/blob/main/Makefile#L111 as well please tell me.

Signed-off-by: Franz Nemeth <franz.nemeth@dynatrace.com>

Closes #15719


stonezdj:
It seems that the distribution v2.8 already integrate this PR and we could bump it up to v2.8 instead of patch.

franznemeth:
So that means waiting for #16190 instead of this fix?

Vad1mo:
I think we need to reopen that or do you see other results @franznemeth?

```sh
/usr/bin/registry_DO_NOT_USE_GC -v    
/usr/bin/registry_DO_NOT_USE_GC github.com/docker/distribution v2.8.0
```

I tried to push 
- `docker pull vad1mo/10gb-random-file:multilayer`  10 layer a 1GB result: **WORKS** 🎉  
- `docker pull vad1mo/10gb-random-file:latest` 1 layer a 10 GB  result: **FAIL 🔥**

franznemeth:
No I've had the exact same results. At the time of testing there was only a beta release of 2.8.0 so I was assuming that it would be fixed in the release. 
I have since used my patch in our deployment since this is a real world problem blocking us from using certain images in our environment (e.g. windows images with large application installations like visual studio).

I would suggest reopening the PR

Vad1mo:
Ok, I figured that out now.

The patch https://github.com/distribution/distribution/pull/2815 was merged into main but did not make it into the 2.8.0 and 2.8.1 release. It seems like this will be part of 3.x

@stonezdj can you please reopen this ticket?

Vad1mo (approved):

Vad1mo (approved):

Vad1mo (approved):

stonezdj (approved):

OrlinVasilev:
@franznemeth can you rebase please and push!

Vad1mo:
@franznemeth, would you mind pushing again so that the CI runs again. Somehow we can't rerun it ourself 😞

OrlinVasilev (approved):

PR #16324, opened 2022-02-02
[FIX] avoid spam debug message in logs (Update unauthorized.go)

gabo89:
avoid spam debug message about  "an unauthorized security context generated for request GET /api/v2.0/ping"  when liveness probe is set to  /api/v2.0/ping on the helm charts

Signed-off-by: gabriel cuadros gabrielcuadros@gmail.com


MinerYang:
You might edit the log-level if you don't need it.

PR #16333, opened 2022-02-08
Remove noise in log

stonezdj:
DB Config Manager could be registered twice if need to enable cache
  Get trace config only when the trace is enabled


chlins (approved):
lgtm

wy65701436 (approved):
lgtm

PR #16351, opened 2022-02-11
bumup redigo version

wy65701436:
To upgrade redigo version to v1.8.8 to support redis 6.0 for username configuration


zyyw:
Currently harbor using beego v1.12.1, which uses `github.com/gomodule/redigo v2.0.0+incompatible` as the client to connect redis. This redigo version does not supporting username+password authentication.

reference: https://github.com/beego/beego/issues/4861

wy65701436:
close it as beego doesn't support the latest redigo version.

PR #16378, opened 2022-02-16
fix accessory copy for index

wy65701436:
When issue an copy, it has to copy the reference's accessory as well


chlins (approved):
lgtm

heww (approved):

PR #16401, opened 2022-02-22
fix replicate reference accessory

### LINKED ISSUE #16375: replication and copy manifest list cosign accessory issue
replication
expected:The cosign accessory can be replicated with or without a tag
actual:Sub-Artifact cannot replicate without tag

copy
expected:You can copy all the cosign accessories
actual:The cosign accessory of the parent Artifact is not copied, and the cosign accessory of the child Artifact has one passed and one has not passed.

replication
1.HarborA HarborB
2.Create a project on HarborA as test01, and push manifest list
3.The manifest list has two sub-Artifacts, one for A and one for B
4.Use cosign sign A and B
5.Trigger replication to HarborB
6.Let A have tag and B have no tag
7.Trigger replication to HarborB to see that A on HarborB has a cosign accessory, while B does not
8.If you make B have a tag in the starting replication, you can see that B in HarborB has a cosign accessory
copy
1.Create a project as test01, and push manifest list
2.Create a project as test02
3.The manifest list has two sub-Artifacts, one for A and one for B
4.Use cosign sign A and B and Parent Artifact
5.Copy manifest list from test01 to test02
6.Parent Artifact has no cosign accessory, A has cosign accessory, B has no cosign accessory

**Versions:**
- harbor version: v2.5.0-fd92efe1

wy65701436:
fixes #16375

### DIFF
--- a/src/pkg/reg/adapter/harbor/v2/client.go
+++ b/src/pkg/reg/adapter/harbor/v2/client.go
@@ -55,23 +55,13 @@ func (c *client) listArtifacts(repo string) ([]*model.Artifact, error) {
 		return nil, err
 	}
 	var arts []*model.Artifact
-	for _, artifact := range artifacts {
-		art := &model.Artifact{
-			Type:   artifact.Type,
-			Digest: artifact.Digest,
-		}
-		for _, label := range artifact.Labels {
-			art.Labels = append(art.Labels, label.Name)
-		}
-		for _, tag := range artifact.Tags {
-			art.Tags = append(art.Tags, tag.Name)
-		}
-		arts = append(arts, art)
 
-		// For Harbor v2 clients, it has to append the accessory objects behind the subject artifact it has.
-		for _, acc := range artifact.Accessories {
-			art := &model.Artifact{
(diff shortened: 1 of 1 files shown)


chlins (approved):
lgtm

src/pkg/reg/adapter/harbor/v2/client.go
```diff
@@ -55,34 +55,56 @@ func (c *client) listArtifacts(repo string) ([]*model.Artifact, error) {
...
-			art.Labels = append(art.Labels, label.Name)
-		}
-		for _, tag := range artifact.Tags {
-			art.Tags = append(art.Tags, tag.Name)
-		}
-		arts = append(arts, art)
 
-		// For Harbor v2 clients, it has to append the accessory objects behind the subject artifact it has.
+	// append the accessory objects behind the subject artifact if it has.
+	var appendAcc = func(artifact *artifact.Artifact) error {
```
heww, line 60: move it out as a method of this struct,  eg `getAccessoriyArtifacts`

heww (approved):

PR #16421, opened 2022-02-26
jobservice/job: fix dropped error

alrs:
This fixes a dropped `err` variable in `src/jobservice/job`.

Signed-off-by: Lars Lehtonen <lars.lehtonen@gmail.com>

### DIFF
--- a/src/jobservice/job/tracker.go
+++ b/src/jobservice/job/tracker.go
@@ -232,13 +232,14 @@ func (bt *basicTracker) CheckIn(message string) error {
 
 	bt.refresh(current, message)
 	err := bt.fireHookEvent(current, message)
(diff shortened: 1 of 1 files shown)


src/jobservice/job/tracker.go
```diff
@@ -232,13 +232,14 @@ func (bt *basicTracker) CheckIn(message string) error {
 
 	bt.refresh(current, message)
 	err := bt.fireHookEvent(current, message)
```
steven-zou, line 234: The consideration here is the status hook event sending error should not block the following status update because hook failure can be repaired by the reaper process which depends on the status data updated here.

@alrs and also cc @wy65701436

PR #16476, opened 2022-03-06
chore: remove deplicate sql

hyy0322:
Remove duplicate sql exist in 0053_2.2.3_schema.up.sql


wy65701436:
The duplication is to cover all the upgrade path. Close it, and thanks for the contribution.

Issue #16455, opened 2022-03-03
Replace harbor-core redis library.


chlins:
Replace harbor core underlying redis library from https://github.com/gomodule/redigo to https://github.com/go-redis/redis because the latter has better performance as well as activity.

PR #16492, opened 2022-03-09
refactor: import go-redis to core as replacement of redigo

chlins:
Closes: https://github.com/goharbor/harbor/issues/16455


src/lib/orm/orm.go
```diff
@@ -88,6 +97,12 @@ func Clone(ctx context.Context) context.Context {
 	return NewContext(ctx, orm.NewOrm())
 }
 
+// DeepClone returns new context with orm and value from parent context but no
+// linkage of parent.
+func DeepClone(ctx context.Context) context.Context {
```
stonezdj, line 102: DeepClone may be inappropriate, it just copies some values, not all values.

stonezdj (approved):

src/lib/orm/orm.go
chlins: `CloneValue` is better?

src/lib/cache/redis/redis.go
```diff
@@ -107,23 +91,45 @@ func New(opts cache.Options) (cache.Cache, error) {
...
-	name := fmt.Sprintf("%x", sha256.Sum256([]byte(opts.Address)))
-
-	param := &libredis.PoolParam{
-		PoolMaxIdle:           100,
-		PoolMaxActive:         1000,
-		PoolIdleTimeout:       10 * time.Minute,
-		DialConnectionTimeout: time.Second,
-		DialReadTimeout:       time.Second * 2,
-		DialWriteTimeout:      time.Second * 5,
+	u, err := url.Parse(opts.Address)
```
wy65701436, line 94: please comment the defaults values in go-redis.

src/lib/cache/redis/redis.go
```diff
@@ -107,23 +91,45 @@ func New(opts cache.Options) (cache.Cache, error) {
...
-	return &Cache{opts: &opts, pool: pool}, nil
+	if t := values.Get("idle_timeout_seconds"); t != "" {
+		values.Del("idle_timeout_seconds")
+		values.Set("idle_timeout", t)
+		u.RawQuery = values.Encode()
+	}
+
+	var client *redis.Client
+
+	switch u.Scheme {
```
wy65701436, line 113: since newer version redis supports username & password in the url, not sure whether the error raised here contains the sensitive message.

src/lib/cache/redis/redis.go
```diff
@@ -107,23 +91,45 @@ func New(opts cache.Options) (cache.Cache, error) {
 		opts.Address = "redis://localhost:6379/0"
```
wy65701436, line 91: since we hardcode 0 here, do we mean that only core can use it without specify address?

src/lib/cache/redis/util_test.go
```diff
@@ -0,0 +1,40 @@
...
+package redis
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseSentinelURL(t *testing.T) {
```
wy65701436, line 24: please add more test cases to cover others parameters.

src/lib/cache/redis/redis.go
```diff
@@ -107,23 +91,45 @@ func New(opts cache.Options) (cache.Cache, error) {
...
+		client = redis.NewClient(rdbOpts)
+	case cache.RedisSentinel:
+		failoverOpts, err := ParseSentinelURL(u.String())
+		if err != nil {
+			return nil, err
+		}
+
+		client = redis.NewFailoverClient(failoverOpts)
+	default:
+		return nil, fmt.Errorf("redis: invalid URL scheme: %s", u.Scheme)
```
wy65701436, line 129: use lib/errors for instead for this PR.

src/lib/cache/redis/util.go
```diff
@@ -0,0 +1,187 @@
...
+	"strings"
+	"time"
+
+	"github.com/go-redis/redis/v8"
+)
+
+// ParseSentinelURL parses sentinel url to redis FailoverOptions.
+// It's a modified version of go-redis ParseURL(https://github.com/go-redis/redis/blob/997118894af9d4244d4a471f2b317eead9c9ca62/options.go#L222) because official version does
+// not support parse sentinel mode.
+func ParseSentinelURL(redisURL string) (*redis.FailoverOptions, error) {
```
wy65701436, line 31: Can we raise anPR to upstream? But it's ok to merge it.

wy65701436 (approved):
lgtm

src/lib/cache/redis/util.go
chlins: ok

src/lib/cache/redis/redis.go
chlins: updated

src/lib/cache/redis/util_test.go
chlins: updated.

src/lib/cache/redis/redis.go
chlins: yes, but in normal the address will not be empty.

src/lib/cache/redis/redis.go
chlins: the parse function will never print any username/password related fields.

src/lib/cache/redis/redis.go
chlins: updated.

src/lib/orm/orm.go
chlins: let's use `Copy`.

PR #16506, opened 2022-03-10
add transaction for artifact delete

wy65701436:
Add transaction for artifact deletion, given API has the transaction when to call artifact controller but other object may not, for example jobservice job.
Here, force add the tx to ensure all the things can be rolled back.


chlins (approved):
lgtm

PR #16510, opened 2022-03-11
fix: check the existence of the tag before updating pull time

chlins:
Check tag before update pull time, there are two purposes:

1. avoid noise log in the caller.("failed to update pull time, tag 0 not found")
2. reduce the db connection because `GetTag` needs query tag from db.


stonezdj (approved):

wy65701436 (approved):
lgtm

PR #16549, opened 2022-03-18
enhance health validataion

wy65701436:

### DIFF
--- a/src/pkg/scan/dao/scanner/model.go
+++ b/src/pkg/scan/dao/scanner/model.go
@@ -16,10 +16,9 @@ package scanner
 
 import (
 	"encoding/json"
-	"net/url"
-	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


src/pkg/scan/dao/scanner/model.go
```diff
@@ -105,10 +106,11 @@ func (r *Registration) Validate(checkUUID bool) error {
 		return errors.New("missing registration name")
 	}
 
-	err := checkURL(r.URL)
+	url, err := lib.ValidateHTTPURL(r.URL)
```
reasonerjt, line 109: if the unexported func `checkURL` still needed?

wy65701436: ok, let me check. If no use, I'll remove it.

heww (approved):

reasonerjt (approved):

Issue #16408, opened 2022-02-24
List user groups API should support  group_name query string


AllForNothing:
For 2.4.x and 2.5.x, the list user groups API does not support group_name query string, we need to support this so that users can search user groups by group_name
Both front and back end need to be adjusted for this

PR #16580, opened 2022-03-24
Update usergroups API to support search by group_name

stonezdj:
Add group_name parameter to GET usergroups API
  Related to issue #16408
  UI should also switch from search api to usergroups list api

### DIFF
--- a/src/server/v2.0/handler/usergroup.go
+++ b/src/server/v2.0/handler/usergroup.go
@@ -110,6 +110,9 @@ func (u *userGroupAPI) ListUserGroups(ctx context.Context, params operation.List
 	if err != nil {
 		return u.SendError(ctx, err)
 	}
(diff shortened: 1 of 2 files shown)


src/server/v2.0/handler/usergroup.go
```diff
@@ -110,6 +110,9 @@ func (u *userGroupAPI) ListUserGroups(ctx context.Context, params operation.List
 	if err != nil {
 		return u.SendError(ctx, err)
 	}
+	if params.GroupName != nil && len(*params.GroupName) > 0 {
```
wy65701436, line 113: https://github.com/goharbor/harbor/blob/main/src/server/v2.0/handler/replication.go#L187

Empty should an condition?

wy65701436 (approved):
lgtm

chlins (approved):
lgtm

PR #16585, opened 2022-03-25
Remove state restrictions for gc log button

### LINKED ISSUE #16339: GC log button can be shown not just for final status
Currently, the log button only shows on the GC complete. We should provide it once UI got the execution id, then user can see the process log of GC when it's running.

AllForNothing:
Fixes #16339


wy65701436 (approved):
lgtm

YangJiao0817 (approved):

Issue #16448, opened 2022-03-02
Metrics suddenly doesn't work


jorgemansillacriado:
If you are reporting a problem, please make sure the following information are provided:

Expected behavior and actual behavior:
curl localhost:9200/metrics suddenly doesn't work

Steps to reproduce the problem:

Versions:
Please specify the versions of following systems.

harbor version: 2.x
docker engine version: 17.12.1-ce
docker-compose version: 1.18.0
Additional context:

I don't see logs with error

my configuration is:

chlins:
@jorgemansillacriado Could you fill your detail information?

jorgemansillacriado:
Hi

exporter container is running even I can check metrics for components but work well
`curl localhost:9090/metrics?comp=core // work successfully`

``curl localhost:9090/metrics`   // no work
An error has occurred while serving metrics:

collected metric "harbor_project_total" { label:<name:"public" value:"true" > gauge:<value:4 > } was collected before with the same name and label values

chlins:
It's not work when you don't carry comp parameter? I didn't reproduce your problem in my environment, and could you share more configurations?

jorgemansillacriado:
> It's not work when you don't carry comp parameter? I didn't reproduce your problem in my environment, and could you share more configurations?

Hi what configuration do you need??
metric:
  enabled: true
  port: 9090
  path: /metrics

jorgemansillacriado:
> curl localhost:9090/metrics?comp=core

it's straing because when I install registry metrics works well but after some hours it levave of work

flaviomoringa:
I had already opened a similar issue here:
https://github.com/goharbor/harbor/issues/16057

And there is another similar one here:
https://github.com/goharbor/harbor/issues/16312

So maybe it should really be looked into

chlins:
Do you have any error log got from the exporter?

Lzhang-hub:
@flaviomoringa @chlins 
I have the same issue, is there any update?

------------------
exporter version: v2.5.0

this is exporter log:
```
Appending internal tls trust CA to ca-bundle ...
find: '/etc/harbor/ssl': No such file or directory
Internal tls trust CA appending is Done.
2022-03-25T07:56:59Z [INFO] [/common/dao/base.go:66]: Registering database: type-PostgreSQL host-aicloud-harbor-database port-5432 database-registry sslmode-"disable"
2022-03-25T07:56:59Z [INFO] [/common/dao/base.go:71]: Register database completed
[...]

PR #16596, opened 2022-03-28
fix: validate project metadata public value

chlins:
Fixes: https://github.com/goharbor/harbor/issues/16448

### DIFF
--- a/src/server/v2.0/handler/project.go
+++ b/src/server/v2.0/handler/project.go
@@ -147,6 +147,13 @@ func (a *projectAPI) CreateProject(ctx context.Context, params operation.CreateP
 		req.Metadata.Public = strconv.FormatBool(false)
 	}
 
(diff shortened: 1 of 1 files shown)


src/server/v2.0/handler/project.go
```diff
@@ -147,6 +147,13 @@ func (a *projectAPI) CreateProject(ctx context.Context, params operation.CreateP
 		req.Metadata.Public = strconv.FormatBool(false)
 	}
 
+	// validate metadata.public value, should only be "true" or "false"
```
wy65701436, line 150: use type assertion?

wy65701436 (approved):
lgtm

zyyw (approved):

src/server/v2.0/handler/project.go
chlins: we only support two value, true or false.

PR #16609, opened 2022-03-30
Fix install failed by custom bundle path

nobiit:


MinerYang:
Hi @nobiit ,

I assume you are using this when the install files and other bundle files(including harbor.yml,, etc) are in different directories,
I just wondering why we need to put these file separately since you have to download the whole offline-installer bundle when you installing a new version harbor?

And for the migration and upgrade scenario, we suggest to load the whole offline installer package with new tag, not just replace the values.yml or part of these files. 

you can refer to this doc [Upgrade Harbor and Migrate Data](https://goharbor.io/docs/1.10/administration/upgrade/)
If there's any scenarios that I haven't covered yet, please point me out.

Best.

nobiit:
I'm trying to deploy on Swam cluster, so I need to run DIND, and I want to package Harbor into a single image.
So, I want the config and the base file (installation package) to be in 2 different folders.

MinerYang:
> I'm trying to deploy on Swam cluster, so I need to run DIND, and I want to package Harbor into a single image. So, I want the config and the base file (installation package) to be in 2 different folders.

We understand, but this is not a common case for harbor users. I strongly suggest you to fork your own repository and customize the functionality you want.

Thanks for your effort!

wy65701436:
Close it per @MinerYang 's comment. If you still think it's a valid requiret, please file discussion on it.

PR #16610, opened 2022-03-30
Use list user groups API to search groups

### LINKED ISSUE #16408: List user groups API should support  group_name query string

AllForNothing:
Fixes #16408


YangJiao0817 (approved):

wy65701436 (approved):
lgtm

Issue #16633, opened 2022-04-01
Automated release notes


OrlinVasilev:
**Is your feature request related to a problem? Please describe.*
The current release notes are not complete and we can do better

**Describe the solution you'd like**
Introduce proper labels and rules for writing new PRs so we can generate Release notes automagicaly
Add release.yml to ease out a bit the process

labels to be used in the future:
For release notes:
release-note/enhancement - make something better
release-note/new-feature - completely new feature
release-note/update - update current component - e.g. Bump Golang version
release-note/breaking-change - breaking changes
release-note/docs - docs update
release-note/community - change related to community 
release-note/deprecation - deprecation change

To be skipped:
release-note/ignore-for-release- no need to be added into release notes

**Describe the main design/architecture of your solution**
Once we have proper setup for the PRs with Labels and good description it will be just click of a button to create release notes!
Add GitHub  Action on PR creation to check if you have at least one label selected - mheap/github-action-required-labels@v1

**Describe the development plan you've considered**
Once we have that implemented we can invoke it from actions or manual on new tag/release created!

New release notes can look like:
-------------------------------------------------
Here is the example repo: 
https://github.com/OrlinVasilev/release/releases/tag/v0.1

-------------------------------------------------

### Exciting New Features 🎉
* add new feature feature by @OrlinVasilev in https://github.com/OrlinVasilev/release/pull/1
### Component updates ⬆️
* update workflow by @OrlinVasilev in https://github.com/OrlinVasilev/release/pull/5
### Breaking Changes 🛠
* Breaking Change for demo #6 by @OrlinVasilev in https://github.com/OrlinVasilev/release/pull/7
### Deprecations ❌
* deprecation #2 by @OrlinVasilev in https://github.com/OrlinVasilev/release/pull/4
### Other Changes
* add action for label check by @OrlinVasilev in https://github.com/OrlinVasilev/release/pull/3

## New Contributors
* @OrlinVasilev made their first contribution in https://github.com/OrlinVasilev/release/pull/1

**Full Changelog**: https://github.com/OrlinVasilev/release/commits/v0.1

a-mccarthy:
@OrlinVasilev we should also update the docs with this new process and what the labels mean. Maybe in the contributing guide? https://github.com/goharbor/harbor/blob/main/CONTRIBUTING.md#push-and-create-pr

a-mccarthy:
[...]

PR #16634, opened 2022-04-01
Add release.yml to automate release notes #16633

OrlinVasilev:
Use github autogeneration release notes and add action to check if the PR has labels

Complete and more readable relase notes
https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes
Signed-off-by: OrlinVasilev <ovasilev@vmware.com>


.github/release.yml
```diff
@@ -0,0 +1,36 @@
...
+
+    - title: Breaking Changes 🛠
+      labels:
+        - release-note/breaking-change
+
+    - title: Deprecations ❌
+      labels:
+        - release-note/deprecation
+
+    - title: Other Changes
```
Vad1mo, line 34: 👍 like that we can adapt and change as needed.

Vad1mo (approved):
Great idea @OrlinVasilev 🚀 , this helps offload some repetitive tasks and give the release notes a constant structure without overhead!

.github/release.yml
```diff
@@ -0,0 +1,36 @@
...
+    - title: Enhancement 🚀
+      labels:
+        - release-note/enhancement
+    - title: Component updates ⬆️
+      labels:
+        - release-note/update
+    - title: Docs update 🗄️
+      labels:
+        - release-note/docs
+    - title: Community update 🧑🏻‍🤝‍🧑🏾
```
a-mccarthy, line 22: I'd like to make sure i understand can you provide an example of what type of PR would get this label? 

And how, does a maintainer/community member know when to apply this label?

.github/release.yml
```diff
@@ -0,0 +1,36 @@
...
+    - title: Exciting New Features 🎉
+      labels:
+        - release-note/new-feature
+    - title: Enhancement 🚀
+      labels:
+        - release-note/enhancement
+    - title: Component updates ⬆️
+      labels:
+        - release-note/update
+    - title: Docs update 🗄️
```
a-mccarthy, line 19: @OrlinVasilev I'm not sure if we need the docs update feature here. Or, rather, i dont think this will end up pulling in anything from this repo, bc all docs change PRs are made in the website repo. 

Is there a way to make this automation across repos? 

Does it make sense to plan on adding in docs updates to release notes manually for now, until we can figure out the automation?

OrlinVasilev: that's just a label definition which will help us in the future, cross repo is something that I started looking today will continue on Monday, but in general we should have as many as possible ways not to miss any change or update!

a-mccarthy:
this is great @OrlinVasilev! thanks for putting this together. I had 2 questions around process for docs and community above.

I wrote up some more details around the process below, more for my own understanding :) 

How this would work on an PR
* A user opens a new PR
* Once a PR is submitted, a github action will run to check if a PR has a release label
* The user or maintainer adds the correct release label to the PR (before or after the PR is created)
* Having a release label is not required before merging a PR. The github action that check labels only flags that the PR has a failed check of missing a release label.
* This check will run whenever a PR is opened, or a label is added or removed from the PR

At release time:
* When it comes time to release, a maintainer will create the release and click the built in autogenerate release option (following the steps [here](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes))
* Github will create release notes which consist of
  * The title of the PR will be pulled into release notes
  * The Github handle of the user who opened the PR will be pulled into release notes 

This this correct?

Some process questions:
* At what point in the process do we make sure the PR title is ok for the release notes? When reviewing the PR? at release time, when the release notes are generated?

.github/release.yml
OrlinVasilev: We will add descriptions in the contributors guide in which cases which labels to be used, but I think the labels are quite self explanatory, the docs and community labels are in case we wanna do some cross repo, to include changes from community and website repos, which BTW are something else that we need to fix (future ISSUE) :)

OrlinVasilev:
@a-mccarthy 
only one thing : Having a release label is not required before merging a PR. The github action that check labels only flags that the PR has a failed check of missing a release label.

Due to the added check, labels will be must on creating PRs otherwise the check will fail and the PR won't be able to merge!
We need to enforce that labeling to make sure: we do not forget label, we have all changes that needs to be added

As a next step we have to work on the PR Titles and to apply some good practices there as well. The title and description needs to be be part of the review process. PR should NOT be approved without well written title and description!

a-mccarthy:
@OrlinVasilev thanks for the clarification! Do you mind creating an issue with the next steps and what else we should update? (if you dont have one already) We should also make sure to bring up the PR and review requirements in a community meeting and over slack, etc.

a-mccarthy (approved):
lgtm! thanks @OrlinVasilev

OrlinVasilev:
I would love to merge that so we have 2.6 with all that in place :)

Vad1mo (approved):

OrlinVasilev:
BTW All labels are created so just need to assign them :)

PR #16642, opened 2022-04-03
Use exec in registryctl so signals are passed properly

mac-chaffee:
When containers are terminated in Kubernetes, Kubernetes will send a SIGTERM to PID1, wait for terminationGracePeriodSeconds, then SIGKILL it.

But since PID1 in registryctl is `bash`, that SIGTERM signal is not automatically passed to the registryctl process, forcing Kubernetes to wait the whole terminationGracePeriodSeconds to update the registry Pod.

This PR fixes that issue by using `exec`, which makes `registryctl` itself be PID1 so that it receives the SIGTERM signal.

Wasn't able to test this locally, I think the build system needs some TLC, especially for testing individual components. Like for example, I need a local copy of `registry` to build the registryctl image which is downloaded from `https://storage.googleapis.com/harbor-builds/bin/registry/release-v2.8.0-patch-redis/registry`, but that URL 404s.

### DIFF
--- a/make/photon/registryctl/start.sh
+++ b/make/photon/registryctl/start.sh
@@ -10,4 +10,4 @@ set -e
 
 /home/harbor/install_cert.sh
 
-/home/harbor/harbor_registryctl -c /etc/registryctl/config.yml
+exec /home/harbor/harbor_registryctl -c /etc/registryctl/config.yml


mac-chaffee:
Friendly bump :)

Vad1mo (approved):
thx, this was super annoying.

wy65701436 (approved):
lgtm

mac-chaffee:
So it seems that code for handing graceful shutdown might be bugged: https://github.com/goharbor/harbor/blob/77df4d92fd3c56baea5b780db3e0e84212c06cd3/src/registryctl/main.go#L72

When registryctl receives SIGTERM, it just logs the following but doesn't actually shut down:

```
2022-11-10T23:10:17Z [INFO] [/registryctl/main.go:68]: Got an interrupt, shutting down...
2022-11-10T23:10:17Z [INFO] [/registryctl/main.go:72]: Registry controller is shut down properly
```

The bug isn't obvious to me as a golang novice, but is there some context we're not cancelling or a goroutine still running or something?

PR #16658, opened 2022-04-06
skip policy check on pull cosign signature

wy65701436:
When user enables the cosign policy and triggers the replication, the harbor adapter will try to  pull the cosign siguature if it has to do the further push.
In this case, it has to skip policy check.


chlins (approved):
lgtm

heww (approved):

PR #16661, opened 2022-04-06
Update/improve grafana dashboard

mac-chaffee:
The existing grafana dashboard had some issues which I have fixed in this version:

* Something was causing the graphs not to render at all until I went in and executed each query. That appears to be fixed now by just re-generating the dashboard.
* Converted the "component up status" into a State Timeline
* Converted quota usage into a pie chart
* Fixed some spelling/grammar mistakes
* Migrated to the new timeseries graph type instead of the legacy graph type
* Added units where appropriate
* Set better legends
* For the Golang metrics, added filters to only show Harbor (instead of every golang exported in the whole cluster)
* Removed some panels that were pulling metrics that no longer existed
* Removed some panels that showed the same data but just at different quantiles
* Minimized the json. This is helpful for inclusion in ConfigMaps which is how the grafana helm chart works.

New images:
Screen Shot 2022-04-06 at 11 42 11 AM
Screen Shot 2022-04-06 at 11 48 25 AM
Screen Shot 2022-04-06 at 11 48 34 AM


a-mccarthy:
@mac-chaffee Thanks for updating this! I have a question, is it worth mentioning which version of Granfana or Harbor was used to create/test this example? I'm not very familiar with Grafana, so i'm not sure if that matters.

I'm thinking it might be useful to create a README file for this folder that gives more information about what this example is and why someone would want to use it. For example, point folks to the the docs site for more info on setting up metrics on their Harbor instance?  

Would be interested in hearing your thoughts :)

cc: @OrlinVasilev

a-mccarthy:
@mac-chaffee The readme looks awesome, thanks so much for adding!

OrlinVasilev:
@goharbor/maintainers  please review!

chlins (approved):
@mac-chaffee Awesome, thanks for your contribution!

wy65701436 (approved):
lgtm

chlins:
@mac-chaffee We'd better change the UID of harbor? otherwise, if I want to keep old dashboard and import this as new dashboard will conflict. And more question is I can't see the golang metrics of harbor component when import new dashboard.

chlins (commented):
Hold for update.

OrlinVasilev:
@mac-chaffee  are you able to provide the update?

chlins:
@mac-chaffee The dashboard is empty when I import it, but the data source is correct.

image

wy65701436:
@mac-chaffee can you see the problem as @chlins mentioned above? I'd like to merge this PR after you resolve it.

wy65701436 (approved):
lgtm

OrlinVasilev:
@chlins and @Vad1mo can you check this one please :)

wy65701436:
@mac-chaffee According to our debugging results, the dashboard does not present the expected results, can you help to verify this?

mac-chaffee:
I also can't get the dashboard to work properly when initially imported. Even parameterizing the data source didn't fix the issue; the dashboards stay blank until edited and re-saved (which doesn't actually change the json).

chlins:
> I also can't get the dashboard to work properly when initially imported. Even parameterizing the data source didn't fix the issue; the dashboards stay blank until edited and re-saved (which doesn't actually change the json).

@mac-chaffee If we can not resolve the issue, I think we need to pending this PR until solving it because currently the user experience is not friendly, from user side I think we should get the proper dashboards after import it like other components instead of doing additional data.

rgarcia6520:
> > I also can't get the dashboard to work properly when initially imported. Even parameterizing the data source didn't fix the issue; the dashboards stay blank until edited and re-saved (which doesn't actually change the json).
> 
> @mac-chaffee If we can not resolve the issue, I think we need to pending this PR until solving it because currently the user experience is not friendly, from user side I think we should get the proper dashboards after import it like other components instead of doing additional data.

@mac-chaffee @chlins Testing this newer Dashboard it is more consistent at not loading blank panels compared to what's in `main`. Also when troubleshooting the dashboard that's in `main` I noticed it was more reliable at showing data immediately when the `expr` keys in panels were closed out with `{}` eg: `harbor_project_member_total{}`

bdols:
this dashboard works for me on the first import attempt with the "importing via panel json" method with grafana 9.3.8 provided by kube-prometheus-stack.

OrlinVasilev (approved):

wy65701436:
@chlins could you please have a double check on this change?

OrlinVasilev:
@chlins any update on that we have this PR for too long now

chlins:
Verification passed on the latest Grafana, will merge it after CI passed, thanks for your contribution! @mac-chaffee 
image

cc @OrlinVasilev @wy65701436

OrlinVasilev:
Hurray!!!!!!!!!!!! @chlins thank you so much!!!!!!

OrlinVasilev:
@mac-chaffee congrats on finally merging that :)))

PR #16670, opened 2022-04-07
Add CODEOWNERS all maintainers

OrlinVasilev:
When we PR comes in all will be assigned as reviewers so we will not miss PR

Signed-off-by: OrlinVasilev <ovasilev@vmware.com>


Vad1mo (approved):
why not @goharbor/maintainers?

wy65701436 (approved):
lgtm

Issue #16540, opened 2022-03-16
Task: Postgres SQL schema file changes to create System artifact tables


prahaladdarkin:
Create Postgres schema file changes to create System artifact tables. This issue is part of the engineering epic for System Artifact Manager development:

https://github.com/goharbor/harbor/issues/16525

prahaladdarkin:
PR - https://github.com/goharbor/harbor/pull/16646

Issue #16542, opened 2022-03-16
Task: DAO layer implementation for System Artifact Manager entities


prahaladdarkin:
This issue is part of epic https://github.com/goharbor/harbor/issues/16525

prahaladdarkin:
PR- https://github.com/goharbor/harbor/pull/16646

prahaladdarkin:
Completed merge. Closing issue

Issue #16541, opened 2022-03-16
Task:  Model struct definition for System Manager entities


prahaladdarkin:
Create struct definitions for the system artifact manager database tables. This issue is part of epic : https://github.com/goharbor/harbor/issues/16525

prahaladdarkin:
PR - https://github.com/goharbor/harbor/pull/16646

prahaladdarkin:
Completed merge. Closing issue

PR #16678, opened 2022-04-09
System Artifact Manager - database and model changes

prahaladdarkin:
Implementation: database schema creation, model definitions, DAO layer and associated tests

Closes:
https://github.com/goharbor/harbor/issues/16540
https://github.com/goharbor/harbor/issues/16541
https://github.com/goharbor/harbor/issues/16542

Signed-off-by: prahaladdarkin <prahaladd@vmware.com>


make/migrations/postgresql/0080_2.5.0_schema.up.sql
```diff
@@ -32,3 +32,21 @@ ALTER TABLE report_vulnerability_record ALTER COLUMN id TYPE BIGINT;
 ALTER SEQUENCE report_vulnerability_record_id_seq AS BIGINT;
 
 CREATE INDEX IF NOT EXISTS idx_task_job_id ON task (job_id);
+
```
chlins, line 35: Why this script in the 2.5.0 migration files, should not 2.6.0?

src/pkg/systemartifact/cleanupcriteria.go
```diff
@@ -0,0 +1,45 @@
...
+	"context"
+	"github.com/goharbor/harbor/src/jobservice/logger"
+	"github.com/goharbor/harbor/src/lib/q"
+	"github.com/goharbor/harbor/src/pkg/systemartifact/dao"
+	"github.com/goharbor/harbor/src/pkg/systemartifact/model"
+	"time"
+)
+
+var (
+	DefaultCleanupWindowSeconds = 86400
```
chlins, line 13: Do we need to consider exposing this config to user?

chlins, line 13: Do we need to consider exposing this config to user?

src/pkg/systemartifact/cleanupcriteria.go
```diff
@@ -0,0 +1,45 @@
...
+	daoLayer dao.DAO
+}
+
+func (cleanupCriteria *defaultCleanupCriteria) List(ctx context.Context) ([]*model.SystemArtifact, error) {
+
+	currentTime := time.Now()
+	duration := time.Duration(DefaultCleanupWindowSeconds) * time.Second
+	timeRange := q.Range{Max: currentTime.Add(-duration).Format(time.RFC3339)}
+	logger.Infof("Cleaning up system artifacts with range: %v", timeRange)
+	query := q.Query{Keywords: map[string]interface{}{"create_time": &timeRange}}
```
chlins, line 43: q.New()?

chlins, line 43: q.New()

src/pkg/systemartifact/dao/dao.go
```diff
@@ -0,0 +1,107 @@
+package dao
+
+import (
+	"context"
+	"github.com/goharbor/harbor/src/lib/orm"
+)
+import "github.com/goharbor/harbor/src/pkg/systemartifact/model"
```
chlins, line 7: Please use go imports to format.

chlins, line 7: Please use go imports to format.

src/pkg/systemartifact/manager.go
```diff
@@ -0,0 +1,231 @@
...
+	artifacts, err := mgr.dao.List(ctx, &listQuery)
+
+	if err != nil {
+		return int64(0), err
+	}
+
+	size := int64(0)
+
+	for _, artifact := range artifacts {
+		size += artifact.Size
```
chlins, line 156: We need to consider potential performance issue if it has a lot of artifacts, `List` will select * from database but we only need size column, maybe could use SUM SQL to aggregate result?

chlins, line 156: We need to consider potential performance issue if it has a lot of artifacts, `List` will select * from database but we only need size column, maybe could use SUM SQL to aggregate result?

src/pkg/systemartifact/manager.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	return size, nil
+}
+
+func (mgr *systemArtifactManager) RegisterCleanupCriteria(vendor string, artifactType string, criteria CleanupCriteria) {
+	key := fmt.Sprintf(keyFormat, vendor, artifactType)
```
chlins, line 162: Need lock?

chlins, line 162: Need lock?

src/pkg/systemartifact/manager.go
```diff
@@ -0,0 +1,231 @@
...
+	mgr.cleanupCriteria[key] = criteria
+}
+
+func (mgr *systemArtifactManager) GetCleanupCriteria(vendor string, artifactType string) CleanupCriteria {
+	key := fmt.Sprintf(keyFormat, vendor, artifactType)
+	if criteria, ok := mgr.cleanupCriteria[key]; ok {
```
chlins, line 168: Need lock?

chlins, line 168: Need lock?

src/pkg/systemartifact/manager.go
prahaladdarkin: My bad. I somehow was under the incorrect assumption that when invoked in a package `init()` method, the access is implicitly synced. I have corrected my understanding and now used a lock

src/pkg/systemartifact/manager.go
prahaladdarkin: My bad. I somehow was under the incorrect assumption that when invoked in a package `init()` method, the access is implicitly synced. I have corrected my understanding and now used a lock

prahaladdarkin: My bad. I somehow was under the incorrect assumption that when invoked in a package `init()` method, the access is implicitly synced. I have corrected my understanding and now used a lock

prahaladdarkin (commented):
Incorporated all review comments

make/migrations/postgresql/0090_2.6.0_schema.up.sql
```diff
@@ -1,2 +1,18 @@
 /* Correct project_metadata.public value, should only be true or false, other invaild value will be rewrite to false */
-UPDATE project_metadata SET value='false' WHERE name='public' AND value NOT IN('true', 'false');
\ No newline at end of file
+UPDATE project_metadata SET value='false' WHERE name='public' AND value NOT IN('true', 'false');
+
+/*
+System Artifact Manager
+Github proposal link : https://github.com/goharbor/community/pull/181
+*/
+ CREATE TABLE IF NOT EXISTS "system_artifact" (
```
wy65701436, line 8: to keep code style consistent, could you use the raw attributie without double quotation marks？

make/migrations/postgresql/0090_2.6.0_schema.up.sql
```diff
@@ -1,2 +1,18 @@
...
+        "id" serial NOT NULL PRIMARY KEY,
+        "repository" text NOT NULL DEFAULT '' ,
+        "digest" text NOT NULL DEFAULT '' ,
+        "size" bigint NOT NULL DEFAULT 0 ,
+        "vendor" text NOT NULL DEFAULT '' ,
+        "type" text NOT NULL DEFAULT '' ,
+        "create_time" timestamp with time zone NOT NULL,
```
wy65701436, line 15: creation_time timestamp DEFAULT CURRENT_TIMESTAMP

make/migrations/postgresql/0090_2.6.0_schema.up.sql
```diff
@@ -1,2 +1,18 @@
...
+System Artifact Manager
+Github proposal link : https://github.com/goharbor/community/pull/181
+*/
+ CREATE TABLE IF NOT EXISTS "system_artifact" (
+        "id" serial NOT NULL PRIMARY KEY,
+        "repository" text NOT NULL DEFAULT '' ,
+        "digest" text NOT NULL DEFAULT '' ,
+        "size" bigint NOT NULL DEFAULT 0 ,
+        "vendor" text NOT NULL DEFAULT '' ,
+        "type" text NOT NULL DEFAULT '' ,
```
wy65701436, line 14: varchar is more appropriate?

make/migrations/postgresql/0090_2.6.0_schema.up.sql
```diff
@@ -1,2 +1,18 @@
...
\ No newline at end of file
+UPDATE project_metadata SET value='false' WHERE name='public' AND value NOT IN('true', 'false');
+
+/*
+System Artifact Manager
+Github proposal link : https://github.com/goharbor/community/pull/181
+*/
+ CREATE TABLE IF NOT EXISTS "system_artifact" (
+        "id" serial NOT NULL PRIMARY KEY,
+        "repository" text NOT NULL DEFAULT '' ,
```
wy65701436, line 10: varchar is more appropriate?

src/pkg/systemartifact/cleanupcriteria.go
```diff
@@ -0,0 +1,45 @@
...
+// provide a custom clean-up criteria. This allows producers of the
+// system artifact data to control the lifespan of the generated artifact
+// records and data.
+// Every system data artifact produces must register a cleanup criteria.
+
+type CleanupCriteria interface {
+	List(ctx context.Context) ([]*model.SystemArtifact, error)
+}
+
+var DefaultCleanupCriteria = NewCleanupCriteria()
```
wy65701436, line 27: Default()? then the caller will be cleanupcriteria.Default().

This rule could apply to other method name in this go file.

src/pkg/systemartifact/cleanupcriteria.go
```diff
@@ -0,0 +1,45 @@
...
+
+func NewCleanupCriteria() CleanupCriteria {
+	return &defaultCleanupCriteria{daoLayer: dao.NewSystemArtifactDao()}
+}
+
+type defaultCleanupCriteria struct {
+	daoLayer dao.DAO
```
wy65701436, line 34: dao is ok, IMO.

src/pkg/systemartifact/dao/dao.go
```diff
@@ -0,0 +1,134 @@
...
+	sizeQuery       = "select sum(size) as total_size from system_artifact"
+	totalSizeColumn = "total_size"
+)
+
+// DAO defines an data access interface for manging the CRUD and read of system
+// artifact tracking records
+type DAO interface {
+
+	// Create a system artifact tracking record.
+	Create(ctx context.Context, artifactRecord *model.SystemArtifact) (int64, error)
```
wy65701436, line 22: artifactRecord is a little confusing here since we have artifact table.

src/pkg/systemartifact/dao/dao.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// Get a system artifact tracking record identified by vendor, repository and digest
+	Get(ctx context.Context, vendor string, repository string, digest string) (*model.SystemArtifact, error)
```
wy65701436, line 25: Get(ctx context.Context, vendor, repository, digest string) (*model.SystemArtifact, error)?

src/pkg/systemartifact/dao/dao.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// Delete a system artifact tracking record identified by vendor, repository and digest
+	Delete(ctx context.Context, vendor string, repository string, digest string) error
```
wy65701436, line 28: ditto

src/pkg/systemartifact/dao/dao.go
```diff
@@ -0,0 +1,134 @@
...
+
+	sa := model.SystemArtifact{
+		Repository: repository,
+		Digest:     digest,
+		Vendor:     vendor,
+	}
+
+	_, err = ormer.Delete(&sa, "vendor", "repository", "digest")
+
+	return nil
```
wy65701436, line 99: why not return the err?

src/pkg/systemartifact/dao/dao.go
```diff
@@ -0,0 +1,134 @@
...
+
+	return systemArtifactRecords, nil
+}
+
+func (d *systemArtifactDAO) Size(ctx context.Context) (int64, error) {
+	ormer, err := orm.FromContext(ctx)
+	if err != nil {
+		return int64(0), err
+	}
+	var maps []orm.Params
```
wy65701436, line 124:
```
var totalSize int64
if err := o.Raw(sql).QueryRow(&totalSize); err != nil {
	return 0, err
}

return totalSize, nil
```

src/pkg/systemartifact/manager.go
```diff
@@ -0,0 +1,235 @@
...
+	if err != nil {
+		return int64(0), err
+	}
+
+	repoName := mgr.getRepositoryName(artifactRecord.Vendor, artifactRecord.Repository)
+	err = mgr.regCli.PushBlob(repoName, artifactRecord.Digest, artifactRecord.Size, reader)
+
+	if err != nil {
+
+		mgr.dao.Delete(ctx, artifactRecord.Vendor, artifactRecord.Repository, artifactRecord.Digest)
```
wy65701436, line 97: log error?

src/pkg/systemartifact/manager.go
```diff
@@ -0,0 +1,235 @@
...
+	repoName := mgr.getRepositoryName(vendor, repository)
+	err := mgr.regCli.DeleteBlob(repoName, digest)
+
+	if err != nil {
+		return err
+	}
+
+	err = mgr.dao.Delete(ctx, vendor, repository, digest)
+
+	return err
```
wy65701436, line 127: return mgr.dao.Delete(ctx, vendor, repository, digest)

src/pkg/systemartifact/manager.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+	return readCloser, nil
+}
+
+func (mgr *systemArtifactManager) Delete(ctx context.Context, vendor string, repository string, digest string) error {
+
+	repoName := mgr.getRepositoryName(vendor, repository)
+	err := mgr.regCli.DeleteBlob(repoName, digest)
+
+	if err != nil {
```
wy65701436, line 121: if err := **; err != nil {
  return err
}

src/pkg/systemartifact/manager.go
```diff
@@ -0,0 +1,235 @@
...
+	defer mgr.lock.Unlock()
+	mgr.lock.Lock()
+	if criteria, ok := mgr.cleanupCriteria[key]; ok {
+		return criteria
+	}
+	return DefaultCleanupCriteria
+}
+
+func (mgr *systemArtifactManager) Cleanup(ctx context.Context) (int64, int64, error) {
+	logger.Info("Starting system artifact cleanup")
```
wy65701436, line 168: to avoid noise in the log, should we use logger.Debug?

src/pkg/systemartifact/manager.go
```diff
@@ -0,0 +1,235 @@
...
+	key := fmt.Sprintf(keyFormat, vendor, artifactType)
+	defer mgr.lock.Unlock()
+	mgr.lock.Lock()
+	if criteria, ok := mgr.cleanupCriteria[key]; ok {
+		return criteria
+	}
+	return DefaultCleanupCriteria
+}
+
+func (mgr *systemArtifactManager) Cleanup(ctx context.Context) (int64, int64, error) {
```
wy65701436, line 167: seems that this method will never return err but nil, do we need this?

src/pkg/systemartifact/cleanupcriteria.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+// CleanupCriteria provides an interface that can be implemented
+// by consumers of the system artifact management framework to
+// provide a custom clean-up criteria. This allows producers of the
+// system artifact data to control the lifespan of the generated artifact
+// records and data.
+// Every system data artifact produces must register a cleanup criteria.
+
+type CleanupCriteria interface {
```
wy65701436, line 23: beside the defaultCriteria, can you give another example about the imple of CleanupCriteria?

src/pkg/systemartifact/cleanupcriteria.go
wy65701436: Please add some comments about default beharior

src/pkg/systemartifact/cleanupcriteria.go
wy65701436: And for the interface name, I would like to propose to be a straightforward thing, like:

Selector
CandidateSelector

src/pkg/systemartifact/manager.go
```diff
@@ -0,0 +1,235 @@
...
+	id, err := mgr.dao.Create(ctx, artifactRecord)
+
+	if err != nil {
+		return int64(0), err
+	}
+
+	repoName := mgr.getRepositoryName(artifactRecord.Vendor, artifactRecord.Repository)
+	err = mgr.regCli.PushBlob(repoName, artifactRecord.Digest, artifactRecord.Size, reader)
+
+	if err != nil {
```
wy65701436, line 95: we can do the storage operation firstly, then DB. By doing this, no need to call the delete here.

prahaladdarkin: The reason I had to adopt this sequence is to ensure that there are no stray blobs within the repository that do not have a corresponding tracking record - can happen if we were to encounter a DB error when inserting the record. 
This makes the job of clean-up simpler as we only have to delete records that do not have a blob really stored on the registry filesystem.
To the best of my knowledge, there is no job/background processing right now that would automatically clean up aging blobs from the disk.
Also by calling Delete immediately following a failure to write the BLOB, we are able to ensure that our system space calculations account only for the BLOBS that are **actually** present on the disk as opposed to orphaned stray records in the DB.

src/pkg/systemartifact/cleanupcriteria.go
prahaladdarkin: An example of a non-default cleanup criteria - consider a harbor component that creates the system artifact record of a particular type that must be valid for 7 days. In such a case, the component would register a custom implementation of the cleanup criteria where-in the `List()` method would return all system artifact records greater than 7 days.
I will add additional methods for filtering for `List()`

src/pkg/systemartifact/manager.go
prahaladdarkin: The current implementation of `Cleanup()` does not return any error. My earlier PR actually had an approach where the cleanup would stop at the first encountered error. However, this would be a problem, since we prevent the deletion of other system artifacts in case any one of the system artifacts could not be deleted (say for e.g. due to an incorrectly implemented cleanup criteria). Hence I only log the error
 `Cleanup()` is an interface method and there could be another implementation provided that would require fail fast and hence the error return.

src/pkg/systemartifact/manager.go
wy65701436: so, then we can use orm.WithTransaction(). When failure happens, the creation will be rolled back.

src/pkg/systemartifact/dao/dao.go
prahaladdarkin: Renamed to `systemArtifact` to be consistent with the model naming conventions and avoid any confusions

src/pkg/systemartifact/cleanupcriteria.go
prahaladdarkin: Done. I have renamed the method to `Default()`. I however could not get the second part of the comment `This rule could apply to other method name in this go file.`.

make/migrations/postgresql/0090_2.6.0_schema.up.sql
prahaladdarkin: Done

make/migrations/postgresql/0090_2.6.0_schema.up.sql
prahaladdarkin: Done

make/migrations/postgresql/0090_2.6.0_schema.up.sql
prahaladdarkin: Done

make/migrations/postgresql/0090_2.6.0_schema.up.sql
prahaladdarkin: Done

src/pkg/systemartifact/dao/dao.go
prahaladdarkin: Done

src/pkg/systemartifact/manager.go
prahaladdarkin: Done

src/pkg/systemartifact/manager.go
prahaladdarkin: Done

src/pkg/systemartifact/manager.go
prahaladdarkin: Done. I have now moved the artifact record creation and the corresponding write to the blob store as a closure to the `WithTransaction` method.

src/pkg/systemartifact/cleanupcriteria.go
prahaladdarkin: Renamed interface to `Selector`. Also, added additional interface methods to allow more filters to be specified by the clients.

src/pkg/systemartifact/manager.go
prahaladdarkin: Done

prahaladdarkin:
@wy65701436 I have incorporated all the review comments. Please take a look and let me know if the changes  can go ahead to master after the CICD run.
I will send the second set of changes based on the current changelist for calculations of space used by system and protection within the project layer that the `sys_harbor` name is not used.

src/pkg/systemartifact/manager.go
```diff
@@ -0,0 +1,239 @@
...
+	"github.com/goharbor/harbor/src/pkg/systemartifact/dao"
+	"github.com/goharbor/harbor/src/pkg/systemartifact/model"
+	"io"
+	"sync"
+)
+
+var (
+	Mgr              = NewManager()
+	keyFormat        = "%s:%s"
+	repositoryFormat = "sys_harbor/%s/%s"
```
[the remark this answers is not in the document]
prahaladdarkin: Ack. Done. I have instead added that as a `const` since the format is going to remain fixed for it's use within the code base.

chlins (approved):
lgtm

src/pkg/systemartifact/manager.go
```diff
@@ -0,0 +1,253 @@
...
+	// get a copy of the registered cleanup criteria and
+	// iterate through this copy to invoke the cleanup
+	registeredCriteria := make(map[string]Selector, 0)
+	mgr.lock.Lock()
+	for key, val := range mgr.cleanupCriteria {
+		registeredCriteria[key] = val
+	}
+	mgr.lock.Unlock()
+
+	for key, val := range registeredCriteria {
```
wy65701436, line 187: This will be better to understand?

```
for key, val := range registeredCriteria {
		logger.Infof("Executing cleanup for 'vendor:artifactType' : %s", key)
		deleted, size, err := mgr.cleanup(ctx, val)

		if err != nil {
			// one vendor error should not impact the clean-up of other vendor types. Hence the cleanup logic would continue
			// after logging the error
			logger.Errorf("Error when cleaning up system artifacts for 'vendor:artifactType':%s, %v", key, err)
                        continue
		}

		totalRecordsDeleted += deleted
		totalReclaimedSize += size
}
```

wy65701436 (approved):
lgtm

PR #16682, opened 2022-04-11
support acceleration service

wy65701436:
1, support acceleration service endpoints manage.
2, support auto convert.
3, support nydus as a new kinds of accessory.


OrlinVasilev:
Hey @wy65701436 can you add a bit more explanation :)

PR #17012, opened 2022-06-16
Unify the process of job schedule/task retrieve and update

stonezdj:
Change regular expression in robot account

Signed-off-by: stonezdj <stonezdj@gmail.com>

Thank you for contributing to Harbor!

# Comprehensive Summary of your change

# Issue being fixed
Fixes #(issue)

Please indicate you've done the following:
- [ ] Well Written Title and Summary of the PR
- [ ] Label the PR as needed. "release-note/ignore-for-release, release-note/new-feature, release-note/update, release-note/enhancement, release-note/community, release-note/breaking-change, release-note/docs, release-note/infra, release-note/deprecation"
- [ ] Accepted the DCO. Commits without the DCO will delay acceptance.
- [ ] Made sure tests are passing and test coverage is added if needed.
- [ ] Considered the docs impact and opened a new docs issue or PR with docs changes if needed in [website repository](https://github.com/goharbor/website).

PR #17117, opened 2022-07-04
fix updating robots with wildcard project access

elenz97:
Updating an existing robot account that has system-level access on all projects will fail in any case.
This was introduced in https://github.com/goharbor/harbor/pull/17012/files#diff-431ad1ebfab4b93f5a2e26dcc7462cb6805e4eb38726c30791f8e36891a276beR302-R306
Checking access to a project is only considerable when one or more projects are selected. If all projects are selected by wildcard, the check will fail as it takes the wildcard `*` as input for the project name.

### DIFF
--- a/src/server/v2.0/handler/robot.go
+++ b/src/server/v2.0/handler/robot.go
@@ -298,18 +298,17 @@ func (rAPI *robotAPI) updateV2Robot(ctx context.Context, params operation.Update
 	if err := rAPI.validate(params.Robot.Duration, params.Robot.Level, params.Robot.Permissions); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


Vad1mo:
@elenz97 DCO is missing, can you add that.

elenz97:
> @elenz97 DCO is missing, can you add that.

@Vad1mo Done, thanks for the reminder :+1:

wy65701436:
thanks @elenz97 for your contribution and the issue founding, it's a regression on v2.5.3 and fixed in the main branch by https://github.com/goharbor/harbor/pull/17248. 

Close it by fixed.

Issue #9728, opened 2019-11-02
Non-root permission installation failed


xuexiaoao:
### I failed to install version 1.9  offline.My account does not have root privileges. How do I install it?
Error message is:
```shell
Traceback (most recent call last):
  File "bin/docker-compose", line 6, in <module>
  File "compose/cli/main.py", line 71, in main
  File "compose/cli/main.py", line 124, in perform_command
  File "compose/cli/command.py", line 64, in project_from_options
  File "compose/cli/command.py", line 139, in get_project
  File "compose/config/config.py", line 404, in load
  File "compose/config/config.py", line 504, in load_services
  File "compose/config/config.py", line 483, in build_services
  File "compose/config/config.py", line 483, in <listcomp>
  File "compose/config/config.py", line 475, in build_service
  File "compose/config/config.py", line 881, in finalize_service
  File "compose/config/config.py", line 681, in resolve_environment
  File "compose/config/environment.py", line 43, in env_vars_from_file
  File "/code/.tox/py37/lib/python3.7/codecs.py", line 904, in open
PermissionError: [Errno 13] Permission denied: '/centos/harbor/common/config/registryctl/env'
```
File permissions is:
```shell
$ ls -alh
total 0
drwxr-xr-x 9 root  root  105 Nov  2 16:16 .
drwxr-xr-x 3 root  root   20 Nov  2 16:16 ..
drwxr-xr-x 3 root  root   53 Nov  2 16:16 core
drwxr-xr-x 2 root  root   17 Nov  2 16:16 db
drwxr-xr-x 2 10000 10000  35 Nov  2 16:16 jobservice
drwxr-xr-x 2 root  root   55 Nov  2 16:16 log
drwxr-xr-x 3 root  root   38 Nov  2 16:16 nginx
drwxr-xr-x 2 root  root   24 Nov  2 16:16 registry
drwxr-xr-x 2 root  root   35 Nov  2 16:16 registryctl

[ registryctl]$ ls -alh
total 8.0K
drwxr-xr-x 2 root root  35 Nov  2 16:16 .
drwxr-xr-x 9 root root 105 Nov  2 16:16 ..
-rw-r--r-- 1 root root 108 Nov  2 16:16 config.yml
-rw-r----- 1 root root  64 Nov  2 16:16 env
```

tvvignesh:
I have the same problem as well. Executing with root permissions worked though.

ferrandpaul:
Hello, I have exactly the same issue, 
[...]

PR #17208, opened 2022-07-19
fix issue #9728: check if script is being run by root 

freebased:


MinerYang:
Hi @freebased,

This install process can be conducted by not only root user (`$EUID eq 0`), but also by the users who has root permissions, for example `sudo ./install.sh`.

So I think it's not appropriate enough to do this check. We will get `permission denied` errors  if you don't have root permissions.

Issue #17552, opened 2022-09-15
Phase-1 dev for replication by chunk


chlins:
The phase-1 develop for replication by chunk, the scope of phase-1 is supporting copy by chunk in replication job and retry copy chunk when error occurred inside job self.

PR #17602, opened 2022-09-27
feat: introduce the copy by chunk for replication

chlins:
1. Add sql migration to alter replication policy table
2. Implement the PullBlobChunk and PushBlobChunk for the underlying v2 registry client
3. Update image transfer logic to support copy by chunk
4. Update the replication policy API handler

# Issue being fixed
Fixes https://github.com/goharbor/harbor/issues/17552


src/controller/replication/transfer/options.go
```diff
@@ -0,0 +1,43 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package transfer
+
+type Option func(*Options)
+
+type Options struct {
```
wy65701436, line 19: add descriptions for these options.

src/pkg/registry/client.go
```diff
@@ -371,6 +375,34 @@ func (c *client) PullBlob(repository, digest string) (int64, io.ReadCloser, erro
 	return size, resp.Body, nil
 }
 
+// PullBlobChunk pulls the specified blob, but by chunked
+func (c *client) PullBlobChunk(repository, digest string, blobSize int64, start, end int64) (int64, io.ReadCloser, error) {
```
wy65701436, line 379: can we add a link to the OCI spec for these two apis?

src/pkg/registry/client.go
```diff
@@ -371,6 +375,34 @@ func (c *client) PullBlob(repository, digest string) (int64, io.ReadCloser, erro
...
+	req, err := http.NewRequest(http.MethodGet, buildBlobURL(c.url, repository, digest), nil)
+	if err != nil {
+		return 0, nil, err
+	}
+
+	req.Header.Add("Accept-Encoding", "identity")
```
wy65701436, line 385: is it defined by spec?

src/pkg/registry/client.go
```diff
@@ -371,6 +375,34 @@ func (c *client) PullBlob(repository, digest string) (int64, io.ReadCloser, erro
...
+	req.Header.Add("Range", fmt.Sprintf("bytes=%d-%d", start, end))
```
wy65701436, line 386: we do not need to validate the `start` and `end`? what about if start > end?

src/lib/redis/instance.go
```diff
@@ -33,7 +33,7 @@ var (
 // Instance returns the redis instance.
 func Instance() *redis.Client {
 	_once.Do(func() {
-		url := os.Getenv("_REDIS_URL_REG")
+		url := os.Getenv("_REDIS_URL_CORE")
```
wy65701436, line 36: mistake?

src/pkg/registry/client.go
```diff
@@ -371,6 +375,34 @@ func (c *client) PullBlob(repository, digest string) (int64, io.ReadCloser, erro
...
+	resp, err := c.do(req)
+	if err != nil {
+		return 0, nil, err
+	}
+
+	var size int64
```
wy65701436, line 392: do we need to verify size equals `end - start`?

src/pkg/registry/client.go
```diff
@@ -371,6 +375,34 @@ func (c *client) PullBlob(repository, digest string) (int64, io.ReadCloser, erro
...
+		return 0, nil, err
+	}
+
+	req.Header.Add("Accept-Encoding", "identity")
+	req.Header.Add("Range", fmt.Sprintf("bytes=%d-%d", start, end))
+	resp, err := c.do(req)
+	if err != nil {
+		return 0, nil, err
+	}
+
```
wy65701436, line 391: defer resp.Body.Close() ?

src/controller/replication/transfer/image/transfer.go
```diff
@@ -282,36 +300,79 @@ func (t *transfer) copyBlobWithRetry(srcRepo, dstRepo, digest string, sizeFromDe
...
+		}
+		t.logger.Infof("will retry %v later", backoff)
+		time.Sleep(backoff)
+	}
+
+	return err
+}
+
+// preCopyBlob checks whether need to copy by check existence and mount, return true if need copy.
+func (t *transfer) preCopyBlob(srcRepo, dstRepo, digest string) (bool, error) {
```
wy65701436, line 329: it needs to rename this func.

wy65701436 (approved):
lgtm

src/pkg/registry/client.go
chlins: updated.

src/pkg/registry/client.go
chlins: Not found in spec, but keep same with copy by blob.

src/lib/redis/instance.go
chlins: updated

src/pkg/registry/client.go
chlins: Here we can not close it since in the next step, the resp need to input as parameter for push.

src/controller/replication/transfer/image/transfer.go
chlins: renamed to `tryMountBlob`.

src/pkg/registry/client.go
```diff
@@ -371,6 +375,34 @@ func (c *client) PullBlob(repository, digest string) (int64, io.ReadCloser, erro
...
+	// no content-length is acceptable, which can taken from manifests
+	if len(n) > 0 {
+		size, err = strconv.ParseInt(n, 10, 64)
+		if err != nil {
+			defer resp.Body.Close()
+			return 0, nil, err
+		}
+	}
+
+	return size, resp.Body, nil
```
ywk253100, line 403: Why we need to return the `size`? Seems it is not used by the caller

src/pkg/registry/client.go
```diff
@@ -379,6 +411,45 @@ func (c *client) PushBlob(repository, digest string, size int64, blob io.Reader)
...
+	var err error
+	// first chunk need to initialize blob upload location
+	if start == 0 {
+		location, _, err = c.initiateBlobUpload(repository)
+		if err != nil {
+			return location, err
+		}
+	}
+
+	lastChunk := end == blobSize
```
ywk253100, line 425: Does `start` start from 0? I mean that if a blob is only 2 bytes, what are the values of `start`, `end` and `blobSize`? Is `0`, `1`, `2` or `0`, `2`, `2`?

src/pkg/registry/client.go
chlins: Yes, not used for caller, just keep it unify with original copy by blob interface.

src/pkg/registry/client.go
chlins: Yes, the start should from 0, if the blob size is 2 bytes, and if the chunk size greater than 2 bytes, then the request will fallback to copy by blob, if the chunk size less than blob size, such as 1 byte, the 1st chunk is start(0), end(1), blobSize(2), the 2st chunk is start(2), end(2), blobSize(2).

ywk253100: Please double check this part as we discussed offline

ywk253100 (approved):

Issue #17643, opened 2022-10-12
httputil: ReverseProxy read error during body copy: unexpected EOF


wyb-2021:
"unexecution EOF" appears when I execute "Docker Pull".
![image](https://user-images.githubusercontent.com/114282105/195266946-0946bea5-23bd-4c2b-99b8-76cca38b91dd.png)
![image](https://user-images.githubusercontent.com/114282105/195266847-bee103c3-b04a-4f2e-ad16-75530db407b7.png)

**Log:**
2022/10/12 06:18:39 httputil: ReverseProxy read error during body copy: unexpected EOF
2022/10/12 06:18:39.817 [C] [panic.go:969]  the request url is  /v2/xxxx/blobs/sha256:46732a3021700b3826f4542dab40c877f34f9e4c2ab8d1a4bfecb39a1617b20b
2022/10/12 06:18:39.817 [C] [panic.go:969]  Handler crashed with error net/http: abort Handler
2022/10/12 06:18:39.817 [C] [panic.go:969]  /usr/local/go/src/runtime/panic.go:969
2022/10/12 06:18:39.817 [C] [panic.go:969]  /usr/local/go/src/net/http/httputil/reverseproxy.go:338
2022/10/12 06:18:39.817 [C] [panic.go:969]  /harbor/src/server/middleware/repoproxy/proxy.go:63
2022/10/12 06:18:39.817 [C] [panic.go:969]  /harbor/src/server/middleware/repoproxy/proxy.go:50
2022/10/12 06:18:39.817 [C] [panic.go:969]  /harbor/src/server/middleware/middleware.go:57
2022/10/12 06:18:39.817 [C] [panic.go:969]  /usr/local/go/src/net/http/server.go:2042
2022/10/12 06:18:39.817 [C] [panic.go:969]  /harbor/src/server/middleware/metric/metric.go:84
2022/10/12 06:18:39.817 [C] [panic.go:969]  /usr/local/go/src/net/http/server.go:2042
2022/10/12 06:18:39.817 [C] [panic.go:969]  /harbor/src/server/middleware/v2auth/auth.go:171
2022/10/12 06:18:39.817 [C] [panic.go:969]  /usr/local/go/src/net/http/server.go:2042
2022/10/12 06:18:39.817 [C] [panic.go:969]  /harbor/src/server/router/router.go:90
2022/10/12 06:18:39.817 [C] [panic.go:969]  /harbor/src/vendor/github.com/astaxie/beego/router.go:769
[...]

Issue #16556, opened 2022-03-18
Handler crash panic


stonezdj:
some times found panic in the harbor core log
```
OccurAt-2022-03-18 08:39:54
2022/03/18 08:39:54.670 [1;35m[C][0m [panic.go:969]  the request url is  /v2/solution_builder_staging/sb-e2eparallel2/edm/blobs/sha256:07b195254c6fbc4ab6f5c3acd79a86553ff2b4f0accc3864570b1610d7d8b9d8
2022/03/18 08:39:54.670 [1;35m[C][0m [panic.go:969]  Handler crashed with error net/http: abort Handler
2022/03/18 08:39:54.670 [1;35m[C][0m [panic.go:969]  /usr/local/go/src/runtime/panic.go:969
2022/03/18 08:39:54.670 [1;35m[C][0m [panic.go:969]  /usr/local/go/src/net/http/httputil/reverseproxy.go:338
2022/03/18 08:39:54.670 [1;35m[C][0m [panic.go:969]  /harbor/src/server/middleware/repoproxy/proxy.go:63
2022/03/18 08:39:54.670 [1;35m[C][0m [panic.go:969]  /harbor/src/server/middleware/repoproxy/proxy.go:50
2022/03/18 08:39:54.670 [1;35m[C][0m [panic.go:969]  /harbor/src/server/middleware/middleware.go:57
2022/03/18 08:39:54.670 [1;35m[C][0m [panic.go:969]  /usr/local/go/src/net/http/server.go:2042
2022/03/18 08:39:54.670 [1;35m[C][0m [panic.go:969]  /harbor/src/server/middleware/metric/metric.go:84
2022/03/18 08:39:54.670 [1;35m[C][0m [panic.go:969]  /usr/local/go/src/net/http/server.go:2042
2022/03/18 08:39:54.670 [1;35m[C][0m [panic.go:969]  /harbor/src/server/middleware/v2auth/auth.go:171
2022/03/18 08:39:54.670 [1;35m[C][0m [panic.go:969]  /usr/local/go/src/net/http/server.go:2042
2022/03/18 08:39:54.670 [1;35m[C][0m [panic.go:969]  /harbor/src/server/router/router.go:90
2022/03/18 08:39:54.670 [1;35m[C][0m [panic.go:969]  /harbor/src/vendor/github.com/astaxie/beego/router.go:769
[...]

PR #17716, opened 2022-10-29
fix: panic recover when stream

hunshcn:

when download a layer and the http request between core and registry is aborted, the ReverseProxy handler will panic. [net/http/httputil: ReverseProxy should use ErrAbortHandler on read error while copying body · Issue ](https://github.com/golang/go/issues/23643)

but beego will recover this panic and show a error page. so the layer will like this 

tar content

If this is a client which has auto-retry with broken-point continuingly-transferring, the content will be dirty.(include error html page), such as wget and containerd [link to code](https://github.com/containerd/containerd/blob/8167751f569c9374a6c3ef5b39a42629577a543a/remotes/docker/fetcher.go#L160)

wget

after retry with breakpoint transmission, the content will be wrong.
sha256sum

maybe related to https://github.com/containerd/containerd/issues/3974#issuecomment-989510035 , https://github.com/goharbor/harbor/issues/17643 , https://github.com/goharbor/harbor/issues/16556

# Issue being fixed
Fixes #(issue)

- [ ] Considered the docs impact and opened a new docs issue or PR with docs changes if needed in [website repository]([GitHub - goharbor/website: Source for the main Harbor website](https://github.com/goharbor/website)).

### DIFF
--- a/src/server/middleware/repoproxy/proxy.go
+++ b/src/server/middleware/repoproxy/proxy.go
@@ -22,6 +22,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/beego/beego"
+
 	"github.com/goharbor/harbor/src/common/security"
 	"github.com/goharbor/harbor/src/common/security/proxycachesecret"
 	"github.com/goharbor/harbor/src/controller/project"
@@ -73,6 +75,12 @@ func handleBlob(w http.ResponseWriter, r *http.Request, next http.Handler) error
 	}
 
 	if !canProxy(r.Context(), p) || proxyCtl.UseLocalBlob(ctx, art) {
+		defer func() {
+			if err := recover(); err != nil {
+				log.Error(err)
+				panic(beego.ErrAbort)
+			}
+		}()
 		next.ServeHTTP(w, r)
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


Vad1mo:
@hunshcn thank you for your contribution, we will review the issue.
is there any possibility we might be able to reproduce the error?

hunshcn:
> @hunshcn thank you for your contribution, we will review the issue. is there any possibility we might be able to reproduce the error?

This error can be reproduced by using unstable s3 storage.

But the most convenient way is to panic http.ErrAbortHandler in the code.

hunshcn:
Actually, there should be no error message after the file download starts.

hunshcn:
> @hunshcn So this issue only happened when the `runmode=dev`?

yes

chlins:
@hunshcn Hi, personally I think since current dev mode is prod which will not pose to this issue, and also the behavior is designed by upstream beego side, we can add your modification for this case, but there are many other similar cases, from my side the behavior from beego may also make sense for dev && debug, so I suggest keep current behavior and resolve the issue from update the mode to prod, do you think so? But great thanks for pointing this.

hunshcn:
> @hunshcn Hi, personally I think since current dev mode is prod which will not pose to this issue, and also the behavior is designed by upstream beego side, we can add your modification for this case, but there are many other similar cases, from my side the behavior from beego may also make sense for dev && debug, so I suggest keep current behavior and resolve the issue from update the mode to prod, do you think so? But great thanks for pointing this.

you are right, but I still think it's a very bad thing to download files with abnormal content. kubelet cannot handle this  situation.

chlins:
@hunshcn I think the behavior is designed by beego dev mode, although we can handle this case from harbor side as your pr described, but such a change is not universal, I suggest open issue on the upstream beego project or keep the current behavior.

PR #17776, opened 2022-11-10
Bugfix: add "default"

xzxiaoshan:
Bugfix: If there is no default, the new configuration file is included and cannot run normally, or include will be placed last.

I try to add a server with the same port. Other domain names are processed. The default values that are not modified by default cannot be recognized normally.

Signed-off-by: xzxiaoshan <365384722@qq.com>

### DIFF
--- a/make/photon/prepare/templates/nginx/nginx.https.conf.jinja
+++ b/make/photon/prepare/templates/nginx/nginx.https.conf.jinja
@@ -50,7 +50,7 @@ http {
   include /etc/nginx/conf.d/*.server.conf;
 
   server {
-    listen 8443 ssl;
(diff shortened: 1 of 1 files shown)


xzxiaoshan:
The code that fails the detection has nothing to do with the code I changed. I only modified a small amount.

MinerYang:
Could you provide the related nginx document to refer?

xzxiaoshan:
官方不计划处理吗？

wy65701436:
Hi @xzxiaoshan, thank you for your contribution. As @MinerYang mentioned earlier, would you please provide us with any link or document that we can use as a reference for this fix? This would be very helpful for us understanding it in deep.

Issue #18060, opened 2023-01-04
Specific project detail information endpoint returns 404 even if the project exist


batusekerci:
Hi, I encountered a bug when I was trying to use "GET [​/projects​/{project_name_or_id}] Return specific project detail information" endpoint. We are keeping project names as integers. There was no problem when I was creating project with integer name from API or GUI. Whenever I used the "detail info" endpoint, it returns me 404 Not Found error, even if the project exists and I can push and pull images. 

If I create another project with just one char and integers there is no problem.

Example:
Project Name: 555  -> 404 Not found
Project Name: 555s  -> 200 Successful

BTW I know that our version is old, I do not know that this bug is fixed in future releases. I could not find any issues about this bug so I decided to create one.

Note: I noticed that other project endpoints are also not working which have "project_name_or_id" parameter.
 
Harbor version: v2.2.0-ec0ba116

stonezdj:
what is the result this query?
```
select * from project where name = '555';
select * from project where name like '%555%';
```

j4ckstraw:
I have a similar problem. 
I got `project not found`, mostly, when new image pushed.

batusekerci:
> Do you mean all the project named with digits will return 404 or just one of them? What is the result this query?
> 
> ```
> select * from project where name = '555';
> select * from project where name like '%555%';
> ```

I mean **all** projects named with digits

batusekerci:
The issue is still active.

chlins:
@batusekerci Hi, please try to add the header `X-Is-Resource-Name: true` in request if your project name is numeric.

PR #18268, opened 2023-02-22
bugfix: unknown: cannot convert {project named using numbers} to Text

llaoj:

When named a project using numbers, Calling this API will cause an error:

image

When I surround the project name with double quotes, the bug disappears:

image

Tail the CORE component log found the following error:

```
[ERROR] [/lib/http/error.go:56]: {"errors":[{"code":"UNKNOWN","message":"unknown: cannot convert 10042620 to Text"}]}
```

Similar problems occur when using the Replications function:

image

A shorter error message on the page:

```
failed to do the prepare work for pushing/uploading resources: list projects with query name={10042620}: http error: code 500, message {"errors":[{"code":"UNKNOWN","message":"internal server error"}]}
```

So, I solved this problem by using double quotes, and the test passed. Few changes.

- [✅ ] Well Written Title and Summary of the PR
- [✅ ] Label the PR as needed. "release-note/ignore-for-release, release-note/new-feature, release-note/update, release-note/enhancement, release-note/community, release-note/breaking-change, release-note/docs, release-note/infra, release-note/deprecation"
- [ ✅] Accepted the DCO. Commits without the DCO will delay acceptance.
- [ ✅] Made sure tests are passing and test coverage is added if needed.

### DIFF
--- a/src/pkg/reg/adapter/harbor/base/adapter.go
+++ b/src/pkg/reg/adapter/harbor/base/adapter.go
@@ -176,7 +176,7 @@ func (a *Adapter) PrepareForPush(resources []*model.Resource) error {
 	for p := range projects {
 		ps = append(ps, p)
 	}
-	q := fmt.Sprintf("name={%s}", strings.Join(ps, " "))
+	q := fmt.Sprintf("name={\"%s\"}", strings.Join(ps, "\" \""))
 	// get exist projects
 	queryProjects, err := a.Client.ListProjectsWithQuery(q, false)
 	if err != nil {


llaoj:
release-note/update

Vad1mo (approved):

chlins:
It's also similar with https://github.com/goharbor/harbor/issues/18060, better to resolve the problem from server side instead of client side.

llaoj:
yep, this script aim to fix this bug:

```
#!/bin/sh

sed -i 's#goharbor/harbor-core:v2.6.2#registry.cn-beijing.aliyuncs.com/llaoj/goharbor_harbor-core:v2.6.2-230222#g' ./docker-compose.yml

nginxconf="./common/config/nginx/nginx.conf"
if grep -q "large_client_header_buffers" $nginxconf; then
    sed -i 's/large_client_header_buffers.*/large_client_header_buffers 16 5120k;/' $nginxconf
else
    sed -i '/http {/a\  large_client_header_buffers 16 5120k;' $nginxconf
fi
```

Maybe my approach is not entirely compliant with the standards, but it works well in this version. If you only want to address this issue in this version (v2.6.2), you can run this script after the 'prepare' command.
Additionally, this script also resolves the issue of URI length limitation. When using the /api/v2.0/projects interface, the length of the q={"p1" "p2" "p3" ...} parameter becomes too long due to the large number of projects we want to query. This is restricted by nginx.

PR #18269, opened 2023-02-22
enable gzip in the nginx config

yankay:
Enable gzip in the nginx config for slow networks.

# Issue being fixed
Fixes #(issue)

- [x]  Accepted the DCO. Commits without the DCO will delay acceptance.
- [x]  Made sure tests are passing and test coverage is added if needed.

### DIFF
--- a/make/harbor.yml.tmpl
+++ b/make/harbor.yml.tmpl
@@ -8,6 +8,7 @@ hostname: reg.mydomain.com
 http:
   # port for http, default is 80. If https enabled, this port will redirect to https port
   port: 80
+  gzip_enabled: false
 
 # https related config
 https:
@@ -16,6 +17,7 @@ https:
   # The path of cert and key files for nginx
   certificate: /your/certificate/path
   private_key: /your/private/key/path
+  gzip_enabled: false
 
 # # Uncomment following will enable tls communication between all harbor components
 # internal_tls:
--- a/make/photon/prepare/templates/nginx/nginx.http.conf.jinja
+++ b/make/photon/prepare/templates/nginx/nginx.http.conf.jinja
@@ -59,6 +59,14 @@ http {
     # customized location config file can place to /etc/nginx/etc with prefix harbor.http. and suffix .conf
     include /etc/nginx/conf.d/harbor.http.*.conf;
 
+{% if gzip_enabled %}
+    # enable gzip
+    gzip on;
+    gzip_min_length 1000;
+    gzip_proxied expired no-cache no-store private auth;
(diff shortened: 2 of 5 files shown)


OrlinVasilev:
@yankay thank you for your contribution!
Does that fix any specific issue or you have faced it can you please elaborate a bit more!
I like the idea but we need to be a bit more specific when we introduce new stuff !

Thank you!

yankay:
> @yankay thank you for your contribution! Does that fix any specific issue or you have faced it can you please elaborate a bit more! I like the idea but we need to be a bit more specific when we introduce new stuff !
> 
> Thank you!

Thanks @OrlinVasilev 

When we use the harbor to serve the helm charts, the `helm repo update` needs to download the index.yaml. 

In the scenario, the index.yaml is 2MB(compressed size 180KB), so when the Nginx use gzip compress, the network IO would be reduced 10x,  it's very effective in the Slow Network.

image

panpan0000:
it saves a lot of public cloud bandwidth cost :-)

stonezdj (changes requested):
Could please add an option in the harbor.yaml.tmpl so that we could enable or disable it?

wy65701436:
Regarding Docker images, this option may only be beneficial for the manifest rather than the blobs, which typically make up the majority of an image's size. I would prefer to add an option that allows the user to decide whether to enable it.

Harbor provides a default configuration for Nginx, but users can enable or disable any configurations based on their specific requirements. However, this default configuration is only applicable for online/offline installations. What if a user installs Harbor on Kubernetes and prefers to use Contour as the proxy?

stonezdj:
Because most of the content type in the Harbor server might be application/octet-stream(blob) or application/vnd.docker.distribution.manifest.v2+json (manifest), it is not in the list of the gzip_types and won't be compressed.

As for the helm chart's index.yaml, it is the output of chartmuseum server, it is already removed from Harbor 2.8.0

So the performance improvement is limited to the REST API and javascript level.

wy65701436:
hi @yankay would you mind to reply the above two comments? Or we can go ahead to stale this pull request.

yankay:
> hi @yankay would you mind to reply the above two comments? Or we can go ahead to stale this pull request.

Thanks @wy65701436 

The PR uses the `gzip_types ` to filter the mime-type of the content. So the blob would not be compressed. :-)

But if the chartmuseum be removed , the compress is not very useful.

So the PR may should be closed :-)

wy65701436:
@yankay thanks for your input.

Issue #18381, opened 2023-03-20
GC retries forever on permanent errors, most commonly on "manifest unknown" (404)


dkulchinsky:
**Expected behavior and actual behavior:**
When GC encounters a permanent error such as 404 when deleting Blobs or Manifests, it should not retry as there's no expectation that an object that was not found would appear if we retried.

GC Job log shows:
```
2023-03-20T18:37:49Z [INFO] [/jobservice/job/impl/gc/util.go:67]: failed to exec v2DeleteManifest, error: http status code: 404, body: {"errors":[{"code":"MANIFEST_UNKNOWN","message":"manifest unknown"}]}
, will retry again after: 171.764815ms
2023-03-20T18:37:50Z [INFO] [/jobservice/job/impl/gc/util.go:67]: failed to exec v2DeleteManifest, error: http status code: 404, body: {"errors":[{"code":"MANIFEST_UNKNOWN","message":"manifest unknown"}]}
, will retry again after: 152.306794ms
2023-03-20T18:37:50Z [INFO] [/jobservice/job/impl/gc/util.go:67]: failed to exec v2DeleteManifest, error: http status code: 404, body: {"errors":[{"code":"MANIFEST_UNKNOWN","message":"manifest unknown"}]}
, will retry again after: 623.007798ms
```

so it keeps retrying ad gets the same 404 every time, eventually it gives up, but since it's still in `artifact_thrash` it will go through the same process again next time.

This causes several issues:
1. when a 404 error occurs, the GC job enters a retry loop with an exponential backoff, each time getting the same 404 error back
2. this greatly increases the time it takes for this job to run
3. when it finally gives up, the thrash artifact that had the 404 error is not removed so it lingers in `artifact_thrash` table in the DB and when GC runs the next time the above is repeated all over again.

Since the goal of GC is to delete the referenced manifest or blob, if the response from the storage is that it in fact doesn't exist (404), then that's the desired end result anyway and shouldn't be treated as a retryable error, instead it should identify this condition, remove it from `artifact_thrash` and proceed.

**Steps to reproduce the problem:**
I don't have an exact repro as there could be different scenarios for why the blob is no longer in the backend storage of the registry, but on a given 404 error, the GC should should mark the manifest/blob is deleted, not retry and remove it from `artifact_thrash` table.

**Versions:**
Please specify the versions of following systems.

- harbor version: 2.7.1
- docker engine version: N/A
- docker-compose version: N/A

**Additional context:**

[...]

PR #18662, opened 2023-05-10
fix: add retry on the caller of v2DeleteManifest instead within v2DeleteManifest

zyyw:

# Issue being fixed
Fixes #(issue)


dkulchinsky:
@zyyw I'm going to build this on top of v2.7.1 (this is the version we are currently running) and will take it for a spin

wy65701436 (approved):
lgtm

dkulchinsky:
@zyyw @wy65701436 I running local build with this change in our sandbox, AFAICT this works as expected, I'm doing additional tests in the next ~12 hours so will provide more feedback, but looking at the implementation I think this is good to go!

one more question: once this is merged, I'm assuming this fix can be cherry-picked into 2.7 & 2.8 releases? we're still not ready to upgrade to 2.8 since we have work around switching charts from chartmuseum to OCI and would be great to have this fix in 2.7

zyyw:
@dkulchinsky , thank you for the feedback. Really appreciate it!

We think this is essentially an enhancement, not a bug fix. I'm afraid we will not back port this change to 2.7 & 2.8 releases.

MinerYang (approved):
lgtm

dkulchinsky:
> @dkulchinsky , thank you for the feedback. Really appreciate it!
> 
> We think this is essentially an enhancement, not a bug fix. I'm afraid we will not back port this change to 2.7 & 2.8 releases.

Hey @wy65701436, that's unfortunate 😕 I'm curious why you consider this an enhancement only and not a bug fix?

I think I've shown in my [analysis](https://github.com/goharbor/harbor/issues/18381#issuecomment-1477840214) in the original issue that the current behavior is improperly handling 404 when it is returned inside the retry loop which I'm sure you'll agree is not correct behavior and is fixed by this change.

The fix is small and adresses a real issue, I think it would be beneficial (not just for my sake) to include it as a fix and backport it to the current stable versions (2.7 & 2.8).

wy65701436:
@dkulchinsky Since the root cause of this problem could be either a network issue or a storage problem, in the harbor side, we simply added a retry mechanism to temporarily conceal the problem. It is important to note that this is not a bug in Harbor, but rather a failure tolerance in the garbage collector (GC).

dkulchinsky:
> @dkulchinsky Since the root cause of this problem could be either a network issue or a storage problem, in the harbor side, we simply added a retry mechanism to temporarily conceal the problem. It is important to note that this is not a bug in Harbor, but rather a failure tolerance in the garbage collector (GC).

Thanks @wy65701436, I agree that the retry mechanism was added a while back to allow GC to be more resilient with temporal issues in the upstream backend storage or network and it has improved GC considerably.

My main point here is there was a gap in the retry implementation where a specific edge case was not correctly handled and is now addressed by this PR, I believe it would be highly beneficial to backport this small and low risk change to 2.7 & 2.8, and happy to work on the PRs if that helps.

PR #18728, opened 2023-05-24
Return error when proxy cache get too many request error(429)

### LINKED ISSUE #18707: Return 429 when upstream registry return 429 in proxy cache
In proxy cache project, if the upstream registry return 429, this error should return to the local with the same error and message so that the client could understand what is happening.

stonezdj:
Add 429 too many request error in http error
  Fixes #18707

# Issue being fixed
Fixes #(issue)

### DIFF
--- a/src/controller/proxy/controller.go
+++ b/src/controller/proxy/controller.go
@@ -166,6 +166,9 @@ func (c *controller) UseLocalManifest(ctx context.Context, art lib.ArtifactInfo,
 	remoteRepo := getRemoteRepo(art)
 	exist, desc, err := remote.ManifestExist(remoteRepo, getReference(art)) // HEAD
 	if err != nil {
+		if errors.IsRateLimitError(err) && a != nil { // if rate limit, use local if it exists, otherwise return error
+			return true, nil, nil
+		}
 		return false, nil, err
 	}
 	if !exist || desc == nil {
--- a/src/server/middleware/repoproxy/proxy.go
+++ b/src/server/middleware/repoproxy/proxy.go
@@ -46,6 +46,8 @@ const (
 	ensureTagMaxRetry   = 60
 )
 
+var tooManyRequestsError = errors.New("too many requests to upstream registry").WithCode(errors.RateLimitCode)
+
 // BlobGetMiddleware handle get blob request
 func BlobGetMiddleware() func(http.Handler) http.Handler {
 	return middleware.New(func(w http.ResponseWriter, r *http.Request, next http.Handler) {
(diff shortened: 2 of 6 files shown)


src/server/middleware/repoproxy/proxy.go
```diff
@@ -46,6 +46,8 @@ const (
 	ensureTagMaxRetry   = 60
 )
 
+var tooManyRequestsError = errors.New("too many requests to upstream registry").WithCode(errors.RateLimitCode)
```
wy65701436, line 49: You have exceeded the pull rate limit for the proxy cache.

wy65701436 (approved):
lgtm

src/controller/proxy/controller.go
```diff
@@ -155,6 +155,9 @@ func (c *controller) UseLocalManifest(ctx context.Context, art lib.ArtifactInfo,
 	remoteRepo := getRemoteRepo(art)
 	exist, desc, err := remote.ManifestExist(remoteRepo, getReference(art)) // HEAD
 	if err != nil {
+		if errors.IsRateLimitError(err) && a != nil { // if rate limit, use local if it exists, otherwise return error
```
reasonerjt, line 158: Why do we need to treat 429 as a special case?  Generally, if there's an error when refreshing the cache, and the data exists in the cache and the cache is not expired, we should use the data in the cache

src/controller/proxy/controller.go
```diff
@@ -155,6 +155,9 @@ func (c *controller) UseLocalManifest(ctx context.Context, art lib.ArtifactInfo,
...
+			return true, nil, nil
+		}
```
reasonerjt, line 160: IMHO, what needs to be refined is in the path when error happens and the cache is expired or the data is not cached.

In this case, we better return the error more explicitly instead of a 404.  I think a better choice here is to return 500 which wraps the root cause (in this case 429), if 500 can't be rendered in a friendly way for mainstream clients, we find a proper error so most clients can handle it.

stonezdj: To avoid leaking any server information. if return 500 error to the client, it just sends the "internal server error" to the client, no other useful information.
see
https://github.com/goharbor/harbor/blob/cc4c08b0a1cef086fdb47d90be8f6b32efe170cf/src/lib/http/error.go#L55-L54

reasonerjt:
I'm not quite convinced why we fix 429 but leave other error-handling situations as-is.  
Returning a 404 for whatever backend error when the local cache does not exist is as confusing as the problem this PR is trying to solve.

That said it's up to @stonezdj to make decision.

stonezdj:
> I'm not quite convinced why we fix 429 but leave other error-handling situations as-is. Returning a 404 for whatever backend error when the local cache does not exist is as confusing as the problem this PR is trying to solve.
> 
> That said it's up to @stonezdj to make decision.

For any other error:
If the Harbor can return an internal server error 500 to the client when there is error and no cache can be served in local
return 404 if fallback to local and no cache can be served, 
It doesn't make too much difference.

For the least impact to the current system,  keep the current implementation.

zyyw (approved):

PR #18735, opened 2023-05-25
Add a text to explain the time window for GC

### LINKED ISSUE #18692: Clarifying the action of button "GC NOW"
It is not obvious that the "GC NOW" button on the "Clean up" page does not free disk space for all images. This is especially disconcerting when Quotas are activated because users can see that there is some disk space still being used, even after GC. This is most obvious when you delete all images in Harbor. I think most new users expect to see 0 bytes used after GC on an empty project.

I suggest adding some text next to the "GC NOW" buttton. This text could be "Note : Any garbage collection operation (including the dry run) only covers images that were not uploaded within the previous two hours." This text is directly copied from paragraph [Handling Images That Are Still Uploading](https://container-registry.com/docs/administration-manual/terminal-deletion-of-images/)
In the case CG ignored some images, they could be mentionned in the log with something like "GC ignored the following images : ..."

AllForNothing:
1.Fixes #18692
UI change:
image

# Issue being fixed
Fixes #(issue)

### DIFF
--- a/src/portal/src/app/base/left-side-nav/clearing-job/gc-page/gc/gc.component.html
+++ b/src/portal/src/app/base/left-side-nav/clearing-job/gc-page/gc/gc.component.html
@@ -59,6 +59,14 @@
     <div class="clr-row">
         <div class="clr-col-2 flex-200"></div>
         <div class="clr-col">
+            <span class="explain">{{
+                'GC.EXPLAIN_TIME_WINDOW' | translate
+            }}</span>
+        </div>
+    </div>
+    <div class="clr-row">
+        <div class="clr-col-2 flex-200"></div>
+        <div class="clr-col">
             <clr-toggle-container class="mt-05">
                 <clr-toggle-wrapper>
                     <input
--- a/src/portal/src/i18n/lang/de-de-lang.json
+++ b/src/portal/src/i18n/lang/de-de-lang.json
@@ -1228,6 +1228,7 @@
         "PARAMETERS": "Parameter",
         "DELETE_UNTAGGED": "Erlaube Speicherbereinigung auf Artefakte ohne Tag",
         "EXPLAIN": "Speicherbereinigung (Garbage Collection / GC) ist eine rechenintensive Operation, die die Registry-Leistung beeinflussen kann",
+        "EXPLAIN_TIME_WINDOW": "Artifacts uploaded in the past 2 hours(the default window) are excluded from garbage collection",
         "DRY_RUN_SUCCESS": "Probelauf erfolgreich gestartet"
     },
     "RETAG": {
--- a/src/portal/src/i18n/lang/en-us-lang.json
+++ b/src/portal/src/i18n/lang/en-us-lang.json
@@ -1228,6 +1228,7 @@
         "PARAMETERS": "Parameters",
         "DELETE_UNTAGGED": "Allow garbage collection on untagged artifacts",
         "EXPLAIN": "GC is a compute intensive operation that may impact registry performance",
+        "EXPLAIN_TIME_WINDOW": "Artifacts uploaded in the past 2 hours(the default window) are excluded from garbage collection",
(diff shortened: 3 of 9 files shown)


AllForNothing:
@OrlinVasilev Can help to review the English text in this PR?
We add this text to explain the time window for GC

Vad1mo:
IMO the text is vague.

- How long is the time window
- What is happens inside and outside this time window?
- What does "covers" mean exactly?

Isn't the time window is hard-coded to 2h? 
Why not write it in more simple words, explaining the restriction.

_Any garbage collection operation (including the dry run) only covers images that were not uploaded within the specified time window"_

Here is the text in more in simple words, straight to the point:

- EN: Artifacts uploaded in the past 2 hours are excluded from garbage collection.
- DE: Artefakte, die in den vergangenen 2 Stunden hochgeladen wurden, sind von der Garbage Collection ausgenommen.
- ES: Los artefactos cargados en las últimas 2 horas se excluyen de la recogida de basura.

Vad1mo (changes requested):
IMO the text is vague.

How long is the time window
What is happens inside and outside this time window?
What does "covers" mean exactly?
Isn't the time window is hard-coded to 2h?
Why not write it in more simple words, explaining the restriction.

Any garbage collection operation (including the dry run) only covers images that were not uploaded within the specified time window"

Here is the text in more in simple words, straight to the point:

EN: Artifacts uploaded in the past 2 hours are excluded from garbage collection.
DE: Artefakte, die in den vergangenen 2 Stunden hochgeladen wurden, sind von der Garbage Collection ausgenommen.
ES: Los artefactos cargados en las últimas 2 horas se excluyen de la recogida de basura.

AllForNothing:
@OrlinVasilev @Vad1mo Thanks for the comments.
But I don't know if '2 hrs' is proper here, because the time window is an environment variable that is hard-coded as 2 hrs by default,  and users can modify it.
What about "Artifacts uploaded within the specified time window(the default is 2 hrs) are excluded from garbage collection"?

AllForNothing:
> Can we read that setting and display the value that the user has setup ?

@OrlinVasilev No, users can not read it. because we don't want to expose this to the user, but the user can surely override this value in the environment variable:
I think the explanation in the documentation is very clear. 
So would it be better to be consistent with the documentation here(say 'a time window(2 hours)')?

OrlinVasilev:
@AllForNothing can you point me to that part of the documentation where is listed that you can tune that variable ?
So the messaage can be :

> Artifacts uploaded in the past 2 hours(the default window) are excluded from garbage collection.

AllForNothing:
@OrlinVasilev This is not included in our doc. 
This should be a part of development instruction(Unfortunately, we don't have it)
Can modify it by below steps
`sudo -E make -e down
add GC_TIME_WINDOW_HOURS=2 to make/common/config/core/env
sudo -E make -e start`

AllForNothing:
@OrlinVasilev @Vad1mo I will update my PR later according to the text 'Artifacts uploaded in the past 2 hours(the default window) are excluded from garbage collection.'

wy65701436:
@Vad1mo @OrlinVasilev The original intention of this tooltip is to inform users that the garbage collector will not remove deleted artifacts within the specified time window, which is set to 2 hours by default. This is done to prevent any potential damage to artifacts that are currently being uploaded. Many users do not read the official documentation and often inquire about this behavior.

Regarding the tuning aspect, it was not originally intended for users to customize. Instead, it was primarily designed for **debugging purposes**. In other words, if a user sets the tuning value to 0 (disabled), it may cause issues in concurrent scenarios where the GC could potentially disrupt the pushing of artifacts. This pushing process is a critical pathway within the registry. So, in any of production environment, we don't recommend user to reset it. If we would like to doc it, @AllForNothing please file another issue specifically addressing this matter. Let's keep the review focused on the wording of the tooltip.

OrlinVasilev (approved):
lgtm

Vad1mo (approved):

PR #18766, opened 2023-05-31
build image: db report error Permission denied. 

AgoCan:
After I built the image in ubuntu arm64, executing install.sh reported the following error：

use command “docker logs harbor-db” get msg:

```
root@ubuntu:/opt/harbor# docker logs harbor-db
/docker-entrypoint.sh: line 4: //initdb.sh: Permission denied
/docker-entrypoint.sh: line 4: //initdb.sh: Permission denied
...
```


OrlinVasilev:
@AgoCan if you build it with some other underlying os does it work ? or if you build it for x68?

PR #18996, opened 2023-07-25
add validate webhhook address

lengrongfu:

The currently created wehook does not verify whether the address is available, and the jobserver fails to execute after saving.

# Issue being fixed
Fixes #(issue)

### DIFF
--- a/src/server/v2.0/handler/webhook.go
+++ b/src/server/v2.0/handler/webhook.go
@@ -17,7 +17,9 @@ package handler
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"strings"
+	"time"
 
 	"github.com/go-openapi/runtime/middleware"
 	"github.com/go-openapi/strfmt"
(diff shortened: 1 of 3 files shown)


chlins:
From security concern, I am not preferring to use net Dial for connection testing with the outside world. An attacker may use this scenario to perform insecure behaviors such as port scanning.

chlins:
> use `http.get` to verify, what do you think?

The webhook endpoint usually serves with the `POST` method, so use the `GET` method to check as the health check I don't think it's rigorous, for the webhook we don't have restricted methods like replication, and per the initial design, there's no `Test Connection` for this feature, I prefer to keep current design before we figure out the better solution.

lengrongfu:
Because if the `webhook` address filled in is wrong, the `jobserver` will make a wrong call and need to retry many times.

chlins:
> Because if the `webhook` address filled in is wrong, the `jobserver` will make a wrong call and need to retry many times.

The retry mechanism is a common way to ensure the guarantee, it's not make sense to configure an unavailable endpoint and also the retry times can be customized by user. https://github.com/goharbor/harbor/blob/560e6cd1757e50a836018bde4d3a1a68d227f8e9/make/harbor.yml.tmpl#L127

PR #19071, opened 2023-08-01
fix: cron string validation

### LINKED ISSUE #18091: Validators for Harbor APIs
In order to improve user experience and strengthen security, let's make the validators of the front and back ends consistent:
- [ ] 1:  Post `/api/2.0/projects`,    `storage_limit`  should  range  from -1 to 1024TiB
- [ ] 2: For all cron string,    the first field should always be `0`
- [x] 3: Cron string for replication rule, the second field can not be `*`
- [x] 4: Tag name, see #18073
- [x] 5: Secret length for robot account, should range from 8 to 128
- [x] 6: P2p-preheat policy, `name` should follow the pattern `^[a-z0-9]+(?:[._-][a-z0-9]+)*$`

zyyw:

# Issue being fixed
Fixes #18091
task 2

### DIFF
--- a/src/common/utils/utils.go
+++ b/src/common/utils/utils.go
@@ -301,6 +301,22 @@ func CronParser() cronlib.Parser {
 	return cronlib.NewParser(cronlib.Second | cronlib.Minute | cronlib.Hour | cronlib.Dom | cronlib.Month | cronlib.Dow)
 }
 
+// ValidateCronString check whether it is a valid cron string and whether the 1st field (indicating Seconds of time) of the cron string is a fixed value of 0 or not
+func ValidateCronString(cron string) error {
+	if len(cron) == 0 {
+		return fmt.Errorf("empty cron string is invalid")
+	}
+	_, err := CronParser().Parse(cron)
+	if err != nil {
+		return err
+	}
+	cronParts := strings.Split(cron, " ")
+	if len(cronParts) == 6 && cronParts[0] != "0" {
+		return fmt.Errorf("the 1st field (indicating Seconds of time) of the cron setting must be 0")
+	}
+	return nil
+}
+
 // MostMatchSorter is a sorter for the most match, usually invoked in sort Less function
 // usage:
 //
--- a/src/pkg/p2p/preheat/models/policy/policy.go
+++ b/src/pkg/p2p/preheat/models/policy/policy.go
@@ -118,6 +118,18 @@ type Trigger struct {
 	} `json:"trigger_setting,omitempty"`
 }
 
+// ValidatePreheatPolicy validate preheat policy
+func (s *Schema) ValidatePreheatPolicy() error {
+	// currently only validate cron string of preheat policy
+	if s.Trigger != nil && s.Trigger.Type == TriggerTypeScheduled && len(s.Trigger.Settings.Cron) > 0 {
+		if err := utils.ValidateCronString(s.Trigger.Settings.Cron); err != nil {
+			return errors.New(nil).WithCode(errors.BadRequestCode).
+				WithMessage("invalid cron string for scheduled preheat: %s, error: %v", s.Trigger.Settings.Cron, err)
+		}
+	}
+	return nil
+}
+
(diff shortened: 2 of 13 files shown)


AllForNothing (approved):
lgtm

src/common/utils/utils.go
```diff
@@ -301,6 +301,25 @@ func CronParser() cronlib.Parser {
...
+// ValidateCronString check whether it is a valid cron string and whether the 1st field (indicating Seconds of time) of the cron string is a fixed value of 0 or not
+func ValidateCronString(cron string) (bool, error) {
+	if len(cron) == 0 {
+		return false, fmt.Errorf("empty cron string is invalid")
+	}
+	_, err := CronParser().Parse(cron)
+	if err != nil {
+		return false, err
+	}
+	cronParts := strings.Split(cron, " ")
```
wy65701436, line 313: CronParser().Parse should cover this situation, there is no need to check this.

wy65701436, line 313:
```
_, err := CronParser().Parse(cron)
if err != nil {
   return false, err
}

t, ok := s.(*cronlib.SpecSchedule)
if ok {
    if t.Second == 0 {
	    return false, err
    }
}
```

src/pkg/p2p/preheat/models/policy/policy.go
```diff
@@ -118,6 +118,19 @@ type Trigger struct {
 	} `json:"trigger_setting,omitempty"`
 }
 
+// ValidatePreheatPolicy validate preheat policy
+func (s *Schema) ValidatePreheatPolicy() error {
+	// currently only validate cron string of preheat policy
+	if s.Trigger != nil && s.Trigger.Type == TriggerTypeScheduled && len(s.Trigger.Settings.Cron) > 0 {
```
wy65701436, line 124: duplicate validation in the length of cron string

src/pkg/p2p/preheat/models/policy/policy.go
```diff
@@ -118,6 +118,19 @@ type Trigger struct {
...
+		isValid, err := utils.ValidateCronString(s.Trigger.Settings.Cron)
+		if !isValid || err != nil {
```
wy65701436, line 126:
```
if err != nil {
 return err --- this is 500 error
}
if !isValid {
 return BadRequest -- this is 400 error
}
```

src/pkg/retention/policy/models.go
```diff
@@ -58,6 +60,24 @@ type Metadata struct {
 	Scope *Scope `json:"scope" valid:"Required"`
 }
 
+// ValidateRetentionPolicy validate the retention policy
+func (m *Metadata) ValidateRetentionPolicy() error {
```
wy65701436, line 64: ditto

src/pkg/p2p/preheat/models/policy/policy.go
zyyw: This is because we need to skip `utils.ValidateCronString()` validation for cron with empty string "". In preheat and tag retention, it uses empty string cron to unschedule an existing job.

src/pkg/retention/policy/models.go
zyyw: Same as above. This is because we need to skip `utils.ValidateCronString()` validation for cron with empty string "". In preheat and tag retention, it uses empty string cron to unschedule an existing job.

src/pkg/p2p/preheat/models/policy/policy.go
zyyw: actually，when the first field of cron is not 0, like "1 0 0 1 1 *", `utils.ValidateCronString` will return `false` and with error message. This is a BadRequest.

zyyw: maybe for `utils.ValidateCronString()` we should only return `error`, instead of return `(bool, error)`

stonezdj (approved):

PR #19085, opened 2023-08-02
Install tar package in database dockerfile

guoard:
Having the tar package in the database container would be beneficial for creating an archive of dump files.


chlins:
Thanks for your contribution, but install other dependencies may bring potential CVEs and increase the image size, so harbor only want to keep the minimal package lists. You can create additional containers which installed any required packages and mount the harbor volume or use some community solutions such as `kubectl debug` to satisfy your requirements.

PR #19321, opened 2023-09-06
Enabling flatenning on empty namespace

### LINKED ISSUE #19319: Flattening not working with empty namespace (Replication rule for ECR)
I expect flattening to work regardless of if a namespace is specified or not.

Actual behavior is that no matter which level of flatenning is chosen, the project name is always being appended to the image, except when a namespace is given, then the project name is properly replaced by the namespace.

Following [these steps](https://anywhere.eks.amazonaws.com/docs/packages/harbor/harboruse/#repository-replication-from-harbor-to-a-private-amazon-elastic-container-registry-ecr-repository), when one starts replication, if my ecr repo is named SOFTWARE, and I create a non-proxy project named DEV, even if I specify Flatten all, harbor will try to push the image as DEV/SOFTWARE, and fail as that repo does not exist in ecr.

If I specify a namespace as TEST in replication, and specify a flatenning level, such as Flatten 1 Level, then the image that's being pushed into ecr is TEST/SOFTWARE, which also fails because no such repo exists, as the repo SOFTWARE is the one in ecr.

I would guess flatten and no namespace would give one the ability to push just SOFTWARE to ecr, without either the project or namespace being attached.

**Versions:**

- harbor version: 2.9.0
- Deployed by helm chart 1.13.0

NicoForce:
- Removes the if condition that skips flatenning when namespace length is 0.
- Adds case in switch for dstRepoPrefix value when namespace length is 0.
- Removed empty namespace test.
- Added 6 new tests for empty namespace behavior.

# Issue being fixed
Fixes #19319 

# Expected behavior and actual behavior
I expect flattening to work regardless of if a namespace is specified or not.

Actual behavior is that no matter which level of flatenning is chosen, the project name is always being appended to the image, except when a namespace is given, then the project name is properly replaced by the namespace.

Let's take this replication rule for example where the namespace is empty and the flattening is on:
image

In the following screenshot, I have pushed an image to harbor, which we will try to replicate to ECR in the next step:
image

After activating manual replication, we can see it fails with the following error:
image

In the error image, we can observe that it adds the namespace dev to the image repository, even if we specified empty namespace and flattening in the replication rule. Here's what the repository looks like in AWS with no namespace.
image

In summary we expect harbor to push to repository **atlas** and not **dev/atlas** when namespace is empty in the replication rule.


NicoForce:
@wy65701436 @chlins Is there anything missing that I can do to move this forward? Please let me know, thanks!

Vad1mo:
Can you elaborate and confirm that your PR is backward compatible and won't break existing setups when applied?

Vad1mo (commented):
Can you elaborate and confirm that your PR is backward compatible and won't break existing setups when applied?
Also the impact on the docs.

NicoForce:
Commenting to remove staleness

NicoForce:
Commenting to remove stale label

mkhpalm:
I would like to see this merged since the existing behavior feels/acts more like a bug than a feature.

mmonaco:
What was the rationale for not supporting empty namespaces? It seems like a bug, as a new user.

Vad1mo:
@NicoForce to be honestly have a hard time understanding the problem. On the surface It seems to be some minor edge case, as there is no background information why this might be important. 

can you ping me on slack and we can discuss this issue, so I can better understand it, and I'll try to help you moving forward.

NicoForce:
@Vad1mo I've added screenshots and a more detailed explanation, can you double check it looks good and easy to understand?

fernandocarletti:
I'm having the exact same issue as described by @NicoForce. It feels off this is the default behavior, I get that it will be a breaking change, but it seems so small since the only requirement is to actually set the namespace as the desired one (for those relying on the replication, literally copying the namespace) or, in the worst scenario, create multiple replications to match what was previously matched by the resource filter.

@Vad1mo If keeping BC is important, maybe adding a select to the namespace setting where we could choose between `Project name`, `Custom` or `Empty`, where the `Project Name` can remain the default setting. Choosing `Custom` opens the current text box and `Empty` the new behavior.

I'm fairly new to Harbor and never looked into the codebase. I may give it a shot if the solution seems fine and @NicoForce is not available to implement it.

chlins:
I'm not sure if this change will introduce any breaking changes to the current behavior, or if it will simply address situations without a namespace without affecting any previous behavior. From the modifications in the PR, I can't clearly determine this.

StephenButcherTXST:
> @NicoForce to be honestly have a hard time understanding the problem. On the surface It seems to be some minor edge case, as there is no background information why this might be important.
> 
> can you ping me on slack and we can discuss this issue, so I can better understand it, and I'll try to help you moving forward.

Being in a similar boat, and waiting on this fix, here's my 2c:

Our issue is that our deployment registry (and afaik, most traditional image names) are in the format
`dns.host.name/image_name:tag` eg: `registry.example.com/image-name:1.0.0`
So many tutorials on private registries use/recommend this format.

We are working on migrating our build tools to push to harbor, but would like to keep our deployment registry, but because of this issue our images would be replicated as `registry.example.com/library/alpine:3` 
Here `library` can be substituted for any value you place in the 'Namespace' field in the replication rule, but it cannot be removed. The expectation is that if Namespace is blank, and we have "Flatten all levels" enabled, you'd expect the common format of `destination.registry/image-name:tag`. Yes, the docs and current tool tip does state that when empty, the source namespace is preserved, just the current method doesn't seem intuitive.

Since replication rules are not project (namespace) specific, and without a larger refactor of the replication process, I think the lowest impact would be to add a checkbox for `Remove Namespace` instead, with default value of 'False'. IMHO, it would expand the UI, offer a resolution to the bug, and should not impact existing replication rules.

fredgate:
Yes this a very valuable pull request :+1: 

This is a very annoying bug in harbor :cry: 
We want to replicate harbor projects (like `library`, `production`, `preproduction`) to cloud registries but we don't want to have the name of the harbor project in the name of repository. The `mysql` image should have the same repository path on docker hub and on our private registry, ia simply `mysql`. The same repository in our harbor `production` and `preproduction` projects should have the same repositories name in two cloud registries for production and preproduction (the name of repositories should not change accross environments).

outofrange:
For avoiding potential breaking changes - maybe opt-in into this behaviour through a "flatten to root namespace" checkbox (can't think of a better name right now...) shown when flattening is used?

I hope this change gets a bit more traction, since the current behaviour complicates our artifact delivery workflow.

PR #19910, opened 2024-01-25
Fix manifest list never being stored in proxy-cache projects 

### LINKED ISSUE #19609: Images exists in a proxy cache project but it do not show in project repositories list.
I set the library project to the Proxy Cache type and pointed it to Docker Hub. Then I configured the docker client to pull the image from the harbor's library. I am sure that the image has been cached in the library project, but I cannot find it in the repositories list of the library project.
bug

cnmcavoy:
Fix manifest list never being stored in proxy-cache projects  by looking up the manifest on HEAD requests

My understanding is that when docker (or containerd) requests a manifest list in a proxy-cache project, the client never makes a GET request for the manifest list. It makes a HEAD request for the manifest list, then makes a GET request for the platform architecture specific digest. As a result, the manifest list is never persisted in the storage layer, only in the redis cache layer. When the cache expires, this forces a new remote lookup for future requests instead of being able to rely on the storage, which is the purpose of the proxy-cache.

# Issue being fixed
This may or may not also fix #19609, but it does seem related.

### DIFF
--- a/src/server/middleware/repoproxy/proxy.go
+++ b/src/server/middleware/repoproxy/proxy.go
@@ -304,6 +304,10 @@ func proxyManifestHead(ctx context.Context, w http.ResponseWriter, ctl proxy.Con
 		// Then GET the image by digest, in order to associate the tag with the digest
 		// Ensure tag after head request, make sure tags in proxy cache keep update
 		bCtx := orm.Context()
+		_, err := ctl.ProxyManifest(ctx, art, remote)
+		if err != nil {
+			log.Debugf("Failed to ensure proxy manifest %+v , error %v", art, err)
+		}
 		for i := 0; i < ensureTagMaxRetry; i++ {
 			time.Sleep(ensureTagInterval)
 			bArt := lib.ArtifactInfo{ProjectName: art.ProjectName, Repository: art.Repository, Digest: string(desc.Digest)}


Vad1mo (approved):
Interesting observation, and hard to spot edge case. Do you have a step-by-step guide on how to reproduce the bug?

cnmcavoy:
> Interesting observation, and hard to spot edge case. Do you have a step-by-step guide on how to reproduce the bug?

We detected this because we have a Harbor / Docker-registry setup that is sensitive the filesystem layout that Harbor owns. We have a primary Harbor that runs in one primary kubernetes cluster. We configure Harbor to use S3 as the filesystem, and we have other clusters which run [docker-registry](https://github.com/distribution/distribution)'s which are given limited read access out of that S3 bucket. Those docker-registry can read Harbor's filesystem, because both are the same fundamental component, and so registries are affected by the filesystem, rather than the contents of Harbor's middleware (redis, in our case). In our case, we detected that manifest-list images were only partially mirrored, that the manifest layers pointed to by the manifests in the manifest-list were copied into the S3 bucket, but the manifest-list itself was absent.

To give a concrete example, if I inspect the current ubuntu focal manifest-list, it returns something like this:
```
{
  "manifests": [
    {
      "digest": "sha256:48c35f3de33487442af224ed4aabac19fd9bfbd91ee90e9471d412706b20ba73",
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      },
      "size": 424
    },
...
}
```
When we made requests to Harbor it would have no problem proxying the ubuntu:focal manifest-list and containers to the client, and the S3 bucket would contain the architecture of the image we requested, for example, the amd arch and "sha256:48c35f3de33487442af224ed4aabac19fd9bfbd91ee90e9471d412706b20ba73". This is expected, but what was missing is that the manifest-list wasn't stored. So pulling directly from the S3 bucket failed, but pulling through Harbor succeeded, even though Harbor was replying with layers that pointed at the S3 bucket. Harbor was caching the container layers, but not the manifest-list. 

This confused me initially because when I read the Harbor proxy cache code, the code seems to explicitly handle manifest-lists correctly. But when I inspected the layers in the docker-registry after pulling a manifest-list image for the first time through the proxy cache, you will find the layers for the inner container, but not the manifest list or any of the file pointers created for that resource. And the reason I found for this is that the docker client at some point in the past, used to make a request flow for an image that looked something like this: client -> HEAD $MANIFEST, client -> GET $MANIFEST, client fetches the layers. And when docker added manifest-lists, they added a step of indirection on the client side. So the client -> HEAD $MANIFESTLIST, then client GET -> $MANIFEST, and never executes a get on a manifest list resource directly. The client uses the list to pick out it's arch specific manifest and never refers to the list again. Harbor never sees the GET request on the manifest-list, and never saves the manifest list to disk.

Piccirello:
@cnmcavoy @Vad1mo is this PR still relevant? I'm facing the same issue as described in #19609 and am looking for a fix.

mmurhamm:
I am still facing the same issue, whether it is the same symptom or same underlying behavior I cannot tell. 
I am running Harbor v2.12.1-80219b7d on Docker compose (Ubuntu 22.04.5 LTS) and local file system (ext4). 
I pull images from a Harbor proxy cache repo using local Docker client or a k3s cluster; some images appear in the proxy projects - even with several minutes delay, others do not, even though they show as having been pulled (registry/repo:tag showing proxy instead of origin).
It does seem, though, that images that have been pulled from origin before do not appear in proxy project if pulled from same host while image has not been updated (digest has not changed). 
Strange, though, that such images also do not appear in the filesystem nor in the Harbor logs, so maybe no actual pull ever happened?
It would be good to see all images pulled through proxy projects, and add an indicator as to how they ended up there - actuall pull or otherwise.
thanks.

src/server/middleware/repoproxy/proxy.go
```diff
@@ -304,6 +304,10 @@ func proxyManifestHead(ctx context.Context, w http.ResponseWriter, ctl proxy.Con
 		// Then GET the image by digest, in order to associate the tag with the digest
 		// Ensure tag after head request, make sure tags in proxy cache keep update
 		bCtx := orm.Context()
+		_, err := ctl.ProxyManifest(ctx, art, remote)
```
stonezdj, line 307: Loggically, there is no need to call the ctl.ProxyManifest inside this go routine for head request.

PR #19962, opened 2024-02-06
remove query.Sorting from registration.go

### LINKED ISSUE #19569: Removing `Query.Sorting` as it we are deprecating it
How can we help you?
Since we are deprecating `Query.Sorting`, let's remove the corresponding logic that's still using `Query.Sorting`.
https://github.com/goharbor/harbor/blob/219e4ba48d4e88e5eb3335a09ddea9dcf542b5c9/src/lib/q/query.go#L30-L31

Its-Maniaco:
Fixes #19569


Its-Maniaco:
@zyyw @stonezdj can you please review the pr

src/pkg/scan/dao/scanner/registration.go
```diff
@@ -123,8 +123,14 @@ func ListRegistrations(ctx context.Context, query *q.Query) ([]*Registration, er
...
 	// Order the list
-	if query.Sorting != "" {
-		qs = qs.OrderBy(query.Sorting)
+	if len(query.Sorts) > 0 {
+		for _, sort := range query.Sorts {
+			sortKey := sort.Key
+			if sort.DESC {
+				sortKey += " DESC"
+			}
+			qs = qs.OrderBy(sortKey)
```
stonezdj, line 132: Supposing there are two columns in the query.Sorts, the query will always sort with the last column

PR #19965, opened 2024-02-08
replacing gorilla mux in router.go

### LINKED ISSUE #18077: Remove the dependency of gorilla/mux.
How can we help you?

We should find out the replacement of `gorilla/mux` as it has been archived.

> https://github.com/gorilla/mux

Its-Maniaco:
Replaced gorilla/mux with net/http

Partially Fixes #18077


Its-Maniaco:
@zyyw I tried my hand at this. If this looks correct I can proceed with other instances.

chlins:
This library is not only used in this one file in harbor, but also in many other places, so replacing only this file does not remove the dependency on this library, to solve this issue you need to find all references to this library and choose the appropriate or unified library or way to do the implementation, not just replacing this one file.

Vad1mo:
@Its-Maniaco, can you respond?

> Thanks for contributing! I recommend organizing all the references for this library into separate modules or components. This will make it easier to review each module or component individually for each PR.

Its-Maniaco:
@chlins @Vad1mo 
Thank you for your feedback and patience. I'll organize the references into separate modules and create individual PRs for each. Please let me know if you have any specific guidelines.

PR #20153, opened 2024-03-22
delete membership=0 in getProjectsByName

### LINKED ISSUE #20056: Pull Replication from GitLab does not work if not member of source group
Hello,

I'm trying to pull replicate images from a public gitlab instance to a local harbor instance, but it does not find the images (hence the job ends with 0 Tasks).

I believe I tracked this down to:
The filter `membership=true` or `membership=1` causes to return only projects of which one is a member. But in case of public repos/groups this is usually not the case.

My proposal is to remove the `membership=true` flag.

https://github.com/goharbor/harbor/blob/29cdc398e048b739c19a23fa0ffb5073df256009/src/pkg/reg/adapter/gitlab/client.go#L77
https://github.com/goharbor/harbor/blob/29cdc398e048b739c19a23fa0ffb5073df256009/src/pkg/reg/adapter/gitlab/client.go#L86

prima101112:
delete membership=0 in getProjectsByName but lets getProjects still using membership=1 for reserve if getProjectsByName not found any

# Issue being fixed
Fixes #20056


stonezdj (approved):
LGTM

stonezdj:
@prima101112 could you please give us the document link for it?

chlins:
@lxShaDoWxl Kindly review this, thank you!

prima101112:
> @prima101112 could you please give us the document link for it?

whats document, is this whay you mean https://docs.gitlab.com/ee/api/projects.html
from the issue just state that when using public gitlab **it should not require the member of the project**

please assist me if its missing something or need to be fixed

lxShaDoWxl:
you can also remove the "membership" here, only the search becomes more extensive. The disadvantage is only in performance

prima101112:
> you can also remove the "membership" here, only the search becomes more extensive. The disadvantage is only in performance

I think yes we can remove the "membership" but adjusting the fallback mechanism is risky. Currently, if the repository is not found, the search extends to the entire GitLab instance. However, it would be more efficient to limit the search to projects where the user is a member. This way, the search would only include projects that the user has access to. so i will leave it like that. please advice

prima101112:
@lxShaDoWxl could we get approval for this

lxShaDoWxl (approved):

Vad1mo (approved):

PR #20261, opened 2024-04-10
Makefile: Enable reusing local Go module cache during compile

palhotel:
# Enable Reusing Local Go Modules During Compile by Mapping the Cache Directory to Build Container

This commit adds support for leveraging the local `go mod cache` when building Harbor components. By mounting the host's Go module cache directory, as specified by `go env GOMODCACHE`, into the build container at `/go/pkg/mod`, subsequent builds can benefit from cached dependencies. This leads to faster build times due to reduced network activity and redundant downloads.

Updated a file: `Makefile` in the root directory of the Harbor source code.
- Add a parameter `GOLOCALMODCACHE=$(shell go env GOMODCACHE)`.
- Modify the compile commands, such as:
```makefile
@$(DOCKERCMD) run --rm -v $(GOLOCALMODCACHE):/go/pkg/mod -v $(BUILDPATH):$(GOBUILDPATHINCONTAINER) -w $(GOBUILDPATH_CORE) $(GOBUILDIMAGE) $(GOIMAGEBUILD_CORE) -o $(GOBUILDPATHINCONTAINER)/$(GOBUILDMAKEPATH_CORE)/$(CORE_BINARYNAME)


zyyw:
Thank you for the PR. But the local Go module cache doesn't have to be this path `/go/pkg/mod`.

MinerYang:
Hi @palhotel,

Thanks for your contributions!
We prefer to add this as an option and do not modified current behavior.

Best,
Miner

wy65701436:
While this could offer benefits for developers, from a build perspective, we prioritize ensuring the binary is built from scratch to mitigate potential issues stemming from cached builds. 

So, I propose providing developers with an option to accelerate the build speed if desired.

palhotel:
Thanks for the feedback, @MinerYang @wy65701436. I've added the option as suggested. Could you please review it again?

Issue #19817, opened 2024-01-09
Some labels are not displayed in the label tooltip box in Artifacts list page


xuelichao:
If you are reporting a problem, please make sure the following information are provided:

**Expected behavior and actual behavior:**
A clear and concise description of what you expected to happen and what's the actual behavior. If applicable, add screenshots to help explain your problem.

User should be able to check all the labels in the tooltip box. maybe should add a vertical scroll bar there and limit the max-height
of the tooltip box.

![Screenshot 2024-01-09 at 15 29 51](https://github.com/goharbor/harbor/assets/68891670/119313e4-49cd-4241-ba03-c220bba0ac9c)

**Steps to reproduce the problem:**
Please provide the steps to reproduce this problem.

1. Navigate to Artifacts page and add 12 labels and assign the labels to one Artifacts item.
2. Then in the Lables column, click the  Icon '...' to open label tooltip box, some labels are not displayed or maybe hidden by other
layer.

**Versions:**
Please specify the versions of following systems.

- harbor version: [v2.10.0-aa1f79cd]
- docker engine version: [y.y.y]
- docker-compose version: [z.z.z]

**Additional context:**

- **Harbor config files:** You can get them by packaging `harbor.yml` and files in the same directory, including subdirectory.
- **Log files:** You can get them by package the `/var/log/harbor/` .

PR #20458, opened 2024-05-21
fix: label pop up with overflow control

ZOTBLX:
# Summary of your change
- We tried to fix the style issue mentioned in issue #19817.
- We add some inner style to overwrite the Clarity UI.
- We set the height of the pop-up to be fixed, so the scroll bar will be displayed.

# Issue being fixed
Fixes #(19817)


Vad1mo:
@ZOTBLX dco is missing, 

can you also attach before after screenshots to easier verify the changes?

xuelichao (approved):
LGTM

PR #20491, opened 2024-05-28
Add update role test case

ZOTBLX:
# Summary of your change

We add a new test case in test_user_crud, to test the ability to update a user’s role to sysadmin and checks the reflection of this change


zyyw:
This test case has been covered here:
https://github.com/goharbor/harbor/blob/6d782ae69572af58fcdef6acc2ee8faa6dc7242a/tests/apitests/python/test_user_crud.py#L69-L72

PR #20492, opened 2024-05-28
Test update user password

ZOTBLX:
#  Summary of your change

Test the password update functionality, make sure that users can only login with the updated credentials.


zyyw:
Update user password has been covered here:
https://github.com/goharbor/harbor/blob/6d782ae69572af58fcdef6acc2ee8faa6dc7242a/tests/apitests/python/test_user_crud.py#L74-L84

PR #20493, opened 2024-05-28
add delete user test

ZOTBLX:
# Summary of your change

Test that the user can be deleted successfully and after being deleted, accessing to that user will throw error.

### DIFF
--- a/tests/apitests/python/test_user_crud.py
+++ b/tests/apitests/python/test_user_crud.py
@@ -90,6 +90,24 @@ class TestUser(unittest.TestCase):
         # 11. Delete user(UA);
         self.user.delete_user(user_id, **ADMIN_CLIENT)
         self.user.get_user_by_id(user_id, expect_status_code=404, expect_response_body="user {} not found".format(user_id), **ADMIN_CLIENT)
+        
+    def test_delete_user_and_error_handling(self):
+        """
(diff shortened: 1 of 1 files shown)


tests/apitests/python/test_user_crud.py
```diff
@@ -90,6 +90,24 @@ def testUser(self):
...
+            3. Attempt to retrieve the deleted user should return an error.
+        """
+        # 1. Create a new user;
+        user_id, user_name = self.user.create_user(**ADMIN_CLIENT)
+        #  2. Delete the user;
+        self.user.delete_user(user_id, **ADMIN_CLIENT)
+        # 3. Attempt to retrieve the deleted user should return an error.
+        with self.assertRaises(Exception) as context:
+            self.user.get_user_by_id(user_id, expect_status_code=404, **ADMIN_CLIENT)
+        self.assertIn("user not found", str(context.exception), "Expected error message not found.")
```
zyyw, line 110: line 110 might need to be aligned with the same indentation level as line 109

zyyw:
Thank you for contributing to Harbor! This delete user test case has been covered here:
* https://github.com/goharbor/harbor/blob/main/tests/apitests/python/test_user_crud.py#L90-L92

Issue #19401, opened 2023-09-27
Loss of images during GC when Postgresql server runs out of connections


dmitry-g:
Hello,

It has been the second time this month when we faced an issue with image loss during garbage collection. It probably has something to do with error handling, since the issue happens only when our database server runs out of connections.

Retention policy is configured as `For the repositories matching **, retain the most recently pushed 30 artifacts with tags matching **`, but garbage collector deletes the latest artifact, even regardless the fact that the repository had only 3 artifacts in total.

1. Garbage collection starts at the defined schedule `01:00 UTC`

2. Postgresql database runs out of available connections `sorry, too many clients already (SQLSTATE 53300)`

3. Instead of failing/retrying/handling the error, garbage collector choses to delete the latest artifact.

Also in the logs provided below you can see that tags of some artifacts have not been loaded, though all of them were actually tagged.

<details>
<summary>Image loss on 2023-09-06</summary>

```
2023-09-06T01:00:03Z [INFO] [/pkg/retention/job.go:86]: Run retention process.
 Repository: project-name/repository-name/image-name 
 Rule Algorithm: or 
 Dry Run: false
2023-09-06T01:00:03Z [INFO] [/pkg/retention/job.go:101]: Load 3 candidates from repository project-name/repository-name/image-name
2023-09-06T01:00:04Z [ERROR] [/pkg/retention/job.go:126]: Refresh quota error after deleting candidates, error: failed to connect to `host=db-host user=db-user database=db-name`: server error (FATAL: sorry, too many clients already (SQLSTATE 53300))
2023-09-06T01:00:04Z [INFO] [/pkg/retention/job.go:212]: 
|                                 Digest                                  | Tag | Kind  | Labels |     PushedTime      |     PulledTime      |     CreatedTime     | Retention |
|-------------------------------------------------------------------------|-----|-------|--------|---------------------|---------------------|---------------------|-----------|
| sha256:ea8cf3b41ddd07879296245f8fd66892cccf240a7fea6d6bedf01c519b224091 |     | image |        | 2023/09/04 07:14:10 | 2023/09/05 12:50:27 | 2023/09/04 07:14:10 | DEL       |
| sha256:76f8270aafef4cad4470e73a0a374f8367500c7fa0c02a036f17b5effa42dd64 |     | image |        | 2023/04/21 10:20:51 | 2023/09/05 11:15:07 | 2023/04/21 10:20:51 | ERR       |
[...]

PR #20555, opened 2024-06-05
bump up beego to v2.2.1

MinerYang:

 Upstream behavior change:
 - https://github.com/uzziahlin/beego/blob/b9b8b8d2a9ec9f1b6e1a76b53b053d9cb7187b91/client/orm/db.go#L1375


zyyw:
@MinerYang please resolve the conflict.

src/lib/orm/query.go
```diff
@@ -64,7 +64,7 @@ import (
 //			},
 //		 }
 //	}
-func QuerySetter(ctx context.Context, model interface{}, query *q.Query) (orm.QuerySeter, error) {
+func QuerySetter(ctx context.Context, model interface{}, query *q.Query, option ...string) (orm.QuerySeter, error) {
```
Vad1mo, line 67: 1. **Error Messages**: The error messages could be more descriptive. For example, instead of saying `"<orm.QuerySetter> cannot use non-ptr model struct `%s`"`, you could say something like `"Expected a pointer to a model struct, but got a non-pointer struct: %s"`.

2. **Magic String**: The string "disableSort" is a magic string. It would be better to define it as a constant at the top of your file or package.

3. **Function Complexity**: The function is doing a lot of things - it's setting filters, sorting, and handling pagination. It might be better to split these into separate functions to reduce the complexity and improve readability.

4. **Type Checking**: The function assumes that the `model` parameter is a pointer. It might be better to add a check to ensure that it is a pointer before trying to use it.

Here's how you might refactor the function:

```go
type Option int

const (
	DisableSort Option = iota
	// Add other options here
)

func QuerySetter(ctx context.Context, model interface{}, query *q.Query, option ...Option) (orm.QuerySeter, error) {
	if err := validateModel(model); err != nil {
		return nil, err
	}

	ormer, err := FromContext(ctx)
	if err != nil {
		return nil, err
	}

	qs := ormer.QueryTable(model)
	if query == nil {
		return qs, nil
	}

	metadata := parseModel(model)
	qs = setFilters(ctx, qs, query, metadata)

	if !isSortDisabled(option) {
		qs = setSorts(qs, query, metadata)
	}

	return setPagination(qs, query), nil
}

func validateModel(model interface{}) error {
	t := reflect.TypeOf(model)
	if t.Kind() != reflect.Ptr {
		return fmt.Errorf("Expected a pointer to a model struct, but got a non-pointer struct: %s", getFullName(t.Elem()))
	}
	return nil
}

func isSortDisabled(option []Option) bool {
	for _, opt := range option {
		if opt == DisableSort {
			return true
		}
	}
	return false
}

func setPagination(qs orm.QuerySeter, query *q.Query) orm.QuerySeter {
	if query.PageSize > 0 {
		qs = qs.Limit(query.PageSize)
		if query.PageNumber > 0 {
			qs = qs.Offset(query.PageSize * (query.PageNumber - 1))
		}
	}
	return qs
}
```

This refactored version separates the concerns into smaller functions, making it easier to understand and maintain. It also uses a constant for the "disableSort" option, which makes the code more robust.

Vad1mo: Magic Strings are bad thing to do.

Vad1mo (dismissed):

src/lib/orm/query.go
wy65701436: @Vad1mo regarding the code refactor, we could raise another PR to handle it, and let's use this to focus on the beego upgrade. I'd like to have it be merged, so then we could do more testings, especially the regression parts.

src/lib/orm/query.go
```diff
@@ -64,7 +64,14 @@ import (
 //			},
 //		 }
 //	}
-func QuerySetter(ctx context.Context, model interface{}, query *q.Query) (orm.QuerySeter, error) {
+
+type Option int
```
wy65701436, line 68: type Option struct {
      DisableSort bool
}

wy65701436 (approved):
lgtm

chlins (approved):
lgtm

src/lib/orm/query.go
Vad1mo: @wy65701436, Its never going to happen, I know it you know it. Another missed oppertunity, to get better.

Vad1mo:
resolves #19401

PR #20563, opened 2024-06-08
New feature: star project

ZOTBLX:
# Summary
This pull request introduces the "Star Project" feature to the project management interface. This new feature allows users to mark specific projects as "starred," making them quickly accessible from a potentially large list of projects. This functionality aims to enhance user experience by allowing users to easily keep track of and navigate to their frequently accessed or favorite projects.

# Key functionalities
- We make it like the delete functionality, user can select the projects they want to star, or say pinpoint, and add these projects as star projects.

# Issues
- We cannot add the database fields so this feature is not complete.

### DIFF
--- a/src/controller/project/controller.go
+++ b/src/controller/project/controller.go
@@ -64,6 +64,7 @@ type Controller interface {
 	Update(ctx context.Context, project *models.Project) error
 	// ListRoles lists the roles of user for the specific project
 	ListRoles(ctx context.Context, projectID int64, u *commonmodels.User) ([]int, error)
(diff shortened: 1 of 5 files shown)


Vad1mo:
@ZOTBLX the DCO is missing, 
Can you provide some more background as wells as designs and mockups that might help understand the problem and asses whether the provided solution meet users' demands?

wy65701436:
Thanks for the contribution. However, as for a new feature, we prefer to have a proposal first. Once the maintainer remove and approve the proposal, then we can start code working.

PR #20622, opened 2024-06-19
Simplify registry build

kariya-mitsuru:

1. `make/photon/Makefile`

   Remove `cd` command since the current directory is changed in `make/photon/registry/builder`.

2. `make/photon/registry/Dockerfile.base`

   Change the owner of `/etc/pki/tls/certs` when building the base image.
   (Since files under `/etc/pki/tls/certs` are not affected by binary build.)

3. `make/photon/registry/Dockerfile`

   - Remove changing the owner of `/etc/pki/tls/certs` since the change has been moved to the base image.
   - The owner/permission changes of files copied from the context are now performed simultaneously when the `COPY` command is executed.
     (If `COPY` and the owner/permission changes were separated, both image layers before and after change would be created, making the image unnecessarily large.)
   - Add `--link` option to `COPY` command.
     (This will improve image build efficiency since the base image will not be extracted at build time.)

4. `make/photon/registry/builder`.

   - Move `set -e` (exit immediately on error) to the top.
   - There is no `error` command, so change it to the `echo` command.
   - Remove `cur` variables that are no longer used by using `~-` and `cd -`.
   - Add `--depth 1` option to `git clone` command.
     (Since we only need the specified version of the source to build, we don't need the whole history, and this reduces the amount of transfer at clone time.)
   - Change the `docker build` command to specify the source file directly with the `-f` option instead of copying `Dockerfile.binary`.
   - Change the `docker build` command to directly output the binary file without creating a container image and container by specifying the output directory with the `-o` option.

5. `make/photon/registry/Dockerfile.binary`.

   - Remove `PREFIX` variable specified at `make` command line, since it is not used.
   - Change `make` build target from `binaries` to `bin/registry`.
     (Since `binaries` other than `bin/registry` are not used.)
   - Add a stage to extract only binary files, since `make/photon/registry/builder` now outputs binary files directly.

# Issue being fixed


Vad1mo (approved):

make/photon/registry/Dockerfile.base
```diff
@@ -3,4 +3,5 @@ FROM photon:5.0
 RUN tdnf install -y shadow >> /dev/null \
     && tdnf clean all \
     && mkdir -p /etc/registry \
-    && groupadd -r -g 10000 harbor && useradd --no-log-init -m -g 10000 -u 10000 harbor
+    && groupadd -r -g 10000 harbor && useradd --no-log-init -m -g 10000 -u 10000 harbor \
+    && chown -R harbor:harbor /etc/pki/tls/certs
```
wy65701436, line 7: why moves chown -R harbor:harbor /etc/pki/tls/certs into Dockerfile.base?

kariya-mitsuru: The files under `/etc/pki/tls/certs` are not affected by binary builds (or any changes to the harbor sources).
They are only affected by the base Photon image.
Therefore, it is not necessary to chown the registry-photon image every time it is created.

make/photon/registry/Dockerfile
```diff
@@ -2,14 +2,9 @@ ARG harbor_base_image_version
...
-COPY ./make/photon/common/install_cert.sh /home/harbor
-COPY ./make/photon/registry/entrypoint.sh /home/harbor
-COPY ./make/photon/registry/binary/registry /usr/bin/registry_DO_NOT_USE_GC
-
-RUN chown -R harbor:harbor /etc/pki/tls/certs \
-    && chown harbor:harbor /home/harbor/entrypoint.sh && chmod u+x /home/harbor/entrypoint.sh \
-    && chown harbor:harbor /home/harbor/install_cert.sh && chmod u+x /home/harbor/install_cert.sh \
-    && chown harbor:harbor /usr/bin/registry_DO_NOT_USE_GC && chmod u+x /usr/bin/registry_DO_NOT_USE_GC
+COPY --link --chown=10000:10000 --chmod=755 ./make/photon/common/install_cert.sh /home/harbor/
+COPY --link --chown=10000:10000 --chmod=744 ./make/photon/registry/entrypoint.sh /home/harbor/
```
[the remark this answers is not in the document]
kariya-mitsuru: I don't know why this is the case, but in the current image, the permission of `entrypoint.sh` is `744`.
The permission of `entrypoint.sh` is `644` on git, and execute `chmod u+x`(= +100) on it, so the result is `744`.
I don't think changing the permissions of `entrypoint.sh` to `755` will cause any problems, but I'm not sure.

Note that the permission of `install_cert.sh` is `755` on git, so it will be `755` whether or not `chmod u+x` is executed.

make/photon/registry/Dockerfile.base
wy65701436: this is the dockerfile for base image, it will no be built every time but on demand to refresh the OS libraries. 

I prefer to put all the permission handling together, it's easy to check and maintain. Otherwise, we have to know that there is a separate one in the base image.

make/photon/registry/builder
```diff
@@ -1,44 +1,32 @@
...
 mkdir -p binary
 rm -rf binary/registry || true
 
 cd `dirname $0`
-cur=$PWD
 
 # the temp folder to store distribution source code...
 TEMP=`mktemp -d ${TMPDIR-/tmp}/distribution.XXXXXX`
-git clone -b $VERSION https://github.com/distribution/distribution.git $TEMP
+git clone -b $VERSION --depth 1 https://github.com/distribution/distribution.git $TEMP
```
wy65701436, line 20: why is it necessary to specify the depth here?

make/photon/registry/Dockerfile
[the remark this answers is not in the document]
wy65701436: In case any regression problem, I prefer to keep the original writing. But I agree that we optimize the docker file to reduce size.

make/photon/registry/builder
```diff
@@ -1,44 +1,32 @@
...
 
 # add patch redis
 cd $TEMP
-git apply $cur/redis.patch
-cd $cur
+git apply ~-/redis.patch
```
wy65701436, line 24: does it support all the linux OS versions?

make/photon/registry/Dockerfile.binary
```diff
@@ -7,4 +7,7 @@ ENV GO111MODULE auto
 WORKDIR $DISTRIBUTION_DIR
 COPY . $DISTRIBUTION_DIR
 
-RUN CGO_ENABLED=0 make PREFIX=/go clean binaries
+RUN CGO_ENABLED=0 make clean bin/registry
+
+FROM scratch
+COPY --from=build /go/src/github.com/docker/distribution/bin/registry /
```
wy65701436, line 13: to /bin/binary?

make/photon/registry/Dockerfile.binary
```diff
@@ -7,4 +7,7 @@ ENV GO111MODULE auto
 WORKDIR $DISTRIBUTION_DIR
 COPY . $DISTRIBUTION_DIR
 
-RUN CGO_ENABLED=0 make PREFIX=/go clean binaries
+RUN CGO_ENABLED=0 make clean bin/registry
+
```
wy65701436, line 11: why is this change needed?

make/photon/registry/Dockerfile.base
kariya-mitsuru: > this is the dockerfile for base image, it will no be built every time but on demand to refresh the OS libraries.

I understand that.

If `chown -R harbor:harbor /etc/pki/tls/certs` is in the `Dockerfile` instead of the `Dockerfile.base` (before this fix), it will be executed on every image build even if the OS libraries are not changed. This is obviously a wasteful process.
With this fix, `chown -R harbor:harbor /etc/pki/tls/certs` will be executed only if there are changes in OS libraries.

However, it's not a significant process, and I'm not particularly particular about it, so I'll revert it.

make/photon/registry/Dockerfile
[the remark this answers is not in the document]
kariya-mitsuru: If we are willing to use a bind mount, we can do the following.
I have confirmed on local environment that image size can be reduced this way.
```
RUN --mount=type=bind,source=make/photon,target=/root \
    chown -R harbor:harbor /etc/pki/tls/certs \
    && cp /root/common/install_cert.sh /home/harbor \
    && cp /root/registry/entrypoint.sh /home/harbor \
    && cp /root/registry/binary/registry /usr/bin/registry_DO_NOT_USE_GC \
    && chown harbor:harbor /home/harbor/entrypoint.sh && chmod u+x /home/harbor/entrypoint.sh \
    && chown harbor:harbor /home/harbor/install_cert.sh && chmod u+x /home/harbor/install_cert.sh \
    && chown harbor:harbor /usr/bin/registry_DO_NOT_USE_GC && chmod u+x /usr/bin/registry_DO_NOT_USE_GC
```
Would this method be better?

make/photon/registry/builder
kariya-mitsuru: If `--depth 1` is not specified, then the entire history of all the commits to the `$VERSION` will be cloned, which is unnecessary and is wasteful in terms of transfer amount and processing time (and very painful for a poor environment like mine).
By specifying `--depth 1`, only the commits specified by `$VERSION` can be retrieved.

make/photon/registry/builder
kariya-mitsuru: Yes.
It is supported by all currently used bash.

make/photon/Makefile
```diff
@@ -178,7 +178,7 @@ _build_registry:
 		rm -rf $(DOCKERFILEPATH_REG)/binary && mkdir -p $(DOCKERFILEPATH_REG)/binary && \
 		$(call _get_binary, $(REGISTRYURL), $(DOCKERFILEPATH_REG)/binary/registry); \
 	else \
-		cd $(DOCKERFILEPATH_REG) && $(DOCKERFILEPATH_REG)/builder $(REGISTRY_SRC_TAG) && cd - ; \
+		$(DOCKERFILEPATH_REG)/builder $(REGISTRY_SRC_TAG) ; \
```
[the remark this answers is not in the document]
kariya-mitsuru: No.
The builder script can be executed in the location of `$(DOCKERFILEPATH_REG)`, but it also can be executed in any other location, so it need not to cd to `$(DOCKERFILEPATH_REG)`.

make/photon/registry/Dockerfile.binary
kariya-mitsuru: The `PREFIX` variable is not used anywhere, which leads to confusion.
Therefore, I think it is better to remove it.

make/photon/registry/Dockerfile.binary
kariya-mitsuru: It was originally created in `binary`, not `bin/binary`, does that mean it should be changed to `bin/binary`?

make/photon/registry/Dockerfile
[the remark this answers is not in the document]
chlins: I think this way looks more clean and similar with previous.

make/photon/registry/Dockerfile.base
kariya-mitsuru: Pushed.

make/photon/registry/Dockerfile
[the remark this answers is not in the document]
kariya-mitsuru: Pushed.

chlins (approved):
lgtm

chlins:
Take note that this modification relies on certain new features of Docker Buildkit. As a result, you might encounter an error when building with an older version of Docker.

Vad1mo:
> Take note that this modification relies on certain new features of Docker Buildkit. As a result, you might encounter an error when building with an older version of Docker.

Do you have an idea how new the docker version should be?

I scanned the Dockerfile, and based on the output it should run on Docker 17.06 CE (June 2017)

We should be pritty save on that side

chlins:
AFAIK, `RUN --mount` was introduced since https://docs.docker.com/build/buildkit/dockerfile-release-notes/#130, and the corresponding docker version should be [v19.03](https://docs.docker.com/engine/release-notes/19.03/).

kariya-mitsuru:
> I scanned the Dockerfile, and based on the output it should run on Docker 17.06 CE (June 2017)

Thank you for pointing this out.

According to `make/common.sh`, docker command must be `20.10.10` or later, so I checked with `20.10.10` and found that `DOCKER_BUILDKIT=1` must be specified.

So, I have pushed it to specify `DOCKER_BUILDKIT=1`.

Although buildkit is enabled by default in docker `23.0.0` or later, even if `DOCKER_BUILDKIT=0` is specified in the build environment (which is not usually the case), this push allows normal builds.

bupd:
merge conflicts!

PR #20623, opened 2024-06-19
Simplify trivy-adapter build

kariya-mitsuru:

1. `make/photon/Makefile`

   Remove `cd` command since the current directory is changed in `make/photon/trivy-adapter/builder.sh`.

2. `make/photon/trivy-adapter/Dockerfile.base`

   Change the owner of `/etc/pki/tls/certs` when building the base image.
   (Since files under `/etc/pki/tls/certs` are not affected by binary build.)

3. `make/photon/trivy-adapter/Dockerfile`

   - Remove changing the owner of `/etc/pki/tls/certs` since the change has been moved to the base image.
   - The owner/permission changes of files copied from the context are now performed simultaneously when the `COPY` command is executed.
     (If `COPY` and the owner/permission changes were separated, both image layers before and after change would be created, making the image unnecessarily large.)
   - Add `--link` option to `COPY` command.
     (This will improve image build efficiency since the base image will not be extracted at build time.)

4. `make/photon/trivy-adapter/builder.sh`.

   - Move `set -e` (exit immediately on error) to the top.
   - There is no `error` command, so change it to the `echo` command.
   - Remove `cur` variables that are no longer used.
   - Add `-b $VERSION` option to `git clone` command and remove `git checkout` command.
   - Add `--depth 1` option to `git clone` command.
     (Since we only need the specified version of the source to build, we don't need the whole history, and this reduces the amount of transfer at clone time.)
   - Change the `docker build` command to specify the source file directly with the `-f` option instead of copying `Dockerfile.binary`.
   - Change the `docker build` command to directly output the binary file without creating a container image and a container by specifying the output directory with the `-o` option.

5. `make/photon/trivy-adapter/Dockerfile.binary`.

   Add a stage to extract only binary files, since `make/photon/trivy-adapter/builder.sh` now outputs binary files directly.

# Issue being fixed

### DIFF
--- a/make/photon/trivy-adapter/Dockerfile
+++ b/make/photon/trivy-adapter/Dockerfile
@@ -4,17 +4,11 @@ FROM ${harbor_base_namespace}/harbor-trivy-adapter-base:${harbor_base_image_vers
 
 ARG trivy_version
 
-COPY ./make/photon/common/install_cert.sh /home/scanner
-COPY ./make/photon/trivy-adapter/entrypoint.sh /home/scanner
-COPY ./make/photon/trivy-adapter/binary/trivy /usr/local/bin/trivy
-COPY ./make/photon/trivy-adapter/binary/scanner-trivy /home/scanner/bin/scanner-trivy
+COPY --link --chown=10000:10000 --chmod=755 ./make/photon/common/install_cert.sh /home/scanner/
+COPY --link --chown=10000:10000 --chmod=744 ./make/photon/trivy-adapter/entrypoint.sh /home/scanner/
+COPY --link --chown=10000:10000 --chmod=755 ./make/photon/trivy-adapter/binary/trivy /usr/local/bin/
+COPY --link --chown=10000:10000 --chmod=755 ./make/photon/trivy-adapter/binary/scanner-trivy /home/scanner/bin/
 
-
-RUN chown -R scanner:scanner /etc/pki/tls/certs \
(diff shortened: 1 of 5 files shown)


make/photon/trivy-adapter/Dockerfile
```diff
@@ -4,18 +4,12 @@ FROM ${harbor_base_namespace}/harbor-trivy-adapter-base:${harbor_base_image_vers
 
 ARG trivy_version
 
-COPY ./make/photon/common/install_cert.sh /home/scanner
-COPY ./make/photon/trivy-adapter/entrypoint.sh /home/scanner
-COPY ./make/photon/trivy-adapter/binary/trivy /usr/local/bin/trivy
-COPY ./make/photon/trivy-adapter/binary/scanner-trivy /home/scanner/bin/scanner-trivy
+COPY --link --chown=10000:10000 --chmod=755 ./make/photon/common/install_cert.sh /home/scanner/
```
wy65701436, line 7: the same comments on the registry pull request.

make/photon/trivy-adapter/Dockerfile.base
```diff
@@ -3,4 +3,5 @@ FROM photon:5.0
 RUN tdnf install -y rpm shadow >> /dev/null \
     && tdnf clean all \
     && groupadd -r -g 10000 scanner \
-    && useradd --no-log-init -m -r -g 10000 -u 10000 scanner
+    && useradd --no-log-init -m -r -g 10000 -u 10000 scanner \
+    && chown -R scanner:scanner /etc/pki/tls/certs
```
wy65701436, line 7: ditto

PR #20709, opened 2024-07-07
`gorilla/mux` reorganize

Its-Maniaco:
Reorganized all the references of `gorilla/mux` into seperate file for each package

Partial Fix #19965


Its-Maniaco:
@chlins @Vad1mo based on my understanding I made these changes.
There is a bit of code that I reused in different files. Should I create a separate package for `gorilla/mux `related code or is there any other alternative you would recommend?

PR #20723, opened 2024-07-10
replaced query.Sorting with query.Sorts

### LINKED ISSUE #19569: Removing `Query.Sorting` as it we are deprecating it

Its-Maniaco:
Fixes #19569 

Replaced deprecated `query.Sorting` with `query.Sorts` in `src/pkg/scan/dao/scanner/registration.go`

### DIFF
--- a/src/pkg/scan/dao/scanner/registration.go
+++ b/src/pkg/scan/dao/scanner/registration.go
@@ -123,10 +123,18 @@ func ListRegistrations(ctx context.Context, query *q.Query) ([]*Registration, er
 	}
 
 	// Order the list
-	if query.Sorting != "" {
-		qs = qs.OrderBy(query.Sorting)
+	if len(query.Sorts) > 0 {
+		sortKey := ""
+		for _, sort := range query.Sorts {
+			if sort.DESC {
+				sortKey += " -" + sort.Key
+			} else {
+				sortKey += " " + sort.Key
+			}
+		}
+		qs = qs.OrderBy(sortKey)
 	} else {
-		qs = qs.OrderBy("-is_default", "-create_time")
+		qs = qs.OrderBy("-is_default", "-create_time") // -column means descending
 	}
 
 	l := make([]*Registration, 0)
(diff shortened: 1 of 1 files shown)


Its-Maniaco:
@stonezdj @zyyw Can you please re-review?
I have made changes based on last suggestion from [here](https://github.com/goharbor/harbor/pull/19962#discussion_r1557157603).
Apologies for the delay.
My last PR was closed due to inactivity, so reopened a new one

src/pkg/scan/dao/scanner/registration.go
```diff
@@ -123,10 +123,18 @@ func ListRegistrations(ctx context.Context, query *q.Query) ([]*Registration, er
 	}
 
 	// Order the list
-	if query.Sorting != "" {
-		qs = qs.OrderBy(query.Sorting)
+	if len(query.Sorts) > 0 {
```
chlins, line 126: I think the following logic is duplicated because when call the `orm.QuerySetter`, there will be a common `setSorts` function to handle it. https://github.com/goharbor/harbor/blob/0da13ebd28d702e35751860327b8dec38a313ad0/src/lib/orm/query.go#L153

Its-Maniaco: So, the `Query Setter` is already returning in sorted order and we just have to replace the explicit Sorting from the code?

chlins: Yes, since there is a common logic processing, there is no need for redundant processing here.

src/pkg/scan/dao/scanner/registration.go
```diff
@@ -121,14 +121,6 @@ func ListRegistrations(ctx context.Context, query *q.Query) ([]*Registration, er
 	if err != nil {
 		return nil, err
 	}
-
-	// Order the list
-	if query.Sorting != "" {
-		qs = qs.OrderBy(query.Sorting)
-	} else {
-		qs = qs.OrderBy("-is_default", "-create_time")
```
chlins, line 129: Please check if we need to keep this line to ensure backward compatibility without introducing inconsistent results from previous versions, because the common query logic processor will not handle this special case.

PR #20778, opened 2024-07-26
registry-photon: transparently pass arguments in entrypoint.sh.

aznashwan:

Transparently pass arguments of the `registry-photon` image's `entrypoint.sh` to the registry binary to allow for user-overridable arguments for it.

# Issue being fixed

This issue on `harbor-helm` fully details the problem: https://github.com/goharbor/harbor-helm/issues/1801

TL;DR: it'd be nice to be able to pass arguments to the `registry` binary in the `registry-photon` Docker image (most notably the config path) while still ensuring all the current and future setup steps in `entrypoint.sh` will still get executed.

     - the linked issue should explain the problem clearly enough
    - I can't label the PR from the creation stage...
    - works with current state of the upstream `goharbor/harbor-helm` chart on my machine ™️ 
    - this change should be completely backwards-compatible.

### DIFF
--- a/make/photon/registry/Dockerfile
+++ b/make/photon/registry/Dockerfile
@@ -16,5 +16,6 @@ HEALTHCHECK CMD curl --fail -s http://localhost:5000 || curl -k --fail -s https:
 USER harbor
 
 ENTRYPOINT ["/home/harbor/entrypoint.sh"]
+CMD ["serve", "/etc/registry/config.yml"]
 
 VOLUME ["/storage"]
--- a/make/photon/registry/entrypoint.sh
+++ b/make/photon/registry/entrypoint.sh
@@ -10,4 +10,4 @@ set -e
 
 /home/harbor/install_cert.sh
 
-exec /usr/bin/registry_DO_NOT_USE_GC serve /etc/registry/config.yml
+exec /usr/bin/registry_DO_NOT_USE_GC $@


Kajot-dev:
I don't know if the condition with the default args is needes. Combining `ENTRYPOINT` (which just passes all args) + `CMD` (with the defeault values) in Dockerfile would be enough

aznashwan:
> I don't know if the condition with the default args is needes. Combining ENTRYPOINT (which just passes all args) + CMD (with the defeault values) in Dockerfile would be enough

Done, PTAL!

MinerYang:
Hi @aznashwan ,

Thanks for your contribution and as you proposed that this `args` settings in the [harbor-helm-registry.template ](https://github.com/goharbor/harbor-helm/blob/618295199a74448626a8cafe3beb84dfee1378b9/templates/registry/registry-dpl.yaml#L99)would not override the config yml you changed in registry-dpl indeed.

However this `/etc/registry/config.yml` file more like a intermediate things that generate from exposed configurations provided in `values.yml` and we preferred do not let this file to be totally customized from outside to keep all the registry configurations to be compatible with corresponding upstream distribution versions.

Instead we prefer to change the registry related config from values.yml or at least registry configmap.

Best,
Miner

aznashwan:
> However this /etc/registry/config.yml file more like a intermediate things that generate from exposed configurations provided in values.yml

That is fair, I will close this PR and open one against the Helm chart to remove the ignored args.

PR #20851, opened 2024-08-14
Extend oidc_cli security context generator to include calls to the v2…

### LINKED ISSUE #14236: Harbor API access via OIDC from other clients not possible
We have configured OIDC as authentication mechanism for Harbor. Also, we have several other applications that are consuming the Harbor API.

With OIDC enabled, authentication against the Harbor API is only possible with a valid OpenID Connect ID Token as Bearer Authentication header.

As far as I understood the OpenID Connect specification, the ID Token should not be used to grant access to ressources - like the Harbor API. The Access Token is meant for that.

Harbor validates the ID Token for API requests using the go-oidc library internally. This library has [SkipClientIDCheck](https://github.com/goharbor/harbor/blob/51c8375425f2677ca683bc94f04c9bd423f2d287/src/vendor/github.com/coreos/go-oidc/verify.go#L257) enabled by default (which is totally fine for ID Tokens). 

This leads to the fact, that Harbor does not allow access to the API with valid tokens, issued by other clients from the same issuer. You are forced to use Harbors OIDC client credentials in other applications to be able to interact with the Harbor API.

**Expected behavior:**
I would have expected that Harbor uses an Access Token to grant access to API resources rather than the ID Token. Then this should not be an issue, right?

I can see two possible solutions:
- Change authorization of Harbor API to use Access Tokens instead of ID Tokens
- Allow setting the `SkipClientIDCheck` via Harbor Configuration (leads to a possible security leak)

- Enable OIDC auth in Harbor with credentials for ClientA
- Fetch an ID Token from the same Issuer with client credentials for ClientB
- Make an API call against the Harbor API with that ID Token issued by ClientB

Harbor will fail with `failed to verify token: oidc: expected audience ClientA got ClientB`

**Versions:**

- harbor version: [v2.1.3]
- docker engine version: [19.03.13]
- docker-compose version: [1.24.0]

LGhoull:

- Extend oidc_cli security context generator to include calls to the v2 API
- remove idtoken security context generator
- rename and consolidate tests and names accordingly

Id Tokens can no longer be used to authenticate against the v2 API
Instead, basic auth with the CLI secret as password must be used.

The use of the OIDC ID token as a bearer token for authentication against the API is inconvenient since the ID token is hard to get as a normal user.

# Issue being fixed
Fixes #14236


LGhoull:
Definitely a breaking change, but I can't add labels.

reasonerjt:
Although the change is relatively small, b/c this is a break change and involves security. I wish to suggest that a proposal is needed for review and get approval from maintainers.

If we reach a consensus to do this, there should also be some change needed in UI and Documentation.

reasonerjt (changes requested):
Please write a proposal and get approval first.

LGhoull:
Proposal: https://github.com/goharbor/community/pull/250

LGhoull:
Still relevant

Issue #20532, opened 2024-05-31
Maximum parallel replications for pull


Joseph94m:
**Is your feature request related to a problem? Please describe.**
It's annoying and causes an avalanche effect when my pull based replications for the same Job start before the previous one is over.
![image](https://github.com/goharbor/harbor/assets/40126230/92a5aeea-0e61-4480-b07e-1dce1dec9e0b)

It starts off by taking 2 minutes, then it slowly starts piling up until it reaches hours...
![image](https://github.com/goharbor/harbor/assets/40126230/06751d33-a09b-410d-9bd3-2bf997d0a3f2)

**Describe the solution you'd like**
The ability to specify, in the replication configuration: 
* Maximum of replications waiting for an inProgress replication to finish
* Maximum replications that can be inProgress

**Describe the main design/architecture of your solution**

<img width="311" alt="image" src="https://github.com/goharbor/harbor/assets/40126230/5ee952c3-3fb2-401f-a3a0-b3abdd6a847c">

What do you guys think before we discuss implementation

stonezdj:
The cron string is `0 */2 * * * *`, it means that the replication is start every 2 minutes,  you should always keep the schedule interval longer than a single job complete time.

Joseph94m:
Hello, that's the fallback approach indeed.
But you cannot always guarantee a replication's time due to occasional bulk insertions in the source registry and network variability.

stonezdj:
In your case, most of the InProgress replication job is actually Pending, you could check the job in the job service dashboard,  they should in the job queue. you could adjust the job service worker count to control the max Parallel job.

bupd:
But at this point we can clearly see. It creates huge load on system and lot of wasted bandwidth. 
It is better to have a replication skip option.

bupd:
> The cron string is 0 */2 * * * *, it means that the replication is start every 2 minutes, you should always keep the schedule interval longer than a single job complete time.

The workaround of setting a longer replication interval, like once a day, fails to address the need for timely synchronization across registries. For users who rely on Harbor to maintain identical registries at different locations, frequent replication (e.g., every 5 minutes) is necessary to ensure minimal discrepancies between registries. By suggesting a longer interval, users may end up with outdated or inconsistent images, undermining the core functionality of replication.

Thanks.

Issue #19937, opened 2024-02-01
Replication job duplication


XANi:
Currently the replication job scheduler does not check whether the previous run ended before next one started which leads to it trying to sync the same repo multiple times if the previous job is not "fast enough" (say syncing big one the first time)

![image](https://github.com/goharbor/harbor/assets/147394/8edaa4b5-3dac-41c7-9ff9-6a9c9f23983b)

It would be nice if job option included "do not allow duplicate jobs"

zyyw:
@wy65701436 It might be a good idea to check if there is `In-Progress` replication job running before the same job being triggered again, resulting to multiple `In-Progress` jobs instance of the same replication job rule.

wy65701436:
I believe we should refrain from adding any checks for this, as each execution should ideally be idempotent and not impact any others.

XANi:
@wy65701436 They are already impacting eachother if the bandwidth is not enough to finish one before starting the next.

If job A1 can't finish in time, now the job A2 will have lower bandwidth, which will make it take even longer, so it will not finish before A3 start, which then will delay A3 etc. And if both start downloading same big image that's just wasted bandwidth.

Having option to specify "just single job at once" gives sort of dynamic balancing; you can set up sync to run often to get new images quickly but if one of them happens to be big all it will happen is next sync being a bit later.

Issue #17269, opened 2022-07-29
Replication: avoid overlapping replication jobs


okin:
In one of our instances one of our users had a replication set up that, in combination with #16656, led to nearly a thousand running replication jobs.
None of which was able to finish.
The logs showed similar entries like #15415.

The replication pulled every few minutes from another registry and when delays occured it lead to an cascading effect where the first replications weren't able to properly finished but new jobs have been started nontheless.

In order to prevent this I would like the possibility to avoid starting new jobs when for a specific replication when there is already a replication running.

I think one way to address this is to switch the way jobs are spawned from a specified schedule to creating jobs _X_ after the previous job has been finished.

chlins:
Currently the replication jobs are launched and executed by `jobservice`, harbor has the limit for `jobservice` workers, default value is 10 which control the max concurrency workers number. If all the workers are occupied and other jobs will be in pending status then wait until have free worker, so your proposal is expecting to run one replication job which has subtasks sequential rather than parallel?

chlins:
For the design of harbor replication, from the server side all the tasks each other as independent individuals, so it's not suitable to add dependency with each other such as wait for previous task, that's too tightly coupled a mechanism. But to solve your issue, i think you could consider it from client side, for example, you could call harbor API to fetch task status and only trigger replication when match your customize conditions.

okin:
@chlins I think that sequential processing of replication jobs makes sense at least if they are grouped per replication policy.
Right now I can't think of a reason why I would want to have overlapping replications that stem from the same policy. Different replication policies should still be able to run overlapping.

Thanks for the suggestion with working through the API. I'll have a look into this.

bupd:
> do we have any update on this topic - to enable sequential replications to a job?

Yes, I am currently working on it. Since a lot of people facing this problem. Will soon raise a PR with changes needed to fix this.

Thanks

Issue #16656, opened 2022-04-06
Replication taking longer each time


piroh1990:
Hello

Since we Updated from Harbor 2.2.1 to 2.4.1 and now to 2.4.2 there is a Problem with the replication. 
Every time the replication is running it takes slightly longer. 
Over time this will lead to a extremely long replication time. 

Syncing is done with another Harbor instance that runs on Version v2.1.4-e2426603

The Settings of the Replication is as following: 
![Replication_Rule](https://user-images.githubusercontent.com/57915454/161968526-c32febc3-e377-4cfb-8b75-111d9d8c78ea.jpg)

As a workaround we are running a cronjob that is deleting all the Charts with the following two lines. 
rm -f  /data/chart_storage/cm2/* 
rm -f /data/redis/dump.rdb

Does anyone know why this is happening and how we can fix it? 
If you need additional log files i will of course provide everything you need.

chlins:
@piroh1990 How many artifacts did you have, and also could you share the execution tab and job logs?

dioguerra:
I'm running the same version and the same happened to me while replicating helm charts from https://artifacthub.io/ locally.
It also happens that the chart replication fails, but then it starts again looping again and again (maybe this is related on me deleting the core pods)
Had to delete the core pods because this problem caused the service to go unresponsive with many in-flight requests pending
![image](https://user-images.githubusercontent.com/5084445/162786440-b6eeeb42-dbca-4d04-b667-8a27a4433ea8.png)

We saw an issue similar to the inflight requests accummulating here: https://github.com/goharbor/harbor/issues/15474 but we since updated and this didn't happen for a long time. now, this issue. Our configuration below:
![image](https://user-images.githubusercontent.com/5084445/162786970-c4e202b4-9d21-4997-8ec8-7069dac55290.png)
Our filter is set to: `{ceph-csi,cert-manager,cluster-autoscaler,coredns,deliveryhero,ingress-nginx,metrics-server,node-feature-discovery,kubernetes,prometheus-community,traefik}/**`

```
2022-04-11T01:10:14Z [ERROR] [/pkg/notifier/notifier.go:203]: Error occurred when triggering handler *artifact.ReplicationHandler of topic REPLICATION: project prometheus-community not found
2022-04-11T01:10:14Z [ERROR] [/controller/event/handler/webhook/artifact/replication.go:195]: failed to get project prometheus-community, error: project prometheus-community not found
[...]

PR #21347, opened 2024-12-22
Add Option to Skip Replication Policy Execution If an Execution is already running

bupd:
This PR fixes the below mentioned issues.
## Issues
- https://github.com/goharbor/harbor/issues/19937
- https://github.com/goharbor/harbor/issues/20532
- https://github.com/goharbor/harbor/issues/17269
- https://github.com/goharbor/harbor/issues/16656

## Related Issues
- https://github.com/goharbor/harbor/issues/15415
- https://github.com/goharbor/harbor/issues/7842

# Why do we need this

1. Users have for long requesting this feature to have single replication execution per policy.
2. Avoid overlapping replications which is common in using scheduled replications.
3. Without this replication starts piling up.
4. And at some point it becomes un-handleable... leading to a snowball effect on bandwidth and the queue. 
5. It makes no sense to run replication in parallel for same replication policy. 
6. It is better to atleast have an option to set replication to single execution at a time per policy.

## Changes Made
- Added `Skip if running` Checkbox in Replication UI.
- Added Execution skipping.
- Updated Replication Policy.
- Added `skip_if_running` column in sql.

### Todo
- [ ] Add Tests

Which also leads to greater cloud bills because of more bandwidth usage. Also it takes down deployments too.

it is better if we can fix this soon.

# Issue being fixed
Fixes #(issue)


wy65701436:
Thanks, @bupd, for your contribution! I believe this is a valuable new feature for Harbor that should follow the proposal process.

There are several questions we need to briefly discuss:

- What happens if the execution gets stuck in the 'running' status?
- How do we determine the relationship between the replication policy and the execution instance — by ID? And what happens if the policy content is updated after the execution begins?
- During the replication job (at least in Harbor-to-Harbor replication), Harbor skips artifacts that have already been successfully replicated. So, we should consider how much additional benefit this feature will bring.

Vad1mo (changes requested):

tpoxa:
> Thanks, @bupd, for your contribution! I believe this is a valuable new feature for Harbor that should follow the proposal process.
> 
> There are several questions we need to briefly discuss:
> 
> * What happens if the execution gets stuck in the 'running' status?

We were planning to use `JobServiceMonitorClient` to get  the current state of job observations. As long observation says it's running - a new execution will be skipped.  I believe we should fix the problem with stuck separately. What are your thoughts about that?

> * How do we determine the relationship between the replication policy and the execution instance — by ID? And what happens if the policy content is updated after the execution begins?

Indeed PolicyID was never needed before. We can add it anyway to the job arguments so we can scan observations, parse job arguments and match policyID with the current.

> * During the replication job (at least in Harbor-to-Harbor replication), Harbor skips artifacts that have already been successfully replicated. So, we should consider how much additional benefit this feature will bring.

I believe even checking an artefact for existence produces an additonal HTTP request. So, we still adding some extra efficiency.

Vad1mo:
> * During the replication job (at least in Harbor-to-Harbor replication), Harbor skips artifacts that have already been successfully replicated. So, we should consider how much additional benefit this feature will bring.

Regarding that statement. there is a fundamental deficit with harbor replications, when the layer is in the copy process, new replication will try to copy the same blob again. If you have a slow connection like DSL the same blob will be copied over and over again. This results in the domino effect. Each new replication halves the throughput, resulting in slower transfers, and because all replications run in parallel, the same effect happens with the next blob/layer. 


We have made an experiment with: 

- Copying 1 GiB image (https://hub.docker.com/r/vad1mo/1gb-random-file)
- 1 Mbit DSL connection
- Replication every minute
- Replication time with "Prevent Overlapping Replications" enabled: **xxx min**
- Replications time with "Prevent Overlapping Replications" disabled: **xxx min**

bupd:
Hello @wy65701436 

The below are my observations.
> - What happens if the execution gets stuck in the 'running' status?

While single active replication is checked. Replication executions per policy will become to atmost one. So if an execution is already running. other scheduled executions will be defered. 

> - How do we determine the relationship between the replication policy and the execution instance — by ID? And what happens if the policy content is updated after the execution begins?

1. each execution will have a replication policyID attached to it. so on executing the new jobs we check if there is any jobs that are already running with same execution policyID. if we find any jobs that are in progress. we will defer the executions until the previous job completes. 
2. By design, once harbor replication execution starts the policy content is fixed. and changes made to replication policy will only reflect in upcoming replication executions.


> - During the replication job (at least in Harbor-to-Harbor replication), Harbor skips artifacts that have already been successfully replicated. So, we should consider how much additional benefit this feature will bring.

**Yes exactly,** Harbor only skips artifacts that have already been successfully replicated. 
So in typical scenario harbor will try copying the same layers with different jobs pushing the same layer.

In the below scenario I tried replicating three images each of size: 512mb. but the resulting repository quota
resulting something similar like this. 

### Project Quota
Where the quota used should only be less than 1.5gb.

Screenshot_2025-01-08_16-02-46
**4.77GB is used** here for project due to replication jobs copying the same layer simultaneously. 
this clearly shows the quota consumed is greater than the images needed to be replicated.

### The below is the overall quota
Attaching here for reference.

Screenshot_2025-01-08_16-13-42


Thanks 

You can check the two instances used on this experiment: [reg1.bupd.xyz](https://reg1.bupd.xyz) & [reg2.bupd.xyz](https://reg2.bupd.xyz)

bupd:
# Observation on Replication Performance with and without Feature

### Speed Test Results:
Upload Speed: 9.50 Mbps (Data Used: 4.3 MB)
Latency: 37.96 ms (Jitter: 1.76 ms, Min: 25.11 ms, Max: 43.32 ms)
Packet Loss: 0.0%
Result URL: [Speedtest Result](https://www.speedtest.net/result/c/1a0cd346-17f1-4420-b563-b96c9f24b1e0)

##  Images to Replicate: Variations of [docker.io/vad1mo/1gb-random-file](docker.io/vad1mo/1gb-random-file)
## Workflow 1 (with Feature):

1 x image (512 MB)
3 x images (~1.5 GB)
From: reg1.bupd.xyz to reg2.bupd.xyz
Replication: Normal
Started: 2:27 PM
Completed: 2:49 PM
Bandwidth Used: 1.42 GB
Theoretical Time: 22.5 minutes for 1.5 GB
Actual Time: 22 minutes (No packet loss or bandwidth issues)

## Insight 
Replication with the feature enabled completed within the expected timeframe, showing stable upload speed and no packet loss.

## Workflow 2 (without Feature):
Started: 2:55 PM
Name: Destroyer
Bandwidth Used: 6 GB
Result: Failed
Time Taken: ~4+hrs

## Insight
The replication process failed without the feature enabled, highlighting that the feature is likely essential for successful image transfers between reg1.bupd.xyz and reg2.bupd.xyz. This indicates that the feature could be critical for ensuring stable uploads and completing transfers within the expected time.

## Conclusion:
This observation suggests that this feature is necessary for reliable replication between registries. The data transfer rates and lack of packet loss when the feature is used make it an essential component for stable image replication.

bupd:
you can say a workaround here:
> please set the scheduled execution more than enough time it needs. to be safe. 

The workaround of setting a longer replication interval, like once a day, **fails to address** the need for timely synchronization across registries. For users who rely on Harbor to maintain identical registries at different locations, frequent replication (e.g., every 5 minutes) is necessary to ensure minimal discrepancies between registries. By suggesting a longer interval, users may end up with outdated or inconsistent images, undermining the core functionality of replication.

Thanks.

Vad1mo (approved):

src/pkg/retention/policy/rule/index/index.go
```diff
@@ -163,6 +163,20 @@ func init() {
...
+		Action:     "immutable",
+		Parameters: []*IndexedParam{
+			{
+				Name:     daysps.ParameterN,
+				Type:     "int",
+				Unit:     "days",
+				Required: true,
+			},
+		},
+	}, daysps.New, daysps.Valid)
```
bupd, line 179: why do we need this??

src/jobservice/worker/cworker/c_worker.go
```diff
@@ -66,7 +66,9 @@ type basicWorker struct {
 
 // workerContext ...
 // We did not use this context to pass context info so far, just a placeholder.
-type workerContext struct{}
+type workerContext struct {
+	client *work.Client
+}
```
bupd, line 71: also we don't need this right?

src/jobservice/worker/cworker/c_worker.go
```diff
@@ -146,9 +148,12 @@ func (w *basicWorker) Start() error {
 		w.pool.Stop()
 	}()
 
+	workCtx := workerContext{
+		client: w.client,
+	}
 	// Start the backend worker pool
 	// Add middleware
-	w.pool.Middleware((*workerContext).logJob)
+	w.pool.Middleware(workCtx.logJob)
```
bupd, line 156: also this.

src/controller/replication/flow/copy.go
bupd, line 1: also this needs to be reverted back. I don't know why we need this

bupd (commented):
need some cleanup.

src/controller/jobmonitor/monitor.go
bupd, line 1: revert this back

src/controller/replication/flow/copy.go
bupd, line 1: revert this

src/controller/replication/execution.go
```diff
@@ -109,10 +109,33 @@ func (c *controller) Start(ctx context.Context, policy *replicationmodel.Policy,
...
+		}
+
+		if count > 1 {
+			if err = c.execMgr.MarkSkipped(ctx, id, "Execution skipped: active replication still in progress."); err != nil {
+				return 0, err
+			}
+			return id, nil
+		}
+	}
+
```
bupd, line 138: core logic lies here.

src/pkg/task/execution.go
```diff
@@ -139,6 +141,17 @@ func (e *executionManager) UpdateExtraAttrs(ctx context.Context, id int64, extra
...
+	now := time.Now()
+	return e.executionDAO.Update(ctx, &dao.Execution{
+		ID:            id,
+		Status:        job.SkippedStatus.String(),
+		StatusMessage: message,
+		UpdateTime:    now,
+		EndTime:       now,
+	}, "Status", "StatusMessage", "UpdateTime", "EndTime")
+}
+
```
bupd, line 154: function used to mark the duplicate executions as skipped.

src/testing/pkg/task/execution_manager.go
```diff
@@ -213,6 +213,24 @@ func (_m *ExecutionManager) MarkError(ctx context.Context, id int64, message str
...
+	var r0 error
+	if rf, ok := ret.Get(0).(func(context.Context, int64, string) error); ok {
+		r0 = rf(ctx, id, message)
+	} else {
+		r0 = ret.Error(0)
+	}
+
+	return r0
+}
+
```
bupd, line 233: mock for the mark skipped function

bupd (commented):
To ease up the review process and also reduce the burden on maintainers. I have reviewed the PR myself once again. 

And I have marked the important core parts that this PR changes. 

Since this PR is ~178 lines of change. Which includes both the Backend and UI changes along with added tests for the new feature.

Hope this helps!! 
Thanks

wy65701436:
@Vad1mo We've already reached the FC for version 2.13. Could we consider this feature as a candidate in version 2.14? It appears that the proposal hasn't been accepted yet.

cc @bupd

reasonerjt:
I think the summary is incorrect, it only solves the problem when there are overlaps of the same replication policy.  I suggest shorten the description and use the proposal as single source of truth of this feature, and reference the proposal in this implementation PR.

src/jobservice/runner/redis.go
```diff
@@ -174,7 +174,7 @@ func (rj *RedisJob) Run(j *work.Job) (err error) {
 	case job.PendingStatus, job.ScheduledStatus:
 		// do nothing now
 		break
-	case job.StoppedStatus:
+	case job.StoppedStatus, job.SkippedStatus:
```
reasonerjt: Is there ever a case that a job in redis is marked as "skipped"?

src/controller/replication/execution.go
```diff
@@ -109,10 +109,33 @@ func (c *controller) Start(ctx context.Context, policy *replicationmodel.Policy,
...
+	// If running executions are found, skip the current execution and mark it as skipped.
+	if policy.SingleActiveReplication {
+		count, err := c.execMgr.Count(ctx, &q.Query{
+			Keywords: map[string]interface{}{
+				"VendorType": job.ReplicationVendorType,
+				"VendorID":   policy.ID,
+				"Status":     job.RunningStatus.String(),
+			},
+		})
+		if err != nil {
```
reasonerjt, line 124: When the `Start` function is triggered via a schedule and there's an error returned by `Count`.  The `Create` is already called, will the new row be deleted?  I didn't find that there's a new transaction started if it's triggered via a schedule.

BTW, we should make sure the error is wrapped to have enough context in the error message, and double check the error is logged properlly.

src/controller/replication/execution.go
```diff
@@ -109,10 +109,33 @@ func (c *controller) Start(ctx context.Context, policy *replicationmodel.Policy,
...
+			return 0, err
+		}
+
+		if count > 1 {
+			if err = c.execMgr.MarkSkipped(ctx, id, "Execution skipped: active replication still in progress."); err != nil {
```
reasonerjt: Would it be better to check first and create the execution as "Skipped" rather than create it as "Running" and Mark it as "Skipped" later?

src/jobservice/job/status.go
```diff
@@ -62,6 +64,8 @@ func (s Status) Code() int {
 		return 3
 	case "Success":
 		return 3
+	case "Skipped":
```
reasonerjt: Is it needed?

src/controller/replication/flow/copy.go
```diff
@@ -136,7 +136,7 @@ func (c *copyFlow) createTasks(ctx context.Context, srcResources, dstResources [
 			Metadata: &job.Metadata{
 				JobKind: job.KindGeneric,
 			},
-			Parameters: map[string]any{
+			Parameters: map[string]interface{}{
```
reasonerjt: Is there a warning in your IDE?  I thought "any" and "interface{}" were equivalent?

reasonerjt:
@wy65701436 @chlins Could you please also take a look?

Is it also required to update `src/controller/replication/flow/copy.go` to check the status of execution and skip the flow?

src/controller/replication/execution.go
```diff
@@ -109,10 +109,33 @@ func (c *controller) Start(ctx context.Context, policy *replicationmodel.Policy,
...
 		extra["operator"] = op
 	}
+
 	id, err := c.execMgr.Create(ctx, job.ReplicationVendorType, policy.ID, trigger, extra)
 	if err != nil {
 		return 0, err
 	}
+
+	// If running executions are found, skip the current execution and mark it as skipped.
+	if policy.SingleActiveReplication {
```
wy65701436, line 115: Why do we create an empty execution (without any tasks) for a skipped action? Is it only for the UI to show a "skipped" status?

Shouldn't we return the currently running execution ID with an HTTP 409 response to the API caller, instead of creating a new execution record with no actual processing?

reasonerjt: I think 409 may not work b/c the execution may be triggered by schedule.

It's in the design:
https://github.com/goharbor/community/pull/256

src/controller/replication/execution.go
```diff
@@ -109,10 +109,33 @@ func (c *controller) Start(ctx context.Context, policy *replicationmodel.Policy,
...
+				"VendorType": job.ReplicationVendorType,
+				"VendorID":   policy.ID,
+				"Status":     job.RunningStatus.String(),
+			},
+		})
+		if err != nil {
+			return 0, err
+		}
+
+		if count > 1 {
```
wy65701436: Regarding the `skipped` status — is it considered a final status for an execution?

I’m asking because this affects how we handle it:

- If skipped is not a final status, the API should eventually transition it to a final state like success, error, or failed.

- If it is a final status, we need to ensure it’s treated accordingly:

1. The "stop execution" API should not attempt to stop a skipped execution.
2. Our system utility that cleans up executions in final status should also include skipped. If so, that logic needs to be updated.

src/jobservice/job/status.go
wy65701436: it's the status of job, not the execution.

src/jobservice/runner/redis.go
wy65701436: ditto

make/migrations/postgresql/0170_2.14.0_schema.up.sql
```diff
@@ -6,4 +6,6 @@ ALTER SEQUENCE permission_policy_id_seq AS BIGINT;
 
 ALTER TABLE role_permission ALTER COLUMN permission_policy_id TYPE BIGINT;
 
-ALTER TABLE vulnerability_record ADD COLUMN IF NOT EXISTS status text;
\ No newline at end of file
+ALTER TABLE vulnerability_record ADD COLUMN IF NOT EXISTS status text;
```
chlins, line 9: You should rebase the main branch.

src/controller/replication/execution.go
```diff
@@ -109,10 +109,33 @@ func (c *controller) Start(ctx context.Context, policy *replicationmodel.Policy,
...
 		return 0, err
 	}
+
+	// If running executions are found, skip the current execution and mark it as skipped.
+	if policy.SingleActiveReplication {
+		count, err := c.execMgr.Count(ctx, &q.Query{
+			Keywords: map[string]interface{}{
+				"VendorType": job.ReplicationVendorType,
+				"VendorID":   policy.ID,
+				"Status":     job.RunningStatus.String(),
```
chlins, line 121: The current status update for exec is asynchronous, which means that while the actual task may have already finished, there could be a delay in the status refresh. As a result, the database might still show "running" even though the task is actually complete. This could lead to scenarios where the real job has finished, but because the status hasn't been updated yet, a new exec is mistakenly identified as skipped.

src/controller/replication/execution.go
chlins: Do we really need a new "skipped" status for representation, or is it feasible to reuse the Error status and include the skip information in the error message?

src/server/v2.0/handler/replication.go
```diff
@@ -113,6 +113,14 @@ func (r *replicationAPI) CreateReplicationPolicy(ctx context.Context, params ope
 		policy.CopyByChunk = *params.Policy.CopyByChunk
 	}
 
+	if params.Policy.SingleActiveReplication != nil {
+		// Validate and assign SingleActiveReplication only for non-event_based triggers
+		if params.Policy.Trigger.Type == "event_based" && *params.Policy.SingleActiveReplication == true {
```
chlins: Don't we have a const definition for event based trigger in current codebase?

src/controller/replication/execution.go
chlins: Since the execution is used for all type jobs, but actually the "skipped" is only for replication, other job has no such scenario.

src/jobservice/job/status.go
bupd: @reasonerjt Yes, it's needed. Without handling SkippedStatus, the job might continue running. There's logic that checks the job status and proceeds with execution unless the status code is 3, which marks a final state. We explicitly create some job executions with SkippedStatus, and this case ensures the runner treats them as finalized, preventing unintended execution.

src/controller/replication/execution.go
bupd: It is fine. this delay doesn't cause issues in practice and doesn't require special handling.

src/controller/replication/execution.go
bupd: @reasonerjt will add more context to the error.

src/controller/replication/execution.go
bupd: > Regarding the skipped status — is it considered a final status for an execution?

Yes

> If it is a final status, we need to ensure it’s treated accordingly:

Yes, its included it already

I believe this PR already addresses points 1 and 2.

src/controller/replication/execution.go
bupd: > Do we really need a new "skipped" status for representation, or is it feasible to reuse the Error status and include the skip information in the error message?

No, it would not be wise. Having skipped as a status is clear.

bupd: Having separate functions for creating skipped and normal executions adds unnecessary operational overhead and code duplication. I had implemented it before but removed it to reduce clutter. It's cleaner and more maintainable to create the execution normally and update its status to `skipped`.

src/controller/replication/flow/copy.go
bupd: @reasonerjt 

Thanks for the heads up. will update it.

src/jobservice/runner/redis.go
bupd: I have added this to ensure that skipped jobs are not being unintentionally taken up for execution

make/migrations/postgresql/0170_2.14.0_schema.up.sql
bupd: done

src/server/v2.0/handler/replication.go
bupd: @chlins Thanks for the heads up fixed it.

src/jobservice/job/status.go
reasonerjt: @bupd But I still don't quite understand, is there anywhere setting the status of a job as "Skipped"?

src/jobservice/runner/redis.go
reasonerjt: @bupd 
I think there's still misunderstanding.
There is no such thing as "skipped job", b/c no code will set the status of a job as "skipped"

src/controller/replication/execution.go
reasonerjt: @bupd But what's your opinion about my first concern:

> When the Start function is triggered via a schedule and there's an error returned by Count. The Create is already called, will the new row be deleted? I didn't find that there's a new transaction started if it's triggered via a schedule.

src/jobservice/job/status.go
bupd: We set the job as skipped in the `MarkSkipped()` function

https://github.com/container-registry/harbor-next/blob/aae5523e39e1376a5ae7b0ee39f628fcf39b5a8e/src/pkg/task/execution.go#L148

src/jobservice/runner/redis.go
bupd: no we set the job as skipped. 

https://github.com/container-registry/harbor-next/blob/aae5523e39e1376a5ae7b0ee39f628fcf39b5a8e/src/pkg/task/execution.go#L148

src/controller/replication/execution.go
wy65701436: Sorry, I need to reopen this. Since we've introduced a new final status, we must ensure all related logic that handles final execution states is updated accordingly.

1. Please check and update the STOP operation if necessary:

    https://github.com/goharbor/harbor/blob/main/src/pkg/task/execution.go#L182

2. Review the SWEEP operation logic here:
    https://github.com/goharbor/harbor/blob/main/src/pkg/task/sweep_manager.go#L95

3. Also, confirm whether the LIST operation needs to be updated:
    https://github.com/goharbor/harbor/blob/main/src/server/v2.0/handler/replication.go#L287

The items above are what I’ve identified so far, but they may not cover all affected areas. Please do a thorough review of the relevant code to ensure full alignment with the new status.

src/controller/replication/execution.go
bupd: I have updated the flow to not create ghost executions. so now the execution is created only after the count.

src/jobservice/runner/redis.go
reasonerjt: Sorry but this is a misconception.
The link you pasted is updating an execution, not a job, like I explained in the design review....

src/jobservice/job/status.go
reasonerjt: Please see my comment:
https://github.com/goharbor/harbor/pull/21347#discussion_r2244729292

src/controller/replication/execution.go
reasonerjt: I checked the code, but it's still incorrect. I think the condition in line 135 should be "count > 0" or "count >=1 " b/c when you count the running execution the new execution is not created, and if there's one execution is running the new one should be marked.

bupd: yeah, fixed the count issue. 

Thanks for the heads up.

src/controller/replication/execution.go
bupd: I have removed `skipped` from being a status. and used error status to report skipped

src/jobservice/job/status.go
bupd: I have removed `skipped` as a status

src/jobservice/runner/redis.go
bupd: I have removed skipped as a status. and used error status as final.

src/controller/replication/execution.go
```diff
@@ -109,10 +109,37 @@ func (c *controller) Start(ctx context.Context, policy *replicationmodel.Policy,
 	if op := operator.FromContext(ctx); op != "" {
 		extra["operator"] = op
 	}
+
+	var count int64
+	// If running executions are found, skip the current execution and mark it as skipped.
```
wy65701436: mark it as error?

src/server/v2.0/handler/replication.go
```diff
@@ -113,6 +113,14 @@ func (r *replicationAPI) CreateReplicationPolicy(ctx context.Context, params ope
 		policy.CopyByChunk = *params.Policy.CopyByChunk
 	}
 
+	if params.Policy.SingleActiveReplication != nil {
+		// Validate and assign SingleActiveReplication only for non-event_based triggers
+		if params.Policy.Trigger.Type == model.TriggerTypeEventBased && *params.Policy.SingleActiveReplication == true {
```
reasonerjt: Is this necessary?  If we don't allow single active replication for event-based replication, we should also validate it in 
`UpdateReplicationPolicy`

Also a nit: `*params.Policy.SingleActiveReplication == true`  -> `*params.Policy.SingleActiveReplication`

src/controller/replication/execution.go
bupd: updated

src/server/v2.0/handler/replication.go
bupd: yes. because "event_based_replication" is already replicates only the changed artifacts based on events and not the entire artifact list. 

and I have updated `UpdateReplicationPolicy` function. and also fixed nit

src/server/v2.0/handler/replication.go
```diff
@@ -181,6 +189,14 @@ func (r *replicationAPI) UpdateReplicationPolicy(ctx context.Context, params ope
 		policy.CopyByChunk = *params.Policy.CopyByChunk
 	}
 
+	if params.Policy.SingleActiveReplication != nil {
+		// Validate and assign SingleActiveReplication only for non-event_based triggers
+		if params.Policy.Trigger.Type == model.TriggerTypeEventBased && *params.Policy.SingleActiveReplication {
```
reasonerjt: Sorry, I didn't raise it earlier but checking the line 171 there's `if params.Policy.Trigger != nil {` 

It seems there's a risk the `params.Policy.Trigger ` is nil, I think we also need to ensure it's not nil here.

src/server/v2.0/handler/replication.go
```diff
@@ -113,6 +113,14 @@ func (r *replicationAPI) CreateReplicationPolicy(ctx context.Context, params ope
...
```
reasonerjt: Sorry, I didn't raise it earlier but checking the line 95 there's `if params.Policy.Trigger != nil {` 

It seems there's a risk the `params.Policy.Trigger ` is nil, I think we also need to ensure it's not nil here.

bupd: fixed

PR #21376, opened 2025-01-03
Add middleware for audit log

stonezdj:

# Issue being fixed
Fixes #(issue)

### DIFF
--- a/src/controller/event/metadata/commonevent/model.go
+++ b/src/controller/event/metadata/commonevent/model.go
@@ -1,0 +1,89 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package commonevent
+
+import (
+	"context"
+	"regexp"
+	"sync"
+
+	"github.com/goharbor/harbor/src/pkg/notifier/event"
+)
+
+// Resolver the interface to resolve Metadata to CommonEvent
+type Resolver interface {
+	Resolve(*Metadata, *event.Event) error
+	PreCheck(ctx context.Context, url string, method string) (bool, string)
+}
+
+var urlResolvers = map[string]Resolver{}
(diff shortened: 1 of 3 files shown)


src/server/middleware/log/log.go
```diff
@@ -40,6 +45,61 @@ func Middleware() func(http.Handler) http.Handler {
...
 		}
 
-		next.ServeHTTP(w, r)
+		e := &commonevent.Metadata{
+			Ctx:           r.Context(),
+			Username:      "unknown",
+			RequestMethod: r.Method,
+			RequestURL:    r.URL.String(),
+		}
+		if matched, resName := e.PreCheck(); matched {
```
wy65701436, line 54: we can leverage the  middleware.BeforeRequest and  middleware.AfterRequest to make it more readable.

src/server/middleware/log/log.go
```diff
@@ -40,6 +45,61 @@ func Middleware() func(http.Handler) http.Handler {
...
+			lib.NopCloseRequest(r)
```
wy65701436, line 55: If we add this line, each request(almost each) will become a no-op close, and for certain requests, such as PUT manifest, it will be called twice.

Does that make sense?

src/controller/event/metadata/commonevent/model.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	return nil
+}
+
+// PreCheck check if current event is matched and return the prefetched resource name when it is delete operation
+func (c *Metadata) PreCheck() (bool, string) {
+	for urlPattern, r := range Resolvers() {
+		p := regexp.MustCompile(urlPattern)
+		if p.MatchString(c.RequestURL) {
+			return r.PreCheck(c.Ctx, c.RequestURL, c.RequestMethod)
```
chlins, line 71: Here might seem confusing, `Metadata` implements two methods, ` Resolve` and `PreCheck`, and the `Resolver` interface definition above also has these two methods. However, `Metadata` doesn't actually implement this interface because the method parameters are different, which can easily lead to misunderstanding, can we have different naming ways to distinguish easily.

src/server/middleware/log/log.go
stonezdj: For user login event, it requires to parse the request content to get the username, use the BeforeRequest and AfterRequest can't read the request body.

src/server/middleware/log/log.go
stonezdj: The `lib.NopCloseRequest(r)` is called after the PreCheck() is matched, the audit log middleware just handles the request of `/api/v2.0/*`, the PUT manifest is the request to `/v2/`

chlins (approved):
lgtm

wy65701436 (approved):
lgtm

PR #21385, opened 2025-01-07
feat: allow for configuring the docker hub registry endpoint via API

tuunit:
This PR adds the possibility to change the registry endpoint of DockerHub via API. This is useful for Harbor instances that run in airgapped environments and have to go through a DMZ or use a reverse proxy to access DockerHub.

**Why not just use a normal Docker Registry type instead of the DockerHub adapter?**

Because of the logic for `library` images which only applies to the Registry Type DockerHub:
 https://github.com/goharbor/harbor/blob/8bf710a4059415096a2fe882af1ab019eb0079b6/src/server/middleware/repoproxy/proxy.go#L128-L144

By allowing to change the registry endpoint for DockerHub, those kinds of environments can still benefit from the `library` substitution logic.

### DIFF
--- a/src/pkg/reg/adapter/dockerhub/adapter.go
+++ b/src/pkg/reg/adapter/dockerhub/adapter.go
@@ -119,7 +119,7 @@ func (a *adapter) Info() (*model.RegistryInfo, error) {
 func getAdapterInfo() *model.AdapterPattern {
 	info := &model.AdapterPattern{
 		EndpointPattern: &model.EndpointPattern{
-			EndpointType: model.EndpointPatternTypeFix,
+			EndpointType: model.EndpointPatternTypeStandard,
 			Endpoints: []*model.Endpoint{
 				{
 					Key:   "hub.docker.com",
--- a/src/pkg/reg/adapter/dockerhub/adapter.go
+++ b/src/pkg/reg/adapter/dockerhub/adapter.go
@@ -46,6 +46,13 @@ func newAdapter(registry *model.Registry) (adp.Adapter, error) {
 		return nil, err
 	}
 
+	var registryURL string
+	if registry.URL != baseURL {
+		registryURL = registry.URL
+	} else {
+		registryURL = defaultRegistryURL
+	}
+
 	return &adapter{
 		client:   client,
 		registry: registry,
--- a/src/pkg/reg/adapter/dockerhub/consts.go
+++ b/src/pkg/reg/adapter/dockerhub/consts.go
@@ -18,7 +18,7 @@ import "fmt"
 
 const (
 	baseURL             = "https://hub.docker.com"
-	registryURL         = "https://registry-1.docker.io"
+	defaultRegistryURL  = "https://registry-1.docker.io"
 	loginPath           = "/v2/users/login/"
 	listNamespacePath   = "/v2/repositories/namespaces"
 	createNamespacePath = "/v2/orgs/"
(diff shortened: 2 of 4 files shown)


tuunit:
@Vad1mo any feedback on the approach? Do you want me to write an apitest?

Vad1mo:
It might be the stupid questions. but why is the URL hardcoded at all? why not preset and allow the user to enter what he wants.

tuunit:
I have no idea why it is hardcoded 😅 If you want me to change the UI as well to allow changing it. I can do it as part of this PR or another. Just hit me up what you guys would prefer how to continue

wy65701436:
We're doing this due to the behavior in the upstream. You can refer to the following link for more details: https://github.com/moby/moby/blob/master/registry/config.go#L43.

@tuunit, could you clarify why you'd like to change the Docker Hub URL? Does the default URL not meet your requirements?

tuunit:
@wy65701436 thanks for actually finding the original place in the moby code base 😅

Nevertheless, I already know that is the official registry URL of docker hub. But as explained in my PR description, it is sometimes necessary to run Harbor inside an air gapped / private environment without direct internet access and therefore replications / caching might need to go through a simple reverse proxy to communicate with the public web. For this purpose it would be great to keep the logic of the `library` completion if missing  despite running behind a proxy. More details in the PR summary. (which is only done for the docker-hub adapter)

Vad1mo:
> We're doing this due to the behavior in the upstream. You can refer to the following link for more details: https://github.com/moby/moby/blob/master/registry/config.go#L43.

The strategy Docker is following is for commercial reasons and backwards compatibility only. We should not limit ourselves to that. 

If we ask ourselves why people user Harbor. One of the big arguments for Harbor is replication. 

1. On-Prem/air gapped environments
2. No Cloud 
3. non Standards setups

Regarding this PR or its further thinking

1. Will a prefilled and changeable URL set to `registry-1.docker.io` be backwards incompatible? → No
2. Will it restrict users? → No
3. Will it open up new possibilities and use cases? → yes


Personally, I don't see any strong argument against it.

wy65701436:
> . For this purpose it would be great to keep the logic of the `library` completion if missing despite running behind a proxy. More details in the PR summary. (which is only done for the docker-hub adapter)

I am still curious, for this case, why the proxy setting cannot work for you? https://github.com/goharbor/harbor/blob/main/make/harbor.yml.tmpl#L232

MinerYang:
Would you help us to understand more why you need a reverse-proxy for dockerhub only instead of adding it into something like a while list in your air-gapped env?

wy65701436 (approved):
lgtm

reasonerjt:
>  after updating the URL through the API, the UI still shows the default value.

This looks like a show-stopper...

tuunit:
> Hi @tuunit Would you help us to understand more why you need a reverse-proxy for dockerhub only instead of adding it into something like a while list in your air-gapped env?

No offence but have you ever worked in government project? Or any air gap environments that doesn't allow ANY internet access except from one single proxy that is highly monitored and controlled?

I've seen setups like this often enough at my previous employer. Harbor might run inside an internal network without any internet access whatsoever so re-routing to a proxy is the only option which will most likely be behind multiple other hardware loadbalancers and firewalls.

tuunit:
> It looks good to me, but could you provide additional information from your side? For example, a screenshot and detailed logs after modifying the URL and successfully replicating images through the reverse proxy would be really helpful. Since we don't have a test case to validate this, it would give us more confidence in the implementation.
> 


I will provide some screenshots and logs later.

> I have one concern: after updating the URL through the API, the UI still shows the default value. I believe we could either update the front-end code accordingly or fork Harbor on your side to address this.

This isn't correct, it's even worse right now. The UI shows a different URL after updating it via the API but the registry URL is not used inside the code. So it currently seems like you can change it via the API but it doesn't do anything.

tuunit:
@Vad1mo @reasonerjt @wy65701436 could you reopen the PR?

PR #21473, opened 2025-01-31
fix: ensure data_volume isn't surrounded by quotes

Aletor93:

I'm generating `harbor.yaml` file with erb function `to_yaml`, where strings containing `/` are quoted. It can't be configured. 

When I run `./install.sh` script, it fails on docker volume command due to quotes which aren't removed: 

```sh
prepare base dir is set to /srv/harbor/install
docker: Error response from daemon: create "/srv/harbor/data": "\"/srv/harbor/data\"" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.
See 'docker run --help'.
```

After editing the `prepare` script to remove quotes on data_volume if present, it works like a charm. The `sed` is already used in other scripts, it shouldn't be a problem to use it.

# Issue being fixed

n/a (no issue opened)

### DIFF
--- a/make/prepare
+++ b/make/prepare
@@ -30,7 +30,7 @@ else
     fi
 fi
 
-data_path=$(grep '^[^#]*data_volume:' $input_dir/harbor.yml | awk '{print $NF}')
+data_path=$(grep '^[^#]*data_volume:' $input_dir/harbor.yml | awk '{print $NF}' | sed 's/"//g')
 
 # If previous secretkeys exist, move it to new location
 previous_secretkey_path=/data/secretkey


make/prepare
```diff
@@ -30,7 +30,7 @@ else
     fi
 fi
 
-data_path=$(grep '^[^#]*data_volume:' $input_dir/harbor.yml | awk '{print $NF}')
+data_path=$(grep '^[^#]*data_volume:' $input_dir/harbor.yml | awk '{print $NF}' | sed 's/"//g')
```
stonezdj, line 33: The intention is to remove the quote when there is a quote, sometimes the path may contain space, to remove quote will cause problem?

Aletor93: > The intention is to remove the quote when there is a quote, sometimes the path may contain space, to remove quote will cause problem?

In my opinion, its already impossible to use a path with spaces due to: `| awk '{print $NF}'`. 

Here is a test: 

```sh
# grep data_volume harbor.yml
data_volume: "/srv/harbor/data directory with space"

# grep '^[^#]*data_volume:' harbor.yml | awk '{print $NF}'
space"
```

There is no change to expect about this.

OrlinVasilev:
@Aletor93 what shell are you using ?

Aletor93:
> @Aletor93 what shell are you using ?

For the [test made here](https://github.com/goharbor/harbor/pull/21473#discussion_r1949349647), I've used `bash`. It seems to be the shell used inside `prepare` script. 

For a better test case, I've commented my modification on `prepare` script and added an echo to debug. 

```sh
data_path=$(grep '^[^#]*data_volume:' $input_dir/harbor.yml | awk '{print $NF}') # | sed 's/"//g')
echo ${data_path}-
```

Then tested to set `data_volume` with a path containing spaces:  `data_volume: "/srv/harbor/data directory with space"`

```sh
# ./install.sh 
...
[Step 3]: preparing harbor configs ...
prepare base dir is set to /srv/harbor/install
space"-
docker: Error response from daemon: create space": "space\"" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.
See 'docker run --help'.
```

I observe that `awk` get the last part of the string (`$NF`), using `space` as its default separator.

make/prepare
bupd, line 33: The actual solution may involve adding an escape sequence for spaces in paths. Whenever a space is encountered, it should be properly escaped to ensure that paths with spaces are handled correctly.

bupd (changes requested):

make/prepare
Aletor93: Could you please provide me a string example to help me to reproduce the expected result ? I tried to escape spaces, awk's `$NF` still work the same.

PR #21474, opened 2025-01-31
fix: fix replication of multiple projects with numeric names

### LINKED ISSUE #19027: When the project name is a number, the replication error
image
image
image

error.log
```
2023-07-26T09:30:00Z [ERROR] [/lib/http/error.go:57]: {"errors":[{"code":"UNKNOWN","message":"unknown: cannot convert 555 to Text"}]} 
```

cgirard-mir:

The original fix to #19027 contained in #19090 worked well for replicating a single project whose name was numeric, but did not resolve the issue when multiple numeric project names were provided.

The issue was in the construction of the API call. The query parameter was previously being constructed as `name={'123 456 789'}`. With the addition of some debug logging ~(also included in this PR)~, I noticed that the server was parsing this as `'123`, `456`, `789'`. Note that the single quotes were being associated with the first and last values of the list, so these values were forced into parsing as string values. However, the middle value contains only numbers and so was parsed as such. This doesn't cause a problem when only 1 or 2 values are being passed, but can show itself once there are 3 or more values.

The solution is to surround each project name in quotes, `name={'123' '456' '789'}`, so they are all handled as strings by the server.

# Issue being fixed
Further fixes #19027.


cgirard-mir:
FYI, all of the tests were previously passing on this, but after I updated the branch for "out-of-date with base", the test statuses have been reset. Since I don't have permission to run those tests, I'm not planning to do another "Update branch" until they've been re-run and the PR has 2 approvals from maintainers.

chlins (approved):
lgtm

src/lib/q/builder.go
```diff
@@ -189,19 +189,22 @@ func parseList(value string, c rune) ([]interface{}, error) {
 
 // try to parse value as time first, then integer, and last string
 func parseValue(value string) interface{} {
+	var ret any
 	value = strings.TrimSpace(value)
```
wy65701436, line 193: Thanks @cgirard-mir for the fix, great catch!

Regarding the refactor, I feel the original implementation is more straightforward. If there’s no change in logic, could we keep it as is?

cgirard-mir: The refactor adds a debug-level logging line to the function (see below). I can remove this commit, though, if you aren't interested in keeping that.

wy65701436: If you'd like to add the debug log, could you please include it in the original code?

I still prefer to keep the original structure, as it returns early and avoids excessive if-else conditions.

cgirard-mir: I'll just drop the commit. I don't want to have a logging line in each `if` block.

cgirard-mir: The refactoring has been reverted. Only the bug fix commit remains.

bupd (approved):
lgtm

bupd:
Thanks for your contribution. 
I can confirm this sure is a problem in replication and this PR can be a simple fix for the problem
  
i can confirm

wy65701436 (approved):
lgtm

PR #21659, opened 2025-02-21
Add `CNAI` processor

s3rj1k:
implements https://github.com/goharbor/community/pull/257


chlins:
@s3rj1k Thanks for your work!  It seems duplicated with https://github.com/goharbor/harbor/pull/21663, maybe we can collaborate on that.

s3rj1k:
@chlins sure, how you want to handle that, Slack?

chlins:
First of all, thank you very much for your attention to this proposal and your investment in its implementation. Here, I would like to introduce some of the design background and original intention of Harbor. Harbor itself is positioned as an OCI artifact repository. Since the OCI Artifact Spec has a relatively broad and flexible definition of OCI artifacts, there may be various artifacts in different standard formats. Initially, Harbor only supported storing container images, so the overall design style and code design were actually based on container images. Later, with the increase in the types of artifacts, such as helm charts, in order to better support the display of relevant metadata of these different artifacts, Harbor designed a 1:1 mapping implementation of artifact types and corresponding processors. Different artifacts can return corresponding data and customize the UI according to their respective standards.

Therefore, in the short term, in the relatively new field of AI, there is no spec that unifies the world like OCI spec (but the vision of CNAI model spec is to unify this part), so I would prefer to use the CNAI model spec as an independent processor dedicated to CNAI. This is the simplest and least impactful way to connect to Harbor currently. Of course, it would be great if it could be compatible with multiple specs in the design, but from the current architecture, it may require some major restructuring of the original design. Therefore, I would prefer to discuss this unified implementation as another topic or proposal, which will be a long-term plan. Because the development of AI itself is very fast, we don't know what will happen in the future and what changes will occur, so it is difficult for us to make a complete abstraction immediately. Therefore, I think we can first use CNAI as an experimental AI scenario access, and then optimize and abstract it at a higher level in future.

Issue #21229, opened 2024-11-22
Implementation for AI model artifact processor


chlins:
### Background

https://github.com/CloudNativeAI/model-spec

The model spec extends the OCI image spec for better building and serving the AI models to the OCI Artifact, since Harbor is the OCI Artifact registry, so we propose that Harbor adopts the model specification standards to ensure consistency and compatibility with AI model-related workflows. Additionally, we request the development of an experimental feature to enable the display of foundational AI model metadata, as previous design in the Harbor codebase, we only need to implement an AI model processor to handle related logics and some UI works for better user interaction experience.

reasonerjt:
What does `kind/experimental` mean?  If this spec is not GAed will we merge the change to `main` branch.

chlins:
I believe that AI model may require more intricate logic in the future, and the current enhanced default processor is only capable of customizing icons or metadata, which falls short of handling certain extraction tasks such as helm charts. Therefore, I propose that we develop a new processor specifically for the AI model, which will offer enhanced customization and flexibility. At the very least, if we intend to incorporate this modification, we must publish one release version of model specification.

For the `kind/experimental` label, I am adding it to indicate a feature that we can consider as experimental.

chlins:
If we don't currently have a concept like `experimental feature` on harbor, I think it's also possible to not have this label, because adding a new model type processor is not going to have any impact on the existing codebase or existing functionality, but I agree that we could have a short proposal to explain why we need a new processor to better handle model parsing.

PR #21663, opened 2025-02-24
feat: implement the CNAI model processor

chlins:

# Issue being fixed
Fixes https://github.com/goharbor/harbor/issues/21229


src/controller/artifact/processor/model/model.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/goharbor/harbor/src/controller/artifact/processor/model/parser"
+	"github.com/goharbor/harbor/src/lib/errors"
+	"github.com/goharbor/harbor/src/lib/log"
+	"github.com/goharbor/harbor/src/pkg/artifact"
+)
+
+// const definitions
+const (
+	// ArtifactTypeModel defines the artifact type for AI model.
+	ArtifactTypeModel = "MODEL"
```
s3rj1k, line 35: I also have this here https://github.com/goharbor/harbor/blob/01b4b23856ce1e1fb9a931c763ecaa040f366ef7/src/controller/artifact/processor/cnai/cnai.go#L36 and I feel like we should consider changing `MODEL` to something like `CNAI`, Harbor can possibly support multiple ML models in future

src/pkg/artifact/model.go
```diff
@@ -48,6 +48,15 @@ type Artifact struct {
 	References        []*Reference           `json:"references"` // child artifacts referenced by the parent artifact if the artifact is an index
 }
 
+// ResolveArtifactType returns the artifact type of the artifact, prefer ArtifactType, use MediaType if ArtifactType is empty.
+func (a *Artifact) ResolveArtifactType() string {
```
s3rj1k, line 52: I am pretty-sure that this change is not necessary and can be avoided to reduce scope of change to only adding artifact processor, but let's see what code-owners have to say.

stonezdj (approved):

Vad1mo (dismissed):

src/controller/artifact/processor/model/model.go
wy65701436: +1 to update it to `CNAI`

src/controller/artifact/processor/model/parser/base.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// base provides a default implementation for other parsers to build upon.
+type base struct {
+	regCli registry.Client
+}
+
+// Parse is the common implementation for parsing layer.
+func (b *base) Parse(_ context.Context, artifact *artifact.Artifact, layer *ocispec.Descriptor) (string, []byte, error) {
+	if artifact == nil || layer == nil {
```
wy65701436, line 50: please add the blob size checking, if the blob is larger than a specific size, then UI shows unable to show the file content.

src/pkg/artifact/model.go
wy65701436: The logic of this method is correct, and wrapping it into a shared method would reduce duplication. Works for me.

src/controller/artifact/processor/model/parser/util.go
```diff
@@ -0,0 +1,150 @@
...
+		if _, err := io.Copy(&buf, tr); err != nil {
+			return nil, fmt.Errorf("failed to copy content to buffer: %w", err)
+		}
+	}
+
+	return buf.Bytes(), nil
+}
+
+// FileType represents the type of a file.
+type FileType = string
```
wy65701436, line 54: I’m curious—aren’t there any existing Go libraries to handle this?

src/controller/artifact/processor/model/parser/base.go
s3rj1k: UI can show/will show arbitrary blob content or are we talking about README and License?

src/controller/artifact/processor/model/parser/util.go
s3rj1k: Traversing nodes?
This implementation is very specific for CNAI, it won't scale for other (ML) artifacts

src/controller/artifact/processor/model/parser/base.go
chlins: @wy65701436 Thanks, it's a good catch, we should check the size in the backend to avoid parse big blob, will make the follow-up PR to fix this issue.

chlins: Currently only README and LICENSE can be displayed in the UI, other files only display the file tree, open or download other files needs to be another proposal to discuss in the future, because based on the current API design of harbor, it's hard to open or download other files.

src/controller/artifact/processor/model/parser/util.go
chlins: > I’m curious—aren’t there any existing Go libraries to handle this?

I don't found the effective go library to handle this case when implement this part, but we can migrate to existing library if we found it in the future.

chlins: > Traversing nodes? This implementation is very specific for CNAI, it won't scale for other (ML) artifacts

This code and logic is implemented in the CNAI model processor, why I need to care other ML artifacts? other ML artifacts can implement their own processor, that's why harbor design the processor mechanism for extension.

PR #21679, opened 2025-02-27
remove redis sentinel patch from builder

MinerYang:

Since we will use harbor/distribution since release-2.13, the redis sentinel patch will be removed from builder.


stonezdj (approved):

chlins (approved):
lgtm

Issue #21632, opened 2025-02-13
Correctly Prepare redis url and CA for offline-installer.


MinerYang:
To allow TLS connection when using cache like Azure Redis or Elastic cache service, we need users to configure `enableTLS` while using external redis service and add prepare logic for harbor offline-installer.
- pass correct `rediss://` to each of the config file of each harbor component need redis tls connection.
- prepare self-sighed ca.crt in `external.redis.tlsOption.cacrt` for install_certs

reasonerjt:
Fixed via #21667 and #21681

PR #21681, opened 2025-02-28
update tlsOptions for external redis

MinerYang:
# Issue being fixed
We should allows using both public trusted certificate and a self-signed certificate for redis tls connection
https://github.com/goharbor/harbor/issues/21632


stonezdj (approved):

chlins (approved):
lgtm

PR #21718, opened 2025-03-10
oidclogout

wy65701436:
enable oidc session logout

1, give the option of logging out user session from OIDC provider. 2, give the option of logging out user offline session from OIDC provider.

# Issue being fixed
Fixes #(issue)


src/core/controllers/base.go
```diff
@@ -108,8 +108,46 @@ func (cc *CommonController) Login() {
...
+				if err != nil {
+					log.Errorf("Failed to get the external endpoint, error: %v", err)
+					cc.CustomAbort(http.StatusUnauthorized, "")
+				}
+				url := strings.TrimSuffix(ep, "/") + common.OIDCLoginoutPath
+				log.Debugf("Redirect user %s to logout page of OIDC provider", u.Username)
+				// Return a json to UI with status code 403, as it cannot handle status 302
+				cc.Ctx.Output.Status = http.StatusForbidden
+				err = cc.Ctx.Output.JSON(struct {
+					Locatioon string `json:"redirect_location"`
```
reasonerjt, line 141: "locatioon" is a typo.

src/core/controllers/oidc.go
```diff
@@ -338,3 +434,52 @@ func secretAndToken(tokenBytes []byte) (string, string, error) {
...
+	auth := base64.StdEncoding.EncodeToString([]byte(clientID + ":" + clientSecret))
+	req, err := http.NewRequest("POST", revokeURL, bytes.NewBufferString(data.Encode()))
+	if err != nil {
+		return errors.Errorf("failed to create request: %v", err)
+	}
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	req.Header.Set("Authorization", "Basic "+auth)
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
Vad1mo, line 473: This always skips certificate verification regardless of the oidc_verify_cert setting, creating a security vulnerability.

src/core/controllers/oidc.go
```diff
@@ -244,6 +248,98 @@ func (oc *OIDCController) Callback() {
...
+		return
+	}
+	endSessionURL := oidc.EndpointsClaims.EndSessionURL
+	baseUrl, err := config.ExtEndpoint()
+	if err != nil {
+		log.Errorf("Failed to get external endpoint: %v", err)
+		oc.SendInternalServerError(err)
+		return
+	}
+	postLogoutRedirectURI := fmt.Sprintf("%s/harbor/projects", baseUrl)
```
Vad1mo, line 332: The redirect URI after OIDC logout is hardcoded to /harbor/projects, this makes no sense. I would redirect to root page...

src/core/controllers/base.go
```diff
@@ -108,8 +108,46 @@ func (cc *CommonController) Login() {
...
 }
 
-// LogOut Habor UI
+// LogOut Harbor UI
 func (cc *CommonController) LogOut() {
+	// redirect for OIDC logout, excludes the admin user.
+	securityCtx, ok := security.FromContext(cc.Context())
+	if !ok {
+		log.Error("Failed to get security context")
+		cc.CustomAbort(http.StatusInternalServerError, "Internal error.")
```
Vad1mo, line 117: Can we not have a precise error message, explaining what actually went wrong?

src/core/controllers/base.go
```diff
@@ -108,8 +108,46 @@ func (cc *CommonController) Login() {
...
+	}
+	principal := securityCtx.GetUsername()
+	if principal != "" {
+		if redirectForOIDC(cc.Ctx.Request.Context(), principal) {
+			u, err := user.Ctl.GetByName(cc.Context(), principal)
+			if err != nil {
+				log.Errorf("Failed to get user by name: %s, error: %v", principal, err)
+				cc.CustomAbort(http.StatusInternalServerError, "Internal error.")
```
Vad1mo, line 125: same here, in the error message we have better error message why a generic in the reponse?

src/core/controllers/base.go
```diff
@@ -108,8 +108,46 @@ func (cc *CommonController) Login() {
...
+			}
+			if u == nil {
+				cc.CustomAbort(http.StatusInternalServerError, "Internal error.")
```
Vad1mo, line 128: same here, out of the 3 different error messages, the response it always the gneric message, I don't think this is helping the users.

src/lib/config/metadata/metadatalist.go
```diff
@@ -147,6 +147,8 @@ var (
 		{Name: common.OIDCVerifyCert, Scope: UserScope, Group: OIDCGroup, DefaultValue: "true", ItemType: &BoolType{}, Description: `Verify the OIDC provider's certificate'`},
 		{Name: common.OIDCAutoOnboard, Scope: UserScope, Group: OIDCGroup, DefaultValue: "false", ItemType: &BoolType{}, Description: `Auto onboard the OIDC user`},
 		{Name: common.OIDCExtraRedirectParms, Scope: UserScope, Group: OIDCGroup, DefaultValue: "{}", ItemType: &StringToStringMapType{}, Description: `Extra parameters to add when redirect request to OIDC provider`},
+		{Name: common.OIDCLogout, Scope: UserScope, Group: OIDCGroup, DefaultValue: "false", ItemType: &BoolType{}, Description: `Logout the OIDC user session`},
```
Vad1mo, line 150: The option is a checkbox, but the message is an action..  Its not clear what happens if the checkbox is enabled.

src/lib/config/metadata/metadatalist.go
```diff
@@ -147,6 +147,8 @@ var (
...
+		{Name: common.OIDCLogoutOffline, Scope: UserScope, Group: OIDCGroup, DefaultValue: "false", ItemType: &BoolType{}, Description: `Logout the OIDC user offline session`},
```
Vad1mo, line 151: The option is a checkbox, but the message is an action..  Its not clear what happens if the checkbox is enabled.

src/core/controllers/oidc.go
```diff
@@ -244,6 +248,98 @@ func (oc *OIDCController) Callback() {
 	}
 }
 
+func (oc *OIDCController) RedirectLogout() {
```
Vad1mo, line 251: Token Verification First: IMO, the token verification should happen before any session destruction to ensure a valid token for the OIDC provider logout.

Session Destruction :

1. For non-OIDC logout flows: Destroy session and redirect immediately
1. For invalid tokens: Destroy local session and redirect, with appropriate warnings
1. For normal OIDC logout: Verify token, prepare logout URL, and only then destroy the session

Vad1mo: This code changes the flow a bit: 


```go
func (oc *OIDCController) RedirectLogout() {
	sessionData := oc.GetSession(tokenKey)
	ctx := oc.Ctx.Request.Context()
	if sessionData == nil {
		log.Error("OIDC session token not found.")
		oc.SendInternalServerError(fmt.Errorf("OIDC session token not found"))
		return
	}

	oidcSettings, err := config.OIDCSetting(ctx)
	if err != nil {
		log.Errorf("Failed to get OIDC settings: %v", err)
		oc.SendInternalServerError(err)
		return
	}
	if oidcSettings == nil {
		log.Error("OIDC settings is missing.")
		oc.SendInternalServerError(fmt.Errorf("OIDC settings is missing"))
		return
	}
	
	// If OIDC logout is not enabled, just destroy the local session and redirect
	if !oidcSettings.Logout {
		if err := oc.DestroySession(); err != nil {
			log.Errorf("Error occurred in LogOut: %v", err)
			oc.SendInternalServerError(err)
			return
		}
		oc.Controller.Redirect("/", http.StatusFound)
		return
	}
	
	tk, ok := sessionData.([]byte)
	if !ok {
		log.Error("Invalid OIDC session data format.")
		oc.SendInternalServerError(fmt.Errorf("invalid OIDC session data format"))
		return
	}
	
	token := oidc.Token{}
	if err := json.Unmarshal(tk, &token); err != nil {
		log.Errorf("Error occurred in Unmarshal: %v", err)
		oc.SendInternalServerError(err)
		return
	}
	
	// Verify the token before proceeding with OIDC logout
	if token.RawIDToken == "" {
		log.Warning("Empty ID token for offline session.")
		// Even with empty token, we should destroy the local session
		if err := oc.DestroySession(); err != nil {
			log.Errorf("Error occurred in LogOut: %v", err)
			oc.SendInternalServerError(err)
			return
		}
		oc.Controller.Redirect("/", http.StatusFound)
		return
	}
	
	// Verify the token before proceeding with logout
	if _, err := oidc.VerifyToken(ctx, token.RawIDToken); err != nil {
		log.Warningf("Failed to verify token: %v. Only destroying local session.", err)
		// Even with invalid token, we should destroy the local session
		if err := oc.DestroySession(); err != nil {
			log.Errorf("Error occurred in LogOut: %v", err)
			oc.SendInternalServerError(err)
			return
		}
		oc.Controller.Redirect("/", http.StatusFound)
		return
	}
	
	// Handle offline session logout if enabled
	if oidcSettings.LogoutOffline && token.RefreshToken != "" {
		sessionType, err := getSessionType(token.RefreshToken)
		if err == nil {
			// If the session is offline, revoke the refresh token
			if strings.ToLower(sessionType) == "offline" {
				if oidc.EndpointsClaims.RevokeURL != "" {
					if err := revokeOIDCRefreshToken(oidc.EndpointsClaims.RevokeURL, token.RefreshToken, oidcSettings.ClientID, oidcSettings.ClientSecret); err != nil {
						log.Errorf("Failed to revoke the offline session: %v", err)
						// Don't return error here, continue with the normal logout flow
					}
				} else {
					log.Warning("Unable to logout OIDC offline session since the 'revoke_endpoint' is not set.")
				}
			}
		} else {
			log.Warningf("Invalid refresh token for offline session: %s, error: %v", token.RefreshToken, err)
		}
	}
	
	// Check if OIDC provider supports end session endpoint
	if oidc.EndpointsClaims.EndSessionURL == "" {
		log.Warning("Unable to logout OIDC session since the 'end_session_point' is not set.")
		// Still destroy the local session even if provider doesn't support end session
		if err := oc.DestroySession(); err != nil {
			log.Errorf("Error occurred in LogOut: %v", err)
			oc.SendInternalServerError(err)
			return
		}
		oc.Controller.Redirect("/", http.StatusFound)
		return
	}
	
	// Construct the logout URL
	endSessionURL := oidc.EndpointsClaims.EndSessionURL
	baseUrl, err := config.ExtEndpoint()
	if err != nil {
		log.Errorf("Failed to get external endpoint: %v", err)
		oc.SendInternalServerError(err)
		return
	}
	
	postLogoutRedirectURI := fmt.Sprintf("%s/harbor/projects", baseUrl)
	logoutURL := fmt.Sprintf(
		"%s?id_token_hint=%s&post_logout_redirect_uri=%s",
		endSessionURL,
		url.QueryEscape(token.RawIDToken),
		url.QueryEscape(postLogoutRedirectURI),
	)
	
	// Now destroy the Harbor session after verifying the token and preparing the redirect
	if err := oc.DestroySession(); err != nil {
		log.Errorf("Error occurred in LogOut: %v", err)
		oc.SendInternalServerError(err)
		return
	}
	
	log.Info("Redirecting user to OIDC logout:", logoutURL)
	oc.Controller.Redirect(logoutURL, http.StatusFound)
}
```

src/core/controllers/base.go
wy65701436: Regarding the 500 error, we used to only provide more details in the log and not expose them further.

src/core/controllers/base.go
wy65701436: ditto

src/core/controllers/base.go
wy65701436: ditto

src/lib/config/metadata/metadatalist.go
wy65701436: sure, I will revise the description.

src/lib/config/metadata/metadatalist.go
wy65701436: ditto

wy65701436:
> Could you explain the difference between "oidc session" and "oidc offline session"?

According to the [OpenID Connect RP-Initiated Logout specification](https://openid.net/specs/openid-connect-rpinitiated-1_0.html), RP-Initiated Logout should be performed by sending a request to the end_session_endpoint with the ID token. This approach is generally defined in the OpenID specification for all IdPs to properly terminate user sessions.

Based on the Keycloak documentation and what I observed, specifically the section on [offline access](https://www.keycloak.org/docs/latest/server_admin/index.html#_offline-access), an offline session is treated like an active session, created during user authentication with the offline_access scope. To revoke an offline session, a POST request must be made to the revoke endpoint. Since I’ve only tested with Keycloak, I’m uncertain whether other IdPs follow the same flow. As this behavior isn't explicitly defined in the OpenID specification, I’ve included an option (checkbox) for logging out offline sessions.

src/core/controllers/oidc.go
wy65701436: > Session Destruction :
> 
>     1. For non-OIDC logout flows: Destroy session and redirect immediately
> 
>     2. For invalid tokens: Destroy local session and redirect, with appropriate warnings
> 
>     3. For normal OIDC logout: Verify token, prepare logout URL, and only then destroy the session

1, This function is specifically designed for OIDC logout, so we don't need to account for non-OIDC logout flows as well.

2,3 My point is that, regardless of whether the admin has enabled OIDC logout, the local session must always be removed first before redirecting to the IdP for logout. Even if the logout process with the IdP fails, the local session should still be successfully removed.

reasonerjt:
> Based on the Keycloak documentation and what I observed, specifically the section on [offline access](https://www.keycloak.org/docs/latest/server_admin/index.html#_offline-access), an offline session is treated like an active session, created during user authentication with the offline_access scope. To revoke an offline session, a POST request must be made to the revoke endpoint. Since I’ve only tested with Keycloak, I’m uncertain whether other IdPs follow the same flow. As this behavior isn't explicitly defined in the OpenID specification, I’ve included an option (checkbox) for logging out offline sessions.

I need to look into the details of the code for other parts, but IMO in this release we don't need to distinguish "OIDC session" and "offline session".  I don't think of a use case where the user wants to end his "OIDC session" but keep the "offline session".  I think we should just provide the option for the user to end "OIDC session" when he logs out from Harbor, and we always end the "offline session" when the "OIDC session" is terminated.  This seems intuitive enough for me, and we need to debug the corner case like user log out without ending the "offline session" and login again which the ID provider may or may not terminate the existing offline session, which may impact the use case for "CLI secret"

src/portal/src/app/shared/components/navigator/navigator.component.ts
```diff
@@ -199,16 +208,43 @@ export class NavigatorComponent implements OnInit {
...
+                this.router.navigate(
+                    [CommonRoutes.EMBEDDED_SIGN_IN],
+                    navigatorExtra
+                );
+                // Confirm search result panel is close
+                this.searchTrigger.closeSearch(true);
+            },
+            error => {
+                // 403 oidc logout no body;
+                if (this.isOidcLoginMode && error && error.status === 403) {
```
stonezdj, line 230: Does it mean under oidc auth mode, the UI can not handler the 403 error other than redirect?

src/core/controllers/oidc.go
```diff
@@ -244,6 +248,98 @@ func (oc *OIDCController) Callback() {
...
+		return
+	}
+	endSessionURL := oidc.EndpointsClaims.EndSessionURL
+	baseURL, err := config.ExtEndpoint()
+	if err != nil {
+		log.Errorf("Failed to get external endpoint: %v", err)
+		oc.SendInternalServerError(err)
+		return
+	}
+	postLogoutRedirectURI := fmt.Sprintf("%s/harbor/projects", baseURL)
```
stonezdj, line 332: Because user is already logout, the login page seems more reasonable than the /harbor/projects as the default page.

wy65701436: sure, will redirect to the root page.

stonezdj (approved):

src/core/controllers/base.go
Vad1mo: does exposing the error, provide a threat, or might help the user to resolve the issue himself?

src/core/controllers/base.go
```diff
@@ -108,8 +108,46 @@ func (cc *CommonController) Login() {
...
+				ep, err := config.ExtEndpoint()
+				if err != nil {
+					log.Errorf("Failed to get the external endpoint, error: %v", err)
+					cc.CustomAbort(http.StatusUnauthorized, "")
+				}
+				url := strings.TrimSuffix(ep, "/") + common.OIDCLoginoutPath
+				log.Debugf("Redirect user %s to logout page of OIDC provider", u.Username)
+				// Return a json to UI with status code 403, as it cannot handle status 302
+				cc.Ctx.Output.Status = http.StatusForbidden
+				err = cc.Ctx.Output.JSON(struct {
```
reasonerjt, line 140: this redirect is unnecessary, we can move the code to redirect to logout URL into this func.

src/core/controllers/oidc.go
```diff
@@ -244,6 +248,98 @@ func (oc *OIDCController) Callback() {
...
+			log.Warningf("Invalid refresh token for offline session: %s, error: %v", token.RefreshToken, err)
+		}
+	}
+
+	if token.RawIDToken == "" {
+		log.Warning("Unable to log out user session due to the empty ID token.")
+		oc.Controller.Redirect("/", http.StatusFound)
+		return
+	}
+	if _, err := oidc.VerifyToken(ctx, token.RawIDToken); err != nil {
```
reasonerjt, line 316: I don't think we need to verify the token here, we can just redirect the user to the `EndSessionURL`.  

Verifying the token here may cause negative impact, it's possible the ID token is expired at this moment and this should not block user from logging out from Harbor's perspective..

src/core/controllers/oidc.go
reasonerjt: Agree we should respect the setting in OIDC config here since we are interacting with OIDC provider.

I also think this should be moved to "oidc helper" if we think it's general process.

src/core/controllers/oidc.go
```diff
@@ -338,3 +434,52 @@ func secretAndToken(tokenBytes []byte) (string, string, error) {
...
+	data := url.Values{}
+	data.Set("token", refreshToken)
+	data.Set("token_type_hint", "refresh_token")
+	auth := base64.StdEncoding.EncodeToString([]byte(clientID + ":" + clientSecret))
+	req, err := http.NewRequest("POST", revokeURL, bytes.NewBufferString(data.Encode()))
+	if err != nil {
+		return errors.Errorf("failed to create request: %v", err)
+	}
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	req.Header.Set("Authorization", "Basic "+auth)
```
reasonerjt, line 470: req.SetBasicAuth(clientID, clientSecret)

src/core/controllers/oidc.go
```diff
@@ -244,6 +248,98 @@ func (oc *OIDCController) Callback() {
...
+		return
+	}
+	token := oidc.Token{}
+	if err := json.Unmarshal(tk, &token); err != nil {
+		log.Errorf("Error occurred in Unmarshal: %v", err)
+		oc.SendInternalServerError(err)
+		return
+	}
+	if oidcSettings.LogoutOffline && token.RefreshToken != "" {
+		sessionType, err := getSessionType(token.RefreshToken)
```
reasonerjt, line 292: As we discussed, we should always try to revoke the token. 

I even think checking the session type can be optional, what will happen if the session it NOT offline?  Will the OIDC provider return any error?

src/core/controllers/base.go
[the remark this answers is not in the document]
wy65701436: resolved.

src/core/controllers/base.go
```diff
@@ -108,8 +108,46 @@ func (cc *CommonController) Login() {
...
 func (cc *CommonController) LogOut() {
+	// redirect for OIDC logout, excludes the admin user.
+	securityCtx, ok := security.FromContext(cc.Context())
+	if !ok {
+		log.Error("Failed to get security context")
+		cc.CustomAbort(http.StatusInternalServerError, "Internal error.")
+	}
+	principal := securityCtx.GetUsername()
+	if principal != "" {
+		if redirectForOIDC(cc.Ctx.Request.Context(), principal) {
```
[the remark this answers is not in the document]
wy65701436: We do this because in the base controller, we disable the render method, which is responsible for rendering the UI.
See [this link](https://github.com/goharbor/harbor/blob/main/src/core/controllers/base.go#L46). After discussing offline, we decided to keep the redirection logic as is, with the redirection happening twice.

Since we need to exclude admin users from the redirection, it requires a database operation to check the user role.

src/core/controllers/base.go
wy65701436: We do this because in the base controller, we disable the render method, which is responsible for rendering the UI.
See [this link](https://github.com/goharbor/harbor/blob/main/src/core/controllers/base.go#L46). After discussing offline, we decided to keep the redirection logic as is, with the redirection happening twice.

src/core/controllers/oidc.go
wy65701436: removed.

src/core/controllers/oidc.go
wy65701436: updated

src/core/controllers/oidc.go
wy65701436: If the session type is not offline, there is no need to revoke it.

src/portal/src/app/shared/components/navigator/navigator.component.ts
wy65701436: this actually a workaround for oidc logout since we disable the render method in the base handler.

src/core/controllers/oidc.go
wy65701436: resolved.

src/core/controllers/base.go
```diff
@@ -108,8 +108,47 @@ func (cc *CommonController) Login() {
...
+				cc.CustomAbort(http.StatusInternalServerError, "Internal error.")
+			}
+			if u.UserID != 1 {
+				ep, err := config.ExtEndpoint()
+				if err != nil {
+					log.Errorf("Failed to get the external endpoint, error: %v", err)
+					cc.CustomAbort(http.StatusUnauthorized, "")
+				}
+				url := strings.TrimSuffix(ep, "/") + common.OIDCLoginoutPath
+				log.Debugf("Redirect user %s to logout page of OIDC provider", u.Username)
```
reasonerjt, line 138: wrong message.

src/core/controllers/base.go
```diff
@@ -108,8 +108,47 @@ func (cc *CommonController) Login() {
...
+				log.Errorf("User not found: %s, empty result", principal)
+				cc.CustomAbort(http.StatusInternalServerError, "Internal error.")
+			}
+			if u.UserID != 1 {
+				ep, err := config.ExtEndpoint()
+				if err != nil {
+					log.Errorf("Failed to get the external endpoint, error: %v", err)
+					cc.CustomAbort(http.StatusUnauthorized, "")
+				}
+				url := strings.TrimSuffix(ep, "/") + common.OIDCLoginoutPath
```
reasonerjt, line 137: I don't think you need to endpoint to redirect the user under the same hostname

src/core/controllers/oidc.go
```diff
@@ -244,6 +246,84 @@ func (oc *OIDCController) Callback() {
 	}
 }
 
+func (oc *OIDCController) RedirectLogout() {
+	sessionData := oc.GetSession(tokenKey)
+	ctx := oc.Ctx.Request.Context()
+	if sessionData == nil {
```
reasonerjt, line 252: I don't think there it's necessary to return an error, it's possible the user is not onboarded via OIDC if the token is not in session, therefore, I also don't think you need to check the user ID in `LogOut()` in  `base.go`.

I think it's fine to move on the destroy the session if there's no token in the session.

src/core/controllers/oidc.go
```diff
@@ -244,6 +246,84 @@ func (oc *OIDCController) Callback() {
...
+		oc.SendInternalServerError(err)
+		return
+	}
+	logoutURL := fmt.Sprintf(
+		"%s?id_token_hint=%s&post_logout_redirect_uri=%s",
+		endSessionURL,
+		url.QueryEscape(token.RawIDToken),
+		url.QueryEscape(baseURL),
+	)
+	log.Info("Redirecting user to OIDC logout:", logoutURL)
```
reasonerjt, line 323: message should be `Redirect user to logout page of OIDC provider`

src/core/controllers/oidc.go
```diff
@@ -244,6 +246,84 @@ func (oc *OIDCController) Callback() {
 	}
 }
 
+func (oc *OIDCController) RedirectLogout() {
```
reasonerjt, line 249: I think you also need to check the auth mode here?

src/core/controllers/base.go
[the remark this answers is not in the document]
reasonerjt: 1. I don't think we need to exclude admin user, see my comment:
https://github.com/goharbor/harbor/pull/21718/files#r1998476706
2. Even you wanna check it, what's the intention for the line 122 and line 131?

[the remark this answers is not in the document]
wy65701436: resolved.

src/core/controllers/oidc.go
wy65701436: resolved.

reasonerjt (approved):

src/core/controllers/oidc.go
wy65701436: update logic to return to the home page on empty session.

src/core/controllers/base.go
wy65701436: keep the redirection per the offline discussion, in the base handler, the render is set to return nil which is responsible for redirection. So, it needs to redirect to the OIDCLoginoutPath, and then to the IDP to logout user.

Nicolfo:
I'm following this issue since I want to use Harbor with Keycloak as OIDC provider.
I've a question concerning the use of offline access functionality (to support CLI) in conjunction with OIDC RP-initiated logout.

The first thing I tried was enabling the OIDC Session Logout flag (to complete a correct RP-initiated logout) and requesting an offline_access scope. My expected behaviour is that after a logout from the web portal I can continue to use the CLI secret throught docker cli, but this is valid only for a short amount of time (tokens remaining lifespan).

If I remove the flag from OIDC Session Logout CLI secret continue to work (since I'm not terminate the session/revoking the offline token), but I completely lose the track of my offline sessions (every login create a new session) and the only way to revoke them is manually throught the keycloak console.
I don't think it is accettable to not perform complete logout with OIDC when the user terminate session with Harbor, since offline token exist to solve this specific problem.

Is there a way to setup Harbor to perform somewhere similar to AWS cli, where the two session (docker credential and browser) are generated in two different way (oauth2.0 code flow and cli secret request)?

PR #21772, opened 2025-03-23
fix: prevent deadlocks during concurrent tag deletion

### LINKED ISSUE #21062: Harbor DB Deadlock when deleting tags
Some jobs that should not run in parallel; however, some do run in parallel, even if it is set in the code otherwise.

Running jobs in parallel, here the GC job causes the DB to Lock, and put the system in a not usable state.  
image

Bildschirmfoto 2024-10-18 um 10 43 08

- Run jobs in parallel, specifically the GC job

**Versions:**
Please specify the versions of the following systems.

- harbor version: >=2.11

**Additional context:**

image

ambahman:

# Issue being fixed
Fixes #21062
When multiple processes try to delete tags simultaneously in Harbor, they can acquire locks on the same underlying layers in different orders, leading to deadlocks. This is particularly problematic during cascading deletions as each operation requires different table locks.

### DIFF
--- a/src/controller/tag/controller.go
+++ b/src/controller/tag/controller.go
@@ -17,6 +17,7 @@ package tag
 import (
 	"context"
 	"regexp"
+	"sort"
 	"time"
 
 	"github.com/goharbor/harbor/src/common/utils"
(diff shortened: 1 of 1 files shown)


Vad1mo (changes requested):
Equal important as the fix is the guide/flow to be able to reporduce the bug...

PR #22047, opened 2025-06-03
Add Conditional Immutability Rules Compatible with Retention Policy

### LINKED ISSUE #10506: Tag Retention/Immutability
What can we help you?
I have 2 Tag Retention Rules set:
Rule 1: For the repositories matching **, retain the images pulled within the last 7 days with tags matching **
Rule 2: For the repositories matching **, retain the most recently pulled 5 images with tags matching **

I also have set Tag Immutability for all repos and tags.

---
A tag is set to be deleted due to Tag Retention rules, however Immutability prevents it from being deleted. 

Is this by design or a limitation in Docker itself? 
Is there a way to for delete an Immutable tag if it's deemed deletable according to Tag Retention rules?

### LINKED ISSUE #20989: Retention Policies for Immutability Tags
Harbor only supports adding immutability rules. Once they are set, they need to be removed manually, one by one. 
The only way to delete immutable objects is to turn off immutability, 
run retention again, and turn it back on when you think you've deleted what you wanted.

The Immutability Tag policy should be extended to not only "apply" immutability tags but also remove immutability tags.

It would be false to delete images with Immutability Tags. To work around that problem, we need to introduce rules to  immutability tags. 

- Immutability rules will have one new action → RELEASE, next to the already existing and implicit action APPLY.
    - This can be done as a dropdown or a toggle
- An additional dropdown provides conditions under which the action is applied.
    - Possible options for condition action → APPLY
        - Immediately (this is like the current implicit behavior)
        - Pulled count greater → # count
        - tbd…
    - Possible options for condition action → RELEASE
        - Artifacts not pulled within the last # of days  → # count days
        - Artifacts pushed # days ago → # count days
        - tbd…

image

Mockup of the new tag immutability rule, with actions and conditions.

- under investigation

**Additional context**
There are a few options under which conditions apply or release can happen, but I think there might be even more...

### LINKED ISSUE #22543: Unable to Delete Artifacts marked as Immutable
Using the Harbor UI when I navigate to an artifact that is 'immutable' I am unable to delete the artifact even through other components (ex retention policies) can successfully delete the artifact.

Immutable tags should mean the specific tags can't be edited/removed/duplicated/ect... but not that the artifact itself is permanent. 

1) Create Project
2) Configure Tag Immutability rules for the project
3) Upload artifact with a tag
4) Attempt to Delete the artifact through the UI

**Versions:**
- harbor version: v2.13.1-9384ed0e

**Additional context:**

bupd:
### Default Backwards Compatible Immutable Rule
image

### Updated Rule Selection Similar to Tag Retention Policies
image

### Better Immutability Rules
image

### New and Old Immutability Rule Comparision
image

## Summary
Currently, immutability overrides retention policy, making it impossible to delete tags once they are marked immutable even if a tag retention policy says they should be deleted. The only workaround is to manually deactivate immutability, run the retention job, then re-enable immutability.

Currently the immutable tags that's once immutable, always immutable.

This PR solves that problem by making immutability conditional according to defined rules. which adds better and dynamic lifecycle control to immutability.

## User Scenario
“As a user, I am trying to use both immutability and retention together.”

### Example:
1. I set a retention policy - `delete tags not pulled in the last 10 days`
2. I set an immutability policy - `make tags immutable if they were pulled in the last 15 days`

With this feature:
- Tags pulled in the last 15 days are immutable (protected)
- Tags older than 15 days become mutable
- The retention policy can now clean up the old, unused tags (e.g., pulled 20+ days ago)

Previously, this was not possible.

# Issue being fixed
- Fixes #20989
- Fixes #10506
- Fixes #22543

- [x] Label the PR as needed. "release-note/enhancement"

Thanks


wy65701436:
@bupd Could you please draft a proposal for this issue first? I believe this constitutes a behavior change and could potentially impact all existing Harbor users?

And we've reached the FF of v2.14, so if we plan to have a update, let's put it in the future release?

Vad1mo:
> @bupd Could you please draft a proposal for this issue first? I believe this constitutes a behavior change and could potentially impact all existing Harbor users?
> 
> And we've reached the FF of v2.14, so if we plan to have a update, let's put it in the future release?

this new feature has no impact on backwards compatibility.

Vad1mo (approved):

wy65701436 (changes requested):
We need a proposal for this feature enhancement to document the details first, thanks.

sluebbert:
We are still interested in getting this feature.

tgagor:
I'm really interested in this functionality as our immutable tags stared piling up.
Do you have any predictions when it might be released?

bupd:
This feature is already available in [github.com/container-registry/harbor-next](https://github.com/container-registry/harbor-next), our experimental playground for Harbor.

If you're interested in this functionality, I recommend trying harbor-next rather than waiting for this implementation to land here.

PR #22147, opened 2025-07-02
Fix: Audit Log Eventtype antipattern in System Settings UI

### LINKED ISSUE #22054: Bad UX in Audit Logs Selection
Image

The audit logs UX design follow two bad practices when using checkboxes:

- Double negative
- Horizontal Checkboxes

The heading says, “Disable Audit Log Event Type." Then, each checkbox represents an event type (e.g., "Create artifact," "Login user").

- If a user **checks** "Create artifact," are they disabling the logging of "Create artifact" events, or are they enabling it?
- If a user **unchecks** "Create artifact," does that mean logging for "Create artifact" is enabled?

This ambiguity creates cognitive load and a high potential for user error. Users have to pause and think through the logic, rather than intuitively understanding the action.

- Horizontal Checkboxes

**A better approach would be:**

- A positive framing like "Enable Audit Log Event Types," where checking a box clearly means that event type *will* be logged.
- Or, if the goal is to select specific events *not* to log, perhaps a clearer label like "Select Event Types to Exclude from Audit Log."

Here are some examples 

- https://www.nngroup.com/articles/checkboxes-design-guidelines/

Image

Image

bupd:

- updated logic from disabled to enabled checkboxes.
- update i18n to reflect the logic change.

# Issue being fixed
- Fixes #22054

## After Update
image

- [x] Label the PR as needed. "release-note/update"


chlins (approved):
lgtm

PR #22175, opened 2025-07-14
Delete local cache if the artifact doesn't exist in upstream registry

### LINKED ISSUE #20596: Unreasonable deletion operation logic in proxy cache
How can we help you?

Proxy cache tries to delete remote repo name in the local registry, which is not reasonable because the remote repo name has been trimmed the project name and it maybe a valid image in other non-proxy cache project.

https://github.com/goharbor/harbor/blob/6a38ed3d7769e3598c6cf829aae4e0e152f93a83/src/controller/proxy/controller.go#L176

stonezdj:
fixes #20596

# Issue being fixed
Fixes #(issue)

### DIFF
--- a/src/controller/proxy/controller.go
+++ b/src/controller/proxy/controller.go
@@ -149,6 +149,22 @@ type ManifestList struct {
 	ContentType string
 }
 
+// getManifestDigestInLocal get the artifact digest in local
+func (c *controller) getManifestDigestInLocal(ctx context.Context, art lib.ArtifactInfo) (string, error) {
+	// Get the manifest from local registry
+	a, err := c.local.GetManifest(ctx, art)
+	if err != nil {
+		return "", err
+	}
+	if a == nil {
+		return "", errors.NotFoundError(fmt.Errorf("manifest %v not found in local registry", art.Repository))
+	}
+	if len(a.Digest) == 0 {
+		return "", errors.NotFoundError(fmt.Errorf("manifest %v not found in local registry", art.Repository))
+	}
+	return a.Digest, nil
+}
+
 // UseLocalManifest check if these manifest could be found in local registry,
 // the return error should be nil when it is not found in local and need to delegate to remote registry
 // the return error should be NotFoundError when it is not found in remote registry
@@ -172,6 +188,16 @@ func (c *controller) UseLocalManifest(ctx context.Context, art lib.ArtifactInfo,
 		return false, nil, err
 	}
 	if !exist || desc == nil {
+		dig, err := c.getManifestDigestInLocal(ctx, art)
+		if err != nil {
+			// skip to delete when error, use debug level log to avoid too many logs when the manifest is removed from upstream
+			log.Debugf("failed to get manifest digest in local, error: %v, skip to delete it, art %+v", err, art)
+		} else {
+			go func() {
+				c.local.DeleteManifest(art.Repository, dig)
+				log.Infof("delete manifest %s with digest %s", art.Repository, dig)
+			}()
+		}
 		return false, nil, errors.NotFoundError(fmt.Errorf("repo %v, tag %v not found", art.Repository, art.Tag))
 	}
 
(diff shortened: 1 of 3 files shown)


src/pkg/registry/client.go
```diff
@@ -356,6 +358,19 @@ func (c *client) DeleteManifest(repository, reference string) error {
 	return nil
 }
 
+func (c *client) DeleteManifestQuiet(repository, reference string) error {
```
wy65701436: add UT for this?

src/controller/proxy/controller.go
```diff
@@ -172,6 +172,9 @@ func (c *controller) UseLocalManifest(ctx context.Context, art lib.ArtifactInfo,
 		return false, nil, err
 	}
 	if !exist || desc == nil {
+		go func() {
+			c.local.DeleteManifestQuiet(ctx, art.Repository, art.Tag)
```
wy65701436: log the failure?

src/controller/proxy/local.go
```diff
@@ -135,6 +138,27 @@ func (l *localHelper) DeleteManifest(repo, ref string) {
 	}
 }
 
+// DeleteManifestQuiet delete manifest without existing check
+func (l *localHelper) DeleteManifestQuiet(ctx context.Context, repo, ref string) {
+	log.Infof("Remove tag from repo if it is exist, repo: %v ref: %v", repo, ref)
```
wy65701436: add proxy-cache in the log for easy to understand why the deletion is happen.

src/controller/proxy/local.go
```diff
@@ -135,6 +138,27 @@ func (l *localHelper) DeleteManifest(repo, ref string) {
...
+	// convert tag to digest
+	bCtx := orm.Copy(ctx)
+	opt := &artifact.Option{}
+	art, err := l.artifactCtl.GetByReference(bCtx, repo, ref, opt)
```
wy65701436: what about if the art is a index? Shouldn't we handle the references of this art?

src/controller/proxy/local.go
```diff
@@ -135,6 +138,27 @@ func (l *localHelper) DeleteManifest(repo, ref string) {
...
+	if err != nil {
+		log.Debug("failed to get digest, skip to delete it")
```
wy65701436: add the repo/ref in the log.

chlins: log.Debugf("failed to get artifact, repo: %s, ref: %s, skip to delete it")

src/controller/proxy/local.go
chlins: I think `%s` is enough.

src/pkg/registry/client.go
wy65701436: What is the difference between DeleteManifestQuiet and DeleteManifest?

src/controller/proxy/local.go
```diff
@@ -135,6 +138,27 @@ func (l *localHelper) DeleteManifest(repo, ref string) {
...
+		return
+	}
+	if art == nil {
+		return
+	}
+	log.Infof("artifact does not exist in remote registry, going to remove it in local, artifact: %v@%v", repo, art.Digest)
```
chlins: `%s`

src/controller/proxy/controller.go
stonezdj: If there is a deleted artifact in the upstream, it will always logs the failure message, it may confuse the user because it is not an error.  when deleting artifact in backend, there is another log message.

src/controller/proxy/local.go
```diff
@@ -135,6 +138,27 @@ func (l *localHelper) DeleteManifest(repo, ref string) {
...
+	if err := l.registry.DeleteManifestQuiet(repo, art.Digest); err != nil {
```
wy65701436: if the purpose is to remove the artifact, I believe that this step is not needs since the GC will handle this part.

chlins:
The naming of `DeleteManifestQuiet `  seems not match with the implementation.

chlins:
Do we really need to remove the manifest in distribution, as the GC maybe can handle this part?

stonezdj:
> Do we really need to remove the manifest in distribution, as the GC maybe can handle this part?

The DeleteManifestQuiet deletes the manifest in database, GC will remove it from file system eventually.

chlins (approved):
lgtm

Issue #22184, opened 2025-07-18
Coalesce upstream requests where possible


Strainy:
**Is your feature request related to a problem? Please describe.**
I'm currently using Harbor to service a large number of K8S nodes. I'm facing a problem where occasionally a large number of those nodes concurrently attempt to pull the same artifact from the registry. If Harbor happens to receive N requests at the same time, I typically see it making a commensurate number of requests to the remote repository for manifests/blobs. For relatively large artifacts/blobs, this has the effect of completely saturating our bandwidth and pulls never completing; requiring manual intervention.

**Describe the solution you'd like**
Harbor should coalesce remote requests for the same manifests/blobs. This could work similarly to how pushes to the local registry are de-duplicated [here](https://github.com/goharbor/harbor/blob/main/src/controller/proxy/inflight.go). The end result should basically be **one remote fetch for one artifact**, with all subsequent requests being served from the local registry cache. 

**Describe the main design/architecture of your solution**
I think the ideal solution here is to use the [sync/singleflight](https://pkg.go.dev/golang.org/x/sync/singleflight) package to de-duplicate remote requests within a single core instance. The specific manner in which this can be handled differs based on whether we're dealing with a blob or manifest request.

For the manifest requests the solution is pretty simple. Because we can buffer the manifest response, we can just update the [proxy controller](https://github.com/goharbor/harbor/blob/main/src/controller/proxy/controller.go) to wrap existing manifest fetching implementations. Only one fetch for the manifest will occur at a time, with all concurrent requests waiting on the initial fetch and receiving the same response on completion.

```go
// This singleflight group is used to deduplicate concurrent manifest requests
result, err, _ := manifestFetchGroup.Do(artifactKey, func() (any, error) {
	log.Debugf("Fetching manifest from remote registry, url:%v", remoteRepo)

	man, dig, err := remote.Manifest(remoteRepo, ref)
	if err != nil {
		return nil, err
	}
	ct, _, err := man.Payload()
	if err != nil {
		return nil, err
	}

	// Push manifest in background

    return man
}
```

For the blob requests, this is going to be a bit trickier. This is because the proxy controller implementation currently propagates an `io.ReadCloser` back the caller. The `io.ReadCloser` can obviously only be read once, which necessitates a different approach. 

[...]

PR #22185, opened 2025-07-18
Add singleflight to deduplicate concurrent blob/manifest requests

Strainy:
This PR implements request deduplication in the proxy controller to reduce network bandwidth saturation from concurrent requests for the same large blobs.
- Added [sync/singleflight](golang.org/x/sync/singleflight) to prevent multiple concurrent requests for the same manifest or blob from hitting the remote registry
- Added configuration for enabling this behavior via the ENABLE_ASYNC_LOCAL_CACHING environment variable
- Added a metric to track remote registry requests by operation, registry name, and success status

A more comprehensive overview of the propose solution has been provided in the linked issue. But the TL;DR here is:
- When async caching is enabled, blobs are streamed directly from remote and cached in background (current behavior)
- When async caching is disabled, concurrent blob requests are queued and served from local cache after first request completes
- All concurrent remote manifest requests are de-duplicated

# Issue being fixed
Fixes https://github.com/goharbor/harbor/issues/22184


Strainy:
Thanks the feedback @Vad1mo!

> If two Harbor core pods receive requests for a blob they don’t have cached:
> Each pod will independently fetch from the remote registry (unless a shared storage/cache backend ensures deduplication).

Yep, the limitation of request de-duplication occurring within a single instance is understood. We could implement a global lock in Redis, but I opted to implement it in this manner to optimize for simplicity and availability. In reality, this is still a significant improvement even if you have multiple Harbor instances.

> If caching is async, or if the pods' local caches diverge, one pod might serve cached data while another fetches fresh data.

Sorry, what is the problem here? For a given set of concurrent requests, there should be no difference between the remote and cached results, right?

stonezdj:
@Strainy Thanks for your PR. most of the time, one connection is insufficient to pull proxy cache images; it is better to be a configurable item in the project level, and the max upstream connection should be controlled at the app level (through redis) if harbor is deployed in k8s.  please refer and review this pr: https://github.com/goharbor/harbor/pull/22348.

PR #22207, opened 2025-07-26
Fix: Improper Neutralization of Input During Escaping of Output Reflected cross-site scripting

odaysec:
https://github.com/goharbor/harbor/blob/ebc340a8f7bb0cab4a3a128a5310122b2e895b16/src/lib/response_recorder.go#L38-L38

Directly writing user input (an HTTP request parameter) to an HTTP response without properly sanitizing the input first, allows for a cross-site scripting vulnerability.This kind of vulnerability is also called reflected cross-site scripting, to distinguish it from other types of cross-site scripting.

fix the reflected XSS vulnerability, we should ensure that any user-controlled data written to an HTTP response is properly escaped for HTML context. In this case, the `Write` method of `ResponseRecorder` should escape the data if it is being used to write log output that may contain user input. The best way to do this is to use Go's `html.EscapeString` function to sanitize the data before writing it to the response. Since `Write` receives a byte slice, we should convert it to a string, escape it, and then write the escaped string as bytes. This change should be made in `src/lib/response_recorder.go`, specifically in the `Write` method. need to import the `html` package in this file to use `html.EscapeString`.

### DIFF
--- a/src/lib/response_recorder.go
+++ b/src/lib/response_recorder.go
@@ -14,7 +14,10 @@
 
 package lib
 
-import "net/http"
+import (
+	"net/http"
+	"html"
+)
 
 // NewResponseRecorder creates a response recorder
 func NewResponseRecorder(w http.ResponseWriter) *ResponseRecorder {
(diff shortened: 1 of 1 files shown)


reasonerjt:
Considering the usage case of `ResponseRecorder`, which is mainly for capturing the response of Harbor's API.
Is there really a case to potentially form this attack?

reasonerjt:
@wy65701436 Could you please double check and follow up?

PR #21982, opened 2025-05-08
Arm64 Support and Multi-Architecture Image Publishing for Harbor

ranimandepudi:
Proposal available at: https://github.com/goharbor/community/pull/262

Closes: [#21825](https://github.com/goharbor/harbor/issues/21825)

Feedback and suggestions are welcome!

Looking forward to enabling official Arm64 support in the upcoming Harbor release.

Thanks 
- Rani Chowdary M, Arm.

PR #22229, opened 2025-08-05
Arm64 Official Support 

ranimandepudi:
Tested and confirmed: Harbor functions fully on both Arm64 and Amd64 platforms, including the UI, API, registry, and multi-architecture artifact handling.

Commands: 
- make clean 
- make build -e VERSIONTAG=dev -e BASEIMAGETAG=dev -e PULL_BASE_FROM_DOCKERHUB=false -e  PUSHBASEIMAGE=false -e BUILD_BASE=true
- docker images | grep goharbor | grep dev 
-  cd make/ 
- cp harbor.yml.tmpl harbor.yml 
-  vim harbor.yml 
-   make prepare 
- make prepare 
- sudo ./install.sh  
 
 
Building locally (Harbor components and base images)
These steps will:
	•	Build Harbor entirely from source
	•	Use your own locally built base images
	•	Run natively on Arm64/AMD64

Open for suggestions !! Closes #21982

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,12 +73,14 @@ CORE_PATH=$(BUILDPATH)/src/core
 PORTAL_PATH=$(BUILDPATH)/src/portal
 CHECKENVCMD=checkenv.sh
 ARCH ?= $(shell uname -m)
-
+ifeq ($(ARCH), aarch64)
+  ARCH := arm64
+endif
 # parameters
 REGISTRYSERVER=
 REGISTRYPROJECTNAME=goharbor
 DEVFLAG=true
-TRIVYFLAG=true
+TRIVYFLAG=false
 EXPORTERFLAG=false
 HTTPPROXY=
 BUILDREG=true
@@ -123,32 +125,15 @@ DISTRIBUTION_SRC=https://github.com/goharbor/distribution.git
 
 # dependency binaries
 REGISTRYURL=https://storage.googleapis.com/harbor-builds/bin/registry/release-${REGISTRYVERSION}/registry
-
 ifeq ($(ARCH), arm64)
-
-
-   # Arm64-specific version and source values
-   TRIVYVERSION=v0.58.2
-   TRIVYADAPTERVERSION=v0.32.3
-   REGISTRY_SRC_TAG=v2.8.3
-   DISTRIBUTION_SRC=https://github.com/distribution/distribution.git
-   TRIVY_DOWNLOAD_URL=https://github.com/aquasecurity/trivy/releases/download/$(TRIVYVERSION)/trivy_$(TRIVYVERSION)_Linux-ARM64.tar.gz
-   TRIVY_ADAPTER_DOWNLOAD_URL=https://github.com/goharbor/harbor-scanner-trivy/archive/refs/tags/$(TRIVYADAPTERVERSION).tar.gz
-
+   TRIVY_DOWNLOAD_URL = https://github.com/aquasecurity/trivy/releases/download/$(TRIVYVERSION)/trivy_$(TRIVYVERSION:v%=%)_Linux-ARM64.tar.gz
 
 else
-  
-   # Default (x86_64 or others) versions   
-   TRIVYVERSION=v0.61.0
-   TRIVYADAPTERVERSION=v0.33.0-rc.2
-   REGISTRY_SRC_TAG=release/2.8
-   DISTRIBUTION_SRC=https://github.com/goharbor/distribution.git
-   TRIVY_DOWNLOAD_URL=https://github.com/aquasecurity/trivy/releases/download/$(TRIVYVERSION)/trivy_$(TRIVYVERSION:v%=%)_Linux-64bit.tar.gz
-   TRIVY_ADAPTER_DOWNLOAD_URL=https://github.com/goharbor/harbor-scanner-trivy/archive/refs/tags/$(TRIVYADAPTERVERSION).tar.gz
+   TRIVY_DOWNLOAD_URL = https://github.com/aquasecurity/trivy/releases/download/$(TRIVYVERSION)/trivy_$(TRIVYVERSION:v%=%)_Linux-64bit.tar.gz
 
 endif
 
-
+TRIVY_ADAPTER_DOWNLOAD_URL=https://github.com/goharbor/harbor-scanner-trivy/archive/refs/tags/$(TRIVYADAPTERVERSION).tar.gz
 define VERSIONS_FOR_PREPARE
 VERSION_TAG: $(VERSIONTAG)
 REGISTRY_VERSION: $(REGISTRYVERSION)
--- a/Makefile
+++ b/Makefile
@@ -72,12 +72,13 @@ TOOLSPATH=$(BUILDPATH)/tools
 CORE_PATH=$(BUILDPATH)/src/core
 PORTAL_PATH=$(BUILDPATH)/src/portal
 CHECKENVCMD=checkenv.sh
+ARCH ?= $(shell uname -m)
 
 # parameters
 REGISTRYSERVER=
 REGISTRYPROJECTNAME=goharbor
 DEVFLAG=true
-TRIVYFLAG=false
+TRIVYFLAG=true
 EXPORTERFLAG=false
 HTTPPROXY=
 BUILDREG=true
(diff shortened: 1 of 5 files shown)


make/photon/prepare/templates/docker_compose/docker-compose.yml.jinja
```diff
@@ -1,6 +1,7 @@
 services:
   log:
     image: goharbor/harbor-log:{{version}}
+    platform: {{platform}}
```
AaronDewes, line 4: Why are these changes necessary? Docker should normally automatically pick the correct platform.

Makefile
```diff
@@ -72,12 +72,13 @@ TOOLSPATH=$(BUILDPATH)/tools
...
 PORTAL_PATH=$(BUILDPATH)/src/portal
 CHECKENVCMD=checkenv.sh
+ARCH ?= $(shell uname -m)
 
 # parameters
 REGISTRYSERVER=
 REGISTRYPROJECTNAME=goharbor
 DEVFLAG=true
-TRIVYFLAG=false
+TRIVYFLAG=true
```
AaronDewes: I think this should be reverted?

Makefile
```diff
@@ -122,8 +123,31 @@ DISTRIBUTION_SRC=https://github.com/goharbor/distribution.git
 
 # dependency binaries
 REGISTRYURL=https://storage.googleapis.com/harbor-builds/bin/registry/release-${REGISTRYVERSION}/registry
-TRIVY_DOWNLOAD_URL=https://github.com/aquasecurity/trivy/releases/download/$(TRIVYVERSION)/trivy_$(TRIVYVERSION:v%=%)_Linux-64bit.tar.gz
-TRIVY_ADAPTER_DOWNLOAD_URL=https://github.com/goharbor/harbor-scanner-trivy/archive/refs/tags/$(TRIVYADAPTERVERSION).tar.gz
+
+ifeq ($(ARCH), arm64)
+
+
```
AaronDewes: A lot of empty lines in this block, they should be removed imo

Makefile
```diff
@@ -122,8 +123,31 @@ DISTRIBUTION_SRC=https://github.com/goharbor/distribution.git
...
+   # Arm64-specific version and source values
+   TRIVYVERSION=v0.58.2
+   TRIVYADAPTERVERSION=v0.32.3
+   REGISTRY_SRC_TAG=v2.8.3
+   DISTRIBUTION_SRC=https://github.com/distribution/distribution.git
```
AaronDewes: I think registry version could be the same for all architectures, or was there a specific reason for this?

Makefile
```diff
@@ -122,8 +123,31 @@ DISTRIBUTION_SRC=https://github.com/goharbor/distribution.git
...
 REGISTRYURL=https://storage.googleapis.com/harbor-builds/bin/registry/release-${REGISTRYVERSION}/registry
-TRIVY_DOWNLOAD_URL=https://github.com/aquasecurity/trivy/releases/download/$(TRIVYVERSION)/trivy_$(TRIVYVERSION:v%=%)_Linux-64bit.tar.gz
-TRIVY_ADAPTER_DOWNLOAD_URL=https://github.com/goharbor/harbor-scanner-trivy/archive/refs/tags/$(TRIVYADAPTERVERSION).tar.gz
+
+ifeq ($(ARCH), arm64)
+
+
+   # Arm64-specific version and source values
+   TRIVYVERSION=v0.58.2
+   TRIVYADAPTERVERSION=v0.32.3
```
AaronDewes: Why are you using an older Trivy version for arm64? I think the trivy version can be the same for all architectures (The download URL would of course be different)

AaronDewes (changes requested):

make/photon/prepare/templates/docker_compose/docker-compose.yml.jinja
ranimandepudi: These changes are necessary because Docker doesn't always auto-select the correct platform especially when using locally built, single-architecture images without a manifest. Explicitly setting platform ensures consistent behavior across Arm and AMD during benchmarking, testing, and templated deployments.

Makefile
ranimandepudi: Yeah I will do that !!

Makefile
ranimandepudi: Will update

Makefile
ranimandepudi: You are right, the image version is consistent, but the REGISTRY_SRC_TAG differs. For Arm, we used distribution/distribution@v2.8.3 instead of Harbor's release/2.8 fork to avoid potential compatibility issues during the initial port. If there's no conflict, we can unify this and use the same source tag across architectures.

make/photon/db/Dockerfile.base
```diff
@@ -6,9 +6,16 @@ RUN tdnf install -y shadow >> /dev/null \
     && groupadd -r postgres --gid=999 \
     && useradd -m -r -g postgres --uid=999 postgres
 
-RUN tdnf install -y postgresql14-server >> /dev/null
-RUN tdnf install -y gzip postgresql15-server findutils bc >> /dev/null \
-    && mkdir -p /docker-entrypoint-initdb.d \
+# Detect architecture and install different packages accordingly
+ARG TARGETARCH
+RUN if [ "$TARGETARCH" = "amd64" ]; then \
```
Vad1mo: Why do we install sql14 for AMD64 but not for ARM?

Makefile
Vad1mo: it should be the same

Vad1mo (changes requested):

make/photon/db/Dockerfile.base
AaronDewes: Postgres 14 is only necessary for upgrades from Postgres 14 to Postgres 15 (Because for upgrading, Postgres requires the binaries of both the old and new version).

Because ARM64 support wasn't official before, there will be no users already running Harbor on ARM64 that would need Postgres to be updated.

ranimandepudi: @Vad1mo 
I actually found PostgreSQL 14 also available for arm64:
`tdnf list available | grep postgresql14`
postgresql14.aarch64                         14.18-1.ph5          photon-updates
postgresql14-client.aarch64                  14.18-1.ph5          photon-updates
postgresql14-contrib.aarch64                 14.18-1.ph5          photon-updates
postgresql14-devel.aarch64                   14.18-1.ph5          photon-updates
postgresql14-docs.aarch64                    14.18-1.ph5          photon-updates
postgresql14-i18n.aarch64                    14.18-1.ph5          photon-updates
postgresql14-libs.aarch64                    14.18-1.ph5          photon-updates
postgresql14-llvmjit.aarch64                 14.18-1.ph5          photon-updates
postgresql14-plperl.aarch64                  14.18-1.ph5          photon-updates
postgresql14-plpython3.aarch64               14.18-1.ph5          photon-updates
postgresql14-pltcl.aarch64                   14.18-1.ph5          photon-updates
postgresql14-server.aarch64                  14.18-1.ph5          photon-updates
root [ / ]# 

I can change it to 
`RUN tdnf install -y postgresql14-server postgresql15-server gzip findutils bc`

AaronDewes: As per my comment, I think that is unnecessary.

I'm not a harbor maintainer, so you shouldn't listen to me, but I think it can still be left out (depending on whether the maintainers think it's necessary).

auggie246:
I am facing an issue while trying to build using your fork and branch.

```
Dockerfile:22
--------------------
  20 |     USER harbor
  21 |     ENTRYPOINT ["/harbor/entrypoint.sh"]
  22 | >>> COPY make/photon/prepare/versions /harbor/
  23 |
--------------------
ERROR: failed to build: failed to solve: failed to compute cache key: failed to calculate checksum of ref 75a944fc-8bd9-4b3e-a9f2-76c2afe9a59e::xt3xo8tb9ggr0n4fc5evqgiz0: "/make/photon/prepare/versions": not found
make[1]: *** [/home/ubuntu/harbor/make/photon/Makefile:134: _build_core] Error 1
make[1]: Leaving directory '/home/ubuntu/harbor'
make: *** [Makefile:402: build] Error 2
```

`ls make/photon/prepare` shows that there's no `versions` to copy

EDIT: `versions` file is only created when `make package_offline` or `make package_online` but not `make build`

ocontant:
Can the maintainer provide context to why this PR is not moving forward ? 

The comment show no evidences of possible road block, and other communities have ported Harbor to arm64.
What is preventing Harbor to support arm64 ?

counten:
why does the latest harbor-offline-installer-v2.15.2.tgz not work on arm64

OrlinVasilev:
@ranimandepudi can you check that please?

OrlinVasilev:
@counten did you do any troubleshooting and/or investigation - anything that you did on your side will be useful information :)

PR #22270, opened 2025-08-21
Fix to allow repo replication when using non-standard port for core

nickhyoti:
Use CORE_LOCAL_URL instead of hardcoded setting

In our environment we are using a non-standard port setting for Core.  
Instead of 8080, we are using 8000.
When we try and setup repository replication we get a message 

Failed to create adapter for destination registry http://core:
Get "http://127.0.0.1:8080/api/version": dial tcp http://127.0.0.1:8080:connect: connection refused

This is because there is a hardcoded port in the function GetURL.
There is already an environment variable defined which has the correct URL for core, so use that instead

# Issue being fixed
Fixes unable to setup repo replication when using non-standard port on core

- [X] Label the PR as needed. "release-note/ignore-for-release, release-note/new-feature, release-note/update, release-note/enhancement, release-note/community, release-note/breaking-change, release-note/docs, release-note/infra, release-note/deprecation"
- [X] Made sure tests are passing and test coverage is added if needed.


src/pkg/reg/adapter/harbor/base/client.go
```diff
@@ -147,7 +147,10 @@ func (c *Client) GetURL() string {
 	// inside core, returns the "127.0.0.1" as URL to avoid the issue:
 	// https://github.com/goharbor/harbor-helm/issues/222
 	// when harbor is deployed on Kubernetes with hairpin mode disabled
-	url := "http://127.0.0.1:8080"
+
+	// get the URL from the environment so can use non-standard port
+        url := os.Getenv("CORE_LOCAL_URL")
```
bupd: I would suggest to add the default 8080 if the env is not found as fallback. 

Also you would need to update Helm chart values. So that people can configure this directly from Helm chart.

bupd (changes requested):
@nickhyoti Thanks for your contributions 

I have added my suggestion for improvements.

src/pkg/reg/adapter/harbor/base/client.go
nickhyoti: Thanks for the feedback,
I've added the check and now set the default if the env variable not set.

WRT to changing Helm, I was going to start with altering the  Docker configuration.  As thats likely to be a longer task, I'll do that separately

src/pkg/reg/adapter/harbor/base/client.go
```diff
@@ -147,7 +147,13 @@ func (c *Client) GetURL() string {
...
```
chlins: Indentation seems to be problematic.

src/pkg/reg/adapter/harbor/base/client.go
```diff
@@ -147,7 +147,13 @@ func (c *Client) GetURL() string {
...
+	if (url == "") {
```
chlins: `len(url) == 0` or `url == ""`

chlins (changes requested):

stonezdj:
How do you specify another port other than 8080? How do you deploy the Harbor? docker-compose or kubernetes?

src/pkg/reg/adapter/harbor/base/client.go
nickhyoti: thanks - fixed

src/pkg/reg/adapter/harbor/base/client.go
nickhyoti: Thanks - added condition

bupd:
not stale. @nickhyoti fix DCO.

nickhyoti:
> not stale. @nickhyoti fix DCO.
tried following the instructions to fix this but get a permissions error.

PR #22297, opened 2025-08-28
Harbor-Core: prevent zero-sized blob keys from being cached in Redis

swood:
Prevents Harbor pull failures caused by zero-sized blob size keys in Redis cache.

Changes:
- Skip caching zero/negative blob sizes in upload middleware
- Treat cached zero values as key not found during retrieval
- Add proactive cleanup of corrupted Registry blob keys during reads
- Maintain backward compatibility with comprehensive logging

Fixes Content-Length header 0 does not match expected size and unexpected EOF errors during image pull operations.

### DIFF
--- a/src/server/middleware/blob/get_blob.go
+++ b/src/server/middleware/blob/get_blob.go
@@ -1,0 +1,65 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package blob
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/goharbor/harbor/src/lib"
+	"github.com/goharbor/harbor/src/lib/log"
+	libredis "github.com/goharbor/harbor/src/lib/redis"
+	"github.com/goharbor/harbor/src/server/middleware"
+)
+
+// GetBlobMiddleware cleans up zero-sized blob keys from Redis before serving blob
+func GetBlobMiddleware() func(http.Handler) http.Handler {
+	return middleware.BeforeRequest(func(r *http.Request) error {
+		// Get blob digest from request context
+		blobInfo := lib.GetArtifactInfo(r.Context())
+		if blobInfo.Digest == "" {
+			return nil // No digest, skip cleanup
+		}
+
+		// Clean up zero-sized blob key in Redis
+		key := fmt.Sprintf("blobs::%s", blobInfo.Digest)
+		rc, err := libredis.GetRegistryClient()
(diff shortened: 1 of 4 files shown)


reasonerjt:
@swood 

Thanks for this PR.  Are you saying the "0 sized" layer will fail to pull from Harbor?  
Could you provide an image so we can easily reproduce?

swood:
@reasonerjt.

The problem isn't with "0 sized" layer. The problem occurs when the layer size differs from what is displayed in the Redis cache. In my case, it was always a key in Redis with `size=0`.

src/server/middleware/blob/get_blob.go
Vad1mo, line 1: The CleanupInvalidBlobSizeKeys function should definitely be moved to Harbor's jobservice instead of being implemented in middleware.

**The current middleware approach scans ALL upload keys (upload:*:size) on every blob request, which could impact performance significantly under high load.**



Harbor already uses its jobservice for similar cleanup operations:

- System Artifact Cleanup (src/jobservice/job/impl/systemartifact/cleanup.go) - Cleans up system artifacts periodically
- Audit Log Purge (src/jobservice/job/impl/purge/purge.go) - Purges old audit logs
- Garbage Collection - Handles image cleanup and space reclamation


Jobservice provides controlled execution environment with proper timeouts, retries, and error handling
and can run during maintenance windows or low-traffic periods

Vad1mo:
if Harbor skips caching zero/negative blob sizes and ignores them when retrieving.

@swood, can you explain why it needs a cleanup job at all? 


Did you experience the problem of zero-sized blobs despite not caching them and ignoring them on retrieval?

swood:
@Vad1mo The idea was to clean up the zero-sized cache in Redis to have a proper cache in there rather than rubbish.

src/server/middleware/blob/patch_blob_upload.go
```diff
@@ -39,7 +40,13 @@ func PatchBlobUploadMiddleware() func(http.Handler) http.Handler {
 
 		sessionID := distribution.ParseSessionID(r.URL.Path)
 
-		return blobController.SetAcceptedBlobSize(r.Context(), sessionID, size)
+		// Only cache non-zero sizes in Redis
+		if size > 0 {
```
Vad1mo: You are checking here for >0 zero and in SetAcceptedBlobSize. 
A change here isn't needed

src/server/middleware/blob/get_blob.go
Vad1mo: The more I think about it the more I am convinced we don't need that middleware at all..

swood:
The middle layer was required in the event of a zero-length Redis key. I am happy to change my approach. However, the current middle layer is functioning properly and, despite a substantial load, doesn't create any issues.

swood:
Also, I tried to drop the middle layer and I faced an issue with registry like:
> 1ba42558e44d: Pulling fs layer
> short read: expected 95 bytes but got 0: unexpected EOF

So the middlelayer is suppose to check if the blob key in Redis has a zero size and clean it pro-actively.

src/server/middleware/blob/get_blob.go
swood: Yes, we need.

The request is sent to the `core`, then it is forwarded to the registry. If the registry finds a zero-sized key, the pull process would fail. 
The purpose of the middle layer is to clean a key in Redis before the registry attempts to read it.

swood:
@Vad1mo any further comments here?

PR #22348, opened 2025-09-12
Add max_upstream_conn parameter for each proxy_cache project

### LINKED ISSUE #22184: Coalesce upstream requests where possible

stonezdj:
Limit the proxy connection to the upstream registry
  
  fixes #22184

# Issue being fixed
Fixes #(issue)

### DIFF
--- a/api/v2.0/swagger.yaml
+++ b/api/v2.0/swagger.yaml
@@ -7321,6 +7321,10 @@ definitions:
         type: string
         description: 'The bandwidth limit of proxy cache, in Kbps (kilobits per second). It limits the communication between Harbor and the upstream registry, not the client and the Harbor.'
         x-nullable: true
+      max_upstream_conn:
+        type: string
+        description: 'The max connection per artifact to the upstream registry in current proxy cache project, if it is -1, no limit to upstream registry connections'
+        x-nullable: true
   ProjectSummary:
     type: object
     properties:
--- a/src/pkg/project/models/pro_meta.go
+++ b/src/pkg/project/models/pro_meta.go
@@ -25,4 +25,5 @@ const (
 	ProMetaReuseSysCVEAllowlist     = "reuse_sys_cve_allowlist"
 	ProMetaAutoSBOMGen              = "auto_sbom_generation"
 	ProMetaProxySpeed               = "proxy_speed_kb"
+	ProMetaMaxUpstreamConn          = "max_upstream_conn"
 )
--- a/src/pkg/project/models/project.go
+++ b/src/pkg/project/models/project.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/goharbor/harbor/src/lib/log"
 	"github.com/goharbor/harbor/src/lib/orm"
 	allowlist "github.com/goharbor/harbor/src/pkg/allowlist/models"
 )
@@ -167,6 +168,20 @@ func (p *Project) ProxyCacheSpeed() int32 {
 		return 0
 	}
 	return int32(speedInt)
+}
+
+// MaxUpstreamConnection ...
+func (p *Project) MaxUpstreamConnection() int {
+	countVal, exist := p.GetMetadata(ProMetaMaxUpstreamConn)
+	if !exist {
+		return 0
+	}
+	cnt, err := strconv.ParseInt(countVal, 10, 32)
+	if err != nil {
+		log.Warningf("failed th parse the max_upstream_conn, val:%s error %v", countVal, err)
+		return 0
+	}
+	return int(cnt)
 }
 
 // FilterByPublic returns orm.QuerySeter with public filter
--- a/src/pkg/proxy/connection/limit.go
+++ b/src/pkg/proxy/connection/limit.go
@@ -1,0 +1,79 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package connection
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/go-redis/redis/v8"
+
+	"github.com/goharbor/harbor/src/lib/log"
+)
+
+// ConLimiter is used to limit the number of connections to the upstream service
+type ConnLimiter struct {
+}
+
+// Limiter is a global connection limiter instance
+var Limiter = &ConnLimiter{}
+
+// Used to compare and increase connection number in redis
+//
+// KEYS[1]: key of max_conn_upstream
+// ARGV[1]: max connection limit
+var increaseWithLimitText = `
+local current = tonumber(redis.call('GET', KEYS[1]) or '0')
+local max = tonumber(ARGV[1])
+
+if current + 1 <= max then
+    redis.call('INCRBY', KEYS[1], 1)
+	redis.call('EXPIRE', KEYS[1], 3600) -- set expire to avoid always lock
+    return 1
+else
+    return 0
+end
+`
+
+var acquireScript = redis.NewScript(increaseWithLimitText)
+
+// Acquire tries to acquire a connection, returns true if successful
+func (c *ConnLimiter) Acquire(ctx context.Context, rdb *redis.Client, key string, limit int) bool {
+	result, err := acquireScript.Run(ctx, rdb, []string{key}, fmt.Sprintf("%v", limit)).Int()
+	if err != nil {
+		log.Errorf("failed to get the connection lock in redis, error %v", err)
+		return false
+	}
+	log.Debugf("Acquire script result is %d", result)
+	return result == 1
+}
+
+var decreaseText = `
+local val = tonumber(redis.call("GET", KEYS[1]) or "0")
+if val > 0 then
+    redis.call("DECR", KEYS[1])
+end
+return 0
+`
+
+var decreaseScript = redis.NewScript(decreaseText)
+
+// Release releases a connection in redis
+func (c *ConnLimiter) Release(ctx context.Context, rdb *redis.Client, key string) {
+	_, err := decreaseScript.Run(ctx, rdb, []string{key}).Int()
+	if err != nil {
+		log.Infof("release connection failed:%v", err)
+	}
+}
--- a/src/server/middleware/repoproxy/proxy.go
+++ b/src/server/middleware/repoproxy/proxy.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"os"
 	"strings"
 	"time"
 
@@ -33,18 +34,21 @@ import (
 	httpLib "github.com/goharbor/harbor/src/lib/http"
 	"github.com/goharbor/harbor/src/lib/log"
 	"github.com/goharbor/harbor/src/lib/orm"
+	"github.com/goharbor/harbor/src/lib/redis"
 	proModels "github.com/goharbor/harbor/src/pkg/project/models"
+	"github.com/goharbor/harbor/src/pkg/proxy/connection"
 	"github.com/goharbor/harbor/src/pkg/reg/model"
 	"github.com/goharbor/harbor/src/server/middleware"
 )
 
 const (
-	contentLength       = "Content-Length"
-	contentType         = "Content-Type"
-	dockerContentDigest = "Docker-Content-Digest"
-	etag                = "Etag"
-	ensureTagInterval   = 10 * time.Second
-	ensureTagMaxRetry   = 60
+	contentLength                  = "Content-Length"
+	contentType                    = "Content-Type"
+	dockerContentDigest            = "Docker-Content-Digest"
+	etag                           = "Etag"
+	ensureTagInterval              = 10 * time.Second
+	ensureTagMaxRetry              = 60
+	upstreamRegistryLimitOnProject = "UPSTREAM_REGISTRY_LIMIT_ON_PROJECT" // if UPSTREAM_REGISTRY_LIMIT_ON_PROJECT is true, the upstream registry connection is based on project level, by default it is artifact level
 )
 
 var tooManyRequestsError = errors.New("too many requests to upstream registry").WithCode(errors.RateLimitCode)
@@ -99,6 +103,22 @@ func handleBlob(w http.ResponseWriter, r *http.Request, next http.Handler) error
 		next.ServeHTTP(w, r)
 		return nil
 	}
+
+	if p.MaxUpstreamConnection() > 0 {
+		client, err := redis.GetHarborClient()
+		if err != nil {
+			return errors.NewErrs(err)
+		}
+		key := upstreamRegistryConnectionKey(art)
+		log.Debugf("handle blob, upstream registry connection limit key: %s", key)
+		if !connection.Limiter.Acquire(ctx, client, key, p.MaxUpstreamConnection()) {
+			log.Infof("current connection exceed max connections to upstream registry")
+			// send http code 429 to client
+			return tooManyRequestsError
+		}
+		defer connection.Limiter.Release(context.Background(), client, key) // use background context in defer to avoid been canceled
+	}
+
 	size, reader, err := proxyCtl.ProxyBlob(ctx, p, art)
 	if err != nil {
 		return err
@@ -173,6 +193,15 @@ func defaultBlobURL(projectName string, name string, digest string) string {
 	return fmt.Sprintf("/v2/%s/library/%s/blobs/%s", projectName, name, digest)
 }
 
+// upstreamRegistryConnectionKey get upstream registry connection key
+func upstreamRegistryConnectionKey(art lib.ArtifactInfo) string {
+	limitOnProject := os.Getenv(upstreamRegistryLimitOnProject)
+	if strings.EqualFold("true", limitOnProject) {
+		return fmt.Sprintf("{upstream_registry_connection}:%s", art.ProjectName)
+	}
+	return fmt.Sprintf("{upstream_registry_connection}:%s:%s", art.Repository, art.Digest)
+}
+
 func handleManifest(w http.ResponseWriter, r *http.Request, next http.Handler) error {
 	ctx := r.Context()
 	art, p, proxyCtl, err := preCheck(ctx, true)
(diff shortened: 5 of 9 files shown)


api/v2.0/swagger.yaml
```diff
@@ -7321,6 +7321,10 @@ definitions:
         type: string
         description: 'The bandwidth limit of proxy cache, in Kbps (kilobits per second). It limits the communication between Harbor and the upstream registry, not the client and the Harbor.'
         x-nullable: true
+      max_upstream_conn:
+        type: string
+        description: 'The max connection to the upstream registry in current proxy cache project, if it is -1, no limit to upstream registry connections'
```
reasonerjt: 1. This attribute is nullable, so if it's null, it should also mean no limit?  Maybe we don't need to use "-1" as a special number here.
2. I believe the limit should be set to limit concurrent connections to fetch the same artifact?

stonezdj: 1. keep it the same as bandwidth setting, -1 means no limitation.
2. change to per artifact.

Vad1mo:
@Strainy does this replace your PR https://github.com/goharbor/harbor/pull/22185?

Strainy:
> @Strainy does this replace your PR #22185?

I've had a quick scan through, and it seems like the client will experience 429s when the maximum connection limit it reached? This is different to my change (but perhaps complementary?), requests for the same manifests/blobs will queue behind a single request and return the result from the cache when it completes (i.e., no requests failures when there is contention for the same resources).

stonezdj:
> > @Strainy does this replace your PR #22185?
> 
> I've had a quick scan through, and it seems like the client will experience 429s when the maximum connection limit it reached? This is different to my change (but perhaps complementary?), requests for the same manifests/blobs will queue behind a single request and return the result from the cache when it completes (i.e., no requests failures when there is contention for the same resources).

Queue all subsequent requests will consume all tcp connection on the Harbor server, if there are 500 request, then only 1 request can proceed, other 499 will wait, if the first takes 30 minutes, the other should wait 30 minutes. and meanwhile other connection might fail because TCP connection is a limited resource. return 429 is better than blocking the request.

Strainy:
> > > @Strainy does this replace your PR #22185?
> 
> > 
> 
> > I've had a quick scan through, and it seems like the client will experience 429s when the maximum connection limit it reached? This is different to my change (but perhaps complementary?), requests for the same manifests/blobs will queue behind a single request and return the result from the cache when it completes (i.e., no requests failures when there is contention for the same resources).
> 
> 
> 
> Queue all subsequent requests will consume all tcp connection on the Harbor server, if there are 500 request, then only 1 request can proceed, other 499 will wait, if the first takes 30 minutes, the other should wait 30 minutes. and meanwhile other connection might fail because TCP connection is a limited resource. return 429 is better than blocking the request.
> 
> 

I am in full agreement that returning a 429 is better than blocking the request. Which is why I feel this change is complementary with mine. This change doesn't address the upstream request deduplication issue that I was originally trying to solve on my branch.

Strainy:
To add a bit more context: I have a use case where we are very sensitive to boot times, particularly for new images that need to be pulled from upstream. We also have a very large number of pods often concurrently hitting the registry for the same artifact. 

So I would just like to make sure we're efficient about retrieving resources from the upstream. Which goes further than just rate limiting imo, we should aim to de-duplicate concurrent requests where possible - which is the motivation of my change.

This approach has been working well for us. We're using our fork of Harbor as a DC-local cache for images in Google Artifact Registry. We routinely handle spikes of >1K concurrent pulls with this approach.

But I think if we were to just use this rate limiting approach, we'd have quite a lot of pods just spinning in CrashLoopBackOff... and that'd be a bad time for us.

src/pkg/proxy/connection/limit.go
```diff
@@ -0,0 +1,78 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/go-redis/redis/v8"
+	"github.com/goharbor/harbor/src/lib/log"
+)
+
+// ConnectionLimiter is used to limit the number of connections to the upstream service
+type ConnectionLimiter struct {
```
chlins: I'm not sure can we use the `https://github.com/go-redis/redis_rate` as the limiter directly?

stonezdj:
> To add a bit more context: I have a use case where we are very sensitive to boot times, particularly for new images that need to be pulled from upstream. We also have a very large number of pods often concurrently hitting the registry for the same artifact.
> 
> So I would just like to make sure we're efficient about retrieving resources from the upstream. Which goes further than just rate limiting imo, we should aim to de-duplicate concurrent requests where possible - which is the motivation of my change.
> 
> This approach has been working well for us. We're using our fork of Harbor as a DC-local cache for images in Google Artifact Registry. We routinely handle spikes of >1K concurrent pulls with this approach.
> 
> But I think if we were to just use this rate limiting approach, we'd have quite a lot of pods just spinning in CrashLoopBackOff... and that'd be a bad time for us.

the event is ImagePullBackOff when the image pulling failed, and it will retry pulling image with a back off policy.
I prefer return 429 literaly rather than hang on the connection on server side. if there are 500 requests come to pull the same image at the same time, only 1 is served, but it may take longer time than expected to get the cache ready time, during this interval, other 499 connection is hanging on, because the max connection to a server is a limit resource, it is possible that other client cannot connect to Harbor without any response. it is not a good user experience.

src/pkg/proxy/connection/limit.go
stonezdj: The library of redis_rate is based on events occur at an interval, and this method is suitable for short-term requests.  image pulling sometimes is a long-term requests.
for example:
```
If set 1 request per second.
When request 1 comes, it last 30 seconds to complete.
request 2 comes at 2 second, then the limitation refreshed for the second time, then request 2 is served.
request 3 comes at 3 second, and request 3 is served.
```
So the total concurrent pull request is 3, not 1.

src/pkg/project/models/pro_meta.go
```diff
@@ -25,4 +25,5 @@ const (
 	ProMetaReuseSysCVEAllowlist     = "reuse_sys_cve_allowlist"
 	ProMetaAutoSBOMGen              = "auto_sbom_generation"
 	ProMetaProxySpeed               = "proxy_speed_kb"
+	ProMetaMaxUpstreamConn          = "max_upstream_conn"
```
wy65701436, line 28: general comments, let's add the UT?

src/server/middleware/repoproxy/proxy.go
```diff
@@ -33,18 +34,21 @@ import (
...
-	etag                = "Etag"
-	ensureTagInterval   = 10 * time.Second
-	ensureTagMaxRetry   = 60
+	contentLength              = "Content-Length"
+	contentType                = "Content-Type"
+	dockerContentDigest        = "Docker-Content-Digest"
+	etag                       = "Etag"
+	ensureTagInterval          = 10 * time.Second
+	ensureTagMaxRetry          = 60
+	upstreamRegistryLimitLevel = "UPSTREAM_REGISTRY_LIMIT_LEVEL" // if UPSTREAM_REGISTRY_LIMIT_LEVEL is project, the upstream registry connection is based on project level, by default it is artifact level
```
wy65701436, line 51: it's for the artifact level? can you revise the comments?

src/server/middleware/repoproxy/proxy.go
```diff
@@ -99,6 +103,22 @@ func handleBlob(w http.ResponseWriter, r *http.Request, next http.Handler) error
...
 		return nil
 	}
+
+	if p.MaxUpstreamConnection() > 0 {
+		client, err := redis.GetHarborClient()
+		if err != nil {
+			return errors.NewErrs(err)
+		}
+		key := upstreamRegistryConnectionKey(art)
+		log.Debugf("handle blob key: %s", key)
```
wy65701436: please gives a more concise comments?

src/server/middleware/repoproxy/proxy.go
```diff
@@ -219,6 +248,20 @@ func handleManifest(w http.ResponseWriter, r *http.Request, next http.Handler) e
...
+			return errors.NewErrs(err)
+		}
+		key := upstreamRegistryConnectionKey(art)
+		log.Debugf("handle manifest key %v", key)
+		if !connection.Limiter.Acquire(ctx, client, key, p.MaxUpstreamConnection()) {
+			log.Infof("current connection exceed max connections to upstream registry")
+			// send http code 429 to client
+			return tooManyRequestsError
+		}
+		defer connection.Limiter.Release(context.Background(), client, key) // use background context in defer to avoid been canceled
```
wy65701436, line 263: there is a defer already, we can merge them together?

src/server/middleware/repoproxy/proxy.go
```diff
@@ -99,6 +103,22 @@ func handleBlob(w http.ResponseWriter, r *http.Request, next http.Handler) error
...
+			return errors.NewErrs(err)
+		}
+		key := upstreamRegistryConnectionKey(art)
+		log.Debugf("handle blob key: %s", key)
+		if !connection.Limiter.Acquire(ctx, client, key, p.MaxUpstreamConnection()) {
+			log.Infof("current connection exceed max connections to upstream registry")
+			// send http code 429 to client
+			return tooManyRequestsError
+		}
+		defer connection.Limiter.Release(context.Background(), client, key) // use background context in defer to avoid been canceled
```
wy65701436, line 119: If the harbor-core crashes without calling `Release`, the counter stays incremented.

src/pkg/proxy/connection/limit.go
```diff
@@ -0,0 +1,78 @@
...
+//
+// KEYS[1]: key of max_conn_upstream
+// ARGV[1]: max connection limit
+var increaseWithLimitText = `
+local current = tonumber(redis.call('GET', KEYS[1]) or '0')
+local max = tonumber(ARGV[1])
+
+if current + 1 <= max then
+    redis.call('INCRBY', KEYS[1], 1)
+	redis.call('EXPIRE', KEYS[1], 3600) -- set expire to avoid always lock
```
wy65701436, line 43: please give a variable for ttl, like: 
```

type ConnectionLimiter struct {
	ttlSeconds int
} 
```

src/pkg/proxy/connection/limit.go
```diff
@@ -0,0 +1,78 @@
...
+local val = tonumber(redis.call("GET", KEYS[1]) or "0")
+if val > 0 then
+    redis.call("DECR", KEYS[1])
+end
+return 0
+`
+
+var decreaseScript = redis.NewScript(decreaseText)
+
+// Release releases a connection in redis
```
wy65701436, line 73: we can return the value for debugging purpose.
```

func (c *ConnectionLimiter) Release(ctx context.Context, rdb *redis.Client, artifactDigest string) int {
	key := fmt.Sprintf("artifact_limit:%s", artifactDigest)

	value, err := decreaseScript.Run(ctx, rdb, []string{key}).Int()
	if err != nil {
		log.Infof("release connection failed:%v", err)
	}
    return -1

	log.Debugf("released slot for %s, current usage=%d", key, val)
	return value
}
```

src/server/middleware/repoproxy/proxy.go
```diff
@@ -173,6 +193,15 @@ func defaultBlobURL(projectName string, name string, digest string) string {
 	return fmt.Sprintf("/v2/%s/library/%s/blobs/%s", projectName, name, digest)
 }
 
+// upstreamRegistryConnectionKey get upstream registry connection key
+func upstreamRegistryConnectionKey(art lib.ArtifactInfo) string {
+	level := os.Getenv(upstreamRegistryLimitLevel)
+	if strings.EqualFold("project", level) {
```
wy65701436: Please ensure that the condition only allows upstreamRegistryLimitLevel to be set to either project or artifact.

src/server/middleware/repoproxy/proxy.go
wy65701436: I have a general question, 

1. When the environment is set to project, does this mean that all /v2 requests within this project will be counted?
2. When the environment is set to artifact, does this mean that all /v2 requests for this artifact will be counted?

So, I think we need a concise describe on this env of the meaning.

stonezdj: Yes. 
1. all GET/HEAD request to manifest/blobs to the same project is counted.
2. all GET/HEAD request to manifest/blobs to the same artifact is counted.

src/server/middleware/repoproxy/proxy.go
stonezdj: If the harbor-core crashed, then the previous lock is still in the redis, and it will released after the key expires.

src/server/middleware/repoproxy/proxy.go
stonezdj: They have different meanings and conditions, we have to split them.

api/v2.0/swagger.yaml
```diff
@@ -7321,6 +7321,10 @@ definitions:
         type: string
         description: 'The bandwidth limit of proxy cache, in Kbps (kilobits per second). It limits the communication between Harbor and the upstream registry, not the client and the Harbor.'
         x-nullable: true
+      max_upstream_conn:
```
wy65701436, line 7324: can it be updated after the proxy cache project is created?

src/server/middleware/repoproxy/proxy.go
wy65701436: UPSTREAM_REGISTRY_LIMIT_ON_PROJECT_LEVEL=true?

api/v2.0/swagger.yaml
stonezdj: No, setting on proxy cache creation is read-only

src/server/middleware/repoproxy/proxy.go
stonezdj: Updated

src/pkg/proxy/connection/limit.go
stonezdj: we don't return the usage value in the lua, if need to get the value, we need to check it in redis.

src/server/middleware/repoproxy/proxy.go
stonezdj: updated

src/server/middleware/repoproxy/proxy.go
stonezdj: updated

src/pkg/project/models/project.go
```diff
@@ -169,6 +169,19 @@ func (p *Project) ProxyCacheSpeed() int32 {
...
 
+// MaxUpstreamConnection ...
+func (p *Project) MaxUpstreamConnection() int {
+	countVal, exist := p.GetMetadata(ProMetaMaxUpstreamConn)
+	if !exist {
+		return 0
+	}
+	cnt, err := strconv.ParseInt(countVal, 10, 32)
+	if err != nil {
+		return 0
```
reasonerjt, line 182: nit:  log the error as a warning?

src/pkg/proxy/connection/limit.go
```diff
@@ -0,0 +1,79 @@
...
+end
+`
+
+var acquireScript = redis.NewScript(increaseWithLimitText)
+
+// Acquire tries to acquire a connection, returns true if successful
+func (c *ConnLimiter) Acquire(ctx context.Context, rdb *redis.Client, key string, limit int) bool {
+	result, err := acquireScript.Run(ctx, rdb, []string{key}, fmt.Sprintf("%v", limit)).Int()
+	if err != nil {
+		log.Errorf("failed to get the connection number in redis, error %v", err)
```
reasonerjt: wrong log message?

reasonerjt:
Could you make Limiter an interface and add some testcases?

src/pkg/project/models/project.go
stonezdj: Updated

src/pkg/proxy/connection/limit.go
stonezdj: updated

Issue #22257, opened 2025-08-14
Disable writing to audit log writing to DB by default


reasonerjt:
Since v2.13.0, the audit log feature has been extended to support more events.  There are concerns that when Harbor is heavily used, too many audit logs may exhaust the database.  

We should consider provide an option to disable writing the audit log to database, such that downstream adopter can choose to turn it off when Harbor is installed.

TBD:  Shall we provide flexibility to the end user (system admin) to re-enable writing the audit logs to DB via UI/API?

Vad1mo:
Your proposal is not clear to me: there is already an option to disable audit log to DB.

<img width="883" height="856" alt="Image" src="https://github.com/user-attachments/assets/0c7a279f-ced0-406c-8b10-c1786333b029" />

reasonerjt:
@Vad1mo Thanks.
I'll double check whether we can achieve this by setting a `rest-config-json` env var while harbor is deployed. 

Let's keep this open until it's clarified.

Vad1mo:
I have been investigating and testing on hwo the auditlogs-ext, can be improved especially the inserts.

this is what I found: 

### Very inefficient purge operation 

The condition `lower(operation || '_' || resource_type)` is terrible for performance because it prevents the database from using standard indexes on operation or resource_type. It must perform a full scan of all old rows and compute this combined, lowercased string for every single one before checking if it's in the list.

https://github.com/goharbor/harbor/blob/63b61d6995a7cc3ae3141ccd28837e9c8ab5fea0/src/pkg/auditext/dao/dao.go#L142

The operation can become an instant delete operation with partitions (see below).

### Remove Not Needed Indexes and idex differenlty 
- idx_audit_log_ext_project_id_operation
- idx_audit_log_ext_project_id_resource_type
Those indexes are not used frequently (almost never), and with the following changes, they aren't needed at all

Change indexes to DESC instead of ASC as this is the usual select.
- project_id, op_time DESC
- op_time DESC

### Use Postgres Partitions

This is the most common and effective solution for time-series data like audit logs. Split large logical table into many smaller physical tables (partitions), typically by a time range.

- Isolates Hot Data: All new INSERTs go into the latest, smallest partition (e.g., weekly partition). Indexing a small table is much faster than indexing a massive, billion-row table.

- Faster Maintenance: VACUUM and other maintenance operations run much faster on a small, active partition instead of the entire table.

[...]

PR #22350, opened 2025-09-12
Allow disable audit log to DB when initialize

MinerYang:

# Issue being fixed
Fixes #([22257](https://github.com/goharbor/harbor/issues/22257))


src/core/main.go
```diff
@@ -356,3 +363,32 @@ func getDefaultScannerName() string {
 	}
 	return ""
 }
+
+func initSkipAuditDBbyEnv(ctx context.Context) error {
```
wy65701436, line 367: Should the sequence be: first check whether an update is needed, and then retrieve the value only if the condition is true?

src/core/main.go
```diff
@@ -356,3 +363,32 @@ func getDefaultScannerName() string {
...
+		skipAuditEnv, err = strconv.ParseBool(s)
+		if err != nil {
+			log.Warningf("Failed to parse SKIP_LOG_AUDIT_DATABASE to bool with error: %v, Will use SKIP_LOG_AUDIT_DATABASE env as false", err)
+		}
+	}
+	log.Debugf("get SKIP_LOG_AUDIT_DATABASE from Env is %v", skipAuditEnv)
+
+	// get from db
+	mgr := config.GetCfgManager(ctx)
+	cfg, err := mgr.GetItemFromDriver(ctx, common.SkipAuditLogDatabase)
```
wy65701436, line 381: It can use the mgr.Get(), if the value is not set, this method will return a [pre-defined](https://github.com/goharbor/harbor/blob/main/src/lib/config/metadata/value.go#L32) error.

MinerYang: mgr.Get() get the k-v from the store map instead of the db driver directly. And it will always has a default values once cfgValues loaded

src/core/main.go
```diff
@@ -356,3 +363,32 @@ func getDefaultScannerName() string {
 	}
 	return ""
 }
+
+func initSkipAuditDBbyEnv(ctx context.Context) error {
+	var err error
+	skipAuditEnv := false
+	s := os.Getenv("SKIP_LOG_AUDIT_DATABASE")
```
stonezdj, line 370: suggest to add _INIT suffix so that it can be self explained.

src/core/main.go
stonezdj: You should implement the function to check if k = common.SkipAuditLogDatabase exists in the properties table. not GetItermFromDriver()

MinerYang: For sure I could  implement  either check the key existence in the db or like now return k-v from db.
However I think get a k-v from db can be more generically used in the future compare to return the key existence bool value solely.

MinerYang:
Hi @wy65701436 ,

Due to personal time break, please help to keep working on this issue. Appreciate!

Best,
Miner

wy65701436:
use this [MR](https://github.com/goharbor/harbor/pull/22452) to cover.

PR #22451, opened 2025-10-11
Deploy with podman and podman-compose

### LINKED ISSUE #15105: Support Podman
Today Harbor requires Docker and `docker-compose` for installations and upgrades.

https://github.com/goharbor/harbor/blob/master/make/common.sh#L78-L130
https://github.com/goharbor/harbor/blob/master/README.md#install--run

Red Hat has deprecated and removed Docker since RHEL8 :-(
https://access.redhat.com/solutions/3696691

While Docker (the company) is still shipping CE and EE binaries for RHEL, the lack of official support from Red Hat may be a source of conflict.

This proposal is to support also Podman as a container Engine. This is not only in favor of Red Hat but more generally of the added security features of Podman. 
http://people.redhat.com/abach/OSAW/FILES/DAY1/5%20Moving%20on%20from%20Docker.pdf

Since Podman 3.0, `docker-compose` is fully supported
https://www.redhat.com/sysadmin/podman-docker-compose

Perhaps the logic at `check_docker` in `common.sh` could be reworked to also support Podman

Podman supports `docker` alias but using Podman CLI would be best.
The idea would be to first rework the logic in `check_docker` verifying whether Docker is present and meet the minimum version and same story to Podman.

Assuming the proposal is acceptable, I can work on a prototype solution and have something available in a few weeks (the logic isn't that hard but perhaps there are all sort of documentation references to update)

deamen:
This PR add support for deploying harbor with podman, podman-docker and podman-compose on rhel/centos/rockylinux/almalinux/fedora.

Added `--with-podman` install option:

1. The docker/docker-compose checks have been extended to support podman and podman compose
2. The logging driver is set to `json-file`, configurations are read from harbor.yml, so the log location/names/rotation is the same as using docker 
3. The log container is removed from the rendered `docker-compose.yml` as podman does not support it
4. The `tests/ci/podman_install.sh` is created for testing on redhat distro

# Issue being fixed
Fixes #15105

### DIFF
--- a/make/install.sh
+++ b/make/install.sh
@@ -9,6 +9,7 @@ set +o noglob
 
 usage=$'Please set hostname and other necessary attributes in harbor.yml first. DO NOT use localhost or 127.0.0.1 for hostname, because Harbor needs to be accessed by external clients.
 Please set --with-trivy if needs enable Trivy in Harbor.
+Please set --with-podman if you want to force prepare to treat the Docker CLI as Podman.
 Please do NOT set --with-chartmuseum, as chartmusuem has been deprecated and removed.
 Please do NOT set --with-notary, as notary has been deprecated and removed.'
 item=0
@@ -17,6 +18,8 @@ item=0
 with_clair=$false
 # trivy is not enabled by default
 with_trivy=$false
+# podman option not enabled by default
+with_podman=$false
 
 # flag to using docker compose v1 or v2, default would using v1 docker-compose
 DOCKER_COMPOSE=docker-compose
@@ -28,6 +31,8 @@ while [ $# -gt 0 ]; do
             exit 0;;
             --with-trivy)
             with_trivy=true;;
+            --with-podman)
+            with_podman=true;;
             *)
             note "$usage"
             exit 1;;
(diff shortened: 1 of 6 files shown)


make/install.sh
```diff
@@ -9,6 +9,7 @@ set +o noglob
 
 usage=$'Please set hostname and other necessary attributes in harbor.yml first. DO NOT use localhost or 127.0.0.1 for hostname, because Harbor needs to be accessed by external clients.
 Please set --with-trivy if needs enable Trivy in Harbor.
+Please set --with-podman if you want to force prepare to treat the Docker CLI as Podman.
```
reasonerjt, line 12: I don't think this is the right pattern, b/c with the param like  "--with-xxx", the "xxx" is always a component in Harbor.  Like "--with-trivy"

tests/ci/podman_install.sh
```diff
@@ -0,0 +1,122 @@
+#!/bin/bash
```
reasonerjt, line 1: Is this really triggered in CI when you add this script?

reasonerjt (changes requested):
Blocking this from being accidentally merged.  Because there are some disagreement, and this has to be verified in CI.

We also need to take a closer look, please give us a bit more time. 
Thanks!

make/install.sh
deamen: Yeah, I wasn't sure about this one when I started the PR. What should we use here? Will `--podman` be fine?

tests/ci/podman_install.sh
deamen: No, it is not triggered by CI yet, this is something I need guidance.
I tried to write a CI test with the ubuntu-latest runner but the podman version in ubuntu seems to be too old, it won't start up the containers.

This podman_install.sh will only work on  RHEL/centos/rocky/almalinux/fedora, if we want to have CI test for deploying harbor with podman and podman-compose, we will need a red hat runner, or wait for Ubuntu 25.04.

Which direction should we go?

PR #22487, opened 2025-10-22
add migraiton sql for skip_audit_log_database

wy65701436:
give the value of skip_audit_log_database bases on the real usage.

1. If audit logs exist in the system (either in audit_log or audit_log_ext tables), set skip_audit_log_database to false
2. If no audit logs exist but the tables show evidence of previous usage set skip_audit_log_database to false

# Issue being fixed
Fixes #(issue)

### DIFF
--- a/make/migrations/postgresql/0180_2.15.0_schema.up.sql
+++ b/make/migrations/postgresql/0180_2.15.0_schema.up.sql
@@ -1,0 +1,48 @@
+/*
+Initialize skip_audit_log_database configuration based on existing audit log usage - Only insert the configuration if it doesn't already exist
+1. If audit logs exist in the system (either in audit_log or audit_log_ext tables),
+   set skip_audit_log_database to false
+2. If no audit logs exist but the tables show evidence of previous usage
+   set skip_audit_log_database to false
+3. If tables exist but show no evidence of usage, don't create the configuration record
+
+*/
+DO $$
+DECLARE
+    audit_log_count INTEGER := 0;
+    audit_log_ext_count INTEGER := 0;
+    total_audit_logs INTEGER := 0;
+    config_exists INTEGER := 0;
+    audit_log_seq_value BIGINT := 0;
+    audit_log_ext_seq_value BIGINT := 0;
+    audit_log_table_used BOOLEAN := FALSE;
+    audit_log_ext_table_used BOOLEAN := FALSE;
+    should_set_config BOOLEAN := FALSE;
+    skip_audit_value TEXT := 'false';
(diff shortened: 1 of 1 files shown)


Vad1mo:
Why is this magic happening in the DB schema migration script and not simply via config flags?

When do you think this will be set? All existing users have audit logs.
Only new deployments have no auditlogs

MinerYang (approved):
lgtm

make/migrations/postgresql/0180_2.15.0_schema.up.sql
```diff
@@ -0,0 +1,48 @@
...
+    audit_log_ext_table_used BOOLEAN := FALSE;
+    should_set_config BOOLEAN := FALSE;
+    skip_audit_value TEXT := 'false';
+BEGIN
+    SELECT COUNT(*) INTO config_exists
+    FROM properties
+    WHERE k = 'skip_audit_log_database';
+
+    IF config_exists = 0 THEN
+        SELECT COUNT(*) INTO audit_log_count FROM audit_log;
```
stonezdj: Suggest to use one sql like that, then only one variable should_set_config is used.
```
SELECT
    (EXISTS (SELECT 1 FROM audit_log)
     OR EXISTS (SELECT 1 FROM audit_log_ext)
     OR (SELECT last_value FROM audit_log_id_seq) > 1
     OR (SELECT last_value FROM audit_log_ext_id_seq) > 1
    ) AS should_config INTO should_set_config ;
```

wy65701436:
> Why is this magic happening in the DB schema migration script and not simply via config flags?
> 
> When do you think this will be set? All existing users have audit logs. Only new deployments have no auditlogs



> Why is this magic happening in the DB schema migration script and not simply via config flags?
> 
> When do you think this will be set? All existing users have audit logs. Only new deployments have no auditlogs

This PR sets a value for skip_audit_log_database to handle the scenario where the database is used as the audit log storage.

Previously, we introduced a new env to control whether to skip using the database for audit logs. However, this change could break existing setups when upgrading from older Harbor versions that already use the database.

This PR is the second step in that change — ensuring the new environment variable only takes effect for fresh installations. During upgrades, if the existing setup uses the database for audit logs, give a value to index it and then behavior will be preserved.

make/migrations/postgresql/0180_2.15.0_schema.up.sql
wy65701436: from the debugging perspective, it's not a good practice the combine all the sqls into one. keep the current imple and close it.

Issue #12579, opened 2020-07-26
The data type of the vulnerability severity value is changed to be string after policy modification but it's expected as integer


steven-zou:
![Screen Shot 2020-07-26 at 11 36 49](https://user-images.githubusercontent.com/5753287/88470853-533d5980-cf34-11ea-8edd-0c1103160b8e.png)

PR #22510, opened 2025-10-27
Remove legacy switch from decodeFilters after UI bug #12579 fix

intojhanurag:
### Description
Removed the old type conversion switch in `decodeFilters` that handled inconsistent
filter value types due to UI bug #12579. The bug is now fixed, so manual conversion
is no longer needed.

### Changes
- Removed type-conversion switch
- Removed unused `strconv` import


intojhanurag:
Hey @OrlinVasilev , It is ready for review.

OrlinVasilev:
@bupd @AllForNothing @chlins take a look please!

PR #22634, opened 2025-12-10
fix: Add distinct to query that is causing high CPU on retention jobs

### LINKED ISSUE #22319: High CPU usage on Postgres during Tag Retention (v2.13.1)
**Harbor Version:** v2.13.1
**Database:** Postgres (AWS RDS, 4 vCPU - db.m7g.xlarge)

#### Summary:
When running Tag Retention, our Postgres CPU usage spikes to 90–100% until the job finishes. Normally the DB runs at ~4% CPU, so keeping a large RDS instance only for these jobs is overprovisioning IMHO.

#### Example repository:

- ~3,403 artifacts
- ~138 GB size

Retention run results looks like:
```
Retained/Total
199/424
199/420
196/413
197/421
205/430
205/432
205/432
205/431
```
But during this, Postgres is fully saturated. Scaling from 2 → 4 vCPUs helped only slightly.

#### Culprit SQL.

From db insights we can see the following SQL query during retention run:
```
SELECT b.digest_blob
FROM artifact a, artifact_blob b
WHERE a.digest = b.digest_af
  AND a.project_id = $1
  AND b.digest_blob IN ($2,$3,$4,$5,$6,$7,$8,$9,$10,$11,$12);
```
This query shows up repeatedly and drives CPU usage.

Image
Image

#### References

- Related issue: https://github.com/goharbor/harbor/issues/14708
- Related PR: https://github.com/goharbor/harbor/pull/17296
- And this one: https://github.com/goharbor/harbor/issues/17653

Even with these changes already merged, in v2.13.1 we still observe the same CPU pressure.

#### Questions

- Do the you have any thoughts or recommendations to mitigate this?
- Are there indexes or config tweaks that could help with this specific query pattern?
- Any planned optimizations for retention/GC to avoid repeated heavy scans?

bbytheway-rq:
fixes #22319

Adds a distinct keyword to ensure the query get only the unique digests, not repeated rows for every time the blob is used by multiple artifacts.

### DIFF
--- a/src/pkg/blob/dao/dao.go
+++ b/src/pkg/blob/dao/dao.go
@@ -246,7 +246,7 @@ func (d *dao) FindBlobsShouldUnassociatedWithProject(ctx context.Context, projec
 		return nil, err
 	}
 
-	sql := `SELECT b.digest_blob FROM artifact a, artifact_blob b WHERE a.digest = b.digest_af AND a.project_id = ? AND b.digest_blob IN (%s)`
+	sql := `SELECT DISTINCT b.digest_blob FROM artifact a, artifact_blob b WHERE a.digest = b.digest_af AND a.project_id = ? AND b.digest_blob IN (%s)`
 	params := []any{projectID}
 	for _, blob := range blobs {
 		params = append(params, blob.Digest)


chlins:
@bbytheway-rq Thanks for your investigation, based on https://github.com/goharbor/harbor/issues/22319#issuecomment-3638759498, I didn’t observe a significant improvement when `distinct` was added. Am I misunderstanding something?

PR #22637, opened 2025-12-11
Fix for loop in 0120_2.9.0_schema.up.sql that never populated fixable_cnt

tkatkov:
include fixable count into the scan_report update

### DIFF
--- a/make/migrations/postgresql/0120_2.9.0_schema.up.sql
+++ b/make/migrations/postgresql/0120_2.9.0_schema.up.sql
@@ -68,7 +68,6 @@ $$
                 none_count := 0;
                 low_count := 0;
                 unknown_count := 0;
-                fixable_count := 0;
                 FOR v IN SELECT vr.severity, vr.fixed_version
                          FROM report_vulnerability_record rvr,
                               vulnerability_record vr
@@ -87,8 +86,7 @@ $$
                             none_count = none_count + 1;
                         ELSIF v.severity = 'Unknown' THEN
                             unknown_count = unknown_count + 1;
-                        END IF;
-                        IF v.fixed_version IS NOT NULL THEN
+                        ELSIF v.fixed_version IS NOT NULL THEN
                             fixable_count = fixable_count + 1;
                         END IF;
                     END LOOP;
@@ -97,8 +95,7 @@ $$
                     high_cnt     = high_count,
                     medium_cnt   = medium_count,
                     low_cnt      = low_count,
-                    unknown_cnt  = unknown_count,
-                    fixable_cnt  = fixable_count
+                    unknown_cnt  = unknown_count
                 WHERE uuid = report.uuid;
             END LOOP;
     END
--- a/make/migrations/postgresql/0120_2.9.0_schema.up.sql
+++ b/make/migrations/postgresql/0120_2.9.0_schema.up.sql
@@ -68,6 +68,7 @@ $$
                 none_count := 0;
                 low_count := 0;
                 unknown_count := 0;
+                fixable_count := 0;
                 FOR v IN SELECT vr.severity, vr.fixed_version
                          FROM report_vulnerability_record rvr,
                               vulnerability_record vr
@@ -86,7 +87,8 @@ $$
                             none_count = none_count + 1;
                         ELSIF v.severity = 'Unknown' THEN
                             unknown_count = unknown_count + 1;
-                        ELSIF v.fixed_version IS NOT NULL THEN
+                        END IF;
+                        IF v.fixed_version IS NOT NULL THEN
                             fixable_count = fixable_count + 1;
                         END IF;
                     END LOOP;
@@ -95,7 +97,8 @@ $$
                     high_cnt     = high_count,
                     medium_cnt   = medium_count,
                     low_cnt      = low_count,
-                    unknown_cnt  = unknown_count
+                    unknown_cnt  = unknown_count,
+                    fixable_cnt  = fixable_count
                 WHERE uuid = report.uuid;
             END LOOP;
     END
--- a/make/migrations/postgresql/0180_2.15.0_schema.up.sql
+++ b/make/migrations/postgresql/0180_2.15.0_schema.up.sql
@@ -15,3 +15,31 @@ BEGIN
         INSERT INTO properties (k, v) VALUES ('skip_audit_log_database', 'false');
     END IF;
 END $$;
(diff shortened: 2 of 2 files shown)


Vad1mo:
Thank you for the PR, this look solid: 

1. Please reference an issue
2. **The Problem with Your Approach**                                                                                                                                                                                                                                                                                                           

                                                                                                                                                                                                                                                                                                                                           
  You modified 0120_2.9.0_schema.up.sql, but:                                                                                                                                                                                                                                                                                              
  - Migrations only run once per database                                                                                                                                                                                                                                                                                                  
  - Users who already upgraded past 2.9.0 have the buggy migration applied                                                                                                                                                                                                                                                                 
  - They won't re-run it, so their fixable_cnt column will remain NULL forever                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                           
  **The Correct Approach**                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                           
  Create a new migration file fixes it for existing installations.                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                           
  Naming convention: {SEQUENCE}_{VERSION}_schema.up.sql                                                                                                                                                                                                                                                                                    
  - Current latest: 0170_2.14.0_schema.up.sql                                                                                                                                                                                                                                                                                              
  - Your fix should be in  0171_2.15.0_schema.up.sql (or next available sequence number)

tkatkov:
@Vad1mo yes, but say me, I am in process upgrading from 2.7 and noticed this problem. Not always people upgrade to the latest and greatest, so probably best to fix it here and I will also create new PR as latest upgrade which would update only empty fixed counts.

Vad1mo:
Thanks for identifying this bug! The fixable_cnt column was indeed never populated due to these issues:
  1. fixable_count variable not initialized
  2. Fixable counting in ELSIF branch (only counted when severity didn't match)
  3. fixable_cnt missing from UPDATE statement

##  Issue with Your Initial Approach

Modifying 0120_2.9.0_schema.up.sql won't fix the issue for most users. (only <2.9)

Harbor uses https://github.com/golang-migrate/migrate which tracks applied migrations in the schema_migrations table. Once a migration runs, it's marked as complete and never re-runs - even if the file content changes.

  | User Scenario          | Effect of This PR                 |
  |------------------------|-----------------------------------|
  | Fresh install          | ✅ Gets the fix                   |
  | Upgrade from <2.9.0 | ✅ Gets the fix                   |
  | Already on 2.9.0+      | ❌ No fix - migration already ran |

  Most existing Harbor installations are already past v2.9.0, so this PR would not help them.

## Additional Bug Found

 The condition v.fixed_version IS NOT NULL is insufficient. The fixed_version column stores empty strings ('') for non-fixable vulnerabilities, not NULL:

```sql
  -- Wrong (counts all records)
  IF v.fixed_version IS NOT NULL THEN
```

```sql
  -- Correct (counts only fixable)
  IF v.fixed_version IS NOT NULL AND v.fixed_version != '' THEN
```

  Verified with Testing

  Tested using node:17 image (7997 total vulnerabilities, 5210 fixable):

  | Test                       | Result                                     |
  |----------------------------|--------------------------------------------|
  | Before fix                 | fixable_cnt = NULL                         |
  | With IS NOT NULL only      | fixable_cnt = 7997 ❌ (wrong - counts all) |
  | With IS NOT NULL AND != '' | fixable_cnt = 5210 ✅                      |
  | API reported               | fixable: 5210 ✅                           |

Vad1mo:
> @Vad1mo yes, but say me, I am in process upgrading from 2.7 and noticed this problem. Not always people upgrade to the latest and greatest, so probably best to fix it here and I will also create new PR as latest upgrade which would update only empty fixed counts.

All users who are on version >2.9 will be left out of the feature.

We don't do that.

Vad1mo (approved):
I completed tests, updated the PR to meet schema update requirements, and found another bug. 

This PR is good to go.

bupd (approved):
please update the PR title to 0180_2.15.0_schema.up.sql 

/lgtm

stonezdj (changes requested):
Thanks for your contribution, it is a bug for v2.9.0, but current we are in Harbor v2.15.0,  v2.9.0 seems to be an old release, there is little chance that the scan_report before 2.9.0 still exist, and the for loop iterate all scan_report during the migration would takes long time, we don't risk the upgrade process, suggest users who have this issue to run this script manually.

tkatkov:
@stonezdj During migration from 2.7 modified script found a LOT of fixable vulnerabilities.  And that sql is already incredible slow.

wy65701436:
> @stonezdj During migration from 2.7 modified script found a LOT of fixable vulnerabilities. And that sql is already incredible slow.

Thanks for raising this issue—we weren’t aware of it previously.

Re-running scan-all will resolve the issue, even if you did not run the SQL manually. 

We do not recommend adding a fix for this, as it may introduce potential performance issues and is easy to work around. Given that this affects a very old version, introducing a fix would be risky and not beneficial for most users.

tkatkov:
@wy65701436  re-running scan-all is not a feasible solution when you have ~500.000 images.

Also if, as you say it affect very old version, how it is that risky? It would not be even used by most users, but certainly will benefit some.

wy65701436:
> re-running scan-all is not a feasible solution when you have ~500.000 images.

1, Workaround without code changes
For users who need to fix existing data without re-scanning, the SQL fix can be executed manually. Anyone who encounters this problem can run the SQL script directly against their database.

2, As for the re-scanning

From a CVE scanning perspective, re-scanning is actually the better approach:
- Vulnerability data changes constantly
- Users generally care about the latest scan results, not outdated data
- Regarding performance concerns, we already have open issues and plans to improve Scan All performance. We can discuss it on there.

3, Low Impact for Most Users

Running this on every Harbor upgrade (even for users who don't need it) would increase upgrade times; Upgrade time becomes dependent on the volume of historical scan data, making upgrade planning difficult for operators. Most existing Harbor installations are already past v2.9.0, and while this migration would run for them, it only affects users who haven't re-scanned their artifacts since the original buggy migration


IMO, we can write a KB(FAQ) for the community to tell the condition and steps to fix it. So anyone who cares about it, can follow the KB.

wy65701436:
@tkatkov Let me know if you’re okay with closing this PR, or feel free to share your thoughts.

PR #22669, opened 2025-12-23
feat: Add Error Handling Linters

### LINKED ISSUE #22665: Add more linters to catch bugs earlier
The recent typo fix in cd46a58 (`colume` → `column` in a struct tag) got me thinking, we could catch more of these issues with better linting.

Our current golangci config is pretty minimal. Adding linters like `gosec` (security), `dupl` (duplicate code), and a few others would help us:
- Catch bugs before they hit main
- Enforce better Go practices
- Reduce manual review burden

### LINKED ISSUE #22666: Add errorlint to enforce proper Go 1.13+ error handling
## Description

Go 1.13 introduced error wrapping with `%w` and `errors.Is`/`errors.As`. Code that doesn't follow these patterns breaks when errors get wrapped and these bugs are silent until runtime.

`errorlint` catches these issues at lint time:

## What it catches

| Bad | Good | Why |
|-----|------|-----|
| `err == ErrNotFound` | `errors.Is(err, ErrNotFound)` | Direct comparison fails on wrapped errors |
| `err.(*MyError)` | `errors.As(err, &myErr)` | Type assertion fails on wrapped errors |
| `fmt.Errorf("failed: %v", err)` | `fmt.Errorf("failed: %w", err)` | `%v` loses the error chain |

## Adoption

This linter is widely adopted across the Go ecosystem:

- **[kubernetes/release](https://github.com/kubernetes/release)** 
- **[kubernetes-sigs/kwok](https://github.com/kubernetes-sigs/kwok)**  
- **[stackrox/kube-linter](https://github.com/stackrox/kube-linter)**
- **[golangci-lint itself](https://github.com/golangci/golangci-lint)**

## Benefits for Harbor

- **Catches bugs early** - error handling issues found at CI, not production
- **Reduces tech debt** - enforces modern Go patterns consistently
- **Auto-fixable** - errorlint can automatically fix most issues
- **Future-proof** - as we wrap more errors for context, existing code won't silently break

## Configuration

Can be enabled incrementally:
```yaml
linters:
  enable:
    - errorlint
```

bupd:
Add error handling linters to catch common bugs:
- `nilerr` - Catches returning nil when error is not nil
- `nilnil` - Catches returning (nil, nil) in (value, error) functions
- Fixed all violations with proper error handling
- Added `//nolint` comments for intentional patterns with justification

## Issue being fixed
Fixes #22665
Fixes #22666

- [ ] Label the PR as needed. "release-note/update"

### DIFF
--- a/src/.golangci.yaml
+++ b/src/.golangci.yaml
@@ -9,6 +9,7 @@ linters:
     - ineffassign
     - misspell
     - nilerr
+    - nilnil
     - revive
     - staticcheck
     - whitespace
--- a/src/.golangci.yaml
+++ b/src/.golangci.yaml
@@ -8,6 +8,7 @@ linters:
     - govet
     - ineffassign
     - misspell
+    - nilerr
     - revive
     - staticcheck
     - whitespace
--- a/src/.golangci.yaml
+++ b/src/.golangci.yaml
@@ -28,6 +28,11 @@ linters:
       - common-false-positives
       - legacy
       - std-error-handling
+    rules:
+      - linters:
+          - revive
+        text: "var-naming: avoid meaningless package names"
+        path: "(common/|controller/event/handler/util/|core/utils/|jobservice/common/utils/|pkg/permission/types/|pkg/quota/types/|server/middleware/util/|server/registry/util/|pkg/reg/util/)"
     paths:
       - third_party$
       - builtin$
--- a/src/controller/artifact/processor/chart/chart.go
+++ b/src/controller/artifact/processor/chart/chart.go
@@ -119,6 +119,7 @@ func (p *processor) AbstractAddition(_ context.Context, artifact *artifact.Artif
 			}, nil
 		}
 	}
+	//nolint:nilnil // addition type is not found
 	return nil, nil
 }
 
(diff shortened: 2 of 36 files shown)


reasonerjt:
I see you are adding tens of "// nolint:nilnil", so is it really needed?

bupd:
> I see you are adding tens of "// nolint:nilnil", so is it really needed?

Yeah, it's not a best practice to return nil, nil 

And currently we are having this and the problem is, when I change it creates differences in the entire logic of how this handled in some cascading into more and more changes.

Just added nolint for the already present, in hopes of preventing this kind of antipatterns. Any new additions will fail the linter.

reasonerjt:
@bupd 
I don't think we should add a linter, skip the check of linter in tens of places.
If this indeed is an issue we should fix it and enable the linter, otherwise, the linter makes little sense.

bupd:
> @bupd 
> I don't think we should add a linter, skip the check of linter in tens of places.
> If this indeed is an issue we should fix it and enable the linter, otherwise, the linter makes little sense.

Yes I agree, let me see how I can improve that, will put this pr in draft mode until then 

Thanks 👍

reasonerjt:
@bupd 

Thanks.

From another perspective, I find many linters are quite opinionated.  We may also consider whether we need to enforce the check of `nilnil`.

PR #22678, opened 2025-12-24
bugfix/max-upstream-comm

### LINKED ISSUE #22676: Add range validation for max_upstream_conn metadata
## Summary
Backend validation for `max_upstream_conn` only checks if value parses as int, not valid range.

## Expected Behaviour
Backend rejects invalid values:
- Only accept `-1` (unlimited) or positive integers > 0
- Reject `0` and negative numbers other than `-1`
- Match frontend validation rules

## Actual Behaviour
`src/server/v2.0/handler/project_metadata.go:164-169`

Only validates `strconv.ParseInt` succeeds:
```go
case proModels.ProMetaMaxUpstreamConn:
    v, err := strconv.ParseInt(value, 10, 32)
    if err != nil {
        return nil, errors.New(nil).WithCode(errors.BadRequestCode).WithMessagef("invalid value: %s", value)
    }
    metas[proModels.ProMetaMaxUpstreamConn] = strconv.FormatInt(v, 10)
```

Values like `0`, `-2`, `-100` pass backend validation but have undefined behavior.

intojhanurag:
This PR adds backend range validation for the `max_upstream_conn` project metadata. Previously, the backend only validated that the value could be parsed as an integer, allowing invalid values such as `0` or negative numbers other than `-1`, which could lead to undefined behavior. This change ensures the backend validation matches the frontend rules.

### Issue 
Fixes #22676

- [x] Label the PR as needed. `release-note/bug-fix`
- [x] Made sure tests are passing and test coverage is added or updated as needed.
- [x] Considered the docs impact and confirmed no documentation changes are required.


src/server/v2.0/handler/project_metadata.go
```diff
@@ -164,7 +164,14 @@ func (p *projectMetadataAPI) validate(metas map[string]string) (map[string]strin
 	case proModels.ProMetaMaxUpstreamConn:
 		v, err := strconv.ParseInt(value, 10, 32)
 		if err != nil {
-			return nil, errors.New(nil).WithCode(errors.BadRequestCode).WithMessagef("invalid value: %s", value)
+			return nil, errors.New(nil).
+				WithCode(errors.BadRequestCode).
+				WithMessagef("invalid value: %s", value)
+		}
+		if v == 0 || v < -1 {
```
wy65701436: `if v<=0?`

wy65701436 (changes requested):

src/server/v2.0/handler/project_metadata.go
intojhanurag: image
Hey @wy65701436 , it was mentioned in the issue other than -1 :)

src/server/v2.0/handler/project_metadata.go
```diff
@@ -164,7 +164,14 @@ func (p *projectMetadataAPI) validate(metas map[string]string) (map[string]strin
...
+			return nil, errors.New(nil).
+				WithCode(errors.BadRequestCode).
+				WithMessagef("invalid value for %s: %d", key, v)
```
Vad1mo: The error message is useless. Imagine you see the message outside this context, when the server boots up.

src/server/v2.0/handler/project_metadata.go
chlins: If we want to exclude all other negative values except -1, we can use a more straightforward condition: `v <= 0 && v != -1`.

chlins: Add an empty line before this condition.

intojhanurag:
Hey @chlins , Is this good to go ?

chlins (approved):
lgtm

OrlinVasilev:
@wy65701436 can you review again please @bupd

bupd (approved):
/lgtm

bupd (approved):
/lgtm

bupd:
I believe we can ignore the codecov for this one.

wy65701436 (approved):
lgtm

Issue #22134, opened 2025-06-30
Migrate Internal Pipeline to Github Actions


bupd:
## Goal
- [ ] Migrate the internal E2E pipeline to GitHub Actions under goharbor.
- [ ] Proposal https://github.com/goharbor/community/pull/268

This goal was discussed in our meeting, recording is available in the Slack:
https://cloud-native.slack.com/files/U051J3DBMJS/F08MSD1P739/e2e_pipeline.mp4

The updates will be tracked on this issue.

## Migration Plan
After consideration, it’s best to run these pipelines on PRs instead of nightly runs.

The rationale: if 3 - 4 PRs get merged in a day and the nightly run fails, it becomes harder to identify the issue, and the main branch ends up broken. My goal is to keep main working as much as possible.

## Strategy
The idea is to:
- Break down the large test suite into smaller, isolated components.
- Run these subtests concurrently.
- Each test currently takes around 10–30 mins.
- The existing pipeline takes 45 mins or more (worst case scenario), So, I believe this approach is feasible.

If we split and run separately, total runtime may reduce to ~20-30 mins.

## These tests include components such as:
- [ ] Playwright Setup (#22462) 
- [ ] DB.robot (#22472)
- [ ] Routing.robot (#22533) 
- [ ] Teardown.robot ( )
- [ ] Common_GC.robot ( )
- [ ] Common.robot
- [ ] #22150
- [ ] P2P_Preheat.robot
- [ ] Trivy.robot

### !!Webhook.robot
This e2e test involves infra requirement such as 
1. Webhook endpoint 
2. Webhook endpoint UI. where it can find the triggered webhook.

So, we need to setup webhook server and with an accessible ui as defined in the `robot` tests.

## Note
if the time to setup the environment in the github actions is too high (i.e time to setup the environemnt take more than 20 - 30 mins). Because at this stage people can't wait an hour to get the tests complete on PR. This is the overall idea. hope this make sense. we will create a deployed service under Harbor AWS acc.

The migration to playwright testcases would include all keywords inside - tests/resources/Harbor-Pages, and 
All testcases inside 
- Only these two testcases in https://github.com/goharbor/harbor/blob/b0e54cad7645ab9fa64180dd6589087391d04b01/tests/robot-cases/Group0-BAT/BAT.robot#L25C1-L46C18
- Group1-Nightly
- Group2-Longevity (*if needed)

Tests that will NOT be migrated.
- all of `tests/apitests`
[...]

PR #22472, opened 2025-10-17
Feat: Add Playwright tests for DB scenarios

ramu-nukavarapu:
# Comprehensive Summary of your change
This PR adds Playwright E2E tests for DB scenarios.

Included test coverage:

- Create user
- Update user comment
- Update user password
- Delete multiple users
- Edit self registration
- Admin add new user

# Issue being fixed
Fixes #22134

### [Stack](https://github.com/kitlangton/stack)

1. #22462
2. **#22472** 👈 current
3. #22707

PR #22707, opened 2026-01-04
Migration of Common_gc tests from Robot to Playwright

adityachopra29:
This PR implements the migration of Common_GC.robot to Playwright.

Changes made:

- Adds Common_GC Playwright tests
- Adds setup, cleanup, navigation, image push/pull/tag helper flows needed by the suite
- Keeps tests independent and self-contained

# Issue being fixed
Fixes #22134

### Test Migration Status

- [x] Project Quota Sorting
- [x] Garbage Collection
- [x] GC Untagged Images
- [x] Project Quotas Control Under GC
- [x] Garbage Collection Accessory

### [Stack](https://github.com/kitlangton/stack)

1. #22462
2. #22472
3. **#22707** 👈 current

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,6 @@
 src/portal/node_modules/
+src/portal/playwright-report/
+src/portal/test-results/
 doc/
 docs/
 .git
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -1,0 +1,91 @@
+name: Playwright Tests
+
+env:
+   POSTGRESQL_HOST: localhost
+   POSTGRESQL_PORT: 5432
+   POSTGRESQL_USR: postgres
+   POSTGRESQL_PWD: root123
+   POSTGRESQL_DATABASE: registry
+   DOCKER_COMPOSE_VERSION: 2.27.1
+   HARBOR_ADMIN: admin
+   HARBOR_ADMIN_PASSWD: Harbor12345
+   CORE_SECRET: tempString
+   KEY_PATH: "/data/secret/keys/secretkey"
+   REDIS_HOST: localhost
+   REG_VERSION: v2.7.1-patch-2819-2553
+   UI_BUILDER_VERSION: 1.6.0
+
+on:
+  pull_request:
+    branches: [ main ]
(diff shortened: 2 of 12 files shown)


adityachopra29:
To the maintainers @bupd @Vad1mo @wy65701436 @OrlinVasilev. This pr is ready for review. Please take a look.

One thing to note here is that the tests [Garbage Collection](https://github.com/goharbor/harbor/blob/b0e54cad7645ab9fa64180dd6589087391d04b01/tests/robot-cases/Group1-Nightly/Common_GC.robot#L45) and [Garbage Collection Accessory](https://github.com/goharbor/harbor/blob/b0e54cad7645ab9fa64180dd6589087391d04b01/tests/robot-cases/Group1-Nightly/Common_GC.robot#L112) are check the working of the GC in different senarios, but GC does not consider artifacts uploaded in the past 2 hours, so it does not consider the project and artifacts that we push for testing.
Any suggestions on how to handle this?

adityachopra29:
Also, I found that there are a lot of different utility functions (for cosign/notation operations, docker functions, harbor navigation, etc) that I found will be used by other test cases as well, and everybody is reimplementing them for now...so I guess we need to modularize this at some point in time...so we should probably have an approach for this as well. 
For now I have included them all in one file, but I can separate that out once the plan is decided. What do you suggest?

adityachopra29:
Lastly, currently, we don’t have a good developer setup for building Playwright tests. The playwright.yml file can’t be used to create a dev environment, which made local development inefficient for me. I’ve added a temporary workaround using sample.env, but a Docker-based setup similar to the Robot test cases would be a better long-term solution.

bupd:
So currently the dev workflow is simple you just need to npm install which would install playwright 

And for running playwright use npx playwright 

On the sample env part and the playwright.yml this might be a good addition given that it simplifies the env required to run playwright tests locally 

You can check the proposal i have linked the dockerfile for running playwright.

adityachopra29:
Okay..found the docker file

bupd (approved):
/lgtm

adityachopra29:
Are there any changes you suggest I should make to this PR?

Issue #17122, opened 2022-07-05
Invalid user name or password on fresh install


premsgr6:
![image](https://user-images.githubusercontent.com/46933846/177282943-f147efaa-68d1-4e45-86db-e8582917faf1.png)

After helm install XXX, and wait for all pods running.
I enter the portal using default username/password ( admin/Harbor12345), I got "Invalid user name or password".

I have no idea why I cannot login.

![image](https://user-images.githubusercontent.com/46933846/177283730-7c2793e5-c710-40dd-89f7-49663392e627.png)

i have checked the logs of harbor-core

![image](https://user-images.githubusercontent.com/46933846/177284368-a6390f7a-7227-4d93-b6b5-2570305d5480.png)

i have checked the nginx logs  & all pods are in running state

![image](https://user-images.githubusercontent.com/46933846/177284774-a95c4212-33ef-4ec2-a8f2-248cbe95df56.png)

wangxg422:
I also have the same problem. The password is clearly correct, and there is no error log for each pod. What is the reason?

wangxg422:
Maybe it has something to do with the browser. I reinstalled harbor with helm, and then opened the privacy window of the browser to log in. It was successful

premsgr6:
@wangxg422 Which helm chart version you have used? I have reinstalled it couple of time setting different passwords and harbor-core secrets is able to pull the new password too, still the same issue on browser. Tried it on privacy window too, didn't work.

wangxg422:
@premsgr6  The chart package I downloaded with the official tutorial is "helm fetch harbor/harbor --untar". I am a self signed certificate. I log in with edge browser and prompt the wrong user name and password (as shown in the figure). However, I can log in normally using the privacy window of edge.

![image](https://user-images.githubusercontent.com/35413234/177446222-03c2869c-2daf-4e41-a63c-db220ea54130.png)

prem9966:
@wangxg422  I tried to reinstall with latest version with default username admin / password Harbor12345 , I am facing the same issue with diff version 1.8.0 , 1.9.0 & 1.9.2 (latest)  & tried to loging with diff private browser

premsgr6:
When I expose the harbor with LoadBalancer on http same credentails work. With ingress and cert-manager certificate for https it gives Invalid username and password error.

wangxg422:
[...]

Issue #12639, opened 2020-07-31
How to reset lost admin password


lalithvaka:
What can we help you?
With out looking at the comments in harbor.yml,  I have re ran the ./prepare with updated password to change the Admin password from the default Harbor12345. Now am not able to retrieve. the password even though I tried following https://github.com/goharbor/harbor/issues/8778.

![image](https://user-images.githubusercontent.com/8292387/88999414-5eccaf80-d2a9-11ea-85e2-6d664d98695f.png)

Tried updating the password field directly in the data base per #8778 screenshot but still no use.

lalithvaka:
Just to provide sequence of steps 
1. Installed Harbor on a VM with RHEL 7.X with out Http with default Harbor admin password and default DB password
2. Applied https config using self signed certificates
3. reran the ./prepare process 
4. Harbor is up and running with the default admin password. 
5. Created an additional user and did some settings changes and created a new project / repo.
6. Was able to login using new user and push an image
7. Enabled Clair by running the ./prepare --with-clair option process. everything good so far.
-- at this time, we have taken the VM snapshot and also increased the VM resources (CPU and Memory)
8. Updated both Admin and DB password using Harbor.yml(with out knowing I need to do this through UI) and re ran the ./prepare script.
9. I was not able to login to the browser UI with admin or user 
10. Tried following the steps laid out here https://kb.vmware.com/s/article/76289 to change the admin password and shaw depicted in [here](https://github.com/goharbor/harbor/issues/8778#issuecomment-523729511)
11. Still could not login with either admin or other user.

Hope this helps understand the steps I took.

stonezdj:
Something maybe wrong in step 8.  The DB password cannot be changed in Harbor.yml, Because it is a exsiting database, you should also update the password in database and both in harbor.yml, if you just change the DB password in harbor.yml, it will cause the other component cannot connect to the database, the admin password cannot be reset either.  Change one thing in a time.

lalithvaka:
@stonezdj Can you please suggest how we can restore both the passwords with out impacting the entire setup / projects / images etc?? At this point, admin user and other user on the instance lost access.

stonezdj:
You should rollback the previous username password in harbor.yml and run with prepare, and restart harbor and check the  core.log to make sure the core container can connect to database with the username password.

Then change the admin password in Harbor web admin console. not in the harbor.yml

lalithvaka:
Thank you for all your help

Issue #11600, opened 2020-04-13
Keep getting "Invalid username and password" when trying to login 


kratos81:
I have been trying to install harbor on AWS and GKE and each time , I am not able to login using the default password.  Im not sure if this is a bug

version is v1.10.1 

This is the log from the database 
```
 k logs harbor-harbor-database-0 -f
The files belonging to this database system will be owned by user "postgres".
This user must also own the server process.

The database cluster will be initialized with locales
  COLLATE:  en_US.UTF-8
  CTYPE:    en_US.UTF-8
  MESSAGES: C
  MONETARY: C
  NUMERIC:  C
  TIME:     C
The default text search configuration will be set to "english".

Data page checksums are disabled.

fixing permissions on existing directory /var/lib/postgresql/data ... ok
creating subdirectories ... ok
selecting default max_connections ... 100
selecting default shared_buffers ... 128MB
selecting default timezone ... UTC
selecting dynamic shared memory implementation ... posix
creating configuration files ... ok
running bootstrap script ... ok
performing post-bootstrap initialization ... ok
syncing data to disk ... ok

WARNING: enabling "trust" authentication for local connections
You can change this by editing pg_hba.conf or using the option -A, or
--auth-local and --auth-host, the next time you run initdb.

Success. You can now start the database server using:

    pg_ctl -D /var/lib/postgresql/data -l logfile start

postgres
waiting for server to start....LOG:  database system was shut down at 2020-04-13 00:02:15 UTC
LOG:  MultiXact member wraparound protections are now enabled
LOG:  database system is ready to accept connections
LOG:  autovacuum launcher started
 done
server started
ALTER ROLE


/docker-entrypoint.sh: running /docker-entrypoint-initdb.d/initial-notaryserver.sql
CREATE DATABASE
CREATE ROLE
ALTER ROLE
GRANT


/docker-entrypoint.sh: running /docker-entrypoint-initdb.d/initial-notarysigner.sql
CREATE DATABASE
CREATE ROLE
ALTER ROLE
GRANT


/docker-entrypoint.sh: running /docker-entrypoint-initdb.d/initial-registry.sql
CREATE DATABASE
You are now connected to database "registry" as user "postgres".
CREATE TABLE


LOG:  received fast shutdown request
LOG:  aborting any active transactions
LOG:  autovacuum launcher shutting down
LOG:  shutting down
waiting for server to shut down....LOG:  database system is shut down
 done
server stopped

[...]

PR #22709, opened 2026-01-05
fix(core): add warning when config admin password is ignored

### LINKED ISSUE #22704: change config harbor_admin_password params, start harbor service occur "username or password is not correct"
if not change password, it correct start, but i want when initialization change password,
and i uncomment https params, maybe is affect?
```
# http related config
http:
  # port for http, default is 80. If https enabled, this port will redirect to https port
  port: 80

# https related config
#https:
  # https port for harbor, default is 443
  #port: 443
  # The path of cert and key files for nginx
  #certificate: /your/certificate/path
  #private_key: /your/private/key/path
  # enable strong ssl ciphers (default: false)
  # strong_ssl_ciphers: false
```

thc1006:
## Summary

- Add warning log when `harbor_admin_password` in config is ignored because admin password already exists in database
- Refactor `updateInitPassword` for testability with dependency injection
- Add unit tests for password initialization scenarios

## Background

The `harbor_admin_password` setting only applies during initial installation. When users modify this value after Harbor is running, the change is silently ignored, leading to "username or password is not correct" errors when they try to login with the new password.

This is documented behavior but causes recurring confusion (see related issues below).

## Changes

**src/core/main.go**
- Extract `updateInitPasswordWithMgr` to accept user manager as parameter
- Add warning log when config password is set but will not be applied

**src/core/main_test.go** (new)
- Test: new user password initialization
- Test: existing user password skip
- Test: warning when config password set
- Test: no warning when config password empty

## Test Plan

- [x] Unit tests pass locally
- [x] Build passes
- [ ] Manual verification with Harbor deployment

## Related Issues

Fixes #22704

Related: #21981, #17122, #12639, #11600

### DIFF
--- a/src/core/main.go
+++ b/src/core/main.go
@@ -88,7 +88,10 @@ const (
 )
 
 func updateInitPassword(ctx context.Context, userID int, password string) error {
-	userMgr := pkguser.Mgr
+	return updateInitPasswordWithMgr(ctx, pkguser.Mgr, userID, password)
+}
+
+func updateInitPasswordWithMgr(ctx context.Context, userMgr pkguser.Manager, userID int, password string) error {
 	user, err := userMgr.Get(ctx, userID)
 	if err != nil {
 		return fmt.Errorf("failed to get user, userID: %d %v", userID, err)
--- a/src/core/main_test.go
+++ b/src/core/main_test.go
@@ -1,0 +1,109 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	commonmodels "github.com/goharbor/harbor/src/common/models"
+	usertesting "github.com/goharbor/harbor/src/testing/pkg/user"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUpdateInitPasswordWithMgr_NewUser(t *testing.T) {
+	ctx := context.Background()
+	mockMgr := usertesting.NewManager(t)
+
+	// User without salt (new user needing password initialization)
+	mockMgr.On("Get", ctx, 1).Return(&commonmodels.User{
+		UserID: 1,
+		Salt:   "",
+	}, nil)
+	mockMgr.On("UpdatePassword", ctx, 1, "newpassword").Return(nil)
+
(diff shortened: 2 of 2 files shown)


thc1006:
**TL;DR:** Kindly requesting a re-run of the failed `APITEST_DB_PROXY_CACHE` job and the addition of a `release-note/fix` label. Both failures appear unrelated to the code changes in this PR.

---

Dear maintainers,

Thank you very much for taking the time to review this pull request. I have carefully analyzed the CI failures and would like to respectfully share my findings.

**Regarding `APITEST_DB_PROXY_CACHE` failure:**

The test failure shows `Test Case - Proxy Cache | FAIL | 1 != 0`, which appears to be a flaky test or infrastructure-related issue. This PR only modifies the admin password warning logic in `src/core/main.go` and its corresponding unit tests, and does not touch any Proxy Cache functionality. I have observed that other recent PRs have also experienced similar intermittent failures with this particular test.

I would be most grateful if you could kindly re-run the failed job at your convenience.

**Regarding `Check release-note label set` failure:**

This PR requires a `release-note/*` label to pass the CI check. Since this is a bug fix that addresses issue #22704, I believe `release-note/fix` would be the appropriate label.

Would you be so kind as to add this label when you have a moment?

I sincerely apologize for any inconvenience this request may cause. Please do not hesitate to let me know if there is anything I should modify or if you have any concerns about the implementation.

Thank you very much for your time and consideration.

Best regards

src/core/main.go
```diff
@@ -87,7 +87,10 @@ const (
 )
 
 func updateInitPassword(ctx context.Context, userID int, password string) error {
-	userMgr := pkguser.Mgr
+	return updateInitPasswordWithMgr(ctx, pkguser.Mgr, userID, password)
```
wy65701436, line 91: do we need the updateInitPasswordWithMgr here? why not update the updateInitPassword directly?

src/core/main.go
```diff
@@ -97,10 +100,13 @@ func updateInitPassword(ctx context.Context, userID int, password string) error
 		if err != nil {
 			return fmt.Errorf("failed to update user encrypted password, userID: %d, err: %v", userID, err)
 		}
-
 		log.Infof("User id: %d updated its encrypted password successfully.", userID)
 	} else {
 		log.Infof("User id: %d already has its encrypted password.", userID)
+		if password != "" {
+			log.Warningf("The admin password in configuration (HARBOR_ADMIN_PASSWORD) will not be applied " +
```
wy65701436: How can we determine whether the user is a Harbor admin? At the very least, should we check the user ID?

src/core/main.go
```diff
@@ -97,10 +97,13 @@ func updateInitPassword(ctx context.Context, userID int, password string) error
 		if err != nil {
 			return fmt.Errorf("failed to update user encrypted password, userID: %d, err: %v", userID, err)
 		}
-
 		log.Infof("User id: %d updated its encrypted password successfully.", userID)
 	} else {
 		log.Infof("User id: %d already has its encrypted password.", userID)
+		if userID == adminUserID && password != "" {
```
chlins, line 107: Please simplify to use `Warning` because no format needed and the log message can be use one sentence to avoid `+` cost.

thc1006: Dear @chlins ,
Done. Changed to log.Warning and combined the message into a single line.

PR #22723, opened 2026-01-09
Allows OIDC integration to handle the case where a user only has a single group

### LINKED ISSUE #15416: OIDC | Custom group claim value can be string or array
If only one group is available in the specified group claim, this claim is of type StringOrUri. Only if two or more groups are available, the claim value becomes an array. This behaviour is explained for the "aud" claim at [RFC751](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3). 

C#s System.IdentityModel.Tokens.Jwt for example uses this logic by default for all claims. Single value = StringOrUri. Multiple Values = Array. 

IdentityProviders like IdentityServer4 also handle claims like this.

The current behavior seems to always expect an array, even if only one group is provided. If i log in via OIDC and only have one group as a StringOrUri value, the custom group claim is ignored. If two or more groups are specified, and the claim value is of type array, the group mapping works.

Add group claim mapping. Only provide one group in the claim. The claim is issued as value "string" from the IdentityServer. Group permissions do not work. Only if two or more groups are present and the claim is of type Array[StringOrUri], even if the additional groups are of no relevance, access to projects that have group permissions now works.

**Versions:**
- harbor version: [2.3.1]

lfrancke:

See the linked issue for details on the issue.

This PR first tries to parse all claims as an array first and if that fails it will try to parse it as a string.
The internal UserInfo struct is unchanged.

> [!NOTE]
> I tried to run the tests and I tried to build Harbor and run it locally but I failed miserably.
>
> a)  I couldn't get the tests to run (failed on a Postgres requirement)
> b) I was able to _build_ using `make install` but I couldn't get the docker images to actually run due to some permissions issue (the images seem to require `root`?)

I understand if you don't want to accept this PR untested and I might try again next week but for now...this is it :)

# Issue being fixed
Fixes #15416 


lfrancke:
Thanks for running the tests. I'm not entirely sure what the actual failure is for the UTTEST run. There are a bunch of Errors in there but they don't seem to stop the build. 

I'm not very familiar with this :(

reasonerjt:
@lfrancke Thanks for this PR.

> // According to RFC 7519, a claim can be either a string (single value) or an array (multiple values)

Could you please point out which section in the RFC has such statements?  I tried to find it but only found has statement about `aud` claim 

```
In the general case, the "aud" value is an array of case-
   sensitive strings, each containing a StringOrURI value.  In the
   special case when the JWT has one audience, the "aud" value MAY be a
   single case-sensitive string containing a StringOrURI value.
```
But this does not have to be applicable to `groups` claim IMHO.  Even the wording in the RFC seems to imply that most cases a claim should be either a string or an array.  And "aud" is a special case, which in general should be avoided.

lfrancke:
I have to admit that this (the [original issue](https://github.com/goharbor/harbor/issues/15416)) is where I took that claim from. I'm happy to remove the comment from the PR. I agree: I can't find any reference to it in the RFC itself.

FWIW I've also opened a support case with JumpCloud

Vad1mo:
While the [RFC751](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3) is not clear, many OIDC implementations support both use cases. I looked into Dex, which supports both string for a single entry and array for multiple entries. 

There is no harm in adding this fix, as it won't break backwards compatibility.

reasonerjt:
> many OIDC implementations support both use cases. I looked into Dex, which supports both string for a single entry and array for multiple entries.

@Vad1mo 
Could you elaborate?

I really doubt it, I tested with Dex, which was years ago, but I recall Dex sets an array in "groups" claim, not single string.

My point is, if we have multiple mainstream OIDC providers which set single string in the "groups" claim I'm fine with this PR.  But if it's only JumpCloud we probably wanna keep it in the fork, b/c it seems to me that this is an implementation issue in JumpCloud.  

I don't think we should update the code again when we see some oidc provider tries to put a map in the "groups" claim.

lfrancke:
It's not only JumpCloud. The original issue was opened by someone else using "IdentityServer 4". PingFederate does the same and I'm 99% certain that I've seen others with the same behavior.

There does not seem to be a spec for the groups claim either so to me it seems reasonable to support "what's out there".

We solved it for us by just assigning a second dummy group so it's not a big problem but it'll make the lives of others easier at very little cost.

reasonerjt:
@lfrancke 
Ah, I misremembered.  Thanks for clarifying.

I'll approve and merge this one.  The code LGTM.  Thanks for your time!

reasonerjt (approved):

lfrancke:
Thank you very much!

PR #22893, opened 2026-03-01
feat(proxy): implement local cache with LRU eviction for object storage backends (#21212)

### LINKED ISSUE #21212: Implement local cache with LRU eviction for object storage backends
When object storage (e.g., AWS S3, Google Cloud Storage) is enabled as the backend for Harbor, all image layers and artifacts are fetched from the remote storage every time they are pulled. While using object storage is advantageous for scalability and reduces the need to maintain large local disk volumes—especially in bare-metal setups—it introduces inefficiencies in terms of cost, bandwidth usage, and latency due to frequent data transfers.

Introduce a local caching mechanism for image blobs within Harbor for frequently accessed images when using an object storage backend. The cache would store recently pulled images on local storage, utilizing a Least Recently Used (LRU) eviction policy to manage cache cleanup and ensure optimal use of disk space.

Benefits:
- Cost Reduction: Minimizes bandwidth costs associated with repeatedly fetching the same images from object storage.
- Improved Performance: Decreases latency for image pulls by serving frequently requested images from the local cache.
- Optimized Resource Utilization: Maintains the benefits of object storage (scalability, reduced local storage requirements) while enhancing efficiency.
- Increased Reliability: Allows Harbor to serve cached images even during object storage failures, ensuring continuous availability of frequently used images.

Rajneesh180:
fixes #21212

## What this PR does / why we need it:
This PR introduces a bounded Local Disk LRU Cache layer for the Proxy Cache feature, specifically designed to eliminate redundant egress costs when using Object Storage (S3, GCS).

It intercepts downstream HTTP blob pulls via a custom CachingResponseWriter in the repoproxy middleware and concurrently streams the bits to a local disk buffer (default max 10GB). Subsequent pulls for the same digest bypass the remote backend entirely and stream straight from local disk.

Sync GC evictions are performed using an O(1) in-memory LRU tracking list to purge the oldest disk bytes.

## PR Checklist
- [x] I have read the CONTRIBUTING.md
- [x] My commits are signed off
- [x] I have added unit tests for my changes
- [x] All existing tests pass

### DIFF
--- a/src/controller/proxy/controller.go
+++ b/src/controller/proxy/controller.go
@@ -18,6 +18,9 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"os"
+	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -51,6 +54,9 @@ var (
 	// Ctl is a global proxy controller instance
 	ctl  Controller
 	once sync.Once
+
+	// BlobCache is the global LRU cache instance managing Proxy blob eviction
+	BlobCache *LocalLRUCache
 )
 
 // Controller defines the operations related with pull through proxy
@@ -92,6 +98,18 @@ func ControllerInstance() Controller {
 			cache:           cache.Default(),
 			handlerRegistry: NewCacheHandlerRegistry(l),
 		}
+
+		// Initialize LRU proxy blob cache with an environment variable limit (default 10GB)
+		// Eviction callback physically deletes the cached file
+		sizeLimitMBStr := os.Getenv("PROXY_LOCAL_CACHE_SIZE_MB")
+		var sizeLimit int64 = 10 * 1024 * 1024 * 1024 // 10GB default
+		if parsed, err := strconv.ParseInt(sizeLimitMBStr, 10, 64); err == nil && parsed > 0 {
+			sizeLimit = parsed * 1024 * 1024
+		}
+		BlobCache = NewLocalLRUCache(sizeLimit, func(digest string) error {
+			cachePath := filepath.Join(os.TempDir(), "lru_blob_cache", digest)
+			return os.Remove(cachePath)
+		})
 	})
 
 	return ctl
--- a/src/controller/proxy/controller.go
+++ b/src/controller/proxy/controller.go
@@ -100,15 +100,21 @@ func ControllerInstance() Controller {
 		}
 
 		// Initialize LRU proxy blob cache with an environment variable limit (default 10GB)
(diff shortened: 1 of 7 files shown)


Rajneesh180:
Hey @stonezdj, I went through all the review feedback and pushed a round of fixes addressing the issues raised. Here's a summary:

**Race conditions & security:**
- Switched to `os.CreateTemp` for unique temp file names — eliminates the race when multiple goroutines cache blobs concurrently
- Tightened cache directory permissions to `0700` and file permissions to `0600`

**Error handling:**
- `caching_writer.go`: now handles `file.Write` errors (cleans up temp file and abandons caching gracefully) and checks `os.Rename` — only adds to the LRU tracker on success
- `controller.go`: eviction callback treats `os.IsNotExist` as success so we don't log noise for already-deleted files
- `lru_cache.go`: on eviction failure, re-adds the entry to the tracker to keep disk and memory accounting consistent

**Proxy path correctness:**
- `CachingResponseWriter` is now only created on the actual remote proxy path — the local-blob and cache-hit branches just use the plain `ResponseWriter`
- Response headers (`Content-Length`, `Docker-Content-Digest`) are now set *before* `io.CopyN` so they actually reach the client
- Cache-hit path now also sets proper headers before calling `http.ServeFile`

**Other:**
- Removed the incorrect `BlobCache.Add` from `PushBlob` in `local.go` — those blobs live in the local registry, not the disk cache, so adding them to the LRU tracker would cause mismatches on eviction
- Allow `PROXY_LOCAL_CACHE_SIZE_MB=0` to explicitly disable the cache

**Tests:**
- Added `caching_writer_test.go` with 3 unit tests (cache miss → blob cached to disk, non-success status → skipped, concurrent writers → unique temp files)
- Updated existing LRU cache test for the re-add-on-eviction-failure behavior

All tests pass locally. Let me know if anything else needs attention!

Rajneesh180:
Rebased both commits with DCO sign-off — the DCO check should pass now. @stonezdj would love your review when you get a chance
 C-c
 exit
 printf '\003'

Rajneesh180:
Rebased onto current main (was 691 commits behind) and addressed the CodeQL XSS finding — blob responses now explicitly set `Content-Type: application/octet-stream` and `X-Content-Type-Options: nosniff` in both the cache-hit and proxy paths. The previous empty mediaType let the browser potentially sniff binary blob data as HTML.

All tests pass locally. @stonezdj would appreciate a review when you get a chance — the core implementation and all earlier review feedback haven't changed, this push just brings it up to date and closes the security finding.

PR #22979, opened 2026-03-09
feat: add webhook trigger for tag created event (#22640)

### LINKED ISSUE #22640: Add Webhook Trigger for “Tag Created” Event
### Discussed in https://github.com/goharbor/harbor/discussions/22635

<div type='discussions-op-text'>

<sup>Originally posted by **yashid-mohamed** December 11, 2025</sup>
I’d like to request a new webhook event type that triggers specifically when a tag is created (without requiring the artifact itself to be pushed again).

**Problem**

When performing tag promotion (e.g., promoting an existing digest from 0.403.0 → 0.403.0-prod) using the Harbor API or UI, Harbor currently does not emit any webhook for this action.

This creates a gap for CI/CD pipelines and downstream automation tools that rely on webhook notifications to react to new tags. Since no new artifact is pushed and no scan/build occurs, none of the existing webhook types fire.

**Use Case**

Many promotion workflows rely on copying or creating tags inside Harbor (to preserve digest immutability and avoid re-pushing images). Examples:

- Promotion from dev → prod
- Tagging releases in automation pipelines
- Triggering deployments based on tag creation

Without a webhook event for tag creation, external systems must continuously poll Harbor, which is inefficient.</div>

DCchoudhury15:
# Add Webhook Trigger for "Tag Created" Event

## Summary

When a tag is created without pushing a new artifact (e.g. tag promotion `0.403.0` → `0.403.0-prod`), Harbor didn't fire any webhook. This PR adds `CREATE_TAG` as a webhook-supported event so external systems can react to tag creation.

### What changed

- **`controller/tag/controller.go`** — Fire `CreateTagEventMetadata` from `Create()` so the event triggers for all tag creation paths.
- **`server/v2.0/handler/artifact.go`** — Remove the now-duplicate event firing from the API handler.
- **`controller/event/handler/webhook/artifact/artifact.go`** — Handle `CreateTagEvent` by mapping it to `ArtifactEvent` for webhook dispatch.
- **`controller/event/handler/init.go`** — Subscribe `artifact.Handler` to `TopicCreateTag`.
- **`pkg/notification/notification.go`** — Add `TopicCreateTag` to supported event types (makes it available in webhook policy UI).
- **`pkg/notifier/formats/cloudevents.go`** — Map `TopicCreateTag` → `harbor.tag.created`.

## Fixes #22640

- [x] Accepted the DCO
- [x] Tests passing (`go test`, `go vet`, `golangci-lint`, `-race` all clean)
- [ ] Label: `release-note/new-feature`
- [ ] Docs impact considered

### DIFF
--- a/src/controller/event/handler/init.go
+++ b/src/controller/event/handler/init.go
@@ -42,6 +42,7 @@ func init() {
 	_ = notifier.Subscribe(event.TopicScanningFailed, &scan.Handler{})
 	_ = notifier.Subscribe(event.TopicScanningStopped, &scan.Handler{})
 	_ = notifier.Subscribe(event.TopicScanningCompleted, &scan.Handler{})
+	_ = notifier.Subscribe(event.TopicCreateTag, &artifact.Handler{})
 	_ = notifier.Subscribe(event.TopicReplication, &artifact.ReplicationHandler{})
 	_ = notifier.Subscribe(event.TopicTagRetention, &artifact.RetentionHandler{})
 
--- a/src/controller/event/handler/webhook/artifact/artifact.go
+++ b/src/controller/event/handler/webhook/artifact/artifact.go
@@ -52,6 +52,16 @@ func (a *Handler) Handle(ctx context.Context, value any) error {
 		return a.handle(ctx, v.ArtifactEvent)
 	case *event.DeleteArtifactEvent:
 		return a.handle(ctx, v.ArtifactEvent)
+	case *event.CreateTagEvent:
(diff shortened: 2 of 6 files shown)


yufengjinsong:
| I came across your comment regarding raising a PR for this issue. I'm currently affected by the same problem.
In Harbor image replication, when an image already exists in the destination registry (with the same digest), any new tags added to that image fail to trigger replication and therefore are not synced to the cloud. Only the first tag that was pushed gets replicated.

affected by the same problem in Harbor image replication feature,  should it be upgrade too?

PR #22994, opened 2026-03-15
perf(blob): fix full table scan in blob unassociation check during GC

### LINKED ISSUE #22995: perf: slow delete API caused by full table scan in blob unassociation check during GC

`FindBlobsShouldUnassociatedWithProject` is called during GC to determine which blobs should be unassociated from a project. The query should complete quickly regardless of the number of shared layers across images.

Instead, the underlying `SELECT ... WHERE digest_blob IN (?, ?, ...)` query ends up doing a full hash join on `artifact_blob`, causing the delete API to hang for minutes in affected environments.

1. Run a Harbor instance with a large number of Go-based images (~3M artifacts). Go-heavy stacks are particularly affected because the base Go image (`FROM golang:...`) is the first layer in almost every service's Dockerfile, making that blob digest appear in a very high fraction of rows in `artifact_blob`.
2. Trigger a GC run via the Harbor UI or API.
3. Observe that delete API calls take an unusually long time to complete.
4. On the DB side, run `EXPLAIN ANALYZE` on the query in `FindBlobsShouldUnassociatedWithProject` to confirm the query planner is performing a full sequential scan instead of using the index on `digest_blob`.

**Query plan — before fix (single blob lookup, no LIMIT):**

```sql
EXPLAIN (ANALYZE, BUFFERS)
SELECT b.digest_blob FROM artifact a
JOIN artifact_blob b ON a.digest = b.digest_af
WHERE a.project_id = '6'
AND b.digest_blob = 'sha256:4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1';
```

```
 Gather  (cost=1848447.17..2695809.25 rows=2465262 width=72) (actual time=64042.892..97011.442 rows=2499108 loops=1)
   Workers Planned: 2
   Workers Launched: 0
   Buffers: shared hit=1976626, temp read=115486 written=115856
   ->  Parallel Hash Join  (cost=1847447.17..2448283.05 rows=1027192 width=72) (actual time=64042.491..96292.923 rows=2499108 loops=1)
         Hash Cond: ((a.digest)::text = (b.digest_af)::text)
         ->  Parallel Seq Scan on artifact a  (cost=0.00..516517.77 rows=2446138 width=72) (actual time=0.011..4684.719 rows=5876706 loops=1)
               Filter: (project_id = 6)
               Rows Removed by Filter: 8894
               Buffers: shared hit=485890
         ->  Parallel Hash  (cost=1814347.60..1814347.60 rows=1002765 width=144) (actual time=32392.037..32392.038 rows=2473424 loops=1)
               Buckets: 32768  Batches: 128  Memory Usage: 3648kB
               ->  Parallel Seq Scan on artifact_blob b  (cost=0.00..1814347.60 rows=1002765 width=144) (actual time=0.015..14575.063 rows=2473424 loops=1)
                     Filter: ((digest_blob)::text = 'sha256:...'::text)
                     Rows Removed by Filter: 59682530
                     Buffers: shared hit=1490736
 Planning Time: 0.882 ms
 Execution Time: 97267.675 ms
```

**Key observations:**
- Full sequential scan on `artifact_blob` — **62 million rows scanned** for a single digest lookup
- Index on `digest_blob` not used despite existing
- `Workers Planned: 2, Workers Launched: 0` — parallel workers did not spin up, ran single-threaded
- **~97 seconds for a single blob** — GC calls this for hundreds of blobs at once

**Query plan — after fix (LIMIT per subquery):**

```sql
EXPLAIN (ANALYZE, BUFFERS)
SELECT b.digest_blob FROM artifact a
JOIN artifact_blob b ON a.digest = b.digest_af
WHERE a.project_id = '6'
AND b.digest_blob = 'sha256:4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1'
LIMIT 1;
```

```
 Limit  (cost=0.56..3.76 rows=1 width=72) (actual time=0.049..0.080 rows=1 loops=1)
   Buffers: shared hit=11
   ->  Nested Loop  (cost=0.56..4285510.96 rows=2676577 width=72) (actual time=0.048..0.078 rows=1 loops=1)
         ->  Seq Scan on artifact_blob b  (cost=0.00..2376630.05 rows=2620999 width=144) (actual time=0.017..0.029 rows=1 loops=1)
               Filter: ((digest_blob)::text = 'sha256:...'::text)
               Rows Removed by Filter: 57
               Buffers: shared hit=2
         ->  Index Only Scan using idx_artifact_project_id_digest on artifact a  (cost=0.56..0.72 rows=1 width=72) (actual time=0.022..0.022 rows=1 loops=2)
               Index Cond: ((project_id = 6) AND (digest = (b.digest_af)::text))
               Heap Fetches: 0
               Buffers: shared hit=9
 Planning Time: 1.960 ms
 Execution Time: 0.117 ms
```

**Key observations:**
- `LIMIT` causes the planner to switch to a **Nested Loop** with an **Index Only Scan** on `idx_artifact_project_id_digest`
- Only **59 rows scanned** (vs 62 million)
- Buffer hits dropped from **1,976,626 to 11**
- Execution time: **0.117 ms** (vs 97,267 ms) — ~831,000x faster per blob

**Versions:**

- harbor version: v2.11.0
- docker engine version: n/a
- docker-compose version: n/a

**Additional context:**

The root cause is in `src/pkg/blob/dao/dao.go` in `FindBlobsShouldUnassociatedWithProject`. In environments where many images share common layers, certain blobs end up associated with almost every image in the registry. In our case, the base Go image pulled in the first line of every service's Dockerfile (`FROM golang:...`) produced a blob shared across thousands of artifacts. The query planner's statistics reflect that these digests match a very high fraction of rows in `artifact_blob`, making the selectivity too low for an index scan to be considered worthwhile. As a result, the planner opts for a full hash join instead.

A fix is proposed in PR #22994 — replacing the `IN (...)` query with batched `UNION` queries using `LIMIT 1` per blob. By targeting one digest at a time with `LIMIT 1`, the planner switches to a Nested Loop + Index Only Scan, reducing the per-blob query time from ~97s to ~0.1ms.

paragdudeja:
### Background

We discovered this issue while investigating slow delete API responses in our system. Investigating slow DB queries pointed to `FindBlobsShouldUnassociatedWithProject` as the root cause.

Our projects are heavily Go-based, which means most images share a common set of Go runtime/stdlib layers. These hot blobs end up associated with almost every image in the registry. The query planner's statistics reflect that these digests match a very high fraction of rows in `artifact_blob`, making selectivity too low for an index scan to be considered worthwhile — so the planner opts for a full hash join instead.

**Before this fix:** the query was taking ~4 minutes.
**After this fix:** the same query completes in ~40 milliseconds.

**Evidence (full query plans in issue #22995):**
- Before: Parallel Hash Join, 62M rows scanned, 1.9M buffer hits, ~97s per blob
- After: Nested Loop + Index Only Scan, 59 rows scanned, 11 buffer hits, ~0.1ms per blob

---

### What changed

`FindBlobsShouldUnassociatedWithProject` previously used a single `SELECT ... WHERE digest_blob IN (?, ?, ...)` query. This has two problems:

1. **Full hash join due to low selectivity:** hot blobs shared across many images cause the query planner to bypass the index on `digest_blob` and perform a full hash join on `artifact_blob` instead.
2. **Unbounded query size:** during GC, hundreds or thousands of blobs may be checked in one call, generating a very large `IN` list.

This PR uses a `VALUES + EXISTS` query:

```sql
SELECT v.digest_blob
FROM (VALUES (?), (?), ...) AS v(digest_blob)
WHERE EXISTS (
  SELECT 1 FROM artifact a
  JOIN artifact_blob b ON a.digest = b.digest_af
  WHERE a.project_id = ? AND b.digest_blob = v.digest_blob
)
```

**Benefits:**
- Single round trip for all blobs — no batching overhead
- Near-zero planning time
- `EXISTS` short-circuits on the first match, giving high selectivity per blob and allowing the index to be used
- Simpler, cleaner SQL

Fixes #22995

### DIFF
--- a/src/pkg/blob/dao/dao.go
+++ b/src/pkg/blob/dao/dao.go
@@ -17,6 +17,7 @@ package dao
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/docker/distribution/manifest/schema2"
@@ -246,21 +247,35 @@ func (d *dao) FindBlobsShouldUnassociatedWithProject(ctx context.Context, projec
 		return nil, err
 	}
 
-	sql := `SELECT b.digest_blob FROM artifact a, artifact_blob b WHERE a.digest = b.digest_af AND a.project_id = ? AND b.digest_blob IN (%s)`
-	params := []any{projectID}
-	for _, blob := range blobs {
-		params = append(params, blob.Digest)
-	}
-
-	var digests []string
-	_, err = o.Raw(fmt.Sprintf(sql, orm.ParamPlaceholderForIn(len(blobs))), params...).QueryRows(&digests)
-	if err != nil {
-		return nil, err
-	}
+	// blobBatchSize caps how many blobs are checked per UNION query to avoid
+	// generating excessively large SQL statements.
+	const blobBatchSize = 100
 
 	shouldAssociated := map[string]bool{}
-	for _, digest := range digests {
-		shouldAssociated[digest] = true
+	for i := 0; i < len(blobs); i += blobBatchSize {
+		end := i + blobBatchSize
+		if end > len(blobs) {
+			end = len(blobs)
+		}
+		batch := blobs[i:end]
+
+		// Each subquery uses LIMIT 1 so it short-circuits as soon as one
+		// matching artifact is found, avoiding full scans per blob.
+		var unionParts []string
+		var params []interface{}
+		for _, blob := range batch {
+			unionParts = append(unionParts, `(SELECT b.digest_blob FROM artifact a, artifact_blob b WHERE a.digest = b.digest_af AND a.project_id = ? AND b.digest_blob = ? LIMIT 1)`)
+			params = append(params, projectID, blob.Digest)
+		}
+
+		var digests []string
+		_, err = o.Raw(strings.Join(unionParts, " UNION "), params...).QueryRows(&digests)
+		if err != nil {
+			return nil, err
+		}
+		for _, digest := range digests {
+			shouldAssociated[digest] = true
+		}
 	}
 
 	var results []*models.Blob
--- a/src/pkg/blob/dao/dao.go
+++ b/src/pkg/blob/dao/dao.go
@@ -247,35 +247,33 @@ func (d *dao) FindBlobsShouldUnassociatedWithProject(ctx context.Context, projec
 		return nil, err
 	}
 
-	// blobBatchSize caps how many blobs are checked per UNION query to avoid
-	// generating excessively large SQL statements.
-	const blobBatchSize = 100
+	// Build a VALUES list for all blob digests and use EXISTS to check
+	// association. EXISTS short-circuits on the first match, and targeting
+	// one digest at a time gives the planner high selectivity to use the index.
+	valueParts := make([]string, len(blobs))
+	params := make([]interface{}, len(blobs)+1)
+	for i, blob := range blobs {
+		valueParts[i] = "(?)"
+		params[i] = blob.Digest
+	}
+	params[len(blobs)] = projectID
+
+	sql := `SELECT v.digest_blob FROM (VALUES ` + strings.Join(valueParts, ", ") + `) AS v(digest_blob)
+WHERE EXISTS (
+  SELECT 1 FROM artifact a
+  JOIN artifact_blob b ON a.digest = b.digest_af
+  WHERE a.project_id = ? AND b.digest_blob = v.digest_blob
+)`
+
+	var digests []string
+	_, err = o.Raw(sql, params...).QueryRows(&digests)
+	if err != nil {
+		return nil, err
+	}
 
 	shouldAssociated := map[string]bool{}
-	for i := 0; i < len(blobs); i += blobBatchSize {
-		end := i + blobBatchSize
-		if end > len(blobs) {
-			end = len(blobs)
-		}
-		batch := blobs[i:end]
-
-		// Each subquery uses LIMIT 1 so it short-circuits as soon as one
-		// matching artifact is found, avoiding full scans per blob.
-		var unionParts []string
-		var params []interface{}
-		for _, blob := range batch {
-			unionParts = append(unionParts, `(SELECT b.digest_blob FROM artifact a, artifact_blob b WHERE a.digest = b.digest_af AND a.project_id = ? AND b.digest_blob = ? LIMIT 1)`)
-			params = append(params, projectID, blob.Digest)
-		}
-
-		var digests []string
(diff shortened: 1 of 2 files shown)


paragdudeja:
The `APITEST_LDAP` failure (`Test Case - Run LDAP Group Related API Test`, `1 != 0`) appears to be unrelated to the changes in this PR — the same test passes on the latest main CI run. Could a maintainer retrigger the failed job? Thanks!

themavik (commented):
Batched UNION + LIMIT 1 per blob should kill the giant IN plan. nit: 100 subqueries per round trip still grows with N; if GC hits tens of thousands of blobs, total query count might be worth monitoring.

paragdudeja:
The `APITEST_DB` failure is in `Test Case - Replication From Dockerhub` — the CI runner appears unable to reach Docker Hub during the test. This is unrelated to the blob DAO changes in this PR. Could a maintainer retrigger the CI? Thanks!

wy65701436:
cc @chlins to take a look at this performance turning.

paragdudeja:
Updated the implementation to use `VALUES + EXISTS` as discussed with @chlins. See latest commit.

wy65701436:
Thanks @paragdudeja for your contribution. I discussed it offline with @chlins, and the code changes LGTM.

However, I do have some concerns about potential regressions. Since GC is a sensitive operation, we need to make sure there’s no impact or regression introduced.

Could you work with @chlins to validate the changes from an end-to-end perspective?

paragdudeja:
Thanks @wy65701436 and @chlins!

Happy to work with @chlins on E2E validation. A few data points that may help:

1. **Production validation:** We have been running this fix on our Harbor instance (v2.11.0, ~3M artifacts) and GC has completed successfully — blobs are correctly identified for unassociation with no observed regressions.

2. **Unit test coverage:** The existing `TestFindBlobsShouldUnassociatedWithProject` test verifies correctness (associated blobs excluded, unassociated ones returned), and `TestFindBlobsShouldUnassociatedWithProjectBatching` covers a mixed set of 110 blobs. Both pass on the PR CI.

Happy to run any additional E2E scenarios @chlins has in mind, or provide more details about our production validation. Let me know how you'd like to proceed.

chlins (approved):
lgtm

wy65701436 (approved):
lgtm

paragdudeja:
Thank you both @chlins and @wy65701436! Could someone please trigger the CI so we can get this merged? Thanks!

PR #23107, opened 2026-04-14
fix: add nil guard in auditext manager Create function

ipsitapp8:

Added a nil check at the top of the `Create` function in `src/pkg/auditext/manager.go`.
If `audit` is nil, the function now returns early with `0, nil` instead of panicking.
Also added a unit test to cover the nil case.


ipsitapp8:
Could a maintainer please add a release-note/bug label?

wy65701436 (approved):
lgtm

src/pkg/auditext/manager.go
```diff
@@ -77,6 +77,9 @@ func (m *manager) Get(ctx context.Context, id int64) (*model.AuditLogExt, error)
 
 // Create ...
 func (m *manager) Create(ctx context.Context, audit *model.AuditLogExt) (int64, error) {
+	if audit == nil {
```
chlins, line 80: In what scenarios would this audit be nil?

ipsitapp8: I added this as a defensive check in case `audit` ends up being nil from upstream (like unexpected caller behavior or future changes), just to avoid any panic.
I can remove it if you'd prefer enforcing non-nil at the caller instead.

chlins (approved):
lgtm

PR #23152, opened 2026-04-21
Add unscanned artifact policy to project settings

### LINKED ISSUE #16218: missing vulnerability visibiltity and broken "Prevent vulnerable images from running" which is a security issue
Hi
i was trying to get myself a picture about security scan handling and processes around those and so i've set up a harbor 2.4.1. (config pretty standard --with-trivy ) and i hope i did not miss anything...

as posted in slack i think the processes in harbor to handle security vulnerability findings (in particular "Prevent vulnerable images from running.") and the general visibility is missing or even broken:

**broken security concept of "Prevent vulnerable images from running" :**
* images that are not scanned (whatever reason "Unsupported" has) are not blocked from pulling like those that were scanned. if i just want "secure" images (no vulnerabilities above a certain level), an unscanned image is simply not secure and should not be able to pull!
* when used as proxy-cache i get the image before its scanned and possibly not allowed to pull.

so the mechanism to prevent insecure images to go into running state is broken/insecure/pointless.

**missing visibility:**
* there is no global/project dashboard about vulnerabilities findings?
* its not possible to create/schedule/mail/export/... any vulnerabilities report?
* there is no webhook/mail/... that is triggered when new vulnerabilities (of a specified level) are found?
* there is no integration into kubernetes/openshift that would add the missing visibility there (like the quay security operator tries to do (but fails))?

so no visibility at all. it's not possible to give harbor access to the security department as they are not able to use it like this.  all the integrated security scannings are pointless as the results can only be found by clicking through everything.

i know i've covered lots of topics here, but i just want to open discussion about those, that are very important to my organisation and i think in general, as the whole security mechanisms seem a little pointless if implemented like this. i run quay (which is even more broken and incomplete) and would love to use harbor (cncf graduated and oss), but with those findings above its impossible to justify a migration.

cheers.ivo

czifrushanna:
This PR separates the unscanned artifact prevention setting from the vulnerability prevention setting and adds the corresponding UI and translations in the project configuration page.

- Added a dedicated project policy option to prevent unscanned artifacts from running.
- Updated pull-prevention logic so vulnerability severity checks only run when prevent_vul is enabled.
- Kept unscanned enforcement independent, so teams can choose:
   - block unscanned artifacts only,
   - block vulnerable artifacts only,
   - or enforce both.
- Added compatibility normalization for legacy severity = none:
   - normalized to prevent_vul = true, prevent_unscanned = true, severity = low
   - applied in API/model handling and UI policy loading path
   - ensures older projects remain editable and safe without breaking behavior.
- Updated UI policy behavior and tests to align with the migration path and prevent regressions.

Fixes #16218


src/portal/src/app/base/project/project-config/project-policy-config/project-policy-config.component.ts
```diff
@@ -134,7 +138,6 @@ export class ProjectPolicyConfigComponent implements OnInit {
         { severity: 'high', severityLevel: 'VULNERABILITY.SEVERITY.HIGH' },
         { severity: 'medium', severityLevel: 'VULNERABILITY.SEVERITY.MEDIUM' },
         { severity: 'low', severityLevel: 'VULNERABILITY.SEVERITY.LOW' },
```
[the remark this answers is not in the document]
czifrushanna: I added migration instead of having none, so if a project was configured to prevent vulnerabilities with severity "none", now it will be configured with severity "low" and prevent unscanned images will be true.

src/server/middleware/vulnerable/vulnerable_test.go
```diff
@@ -279,6 +280,37 @@ func (suite *MiddlewareTestSuite) TestArtifactScanFailed() {
...
+	mock.OnAnything(suite.checker, "IsScannable").Return(true, nil)
+	mock.OnAnything(suite.scanController, "GetVulnerable").Return(&scan.Vulnerable{ScanStatus: "Error"}, nil)
+	mock.OnAnything(suite.accessMgr, "List").Return([]accessorymodel.Accessory{}, nil)
+
+	req := suite.makeRequest()
+	rr := httptest.NewRecorder()
+
 	Middleware()(suite.next).ServeHTTP(rr, req)
 	suite.Equal(rr.Code, http.StatusPreconditionFailed)
 }
```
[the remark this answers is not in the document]
czifrushanna: I added the test case

api/v2.0/swagger.yaml
```diff
@@ -7320,6 +7320,10 @@ definitions:
         type: string
         description: 'Whether prevent the vulnerable images from running. The valid values are "true", "false".'
         x-nullable: true
+      prevent_unscanned:
```
wy65701436, line 7323: `prevent_vul=true` blocks unscanned and scan-failed artifacts. After this PR, an upgraded project with `prevent_vul=true` and no `prevent_unscanned` key will **start allowing unscanned images to be pulled** — with no admin action and no notice.

src/server/v2.0/handler/model/project.go
```diff
@@ -26,6 +26,32 @@ import (
...
 
+const (
+	metadataTrue       = "true"
+	normalizedSeverity = "low"
+	legacyNoneSeverity = "none"
+)
+
+// NormalizeLegacySeverityPolicy rewrites legacy severity=none to the equivalent
+// supported policy combination.
+func NormalizeLegacySeverityPolicy(md *models.ProjectMetadata) {
```
wy65701436, line 39: NormalizeLegacySeverityPolicy is only applied at the API layer. The enforcement path in `src/server/middleware/vulnerable/vulnerable.go` reads the raw project metadata directly and never sees the normalized values.

Correct me if im wrong.

czifrushanna: Thanks for flagging these issues! I fixed both in the latest commits using a migration backfill, it should be good now. Let me know if you find any other problems.

PR #23155, opened 2026-04-22
Fix tag overlap in artifact list by enabling natural wrapping #23110

### LINKED ISSUE #23110: UI: tag items overlap when there are many tags
On the artifact tab page, when an image has a large number of tags, the UI layout breaks and some tag items overlap with each other. This makes the tag list hard to read and affects usability.

Image

ipsitapp8:
Comprehensive Summary of your change
This PR resolves a visual bug in the Artifact List where multiple tags would overlap and become unreadable.

Layout Refactor: Introduced a localized .tags-wrapper flex container in the SCSS to replace restrictive overflow constraints.
Natural Wrapping: Enabled flex-wrap: wrap and natural white-space handling for tag strings, allowing them to flow onto multiple lines within the datagrid cell.
Minimal Footprint: The fix preserves the existing tagsString() rendering logic and the tag count/tooltip functionality, ensuring zero regression in component behavior while fixing the overlap issue.
Aesthetics: Added a 4px gap between individual wrapping elements to maintain Harbor's clean UI density.
Issue being fixed Fixes #23110

Checklist:

 Well Written Title and Summary of the PR
 Label the PR as needed (release-note/enhancement)
 Accepted the DCO (Please ensure you've signed your commits if required)
 Made sure tests are passing
 Considered the docs impact


ipsitapp8:
Hi @goharbor/maintainers,
The workflow Release Note Label Check is failing because I do not have permission to add the required release-note/* label to my PR ([see the failing job here](https://github.com/goharbor/harbor/actions/runs/24762811197/job/72450144171?pr=23155)).
Could a maintainer please review and add the appropriate release-note label so the check can pass? Thank you!

AllForNothing:
@ipsitapp8 Thanks for the fix! I'd like to suggest an alternative approach that more precisely addresses the root cause.

Issue with the current approach
Setting overflow: visible and height: auto on the outer container un-clips the cell, which means long tag strings can overflow into adjacent columns. And since the inner <span> still uses a fixed width: 75%, the (N) count badge still overlaps the tail of the tag text — the original visual bug isn't fully resolved.

Suggested fix

you can refer to this https://github.com/AllForNothing/harbor/commit/91e5e003d5c48d62cde595b2a5b22289f9052a98

ipsitapp8:
@wy65701436 please check this pr

Issue #10000, opened 2019-11-26
using empty dict to provide pw_dict


tongying123456:
build by sourcecode, warning using empty dict to provide pw_dict, how to resolve it?

reasonerjt:
@tongying123456 please provide more context and output.

PR #23164, opened 2026-04-22
fix(log): remove sudo dependency from harbor-log startup

### LINKED ISSUE #23156: harbor-log fails to start on v2.15.0 because start.sh uses `sudo -u #10000 rsyslogd -n`, which fails with PAM account management error
## Description

  On a fresh Harbor v2.15.0 installation, `harbor-log` keeps restarting and the rest of the Harbor stack cannot start completely.

  The direct cause is that the `harbor-log` container's startup script runs:

 ```sh
sudo -u #10000 -E rsyslogd -n
```

  In my environment this fails inside the official goharbor/harbor-log:v2.15.0 image with:

  sudo: PAM account management error: Authentication service cannot retrieve authentication info
  sudo: a password is required

  Because harbor-log never becomes healthy, other Harbor services fail to start with errors like:

  failed to create task for container: failed to initialize logging driver: dial tcp [::1]:1514: connect: connection refused

  This looks related in outward symptom to #20949, but the root cause here is different: harbor-log itself is crashing due to sudo/PAM inside the
  container.

  ## Expected behavior

  harbor-log should start normally on a fresh install, and the rest of Harbor should come up without requiring any manual compose override.

  ## Actual behavior

  harbor-log restarts continuously with exit code 1, printing:

  sudo: PAM account management error: Authentication service cannot retrieve authentication info
  sudo: a password is required

  and Harbor installation/startup remains incomplete.

  ## Versions

  - Harbor version: v2.15.0
  - Image: goharbor/harbor-log:v2.15.0
  - Docker engine: 29.4.0
  - Docker Compose: v5.1.1
  - Host OS: Linux x86_64

  ## Steps to reproduce

  1. Extract the official Harbor v2.15.0 offline installer.
  2. Configure harbor.yml.
  3. Run:

  bash install.sh

  4. Check container status:

 ```bash
docker ps -a
```

  5. Check harbor-log logs:

  docker logs harbor-log

  ## What I observed

  harbor-log is created with the official image and default command:
```bash
/bin/sh -c /usr/local/bin/start.sh
```
  The startup script in Harbor source is:

```bash
#!/bin/bash
set -e
chown -R 10000:10000 /var/log/docker
crond
rm -f /var/run/rsyslogd.pid
sudo -u #10000 -E rsyslogd -n
set +e
```
  Source:

  - https://raw.githubusercontent.com/goharbor/harbor/v2.15.0/make/photon/log/start.sh

  In my environment, running the same image directly reproduces the failure:

```bash
docker run --rm --entrypoint sh goharbor/harbor-log:v2.15.0 -c 'sudo -u \#10000 -E id'
```

  Output:

  sudo: PAM account management error: Authentication service cannot retrieve authentication info
  sudo: a password is required

  ## Workaround

  I was able to make Harbor start successfully by overriding the log service command and bypassing sudo, for example:
```bash
/bin/sh -c 'chown -R 10000:10000 /var/log/docker && crond && rm -f /var/run/rsyslogd.pid && exec rsyslogd -n'
```

  After doing that, harbor-log became healthy and the entire Harbor stack came up successfully.

  ## Question

  Is this a known compatibility issue between the current Photon base image / PAM / sudo behavior and the harbor-log startup script?

  If so, would Harbor consider replacing this sudo-based user switch in harbor-log with a more container-friendly approach?

ipsitapp8:

## Summary

This PR fixes a `harbor-log` crash loop on `v2.15.0` by removing the failing `sudo -u #10000` startup path.

In the current container environment, that `sudo` call can fail with PAM/account-management errors before `rsyslogd` is launched. When that happens, `harbor-log` never becomes healthy and other Harbor services fail because logging is unavailable.

## What changed

* replace `sudo -u #10000 -E rsyslogd -n` with `exec rsyslogd -n` in `make/photon/log/start.sh`
* replace the same `sudo`-based execution path in `make/photon/log/logrotate` with direct execution
* remove the now-dead `set +e` line after `exec`
* update the outdated `logrotate` comment to match the new behavior

## Why

This removes the failing `sudo`/PAM dependency from the `harbor-log` startup path while keeping the existing control flow otherwise unchanged. The issue report also confirmed that bypassing the `sudo` path allows Harbor to start successfully.

## Notes

This changes `rsyslogd` and `logrotate` to run as the container's default user instead of UID `10000`, so there is a behavior change here. However, it is limited to the `harbor-log` container path and avoids the current failure mode where the container cannot start at all.

## Verification

* build the updated `harbor-log` image
* verify `harbor-log` stays up instead of crash-looping
* verify Harbor services start normally once logging is available
* run `/etc/cron.hourly/logrotate` in the container and confirm it completes without `sudo`/PAM errors
* confirm logs under `/var/log/docker` remain writable and rotation still works

# Issue being fixed
Fixes #23156

/label ~"release-note/bug-fix"

* [done] Well Written Title and Summary of the PR
* [done] Label the PR as needed (`release-note/bug-fix`)
* [done] Accepted the DCO (all commits signed with `-s`)
* [done] Made sure tests are passing (no new tests required for this minimal script fix)
* [done] Considered the docs impact (no documentation changes needed for this fix)

### DIFF
--- a/make/photon/log/logrotate
+++ b/make/photon/log/logrotate
@@ -1,7 +1,7 @@
 #!/bin/sh
 
-# run the logrotate with user 10000, the state file "/var/lib/logrotate/logrotate.status"
-# is specified to avoid the permission error
+# use the explicit state file "/var/lib/logrotate/logrotate.status"
+# to avoid permission issues when rotating logs
 cd /
-sudo -u \#10000 -E /usr/sbin/logrotate -s /var/lib/logrotate/logrotate.status /etc/logrotate.conf
+/usr/sbin/logrotate -s /var/lib/logrotate/logrotate.status /etc/logrotate.conf
 exit 0
--- a/make/photon/log/start.sh
+++ b/make/photon/log/start.sh
@@ -3,5 +3,5 @@ set -e
 chown -R 10000:10000 /var/log/docker
 crond
 rm -f /var/run/rsyslogd.pid
-sudo -u \#10000 -E  'rsyslogd' '-n'
-set +e
+exec rsyslogd -n
+


ipsitapp8:
@Vad1mo , @chlins , @OrlinVasilev 
Just a gentle follow-up on this PR.
I’ve addressed all feedback and verified the changes locally.
Would really appreciate a quick review whenever you get time. Thanks!

ipsitapp8:
Hi @Vad1mo ,
Just a gentle follow-up 🙂
CI is green except for the label requirement, which I believe needs maintainer access.
Could you please add the appropriate label so the checks pass?

Thanks in advance!

stonezdj:
@ipsitapp8   please avoid update the UI code in this PR.

ipsitapp8:
@stonezdj, Thanks for pointing this out!

I’ll remove the UI changes from this PR and keep it focused on the harbor-log fix.
Will open a separate PR for the UI update.

Also, could you please help add the release-note/bug-fix label? I don’t have permission.

Thanks!

PR #23174, opened 2026-04-26
fix: use native clipboard API for artifact link copy

### LINKED ISSUE #23115: Copy Artifact link is not working.
Hello,

It seems that in v2.15 there is an issue while trying to copy the artifact link from Harbor UI.

Image

Steps to reproduce:
1) Go to Harbor
2) Navigate to the image repository
3) Select any image
4) Click on Artifacts
5) Click on the Copy icon 

Not sure if it is related to https://github.com/goharbor/harbor/pull/22136

Best regards,
George

varsha-0007:
Fixes #23115

Replaced `ngx-clipboard` with the browser's native `navigator.clipboard.writeText()` API, which is more reliable and doesn't fail silently.

For cases where the page isn't running in a secure context (HTTP), added a fallback that creates a temporary hidden textarea, selects the text, and attempts the copy that way. `execCommand` is deprecated but only used here as a last resort — open to a better suggestion if maintainers prefer.


stonezdj:
@bupd can you please take a look?

bupd (changes requested):
@varsha-0007 Thanks for your contribution 

I would consider this more of a fallback - I prefer https://material.angular.dev/cdk/clipboard/overview  - using official angular clipboard over this - since much concise and its pretty much the same for all browsers.

varsha-0007:
hey @bupd,
could you please look at pr now? I have made the changes as you requested

package.json
bupd, line 1: Sorry but, this is the wrong way. We should not have multiple package.json files.

I cant merge this PR, Since it touches other parts which will be handled in the Angular Migration.

bupd (changes requested):
Thanks for the contribution. 

But this PR is heading in the wrong direction. If we need to package.json in order for copytoclipboard to work then we should not do this in this PR. and handle it in the Angular Migration task.

bupd:
Thanks again for taking the time to work on this. I really appreciate the contribution.

Although, I agree with the problem this is trying to solve, but I think this PR is a bit heavier than what we need here, and it adds new dependencies that I’d rather avoid for this fix.

I’m going to close this PR for now.

PR #22684, opened 2025-12-25
fix(i18n): resolve missing keys and localize max upstream connection placeholder

Tusharjamdade:
Thank you for contributing to Harbor!

# Comprehensive Summary of your change
Addresses i18n issues in the max upstream connection configuration by adding missing translations and replacing a hardcoded placeholder.

# Issue being fixed
Fixes #22677

Please indicate you've done the following:
- [x] Well Written Title and Summary of the PR
- [x] Label the PR as needed. "release-note/ignore-for-release, release-note/new-feature, release-note/update, release-note/enhancement, release-note/community, release-note/breaking-change, release-note/docs, release-note/infra, release-note/deprecation"
- [x] Accepted the DCO. Commits without the DCO will delay acceptance.
- [x] Made sure tests are passing and test coverage is added if needed.
- [x] Considered the docs impact and opened a new docs issue or PR with docs changes if needed in [website repository](https://github.com/goharbor/website).

PR #23197, opened 2026-05-05
fix(i18n): localize max upstream connection placeholder

bupd:
The author of #22684 has been inactive, and this fix seems valuable for Harbor. As a member of the Harbor community, I am adopting this fix and willing to steward it into Harbor v2.16.0.

Thanks.


Vad1mo (approved):

chlins (approved):
lgtm

Issue #13020, opened 2020-09-09
First and last name from OIDC is set to username


Natanande:
During automatic onboarding using OIDC, the Realname is set to the username as seen [here](https://github.com/goharbor/harbor/blob/master/src/core/controllers/oidc.go#L198).

At my company, we set the username claim to "email", because Harbor requires the username to be unique. This results in the profile showing the email as firstname and last name, which doesn't make sense.

It would be nice to be able to set the Realname attribute to a specified claim (or a combination of multiple ones? given_name + family_name), the same way as we now can [with the username and group](https://github.com/goharbor/harbor/blob/master/src/common/utils/oidc/helper.go#L313).

This is mostly cosmetic, but it does look ridiculous when all three fields (username, email and realname) in the profile all are set to the user's email.

(Even better would be to dynamically update the realname at login/some interval, since that is something that can change)

steven-zou:
cc @reasonerjt

reasonerjt:
IMO to fix this we may allow user to set more attributes in the onboard dialog, which is a common practice.

In addition to that we may consider allow user to update his profile partially after he's onboarded.

Natanande:
Agreed, but since we use the "Automatic onboarding" without any onboard dialog, we would prefer specifying a claim in the oidc JWT to automatically populate the profile. For us all user information is already stored centrally in Azure.

kautkata:
I stumbled on this issue while trying to set up following authentication/authorization chain Harbor-2.1.1 -> Keycloak-10.0.1 -> MS AD.

From my perspective it would be enough to use value of "name" attribute returned by userinfo API.

schuerg:
The OpenID Connect specification defines a set of standard claims for things like the first name, last name and so on.  
See: <https://openid.net/specs/openid-connect-basic-1_0.html#StandardClaims>

I think these claims (`preferred_username`, `given_name`, `family_name`) should be used if no other configuration is proivided to harbor.

keliansb:
Issue still relevant, please do not close.

lusu007:
Any news on this? In my opinion, it would be quite helpful if the `First and Lastname` field could be set automatically during onboarding

Craftoncu:
Still a problem

belfhi:
We also ran into this issue when configuring Keycloak as an OIDC Provider. Not a real problem but weird, especially since the Name is known and used as the username if `Username Claim` is not set.

pniebylski-zilch:
[...]

PR #23198, opened 2026-05-06
feat: allow configurable OIDC claim for user realname mapping

### LINKED ISSUE #23188: OIDC onboarding sets `realname` equal to resolved username instead of using a human-friendly claim
## Description                                                                                          
                                                                                                          
  We found what looks like a bug, or at least an incomplete OIDC onboarding implementation, in Harbor.    
                                                                                                          
  When an OIDC user is onboarded for the first time, Harbor appears to set:                               
                                                                                                          
  - `username` from the configured OIDC username source                                                   
  - `realname` to the exact same value as `username`                                                      
                                                                                                          
  This causes incorrect user display names in environments where the technical identifier and the human-  
  friendly name are intentionally different.                                                              
                                                                                                          
  ## Expected behavior                                                                                    
                                                                                                          
  Harbor should allow the user-facing display name (`realname`) to be populated from a human-friendly OIDC
  claim such as:                                                                                          
                                                                                                          
  - `name`                                                                                                
  - or another explicitly configurable claim                                                              
                                                                                                          
  while still allowing `username` to come from a technical identifier claim such as:                      
                                                                                                          
  - `userid`                                                                                              
  - `preferred_username`                                                                                  
                                                                                                          
  In our environment, this separation is necessary and intentional:                                       
                                                                                                          
  - technical identifier: unique and stable                                                               
  - human-friendly name: readable by users and admins                                                     
                                                                                                          
  ## Actual behavior                                                                                      
                                                                                                          
  Harbor uses the resolved onboarding username for both fields.                                           
                                                                                                          
  As a result, after OIDC onboarding, the Harbor local user record becomes something like:                
                                                                                                          
  ```json                                                                                                 
  {                                                                                                       
    "username": "XXXXXXXXXXX",                                                                       
    "realname": "xxxxxxxxxxxxxxx",                                                                       
    "email": "XXXXXXXXXXXXX"                                                           
  }                                                                                                       
                                                                                                          
  or, depending on configuration, it would become:                                                        
                                                                                                          
  - username = userid                                                                                     
  - realname = userid                                                                                     
                                                                                                          
  or:                                                                                                     
                                                                                                          
  - username = preferred_username                                                                         
  - realname = preferred_username                                                                         
                                                                                                          
  In all cases, realname is not the actual human-friendly full name.                                      
                                                                                                          
  ## Why this is a problem                                                                                
                                                                                                          
  OIDC commonly carries different kinds of identity attributes:

  - technical identifier
  - login name
  - human-friendly full name
  - first name
  - last name

  Harbor currently seems to collapse at least two of these concerns into one field during onboarding.

  That makes OIDC onboarding unsuitable for environments where:

  - username must remain technical and stable
  - realname must remain human-readable


  ## Evidence from behavior

  Our OIDC token/userinfo contains values like:

  {
    "name": "XXXXXXXXXXXX",
    "preferred_username": "XXXXX",
    "userid": "xXXXXXXXXXXX",
    "email": "XXXXXXXXXX"
  }

  But Harbor stores the onboarded user as:

  {
    "username": "XXXXXXXXXXX",
    "realname": "XXXXXXXXXXXX",
    "email": "XXXXXXXXXXXXX"
  }

  ## Source code observation

  The behavior appears to come from the OIDC onboarding flow.

  In src/core/controllers/oidc.go, Harbor creates the user like this:

  user := &models.User{
      Username: username,
      Realname: username,
      Email:    info.Email,
  }

  This means realname is always copied from the resolved onboarding username.

  Also, in src/pkg/oidc/helper.go, Harbor resolves the OIDC username from:

  1. configured oidc_user_claim
  2. userinfo username
  3. fallback name

  So even if the OIDC claim selection is changed, realname still follows username and cannot be
  independently mapped.

  ## Request

  Please consider one of these options:

  1. Add a dedicated configurable OIDC claim for Harbor realname
  2. Default realname to the OIDC name claim when available
  3. At minimum, document clearly that Harbor intentionally sets realname = username for OIDC onboarding

  ## Environment

  - Harbor version: v2.15.0
  - Auth mode: oidc_auth
  - OIDC provider: Keycloak

  ## Notes

  This is not just about cosmetic display.
  In real environments, username and realname serve different purposes and should not be conflated.

sakkshm:

This PR introduces the ability to map a specific OIDC claim to the Harbor `realname` (display name) field during user onboarding. Previously, Harbor hardcoded the `realname` to be identical to the resolved `username`, which caused issues in environments where technical identifiers (like UIDs) and human-friendly names are distinct.

**Key Changes:**
*   **Configuration:** Added a new configuration setting `oidc_realname_claim` to the OIDC settings.
*   **OIDC Helper:** Updated the `UserInfo` struct and extraction logic in `src/pkg/oidc/helper.go` to pull the configured claim from both ID Tokens and the UserInfo endpoint.
*   **Onboarding Logic:** Modified `src/core/controllers/oidc.go` to use the extracted realname during user creation, with a backward-compatible fallback to the username if no claim is configured or found.
*   **Persistence:** Registered the new configuration field in the metadata list to allow persistence in the database and accessibility via the Configuration API.
*   **Testing:** Added unit tests to verify claim extraction and merging priorities.

# Issue being fixed
Fixes #23188 & #13020

- [x] Label the PR as needed. (Label: `release-note/new-feature`)
- [ ] Considered the docs impact and opened a new docs issue or PR with docs changes if needed in [[website repository](https://github.com/goharbor/website)](https://github.com/goharbor/website).


sakkshm:
I've updated the PR to address the feedback from CopilotAI:

* Fixed a bug in the configuration tests where the realname claim wasn't being validated correctly.
* Updated the OIDC helper tests to properly account for the fallback logic (Claim -> Name -> Username).
* Fixed formatting and indentation issues in `src/pkg/oidc/helper.go`.
* Rebased with the latest main branch.

Ready for another look.

src/pkg/oidc/helper.go
```diff
@@ -391,6 +401,25 @@ func userInfoFromClaims(c claimsProvider, setting cfgModels.OIDCSetting) (*UserI
...
+		if realname, ok := allClaims[setting.RealNameClaim].(string); ok {
+			// Extracting real name from RealNameClaim
+			res.Realname = realname
+		} else {
+			log.Warningf("OIDC. Failed to recover Realname from claim. Claim '%s' is invalid or not a string", setting.RealNameClaim)
+		}
+	}
+
+	if res.Realname == "" {
+		if name, ok := allClaims["name"].(string); ok && name != "" {
```
chlins, line 415: Could you explain why, if the real name is empty, we first try to extract it from the `name` claim and then fall back to the username? Could you provide more context?

wy65701436: same here

sakkshm: If we fail to extract the real name from the `RealNameClaim`, it falls back to the `name` claim, which the [OIDC Spec](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) defines as the standard for an "End-User's full name in displayable form." As a final safety net, it uses the `Username`. 

The hierarchy is Explicit Config > Protocol Standard > Username, it guarantees a human-readable identifier is always available, even if the Identity Provider is restrictive or misconfigured.

PR #23223, opened 2026-05-09
fix(portal): remove temporary SBOM permission override

### LINKED ISSUE #23218: Remove temporary SBOM permission override in ArtifactListPageService
## Description

In `artifact-list-page.service.ts`, _hasSbomPermission is already populated using the result returned from:

```ts

hasProjectPermissions(projectId, permissions)

```

However, the value is immediately overridden by a hardcoded assignment:

`this._hasSbomPermission = true;`

This bypasses the actual permission check and always enables SBOM permission in the UI regardless of the user's real project permissions.

## Expected Behavior

- SBOM permission should rely only on the result returned from `hasProjectPermissions(...)`

- Users without SBOM permission should not receive access through the UI

- The temporary hardcoded override should be removed

## Possible Fix

Remove:

```ts

this._hasSbomPermission = true;

```

and verify the permission flow with different project roles/users.

kimjune01:
Fixes #23218.

The static override `this._hasSbomPermission = true` in
`artifact-list-page.service.ts` bypasses RBAC permission checks,
granting all users SBOM access regardless of project role.

Remove the override so the actual permission result from
`userPermissionService.getPermission()` propagates correctly.


chlins:
@bupd Please take a look for this PR, thanks!

bupd (approved):
/lgtm

chlins (approved):
lgtm

wy65701436 (approved):
lgtm

PR #23225, opened 2026-05-11
fix: Bump repository update_time on tag and artifact changes

### LINKED ISSUE #23149: Repository update_time not updated on tag/digest changes
Currently in Harbor:

When a tag is pushed or deleted, the repository update_time is not updated
When a tag is pushed or deleted, the artifact (digest) update_time is not updated

Aloui-Ikram:
## Summary
     
  Fixes #23149.
     
  The repository `update_time` only advanced on docker push, so the Portal's "Last Modified Time" column stayed frozen when users added/deleted tags from the UI or deleted artifacts.

  ## Changes
  
  - Add `Touch()` on the repository manager (and its cached redis wrapper) to refresh `update_time`.
  - `tag.controller` touches the parent repository after `Create`, `Delete`, and `Ensure` rebind. Skipped on conflict so a losing push in a race doesn't falsely bump the timestamp.
  - `artifact.controller` touches the parent repository on artifact create and delete.

  ## Before
  
  "Last Modified Time" stuck on the v1 push even though v3 was pushed minutes later:

  before-repo-list

  before-artifact-detail
  
  ## After
  
  Timestamps advance on every push, tag change, and delete:

  after-repo-list

  after-artifact-detail
  
  ## Tests

  Unit tests added/updated for the new `Touch` paths in `pkg/repository`, `pkg/cached/repository/redis`, `controller/tag`, and `controller/artifact`.

  Verified locally that `update_time` bumps on the tag and artifact operations.


stonezdj:
Because this PR is going to cherry-pick to 2.15.x, and it already updated the migrate SQL, need to change the migrate SQL's file version for 2.15.x cherry-pick.

make/migrations/postgresql/0191_2.16.0_schema.up.sql
```diff
@@ -0,0 +1,4 @@
+/* Track when an artifact row last changed (e.g. tag attached/detached). See #23149. */
```
stonezdj: The file name should be 0190_2.16.0_schema.up.sql, and all operation in the sql should be idempotent.

Aloui-Ikram:
Hi @stonezdj, thanks for the review!
  I've addressed your feedback in the latest commit:
  - Removed the separate 0191_2.16.0_schema.up.sql file and merged its content
  into the existing 0190_2.16.0_schema.up.sql so the version stays consistent
  for the 2.15.x cherry-pick.
  - Made all the SQL statements idempotent: the column is created with IF NOT 
  EXISTS and its default is set at creation time, and the backfill uses
  COALESCE(push_time, CURRENT_TIMESTAMP) with a WHERE update_time IS NULL guard
  so re-runs are safe.
  Please take another look when you have a chance.

wy65701436:
Thanks @Aloui-Ikram for your contribution. I’ll target this for 2.16.

BTW, I may not have described the requirement clearly in my original description. What I’d like to see is that any operation within a specific repository — such as push, pull, or delete operations, including both artifacts and tags — should update the repository’s `update_time`.

I didn’t mean to add an `update_time` field for artifacts themselves. I don’t think there’s a clear scenario where the artifact `update_time` would be useful. I’ll update the description accordingly — sorry for the confusion.

Aloui-Ikram:
hi @wy65701436 , thanks for the clarification , 
to confirm : only the repostry's update_time should be bumped , on any content change inside it (tag and artifact create /delete /update/retag) . artifacts dont need their own update_time
so i ll update the PR to :
   - Drop the artifact update_time column, migration, swagger field, and artifact-side
  Touch().
  - Keep the repository Touch() and the tag controller calls.
  - Add repository Touch() calls in the artifact controller too.

stonezdj (approved):

make/migrations/postgresql/0190_2.16.0_schema.up.sql
```diff
@@ -1 +1,5 @@
 ALTER TABLE artifact_accessory ADD COLUMN IF NOT EXISTS source varchar(50) DEFAULT 'local';
+
+/* Track when an artifact row last changed (e.g. tag attached/detached). See #23149. */
+ALTER TABLE artifact ADD COLUMN IF NOT EXISTS update_time timestamp DEFAULT CURRENT_TIMESTAMP;
```
chlins: Per our latest discussion, I don’t think we need to make the database schema change.

chlins (changes requested):

wy65701436:
> hi @wy65701436 , thanks for the clarification , to confirm : only the repostry's update_time should be bumped , on any content change inside it (tag and artifact create /delete /update/retag) . artifacts dont need their own update_time so i ll update the PR to :
> 
>     * Drop the artifact update_time column, migration, swagger field, and artifact-side
>       Touch().
> 
>     * Keep the repository Touch() and the tag controller calls.
> 
>     * Add repository Touch() calls in the artifact controller too.

yes, please go ahead to have a update, thanks.

chlins:
@Aloui-Ikram Could you please update this PR based on the latest conclusion?

Aloui-Ikram:
Hi @wy65701436  @chlins ,
i've updated the PR to match agreed scope (repo-only update_time , no artifact filed , no swagger/migration changes ; force-pushed as a single commit , could you take a look when u have chance ? Thanks

src/controller/tag/controller.go
```diff
@@ -118,13 +126,29 @@ func (c *controller) Ensure(ctx context.Context, repositoryID, artifactID int64,
...
+	c.touchRepo(ctx, repositoryID)
 
 	return tagID, nil
 }
 
+// touchRepo bumps update_time on the parent repository so the "last modified"
+// timestamp reflects tag push/delete events. Errors are logged only; the tag
+// operation has already succeeded.
+func (c *controller) touchRepo(ctx context.Context, repositoryID int64) {
+	if c.repoMgr == nil {
```
chlins: This logic is not consistent with the `touchRepo` in artifact controller, please align to keep the same behavior. IMO we don't need to double-check for the `repoMgr` or `repositoryID`.

src/controller/artifact/controller.go
```diff
@@ -637,12 +656,27 @@ func (c *controller) AddLabel(ctx context.Context, artifactID int64, labelID int
 			}
 		}
 	}()
-	err = c.labelMgr.AddTo(ctx, labelID, artifactID)
+	if err = c.labelMgr.AddTo(ctx, labelID, artifactID); err != nil {
+		return
+	}
+	if c.artMgr != nil {
```
chlins: I don't think we need to check the nil for `artMgr` as it has been used before.

src/controller/artifact/controller.go
```diff
@@ -637,12 +656,27 @@ func (c *controller) AddLabel(ctx context.Context, artifactID int64, labelID int
...
+	}
 	return
 }
 
 func (c *controller) RemoveLabel(ctx context.Context, artifactID int64, labelID int64) error {
-	return c.labelMgr.RemoveFrom(ctx, labelID, artifactID)
+	if err := c.labelMgr.RemoveFrom(ctx, labelID, artifactID); err != nil {
+		return err
+	}
+	if c.artMgr != nil {
```
chlins: ditto.

src/controller/artifact/controller.go
```diff
@@ -637,12 +656,27 @@ func (c *controller) AddLabel(ctx context.Context, artifactID int64, labelID int
 			}
 		}
 	}()
-	err = c.labelMgr.AddTo(ctx, labelID, artifactID)
+	if err = c.labelMgr.AddTo(ctx, labelID, artifactID); err != nil {
+		return
+	}
+	if c.artMgr != nil {
+		if art, gerr := c.artMgr.Get(ctx, artifactID); gerr == nil {
```
chlins: Log the error.

src/controller/artifact/controller.go
```diff
@@ -637,12 +656,27 @@ func (c *controller) AddLabel(ctx context.Context, artifactID int64, labelID int
...
 	return
 }
 
 func (c *controller) RemoveLabel(ctx context.Context, artifactID int64, labelID int64) error {
-	return c.labelMgr.RemoveFrom(ctx, labelID, artifactID)
+	if err := c.labelMgr.RemoveFrom(ctx, labelID, artifactID); err != nil {
+		return err
+	}
+	if c.artMgr != nil {
+		if art, err := c.artMgr.Get(ctx, artifactID); err == nil {
```
chlins: ditto.

Aloui-Ikram:
Hi @chlins, thanks again for the review. I pushed the changes ,  removed the extra nil checks in both touchRepo functions and added error logging in AddLabel/RemoveLabel. Let me know if anything else needs to change.

chlins (approved):
lgtm

chlins:
@Aloui-Ikram Hi, as discussed offline with @wy65701436, let's keep this minimal and limit the changes to artifact/tag create, update, and delete operations. Please remove any label-related changes. And when user create/delete tag from the harbor UI, the repo update_time should be updated as well.

chlins (changes requested):
pending for update.

Aloui-Ikram:
Hi @chlins , @wy65701436, dropped the label changes.

Also noticed tag create from the UI wasn't bumping update_time, so i added touchRepo to tagCtl. Create (registry push was the only path wired up before).
Tested the tag and artifact operations locally and update_time bumps as expected.

chlins (approved):
lgtm

bupd (approved):
/lgtm

PR #23231, opened 2026-05-13
fix: block pull of unsupported artifacts when vulnerability prevention is on

### LINKED ISSUE #16218: missing vulnerability visibiltity and broken "Prevent vulnerable images from running" which is a security issue

jeiche:
# Summary
Previously, artifacts that were unsupported for vulnerability scanning (not scannable with no existing scan report) were allowed to be pulled even when the project had "Prevent vulnerable images from running" enabled. This created a security gap where unscanned and unscannable artifacts could bypass the vulnerability prevention policy.

Now, unsupported artifacts are blocked with a `PROJECTPOLICYVIOLATION` error, consistent with how other unscanned artifacts are treated.

#### `src/server/middleware/vulnerable/vulnerable.go`

- When a project has "Prevent vulnerable images from running" enabled, artifacts that are unsupported for vulnerability scanning (not scannable, with no existing scan report) were previously allowed to be pulled, bypassing the policy. Now they are blocked with a `PROJECTPOLICYVIOLATION` error and a message indicating the image is unsupported for scanning. Updated the corresponding comment to reflect the new behavior.

#### `src/server/middleware/vulnerable/vulnerable_test.go`

- Updated `TestNonScannerPulling` and `TestArtifactIsNotScannableNotScanned` to expect `HTTP 412 Precondition Failed` instead of `HTTP 200 OK`, matching the new blocking behavior for unsupported artifacts.

# Issues Fixed
Fixes #16218

# Additional Info
I think this PR should be grouped with (merged at the same time as) #23130.

### DIFF
--- a/src/server/middleware/vulnerable/vulnerable.go
+++ b/src/server/middleware/vulnerable/vulnerable.go
@@ -84,7 +84,8 @@ func Middleware() func(http.Handler) http.Handler {
 		vulnerable, err := scanController.GetVulnerable(ctx, art, allowlist, proj.CVEAllowlist.IsExpired())
 		if errors.IsNotFoundErr(err) {
 			// When the scanner is disconnected the artifact will be considered not scannable.
-			// We'll try to check the existing scan report even when it's not scannable, and only if there is no report, we will skip checking the vulnerability.
+			// We'll try to check the existing scan report even when it's not scannable, and if there is no report,
+			// we block the pull for both scannable artifacts (missing report) and unsupported artifacts.
 			checker := scanChecker()
 			scannable, err := checker.IsScannable(ctx, art)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


adrianipopescu:
my 2c is that this is still relevant

PR #23238, opened 2026-05-14
fix: build valkey from source and tag the version

danacr:
## Description
Fix the failing build where the valkey package installation was failing because the package was not found in the Photon repository for arm64, amd looks like its there.

The valkey package is available in EPEL (Extra Packages for Enterprise Linux) but requires the EPEL repository to be enabled first.

## Root Cause
The build was attempting to install `valkey` directly from Photon's default package repository, which doesn't include it for arm64. Valkey is available in EPEL but Photon didn't have EPEL enabled.

## Changes Made
- Updated `make/photon/valkey/Dockerfile.base` to install `epel-release` before installing valkey
- This enables access to EPEL packages in the Photon base image

## Testing
The change was tested on a branch based on the latest upstream `main`. The fix enables the Docker image build to successfully install valkey from EPEL.

## Related Issues
Fixes the failing GitHub Actions job: https://github.com/goharbor/harbor/actions/runs/25845578993/job/75939978452


danacr:
@Vad1mo @ranimandepudi

bupd (approved):
/lgtm

make/photon/valkey/Dockerfile.base
```diff
@@ -3,4 +3,4 @@ FROM photon:5.0
 RUN tdnf install -y shadow >> /dev/null \
     && groupadd -g 999 valkey \
     && useradd -u 999 -g 999 -c "Valkey Database Server" -d /var/lib/redis -s /sbin/nologin -m valkey
-RUN tdnf install -y valkey && tdnf clean all
+RUN tdnf install -y epel-release && tdnf install -y valkey && tdnf clean all
```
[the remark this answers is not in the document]
danacr: this should be fine as we should in any case pin the valkey version

jUDASmILE: Hi
Where the epel-release package comes from? Does it exist in public photon aarch64 repos? Could you share the url to the package? Also did you have tried to build it in x86 environment? I couldn't find the package in photon x86 repos, if it doesn't exist for x86 repo, I'm afraid the tdnf install gonna to exit with error and build process will be broken.

danacr: You're right, looks like it's more tricky than I thought to add expel. Let me just build valkey instead in this case https://github.com/vmware/photon/issues/481

danacr:
@ranimandepudi do you have a better idea on how to solve this?

ranimandepudi:
> @ranimandepudi do you have a better idea on how to solve this?

Hey @danacr - thanks for working on this, your root cause analysis is spot on. Photon doesnt ship valkey for `aarch64`, so building from source is the right approach.

I tested the multi-stage build on an arm64 instance and confirmed it works. A couple of findings:

- Add coreutils and gawk to the builder stage - OpenSSL's bundled configure needs `expr` and jemalloc's configure needs awk. Without them the build either loops on errors or fails.

- Use goharbor/photon:5.0 (not bare photon:5.0) - matches all other Harbor base images.

- No Makefile changes needed - the existing `_build_base` macro handles multi-stage Dockerfiles transparently (BuildKit is already enabled). The `ARG VALKEY_VERSION=9.1.0` default means Docker picks it up without explicit `--build-arg`. This keeps it as a single-file fix that's easier to review and carries less risk.

Tested on arm64:
valkey-server --version >> v=9.1.0 malloc=jemalloc-5.3.0 bits=64
SET/GET operations functional
Image size: ~99MB

I can push a commit to your branch with these fixes, or open a separate PR - whichever you prefer. Let me know how you would like to proceed!

danacr:
closed in favor of https://github.com/goharbor/harbor/pull/23282

PR #23254, opened 2026-05-19
Fix: Slack notification handler timestamp formatting

rossigee:
## Description

Fix Slack notification handler to display timestamps in a human-readable format instead of Slack's complex date markup that was causing display issues.

### Problem
The Slack handler was using Slack's date markup format `<!date^{.OccurAt}^{date} at {time}|...>` which failed to render correctly in many Slack clients.

### Solution
Replace with a simple formatted timestamp using Go's time.Unix().Format() to produce `YYYY-MM-DD HH:MM:SS UTC` format.

### Changes
- `src/pkg/notifier/handler/notification/slack_handler.go`: 
  - Add `time` import
  - Replace `OccurAt` with `OccurAtFormatted` using `time.Unix(payLoad.OccurAt, 0).Format("2006-01-02 15:04:05 UTC")`

## Testing

```bash
cd src && go test ./pkg/notifier/handler/notification/...
```

## Sign-off

Signed-off-by: Ross Golder <ross@golder.org>

## Related Issue

Fixes notification timestamp display in Slack.

### DIFF
--- a/src/pkg/notifier/handler/notification/slack_handler.go
+++ b/src/pkg/notifier/handler/notification/slack_handler.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"strings"
 	"text/template"
+	"time"
 
 	"github.com/goharbor/harbor/src/common/job/models"
 	"github.com/goharbor/harbor/src/jobservice/job"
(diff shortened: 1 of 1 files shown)


rossigee:
Closing this PR — on reflection, the original code was not actually broken. The Slack `<!date^{{.OccurAt}}^{date} at {time}|...>` format is valid and works correctly with the Unix timestamp as-is. This 'fix' would have regressed the user-local-timezone rendering that Slack provides natively, and also introduced a misleading hardcoded 'UTC' label. Apologies for the noise — wasted everyone's time on a non-issue.

PR #23184, opened 2026-04-29
bump pg to v18

wy65701436:
currently harbor is using the pg 15, I will use this pull request to bump it to pg 18.

Thank you for contributing to Harbor!

# Comprehensive Summary of your change

# Issue being fixed
Fixes #(issue)

Please indicate you've done the following:
- [ ] Well Written Title and Summary of the PR
- [ ] Label the PR as needed. "release-note/ignore-for-release, release-note/new-feature, release-note/update, release-note/enhancement, release-note/community, release-note/breaking-change, release-note/docs, release-note/infra, release-note/deprecation"
- [ ] Accepted the DCO. Commits without the DCO will delay acceptance.
- [ ] Made sure tests are passing and test coverage is added if needed.
- [ ] Considered the docs impact and opened a new docs issue or PR with docs changes if needed in [website repository](https://github.com/goharbor/website).

PR #23282, opened 2026-05-25
fix: build valkey from source to support arm64

ranimandepudi:
The Photon OS repos dont ship a valkey package for aarch64, which breaks arm64 builds. This replaces the `tdnf install valkey` approach with a multi-stage build that compiles valkey from the upstream release tarball.

The same Dockerfile now produces working images on both amd64 and arm64 without any Makefile changes, the existing `_build_base` macro already handles multi-stage builds via BuildKit.
No changes to CI pipelines,image names, or downstream Dockerfiles are needed.

What changed:
- Multi-stage builder compiles valkey 9.1.0 with TLS support
- Only the final binaries are copied into the runtime image (~99MB)
- Added redis-server/redis-cli symlinks for backward compat
- Reverted Makefile to main (no changes needed there)

Tested on arm64:
- valkey-server 9.1.0 starts cleanly (jemalloc-5.3.0, 64-bit)
- SET/GET operations work
- No regressions on amd64

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,6 +133,10 @@ PREPARE_VERSION_NAME=versions
 REGISTRYVERSION=v2.8.3-patch-redis
 TRIVYVERSION=v0.69.3
 TRIVYADAPTERVERSION=v0.35.1
+# VALKEYVERSION and VALKEYSHA256 are only used for the arm64 source build.
+# On amd64 the Photon tdnf package version is used directly.
+VALKEYVERSION=9.0.3
+VALKEYSHA256=e220f4b0143292ee6ea6d705aa40d45a0c8a77759b3e94c201cb5c25dbdca42f
 NODEBUILDIMAGE=node:16.18.0
 
 # version of registry for pulling the source code
@@ -432,6 +436,7 @@ build:
 	make -f $(MAKEFILEPATH_PHOTON)/Makefile $(BUILDTARGET) -e DEVFLAG=$(DEVFLAG) -e GOBUILDIMAGE=$(GOBUILDIMAGE) -e NODEBUILDIMAGE=$(NODEBUILDIMAGE) \
 	 -e REGISTRYVERSION=$(REGISTRYVERSION) -e REGISTRY_SRC_TAG=$(REGISTRY_SRC_TAG)  -e DISTRIBUTION_SRC=$(DISTRIBUTION_SRC)\
 	 -e TRIVYVERSION=$(TRIVYVERSION) -e TRIVYADAPTERVERSION=$(TRIVYADAPTERVERSION) \
+	 -e VALKEYVERSION=$(VALKEYVERSION) -e VALKEYSHA256=$(VALKEYSHA256) \
 	 -e VERSIONTAG=$(VERSIONTAG) \
 	 -e DOCKERNETWORK=$(DOCKERNETWORK) \
 	 -e BUILDREG=$(BUILDREG) -e BUILDTRIVYADP=$(BUILDTRIVYADP) \
@@ -457,7 +462,11 @@ build_base_docker:
 	@for name in $(BUILDBASETARGET); do \
 		echo $$name ; \
 		sleep 30 ; \
-		$(DOCKERBUILD) --pull --no-cache -f $(MAKEFILEPATH_PHOTON)/$$name/Dockerfile.base -t $(BASEIMAGENAMESPACE)/harbor-$$name-base:$(BASEIMAGETAG) --label base-build-date=$(date +"%Y%m%d") . ; \
+		valkey_args="" ; \
+		if [ "$$name" = "valkey" ]; then \
+			valkey_args="--build-arg VALKEY_VERSION=$(VALKEYVERSION) --build-arg VALKEY_SHA256=$(VALKEYSHA256)" ; \
+		fi ; \
+		$(DOCKERBUILD) --pull --no-cache -f $(MAKEFILEPATH_PHOTON)/$$name/Dockerfile.base $$valkey_args -t $(BASEIMAGENAMESPACE)/harbor-$$name-base:$(BASEIMAGETAG) --label base-build-date=$(date +"%Y%m%d") . ; \
 		if [ "$(PUSHBASEIMAGE)" != "false" ] ; then \
 			$(PUSHSCRIPTPATH)/$(PUSHSCRIPTNAME) $(BASEIMAGENAMESPACE)/harbor-$$name-base:$(BASEIMAGETAG) $(REGISTRYUSER) $(REGISTRYPASSWORD) || exit 1; \
 		fi ; \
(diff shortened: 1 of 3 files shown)


ranimandepudi:
@danacr @Vad1mo @wy65701436 @chlins @OrlinVasilev 
Please review !! Thanks

make/photon/valkey/Dockerfile.base
```diff
@@ -1,6 +1,28 @@
-FROM goharbor/photon:5.0
+ARG VALKEY_VERSION=9.1.0
```
wy65701436: does this change impact all the ARCH? I suppose it should not.
If it's not possible to have one docker file for all the ARCH, we can have one for each.

ranimandepudi: @wy65701436 
This single Dockerfile works on all architectures, make compiles native code on whatever platform you are building on. `goharbor/photon:5.0` is multi-arch. No separate Dockerfiles needed. One Dockerfile, all arches.

ranimandepudi:
Copilot issues - both fixed !!
|| true only applies to strip (separate RUN layer) - cant mask build failures
Added SHA256 verification of the downloaded tarball

make/photon/valkey/Dockerfile.base
wy65701436: I’m not against fixing ARM, but I’m against changing the Valkey install path globally. The existing Photon package path should remain the default because it preserves package provenance, security update flow, scanner behavior, and existing amd64 behavior. If ARM needs source-built Valkey, please isolate that into an ARM-specific code path and do not impact the existing tdnf install -y valkey path, and documentation for CVE/update ownership.

Btw, did you ask the photon team about the ARM version Valkey?

wy65701436 (changes requested):
please do not change the Valkey installation for non-ARM path.

make/photon/valkey/Dockerfile.base
ranimandepudi: Agree !! Ok let me commit it now and I haven't reached out to any but can raise an issue here: https://github.com/vmware/photon now if that works !!

wy65701436: yes, please open an issue and link to our PR, thanks.

ranimandepudi: https://github.com/vmware/photon/issues/1655
Opened the PR and updated the commit !! Thanks !!

make/photon/valkey/Dockerfile.base
```diff
@@ -1,6 +1,43 @@
-FROM goharbor/photon:5.0
+ARG TARGETARCH
+ARG VALKEY_VERSION=9.1.0
```
wy65701436: do not define the specific valkey version in the docker file, but in the root make file if you want and parse it into the docker file

make/photon/valkey/Dockerfile.base
```diff
@@ -1,6 +1,43 @@
-FROM goharbor/photon:5.0
+ARG TARGETARCH
```
wy65701436: Could you please indicate the specific file that handles the parsing the **TARGETARCH** build argument during the build process?

make/photon/valkey/Dockerfile.base
```diff
@@ -1,6 +1,43 @@
...
+COPY --from=arm64-builder /usr/local/bin/valkey-server /usr/local/bin/valkey-server
+COPY --from=arm64-builder /usr/local/bin/valkey-cli /usr/local/bin/valkey-cli
+COPY --from=arm64-builder /usr/local/bin/valkey-check-aof /usr/local/bin/valkey-check-aof
+COPY --from=arm64-builder /usr/local/bin/valkey-check-rdb /usr/local/bin/valkey-check-rdb
+COPY --from=arm64-builder /usr/local/bin/valkey-benchmark /usr/local/bin/valkey-benchmark
+RUN ln -sf /usr/local/bin/valkey-server /usr/local/bin/redis-server \
+    && ln -sf /usr/local/bin/valkey-cli /usr/local/bin/redis-cli
+
+# ---- Select final image based on target architecture ----
+FROM ${TARGETARCH}-base
```
wy65701436, line 47: Please provide validation for both linux/amd64 and linux/arm64, including starting Valkey with Harbor’s existing redis.conf and confirming the healthcheck works.

ranimandepudi:
> please do not change the Valkey installation for non-ARM path.

Addressed in the latest commit. The non-Arm path is unchanged and still uses Photon package installation with tdnf install -y valkey. Source build is isolated to ARM only.

make/photon/valkey/Dockerfile.base
ranimandepudi: Done in the latest commit. `VALKEYVERSION` and `VALKEYSHA256` are now defined in the root Makefile and passed through `make/photon/Makefile` into valkey [Dockerfile.base] as build args. The Dockerfile no longer hardcodes default values.

make/photon/valkey/Dockerfile.base
ranimandepudi: `TARGETARCH` is provided by Docker BuildKit automatically based on the selected platform. The platform is set in `make/photon/Makefile` via `--platform=$(PLATFORM)`, and Dockerfile.base consumes `ARG TARGETARCH` to select the final stage. So the effective flow is:

make/photon/Makefile sets platform for docker build
BuildKit populates TARGETARCH
valkey Dockerfile.base uses TARGETARCH for stage selection

make/photon/valkey/Dockerfile.base
ranimandepudi: Validation summary:

linux/arm64 (Graviton):

Built valkey base and final images successfully
Started container with Harbor redis.conf
valkey-cli PING returned PONG
Harbor docker-healthcheck passed and container reported healthy

linux/amd64:
Verified build path with the unchanged package-based install flow
Final image build succeeded with Harbor redis.conf included

Makefile
```diff
@@ -133,6 +133,8 @@ PREPARE_VERSION_NAME=versions
 REGISTRYVERSION=v2.8.3-patch-redis
 TRIVYVERSION=v0.69.3
 TRIVYADAPTERVERSION=v0.35.1
+VALKEYVERSION=9.1.0
```
chlins: Should we use the latest `.0` version? Previously we installed the valkey in amd64 and verify it(valkey 9.0.3).

chlins: And please add the comment to clarify this valkey version is only applied to arm64 arch, not for amd64.

wy65701436: I am ok that we use different version for different arch if it's not easy to consolidate it.

ranimandepudi:
@wy65701436 @chlins - updated the commit - please review and open for final suggestions if any !! Thanks

chlins (approved):
lgtm

wy65701436 (approved):
lgtm

ranimandepudi:
@wy65701436 
Our valkey fix is NOT the cause. Its working correctly.
The failure is from PR #23184 which bumped PostgreSQL to v18: postgresql18-server has no arm64 package in Photon.
This needs the same approach we used for valkey (either source-build for arm64, or file an issue on the Photon repo for the missing aarch64 package).

I will open an issue on Harbor referencing this: the build-package arm64 job will fail until postgresql18-server is available for aarch64 or a workaround is added to make/photon/db/Dockerfile.base and work on it !!

PR #23283, opened 2026-05-26
perf(replication): filter event policies in query

### LINKED ISSUE #23277: harbor-core OOMKill when deleting artifacts with large number of replication policies (100k+)
When Harbor has a large number of replication policies (100k+) and artifact deletions happen concurrently, `harbor-core` is killed by OOMKill. This is because every artifact deletion event triggers a full, unbounded load of **all** replication policies into memory with no pagination and no database-level filtering.

The call chain on every artifact deletion:

```
DELETE /api/v2.0/.../artifacts/{ref}
  → artifact.Controller.Delete()
  → event: TopicDeleteArtifact
  → replication.Handler.Handle()
  → event.Handle()
  → getRelatedPolicies()
  → replication.Ctl.ListPolicies(ctx, nil)   ← nil query, no filter
  → dao.List() → qs.All(&policies)           ← entire table loaded into memory
```

### Additional Context

- This use case may be unconventional: we created hundreds of thousands of replication policies with **manual trigger type**, and invoke them programmatically via API. We acknowledge this may not be the intended usage pattern.
- However, we found **no documentation warning about the performance implications** of having a large number of replication policies, nor any documented limits on policy count.
- This issue affects **all trigger types**, not just event-based ones: since `getRelatedPolicies()` loads all policies unconditionally, even 100k manual-trigger policies (which can never be triggered by artifact events) are fully materialized in memory on every artifact deletion.
- This issue scales with the total number of enabled policies AND with the concurrency of deletion requests.
- The manual-trigger policies are never matched by the event handler (they are filtered out in Go code), so all the memory and DB load is effectively wasted.

### Suggestions

We think either of the following improvements would address the root cause:

1. **Push filtering to the database**: Accept filter parameters (e.g., `trigger_type`, `enabled`) in `ListPolicies` and apply them as SQL `WHERE` conditions, avoiding loading irrelevant policies into memory entirely.
2. **Cache the policy list**: After the first full load, cache and reuse the replication policy list (with appropriate TTL or invalidation on policy create/update/delete), so that concurrent artifact deletions share a single load instead of each triggering a full table scan.

chlins:

This pull request improves the efficiency and clarity of the replication event handler by refactoring how related replication policies are queried and filtered. The main change is to move filtering logic for enabled, event-based policies into the query itself, reducing unnecessary in-memory checks.

**Replication policy filtering improvements:**

* Updated `getRelatedPolicies` in `handler.go` to query only enabled, event-based replication policies directly using the `ListPolicies` method with a constructed query, removing redundant in-memory checks for policy status and trigger type.
* Added an import for the `q` package to support the new query construction.
# Issue being fixed
Fixes https://github.com/goharbor/harbor/issues/23277

### DIFF
--- a/src/controller/event/handler/replication/event/handler.go
+++ b/src/controller/event/handler/replication/event/handler.go
@@ -23,6 +23,7 @@ import (
 	"github.com/goharbor/harbor/src/controller/replication"
 	repctlmodel "github.com/goharbor/harbor/src/controller/replication/model"
 	"github.com/goharbor/harbor/src/lib/log"
+	"github.com/goharbor/harbor/src/lib/q"
 	"github.com/goharbor/harbor/src/pkg/reg/filter"
 	"github.com/goharbor/harbor/src/pkg/reg/model"
 	"github.com/goharbor/harbor/src/pkg/task"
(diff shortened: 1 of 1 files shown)


wy65701436 (approved):
lgtm

HirazawaUi (commented):
LGTM

Thank you very much for your quick action.

stonezdj (approved):

PR #23295, opened 2026-05-28
fix(registryctl): avoid fatal log on graceful shutdown

### LINKED ISSUE #23294: registryctl logs http: Server closed as fatal during graceful shutdown
# Issue Title

registryctl logs http: Server closed as fatal during graceful shutdown

# Description

## Summary

registryctl logs a fatal error during normal graceful shutdown:

However, this is caused by http.ErrServerClosed, which is expected behavior when the HTTP server shuts down gracefully.

## Current Behavior

In RegistryCtl.Start():

err = regCtl.ListenAndServe() 
... 
if err != nil {     log.Fatal(err) } 

ListenAndServe() returns http.ErrServerClosed after a successful shutdown, causing Harbor to incorrectly emit a fatal log.

## Expected Behavior

Graceful shutdown should not produce fatal logs.

http.ErrServerClosed should be ignored explicitly.

## Result

After applying the fix:

- graceful shutdown logs are clean
- no misleading fatal message appears
- shutdown behavior remains unchanged

rakshityadav1868:

This PR fixes a misleading fatal log emitted by registryctl during graceful shutdown.

Previously, http.ErrServerClosed returned from ListenAndServe() was treated as a fatal error and logged as:

 [FATAL] http: Server closed 

This behavior occurs during normal graceful shutdown and should not be considered a fatal error.

The fix explicitly ignores http.ErrServerClosed before calling log.Fatal.

# Issue being fixed

Fixes #23294

### DIFF
--- a/src/registryctl/main.go
+++ b/src/registryctl/main.go
@@ -17,6 +17,7 @@ package main
 import (
 	"context"
 	"crypto/tls"
+	"errors"
 	"flag"
 	"net/http"
 	"os"
@@ -86,7 +87,7 @@ func (s *RegistryCtl) Start() {
 		err = regCtl.ListenAndServe()
 	}
 	<-shutdown
-	if err != nil {
+	if err != nil && !errors.Is(err, http.ErrServerClosed) {
 		log.Fatal(err)
 	}
 }


Vad1mo (approved):

wy65701436 (approved):
lgtm

Issue #17709, opened 2022-10-27
Provide meaningfull metrics on artifaces pulled and pushed. 


kbabuadze:
**Is your feature request related to a problem? Please describe.**
Harbor doesn't provide any meaningful metrics regarding upload and download statistics of the artifact. 
`/metric` endpoint provides statistics on the number of downloaded artifacts, however it doesn't really give you the size of the artifact that has been pulled. 
Especially considering layering you can't estimate how much bytes of data has been downloaded. 

**Describe the solution you'd like**
It would be great to have metrics that shows how much data has been downloaded from the 3rd party registry, from harbor itself and how much data has been pushed to the 3rd party registry. 
I would give the visibility on how much is harbor actually helping to save and  help with pricing calculation in case of switching registry providers.

PR #23298, opened 2026-05-28
Add metric of push count per project

### LINKED ISSUE #23200: Add artifacts pushed metric
The exporter currently exposes pull metrics (harbor_artifact_pulled) but lacks a corresponding push metric. This gap limits visibility into artifact push activity when monitoring.

Proposal:
Add harbor_artifact_pushed metric to track push activity per project, following the same pattern as the existing pull metric.

No schema changes required; uses existing artifact.push_time field

czifrushanna:
This PR adds a new Prometheus gauge metric harbor_artifact_pushed{project_name="<name>"} exposed by the Harbor exporter. The metric reports the number of artifact rows with a non-null push_time per project (SQL-based aggregation in the exporter). Implemented in "project_collector.go" and unit tests added in "project_collector_test.go".

Provides visibility into how many artifacts have been pushed per project for monitoring, alerting, and capacity planning.

Fixes #23200


czifrushanna:
It is still needed! Related issues: #23200, #17709

Vad1mo:
@czifrushanna @rochaporto thanks.

The implementation doesn't measure pushes. It counts current artifact rows:

1. push_time is non-null on every artifact row. The schema default is `CURRENT_TIMESTAMP` and the controller always sets it on create. So `COUNT(artifact.push_time)` is identical to `COUNT(*)`.
2. Re-pushing an existing digest is a no-op for this metric. ensureArtifact returns early when the digest already exists, no row is created and push_time is not updated. N pushes of the same artifact count as 1.
3. Deleted artifacts drop out of the count, so the value decreases over time. increase()/rate() over this metric goes negative after deletion or GC.

The result is exactly sum by` (project_name)(harbor_project_artifact_total)`, an existing metric under a new, misleading name. It also doesn't answer the use case from #23200 (artifacts pushed in a given period, robust to deletions).

This is also not "the same pattern as the existing pull metric": artifact_pulled sums repository.pull_count, which is a real per-event counter incremented on every pull. There is no push-side equivalent column, and this PR doesn't create one.

A correct push counter needs an append-only source, since the artifact table only holds current state. 


Two options:

- Aggregate PUSH operations from the audit log in the exporter (caveat: audit log purge breaks monotonicity).
- Add a push_count column incremented on push, mirroring repository.pull_count. That's a schema change, which #23200 wanted to avoid, but it's what the pull metric actually does.

As is, I'd suggest closing this PR and consolidating the discussion in #17709 (and closing #23200 as duplicate).

Issue #23311, opened 2026-06-02
Refresh Photon/base packages for official v2.15.x images


aldrinjenson:
How can we help you?

A security scan of the official Harbor `v2.15.1` container images found reachable Critical/High CVEs in packages that have available fixes. This appears primarily related to stale Photon/base packages in the published images, plus the Go toolchain/runtime version used for some binaries.

I am not including private scanner links, scan IDs, or a full vulnerability dump here. The intent is to ask whether the Harbor project can publish patched `v2.15.x` images, or confirm that the current `release-2.15.0` branch changes will be released soon.

Observed rollup for official `v2.15.1` images:

| Component | Image | Reachable Critical/High | Fixable reachable Critical/High | Main package families observed |
| --- | --- | ---: | ---: | --- |
| harbor-core | `goharbor/harbor-core:v2.15.1` | 24 | 20 | `stdlib`, `glibc`, `curl`, `distribution`, `libssh2`, `ncurses` |
| trivy-adapter-photon | `goharbor/trivy-adapter-photon:v2.15.1` | 24 | 24 | `stdlib`, `glibc`, `curl`, `go-git`, `libssh2`, `ncurses` |
| harbor-exporter | `goharbor/harbor-exporter:v2.15.1` | 21 | 17 | `stdlib`, `glibc`, `curl`, `distribution`, `libssh2`, `ncurses` |
| harbor-jobservice | `goharbor/harbor-jobservice:v2.15.1` | 21 | 17 | `stdlib`, `glibc`, `curl`, `distribution`, `libssh2`, `ncurses` |
| registry-photon | `goharbor/registry-photon:v2.15.1` | 17 | 17 | `stdlib`, `glibc`, `curl`, `libssh2`, `ncurses`, `expat` |
| harbor-registryctl | `goharbor/harbor-registryctl:v2.15.1` | 12 | 9 | `glibc`, `curl`, `distribution`, `libssh2`, `ncurses`, `expat` |
| harbor-db | `goharbor/harbor-db:v2.15.1` | 10 | 10 | `glibc`, `ncurses`, `curl`, `libssh2`, `expat` |
| harbor-portal | `goharbor/harbor-portal:v2.15.1` | 9 | 9 | `glibc`, `curl`, `libssh2`, `ncurses`, `expat` |
| nginx-photon | `goharbor/nginx-photon:v2.15.1` | 9 | 9 | `glibc`, `curl`, `libssh2`, `ncurses`, `expat` |
[...]

PR #23312, opened 2026-06-02
Refresh Photon packages in component base images

aldrinjenson:
## Summary
- Run `tdnf update -y` in each Photon component base image before installing component-specific packages.
- Keep the existing package installs and user/group setup unchanged.
- Make patched `2.15.x` rebuilds less dependent on whether the inherited `goharbor/photon:5.0` tag or local build cache already contains the latest Photon packages.

## Motivation
A container security scan of the official `v2.15.1` image set showed fixable reachable findings in common Photon package families such as `glibc`, `curl-libs`, `libssh2`, `ncurses-libs`, and `expat-libs`. The release branch already refreshes the shared Photon base, but the component base images can also refresh packages directly at build time to reduce stale-package risk in rebuilt images.

Addresses #23311.

## Testing
- `git diff --check`
- Built representative base image:
  `docker build --platform linux/amd64 -f make/photon/core/Dockerfile.base -t harbor-core-base-security-refresh-test .`

The local Docker environment emitted a platform warning because the host is arm64, but the Dockerfile parsed and the `tdnf update` step completed successfully.

### DIFF
--- a/make/photon/core/Dockerfile.base
+++ b/make/photon/core/Dockerfile.base
@@ -1,5 +1,7 @@
 FROM goharbor/photon:5.0
 
+RUN tdnf update -y && tdnf clean all
+
 RUN tdnf install -y tzdata shadow >> /dev/null \
     && tdnf clean all \
     && groupadd -r -g 10000 harbor && useradd --no-log-init -r -m -g 10000 -u 10000 harbor \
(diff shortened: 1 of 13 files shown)


aldrinjenson:
Closing in favor of signed-off replacement PR #23313. Same patch; the new branch includes DCO sign-off.

PR #23319, opened 2026-06-04
fix(scan): make vulnerability report persistence idempotent

### LINKED ISSUE #23310: Massive database growth in report_vulnerability_record table linked to Trivy scans
**Summary:**
We have detected an uncontrolled and rapid growth in our PostgreSQL database. Through our monitoring probes, we identified that the issue started around May 21, 2026. The `report_vulnerability_record` table is the main driver of this growth. While our database historically maintained a stable size of approximately 16 GB, it has been increasing by roughly 6 GB per day since May 21, recently reaching 96 GB.

**Observed Behavior:**
The database growth is strictly correlated with the execution time of the Trivy vulnerability scan cron jobs. This scan configuration has been in place long before May 21, and there has been no significant increase in the number of images stored in Harbor that would justify such an exponential data influx.

*Note: We performed a routine upgrade from Harbor 2.12.4 to 2.13.5 on May 26, 2026, as part of our standard maintenance policy. This upgrade was unrelated to the issue and has not resolved the behavior.*

**Expected Behavior:**
The database size should remain stable (around 16 GB), consistent with previous usage patterns, despite the frequency of Trivy scans.

**Actual Behavior:**
The database has grown to over 96 GB, with the `report_vulnerability_record` table accounting for 87 GB.

**Database Statistics:**
```sql
         table_name          | pg_size_pretty | pg_total_relation_size
-----------------------------+----------------+------------------------
 report_vulnerability_record | 87 GB          |            93823598592
```

**Versions:**
- **Harbor version:** [2.12.4 -> 2.13.5]
- **Docker engine version:** [24.0.5]
- **Docker-compose version:** [2.20.2]
- **External Redis:** [6.2.20]
- **External PostgreSQL:** [13.10]

**Additional Context:**
This issue seems related to the following reported issue: https://github.com/goharbor/harbor/issues/17927

Vad1mo:
Every re-scan deletes the artifact's `scan_report` row and re-inserts all of its `report_vulnerability_record` rows under a new random UUID, which is what bloats that table on registries with a daily Scan All (#23310). This PR reuses the existing report row and reconciles its associations with a set-diff, so an unchanged result writes nothing, and as a side effect the previous report stays readable while the re-scan is queued or running instead of disappearing.

## Problem

`report_vulnerability_record` grows without bound on a stable image set. The reporter in #23310 saw 16 GB to 96 GB at about 6 GB a day, with the unique `(report_uuid, vuln_record_id)` index alone at 40 GB, correlated with the scheduled Trivy Scan All.

The cause is the contract of `MakePlaceHolder`: delete the existing report, create a new one. The `scan_report` row for `(digest, registration_uuid, mime_type)` is deleted, cascading its `report_vulnerability_record` rows, and recreated with a fresh random UUID. `PostScan` then bulk-inserts every association under that UUID. A digest is immutable and a scanner is deterministic for a given vulnerability DB, so a daily Scan All rewrites an essentially identical join set every day. The foreign keys cascade, so the logical row count stays bounded by `scan_report`; the growth is dead tuples and random-UUID B-tree churn on the unique index that autovacuum cannot keep up with.

The same delete has a second effect. Scan All calls `Scan()` for every artifact at enqueue time, so every report in the registry is gone before the first worker starts, and each artifact stays without a report until its own job finishes. During that window the vulnerabilities addition returns `{}`, which is also the response for a never-scanned artifact, and Security Hub counts the artifact as clean. Measured on a dev build of main with one artifact (nginx:1.19, 424 findings) and the IMAGE_SCAN queue paused so the re-scan stays Pending:

| | before re-scan | while Pending | after re-scan |
|---|---|---|---|
| `GET .../additions/vulnerabilities` | 424 CVEs | `{}` | 424 CVEs |
| `GET /security/summary` total_vuls | 424 | 0 | 424 |
| `GET /security/vul` X-Total-Count | 424 | 0 | 424 |
| `scan_report` row | uuid A | uuid B, empty report | uuid B |
| `report_vulnerability_record` rows | 424 | 0 | 424 |

## Solution

Make report persistence idempotent.

1. Stable report row (`src/pkg/scan/vulnerability/vul.go`). `MakePlaceHolder` reuses the existing `scan_report` row, with its UUID and associations, per `(digest, registration_uuid, mime_type)` instead of deleting and recreating it. The guard that refuses a new scan while a previous one is still running is unchanged.
2. Set-diff instead of bulk insert (`src/pkg/scan/dao/scan/vulnerability.go`, `src/pkg/scan/postprocessors/report_converters.go`). A new `SyncForReport` reconciles `report_vulnerability_record` in one transaction: insert the missing rows, delete the stale ones, and clear all associations for an empty result. `PostScan` calls it instead of `InsertForReport`. An empty raw report is rejected before any write so the last successful result stays intact, and a clean image arrives as a structured report with an empty vulnerability list that reconciles to zero through `toSchema`. A re-scan with an unchanged result writes no rows. A transaction-scoped `pg_advisory_xact_lock(hashtextextended(report_uuid, 0))` serialises two reconciliations of the same report, so overlapping scans cannot collide on the unique index or interleave.
3. Deterministic task resolution (`src/pkg/task/dao/task.go`). A reused report UUID is referenced by more than one task over time, so `ListScanTasksByReportUUID` now runs `ORDER BY id DESC LIMIT 1`. Every caller takes `tasks[0]` for status and log resolution and gets the current scan rather than a stale one, and the matched set no longer grows.

With the PR applied to the same dev build, the table above reads 424 in every column while the task is Pending, and also after the re-scan fails with status Error. The report UUID stays the same across runs.

## Benefits

- `report_vulnerability_record` stops growing on registries whose image set is stable. A re-scan with the same result writes nothing, and the unique index stops churning on random UUIDs.
- No blind window during Scan All. The last successful result stays readable until the new one replaces it, on the API, in Security Hub and in the scan export, and it survives a failed re-scan.
- A saved `GET /scan/{report_id}/log` link keeps working after a re-scan and returns the latest log.

### Behaviour changes

`report_id` is stable across re-scans for a given (artifact, scanner, mime type). It used to be a fresh random UUID per run. Anyone who needs per-run identity should use the execution or task id, or the start and end times. A saved `GET /scan/{report_id}/log` link no longer 404s after a re-scan; it returns the latest scan's log. SBOM `report_id` is unchanged.

The previous report survives a running or failed re-scan. Surfaces that gate on status (scan badge and summary, the pull-time vulnerability gate, the P2P preheat gate) read status from the newest task, show Running or Error as before, and fail closed on anything but Success, so there is no policy bypass. The raw vulnerabilities addition, Security Hub and the scan export, which do not gate on status, return the last successful result until the next scan succeeds. Previously they returned nothing, which read as a clean artifact.

Starting a scan no longer mutates `scan_report`, so two scans of the same artifact can both be submitted in a narrow window. Reconciliation is serialised, so the second one is a redundant scan, never inconsistent associations.

Task rows per report stay bounded by the existing daily `EXECUTION_SWEEP` with retention 1; after each sweep a reused report UUID is back to about one referencing task.

### Downgrade

After re-scans under this change, a report UUID is referenced by several tasks and the `ORDER BY id DESC LIMIT 1` tie-break is what keeps status correct. Rolling `core` back to a build without it can show stale scan status, including a brief fail-open of the pull gate, until the next daily sweep, at most 24 hours. Backport that one-liner to any release operators may roll back to.

### Out of scope

This stops future growth but does not shrink a table that is already bloated. Reclaim once with `VACUUM (FULL)` or `pg_repack` plus `REINDEX` on `report_vulnerability_record`. Per-table autovacuum tuning and an orphan-record sweep are separate follow-ups. The SBOM path still deletes its report before a re-scan, so a failed SBOM regeneration loses the prior SBOM; that is #23320.

### Tests

DAO test for `SyncForReport`: reconcile to a changed set, an idempotent re-sync writes nothing, sync to empty clears. Task DAO test: a reused report UUID resolves to the latest task only. Handler tests: `MakePlaceHolder` reuses an existing report's UUID with no Create or Delete, and creates one only when none exists.

# Issue being fixed
Fixes #23310


Vad1mo:
Pushed 16d268ff5, which replaces the empty-report branch in `ToRelationalSchema`. That branch called `SyncForReport` with no ids and returned a nil error, so on a reused `scan_report` row it cleared every vulnerability association the previous scan had left. It returned before `updateReport`, so the severity counters and the `report` column kept the previous scan's values, and the caller in `job.go` then called `handler.Update` with an empty string, which `UpdateReportData` rejects with "missing report JSON data". The job therefore failed anyway and the row was left half updated: associations gone, counters and report text stale, task Error.

An empty raw report is now rejected before any write, so the last successful result stays intact and the scan is reported as failed. A clean image is not affected, since it arrives as a structured report with an empty `vulnerabilities` array and reconciles to zero through `toSchema`. The path is reachable only through a jobservice shutdown race, where the per mime goroutine returns on `ctx.SystemContext().Done()` without recording an error and `shouldStop` only inspects the OP command, so an empty entry in `rawReports` falls through to `PostScan`. For both vulnerability mime types `fetchScanReportFromScanner` already rejects an empty body through `ResolveData`.

`TestConvertEmptyReportKeepsLastResult` covers it. The `## Solution` section of the description has been updated, since item 2 still described the old behaviour.

chlins (commented):
### 1. Atomic publication of the report

`SyncForReport` commits the association changes before the vulnerability count and report JSON are updated. The advisory lock protects the association reconciliation, but not publication of the complete report.

For example, suppose the previous successful report contains vulnerabilities and a re-scan returns a clean result. The association transaction removes the old entries and commits. If the subsequent count or report JSON update fails, the scan fails, but the previous successful result has already been modified. Concurrent scans can also leave associations from one run combined with the count or JSON from another.

Could we publish the associations, count, and report JSON in a single transaction, with the report-level lock covering that entire operation? Reusing the report UUID makes these writes updates to the currently visible result, so they need to succeed or roll back together to preserve the last successful report on failure.

### 2. Bind report publication to the scan task

Selecting the highest task ID resolves the latest submitted task, but does not guarantee that the stored report was produced by that task.

As noted in the PR description, two scans can still pass the running-scan check concurrently. If task A is created before task B, but B finishes first and A publishes afterward, A overwrites the shared report while its status, timestamps, and logs are resolved from B. The advisory lock serializes association updates, but does not enforce task ordering. This remains possible even if the complete report publication is made transactional.

Could we either prevent concurrent scan submissions atomically for the same report, or bind publication to a task ID/generation and reject superseded writers? Otherwise, consumers can observe a report whose contents and task metadata belong to different scans, so the concurrent case is not merely a redundant scan.

PR #23331, opened 2026-06-09
fix: use postgres 18 in db base image

bupd:
## Summary
- remove the remaining `postgresql15-server` install from the DB base image
- install `postgresql18-server` directly as the only PostgreSQL server package
- keep the change limited to `make/photon/db/Dockerfile.base`

Failing job reference: https://github.com/goharbor/harbor/actions/runs/27182030877/job/80243049233

PR #23332, opened 2026-06-09
[WIP]fix postgres15 removed from photon snapshot issue

jUDASmILE:

# Issue being fixed
Fixes #(issue)
Photon modified their snapshot in recently, Postgres15 releases were removed from snapshot, we need to install it from main repo and switch back to snapshot repo after the installation. This change implement the mentioned fix. 

### DIFF
--- a/make/photon/db/Dockerfile.base
+++ b/make/photon/db/Dockerfile.base
@@ -8,7 +8,19 @@ RUN tdnf install -y shadow >> /dev/null \
     && groupadd -r postgres --gid=999 \
     && useradd -m -r -g postgres --uid=999 postgres
 
-RUN tdnf install -y postgresql15-server >> /dev/null
+RUN if [ "${TARGETARCH:-amd64}" = "amd64" ]; then \
+        cp /etc/yum.repos.d/photon-snapshot.repo /etc/yum.repos.d/photon-snapshot.repo.bak \
+        && sed -i '/^snapshot/d' /etc/yum.repos.d/photon-snapshot.repo \
+        && tdnf clean all \
+        && tdnf makecache \
+        && tdnf install -y postgresql15-server >> /dev/null \
+        && mv /etc/yum.repos.d/photon-snapshot.repo.bak /etc/yum.repos.d/photon-snapshot.repo \
+        && tdnf clean all \
+        && tdnf makecache; \
+    else \
+        tdnf install -y postgresql15-server >> /dev/null; \
+    fi
+
 RUN if [ "${TARGETARCH:-amd64}" = "arm64" ]; then \
         cp /etc/yum.repos.d/photon-snapshot.repo /etc/yum.repos.d/photon-snapshot.repo.bak \
         && sed -i '/^snapshot/d' /etc/yum.repos.d/photon-snapshot.repo \
--- a/.buildbaselog
+++ b/.buildbaselog
@@ -8,6 +8,9 @@
 *   Add date here... Add signature here...
 -   Add your reason here...
 
+*   Jun 09 2026 <aaron.an@broadcom.com>
+-   Refresh base image
+
 *   May 18 2026 <aaron.an@broadcom.com>
 -   Refresh base image
 


make/photon/db/Dockerfile.base
```diff
@@ -8,7 +8,19 @@ RUN tdnf install -y shadow >> /dev/null \
     && groupadd -r postgres --gid=999 \
     && useradd -m -r -g postgres --uid=999 postgres
 
-RUN tdnf install -y postgresql15-server >> /dev/null
+RUN if [ "${TARGETARCH:-amd64}" = "amd64" ]; then \
+        cp /etc/yum.repos.d/photon-snapshot.repo /etc/yum.repos.d/photon-snapshot.repo.bak \
+        && sed -i '/^snapshot/d' /etc/yum.repos.d/photon-snapshot.repo \
+        && tdnf clean all \
+        && tdnf makecache \
+        && tdnf install -y postgresql15-server >> /dev/null \
```
Vad1mo, line 16: there is no sense to install postgresql15 we should install postgresql18, 18 works we should not go back to 15

bupd, line 16:
```suggestion
        && tdnf install -y postgresql18-server >> /dev/null \
```

make/photon/db/Dockerfile.base
```diff
@@ -8,7 +8,19 @@ RUN tdnf install -y shadow >> /dev/null \
...
+        && mv /etc/yum.repos.d/photon-snapshot.repo.bak /etc/yum.repos.d/photon-snapshot.repo \
+        && tdnf clean all \
+        && tdnf makecache; \
+    else \
+        tdnf install -y postgresql15-server >> /dev/null; \
```
bupd, line 21:
```suggestion
        tdnf install -y postgresql18-server >> /dev/null; \
```

bupd (changes requested):
why need to go back to pg 15 - undoing the upgrade done here https://github.com/goharbor/harbor/pull/23184

bupd:
https://github.com/goharbor/harbor/pull/23331 - is better at this.

make/photon/db/Dockerfile.base
wy65701436: PG15 is required for the migration flow: PostgreSQL upgrades need both the old and new binaries present, otherwise the upgrade fails.

https://www.postgresql.org/docs/current/pgupgrade.html

jUDASmILE: please refer to Yan's comments to Vad1mo, we need to keep both versions. thanks.

make/photon/db/Dockerfile.base
jUDASmILE: please refer to Yan's comments to Vad1mo, we need to keep both versions. thanks.

jUDASmILE:
> #23331 - is better at this.

please refer to Yan's comments to Vad1mo, we need to keep both versions. thanks.

wy65701436:
@bupd @Vad1mo the old version pg15 is required for migration flow.  Please checkout the official document here, https://www.postgresql.org/docs/current/pgupgrade.html

@jUDASmILE has experience on Photon, I've asked him to confirm with Photon why they removed the pg 18 is not found and how we resolve this. And we will try to use this PR to fix the build issue.

Any idea from your side?

Vad1mo:
One thing not clear to me is why we are using the pinned photon-snapshot repo instead of rolling photon + photon-updates repos. This makes zero sense to me.

Vad1mo:
> @bupd @Vad1mo the old version pg15 is required for migration flow. Please checkout the official document here, https://www.postgresql.org/docs/current/pgupgrade.html

I understand the need for pgupgrade, but I don't understand why it has to be pg15 pgupgrade and not pg18 pgupgrade?

wy65701436:
> > @bupd @Vad1mo the old version pg15 is required for migration flow. Please checkout the official document here, https://www.postgresql.org/docs/current/pgupgrade.html
> 
> I understand the need for pgupgrade, but I don't understand why it has to be pg15 pgupgrade and not pg18 pgupgrade?

It uses the pg18 pg_upgrade, but even so, it still needs the pg15 binary. That's why we install pg15 in the database image.

jUDASmILE:
> One thing not clear to me is why we are using the pinned photon-snapshot repo instead of rolling photon + photon-updates repos. This makes zero sense to me.

We used to use that way, just photon.repo and photon-updates.repo. But everything changed since last year. Photon created multi branches as stablized releases. the origin two photon repos become development repos which there's no guarantee on stability and compatibility of packages. There're expected failure of package installation due to conflicts or compatibilities in some time when using these two repos, especially on new packages. Given this reason, we have to switch to snapshot based repos to make sure the stability of our builds. And photon deprecates legacy releases very actively as you can seen, they move the old releases to the archive repo frequently. then we need to install it from archive repo and switch back to snapshot repo time to time.

wy65701436:
> One thing not clear to me is why we are using the pinned photon-snapshot repo instead of rolling photon + photon-updates repos. This makes zero sense to me.

it's not a pinned version, but a recommend and stable way to get the latest package from the photon team.

wy65701436:
let's use this one resolve the problem.

https://github.com/goharbor/harbor/pull/23334

PR #23347, opened 2026-06-10
Set max-parallel to 1 to run build package for arm and amd64 sequentially

stonezdj:
fix issue the base image is overwriten by other platform.

related doc: https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#defining-the-maximum-number-of-concurrent-jobs

# Issue being fixed
Fixes #(issue)

### DIFF
--- a/.github/workflows/build-package.yml
+++ b/.github/workflows/build-package.yml
@@ -51,6 +51,7 @@ jobs:
     needs: [BUILD_BASE_IMAGE]
     strategy:
       fail-fast: false
+      max-parallel: 1
       matrix:
         arch: [amd64, arm64]
     env:


Vad1mo:
Thanks @stonezdj — `max-parallel: 1` makes the two jobs sequential, but both still push the same unsuffixed tag (`goharbor/harbor-<name>-base:dev`), so the tag remains **single-arch**: it would just deterministically end up as whatever the last job pushed (arm64, since it runs second in the matrix). Verified on Docker Hub today: `harbor-portal-base:dev` and `harbor-core-base:dev` are arm64-only manifests. It also roughly doubles the workflow wall-clock by removing the parallelism.

The actual fix needs the base tags to become multi-arch manifest lists, same as we already do for the component images: each arch job pushes `harbor-<name>-base:<tag>-amd64` / `-arm64`, and `CREATE_MANIFEST` combines them into the unsuffixed `<tag>` with `docker buildx imagetools create`.

Opened #23348 with that approach — `_build_base` keeps the local unsuffixed tag for component builds and only the push gets the arch suffix, so nothing else in the build flow changes.

PR #23334, opened 2026-06-09
fix: publish goharbor/photon:5.0 as a multi-arch manifest (#23324)

Vad1mo:
# Comprehensive Summary of your change

`goharbor/photon:5.0` was built and pushed independently by **both** the `amd64` and `arm64` `BUILD_PACKAGE` matrix jobs to the **same tag**, so whichever job finished last overwrote the other and the published tag ended up single-arch — "sometimes amd64, sometimes arm64" (#23324).

### Changes

- **`.github/workflows/build-package.yml`** (the #23324 fix) — build and push `goharbor/photon:5.0` **once** as a multi-arch (amd64+arm64) manifest list in a dedicated `BUILD_BASE_IMAGE` job that `BUILD_PACKAGE` depends on, instead of each per-arch job overwriting the tag. `goharbor/photon` is excluded from the component manifest list in `CREATE_MANIFEST` (it is published here and stays at tag `5.0`). The base-rebuild decision is made fail-safe (rebuild when the changed-files lookup is unavailable, e.g. on a new branch).
- **`make/photon/db/Dockerfile.base`** — apply the Photon-team workaround (per @wy65701436 / @jUDASmILE review, same as #23332): keep the stable `photon-snapshot` repo, but temporarily un-pin it (`sed '/^snapshot/d'`) to install the postgresql major that Photon has deprecated out of the current snapshot — `postgresql15-server` on amd64, `postgresql18-server` on arm64 — then restore the snapshot. Both majors are kept: pg15 supplies the old binaries at `/usr/pgsql/15/bin` required by `pg_upgrade`, pg18 is the runtime.
- **`make/photon/common/Dockerfile`** — keep the `photon-snapshot` repo (the stable Photon source) and merge its setup into a single `RUN`.

### Why keep both pg15 and pg18

`make/photon/db/Dockerfile`'s entrypoint is `["/docker-entrypoint.sh", "15", "18"]`. For an existing install, `docker-entrypoint.sh` runs `pg_upgrade` from the old pg15 cluster (`/usr/pgsql/15/bin`) to pg18 (`/usr/bin`) — so the pg15 binaries must be present. Fresh installs init pg18 directly. This is why dropping pg15 (#23331) is not viable.

### Verification

Built locally with BuildKit for **both** `linux/amd64` and `linux/arm64`:
- `goharbor/photon:5.0` (snapshot base) builds on both arches.
[...]

Issue #23324, opened 2026-06-06
goharbor/photon:5.0 tag is overwritten by different architectures (amd64 / arm64) instead of being a multi-arch manifest list


AgoCan:
You keep pushing different architecture images to the same goharbor/photon:5.0 tag. Sometimes it's amd64, sometimes it's arm64. This is not a proper multi-arch manifest list — you are just overwriting the tag.

Please publish it as a real multi-arch manifest list that serves the correct architecture automatically.

MinerYang:
Hi @Vad1mo , 
Could you please take a look at this issue? Seems build. index manifest list is a valid requirement when we build multi-arch platform image.

Appreciate
Miner

jUDASmILE:
Agreed, we need to implement the manifest index for multi-platform usage. @Vad1mo please help to take care of it, could you make it for the upcoming v2.15.2 release? thanks.

Vad1mo:
OK, I know what the problem is. It is in the nature of the matrix build.

I have a fix, but I am not sure about other things! especially why aren't we using rolling photon + photon-updates repos and onyl snapshots?

PR #23348, opened 2026-06-10
fix: publish harbor-*-base images as multi-arch manifests

Vad1mo:
After #23334 fixed `goharbor/photon:5.0`, an audit of every image push path in the pipeline found one remaining last-writer-wins overwrite: the **12 per-component base images** (`goharbor/harbor-core-base`, `harbor-db-base`, `harbor-portal-base`, `harbor-nginx-base`, `harbor-registry-base`, `harbor-registryctl-base`, `harbor-jobservice-base`, `harbor-log-base`, `harbor-prepare-base`, `harbor-exporter-base`, `harbor-trivy-adapter-base`, `harbor-valkey-base`).

`_build_base` in `make/photon/Makefile` pushes `goharbor/harbor-<name>-base:dev` via a plain `docker push` from **both** `BUILD_PACKAGE` matrix jobs — whichever architecture finishes last overwrites the tag. Verified on Docker Hub right now:

| tag | architectures |
|---|---|
| `harbor-portal-base:dev` | **arm64 only** (last push 2026-06-10 09:19) |
| `harbor-core-base:dev` | **arm64 only** (last push 2026-06-10 09:21) |
| `harbor-*-base:v2.x` (release tags) | **amd64 only** |

This breaks any consumer that pulls the bases from Docker Hub (`PULL_BASE_FROM_DOCKERHUB=true` with `BUILD_BASE=false`) on the architecture that lost the race.

### Changes — same pattern as the component images

- **`make/photon/Makefile`** — `_build_base` still builds the local **unsuffixed** tag (component Dockerfiles `FROM harbor-<name>-base:<tag>` keep working unchanged), but now pushes an **arch-suffixed** tag: `harbor-<name>-base:<tag>-amd64` / `-arm64`.
- **`.github/workflows/build-package.yml`** — each matrix job saves the list of `goharbor/*-base` repos; `CREATE_MANIFEST` combines `<tag>-amd64` + `<tag>-arm64` into the unsuffixed `<tag>` multi-arch manifest with `docker buildx imagetools create`. The base tag scheme matches `Harbor_Build_Base_Tag` (`dev` on main, the release version on `release-*`). The base-manifest step is gated on `BUILD_BASE_IMAGE.outputs.build_base` since suffixed tags only exist in runs that rebuilt bases.

After this, pulling `harbor-*-base:dev` resolves the correct architecture via the manifest list on both amd64 and arm64.

### Relation to #23347

#23347 proposes `max-parallel: 1` for the same symptom. Serializing the matrix makes the overwrite deterministic (the tag would always end up arm64-only, since arm64 runs last) but still publishes a single-arch tag and roughly doubles workflow wall-clock. This PR removes the overwrite instead: both arches publish suffixed tags concurrently and the manifest list serves both.

### Verification

- `make -n` dry-run of `_build_base` for `ARCH=amd64` and `ARCH=arm64` confirms: local build tag stays `harbor-db-base:dev`, push becomes `harbor-db-base:dev-amd64` / `dev-arm64`, and the component image build still consumes the local unsuffixed tag.
- The `CREATE_MANIFEST` `imagetools create` pattern is the one already in production for the component images (e.g. `harbor-core:dev` is amd64+arm64 today).
- The full push flow validates post-merge on `main` (the workflow is push-only and needs the Docker Hub secrets).

# Issue being fixed

Follow-up to #23324 / #23334. Alternative to #23347.

- [x] Label the PR as needed. `release-note/infra`

```release-note
Publish the goharbor/harbor-*-base images as multi-arch (amd64+arm64) manifests instead of overwriting the tag per architecture.
```

### DIFF
--- a/.github/workflows/build-package.yml
+++ b/.github/workflows/build-package.yml
@@ -150,14 +150,19 @@ jobs:
       - name: Save repo list
         run: |
           docker images --format '{{.Repository}}' | grep '^goharbor/' | grep -v '\-base' | grep -v '^goharbor/photon$' | grep -v '^goharbor/spectral$' | grep -v '^goharbor/swagger$' | grep -v '^goharbor/mockery$' | sort -u > "$GITHUB_WORKSPACE/_repos_${ARCH}.txt"
+          # base images are pushed arch-suffixed by the _build_base make macro
+          # and combined into multi-arch manifests by CREATE_MANIFEST
+          docker images --format '{{.Repository}}' | grep '^goharbor/.*-base$' | sort -u > "$GITHUB_WORKSPACE/_base_repos_${ARCH}.txt"
       - name: Upload repo list
         uses: actions/upload-artifact@v4
         with:
           name: repos-${{ matrix.arch }}
-          path: ${{ github.workspace }}/_repos_${{ matrix.arch }}.txt
+          path: |
+            ${{ github.workspace }}/_repos_${{ matrix.arch }}.txt
+            ${{ github.workspace }}/_base_repos_${{ matrix.arch }}.txt
 
   CREATE_MANIFEST:
-    needs: [BUILD_PACKAGE]
+    needs: [BUILD_BASE_IMAGE, BUILD_PACKAGE]
     runs-on: ubuntu-latest
     permissions:
       contents: read
@@ -181,6 +186,13 @@ jobs:
           else
             echo "image_tag=${version}" >> $GITHUB_OUTPUT
           fi
+          # base images use BASEIMAGETAG (see Harbor_Build_Base_Tag in BUILD_PACKAGE):
+          # "dev" on main, the release version on release-* branches
+          if [[ $target_branch == "release-"* ]]; then
+            echo "base_tag=${version}" >> $GITHUB_OUTPUT
+          else
+            echo "base_tag=dev" >> $GITHUB_OUTPUT
+          fi
       - name: Login to Docker Hub
         run: echo "${{ secrets.DOCKER_HUB_PASSWORD }}" | docker login -u "${{ secrets.DOCKER_HUB_USERNAME }}" --password-stdin
       - name: Create multi-arch manifests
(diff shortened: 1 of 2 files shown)


Vad1mo:
**Why not just build multi-arch with buildx straight away?**

`buildx build --platform amd64,arm64 --push` is the cleaner model, and it's what we did for `goharbor/photon:5.0` in #23334. But for the rest of the pipeline it loses to the suffix+manifest pattern on three concrete grounds:

**1. The matrix jobs must build the images locally anyway (for the installers).**
`make package_offline` does `docker save` of the locally-built images into the per-arch offline installer tarball (`Makefile:511` → `harbor-offline-installer-*-{amd64,arm64}.tgz`). That's the workflow's primary product. So the per-arch native builds exist regardless; a separate multi-arch buildx job would *duplicate* every image build, not replace it. The `imagetools create` stitching costs seconds — it only shuffles manifests, no rebuild.

**2. QEMU emulation cost on the non-native half.**
A buildx multi-platform build on one runner emulates the other arch:
- `harbor-valkey-base` compiles valkey from source on arm64 — minutes natively, plausibly 30–60+ min under QEMU.
- Component images embed Go binaries (core, jobservice, registry, exporter…) compiled natively per-runner via the golang build image, plus the portal's Node/Angular build. Under emulation these get dramatically slower; the photon base was only viable because it's pure `tdnf install`.
- Today both arches build natively in parallel (`ubuntu-latest` + `ubuntu-24.04-arm`) — that's the fastest configuration available. Single-runner buildx serializes one arch through emulation.

**3. The blast radius.**
Going buildx-everywhere means restructuring `make package_offline/online` (which assumes local single-arch images), the base/component interleaving in `make/photon/Makefile`, and the release-branch flow (versus a 2-file change that reuses the pattern already proven for the 14 component images.)

wy65701436 (approved):
lgtm, thanks.

stonezdj (approved):

Issue #22545, opened 2025-11-05
While pulling images via habor proxy, the images are pulled without tagging


rajkumar-ja:
While pulling hashicorp/vault and hashicorp/vault-k8s images via docker hub proxy it is pulling the image but not with the tagging.

ex: 

harbor.company.com/proxy-testing/hashicorp/vault-k8s:1.5.0
harbor.company.com/proxy-testing/hashicorp/vault:1.18.1

This happens to postgres, python even latest is not reflecting in the repo

rajkumar-ja:
we are running with harbor v2.10.3

stonezdj:
These two images are image indexes, it is possible that image is pulled but the tag is not created. it is a limitation, see: https://github.com/goharbor/harbor/issues/21454

stonezdj:
The solution to fix this issue, it maybe provide an option when creating proxy cache project.

cache multi-arch image for image index

when it is enabled, once the image index is pulled all content will be cached and the tag will also be created.

SA-KIRA:
I have the same issue

stonezdj:
Dup to #20920

PR #23355, opened 2026-06-12
Fix proxy cache manifest list reconcile

### LINKED ISSUE #18335: ProxyCache project doesn't tag saved artifacts
If I pull a new image (for example: _harbor.hrp.extbrain.net/quay_proxy/argoproj/argocd:v2.6.3_) through ProxyCache Project I expect the image is tagged in Harbor after it is pulled and cached. Currently the image is saved without any tag. 

As a container runtime I use _containerd_
1) Prepare Registry Endpoint in Harbor, Provider: Quay, Endpoint: https://quay.io, Access ID and Access Secret is left empty.
2) Prepare ProxyCache Project on this Registry Endpoint with name _quay_proxy_
3) Pull image: `sudo ctr i pull --user admin:<admin_pass> <harbor_url>/quay_proxy/argoproj/argocd:v2.6.3`

After the image is pulled, it is saved in the created ProxyCache Project but without tag v2.6.3
If I Use Docker for pulling the image everything is ok.

The Log message in harbor-core which is probably connected with this is:
```
2023-03-10T11:54:04Z [DEBUG] [/server/middleware/repoproxy/proxy.go:308]: Failed to ensure tag {Repository:quay_proxy/argoproj/argocd Reference:v2.6.3 ProjectName:quay_proxy Digest: Tag:v2.6.3 BlobMountRepository: BlobMountProjectName: BlobMountDigest:} , error the artifact is not ready yet, failed to tag it to v2.6.3`
```
**Versions:**
- Harbor version: 2.7.1
- containerd version: 1.6.8

I compared logs in Harbor-core in case of using Docker and in case of using Containerd.
Significant log message which is logged in case of using containerd and not Docker might be:
```
2023-03-10T11:17:40Z [DEBUG] [/server/middleware/security/unauthorized.go:28][requestID="64018b0b-0684-4f01-871d-2456b0b20c8e"]: an unauthorized security context generated for request GET /v2/quay_proxy/argoproj/argocd/manifests/sha256:0fd690bd7b89bd6f947b4000de33abd53ebcd36b57216f1c675a1127707b5eef
2023-03-10T11:17:40Z [DEBUG] [/lib/http/error.go:61]: {"errors":[{"code":"UNAUTHORIZED","message":"authorize header needed to send HEAD to repository: authorize header needed to send HEAD to repository"}]}
```
I did some research on it and I found out that the containerd didn't set any Authorization Header in the HTTP Request. 

In Postman:
I sent GET request to `https://<harbor_url>/v2/quay_proxy/argoproj/argocd/manifests/sha256:0fd690bd7b89bd6f947b4000de33abd53ebcd36b57216f1c675a1127707b5eef` with No Authorization Header. The response is

But if I made up any Basic Authorization Header with any credentials in it I got a correct response:

I also tried to send a request to quay.io directly and without any Authorization Header and the response was correct too unlike Harbor. 

Code snippet which might be relevant to this (?) https://github.com/goharbor/harbor/blob/a4c577f9ec4f18396207a5e686433a6ba203d4ef/src/server/middleware/v2auth/auth.go#L65-L78

I am not sure if this the cause of the proxy cache tagging issue of course.

I can give any additional information or log messages if needed.
Thank you.

### LINKED ISSUE #20920: Harbor Proxy Cache not creating additional Multi-Architecture Images under OCI Image Index

We are currently utilizing Harbor’s proxy-cache capability. Currently, we are running into issues where when pulling images with the same tag/version, but *different* architectures, it won’t “combine” the two images under the same tag/OCI Image Index.

Using a redis image as an example, I first pulled an image with the tag: 7.4.0 and specified the platform: “linux/arm64” using the following command:

“docker pull --platform linux/arm64 [server name]/ext.hub.docker.com/library/redis:7.4.0”

This created the following item in the Harbor UI. It first appeared as an untagged artifact immediately upon creation and after approximately 7 minutes an OCI Image Index was created and the original untagged artifact was put inside of that “folder”. This delayed creation and tagging is also an issue we are encountering, so insights into how to make this OCI Index creation and tagging more immediate would be greatly appreciated.
screenshot-1
 
Clicking inside of that OCI Index folder I see our image and the architecture that we specified.
screenshot-2
 
Which looks as expected => We now have a structure that mirrors Docker Hub’s where we have a tag/version at the top level, and all of the various OS/ARCH options below that.
screenshot-3
 
But when I try to pull a different architecture for the same tag through the proxy-cache (in this case: amd64), it does not add this additional architecture under the previously created 7.4.0 tag/OCI Image Index like I would have expected. Running the following command:

“docker pull --platform linux/amd64 [server name]/ext.hub.docker.com/library/redis:7.4.0”

This pull created a new item on the top level of the repository.
screenshot-4
 
Clicking into that new artifact’s SHA, you can see that it has all of the information that it should need to determine its OS/ARCH and version:
screenshot-5
screenshot-6
 
And yet it did not get added under our first item’s OCI Image Index. Instead it got created as a new artifact at the top level of the repository and didn’t receive a tag. Not receiving a tag is likely because our first item got the 7.4.0 tag, and since the tag is unique, our second item didn’t get a tag.

**Versions:**
- harbor version: 2.10.1

**Additional context:**
This secondary pull of an architecture that already has an OCI Index and tag also doesn’t seem to “fire” a webhook notification for a new image as we would expect as well. This *may* be a secondary effect of the issues highlighted above, where not having a tag or not going under the proper OCI Index might be the reason for this webhook not to fire when we would have expected it to.

Since Harbor is handling the creation of these items, it is hard to deduce or debug why these two images aren’t combining under the same OCI Index. Any insights into what is going wrong or how to best debug or find logs would be greatly appreciated!

### LINKED ISSUE #21454: Docker.io proxy cache isn't tagging official images
When pulling an image from docker.io, all images should be tagged.  However, we're seeing that any "official" images from docker.io, that contain "library" in the image path are not getting tagged

1. Create a Proxy Cache repository pointed at `docker.io`
2. Pull an "official" docker.io image, which contains "library" in the path or implied by not having an organization tag.  (eg. ${harbor_url}/docker.io/library/httpd:2.4.41)
3. Check the docker.io proxy, locate the pulled image and check that the tag does / does not exist

**Versions:**

- harbor version: v2.12.2-73072d0d (also failed on previous versions of Harbor)
- High-availability / multi-az deployed Helm Chart, using AWS RDS, S3, and Elasticache.

**Additional context:**

- When pulling an image from docker.io, if library isn't specified (eg. ${harbor_url}/docker.io/httpd:2.4.41) it is still implied and will show up with "library" in the path for the image.

- When pulling organizational image (non-official docker.io images), such as ${harbor_url}/docker.io/cypress/included:13.13.0, the image will get tagged

### LINKED ISSUE #21818: Harbor proxy repository doesn't make index manifest

- Expected: Image index manifest is created normally.
- Actual: Image index is not created.
Image

1. The issue is more easily reproduced when the image size is larger.
2. Push an amd64 image to the repository that serves as the proxy target (e.g., ECR), and push the index.
3. Pull the image from the Harbor server (e.g., `docker pull {harbor_server}/ecr_proxy/image:target`).
4. Log in to Harbor and check the image in the proxy repository.

**Versions:**
- harbor version: 2.12.2

**Additional context:**
- **Harbor proxy behavior:** When a manifest-related request comes in, the system checks whether the data is present in the cache. If the data is not in the cache, it asynchronously pushes the manifest to the Harbor repository. (https://github.com/goharbor/harbor/blob/main/src/controller/proxy/controller.go#L228)
- If the user sends a request for the index, the image index manifest is reconstructed by selecting only the manifests that exist locally (in the Harbor repository) and then pushed. If a manifest does not exist locally, the push fails. (https://github.com/goharbor/harbor/blob/main/src/controller/proxy/manifestcache.go#L145)
- The image manifest can only be pushed if all layers are present in the repository. (https://github.com/goharbor/harbor/blob/main/src/controller/proxy/manifestcache.go#L203)
- When a user sends a pull to harbor, the index is requested first, so there is a high probability that the image manifest does not exist locally during the push process in harbor. Moreover, the larger the image, the longer the push to the local repository takes, which increases the chance of failure.
- Since the process goes from cache to push, the index is not regenerated in the database on subsequent pull requests (because the image index has already been cached). This behavior appears to be related to [this issue](https://github.com/goharbor/harbor/issues/20920).

**Suggested modification:**
- Even if the manifest under the index does not exist, the system should continue using the existing retry logic (which currently retries for 20 * 20 seconds).
- https://github.com/goharbor/harbor/blob/main/src/controller/proxy/manifestcache.go#L145

matao-works:
## Summary

Fix proxy-cache multi-arch parent manifest handling when the parent index is pulled before its child manifests are available locally.

Harbor currently caches the parent manifest list payload, retries the initial local push for a bounded time, and then gives up permanently if no child manifest is available yet. Later child pulls never revisit that parent manifest list, which leaves proxy-cache repositories with a mix of symptoms:

- missing OCI index / manifest list artifacts
- child artifacts present without the expected tag
- repeated reports that Docker Hub official images or multi-arch images are cached but not tagged

This change makes parent manifest lists eventually consistent instead of one-shot best effort.

## Root cause

The current flow has two gaps:

1. `ManifestListCache.push` only retries while servicing the original parent pull request. If no child manifest is available locally during that window, Harbor stores the parent payload in cache but drops the local materialization attempt entirely.
2. When the parent manifest list is trimmed to only locally available child manifests, the `TrimmedManifestlist` cache entry is written using the artifact info from context instead of the original parent digest that `EnsureTag` later looks up. That breaks the digest handoff needed to attach the final tag consistently.

## What this PR changes

- Track pending parent manifest lists in cache, keyed by repository + child digest + parent digest.
- When a child manifest is cached successfully, scan and reconcile any pending parent manifest lists that depend on that child.
- Persist trimmed digest mappings against the original parent digest so `EnsureTag` can resolve the final local artifact.
- Always push the reconciled parent manifest list by digest, and by tag when a tag is available.
- Keep pending parent records when only part of the child set is available, so later child pulls can continue reconciling the same parent index.

## Related issues

This addresses the same underlying proxy-cache failure mode reported in:

- Fixes #21818
- Fixes #20920
- Fixes #21454
- Fixes #18335

It is also relevant to the symptoms described in #22545, where cached artifacts could exist without the expected tag.

## Validation

Ran:

- `go test ./controller/proxy -run TestCacheTestSuite -count=1 -v`
- `go test ./controller/proxy -count=1`

Added regression coverage for:

- registering pending parent manifest lists when no child is ready yet
- pushing the trimmed parent manifest list by digest and tag
- reconciling pending parents after a child manifest is cached


blankirigaya:
below are my finding after running this pr locally and testing against expected behavior for https://github.com/goharbor/harbor/issues/21454

---

## Test case

Pulled multi-arch Docker Hub images through a proxy cache project, including:

library/nginx:latest
library/httpd:2.4.41

---

## Before pr behavior
When pulling an image from docker.io, all images should be tagged. However, in reality it is becoming that any images from docker.io, that contain library/ in the image path are not getting tagged at all both for parent and child artifact.

image

---

## After pr Behavior 
My understanding is that this PR fixes proxy-cache reconciliation for multi-arch images by allowing parent manifest lists to be revisited after child manifests are cached. Previously, Harbor could end up with cached child manifests but no properly reconciled or tagged parent manifest list because the reconciliation attempt was effectively one-shot. This change makes parent manifest handling eventually consistent, preserves the digest mapping required for correct tag attachment, and enables the parent manifest list artifact to receive and retain the expected tag once reconciliation succeeds.


image

---

## Additional finding/things to look at 

Across multiple test runs on pr, the behavior was not consistently reproducible. In multiple cases, Harbor only displayed child manifests and no tagged parent manifest list artifact was created, which appears similar to the original issue where reconciliation of the parent manifest list did not complete successfully.

image


It would be helpful to clarify the expected post-fix behavior for proxy-cached multi-arch images. Should Harbor always materialize a tagged parent manifest list artifact once the required child manifests become available locally?

---

that is all of the relevant findings after running the pr locally one final suggestion for better testing would be to add more logging around parent manifest reconciliation, as it was difficult during testing to determine whether reconciliation was not being triggered or was failing partway through.  


---

If any future changes are made to this PR, I would be happy to test them locally and share additional findings and feedback.

matao-works:
Thanks for testing this locally and for sharing the detailed findings.

I re-reviewed the PR after this comment, including the async reconciliation path and the tests. I also re-ran the proxy controller tests:

- `go test ./controller/proxy -count=1` passed: 46 tests
- `go test ./controller/proxy -count=5` passed: 230 total test runs
- `go test -race ./controller/proxy -count=1` passed

A key point for validation is that this proxy-cache path is eventually consistent, not immediately consistent.

The current implementation has two async wait windows:

- child manifest caching waits up to `10 * 20s = 200s`
- parent manifest list / OCI index reconciliation waits up to `20 * 20s = 400s`

Because of that, checking Harbor UI/API/DB too early can legitimately show only the child manifest while the tagged parent manifest list artifact has not been materialized yet. That state alone does not necessarily mean reconciliation failed.

The expected post-fix behavior is:

1. the child manifest is cached first
2. the parent manifest list / OCI index is revisited later
3. once at least one required child manifest is available locally, Harbor can persist a parent index containing the locally available children and attach the tag to that parent artifact
4. if the parent is still partial, pending reconciliation state is kept so later child arrivals can reconcile it again

For reliable validation, I recommend using a never-before-cached multi-arch repository, then waiting for the full async reconciliation window before checking the UI/API/DB state.

I also validated this end-to-end against clean proxy-cache repositories in a production-like Harbor v2.15.0 deployment with this patch backported:

- `library/ruby:3.3-slim-bookworm`
- `library/perl:5.40-slim-bookworm`

For both images, after the async window completed, Harbor had:

- repository row created
- child artifact persisted
- parent OCI index artifact persisted
- tag attached to the parent artifact
- repeat pulls no longer logged `is not found in proxy cache, fetch it from remote repo`

So the behavior you observed where only child manifests are visible can happen during the reconciliation window, especially with multi-arch indexes that include many platforms and attestation manifests.

I agree with your suggestion about observability. Additional debug logs around parent manifest-list reconciliation would make this flow easier to troubleshoot, especially to distinguish "still waiting for reconciliation" from "reconciliation failed".

matao-works:
I pushed a follow-up update to make the PR easier to review and validate:

- merged the latest upstream `main`, so the PR branch is no longer behind
- added focused tests for the manifest-list cache handler registration and manifest-list payload/content-type caching when the parent digest is derived from payload
- re-ran the proxy controller package tests after the update

Local validation completed:

- `go test ./controller/proxy -count=1` passed: 48 tests
- `go test ./controller/proxy -count=5` passed: 240 total test runs
- `go test -race ./controller/proxy -count=1` passed
- `go vet ./controller/proxy` passed
- `make lint` passed with 0 issues after generating the API server code required by the lint target

DCO is also passing on the updated branch.

fabio-t:
This would be very useful for us - is there anything blocking review?

stonezdj (approved):

src/controller/proxy/manifestcache.go
```diff
@@ -106,6 +113,91 @@ func (m *ManifestListCache) cacheTrimmedDigest(ctx context.Context, newDig strin
 	log.Debugf("Saved key:%v, value:%v", key, newDig)
 }
 
+func (m *ManifestListCache) getTrimmedDigest(ctx context.Context, originDig string) (string, error) {
```
wy65701436, line 116: Is this a race condition?If I understand correctly.

When different child artifacts are pulled/pushed in parallel for the first time, will the last successful write overwrite the previously generated index in the DB?

1. When x64 is pulled successfully, it is saved to the DB with an index containing only x64.
2. When other child artifacts are pulled in parallel, a race condition occurs where the last successful write overwrites the previous entry in the DB.

cc @stonezdj

PR #23368, opened 2026-06-13
fix(proxy-cache): redirect tags/list for Docker Hub library images

bartlaarhoven:
## Problem

Listing tags for a Docker Hub official ("library") image through a proxy cache
returns an empty list:

    GET /v2/<proxy>/busybox/tags/list  ->  {"name":"...","tags":[]}

while the same repository with the explicit `library/` prefix works:

    GET /v2/<proxy>/library/busybox/tags/list  ->  full upstream tag list

This breaks every tool that discovers versions through the registry tags/list
API. For example, Renovate can no longer propose image updates for any official
image pulled through the proxy cache (busybox, postgres, nginx, mariadb, redis,
…), because it sees zero tags.

## Root cause

Docker Hub official images live under the implicit `library/` namespace. The
docker client expands `busybox` -> `library/busybox` transparently, but the
registry tags/list API does not.

`ManifestMiddleware` and `BlobGetMiddleware` already handle this: for a
single-segment repository on a Docker Hub registry they detect the
default-library case (`defaultLibrary`) and issue a 301 redirect to the
`library/`-prefixed path (`defaultManifestURL` / `defaultBlobURL`).
`TagsListMiddleware` was missing the equivalent handling, so the request was
proxied upstream as `busybox` (which Docker Hub does not resolve) and an empty
tag list was returned.

The asymmetry is observable today on a Docker Hub proxy-cache project:

    GET /v2/<proxy>/busybox/manifests/<tag>  ->  301 Location: /v2/<proxy>/library/busybox/manifests/<tag>
    GET /v2/<proxy>/busybox/tags/list        ->  200 {"tags":[]}

## Fix

Add the same `defaultLibrary` redirect to `TagsListMiddleware`, pointing at a
new `defaultTagsListURL` helper. The original query string is preserved so the
`n` / `last` pagination parameters keep working after the redirect.

## Tests

- `TestDefaultTagsListURL` — unit test for the URL builder, including query-string preservation.
- `TagsListMiddlewareTestSuite` — verifies that a single-segment Docker Hub image
  is redirected (301 + `Location`, query preserved) and that a multi-segment
  image is not.

Relates to #21454.

### DIFF
--- a/src/server/middleware/repoproxy/proxy.go
+++ b/src/server/middleware/repoproxy/proxy.go
@@ -165,7 +165,7 @@ func ManifestMiddleware() func(http.Handler) http.Handler {
 	})
 }
 
-func defaultLibrary(ctx context.Context, registryID int64, a lib.ArtifactInfo) (bool, string, error) {
+var defaultLibrary = func(ctx context.Context, registryID int64, a lib.ArtifactInfo) (bool, string, error) {
 	if registryID <= 0 {
 		return false, "", nil
 	}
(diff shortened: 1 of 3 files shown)


bartlaarhoven:
I think this is still relevant?

PR #23374, opened 2026-06-14
Set blob response headers before streaming the body

### LINKED ISSUE #23373: Proxy-cache blob responses omit Content-Length (headers set after the body), enabling truncated-fill cache poisoning

  When the proxy-cache (pull-through) feature serves a blob fetched from the upstream registry, the response should carry `Content-Length` (and `Docker-Content-Digest`/`ETag`). The blob size is known up
  front from the descriptor.

  What actually happens: in `handleBlob` (`src/server/middleware/repoproxy/proxy.go`), `setHeaders()` (which sets `Content-Length`, `Docker-Content-Digest`, `ETag`) is called after `io.CopyN(w, reader,
  size)` has already streamed the body. The response is committed on the first write, so `setHeaders` ends up a no-op. As a result, proxied blobs larger than Go's `net/http` pre-chunking buffer (2048 bytes)
  go out with `Transfer-Encoding: chunked` and no `Content-Length`, and the `Docker-Content-Digest`/`ETag` headers are never sent.

  Code (v2.13.0, also present on current `main`), `handleBlob`:

  ```go
  size, reader, err := proxyCtl.ProxyBlob(ctx, p, art)   // size is known here
  if err != nil { return err }
  defer reader.Close()
  written, err := io.CopyN(w, reader, size)              // body streamed first
  if err != nil { return err }
  if written != size { return errors.Errorf("size mismatch ...") }
  setHeaders(w, size, "", art.Digest)                    // Content-Length set here, too late
  ```

  Why it matters: with no `Content-Length`, a truncated upstream read (for example the remote registry stalls or times out mid-stream) is delivered as a complete-looking chunked `200`. Go still writes the
  terminating `0\r\n\r\n` chunk when the handler returns, and `SendError` can't change the status because `200` was already sent. A caching proxy in front of Harbor (we use nginx `proxy_cache` for blob
  serving) then stores the truncated body as a complete blob. Every later pull on that node gets the short blob (digest mismatch, failed pulls) until the entry is evicted. Even with no caching layer,
  proxied blob responses are still missing the `Content-Length` and digest headers they should have.

  1. Configure a proxy-cache project pointing at a remote registry.
  2. Pull an image whose layers are larger than 2 KB through the proxy-cache.
  3. Inspect the blob response from `core` (for example `curl -sI` through the chain, or look at a fronting cache entry): `Transfer-Encoding: chunked`, no `Content-Length`, no
  `Docker-Content-Digest`/`ETag`.
  4. To see the poisoning, make the upstream fetch stall or time out mid-stream. The partial body comes back as a complete chunked `200`, a fronting `proxy_cache` stores it as a complete blob, and later
  pulls on that node return the truncated blob.

  **Versions:**

  - harbor version: 2.13.0 (same code on current `main`)
  - docker engine version: N/A (running on Kubernetes via the Bitnami chart; the bug is in Harbor source and is independent of the deployment)
  - docker-compose version: N/A

  **Additional context:**

  The root cause is the header/body write order in `handleBlob`. Setting the headers before streaming the body fixes it: `net/http` then frames the response with the known `Content-Length`, and a truncated
  read shows up as a short body that clients and caches reject instead of being accepted and cached as complete. I have a fix and unit tests ready and will open a PR referencing this issue.

  For reference, behavior confirmed with Go 1.26: a 100-byte body gets an automatic `Content-Length` (it fits Go's pre-chunking buffer), a 5000-byte body is sent chunked with no `Content-Length`, and a
  truncated chunked stream still ends with a valid terminating `0` chunk.

shrenikgala:
In `handleBlob` (`src/server/middleware/repoproxy/proxy.go`), `setHeaders()` (which sets `Content-Length`, `Docker-Content-Digest`, `ETag`) was called after `io.CopyN(w, reader, size)` had already
  streamed the body. The response is committed on the first write, so `setHeaders` was a no-op: proxied blobs larger than Go's `net/http` pre-chunking buffer (2048 bytes) were sent with `Transfer-Encoding:
  chunked` and no `Content-Length`, and the digest/ETag headers were never sent.

  With no `Content-Length`, a truncated upstream read (for example when the remote registry stalls or times out mid-stream) is delivered as a complete-looking chunked `200`, which a caching proxy in front
  of Harbor can store as a complete blob. Later pulls on that node then receive the truncated blob until the cache entry is evicted.

  This change moves the header write before the body is streamed (extracted into a small `serveBlob` helper). With the length declared up front, `net/http` frames the response with the known
  `Content-Length`, so a short read is surfaced as a truncated body that clients and intermediary caches reject instead of being accepted as complete.

  Added `TestServeBlob`, which drives the helper through a real `httptest.Server` and covers two cases:

  - a complete blob larger than 2 KB is sent with a `Content-Length`, is not chunked, and carries the `Docker-Content-Digest`/`ETag` headers
  - a short read (fewer bytes than the declared size) is surfaced to the client as `io.ErrUnexpectedEOF` instead of a complete response

The post-first-byte error path no longer goes through `SendError`. `serveBlob` now branches on where the failure happens relative to the first write:

  - `err == nil`: full success (`io.CopyN` returns a nil error only when exactly `size` bytes were copied).
  - failure with `written == 0`: headers are not committed yet, so it clears the blob headers and returns the error, and the middleware writes a clean `500`.
  - failure with `written > 0`: the status and `Content-Length` are already committed, so it logs the error and returns `nil`. Nothing is appended to the body, so the response stays short of
  `Content-Length`, `net/http` closes the connection, and the client (and any cache) sees a truncated body. This removes the boundary case where an appended error payload could pad the body up to
  `Content-Length` and be cached as a complete but corrupted blob.

  Added a test for the previously uncovered path: a reader that yields some bytes and then fails. It asserts `serveBlob` returns `nil`, the client reads `io.ErrUnexpectedEOF`, and the delivered bytes equal
  exactly the streamed bytes with nothing appended.

  One design note, and I am open to your preference here: on the `written > 0` path I log the error and return `nil`, since once the `200` is committed there is no faithful way to surface it as an HTTP
  error without writing into the blob body. An alternative is to `panic(http.ErrAbortHandler)` there (`handleBlob` already recovers and suppresses it), which also aborts the connection. Happy to switch to
  that if you prefer an explicit abort over returning `nil`.
  
  # Issue being fixed
  Fixes #23373

  - [x] Label the PR as needed. "release-note/ignore-for-release, release-note/new-feature, release-note/update, release-note/enhancement, release-note/community, release-note/breaking-change,
  release-note/docs, release-note/infra, release-note/deprecation"


stonezdj (approved):
LGTM

shrenikgala:
@stonezdj thanks for your review, what is needed on my end to get this PR over the line, I see the merging criteria requires 2 additional approvers

Vad1mo (approved):

PR #23376, opened 2026-06-14
fix: skip corrupted encrypted config values on decryption failure

rossigee:
## Description
When decrypting password-type config fields fails (e.g., due to encryption key mismatch), the code was logging an error but then silently including the encrypted garbage in the result map. This corrupted value would then be cached and used by the application, causing authentication failures.

For example, an OIDC client secret that failed to decrypt would return the encrypted bytes as a string, which would then be sent to the OIDC provider, resulting in 'unauthorized_client' errors.

## Fix
Skip corrupted password fields during decryption failure, allowing the system to fall back to defaults or environment variables.

## Testing
- Verified OIDC authentication works when client secret fails initial decryption
- Confirmed skipping corrupted fields allows proper fallback behavior


wy65701436 (approved):
lgtm, thanks.

chlins (approved):
lgtm

PR #23420, opened 2026-06-22
fix(deps): sync package-lock.json to resolve missing yaml@2.9.0 entry

zoom2ashish:
fix(deps): sync package-lock.json to resolve missing yaml@2.9.0 entry
package-lock.json was out of sync with package.json — yaml@2.9.0 was
declared as a dependency but missing from the lock file, causing
`npm ci` to fail with EUSAGE in CI/CD. Regenerated the lock file via
`npm install` to add the missing yaml@2.9.0 resolution.

Signed-off-by: Ashish Patel <zoom2ashish@gmail.com>


wy65701436 (approved):
lgtm

PR #23425, opened 2026-06-23
fix: safely handle missing dry_run execution attribute

ashnaaseth2325-oss:
## SUMMARY

This PR fixes `StopScanArtifact()` to use the already validated and defaulted `scanType` value when stopping a scan. The change affects the scan API handler in `src/server/v2.0/handler/scan.go` and aligns its behavior with the existing `ScanArtifact()` implementation.

## FIX

```go 
// Before
if err := s.scanCtl.Stop(ctx, curArtifact, params.ScanType.ScanType); err != nil {
    return s.SendError(ctx, err)
}

// After
if err := s.scanCtl.Stop(ctx, curArtifact, scanType); err != nil {
    return s.SendError(ctx, err)
}
```

### DIFF
--- a/src/controller/retention/controller.go
+++ b/src/controller/retention/controller.go
@@ -343,10 +343,12 @@ func convertExecution(exec *task.Execution) *retention.Execution {
 		EndTime:   exec.EndTime,
 		Status:    exec.Status,
 		Trigger:   exec.Trigger,
-		DryRun:    exec.ExtraAttrs["dry_run"].(bool),
 		Type:      exec.VendorType,
 	}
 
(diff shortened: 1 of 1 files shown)


ashnaaseth2325-oss:
Could a maintainer please add the appropriate release-note label? This is a small bug fix that safely handles a missing dry_run execution attribute in retention execution conversion.

chlins:
@ashnaaseth2325-oss Hi, your code changes don't seem to align with the PR description. Which issue are you trying to resolve?

Vad1mo:
I agree; I don't understand the problem you are trying to address here @ashnaaseth2325-oss

PR #23463, opened 2026-06-29
Constrain /registries/ping to the saved URL for existing registries

Aloui-Ikram:
**What this PR does**

`PingRegistry` lets a caller reference an existing registry by `id` and, in the same request, override its stored connection settings (`url`, `insecure`, `ca_certificate`). Because the ping loads the registry's saved credentials (never exposed through the API) and then sends them to the resolved endpoint, a user with only `read:registries` could point the ping at their own server and exfiltrate those credentials, or downgrade/redirect TLS to have them captured on-path.

This change makes an existing registry's saved connection settings authoritative: when an `id` is supplied, any `url`, `insecure`, or `ca_certificate` in the request body is ignored and the ping runs against the stored configuration. Inline ping requests (no `id`) are unaffected and still honor all supplied fields.

**Behaviour**

- ping by `id` only → unchanged (uses saved settings)
- ping by `id` + `url`/`insecure`/`ca_certificate` overrides → overrides ignored, ping uses the saved settings
- inline ping (no `id`) → unchanged (uses supplied fields)

**Testing**

- Added `registry_test.go` unit tests: a ping by `id` with attacker-supplied `url`, `insecure`, and `ca_certificate` reaches the health check with the *saved* settings; an inline ping still honors the supplied URL.
- Verified the regression test fails against the previous override behaviour (the ping was reaching the attacker URL).


Vad1mo (approved):

bupd (approved):
/lgtm

Issue #22898, opened 2026-03-02
Failed to download trivy adapter in the build


stonezdj:
https://github.com/goharbor/harbor/actions/runs/22566968434/job/65365133539?pr=22885
```
08:15:08 Downloading Trivy scanner v0.69.1...
08:15:08 download https://github.com/aquasecurity/trivy/releases/download/v0.69.1/trivy_0.69.1_Linux-64bit.tar.gz
08:15:08 % Total % Received % Xferd Average Speed Time Time Time Current
08:15:08 Dload Upload Total Spent Left Speed
08:15:09 
 0 0 0 0 0 0 0 0 --:--:-- --:--:-- --:--:-- 0
 0 9 0 0 0 0 0 0 --:--:-- --:--:-- --:--:-- 0
08:15:09 curl: (22) The requested URL returned error: 404
```

abvwag:
https://github.com/aquasecurity/trivy/discussions/10265

Issue #22895, opened 2026-03-01
https://github.com/aquasecurity/trivy/


mgl-pub:
This repository is empty.
Care to check out the [GitHub Channel on YouTube](https://youtube.com/GitHub) while you wait?

TRIVY_DOWNLOAD_URL=https://github.com/aquasecurity/trivy/releases/download/$(TRIVYVERSION)/trivy_$(TRIVYVERSION:v%=%)_Linux-64bit.tar.gz

Aloui-Ikram:
## Update: Confirmed root cause + fix available

This is the result of a **supply chain attack** on the Trivy GitHub 
repository on March 1, 2026, by an automated AI bot (`hackerbot-claw`).

**Official incident report from Aqua Security:**
https://github.com/aquasecurity/trivy/discussions/10265

### What happened
- The attacker obtained a `GITHUB_TOKEN` with write permissions via a 
  compromised GitHub Actions workflow
- All GitHub Releases from v0.27.0 to v0.69.1 were **permanently deleted**
- Harbor's Makefile pins `TRIVYVERSION=v0.69.1` and downloads directly 
  from GitHub Releases → **any `make build` with `TRIVYFLAG=true` fails 
  with HTTP 404**

### Fix
Aqua Security has published **v0.69.2** as an emergency patch release 
(March 1, 2026 at 19:14 UTC). The proposed fix for Harbor is to bump:
```
TRIVYVERSION=v0.69.2
```

PR #23471, opened 2026-06-30
fix: build trivy scanner binary from source instead of release tarball

air1bzz:

Harbor builds the `trivy-adapter` image by downloading the Trivy scanner binary from a **GitHub release tarball**, at a version pinned per Harbor release in the Makefile (`TRIVY_DOWNLOAD_URL`).

Aqua Security **prunes old release assets** (and yanks releases during supply-chain incidents), keeping only a rolling window of recent versions. The git **tags stay**, but the Release objects and their tarballs are deleted — so the pinned URL 404s:

```
curl: (22) The requested URL returned error: 404
gzip: stdin: unexpected end of file
tar: Child returned status 1
make[1]: *** [make/photon/Makefile:147: _build_trivy_adapter] Error 1
```

This is silent for users pulling the prebuilt `goharbor/trivy-adapter-photon` images, but it **permanently breaks build-from-source of older Harbor tags**: a tag's Makefile is frozen, so once its pinned Trivy version is pruned, that tag becomes unbuildable. For example, v2.14.2 pins Trivy `v0.68.2`, whose entire `0.68.x` series has been removed from GitHub releases.

This keeps recurring and is currently fixed *reactively* by bumping the Trivy version forward (most recently the March 2026 supply-chain incident, #22895 / #22898, fixed by #22896).

## Fix

Build the Trivy scanner binary **from its git source tag**, mirroring how Harbor already builds the `registry` binary (`BUILDREG`) and the `scanner-trivy` adapter (`BUILDTRIVYADP`). The scanner is currently the *only* Trivy component sourced from a prebuilt release tarball — this makes it consistent with the rest.

Git **tags are never pruned** (only release *assets* are), so `git clone -b <tag> && go build` keeps older tags buildable.

- New `BUILDTRIVY` flag (default `true`) selects source-build; `BUILDTRIVY=false` falls back to the existing tarball download.
- New `make/photon/trivy-adapter/trivy-builder.sh` + `Dockerfile.trivy-binary`, structured exactly like `make/photon/registry/builder` + its `Dockerfile.binary`.
- The build reproduces Trivy's own `mage` build (`magefiles/magefile.go`): `CGO_ENABLED=0 GOEXPERIMENT=jsonv2`, ldflags `-X=github.com/aquasecurity/trivy/pkg/version/app.ver=<version>`, so `trivy --version` is unchanged.
- `TRIVYVERSION` is unchanged → no scanner behaviour change. The scanner binary still lands in `binary/trivy`, so the adapter `Dockerfile` is untouched.
- Harbor's `GOBUILDIMAGE=golang:1.26.3` matches Trivy v0.71.1's required `go 1.26.3` exactly.

## Verification

The Makefile recipe expands correctly in both modes:

```sh
make -f make/photon/Makefile -n _build_trivy_adapter TRIVYFLAG=true \
     TRIVYVERSION=v0.71.1 TRIVY_SRC=https://github.com/aquasecurity/trivy.git \
     GOBUILDIMAGE=golang:1.26.3 DOCKERNETWORK=default          # BUILDTRIVY=true → trivy-builder.sh
make -f make/photon/Makefile -n _build_trivy_adapter TRIVYFLAG=true BUILDTRIVY=false  # tarball fallback
```

The exact build command run by `Dockerfile.trivy-binary` was executed against the
`v0.71.1` source tag, and the resulting binary reports the expected version:

```sh
CGO_ENABLED=0 GOEXPERIMENT=jsonv2 go build -buildvcs=false \
  -ldflags "-s -w -X=github.com/aquasecurity/trivy/pkg/version/app.ver=0.71.1" \
  -o trivy ./cmd/trivy
./trivy --version
# => Version: 0.71.1   (matches TRIVYVERSION; statically linked, CGO_ENABLED=0)
```

The full containerized path is exercised by the existing build pipeline:

```sh
make build BUILDTARGET=_build_trivy_adapter TRIVYFLAG=true PULL_BASE_FROM_DOCKERHUB=true
docker run --rm --entrypoint /usr/local/bin/trivy goharbor/trivy-adapter-photon:dev --version
# => Version: 0.71.1
```

# Issue being fixed

Related to #22895 and #22898 (recurring Trivy release-asset unavailability during the build). This addresses the root cause for build-from-source of pinned/older tags rather than bumping the version forward.


Vad1mo:
This is a mess; there must be better ways than this shell hell. Any ideas?

air1bzz:
Fair point, and I agree the shell script is heavy. I kept it to stay consistent with the existing implementations in `make/photon/`: the `registry` and `scanner-trivy` adapter builders use the same clone-and-build-in-a-script pattern, so I followed the convention instead of introducing a different mechanism. But there's a much cleaner way and I'm happy to drop the script. Two options:

### Option A: `go install` one-liner (my preference)

Trivy is pure Go with no `replace` directives in its `go.mod`, so the whole script collapses into a single line inside the existing `$(GOBUILDIMAGE)` container:

```makefile
CGO_ENABLED=0 GOEXPERIMENT=jsonv2 GOBIN=$(DOCKERFILEPATH_TRIVY_ADAPTER)/binary go install \
  -ldflags "-s -w -extldflags '-static' -X github.com/aquasecurity/trivy/pkg/version/app.ver=$(TRIVYVERSION:v%=%)" \
  github.com/aquasecurity/trivy/cmd/trivy@$(TRIVYVERSION)
```

This is still a build from source: the Go module proxy serves checksummed source archives (verified via `sum.golang.org`), not binaries. It also keeps working after Aqua prunes the GitHub release assets, which is the root cause this PR addresses: only the git tags survive pruning, and the module proxy keeps the source. If we'd rather resolve straight from the git tag, `GOPROXY=direct` does the same with one line.

I checked this against the pinned `v0.71.1` tag (and `v0.68.2`, the oldest broken one):

- No `replace`/`exclude` directives in `go.mod` at either tag, so `go install ...@version` works.
- The flags above match Trivy's `goreleaser.yml` at both tags: `CGO_ENABLED=0`, `GOEXPERIMENT=jsonv2`, static linking, and `app.ver` set without the `v` prefix so `trivy --version` is correct.
- `go.mod` requires `go 1.26.3` at v0.71.1, which the current `GOBUILDIMAGE` (`golang:1.26.3`) already provides.

I ran it locally: the binary builds and `trivy --version` reports `Version: 0.71.1`.

One thing worth a code comment: the `app.ver` ldflags path and the build env (`GOEXPERIMENT`) have changed across Trivy releases before, so a future `TRIVYVERSION` bump should re-check `goreleaser.yml` at the new tag.

### Option B: multi-stage build in the trivy-adapter Dockerfile

Add a builder stage (`FROM ${golang_image} AS trivy-builder` plus the same `go install`) and `COPY --from` the binary into the adapter image. No Makefile changes, but it diverges from how the other binaries are built and makes Trivy the only component with its own in-Dockerfile builder stage.

Either option removes the shell script and the Copilot nits along with it. Let me know which you prefer and I'll rework the PR.

Vad1mo:
I would go B (with A inside) make sure we that the build is multi-arch for AMD and ARM

air1bzz:
Sorry for the long silence, I was on holiday and only got back to this now. Reworked as Option B with the `go install` from Option A inside it, and pushed to the same branch.

The builder script and its Dockerfile are gone, and so is the `BUILDTRIVY` flag. What is left is one builder stage in `make/photon/trivy-adapter/Dockerfile`:

```dockerfile
FROM ${golang_image} AS trivy-builder

ARG trivy_version
ARG TARGETARCH=amd64

ENV CGO_ENABLED=0
ENV GOOS=linux
ENV GOARCH=${TARGETARCH}
ENV GOEXPERIMENT=jsonv2

RUN GOBIN=/out go install \
    -ldflags "-s -w -extldflags '-static' -X github.com/aquasecurity/trivy/pkg/version/app.ver=${trivy_version#v}" \
    github.com/aquasecurity/trivy/cmd/trivy@${trivy_version}
```

plus `COPY --from=trivy-builder /out/trivy /usr/local/bin/trivy` in the final stage and two extra `--build-arg`s in the Makefile recipe. That removes the 51 lines of shell without adding a single new file, and it resolves the Copilot comments along with the script.

**Multi-arch.** The stage is shaped like `make/photon/exporter/Dockerfile`: no `--platform=$BUILDPLATFORM`, so the builder runs on the target platform, and `GOARCH` comes from `TARGETARCH`, which `make/photon/Makefile` already derives from `ARCH` next to `--platform`. That matches the native amd64/arm64 runner matrix in `build-package.yml`, so nothing is compiled under QEMU emulation. I did try the cross-compiling variant with `--platform=$BUILDPLATFORM`, but `go install` refuses `GOBIN` when cross-compiling (`cannot install cross-compiled binaries when GOBIN is set`), which forces an arch-dependent copy path, more shell than it saves.

**One deliberate removal worth flagging.** I dropped `TRIVY_DOWNLOAD_URL` and the `BUILDTRIVY=false` tarball fallback instead of keeping it behind a conditional `COPY`. Keeping the fallback means keeping alive the exact download path this PR exists to remove, and air-gapped builders are better served by pointing `GOPROXY` at an internal proxy than by mirroring a tarball. Happy to restore the escape hatch if you would rather keep it.

**Verified.**
- `docker build --target trivy-builder --build-arg golang_image=golang:1.26.3 --build-arg trivy_version=v0.71.1`: builds clean, `trivy --version` reports `Version: 0.71.1`, and the binary is statically linked and stripped.
- `make -f make/photon/Makefile -n _build_trivy_adapter ARCH=arm64` expands to `--platform=linux/arm64 --build-arg TARGETARCH=arm64 --build-arg golang_image=golang:1.26.3`.

The PR still needs a `release-note/*` label from a maintainer for `label_check` to go green, I cannot set it myself.

PR #23496, opened 2026-07-05
fix(registry): send fallback scope for base /v2 ping requests

### LINKED ISSUE #23466: Proxy cache endpoint health check fails for registry.istio.io
### Expected behavior

Harbor should successfully validate https://registry.istio.io as a generic Docker Registry endpoint when no credentials are configured.

The registry hosts publicly accessible images. For example, the following anonymous pull succeeds:

docker pull registry.istio.io/release/proxyv2:1.26.4

Consequently, Harbor should:

- consider the endpoint healthy;
- allow it to be registered as a remote registry;
- allow the endpoint to be used by a proxy cache project.

### Actual behavior

Harbor rejects the registry endpoint as unhealthy:

```
{"errors":[{"code":"BAD_REQUEST","message":"the registry is unhealthy"}]}
```

The Harbor core logs contain:

```
 [/pkg/reg/adapter/native/adapter.go:126]: failed to ping registry https://registry.istio.io: http status code: 400, body: {"error":"scope is required"}
```

### Prerequisites

- A Harbor instance with administrator access
- Outbound HTTPS access from harbor-core to registry.istio.io
- No credentials configured for the remote registry
- Docker, Podman, crane, or another OCI client

### Verify that the image is publicly pullable
```
docker logout registry.istio.io 2>/dev/null || true
docker pull registry.istio.io/release/proxyv2:1.26.4
```
Expected result:

```
1.26.4: Pulling from release/proxyv2
...
Status: Downloaded newer image for registry.istio.io/release/proxyv2:1.26.4
```
### Reproduce the failure through the Harbor UI

In the Harbor administration interface:

- Open Administration → Registries.
- Select New Endpoint.
- Select Docker Registry as the provider.
- Enter a name such as istio.
- Enter: https://registry.istio.io
- Leave Access ID and Access Secret empty.
- Click Test Connection or save the endpoint.

Observed result:

The registry is unhealthy

## Versions:

- harbor version: 2.14.3
- docker engine version: 28.5.1

Adityakumar37:

Harbor's `PingSimple()` health check (used when a registry endpoint has no credentials configured) sends an unauthenticated `GET /v2/` to the remote registry. When no resource-specific scope can be derived from that base path, `parseScopes()` previously returned `nil`, so the subsequent bearer token request was sent with no `scope` param at all.

Most token services tolerate a scope-less anonymous token request. `registry.istio.io`'s token service does not — it rejects it with `400 {"error":"scope is required"}`. Since `PingSimple()` only tolerates `401`/`403` as "registry is reachable," this `400` was treated as a genuine failure, so the endpoint was marked unhealthy even though it's fully pullable anonymously (`docker pull registry.istio.io/release/proxyv2:1.26.4` works fine).

This PR makes `parseScopes()` fall back to the same `registry:catalog:*` scope already used for catalog listing when no resource-specific scope can be derived from the path (i.e. for the base `/v2` ping). This gives strict token services a non-empty scope to evaluate, while registries that don't grant it will simply respond `401`/`403` as before — which `PingSimple()` already tolerates.

**Files changed:**
- `src/pkg/registry/auth/bearer/scope.go` — added the base-ping fallback case
- `src/pkg/registry/auth/bearer/scope_test.go` — updated the existing base-path assertion to expect the fallback scope instead of `nil`, and added `TestParseScopesBasePing` as a regression test

No changes were needed in `authorizer.go` (the existing scope-iteration loop in `fetchToken` already handles a single-element slice correctly) or `adapter.go` (`PingSimple`'s 401/403 tolerance is unchanged).

**Follow-up not included in this PR:** a mock-token-server integration test that exercises the full `GET /v2/` → `401` challenge → token fetch → `400`/`200` round trip end-to-end, rather than just the unit-level `parseScopes` behavior. Kept out for now to keep this PR small per the contributing guide; happy to add it here if a reviewer wants broader coverage.

# Issue being fixed
Fixes #23466

### DIFF
--- a/src/pkg/registry/auth/bearer/scope.go
+++ b/src/pkg/registry/auth/bearer/scope.go
@@ -113,7 +113,7 @@ func parseScopes(req *http.Request) []*scope {
 	// minimal registry-level scope used for catalog listing so the ping still
 	// succeeds; a registry that doesn't grant it will simply respond with
 	// 401/403, which PingSimple already treats as "reachable".
-	if path == "/v2" {
+	if strings.HasSuffix(path, "/v2") {
 		return []*scope{
 			{
 				Type:    scopeTypeRegistry,
--- a/src/pkg/registry/auth/bearer/scope.go
+++ b/src/pkg/registry/auth/bearer/scope.go
@@ -107,6 +107,21 @@ func parseScopes(req *http.Request) []*scope {
 			}}
 	}
 
-	// base or no match, return nil
+	// base ping request (e.g. GET /v2/): no resource-specific scope can be
+	// derived from the path, but some token services (e.g. registry.istio.io)
+	// reject a token request that carries no scope at all. Request the same
+	// minimal registry-level scope used for catalog listing so the ping still
+	// succeeds; a registry that doesn't grant it will simply respond with
+	// 401/403, which PingSimple already treats as "reachable".
+	if path == "/v2" {
+		return []*scope{
+			{
+				Type:    scopeTypeRegistry,
+				Name:    "catalog",
+				Actions: []string{scopeActionAll},
+			}}
+	}
+
+	// no match, return nil
 	return nil
 }
--- a/src/pkg/registry/auth/bearer/scope_test.go
+++ b/src/pkg/registry/auth/bearer/scope_test.go
@@ -35,7 +35,11 @@ func TestParseScopes(t *testing.T) {
 	// base
 	req, _ := http.NewRequest(http.MethodGet, "/v2/", nil)
 	scopes := parseScopes(req)
-	require.Nil(t, scopes)
+	require.Len(t, scopes, 1)
+	assert.Equal(t, scopeTypeRegistry, scopes[0].Type)
+	assert.Equal(t, "catalog", scopes[0].Name)
+	require.Len(t, scopes[0].Actions, 1)
+	assert.Equal(t, scopeActionAll, scopes[0].Actions[0])
 
 	// catalog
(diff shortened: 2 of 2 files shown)


Adityakumar37:
Thanks @Copilot both good catches, fixed :)

- `scope.go`: switched from exact `path == "/v2"` to `strings.HasSuffix(path, "/v2")`, so registries mounted under a URL prefix (e.g. `/prefix/v2`) also get the fallback scope instead of silently falling through to a scope-less token request.
- `scope_test.go`: added `TestParseScopesBasePingWithPrefix` to cover the prefixed base-ping case as a regression guard, alongside the existing `TestParseScopesBasePing`.

Ran the full package suite locally (`go test ./pkg/registry/auth/bearer/... -v`) all tests pass, including both base-ping cases.

stonezdj:
It seems that this fix breaks the dockerhub adapter's feature.
after applyed this fix, the dockerhub registry seems unhealthy.
Screenshot 2026-07-22 at 14 56 06
and there is an error in harbor core log

```
Jul 22 06:06:24 zdj-dev core[3334048]: 2026-07-22T06:06:24Z [ERROR] [/pkg/reg/adapter/native/adapter.go:127]: failed to ping registry https://registry-1.docker.io: http status code: 400, body: {"details":"unknown resource type"}
```

stonezdj (changes requested):
it breaks the dockerhub adapter.

Stevenjin8:
Folks, you probably don't want to merge this. https://istio.io/latest/blog/2026/retirement-of-gcr.io-follow-up/ Let me know if you have any questions

PR #23500, opened 2026-07-06
Add a size limit for manifest uploads

Aloui-Ikram:

Image manifests are small in practice (well under 1 MiB). This adds a
reasonable size limit for manifest uploads.

- Add `MaxManifestBodySize` (4 MiB) in `src/common/const.go`.
- `NopCloseRequest` / `copyBody` (`src/lib/request.go`) gains an optional limit
  and caps retained bytes via `io.LimitReader`; passing `0` keeps the existing
  behavior for callers that need the full body.
- The manifest quota and blob middlewares apply the limit; the manifest handler
  reads through an `io.LimitReader` and returns `413 Request Entity Too Large`
  for oversized bodies.
- The audit-log middleware is unchanged (it handles non-manifest API bodies).
- Adds a unit test.

Legitimate manifests are unaffected.

# Issue being fixed

N/A

- [x] Label the PR as needed: `release-note/enhancement`
- [ ] Considered the docs impact — no docs change needed.

### DIFF
--- a/src/lib/request.go
+++ b/src/lib/request.go
@@ -78,11 +78,16 @@ func ReadRequestBody(r *http.Request, limit int64) ([]byte, error) {
 		return nil, nil
 	}
 
-	var reader io.Reader = r.Body
+	// keep a reference to the original body so we can close it safely once we
+	// have buffered its contents (or bailed out on an over-limit body)
+	originalBody := r.Body
+	defer originalBody.Close()
+
+	var reader io.Reader = originalBody
 	if limit > 0 {
 		// read one extra byte so an over-limit body is detected rather than
 		// silently truncated
-		reader = io.LimitReader(r.Body, limit+1)
+		reader = io.LimitReader(originalBody, limit+1)
 	}
 
 	data, err := io.ReadAll(reader)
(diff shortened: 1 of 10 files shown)


Vad1mo (approved):

src/lib/request.go
```diff
@@ -56,11 +58,41 @@ func copyBody(body io.ReadCloser) io.ReadCloser {
...
+// RequestEntityTooLargeError is returned so callers surface a 413 instead of
+// buffering or parsing a truncated body. On success r.Body is replaced with a
+// re-readable buffer holding the same (<= limit) bytes for any downstream
+// consumer.
+func ReadRequestBody(r *http.Request, limit int64) ([]byte, error) {
+	if r == nil || r.Body == nil {
+		return nil, nil
+	}
+
+	var reader io.Reader = r.Body
```
wy65701436: Keep a reference to the original body so we can close it safely

`originalBody := r.Body`

Aloui-Ikram: thanks! Done , I now keep a reference to the original body (originalBody:= r.Body) and close it with a defer, so it gets closed on every path: normal success,the 413 when the body is too big, and read errors too.

wy65701436 (approved):
lgtm

PR #23504, opened 2026-07-07
feat(replication): add option to skip repo pre-creation for ECR

### LINKED ISSUE #22842: Allow replication without auto-creating repositories on target registry (ECR templates support)
When replicating images from Harbor to AWS ECR, Harbor checks if a repository exists and automatically creates it if it does not.
This breaks AWS ECR repository creation templates, because ECR templates are only applied when the repository is created automatically by ECR itself (create-on-push).
As a result, replicated repositories are created with default settings instead of the expected configuration (immutability, lifecycle policies, encryption, tags, etc.).

Add an optional replication setting like “Create repository if not exists” (enabled by default for backward compatibility).
When disabled, Harbor should not call [CreateRepository](https://github.com/goharbor/harbor/blob/536074d658b1e3c846697092a7ccbb0a460664b5/src/pkg/reg/adapter/awsecr/adapter.go#L195) on the target registry. Harbor should attempt to push images directly and let the target registry auto-create the repository. If creation templates are not enabled the replication will fail with "repository not found".

* create_repository_if_not_exists: true|false (default: true)

Behavior:
* If true: current behavior (Harbor checks and creates repo if missing)
* If false:  Harbor skips repository creation and attempts push directly
* If push fails with "repository not found", return error and rely on remote-side auto-creation features

UI:
* Checkbox in replication rule:
[x] Create repository if not exists on target registry

* Add config flag to replication policy model and API
* Update replication executor logic to skip CreateRepository when disabled
* UI checkbox in replication rule settings
* Default to current behavior to avoid breaking existing setups
* Basic tests with ECR

Adityakumar37:

When replicating images to AWS ECR, Harbor always pre-creates the destination repository via an explicit `CreateRepository` call before pushing. This breaks AWS ECR repository creation templates, which apply settings like immutability, lifecycle policies, and encryption — but only when ECR creates the repository itself (on push, pull-through-cache, or replication triggers), not when it's created through an explicit API call from an external tool like Harbor.

This PR adds an optional `create_repo_if_not_exist` flag on the replication policy (default `true`, so existing policies keep behaving exactly as before). When set to `false`, Harbor skips its own pre-creation step in the ECR adapter and pushes directly, letting ECR's native create-on-push path fire and apply the matching creation template.

The flag is threaded from the replication policy → onto each `Resource` object (mirroring the existing `Override` field, which already does the same kind of policy-to-resource threading) → down into the adapter, so no adapter interface signatures had to change. Only the AWS ECR adapter currently advertises support via a new `SupportedCreateRepoConfig` capability flag, so the new checkbox only shows up in the UI when the destination registry is ECR — other adapters are unaffected.

**Note for reviewers/users:** if a user disables this without a matching ECR creation template configured on the AWS side (matched by repository name prefix), the push will fail with "repository not found" rather than falling back to default settings — this is expected AWS ECR behavior, not a bug, and it's called out in the UI tooltip.

**Changes:**
- `create_repo_if_not_exist` field added to replication policy (DB, DAO, controller model, API)
- ECR adapter (`PrepareForPush`) skips repo creation when the flag is `false`
- New `supported_create_repo_config` capability flag so the UI conditionally shows the option
- UI checkbox in the replication rule form, shown only for ECR destinations
- DB migration, swagger spec update, unit tests (flag defaulting, and DAO round-trip via `To()`/`From()`)

## Issue being fixed
Fixes #22842

## Please indicate you've done the following:

### DIFF
--- a/api/v2.0/swagger.yaml
+++ b/api/v2.0/swagger.yaml
@@ -7680,13 +7680,18 @@ definitions:
           Whether to skip execution until the previous active execution finishes,
           avoiding the execution of the same replication rules multiple times in parallel.
         x-isnullable: true # make this field optional to keep backward compatibility
-      create_repo_if_not_exist:
-        type: boolean
+      adapter_options:
+        type: object
+        additionalProperties:
+          type: string
         description: |-
-          Whether Harbor should pre-create the repository on the destination registry
-          before pushing. Defaults to true (current behavior). Set to false to let the
-          destination registry auto-create the repository on push instead, so
-          registry-side automation (e.g. AWS ECR repository creation templates) can apply.
+          Generic key/value options interpreted differently by each registry
+          adapter; unknown keys are ignored by adapters that don't understand
+          them. For example, the AWS ECR adapter reads "skip_repo_creation":
+          "true" to skip pre-creating the destination repository before
+          pushing, letting the destination registry auto-create it on push
+          instead so registry-side automation (e.g. AWS ECR repository
+          creation templates) can apply.
         x-isnullable: true # make this field optional to keep backward compatibility
   ReplicationTrigger:
     type: object
--- a/api/v2.0/swagger.yaml
+++ b/api/v2.0/swagger.yaml
@@ -7680,6 +7680,14 @@ definitions:
           Whether to skip execution until the previous active execution finishes,
           avoiding the execution of the same replication rules multiple times in parallel.
         x-isnullable: true # make this field optional to keep backward compatibility
+      create_repo_if_not_exist:
+        type: boolean
+        description: |-
+          Whether Harbor should pre-create the repository on the destination registry
+          before pushing. Defaults to true (current behavior). Set to false to let the
+          destination registry auto-create the repository on push instead, so
+          registry-side automation (e.g. AWS ECR repository creation templates) can apply.
+        x-isnullable: true # make this field optional to keep backward compatibility
   ReplicationTrigger:
     type: object
     properties:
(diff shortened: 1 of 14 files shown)


Vad1mo (changes requested):
This will not go through; this is too big of a change all over the place. 

Just looking at the code, I feel it might not be backward compatible. 
Also, the code is pushing one concept for ECR down into all other adapters.


What other solutions exist?

some ideas
- two ECR adapters (a new one wiht skip repo pre-creation)
- ENV Vars
- add skip repo pre-creation on adapter level
- generic key/value options on replication or adapter level to add additional parameters, that each adapter can interpret differently.

Adityakumar37:
That's fair, and looking at it again you're right — baking `create_repo_if_not_exist`
into `Resource` means every adapter now carries an ECR-only concept, even ones that
will never care about it. Sorry for the noise.

I think option 4 (generic key/value options that each adapter interprets on its own)
is the best fit here — it solves this specific ask without adding another one-off
field to the shared model every time some adapter needs a small tunable, and it keeps
the blast radius to just the ECR adapter reading its own key.

Plan: replace `create_repo_if_not_exist` with a generic `adapter_options
map[string]string` on the replication policy, threaded down to `Resource` the same
way, with the ECR adapter reading a `skip_repo_creation` key out of it. Everything
else (migration, swagger, UI checkbox scoped to ECR only) stays the same shape, just
reading/writing through the generic map instead of a dedicated field.

Happy to rework it that way unless you'd rather go with one of the other options —
let me know if you have a preference before I redo the diff.
@Vad1mo

Adityakumar37:
Pushed the rework. Replaced create_repo_if_not_exist with a generic adapter_options
key/value map (map[string]string) on the replication policy — same threading path
as before (policy → resource → adapter), but now the Resource/Policy structs only
carry a generic bag instead of an ECR-specific field. The AWS ECR adapter reads
skip_repo_creation=true from it to skip pre-creating the repo before push; every
other adapter just ignores the key since it means nothing to them.

DB column, swagger spec, and UI checkbox are updated to match (the UI still shows
a single "create repo if not exists" checkbox for ECR destinations, it just now
writes into adapter_options under the hood instead of a dedicated field).

PR #23520, opened 2026-07-08
fix: use LEFT JOIN in projectBasicSQL to include projects without quota records

### LINKED ISSUE #23519: exporter: projectBasicSQL INNER JOIN on quota tables drops projects when quota_per_project_enable=false
## Description

When `quota_per_project_enable` is set to `false` in Harbor's configuration, no rows are created in the `quota` or `quota_usage` tables for new projects. The Prometheus exporter's `projectBasicSQL` uses `INNER JOIN` on both tables, silently excluding every project without quota records from the in-memory project map built on each scrape.

The three subsequent collector functions (`updateProjectMemberInfo`, `updateProjectRepoInfo`, `updateProjectArtifactInfo`) query those same projects independently and call `log.Errorf("project not found, ID %d")` for each project ID absent from the map. With a large number of affected projects this produces hundreds of error log writes per scrape, increasing scrape latency and risking Prometheus scrape timeouts that drop all Harbor metrics. Affected projects also have no metrics emitted for quota, member, repo, or artifact dimensions.

## Steps to Reproduce

1. Set `quota_per_project_enable=false` in Harbor configuration.
2. Create one or more projects — no rows are inserted into `quota` or `quota_usage`.
3. Observe the exporter logs: a `project not found` error is logged for each affected project on every Prometheus scrape.
4. With enough affected projects, Prometheus scrapes begin to time out intermittently.

## Expected Behavior

All non-deleted projects appear in the collector map regardless of whether quota records exist. Projects without quota records should emit `0` for storage quota/usage metrics.

## Actual Behavior

Projects without quota records are silently dropped from the map. The exporter floods logs with `project not found` errors on every scrape and Prometheus metrics are incomplete or missing.

## Environment

- Harbor version: v2.13.1
- `quota_per_project_enable`: `false`

## Fix

PR #23520 switches `projectBasicSQL` from `INNER JOIN` to `LEFT JOIN` on `quota`, `quota_usage`, and `project_metadata`, with `COALESCE` defaults so projects without quota records emit `{}` (storage = 0) rather than being dropped.

cdarninsuang-bamfunds:
Fixes #23519

## What does this PR fix?

When `quota_per_project_enable` is `false`, Harbor does not create rows in the `quota` or `quota_usage` tables for new projects. The exporter's `projectBasicSQL` used `INNER JOIN` on both tables, so any project missing those rows was silently excluded from the in-memory project map built on each scrape.

The three subsequent collector functions (`updateProjectMemberInfo`, `updateProjectRepoInfo`, `updateProjectArtifactInfo`) query those same projects via independent JOINs on the `project` table directly, and call `log.Errorf` for each project ID not found in the map. With a large number of projects missing quota records this produces hundreds of error log writes per scrape, increasing scrape latency and risking Prometheus timeouts that drop all Harbor metrics.

## Changes

- **`src/pkg/exporter/project_collector.go`**: Convert `projectBasicSQL` from `INNER JOIN` to `LEFT JOIN` on `quota`, `quota_usage`, and `project_metadata`. Join-condition filtering (e.g. `quota.reference='project'`, `project_metadata.name='public'`) is moved into the `ON` clause as required for correct `LEFT JOIN` semantics. `COALESCE` defaults ensure projects without quota records emit `{}` (storage = 0) rather than being dropped. The quota join now casts `project_id` to text rather than casting `reference_id` to integer, which is safer if the quota table ever contains non-integer reference IDs.

- **`src/pkg/exporter/project_collector_test.go`**: Add `testPro3`, a project created without quota records, and assert it appears in the collector map with empty quota/usage values — covering the previously broken path.

## Testing

Existing unit tests continue to pass for projects with quota records. The new `testPro3` case verifies the fix.

Tested against a live Harbor instance where `quota_per_project_enable=false` and 529 of 611 projects had no quota rows — after the fix, all 611 projects appear in the map and the `project not found` error flood is eliminated.

## Notes

- No database schema changes.
- No behavioural change for projects that have quota records.

Signed-off-by: cdarninsuang <cdarninsuang@bamfunds.com>

### DIFF
--- a/src/pkg/exporter/project_collector.go
+++ b/src/pkg/exporter/project_collector.go
@@ -32,11 +32,11 @@ var (
 	FROM project INNER JOIN project_metadata ON project.project_id=project_metadata.project_id
 	WHERE project.deleted=FALSE AND project_metadata.name='public'
 	GROUP BY project_metadata.value;`
-	projectBasicSQL = `SELECT project.project_id, project.name, project_metadata.value AS public, quota.hard AS quota, quota_usage.used AS usage FROM project
-	INNER JOIN project_metadata ON (project.project_id = project_metadata.project_id)
-	INNER JOIN quota ON project.project_id = CAST(quota.reference_id AS Integer)
-	INNER JOIN quota_usage ON project.project_id = CAST(quota_usage.reference_id AS Integer)
-	WHERE quota.reference='project' AND quota_usage.reference='project' AND project.deleted=FALSE AND project_metadata.name='public';`
+	projectBasicSQL = `SELECT project.project_id, project.name, COALESCE(project_metadata.value, 'false') AS public, COALESCE(quota.hard, '{"storage": -1}') AS quota, COALESCE(quota_usage.used, '{}') AS usage FROM project
+	LEFT JOIN project_metadata ON (project.project_id = project_metadata.project_id AND project_metadata.name='public')
+	LEFT JOIN quota ON (CAST(project.project_id AS text) = quota.reference_id AND quota.reference='project')
+	LEFT JOIN quota_usage ON (CAST(project.project_id AS text) = quota_usage.reference_id AND quota_usage.reference='project')
+	WHERE project.deleted=FALSE;`
 	projectMemberSQL = `SELECT project.project_id, COUNT(project.project_id) AS member_total
 	FROM project INNER JOIN project_member ON project.project_id=project_member.project_id
 	WHERE project.deleted=FALSE AND project_member.entity_type='u'
--- a/src/pkg/exporter/project_collector_test.go
+++ b/src/pkg/exporter/project_collector_test.go
@@ -29,6 +29,7 @@ var (
 	eve      = models.User{Username: "eve", Password: "password", Email: "eve@test.com"}
 	testPro1 = proModels.Project{OwnerID: 1, Name: "test1", Metadata: map[string]string{"public": "true"}}
 	testPro2 = proModels.Project{OwnerID: 1, Name: "test2", Metadata: map[string]string{"public": "false"}}
+	testPro3 = proModels.Project{OwnerID: 1, Name: "test3", Metadata: map[string]string{"public": "false"}}
 	rs1      = qtypes.ResourceList{qtypes.ResourceStorage: 100}
 	rs2      = qtypes.ResourceList{qtypes.ResourceStorage: 200}
 	repo1    = model.RepoRecord{Name: "repo1"}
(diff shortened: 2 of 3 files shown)


src/pkg/exporter/project_collector.go
```diff
@@ -32,11 +32,11 @@ var (
 	FROM project INNER JOIN project_metadata ON project.project_id=project_metadata.project_id
 	WHERE project.deleted=FALSE AND project_metadata.name='public'
 	GROUP BY project_metadata.value;`
-	projectBasicSQL = `SELECT project.project_id, project.name, project_metadata.value AS public, quota.hard AS quota, quota_usage.used AS usage FROM project
-	INNER JOIN project_metadata ON (project.project_id = project_metadata.project_id)
-	INNER JOIN quota ON project.project_id = CAST(quota.reference_id AS Integer)
-	INNER JOIN quota_usage ON project.project_id = CAST(quota_usage.reference_id AS Integer)
-	WHERE quota.reference='project' AND quota_usage.reference='project' AND project.deleted=FALSE AND project_metadata.name='public';`
+	projectBasicSQL = `SELECT project.project_id, project.name, COALESCE(project_metadata.value, 'false') AS public, COALESCE(quota.hard, '{}') AS quota, COALESCE(quota_usage.used, '{}') AS usage FROM project
```
chlins: COALESCE(quota.hard, '{}') emits quota_byte=0, but Harbor's convention for unlimited is -1 (quota enabled with no limit emits -1). This breaks usage/quota dashboards (division by zero) and fires "quota exhausted" alerts. Suggest COALESCE(quota.hard, '{"storage": -1}'). The '{}' default for usage is fine.

cdarninsuang-bamfunds: Makes sense, thanks. Switched the default to `'{"storage": -1}'` and left usage at `{}`. Test updated too.

chlins:
One minor style nit: this introduces the first :: cast in the Go codebase. Everywhere else Harbor uses standard CAST(...) syntax (e.g. the original lines here, and src/pkg/quota/dao/util.go). For consistency, suggest:

```sql
LEFT JOIN quota ON (CAST(project.project_id AS text) = quota.reference_id AND quota.reference='project')
LEFT JOIN quota_usage ON (CAST(project.project_id AS text) = quota_usage.reference_id AND quota_usage.reference='project')
```

Functionally identical, just keeps the SQL style uniform.

cdarninsuang-bamfunds:
Switched both joins to `CAST(... AS text)`. Pushed.

chlins (approved):
lgtm

cdarninsuang-bamfunds:
Hey @chlins would you be able to re-approve a test workflow invocation please? Had to fix the project counts in the unit tests. Thank you.

wy65701436 (approved):
lgtm

PR #23522, opened 2026-07-08
fix: avoid session fallback when Authorization header is present

### LINKED ISSUE #7704: Swagger API explorer should use the authorized user

Pragati5-DEBUG:
## Summary
- Fixes fallback behavior in security middleware when an `Authorization` header is present.
- Prevents invalid Basic credentials (e.g. from API Explorer/Swagger) from silently falling back to the logged-in session user.
- Adds middleware tests for both cases:
  - `Authorization` present + invalid credentials => no session fallback
  - no `Authorization` header => existing session auth behavior remains

## Why
Issue #7704 reports that API Explorer can appear “authorized” with fake credentials while requests are still executed as the current portal session user.  
This change makes auth behavior explicit: when `Authorization` is provided, failed auth does not downgrade to session auth.

## Test Plan
- Added unit tests in `src/server/middleware/security/security_test.go`:
  - `TestSecuritySkipsSessionWhenAuthorizationPresent`
  - `TestSecurityUsesSessionWhenNoAuthorization`
- Local full test run is limited on native Windows due to Harbor Linux-specific dependencies (`syslog`); CI should validate on Linux.

## Related
- Fixes #7704

### DIFF
--- a/src/server/middleware/security/security.go
+++ b/src/server/middleware/security/security.go
@@ -53,7 +53,16 @@ func Middleware(skippers ...middleware.Skipper) func(http.Handler) http.Handler 
 		} else {
 			log.Warningf("failed to get auth mode: %v", err)
 		}
+		// When Authorization is provided (e.g. Swagger UI "Authorize"), do not
+		// fall back to the portal session. Failed credentials must not silently
+		// use the logged-in user (https://github.com/goharbor/harbor/issues/7704).
+		hasAuthorization := r.Header.Get("Authorization") != ""
 		for _, generator := range generators {
+			if hasAuthorization {
+				if _, ok := generator.(*session); ok {
+					continue
+				}
+			}
 			if ctx := generator.Generate(r); ctx != nil {
 				r = r.WithContext(security.NewContext(r.Context(), ctx))
 				break
--- a/src/server/middleware/security/security.go
+++ b/src/server/middleware/security/security.go
@@ -16,6 +16,7 @@ package security
 
 import (
 	"net/http"
+	"strings"
 
 	"github.com/goharbor/harbor/src/common/security"
 	"github.com/goharbor/harbor/src/lib"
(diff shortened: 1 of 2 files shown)


Pragati5-DEBUG:
Hi, could a maintainer please add the `release-note/enhancement` label when you have a moment?

This PR fixes #7704 — prevents invalid Basic auth in the API Explorer from silently falling back to the portal session user. Happy to address any feedback. Thanks!

Nachiket-Roy (changes requested):
Hello @Pragati5-DEBUG,

Overall, the approach makes sense. I have a couple of observations that could further improve the implementation:

1. **Refine the `Authorization` header presence check**

   Currently, `r.Header.Get("Authorization") != ""` skips session authentication for any non-empty `Authorization` header. If a reverse proxy, load balancer, or browser extension injects a non-standard `Authorization` header, users could unexpectedly bypass session authentication even though Harbor doesn't handle that scheme.

   It would be safer to only treat the header as authoritative for the schemes Harbor supports (e.g. `Basic` and `Bearer`):

   ```go
   authHeader := r.Header.Get("Authorization")
   hasAuthorization := strings.HasPrefix(strings.ToLower(authHeader), "basic ") ||
       strings.HasPrefix(strings.ToLower(authHeader), "bearer ")
   ```

2. **Add a test for header precedence**

   The new tests cover the fallback-avoidance cases well, but I didn't see a test verifying the expected precedence when both header- and session-based authentication are valid.

   Could you add a test (e.g. `TestSecurityAuthorizationTakesPrecedenceOverSession` in `security_test.go`) to explicitly verify that when both are present, the authenticated identity is resolved from the `Authorization` header (User A) rather than the session (User B)? This would also help document the intended behavior.

Pragati5-DEBUG:
Thanks @Nachiket-Roy for taking a look!

Updated based on your suggestions:
1. Only skip session for `Basic` / `Bearer`
2. Added a precedence test when both header and session auth are present

Appreciate the review.

Nachiket-Roy (approved):
LGTM. Please make sure all commits are signed off

PR #23527, opened 2026-07-09
feat(backend): implement proxy cache repository filter API

### LINKED ISSUE #13231: Limit Proxy Cache Images
With the new version of Harbor it can now be used as a proxy cache. As administrator I would like to have the possibility to restrict the images that are pulled (whitelist). This would save us a lot of replication rules.

stonezdj:
Introduce repository filter pattern and kind on proxy cache projects to explicitly allow selected repositories. Requests that do not match the filter are denied before Harbor proxies content from upstream.

Fixes #13231

# Issue being fixed
Fixes #(issue)

### DIFF
--- a/src/lib/pattern/matcher.go
+++ b/src/lib/pattern/matcher.go
@@ -1,0 +1,118 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pattern
+
+import (
+	"regexp"
+	"strings"
+
+	"github.com/bmatcuk/doublestar"
+
+	"github.com/goharbor/harbor/src/lib/errors"
+)
+
+const (
+	// KindRegex indicates regular expression pattern matching
+	KindRegex = "regex"
+	// KindDoublestar indicates doublestar (glob) pattern matching
+	KindDoublestar = "doublestar"
+)
+
+// ValidateRepositoryFilter validates the repository filter kind and pattern.
+// Empty pattern is valid and means all repositories are allowed.
+func ValidateRepositoryFilter(filterPattern, kind string) error {
+	_, err := Match("", filterPattern, kind)
+	return err
+}
+
+// Match returns true if the value matches the pattern according to the kind.
+// Empty pattern matches all (returns true).
+func Match(value, filterPattern, kind string) (bool, error) {
+	filterPattern = strings.TrimSpace(filterPattern)
+	if filterPattern == "" {
+		return true, nil
+	}
+	switch kind {
+	case KindRegex:
+		re, err := regexp.Compile("^(?:" + filterPattern + ")$")
+		if err != nil {
+			return false, err
+		}
+		return re.MatchString(value), nil
+	case KindDoublestar, "":
+		if !validateDoublestarPattern(filterPattern) {
+			return false, doublestar.ErrBadPattern
+		}
+		return doublestar.Match(filterPattern, value)
+	default:
+		return false, errors.Errorf("unsupported repository filter kind %q", kind)
+	}
+}
+
+func validateDoublestarPattern(s string) bool {
+	altDepth := 0
+	l := len(s)
+VALIDATE:
+	for i := 0; i < l; i++ {
+		switch s[i] {
+		case '\\':
+			// skip the next byte - return false if there is no next byte
+			if i++; i >= l {
+				return false
+			}
+			continue
+
+		case '[':
+			if i++; i >= l {
+				// class didn't end
+				return false
+			}
+			if s[i] == '^' || s[i] == '!' {
+				i++
+			}
+			if i >= l || s[i] == ']' {
+				// class didn't end or empty character class
+				return false
+			}
+
+			for ; i < l; i++ {
+				if s[i] == '\\' {
+					i++
+				} else if s[i] == ']' {
+					// looks good
+					continue VALIDATE
+				}
+			}
+
+			// class didn't end
+			return false
+
+		case '{':
+			altDepth++
+			continue
+
+		case '}':
+			if altDepth == 0 {
+				// alt end without a corresponding start
+				return false
+			}
+			altDepth--
+			continue
+		}
+	}
+
+	// valid as long as all alts are closed
+	return altDepth == 0
+}
--- a/src/pkg/auditext/event/project/project.go
+++ b/src/pkg/auditext/event/project/project.go
@@ -1,0 +1,118 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package project
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"regexp"
+	"strconv"
+	"time"
+
+	ctlevent "github.com/goharbor/harbor/src/controller/event"
+	"github.com/goharbor/harbor/src/controller/event/metadata/commonevent"
+	"github.com/goharbor/harbor/src/controller/event/model"
+	"github.com/goharbor/harbor/src/controller/project"
+	"github.com/goharbor/harbor/src/lib/config"
+	"github.com/goharbor/harbor/src/lib/orm"
+	"github.com/goharbor/harbor/src/pkg/notifier/event"
+	"github.com/goharbor/harbor/src/pkg/project/models"
+)
+
+var (
+	projectReg        = regexp.MustCompile(`^/api/v2.0/projects/([^/?]+)(?:\?.*)?$`)
+	projectMetaReg    = regexp.MustCompile(`^/api/v2.0/projects/([^/?]+)/metadatas/(?:\?.*)?$`)
+	projectMetaKeyReg = regexp.MustCompile(`^/api/v2.0/projects/([^/?]+)/metadatas/([^/?]+)(?:\?.*)?$`)
+)
+
+func init() {
+	var projectEventResolver = &resolver{}
+	commonevent.RegisterResolver(`^/api/v2.0/projects/[^/?]+(?:\?.*)?$`, projectEventResolver)
+	commonevent.RegisterResolver(`^/api/v2.0/projects/[^/?]+/metadatas/(?:\?.*)?$`, projectEventResolver)
+	commonevent.RegisterResolver(`^/api/v2.0/projects/[^/?]+/metadatas/[^/?]+(?:\?.*)?$`, projectEventResolver)
+}
+
+type resolver struct {
+}
+
+func (r *resolver) Resolve(ce *commonevent.Metadata, evt *event.Event) error {
+	projStr := getProjectNameOrID(ce.RequestURL)
+	if projStr == "" {
+		return fmt.Errorf("failed to parse project name or ID from URL: %s", ce.RequestURL)
+	}
+
+	ctx := ce.Ctx
+	if _, err := orm.FromContext(ctx); err == nil {
+		ctx = orm.Clone(ctx)
+	}
+
+	proj, err := getProject(ctx, projStr)
+	if err != nil {
+		return fmt.Errorf("failed to get project: %v", err)
+	}
+
+	e := &model.CommonEvent{}
+	e.Operation = "update"
+	e.Operator = ce.Username
+	e.ResourceType = "project"
+	e.ResourceName = proj.Name
+	e.ProjectID = proj.ProjectID
+	e.OcurrAt = time.Now()
+	e.OperationDescription = fmt.Sprintf("update project: %s", proj.Name)
+
+	if ce.ResponseCode == http.StatusOK {
+		e.IsSuccessful = true
+	}
+	evt.Topic = ctlevent.TopicCommonEvent
+	evt.Data = e
+	return nil
+}
+
+func (r *resolver) PreCheck(ctx context.Context, url string, method string) (bool, string) {
+	if method != http.MethodPut && method != http.MethodPost && method != http.MethodDelete {
+		return false, ""
+	}
+
+	if method == http.MethodDelete && projectReg.MatchString(url) {
+		return false, ""
+	}
+
+	return config.AuditLogEventEnabled(ctx, "update_project"), ""
+}
+
+func getProjectNameOrID(url string) string {
+	if m := projectMetaKeyReg.FindStringSubmatch(url); len(m) > 1 {
+		return m[1]
+	}
+	if m := projectMetaReg.FindStringSubmatch(url); len(m) > 1 {
+		return m[1]
+	}
+	if m := projectReg.FindStringSubmatch(url); len(m) > 1 {
+		return m[1]
+	}
+	return ""
+}
+
+func getProject(ctx context.Context, str string) (*models.Project, error) {
+	var projectNameOrID any
+	v, err := strconv.ParseInt(str, 10, 64)
+	if err != nil {
+		projectNameOrID = str
+	} else {
+		projectNameOrID = v
+	}
+	return project.Ctl.Get(ctx, projectNameOrID)
+}
--- a/src/server/middleware/repoproxy/proxy.go
+++ b/src/server/middleware/repoproxy/proxy.go
@@ -44,6 +44,7 @@ import (
 	"github.com/goharbor/harbor/src/lib/log"
 	"github.com/goharbor/harbor/src/lib/metric"
 	"github.com/goharbor/harbor/src/lib/orm"
+	"github.com/goharbor/harbor/src/lib/pattern"
 	"github.com/goharbor/harbor/src/lib/q"
 	"github.com/goharbor/harbor/src/lib/redis"
 	"github.com/goharbor/harbor/src/pkg"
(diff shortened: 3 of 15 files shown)


Vad1mo:
Comparing this against the merged proposal (goharbor/community#280), a few deviations:

**1. Pattern validation at the metadata endpoint ignores the configured kind**

The proposal requires: *"Invalid pattern in `proxy_cache_filter_pattern` is rejected with `400 Bad Request` when creating/updating project metadata."*

In `project_metadata.go`, `validate()` checks each key independently, so the pattern is always validated as doublestar:

```go
case proModels.ProMetaProxyCacheFilterPattern:
    if err := pattern.ValidateRepositoryFilter(value, pattern.KindDoublestar); err != nil {
```

With `proxy_cache_filter_kind=regex`, an invalid regex that is a valid glob (e.g. `library/(*`) passes validation here, and since invalid patterns are treated as non-match at runtime, every pull in the project silently 404s. A valid regex that is an invalid glob is wrongly rejected. `validateProxyCacheRepositoryFilter` in `project.go` gets this right — the metadatas endpoint needs the same cross-field logic.

**2. Metadata endpoint accepts the keys on non-proxy projects**

Proposal: *"For non-proxy projects, these keys are ignored on create/update and omitted from response."* Create/Update strip them and `ToSwagger` hides them, but `PUT/POST /projects/{id}/metadatas` stores both keys on any project.

**3. Regex "auto-anchoring" isn't equivalent to real anchoring**

Proposal: regex is *"always auto-anchored."* `RegexMatcher.Match` checks that the leftmost match (`FindStringIndex`) spans the whole string. Go regexp alternation is leftmost-first, so `a|ab` fails to match `"ab"` even though `^(?:a|ab)$` matches. Compiling `^(?:` + pattern + `)$` implements the stated semantics correctly. Also worth documenting explicitly that patterns must match the full repository path (`^library/` matches nothing) — the companion UI PR #23536 currently describes this as "head-anchored", which doesn't match this implementation.

**4. No audit entry when the filter changes**

Proposal, Audit section: *"Record project update audit entries when filter set is changed."* Only the blocked-pull log is implemented.

**5. Blob path is not filtered (matches the proposal's implementation section, but not its Abstract)**

Enforcement is in `handleManifest` only, exactly as the proposal's implementation section specifies. However the Abstract says requests are *"denied before Harbor proxies content from upstream"* — a client that knows a blob digest out-of-band can still pull blobs of a filtered-out repository through the blob proxy path. If the filter is meant as a security boundary rather than a convenience, blob requests need the same check; otherwise this limitation should be documented.

Vad1mo:
backward-compatibility analysis

- The filter applies retroactively to already-cached content: adding a filter to an existing proxy project blocks pulls of previously cached non-matching repos (the proposal explicitly applies the pattern to local content too). Clusters pulling those images break at the moment the filter is saved, not at the next upstream fetch.
- Downgrade/rollback: the metadata rows persist, older Harbor silently ignores them, and filter enforcement vanishes without any warning — a silent security regression if the filter was doing compliance work.

src/lib/pattern/matcher.go
```diff
@@ -0,0 +1,117 @@
...
+func (r *RegexMatcher) Match(value, pattern string) (bool, error) {
+	pattern = strings.TrimSpace(pattern)
+	if pattern == "" {
+		return true, nil
+	}
+	re, err := regexp.Compile(pattern)
+	if err != nil {
+		return false, err
+	}
+	match := re.FindStringIndex(value)
```
wy65701436: `FindStringIndex` returns the leftmost match, not necessarily a full-string match, so this can return false negatives. 

Example: pattern `a|ab` against value `ab` — the leftmost match is `a` (`[0,1]`), so this returns `false` even though the pattern matches the whole string.

src/server/v2.0/handler/project_metadata.go
```diff
@@ -168,6 +169,20 @@ func (p *projectMetadataAPI) validate(metas map[string]string) (map[string]strin
 			return nil, errors.New(nil).WithCode(errors.BadRequestCode).WithMessagef("invalid value: %s", value)
 		}
 		metas[proModels.ProMetaMaxUpstreamConn] = strconv.FormatInt(v, 10)
+	case proModels.ProMetaProxyCacheFilterPattern:
+		// plain pattern string; empty means match all — no further validation needed at metadata level
+		if err := pattern.ValidateRepositoryFilter(value, pattern.KindDoublestar); err != nil {
```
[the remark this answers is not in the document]
wy65701436: Is hard code the KindDoublestar intended?

src/server/middleware/repoproxy/proxy.go
```diff
@@ -193,6 +194,22 @@ func defaultBlobURL(projectName string, name string, digest string) string {
 	return fmt.Sprintf("/v2/%s/library/%s/blobs/%s", projectName, name, digest)
 }
 
+// matchRepositoryFilter returns true if repository matches the filter.
+// filterPattern is the plain pattern string; filterKind is "doublestar" or "regex" (defaults to doublestar when empty).
+// If filterPattern is empty, all repositories are allowed.
+// An invalid pattern is treated as no match.
+func matchRepositoryFilter(repository, filterPattern, filterKind string) bool {
```
wy65701436, line 259: the matchRepositoryFilter is applied only for manifest, should we apply it for blob?

src/lib/pattern/matcher.go
```diff
@@ -0,0 +1,117 @@
...
+func NewMatcher(kind string) Matcher {
+	switch kind {
+	case KindRegex:
+		return &RegexMatcher{}
+	case KindDoublestar:
+		fallthrough
+	default:
+		return &DoublestarMatcher{}
+	}
+}
```
wy65701436, line 62:
```
func ValidateRepositoryFilter(filterPattern, kind string) error {
    filterPattern = strings.TrimSpace(filterPattern)
    if filterPattern == "" {
        return nil
    }
    switch kind {
    case KindRegex:
        _, err := regexp.Compile(filterPattern)
        return err
    case KindDoublestar, "":
        if !doublestar.ValidatePattern(filterPattern) {
            return doublestar.ErrBadPattern
        }
        return nil
    default:
        return errors.Errorf("unsupported repository filter kind %q", kind)
    }
}


```

stonezdj:
> backward-compatibility analysis
> 
> * The filter applies retroactively to already-cached content: adding a filter to an existing proxy project blocks pulls of previously cached non-matching repos (the proposal explicitly applies the pattern to local content too). Clusters pulling those images break at the moment the filter is saved, not at the next upstream fetch.
> * Downgrade/rollback: the metadata rows persist, older Harbor silently ignores them, and filter enforcement vanishes without any warning — a silent security regression if the filter was doing compliance work.

The proxy cache filter policy's behvior should be consitent with the already cached image and not-cached image.  if it use different policy for cached and non-cached iamge, it will cause security issue.

src/server/v2.0/handler/project_metadata.go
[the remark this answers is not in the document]
stonezdj: Updated, it should be KindDoublestart or KindRegEx

src/lib/pattern/matcher.go
stonezdj: Updated

src/server/middleware/repoproxy/proxy.go
stonezdj: Updated for blob

src/server/middleware/repoproxy/proxy.go
```diff
@@ -219,6 +236,16 @@ func handleManifest(w http.ResponseWriter, r *http.Request, next http.Handler) e
 		return nil
 	}
 
+	// Apply repository filter: if proxy_cache_filter_pattern is set, the repository must match
+	if filterPattern, ok := p.GetMetadata(proModels.ProMetaProxyCacheFilterPattern); ok && filterPattern != "" {
```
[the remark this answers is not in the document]
stonezdj: fixed

src/lib/pattern/matcher.go
stonezdj: Updated

src/server/middleware/repoproxy/proxy.go
stonezdj: Updated

stonezdj:
> Comparing this against the merged proposal ([goharbor/community#280](https://github.com/goharbor/community/pull/280)), a few deviations:
> 
> **1. Pattern validation at the metadata endpoint ignores the configured kind**
> 
> The proposal requires: _"Invalid pattern in `proxy_cache_filter_pattern` is rejected with `400 Bad Request` when creating/updating project metadata."_
> 
> In `project_metadata.go`, `validate()` checks each key independently, so the pattern is always validated as doublestar:
> 
> ```go
> case proModels.ProMetaProxyCacheFilterPattern:
>     if err := pattern.ValidateRepositoryFilter(value, pattern.KindDoublestar); err != nil {
> ```
> 
> With `proxy_cache_filter_kind=regex`, an invalid regex that is a valid glob (e.g. `library/(*`) passes validation here, and since invalid patterns are treated as non-match at runtime, every pull in the project silently 404s. A valid regex that is an invalid glob is wrongly rejected. `validateProxyCacheRepositoryFilter` in `project.go` gets this right — the metadatas endpoint needs the same cross-field logic.
> 
> **2. Metadata endpoint accepts the keys on non-proxy projects**
> 
> Proposal: _"For non-proxy projects, these keys are ignored on create/update and omitted from response."_ Create/Update strip them and `ToSwagger` hides them, but `PUT/POST /projects/{id}/metadatas` stores both keys on any project.
> 
> **3. Regex "auto-anchoring" isn't equivalent to real anchoring**
> 
> Proposal: regex is _"always auto-anchored."_ `RegexMatcher.Match` checks that the leftmost match (`FindStringIndex`) spans the whole string. Go regexp alternation is leftmost-first, so `a|ab` fails to match `"ab"` even though `^(?:a|ab)$` matches. Compiling `^(?:` + pattern + `)$` implements the stated semantics correctly. Also worth documenting explicitly that patterns must match the full repository path (`^library/` matches nothing) — the companion UI PR #23536 currently describes this as "head-anchored", which doesn't match this implementation.
> 
> **4. No audit entry when the filter changes**
> 
> Proposal, Audit section: _"Record project update audit entries when filter set is changed."_ Only the blocked-pull log is implemented.
> 
> **5. Blob path is not filtered (matches the proposal's implementation section, but not its Abstract)**
> 
> Enforcement is in `handleManifest` only, exactly as the proposal's implementation section specifies. However the Abstract says requests are _"denied before Harbor proxies content from upstream"_ — a client that knows a blob digest out-of-band can still pull blobs of a filtered-out repository through the blob proxy path. If the filter is meant as a security boundary rather than a convenience, blob requests need the same check; otherwise this limitation should be documented.

Thanks comments, I've fixed these five issues

src/server/middleware/repoproxy/proxy.go
```diff
@@ -193,6 +199,40 @@ func defaultBlobURL(projectName string, name string, digest string) string {
...
+		return false
+	}
+	log.Debugf("repository %q match result: %v (filter: %q, kind: %s)", repository, matched, rf.Filter, rf.Kind)
+	return matched
+}
+
+// checkRepositoryFilter returns a NotFoundError if the project has a proxy_cache_filter_pattern
+// set and the artifact's repository does not match it. It is enforced for both manifest and
+// blob requests so the filter acts as an access boundary rather than a manifest-only convenience,
+// since a client that knows a blob digest out-of-band could otherwise bypass the filter.
```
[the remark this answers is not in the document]
stonezdj: fixed

src/server/v2.0/handler/project_metadata_test.go
```diff
@@ -14,20 +14,55 @@
 
 package handler
 
+// Import context and other required packages
```
chlins: Please remove this AI noise log.

src/server/v2.0/handler/project_metadata.go
```diff
@@ -132,7 +133,7 @@ func (p *projectMetadataAPI) UpdateProjectMetadata(ctx context.Context, params o
 	return operation.NewUpdateProjectMetadataOK()
 }
 
-func (p *projectMetadataAPI) validate(metas map[string]string) (map[string]string, error) {
+func (p *projectMetadataAPI) validate(ctx context.Context, projectID int64, metas map[string]string) (map[string]string, error) {
```
chlins, line 149: The add/update paths now have solid validation (including the nice
cross-check when changing the kind against an existing pattern), but
`DeleteProjectMetadata` (L83-96, untouched by this PR) doesn't get the
same guard — which leaves a way to silently break an entire proxy cache
project:

1. Configure a valid filter:
   `proxy_cache_filter_pattern = "^library/(nginx|redis)$"`,
   `proxy_cache_filter_kind = "regex"` — validated, works as expected.
2. `DELETE /api/v2.0/projects/{id}/metadatas/proxy_cache_filter_kind`
   — succeeds, no validation on this handler.
3. The middleware now sees a non-empty pattern with no kind, so
   `pattern.Match` falls back to doublestar (the `KindDoublestar, ""`
   case). The regex metacharacters `^ ( | ) $` are all literals in
   doublestar, `validateDoublestarPattern` happily accepts the string,
   and the match is always false.
4. Since non-matching repositories are denied (fail-closed by design),
   every pull in the project now returns 404 — with nothing but a
   `log.Infof` in core to explain why.

The update path already solves exactly this class of problem
("existing proxy_cache_filter_pattern %q is invalid for new kind %q"),
so the delete path just needs the same treatment. Suggestion:

- In `DeleteProjectMetadata`, when `MetaName ==
  proxy_cache_filter_kind` and a non-empty
  `proxy_cache_filter_pattern` still exists, reject with 400 and ask
  the caller to clear the pattern first (or re-validate the remaining
  pattern against the default kind and reject if it fails).

Deleting the pattern itself is harmless (empty pattern = allow all),
so only the kind deletion needs guarding.

Would be great to pin this with a test: delete kind while a regex
pattern is present → expect 400.

chlins (approved):
lgtm

src/server/middleware/repoproxy/proxy.go
```diff
@@ -241,6 +247,39 @@ func defaultBlobURL(projectName string, name string, digest string) string {
...
+func checkRepositoryFilter(p *proModels.Project, art lib.ArtifactInfo) error {
+	filterPattern, ok := p.GetMetadata(proModels.ProMetaProxyCacheFilterPattern)
+	if !ok || filterPattern == "" {
+		return nil
+	}
+	filterKind, _ := p.GetMetadata(proModels.ProMetaProxyCacheFilterKind)
+	remoteRepo := strings.TrimPrefix(art.Repository, art.ProjectName+"/")
+	if !matchRepositoryFilter(remoteRepo, filterPattern, filterKind) {
+		log.Infof("blocked proxy cache pull for project %q repository %q: repository does not match filter %q (kind: %s)", p.Name, remoteRepo, filterPattern, filterKind)
+		return errors.NotFoundError(fmt.Errorf("repository %q does not match project repository filter %q (kind: %s)", remoteRepo, filterPattern, filterKind))
```
wy65701436: please don't return the filter pattern in the 404 body.

 For a public proxy-cache project, an unauthenticated client can currently discover the admin's allowlist by probing denied repos.

src/server/middleware/repoproxy/proxy.go
```diff
@@ -241,6 +247,39 @@ func defaultBlobURL(projectName string, name string, digest string) string {
...
+// since a client that knows a blob digest out-of-band could otherwise bypass the filter.
+func checkRepositoryFilter(p *proModels.Project, art lib.ArtifactInfo) error {
+	filterPattern, ok := p.GetMetadata(proModels.ProMetaProxyCacheFilterPattern)
+	if !ok || filterPattern == "" {
+		return nil
+	}
+	filterKind, _ := p.GetMetadata(proModels.ProMetaProxyCacheFilterKind)
+	remoteRepo := strings.TrimPrefix(art.Repository, art.ProjectName+"/")
+	if !matchRepositoryFilter(remoteRepo, filterPattern, filterKind) {
+		log.Infof("blocked proxy cache pull for project %q repository %q: repository does not match filter %q (kind: %s)", p.Name, remoteRepo, filterPattern, filterKind)
```
wy65701436: Please lower this to `log.Debugf`. Kubernetes nodes retry image pulls again and again, so there will be logs of blocked logs.

src/pkg/auditext/event/project/project.go
```diff
@@ -0,0 +1,113 @@
...
+	if ce.ResponseCode == http.StatusOK {
+		e.IsSuccessful = true
+	}
+	evt.Topic = ctlevent.TopicCommonEvent
+	evt.Data = e
+	return nil
+}
+
+func (r *resolver) PreCheck(ctx context.Context, _ string, method string) (bool, string) {
+	if method != http.MethodPut && method != http.MethodPost && method != http.MethodDelete {
```
wy65701436, line 85: This also matches `DELETE /api/v2.0/projects/{id}` (project deletion). After a successful delete, `Resolve` tries to fetch the project — but it's already gone, so no event is emitted.

Is this expected?

src/server/v2.0/handler/project.go
```diff
@@ -838,6 +850,27 @@ func (a *projectAPI) validateProjectReq(ctx context.Context, req *models.Project
...
 }
 
+func validateProxyCacheRepositoryFilter(metadata *models.ProjectMetadata) error {
+	if metadata == nil {
+		return nil
+	}
+
+	filterKind := lib.StringValue(metadata.ProxyCacheFilterKind)
+	if filterKind == "" {
+		filterKind = pattern.KindDoublestar
```
wy65701436: There should be a problem,

1, The admin configures a regex filter. Both fields together, everything is fine:

```
PUT /api/v2.0/projects/dockerhub-proxy
{
  "metadata": {
    "proxy_cache_filter_pattern": "library/(nginx|redis)",
    "proxy_cache_filter_kind": "regex"
  }
}
```

Stored in DB: `pattern = library/(nginx|redis)`, `kind = regex`. 

2, Later, the admin wants to broaden the filter to allow everything, and sends only the pattern:

```
PUT /api/v2.0/projects/dockerhub-proxy
{
  "metadata": {
    "proxy_cache_filter_pattern": "*"
  }
}

```

What happens:

1. kind is not in the request → validator defaults it to `doublestar`.
2. `*` is a valid doublestar pattern → validation passes.
3. But the DB merge keeps the stored `kind = regex`.

src/server/middleware/repoproxy/tag.go
```diff
@@ -35,12 +35,18 @@ func TagsListMiddleware() func(http.Handler) http.Handler {
...
 
-		art, p, _, err := preCheck(ctx, false)
+		art, p, _, err := preCheck(ctx, true)
 		if err != nil {
 			libhttp.SendError(w, err)
 			return
 		}
 
+		// Apply repository filter: if proxy_cache_filter_pattern is set, the repository must match
+		if err := checkRepositoryFilter(p, art); err != nil {
```
wy65701436, line 56: The repository name here is not normalized the same way as in the manifest/blob paths. Please apply the same `library/` normalization before calling the filter here.

src/server/v2.0/handler/project.go
stonezdj: Unlike the /api/v2.0/configurations -- it only send back the updated field.  the project metadata configuration api always requires to send back the full content of the project_metadata. so when you updated the pattern, the  proxy_cache_filter_pattern will send to the backend, if the proxy_cache_filter_kind is not send, it will default to doublestar type. in the UI, it will not send a request like step 2.

stonezdj: will remove the default value for kind=doublestar if it is empty.

wy65701436 (approved):
lgtm

PR #23526, opened 2026-07-09
fix(ui): No SBOM for multi-arch images

mlimardo1984:
Multi-arch image index rows showed "No SBOM" even after a successful scan, because SBOM accessories live on per-architecture child manifests, not on the index itself.

Fix: When a multi-arch artifact has `scan_status: Success` but no index-level SBOM digest, show "SBOM Details" instead of "No SBOM". Clicking it navigates to the per-architecture child list. Single-arch behavior is unchanged.

Changed: artifact list SBOM bar wiring, SBOM histogram display logic, i18n tooltip text, and unit tests. Portal-only; no backend changes.

Please backport to v2.15, v2.14, and v2.13.

# Issue being fixed
Fixes #([22790](https://github.com/goharbor/harbor/issues/22790))

# Manual test

- Deploy Harbor from this PR:
```
{
  "auth_mode": "db_auth",
  "banner_message": "",
  "current_time": "2026-07-09T08:27:14.470Z",
  "external_url": "https://mlimardo-factual-bluejay-worker-1.eastus.cloudapp.azure.com",
  "harbor_version": "v2.16.0-2f47c5de",
  "has_ca_root": false,
  "notification_enable": true,
  "oidc_provider_name": "",
  "primary_auth_mode": false,
  "project_creation_restriction": "everyone",
  "read_only": false,
  "registry_storage_provider_name": "filesystem",
  "registry_url": "mlimardo-factual-bluejay-worker-1.eastus.cloudapp.azure.com",
  "self_registration": false
}
```
- Push multi-arch image and generate SBOM:
Screenshot 2026-07-09 at 10 22 50
Screenshot 2026-07-09 at 10 23 00
Screenshot 2026-07-09 at 10 23 10
Screenshot 2026-07-09 at 10 23 18

Please indicate you've done the following:
- [x] Well Written Title and Summary of the PR
- [x] Label the PR as needed. "release-note/ignore-for-release, release-note/new-feature, release-note/update, release-note/enhancement, release-note/community, release-note/breaking-change, release-note/docs, release-note/infra, release-note/deprecation"
- [x] Accepted the DCO. Commits without the DCO will delay acceptance.
- [ ] Made sure tests are passing and test coverage is added if needed.
- [ ] Considered the docs impact and opened a new docs issue or PR with docs changes if needed in [website repository](https://github.com/goharbor/website).

PR #23540, opened 2026-07-10
fix: implement signature inheritance for OCI index children

Aloui-Ikram:
When a multi-arch OCI index is signed, the signature is attached to the index digest only. The per-arch children carry no signature of their own, so Harbor shows every child as unsigned even though the index above them is signed. This holds for both cosign and notation, since neither one descends into the children of an index by default.

The goal here is to surface the parent's signature on the children without claiming the children are signed.

## Backend

The artifact controller resolves the parent indexes of an artifact and, if one of them carries a signature, exposes it in a new `inherited_accessories` field. Those entries keep the parent index as their `subject_artifact_digest`, so they stay distinguishable from a direct signature and no consumer can mistake one for a signature that verifies against the child.

The artifact's own `accessories` list is left alone. A child digest still returns an empty list, which is the truth about what the registry holds and matches what `cosign verify` and `notation verify` report for that digest.

Resolving the field costs one reference lookup per artifact, so it is opt-in behind a new `with_inherited_accessory` query parameter, off by default. An artifact that no index references short-circuits after a single indexed lookup. `with_accessory` could not serve as the gate, because it defaults to true and is also set on the copy path.

Responses that do not ask for the field are unchanged.

## Portal

The Signed column gains a third state. A directly signed artifact stays green. An artifact covered only by a signature on its parent index gets a distinct grey icon and a tooltip saying where the signature comes from. Anything else stays as it was.

The children view now fetches accessories inline with the artifact instead of issuing one extra request per child.

## Both signing tools

cosign and notation are handled the same way, since the Portal already treats both as signed. Inheritance covers `signature.cosign` and `signature.notation`, and so does the check that skips artifacts already signed in their own right. Accessory kinds that are not signatures stay excluded, because an SBOM or an accelerator on an index says nothing about the children of that index.

## Verification

Tested on a local instance against a signed 16 child index, for both cosign and notation:

- `cosign verify` and `notation verify` pass on the index and fail on a child, which is what the UI and the API now say too
- a child's `accessories` stays empty, and `inherited_accessories` carries the parent's signature with the index digest as `subject_artifact_digest`
- the default response carries no new key

### Before, a signed index with its children shown as unsigned

01-notation-BEFORE-light-grid

### After, the children show the inherited state

02-notation-AFTER-light-grid

### The inherited icon and its tooltip

06-tooltip-dark

### The parent index next to its children

07-parent-index-light

# Issue being fixed
Related discussion: #23526 (signature counterpart of the SBOM inheritance problem discussed there)

- [x] Label the PR as needed. "release-note/update"


wy65701436:
@Aloui-Ikram thanks for your contribution, I would like to open discuss regarding the behavior.

A cosign signature is tied to one specific digest — when you sign a multi-arch image, only the index digest gets signed. The per-arch children have no signature of their own, so `cosign verify` on a child digest would actually fail. Harbor just shows the truth.

If Harbor marked the children as "signed", the UI would say one thing while cosign says another — that's misleading and could break security policies.

And I just look at the cosign(v3.1.1), it provides a flag to sign the index and all children, and Harbor will show everything as signed.

```
    -r, --recursive=false:
        if a multi-arch image is specified, additionally sign each discrete image
```

I will try to learn how other registries work, like ACR, ECR, Quay, and etc.

wy65701436:
@Aloui-Ikram 

After researching this, I’d like to use an example from this [tech blog](https://some-natalie.dev/blog/sigstore-multiarch/) to explain my point.

When you sign a multi-arch image, default Cosign behavior only targets the top-level OCI Index (the parent). If you try to verify a specific child architecture digest directly using the cosign client, it fails:

`
# ✅ verifying the "manifest of manifests" works on the tag
$ cosign verify \
  --certificate-oidc-issuer=https://token.actions.githubusercontent.com \
  --certificate-identity=https://github.com/github/github-mcp-server/.github/workflows/docker-publish.yml@refs/tags/v0.6.0 \
  ghcr.io/github/github-mcp-server:v0.6.0

Verification for ghcr.io/github/github-mcp-server:v0.6.0 --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - Existence of the claims in the transparency log was verified offline
  - The code-signing certificate was verified using trusted certificate authority certificates

# ✅ verifying the "manifest of manifests" works on the shasum too
$ cosign verify \
  --certificate-oidc-issuer=https://token.actions.githubusercontent.com \
  --certificate-identity=https://github.com/github/github-mcp-server/.github/workflows/docker-publish.yml@refs/tags/v0.6.0 \
  ghcr.io/github/github-mcp-server@sha256:db17de30c03103dd8b40ba63686a899270a79a1487d2304db0314e5c716118cf

Verification for ghcr.io/github/github-mcp-server:v0.6.0 --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - Existence of the claims in the transparency log was verified offline
  - The code-signing certificate was verified using trusted certificate authority certificates

# ❌ but verifying any one of the architectures doesn't
$ cosign verify \
  --certificate-oidc-issuer=https://token.actions.githubusercontent.com \
  --certificate-identity=https://github.com/github/github-mcp-server/.github/workflows/docker-publish.yml@refs/tags/v0.6.0 \
  ghcr.io/github/github-mcp-server@sha256:62684e9098550da194d7155d69af2499c402cc57e2302173a148a4ce348a0021

Error: no signatures found
error during command execution: no signatures found
`

As this example demonstrates, if the Harbor UI/API reports that all sub-artifacts (children) are signed, this is a custom abstraction. It behaves differently from GHCR (which shows them as unsigned) and conflicts with how the cosign CLI tool evaluates child digests natively. 

We need to decide how we want to align or document this difference.

wy65701436 (changes requested):
pending for discuss, I've added my comments.

Aloui-Ikram:
Hi @wy65701436, thank you for taking the time to research this and for opening the discussion , I really appreciate it. You're right on the technical facts: with default cosign only the index digest is signed, and `cosign verify` on a child digest fails. I fully agree with that, and I did some digging of my own on both the registry side and the verifier side, which I'd like to share before we decide.

**Registries:** Interestingly, GHCR itself doesn't display any signed/unsigned status in its UI , cosign signatures just appear as extra `sha256-*.sig` package versions, with no signature indicator on any manifest. (The failing check in the blog example is the cosign CLI against a child digest, which behaves the same regardless of registry UI.) Docker Hub, Quay, ECR, ACR and Artifactory are similar: no per-manifest signed status, only raw signature artifacts/referrers listed under the digest they reference. So Harbor seems to be ahead of other registries here ,we have an explicit "Signed" column that most don't. The closest comparable case is GitLab, which has a signed badge like ours, and they have an open issue to display it for multi-arch images based on the index-level signature: https://gitlab.com/gitlab-org/gitlab/-/issues/510998

**Verifiers:** Kubernetes admission controllers (Kyverno, sigstore policy-controller) verify the digest the workload references , for a multi-arch image that is the index digest, and they don't recurse into the per-arch children. In kyverno/kyverno#12630 a user had signed only the per-arch children (not the index) and Kyverno rejected the image; the maintainer explained this is correct sigstore trust-model behavior, because the index pins the children's digests and is the object an attacker would swap. So in Kubernetes enforcement, the index signature is the one that gets verified for every architecture deployed. To be fair, there is a counter-example: podman verifies the per-arch manifest it actually pulls, so index-only signing fails there and needs `--recursive` (containers/podman#21209). And on the cosign side, guidance for image indexes is still an open topic , see sigstore/cosign#3552, where signing both the index and the manifests is described as "the most compatible option" precisely because there's no settled convention yet.

**How images are signed in practice:** I also checked the registry directly for the image from the blog you linked: for `ghcr.io/github/github-mcp-server:v0.6.0`, the index digest (`db17de…`) has a cosign `.sig` tag, while neither the amd64 nor the arm64 child does , index-only signing, no `--recursive`. That seems to be the common pattern for public multi-arch images today.

So I think both statements are true at once: the child digest has no signature of its own (as you said), and the child *is* covered by the parent's signature through content-addressing , which is what Kubernetes policy enforcement relies on in production. **_The PR tried to stay on the safe side of that line: children are not marked as signed in Harbor's data. The parent's signature goes into a separate display-only `InheritedAccessories` field, the real accessories list is untouched (copy/delete/GC behave exactly as before), and the accessory returned by the API still carries the parent's `subject_artifact_digest`, so it remains distinguishable from a direct signature._**

That said, you make a fair point that the UI currently renders it with the same icon as a direct signature, and that can mislead. So to align and document, may I propose:

1. Show a distinct indicator on children , e.g. a different icon/tooltip like "signature inherited from parent index" instead of the standard signed icon, so Harbor never implies `cosign verify <child-digest>` would pass. The portal can detect this with no API change, by comparing the accessory's `subject_artifact_digest` with the child's digest.
2. Document the behavior and the difference vs. the cosign CLI / podman in the accessory docs.

On `-r/--recursive`: agreed, that's the right fix at signing time , but it's publisher-side. A Harbor operator can't re-sign images that are already pushed, mirrored, or third-party, and as the github-mcp-server example above shows, public multi-arch images are commonly signed index-only. That's the gap this PR is trying to close.

Would (1) + (2) address your concern? I'm happy to implement both in this PR, and of course open to other directions if you have a different preference.

wy65701436:
Hi @Aloui-Ikram , thanks for the great research and useful data points!

I agree with proposals (1) and (2)—an inherited signature should definitely look different from a direct one.

Before we decide, I want to quickly verify two things:
- Verifier behavior: How tools like Kyverno and sigstore actually verify multi-arch images during admission.
- Other registries: How ACR, ECR, Quay, and GitLab display signature status for child images.

 I'll follow up once I've completed the investigation.

Vad1mo:
I think there is a misunderstanding about what problem this feature solves.
 
It does not alter the signature or its verification; it is just a UX/UI improvement in the Portal with how/what we display signature icons for users.

The user expectation is the following: I user signs a multi-arch image, but they see in the harbor UI that the artifacts are displayed as not signed, and there is no obvious explanation on how and why this is happening. 

An alternative approach:

1. Do not display any icon (change in the Portal UI needed, unclear to the user if the artifact is signed)
2. Display a green icon with a (tooltip that indicates this is a derived signature from the index)
3. Display a different colored icon (it is not clear to the user what this color means)


For me, options 1 or 2 are equally fine. Option 3 does not resolve users confusion.

wy65701436:
I think we're mostly on the same page about the goal, but one clarification first: the current PR is not portal-only. It changes the backend — `populateAccessories` in the artifact controller does a parent-index lookup, and `ListAccessories` on the public API returns the parent's signature for a child digest that has no signature of its own. Any API consumer querying `type=signature.cosign` (CI gates, policy scripts, not just the Portal) would get back a signature that `cosign verify` fails on for that digest. So this is a semantic change to the API, not just a display tweak. 

I mean, the accessory list should stay truthful to what's actually in the registry.

For the UX problem itself, I agree with your option 2, with one small adjustment: use a visually distinct icon (e.g., outlined instead of solid green) together with the tooltip, rather than the identical green icon. If a child shows the exact same icon as a directly-signed artifact, the UI makes the same misleading claim we're trying to avoid on the API — and a tooltip only helps if the user hovers. 

The nice thing is the Portal already has everything it needs — children are rendered inside the index detail view, where the parent's accessories are already loaded. So we can do this with zero API change: drop the controller/handler changes from the PR, keep the idea, implement it in the Portal.

@Vad1mo @Aloui-Ikram

Aloui-Ikram:
@wy65701436 thanks! One clarification on the icon: the current signed icon is already Clarity's outlined `check-circle`, just tinted green (`.signed { color: #00d40f }`) , there's no solid variant in use today.
So to make inherited visibly distinct, did you have in mind a different color (e.g.grey), a different shape, or making the direct-signature icon solid instead?

Vad1mo:
>The nice thing is the Portal already has everything it needs — children are rendered inside the index detail view, where the parent's accessories are already loaded. So we can do this with zero API change: drop the controller/handler changes from the PR, keep the idea, implement it in the Portal.

Do I understand you correctly that you would want to have it in UI/Portal only?

Vad1mo:
1. The OCI side is unaffected either way. The signature is an accessory whose subject is the index digest. This PR only changes the Harbor REST API view; the tag-scheme lookup (sha256-*.sig) and Harbor's own OCI referrers endpoint are untouched. Cosign, Kyverno, policy-controller all talk OCI, so no verification outcome changes with or without this PR. 

The debate is purely about what Harbor's management API and UI report, not about signature semantics.

2. On "the Portal already has everything it needs": not quite. In the index detail view the parent is fetched with withAccessory: false, and the Signed column comes from a per-child listAccessories call (checkCosignAndSbomAsync). The parent's accessories are not loaded there today. The Portal does have the parent digest (from the depth route param), so a Portal-only fix means one additional call to the existing accessories endpoint, not zero work. Still no API change.

**Option A: Portal-only**

Pros
- Harbor REST API stays consistent with the OCI layer: both give the same answer for a child digest, including Harbor's own referrers endpoint.
- No silent semantic change for existing API consumers that read "non-empty accessory list = signed". Note our own Portal reads it exactly that way, it counts results and never checks subject_artifact_digest, which suggests third-party consumers do the same.
- Small change: one extra request in the index detail view plus a third icon state, ~30 lines in one component.
- Inherited state is naturally distinct in the UI (own icon and tooltip), which is what we all agreed on anyway.

Cons
- One extra API request per index detail page load.
- The inheritance information exists only in the UI. API or Harbor-cli consumers who want it must reimplement the parent lookup themselves.
- Only the parent you navigated through is checked, not every index referencing the child (edge case).

**Option B: Backend (current PR)**

Pros
- Single source of truth: every Harbor API consumer and any future UI surface gets the inheritance for free and consistently.
- Handles children referenced by multiple indexes.
- The inherited accessory carries the parent's subject_artifact_digest, so it is distinguishable by consumers who know to check.

Cons
- Harbor's REST API and Harbor's own OCI referrers endpoint would disagree about the same digest. The accessory list stops being a faithful index of registry content.
- Existing consumers get the changed semantics silently. The only signal is the subject digest mismatch, a check nothing performs today because the mismatch could not previously occur.
- Extra parent lookup on the populateAccessories hot path for all consumers.
- Locks in one inheritance policy ("any signed parent") server-side.

**Option C: explicit inherited_accessories field + Portal icon**

Keep this PR's controller logic, drop the ListAccessories merge. Instead expose the already-computed InheritedAccessories as a new, explicit field: GET /artifacts/{ref}?with_accessory=true (and list) returns inherited_accessories: [...] alongside the existing accessories. The Portal reads it and renders the distinct icon.

Pros
- The accessory list stays truthful to registry content under all circumstances, which addresses the core objection. Inheritance is separate, self-describing data instead of entries smuggled into the old field.
- Additive and non-breaking: default responses are byte-identical to today, old clients ignore the unknown field, new clients opt in by reading it.
- All consumers get the same server-side semantics as the UI, including the multi-parent case, without reimplementing the parent walk.
- The Portal child view can get simpler: flipping the per-child getArtifact to withAccessory: true returns direct and inherited accessories in one call.
- Salvages most of this PR: the controller side survives nearly unchanged, only the handler merge is replaced by a swagger field.

Cons
- More total work: swagger + handler + docs + the Portal work from A, and a real API surface addition that needs proper API review.
- The perf cost needs gating: as written, populateAccessories fires the parent ListReferences lookup for every artifact without a cosign signature, which is most artifacts in most registries, on every list call. It should run only when with_accessory=true, or only for artifacts that actually have parent references.
- Locks in the "any signed parent counts" policy as documented API behavior.

**Proposal**

Both options keep the API truthful and both end user confusion in the UI. A is the fastest merge; C additionally makes the inheritance queryable for CI scripts and dashboards and keeps API and UI telling exactly the same story. 

**My primary concern with the  pure A solution is that nowadays the Harbor UI is only one interface for the user, CLIs, agents, CI-flows and many more 3rd party tools that consume the Harbor REST API.**

**So, in my opinion, the API should expose the UI as close as possible, and the UI should not have business logic.. Option C is the one that can do it all together.** 

@wy65701436 which of the two would you accept? @Aloui-Ikram would you be OK reworking the PR accordingly?

Aloui-Ikram:
@Vad1mo  @wy65701436 
Reworked as Option C , 
On gating: `with_accessory` defaults to true and copyDeeply sets it, so gating on it would still run the parent lookup on nearly every list and copy. I used a separate `with_inherited_accessory` param (default false) instead, plus the short-circuit for artifacts with no parent index.

That's one more public param than you proposed ,  OK with you?

image

wy65701436:
@Aloui-Ikram thanks for the rework — this is exactly the direction we discussed, and it resolves my main concern.

wy65701436:
My key requirement was that UI and API must never disagree with what cosign verify says, and this now holds at every layer:

- API: accessories for a child digest stays empty — truthful to the registry, consistent with cosign verify against the child digest failing. The parent's signature is only exposed in the separate, opt-in inherited_accessories field, and it keeps the parent index as subject_artifact_digest, so it's self-describing: no consumer can mistake it for a verifiable signature on the child.
- UI: the child is not shown as "signed" — it gets a third, visually distinct state (grey icon + tooltip explaining the signature is derived from the parent index).
- Default responses are byte-identical to today, so existing API consumers are unaffected.

Also agree with your gating decision: since with_accessory defaults to true and is set on the copy path, a separate with_inherited_accessory param (default false) is the right call.

wy65701436:
Two minor points before approval:

1. Notation: the inheritance logic only considers signature.cosign, but notation signatures on an index have the same problem, and the portal already treats notation as "signed". Either extend both the own-signature check and the inheritance to include notation, or document cosign-only as the scope and track notation as a follow-up — your preference.
2. nice to have: the inherited icon color is hardcoded, which may not read well in dark theme — consider a Clarity token if a suitable one exists.

Otherwise LGTM

wy65701436:
I will mark this one as 2.16, thanks for the contribution.

Aloui-Ikram:
@wy65701436  , thanks. I extended notation rather than deferring it, so both points are in  `1e4dba6`.

 1. **Notation**. Inheritance now covers signature.notation as well as signature.cosign, in both places you mentioned. The own-signature check mattered on its own: a child with its own notation signature under a cosign-signed index was not counted as directly signed, so it got an inherited entry it should not have. Non-signature accessories stay excluded.

 2. **Icon colour**. Now var(--cds-alias-status-neutral), which resolves to construction-600 in light and construction-300 in dark, so it follows the theme.

The tooltip and the 10 locale strings needed no change, the wording was already tool-agnostic.
Verified on a local instance with a notation-signed 16 child index: notation verify passes on the index and fails on a child, the child's accessories stays empty, inherited_accessories carries the notation signature with the index digest as subject_artifact_digest, and the default response has no new key. Added three controller tests.

  ### Notation, before and after
01-notation-BEFORE-light-grid

02-notation-AFTER-light-grid

  ### Inherited icon, before and after
dark :
05-token-dark-BEFORE
04-token-dark-AFTER

light :

05-token-light-BEFORE
04-token-light-AFTER

These differ only by forcing color: #8c8c8c on the same page.

wy65701436 (approved):
lgtm

wy65701436:
lgtm,

but one follow-up, when both options are enabled and the artifact has no accessories of its own, populateAccessories leaves Accessories as nil, so populateInheritedAccessories re-runs the same SubjectArtifactID query — one redundant lookup for exactly the target case (unsigned children). Normalizing the empty result to a non-nil slice would avoid it.

wy65701436:
thanks @Aloui-Ikram for your contribution.

chlins (approved):
lgtm

PR #23542, opened 2026-07-10
Fix grammar: 'can not be override' -> 'cannot be overridden' in comment

magic-peach:
Fix grammar in a source code comment in utils.go: 'can not be override' should be 'cannot be overridden'.


Vad1mo (approved):

chlins (approved):
lgtm

PR #23545, opened 2026-07-11
ci: migrate all GitHub workflows to CNCF-hosted runners

Vad1mo:

Migrates every workflow job to the CNCF-hosted runner pool, and fixes Go caching in CI, which was silently disabled.

CNCF recommends that graduated projects move their CI onto the CNCF-provided runner pool. Support and coordination for these runners happens in the CNCF Slack channel [#cncf-ci-infra](https://cloud-native.slack.com/archives/C08P4HUFQ6M).

## 1. Runner migration

| Old label | New label |
|---|---|
| `oracle-vm-24cpu-96gb-x86-64` | `cncf-ubuntu-16-64-x86` |
| `ubuntu-latest` | `cncf-ubuntu-16-64-x86` |
| `ubuntu-24.04-arm` | `cncf-ubuntu-16-64-arm` |

All 22 `runs-on` declarations across 11 workflow files are covered. The `pass-CI.yml` jobs are no-op stubs that only satisfy required status checks on skipped paths, so they use the smallest runner, `cncf-ubuntu-2-8-x86`. The `arm64` leg of the `BUILD_PACKAGE` matrix still builds natively, now on `cncf-ubuntu-16-64-arm`.

Note the new runners are **smaller** than the old ones: 16 vCPU / 64 GB versus Oracle's 24 vCPU / 96 GB. They are nevertheless faster on every job (see below).

## 2. Go caching was silently disabled

`actions/setup-go` ran *before* `actions/checkout`, so it looked for `go.sum` in an empty workspace:

```
##[warning]Restore cache failed: Dependencies file is not found in
/home/ubuntu/_work/harbor/harbor. Supported file pattern: go.sum
```

With `cache: true` (the v5 default) this fails open and no-ops, so `GOMODCACHE` and `GOCACHE` were cold on **every job of every run**. The same ordering bug was present in `build-package.yml` and `conformance_test.yml`. All 7 `setup-go` usages now run after checkout with `cache-dependency-path` set.

`$GOPATH/bin` is cached too. `cover` and `goveralls` were installed on every run but never used (`goveralls` appears only in a commented-out line in `ut_run.sh`; `go test -cover` uses the built-in tool, not `golang.org/x/tools/cmd/cover`), so they are dropped. `misspell`, which `make misspell` does invoke, is pinned rather than tracking `@latest`, so the cached tool version is reproducible.

## 3. Measured before / after

Baseline is 12 successful `main` runs on the old Oracle runners (very stable: UTTEST varied only 34.6-35.1 min). All other columns are single runs. All jobs green in the final column.

| Job | A: Oracle 24c/96g, no cache (n=12, median) | B: CNCF 16c/64g, no cache (n=1) | C: CNCF + cache, **cold** (n=1) | E: CNCF + cache, **warm** (n=1) | A → E |
|---|---|---|---|---|---|
| UTTEST | 34.7 | 33.6 | 34.0 | **29.5** | −15% |
| APITEST_DB | 30.0 | 27.9 | 27.6 | **28.4** | −5% |
| APITEST_DB_PROXY_CACHE | 17.4 | 15.9 | 15.7 | **16.0** | −8% |
| OFFLINE | 10.5 | 8.7 | 8.9 | **8.8** | −16% |
| APITEST_LDAP | 10.2 | 8.4 | 8.2 | **8.3** | −19% |
| UI_UT | 1.9 | 1.8 | (failed, see below) | **1.8** | −5% |
| **Total job-minutes** | **104.7** | 96.3 | — | **92.8** | **−11%** |
| **Critical path (wall clock)** | **34.7** | 33.6 | — | **29.5** | **−15%** |

Separating the two effects, so neither is credited with the other's gain:

- **Runner change alone** (A → B): total job-minutes −8%, critical path −3%. The API/OFFLINE jobs gain 16-19%; UTTEST barely moves because it is dominated by a cold Go build.
- **Caching alone** (B → E): critical path −12%. UTTEST 33.6 → 29.5 (`install` 5.0 → 3.7, `script` 29.2 → 24.4).

A cold run (C) is expected to match the baseline: it writes the cache but cannot read it. It confirmed the mechanism works (`Cache saved with the key: setup-go-Linux-x64--go-1.26.4-...`), where previously the same step logged `Restore cache failed`.

The `$GOPATH/bin` tool cache still missed in run E (its key changed with the last edit to `ut_install.sh`), so it will save a little more once that key settles.

## 4. Two real bugs found while measuring

**`make lint` depended on a broken path.** `GOLANGCI_LINT` was `$(shell go env GOPATH)/bin/golangci-lint`, but CI sets `GOPATH` to a colon-separated list, so this expanded to the nonexistent `/home/ubuntu/go:/home/ubuntu/_work/harbor/harbor/bin/golangci-lint`. It only worked because `ut_install.sh` *installed* the binary using the same broken expression, so both sides agreed on the same bogus location. Tools are now resolved against the first `GOPATH` entry, where `go install` actually writes them. This would break for any developer whose `GOPATH` has more than one entry.

**Persisting `GOCACHE` made `go test` stop running the tests.** Go's build cache also caches *test results*. With `GOCACHE` restored across runs, `go test` replayed cached passes instead of executing: a warm run printed test output stamped with timestamps from the *previous* run and finished the test step in 2.2 min instead of 29. It then failed on `TestConfig`, because Harbor's packages share one Postgres instance and a package that did re-run observed state the replayed packages never created. Left unnoticed, this would have turned CI into a rubber stamp that reported a fake 80% speedup.

`tests/coverage4gotest.sh` now passes `-count=1`, which disables test-result reuse while keeping the compilation cache. That is why the honest warm number is −15% and not −80%. The final run was verified: every in-test log timestamp falls after the job's start time, and `-count=1` is applied to all 479 test packages.

## 5. Flaky test fixed

`UI_UT` failed once on `ProjectAuditLegacyLogComponent should support pagination`. The spec queried `.pagination-next` immediately after `whenStable()` and clicked it, so it broke whenever the datagrid had not rendered yet - more likely on 16-core runners than on the old 24-core ones. The lookup now retries until the control appears; the assertion still fails if it never renders. Passed on both subsequent runs.

# Issue being fixed
N/A - infrastructure migration, no associated issue.

### DIFF
--- a/src/portal/src/app/base/project/project-log/audit-legacy-log.component.spec.ts
+++ b/src/portal/src/app/base/project/project-log/audit-legacy-log.component.spec.ts
@@ -165,11 +165,12 @@ describe('ProjectAuditLegacyLogComponent', () => {
     });
     it('should support pagination', async () => {
         fixture.autoDetectChanges(true);
-        await fixture.whenStable();
-        fixture.detectChanges();
-        await fixture.whenStable();
-        const el: HTMLButtonElement =
-            fixture.nativeElement.querySelector('.pagination-next');
+        let el: HTMLButtonElement = null;
+        for (let i = 0; i < 20 && !el; i++) {
+            fixture.detectChanges();
+            await fixture.whenStable();
+            el = fixture.nativeElement.querySelector('.pagination-next');
+        }
         expect(el).toBeTruthy();
         el.click();
         fixture.detectChanges();
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -50,14 +50,20 @@ jobs:
       - cncf-ubuntu-16-64-x86
     timeout-minutes: 100
     steps:
+      - uses: actions/checkout@v7
+        with:
+         path: src/github.com/goharbor/harbor
       - name: Set up Go 1.26
         uses: actions/setup-go@v5
         with:
            go-version: 1.26.4
+           cache-dependency-path: src/github.com/goharbor/harbor/src/go.sum
         id: go
-      - uses: actions/checkout@v7
+      - name: Cache CI Go tools
+        uses: actions/cache@v4
         with:
-         path: src/github.com/goharbor/harbor
+          path: ~/go/bin
+          key: ci-gotools-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('src/github.com/goharbor/harbor/tests/ci/ut_install.sh') }}
       - name: setup env
         run: |
           cd src/github.com/goharbor/harbor
@@ -112,14 +118,15 @@ jobs:
       - cncf-ubuntu-16-64-x86
     timeout-minutes: 100
     steps:
+      - uses: actions/checkout@v7
+        with:
+          path: src/github.com/goharbor/harbor
       - name: Set up Go 1.23
         uses: actions/setup-go@v5
         with:
           go-version: 1.23.2
+          cache-dependency-path: src/github.com/goharbor/harbor/src/go.sum
         id: go
-      - uses: actions/checkout@v7
-        with:
-          path: src/github.com/goharbor/harbor
       - name: setup env
         run: |
           cd src/github.com/goharbor/harbor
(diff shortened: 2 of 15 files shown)


src/portal/src/app/base/project/project-log/audit-legacy-log.component.spec.ts
```diff
@@ -165,11 +165,12 @@ describe('ProjectAuditLegacyLogComponent', () => {
     });
     it('should support pagination', async () => {
         fixture.autoDetectChanges(true);
-        await fixture.whenStable();
-        fixture.detectChanges();
-        await fixture.whenStable();
-        const el: HTMLButtonElement =
-            fixture.nativeElement.querySelector('.pagination-next');
+        let el: HTMLButtonElement = null;
```
wy65701436, line 168: is this needed for the migrating?

Vad1mo: no, but the existing code cause a few time the pipeline to break because of this timing bug. So I dediced to let it fix

wy65701436 (approved):
lgtm

chlins (approved):
lgtm

PR #23571, opened 2026-07-17
fix: address unsafe type assertions causing potential panics across packages

Norway-02:

Thank you for the kind suggestion, @wy65701436! This PR consolidates all previous individual fixes into a single PR that addresses the **generic problem of unsafe type assertions** across the codebase, which can cause runtime panics (HTTP 500) when the underlying interface type does not match the expected type.

## Problem

Throughout several packages, values stored as `interface{}` (in maps, filter structures, ORM results, etc.) were directly type-asserted without a safety check. For example:

```go
// Before: Direct assertion, panics if wrong type
return val.(float64)
cronItem.(string)
filter.Value.(string)
value.(int64)
```

If any of these values contain an unexpected type — due to API anomalies, malformed payloads, ORM driver changes, or JSON unmarshaling quirks — the process panics with a fatal error, causing the entire request to fail and often crashing the goroutine.

## Files Changed

| File | Fix |
|------|-----|
| `src/pkg/retention/manager.go` | Safe type assert for cron trigger setting |
| `src/pkg/scan/postprocessors/report_converters.go` | Safe type assert for CVSS v3 score |
| `src/pkg/retention/policy/models.go` | Safe type assert in retention policy validation |
| `src/pkg/project/dao/dao.go` | Safe type assert for ORM role values |
| `src/pkg/reg/filter/artifact.go` | Safe type assert for artifact filter values |
| `src/pkg/reg/filter/repository.go` | Safe type assert for repository filter values |

## Solution

Each unsafe assertion is now guarded by a two-value type assertion, and failures return a proper error rather than panicking:

```go
// After: Safe assertion, returns error if wrong type
if val, ok := v.(float64); ok {
    return val
}
if cronStr, isStr := cronItem.(string); isStr {
    // use cronStr safely
}
```

### DIFF
--- a/src/pkg/project/dao/dao.go
+++ b/src/pkg/project/dao/dao.go
@@ -197,7 +197,9 @@ func (d *dao) ListRoles(ctx context.Context, projectID int64, userID int, groupI
 
 	var roles []int
 	for _, value := range values {
-		roles = append(roles, int(value.(int64)))
+		if roleInt64, ok := value.(int64); ok {
+			roles = append(roles, int(roleInt64))
+		}
 	}
 
 	return roles, nil
--- a/src/pkg/reg/filter/artifact.go
+++ b/src/pkg/reg/filter/artifact.go
@@ -15,6 +15,7 @@
 package filter
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/goharbor/harbor/src/pkg/reg/model"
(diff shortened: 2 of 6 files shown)


Norway-02:
@wy65701436 As suggested, I've consolidated all the individual unsafe type assertion fixes into this single PR. It addresses the generic problem across 6 files in one go. @chlins @stonezdj @Vad1mo @OrlinVasilev — would you be able to review this?

wy65701436 (approved):
lgtm, thanks.

stonezdj (approved):

PR #23573, opened 2026-07-17
fix: safely type assert context value in SetMetricOpID to avoid panic

Norway-02:
# Summary

This PR fixes a potential **nil pointer panic** in `SetMetricOpID` in `src/server/middleware/metric/metric.go`.

## The Bug

`SetMetricOpID` retrieves a context value and directly asserts it to `*string` without a nil or type check:

```go
func SetMetricOpID(ctx context.Context, value string) {
    if config.Metric().Enabled {
        v := ctx.Value(contextOpIDKey{}).(*string) // <-- PANIC if key not in context
        *v = value
    }
}
```

`ctx.Value()` returns `nil` if the key was never injected into the context (e.g. if `SetMetricOpID` is called from a code path that does not go through the `instrumentHandler` middleware). The direct assertion `nil.(*string)` causes a **runtime panic**.

## The Fix

Replace the bare assertion with a safe two-value type assertion:

```go
func SetMetricOpID(ctx context.Context, value string) {
    if config.Metric().Enabled {
        if v, ok := ctx.Value(contextOpIDKey{}).(*string); ok && v != nil {
            *v = value
        }
    }
}
```

This ensures that if the context key is absent or holds an unexpected type, the function silently no-ops instead of panicking, which is the correct defensive behaviour for a metrics helper.

## Issue being fixed

Fixes a potential runtime panic in the metrics middleware when `SetMetricOpID` is called outside of the `instrumentHandler` context.

### DIFF
--- a/src/server/middleware/metric/metric.go
+++ b/src/server/middleware/metric/metric.go
@@ -48,8 +48,9 @@ const (
 // SetMetricOpID used to set operation ID for metrics
 func SetMetricOpID(ctx context.Context, value string) {
 	if config.Metric().Enabled {
-		v := ctx.Value(contextOpIDKey{}).(*string)
-		*v = value
+		if v, ok := ctx.Value(contextOpIDKey{}).(*string); ok && v != nil {
+			*v = value
+		}
 	}
 }
 


Norway-02:
@chlins @stonezdj @wy65701436 @Vad1mo @OrlinVasilev Could you please take a look at this PR?

Norway-02:
Superseded by #23585 (consolidated PR to simplify review/CI per maintainer request).

PR #23584, opened 2026-07-20
fix: safeguard against missing id_token in OIDC auth

Norway-02:
# Summary

This PR fixes an unsafe type assertion in the OIDC helper logic that could cause a runtime panic during user authentication.

## The Bug

In `src/pkg/oidc/helper.go`, the `id_token` is extracted from the `oauthToken.Extra` properties and blindly cast to a string:

```go
RawIDToken: oauthToken.Extra("id_token").(string)
```

If the OIDC identity provider fails to return an `id_token`, or returns it in an unexpected type (e.g. `nil`), this type assertion will trigger a panic and crash the Harbor core instance during login flows.

## The Fix

Introduced the safe type-assertion idiom (`ok` check) when extracting the `id_token`. If it is missing or invalid, it gracefully returns a formatted error rather than crashing.

## Issue being fixed

Fixes potential runtime panics during OIDC authentication when the identity provider responds with malformed tokens.

Please indicate you've done the following:

- [x] Well Written Title and Summary of the PR
- [x] Label the PR as needed: `release-note/ignore-for-release`
- [x] Accepted the DCO. Commits without the DCO will delay acceptance.
- [x] Made sure tests are passing and test coverage is added if needed.
- [x] Considered the docs impact and opened a new docs issue or PR with docs changes if needed in website repository.

PR #23583, opened 2026-07-20
fix: safeguard against panics in Casbin keyMatch2Func

Norway-02:
# Summary

This PR fixes unsafe type assertions in the Casbin matcher (`keyMatch2Func`), which could cause a runtime panic if invalid arguments are passed from the RBAC policy evaluator.

## The Bug

In `src/pkg/permission/evaluator/rbac/casbin_match.go`, the custom Casbin matcher function blindly asserted its arguments as strings:

```go
func keyMatch2Func(args ...any) (any, error) {
	name1 := args[0].(string)
	name2 := args[1].(string)
	...
```

If the Casbin policy engine passes fewer than two arguments or arguments of different types (e.g. `nil`), this will trigger an immediate runtime panic, crashing the Harbor core server instead of gracefully returning an error.

## The Fix

Introduced standard bounds checking (`len(args) != 2`) and safe type assertions (`ok1`, `ok2`) for the arguments. If the check fails, it now safely returns a formatted error instead of crashing.

## Issue being fixed

Fixes potential runtime panics during RBAC policy evaluation when the Casbin matcher is invoked with malformed arguments.

Please indicate you've done the following:

- [x] Well Written Title and Summary of the PR
- [x] Label the PR as needed: `release-note/ignore-for-release`
- [x] Accepted the DCO. Commits without the DCO will delay acceptance.
- [x] Made sure tests are passing and test coverage is added if needed.
- [x] Considered the docs impact and opened a new docs issue or PR with docs changes if needed in website repository.

PR #23581, opened 2026-07-19
fix: safely type assert LDAP group config values

Norway-02:
# Summary

This PR introduces safe type assertions for LDAP group configurations during validation to prevent runtime panics.

## The Bug

In `src/pkg/config/validate/ldapgroup.go`, the configuration values for LDAP groups are unmarshaled into `map[string]any` via API input. The code previously assumed that certain fields were strings and forcefully cast them without checking:

```go
	if val, exist := cfgs[common.LDAPGroupSearchFilter]; exist {
		cfg.Filter = val.(string)
```

If a user maliciously or accidentally submitted a boolean, integer, or array for these fields, `val.(string)` would trigger a panic, crashing the Harbor core process.

## The Fix

Introduced the safe type-assertion idiom (`v, ok := val.(string)`) for LDAP group string configuration fields. If the provided value is not a string, a descriptive validation error is returned instead of panicking.

## Issue being fixed

Fixes potential runtime panics when processing incorrectly typed LDAP group configurations via the API.

Please indicate you've done the following:

- [x] Well Written Title and Summary of the PR
- [x] Label the PR as needed: `release-note/ignore-for-release`
- [x] Accepted the DCO. Commits without the DCO will delay acceptance.
- [x] Made sure tests are passing and test coverage is added if needed.
- [x] Considered the docs impact and opened a new docs issue or PR with docs changes if needed in website repository.

PR #23580, opened 2026-07-19
fix: safeguard against nil context value in scan data export

Norway-02:
# Summary

This PR fixes an unsafe type assertion in the scan data export execution flow which can cause a runtime panic.

## The Bug

In `src/controller/scandataexport/execution.go`, when starting an execution, the vendor ID is extracted from the context using an unsafe type assertion:

```go
vendorID := int64(ctx.Value(export.CsvJobVendorIDKey).(int))
```

If the context does not contain the `CsvJobVendorIDKey`, `ctx.Value()` returns `nil`, and the type assertion `.(int)` directly on `nil` results in a panic.

## The Fix

Introduced the safe type-assertion idiom (`ok` check) for `ctx.Value(export.CsvJobVendorIDKey)`. If it is not present or invalid, it gracefully returns a formatted error instead of crashing the process.

## Issue being fixed

Fixes a potential runtime panic when processing scan data export executions without a valid vendor ID in the context.

Please indicate you've done the following:

- [x] Well Written Title and Summary of the PR
- [x] Label the PR as needed: `release-note/ignore-for-release`
- [x] Accepted the DCO. Commits without the DCO will delay acceptance.
- [x] Made sure tests are passing and test coverage is added if needed.
- [x] Considered the docs impact and opened a new docs issue or PR with docs changes if needed in website repository.

PR #23585, opened 2026-07-20
fix: consolidate type assertion and nil pointer safeguards

Norway-02:
# Summary

As suggested by maintainers, this PR consolidates several small defensive type assertion and nil pointer guards into a single patch to simplify review and CI runs. It groups fixes for unchecked type assertions (comma-ok pattern) and missing nil-guards across various Harbor packages.

## The Bug

In several packages, context values, interface maps, and pointers were being accessed and directly asserted/dereferenced without checking their validity. 
For example:
- `ctx.Value(...).(string)`
- `ExtraAttrs["key"].(string)`
- `args[0].(string)`
- `p.Trigger.Kind` when `p.Trigger` is nil

These blind assertions and dereferences would trigger a runtime panic (fatal crash) if the underlying value was missing, nil, or of a different type than expected.

## The Fix

Introduced standard safe type-assertion idioms (e.g. `val, ok := obj.(string)`) and nil-guards. 
If an assertion fails or a required pointer is nil, the functions now safely exit or return an appropriately formatted `error` rather than crashing the Harbor process.

This encompasses the fixes from the following PRs:
- #23515: `ExtraAttrs["dry_run"]`
- #23573: `ctx.Value` in metric middleware
- #23576: nil `Trigger` in retention manager
- #23577: `ExtraAttrs` in scan data export
- #23580: `ctx.Value` in scan data export
- #23581: LDAP group config values
- #23583: Casbin `keyMatch2Func` args
- #23584: OIDC `id_token`

## Issue being fixed

Fixes potential runtime panics due to malformed input, missing configuration, or unexpected context values.


Norway-02:
@chlins Done! Thank you for the guidance. I have consolidated all 8 small PRs (Group 1 & Group 2) into this single PR and closed the individual ones to make review and CI much easier.

Norway-02:
@stonezdj Thank you as well for your review! I've consolidated them all here as requested.

patch_23515.patch
```diff
@@ -0,0 +1,96 @@
+diff --git a/src/controller/retention/controller.go b/src/controller/retention/controller.go
```
wy65701436: are they generated by AI?

src/controller/retention/controller.go
```diff
@@ -343,12 +343,17 @@ func convertExecution(exec *task.Execution) *retention.Execution {
 		EndTime:   exec.EndTime,
 		Status:    exec.Status,
 		Trigger:   exec.Trigger,
-		DryRun:    exec.ExtraAttrs["dry_run"].(bool),
```
wy65701436, line 346: why we hard code to false?

Norway-02:
@wy65701436 Thanks for reviewing! The value was actually being overridden to the correct value a few lines further down in the original diff, but I agree it looked confusing as a hardcoded default. I've refactored `convertExecution` in the latest commit to extract `dry_run` and `operator` from `ExtraAttrs` *before* initializing the struct, so it reads much cleaner now.

patch_23515.patch
Norway-02: Yes, I was experimenting with an AI coding assistant to help me find and safely refactor these type assertions across the codebase. I accidentally committed its internal scratchpad/working files in the previous push! I've completely removed them from the commit history now so they won't clutter the PR. Sorry for the noise!

src/controller/retention/controller.go
Norway-02: The value was actually being overridden to the correct value a few lines further down in the original diff, but I agree it looked confusing as a hardcoded default. I've refactored convertExecution in the latest commit to extract dry_run and operator from ExtraAttrs before initializing the struct, so it reads much cleaner now.

Norway-02:

PR #23591, opened 2026-07-21
fix: set MaxIdleConnsPerHost on internal HTTP transports

chlins:
### Comprehensive Summary

Go's `http.Transport` defaults `MaxIdleConnsPerHost` to 2. Harbor's internal traffic is dominated by single-host hotspots (core → registry reverse proxy, registry API client, jobservice → core hooks, scanner adapter client), so under high concurrency almost every connection is closed after use. The resulting churn piles up client-side `TIME_WAIT` sockets and eventually exhausts ephemeral ports:

```
http: proxy error: dial tcp 172.18.0.4:5000: connect: cannot assign requested address
```

Changes:

- `common/http/transport.go`: `newDefaultTransport()` now sets `MaxIdleConns=1000` / `MaxIdleConnsPerHost=200`, covering the registry client, core↔jobservice, webhook jobs and replication adapters.
- `server/registry/proxy.go`: the core → registry reverse proxy previously fell back to `http.DefaultTransport` when internal TLS is disabled; it now always gets a dedicated transport with `MaxIdleConns(PerHost)=1024`, as all `/v2/` data-plane traffic targets a single host.
- `jobservice/hook/hook_client.go` / `pkg/scan/rest/v1/client.go`: per-host idle pool now matches `MaxIdleConns` (20/100), since each client only talks to one endpoint.

These knobs only enable connection reuse — idle connections beyond actual need are still reaped by `IdleConnTimeout`, and peak connection count remains driven by concurrency. No behavior change.

### Issue being fixed

Observed on a docker-compose deployment under high-concurrency pulls; core floods `proxy error: context canceled` followed by `cannot assign requested address` on the core → registry hop.

### DIFF
--- a/src/common/http/transport.go
+++ b/src/common/http/transport.go
@@ -67,8 +67,14 @@ func newDefaultTransport() *http.Transport {
 			KeepAlive: 30 * time.Second,
 			DualStack: true,
 		}).DialContext,
-		TLSClientConfig:       &tls.Config{},
-		MaxIdleConns:          100,
+		TLSClientConfig: &tls.Config{},
+		MaxIdleConns:    1000,
+		// The default value of MaxIdleConnsPerHost is 2, which is too small for
+		// high-concurrency scenarios (e.g. core/jobservice talking to the registry,
+		// or replicating to a single remote endpoint). Idle connections beyond the
+		// need are reaped by IdleConnTimeout, so a generous value is safe. Keeping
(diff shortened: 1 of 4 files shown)


wy65701436 (approved):
lgtm, if we believe it is a bug fix, we can consider it to backport it into branch 2.15?

stonezdj (approved):

PR #23646, opened 2026-07-29
fix: allow resizing of the SBOM column in artifact table

### LINKED ISSUE #23634: The data column dislocated when adjusting the column width in the artifact table
Go to artifact table, click the sbom table column to adjust its size, its size can not adjustable and after that the data field dislocated.

Image

La1itchauhan:
i have  removed the `!important` modifier from the `.sbom-column` width rule in the artifact list table

The SBOM column width is currently defined with `!important`, which overrides the inline width applied by Clarity' s datagrid during column resizing.  As a result, the SBOM column cannot be resized correctly, causing the header and data columns to become misaligned.

Removing " !important " preserves the default width while allowing Clarity's resize logic to update the column width as expected.

## Issue being fixed

Fixes #23634

### DIFF
--- a/src/portal/src/app/base/project/repository/artifact/artifact-list-page/artifact-list/artifact-list-tab/artifact-list-tab.component.scss
+++ b/src/portal/src/app/base/project/repository/artifact/artifact-list-page/artifact-list/artifact-list-tab/artifact-list-tab.component.scss
@@ -163,7 +163,7 @@
 }
 
 .sbom-column {
-  width: 7rem !important;
+  width: 7rem;
 }
 
 .annotations-column {
--- a/src/portal/src/app/base/project/repository/artifact/artifact-list-page/artifact-list/artifact-list-tab/artifact-list-tab.component.scss
+++ b/src/portal/src/app/base/project/repository/artifact/artifact-list-page/artifact-list/artifact-list-tab/artifact-list-tab.component.scss
@@ -147,19 +147,19 @@
 }
 
 .pull-command-column {
-  width: 6rem !important;
+  width: 6rem;
 }
 
 .co-signed-column {
-  width: 7rem !important;
+  width: 7rem;
 }
 
 .size-column {
-  width: 6rem !important;
+  width: 6rem;
 }
 
 .vul-column {
-  width: 11rem !important;
+  width: 11rem;
 }
 
 .sbom-column {
@@ -167,7 +167,7 @@
 }
 
 .annotations-column {
-  width: 5rem !important;
+  width: 5rem;
 }
 
 .signed {


wy65701436:
cc @AllForNothing can you help to review?

src/portal/src/app/base/project/repository/artifact/artifact-list-page/artifact-list/artifact-list-tab/artifact-list-tab.component.scss
```diff
@@ -163,7 +163,7 @@
 }
 
 .sbom-column {
-  width: 7rem !important;
+  width: 7rem;
```
wy65701436, line 166: One question before this merges: all the sibling columns in the same file still use `!important`:

```
.pull-command-column { width: 6rem !important; }
.co-signed-column     { width: 7rem !important; }
.size-column          { width: 6rem !important; }
.vul-column           { width: 11rem !important; }
.annotations-column   { width: 5rem !important; }
```

If `!important` breaks resizing for the SBOM column, wouldn't it break it for these too? Could you confirm whether the issue affects all columns?

AllForNothing:
> cc @AllForNothing can you help to review?

@wy65701436 @La1itchauhan 

Based on my testing, in the current version of Harbor (which is actually tied to the version of the Clarity UI component library), !important can only fix the width of a column — it does not prevent the resize operation itself, as shown in my screenshot. In fact, performing a resize in this state causes the column styles to break/misalign.

Therefore, my suggestion is to remove !important from all column width declarations.

As a side note: in earlier versions, !important did prevent resizing — which is why it was added in the first place. That behavior no longer holds in the current version.


After resizing, the column is misaligned.
image

src/portal/src/app/base/project/repository/artifact/artifact-list-page/artifact-list/artifact-list-tab/artifact-list-tab.component.scss
AllForNothing: As my explanation,  all the "! important" above can be removed

La1itchauhan: Thanks for the detailed explanation! I verified the behavior locally on the current Harbor version as well. After resizing, the columns become misaligned when !important is present. I've updated the PR to remove !important from the related artifact table column width declarations as suggested. Thanks for the guidance!

AllForNothing (approved):
lgtm

wy65701436 (approved):
lgtm, thanks @AllForNothing

PR #22862, opened 2026-02-23
refac(artifact): Add `manifest` package for extensibility using registry pattern

vg006:
# Comprehensive Summary of your change

Fixes #22847

This PR refactors the `artifact` package within the controller as suggested in the issue.

## Summary of changes

1. Removed private methods in the `src/controller/artifact/abstract.go`
   
    - `abstractManifestV1Metadata()` - Docker V1 / schema1
    - `abstractManifestV2Metadata()` - Docker V2 / OCI image manifest
    - `abstractIndexMetadata()` - OCI index / Docker manifest list

2. Add `ManifestAbstractor` interface
    ```go
    type ManifestAbstractor interface {
	    // AbstractManifest abstracts the metadata for the specific artifact type into the artifact model,
	    AbstractManifestMetadata(ctx context.Context, artifact *artifact.Artifact, content []byte) error
    }
    ```
3. Add `image` package to handle different manifests type, referencing an image.
    ```bash  
    manifest/
    ├── image/
    │   ├── docker_v1.go  # Docker V1 / schema1
    │   ├── docker_v2.go  # Docker V2 / OCI image manifest
    │   └── index.go      # OCI index / Docker manifest list
   └── manifest.go   # Manifest Abstract interface & registry
    ```

- [x] Add `ManifestAbstractor` interface and utilities `Register` and `Get`.
- [x] Refactored the `artifact` package with `image` package.
- [ ] Add test files to each manifest abstractors.

---
Please indicate you've done the following:
- [ ] Well Written Title and Summary of the PR
- [ ] Label the PR as needed. "release-note/ignore-for-release, release-note/new-feature, release-note/update, release-note/enhancement, release-note/community, release-note/breaking-change, release-note/docs, release-note/infra, release-note/deprecation"
- [ ] Accepted the DCO. Commits without the DCO will delay acceptance.
- [ ] Made sure tests are passing and test coverage is added if needed.
- [ ] Considered the docs impact and opened a new docs issue or PR with docs changes if needed in [website repository](https://github.com/goharbor/website).

PR #23647, opened 2026-07-30
Refactor manifest abstraction into a registry

### LINKED ISSUE #22847: Refactor manifest abstraction in Abstractor to use registry pattern
**Is your feature request related to a problem?**

`src/controller/artifact/abstractor.go` handles metadata extraction for different manifest types via a hardcoded `switch` statement in `AbstractMetadata()`. Adding support for a new manifest type (e.g., Docker attestation manifests) requires modifying this method directly, violating the Open/Closed Principle.

Resolving this issue will make it easier to support  [Docker attestation manifests. ](https://github.com/moby/buildkit/blob/master/docs/attestations/slsa-provenance.md)

**Background**
Currently, three manifest formats are handled by private methods on the `abstractor` struct:

- `abstractManifestV1Metadata()` — Docker V1 / schema1
- `abstractManifestV2Metadata()` — Docker V2 / OCI image manifest
- `abstractIndexMetadata()` — OCI index / Docker manifest list

Each has different dependency requirements (e.g., V1 needs `blob.Manager`, index needs `artifact.Manager`, V2 needs neither), yet they all live as methods on the same struct.

**Motivation**

Adding support for Docker attestation manifests as a fourth manifest type. 
It is also likely that additional manifest formats will need to be supported in the future. The current switch-based dispatch makes each addition a modification to core logic rather than an extension.

**Proposed solution**

Harbor already has a well-established pattern for this exact problem: the artifact **processor** registry in `src/controller/artifact/processor/processor.go`. Processors use:

The manifest abstraction layer should adopt the same pattern:

**Benefits**

- **Conformity with existing architecture** — uses the same registry + `init()` pattern already proven by artifact processors, reducing cognitive overhead for contributors
- **Open/Closed** — new manifest types are added by registering a new implementation, not by editing dispatch logic
- **Dependency isolation** — each manifest abstractor carries only the dependencies it needs (V1 needs `blob.Manager`, index needs `artifact.Manager`) instead of a single struct holding all of them
- **Testability** — each manifest abstractor becomes independently unit-testable behind an interface

**Files affected**

- `src/controller/artifact/abstractor.go` — replace switch with registry lookup
- New files for registry definition and per-type manifest abstractors (within the same package or a sub-package, depending on maintainer preference)
- `src/controller/artifact/controller.go` — add blank imports if sub-packages are used

**Additional context**

This is a pure refactor of internal dispatch logic. The `Abstractor` interface exposed to the rest of the codebase (`AbstractMetadata(ctx, artifact) error`) remains unchanged. The processor delegation on the last line of `AbstractMetadata()` is also unaffected.

Vad1mo:
Fixes #22847

`AbstractMetadata()` dispatched on manifest media type through a hardcoded `switch`, with the three abstraction bodies as private methods on `abstractor`. Every new manifest format meant editing core dispatch logic, and the struct carried the union of every dependency: `blob.Manager` for schema1, `artifact.Manager` for indexes, neither for v2.

This moves the three formats behind an interface in a new `controller/artifact/manifest` package, keyed by media type, using the registry pattern the artifact processors already use.

```
manifest/
├── manifest.go  # Abstractor interface, Registry, Register, Get
├── v1.go        # docker manifest v1 / schema1     (blob.Manager)
├── v2.go        # OCI image manifest / docker v2   (no deps)
└── index.go     # OCI index / docker manifest list (artifact.Manager)
```

No behaviour change: the same media types are accepted and the same metadata is produced.

## What this opens up

- **New manifest formats become registrations, not edits.** Anything needing its own metadata extraction — future OCI additions, artifact/referrers handling, vendor-specific formats — registers an implementation instead of growing the switch and the dependency union on a single struct.
- **Each abstractor is independently testable.** Constructors are exported and take their dependencies, so an abstractor can be driven directly with mocks (see `manifest/abstractor_test.go`) instead of only through the full `AbstractMetadata` path.
- **Downstream can extend without patching core.** Distributions and forks can register their own abstractor rather than carrying a permanent diff on `abstractor.go`.

## Design notes

- `Get` returns an error instead of falling back to a default. `processor.Get` defaults, which is right for artifact types; guessing at an unknown manifest format would corrupt the artifact model. The error now carries `errors.UNSUPPORTED` rather than being untyped.
- Flat package, no sub-packages. `controller/artifact` imports `manifest` by name, so the `init()` registrations cannot be silently missed the way a forgotten blank import would leave the registry empty.
- `schema1.MediaTypeManifest` is left unregistered to match the current switch exactly. It is accepted elsewhere in the tree (`pkg/registry/client.go`, `replication/transfer/image/transfer.go`), so it may be a real gap — but that is a behaviour change and belongs in its own PR.

Supersedes #22862, with thanks to @vg006.

## Testing

`go test ./src/controller/artifact/...` green; the existing `abstractorTestSuite` is unchanged in behaviour and now populates `manifest.Registry` with mock-backed abstractors, mirroring how it already replaces `processor.Registry`. New tests cover the registry, each abstractor, and the bootstrap itself (an empty registry would break every push). 94.9% statement coverage on the new package. `golangci-lint`, `gofmt` and `go vet` clean.

---

### DIFF
--- a/src/controller/artifact/manifest/index.go
+++ b/src/controller/artifact/manifest/index.go
@@ -1,0 +1,98 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package manifest
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/docker/distribution/manifest/manifestlist"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+
+	"github.com/goharbor/harbor/src/lib/log"
+	"github.com/goharbor/harbor/src/pkg"
+	"github.com/goharbor/harbor/src/pkg/artifact"
+)
+
+func init() {
+	mediaTypes := []string{v1.MediaTypeImageIndex, manifestlist.MediaTypeManifestList}
+	if err := Register(NewIndex(pkg.ArtifactMgr), mediaTypes...); err != nil {
+		log.Errorf("failed to register the abstractor for manifest media types %v: %v", mediaTypes, err)
+	}
+}
+
+// indexAbstractor abstracts artifacts enveloped by OCI index or docker manifest list.
+type indexAbstractor struct {
+	artMgr artifact.Manager
+}
+
+// NewIndex returns an abstractor for OCI indexes and docker manifest lists.
+func NewIndex(artMgr artifact.Manager) Abstractor {
+	return &indexAbstractor{artMgr: artMgr}
+}
+
+func (a *indexAbstractor) Abstract(ctx context.Context, art *artifact.Artifact, content []byte) error {
+	// the identity of index is still in progress, we use the manifest mediaType
+	// as the media type of artifact
+	art.MediaType = art.ManifestMediaType
+
+	index := &v1.Index{}
+	if err := json.Unmarshal(content, index); err != nil {
+		return err
+	}
+
+	/*
+		https://github.com/opencontainers/distribution-spec/blob/v1.1.0/spec.md#listing-referrers
+		For referrers list, If the artifactType is empty or missing in an index, the artifactType MUST be omitted.
+	*/
+	if index.ArtifactType != "" {
+		art.ArtifactType = index.ArtifactType
+	} else {
+		art.ArtifactType = ""
+	}
+
+	// set annotations
+	art.Annotations = index.Annotations
+
+	art.Size += int64(len(content))
+	// populate the referenced artifacts
+	for _, mani := range index.Manifests {
+		digest := mani.Digest.String()
+		// make sure the child artifact exist
+		ar, err := a.artMgr.GetByDigest(ctx, art.RepositoryName, digest)
+		if err != nil {
+			return err
+		}
+		art.Size += ar.Size
+		art.References = append(art.References, &artifact.Reference{
+			ChildID:     ar.ID,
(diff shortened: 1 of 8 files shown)


src/controller/artifact/manifest/index.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/goharbor/harbor/src/lib/log"
+	"github.com/goharbor/harbor/src/pkg"
+	"github.com/goharbor/harbor/src/pkg/artifact"
+)
+
+func init() {
+	mediaTypes := []string{v1.MediaTypeImageIndex, manifestlist.MediaTypeManifestList}
+	if err := Register(NewIndex(pkg.ArtifactMgr), mediaTypes...); err != nil {
+		log.Errorf("failed to register the abstractor for manifest media types %v: %v", mediaTypes, err)
```
wy65701436: log.fatalf?

If Register fails here we only log the error and continue, which means the media types stay unregistered and every push/pull of that manifest type fails later at runtime with "unsupported manifest media type" — pretty far from the real cause.

Vad1mo: Agreed on the failure mode, fixed in d3df3f8 — with `panic` rather than `log.Fatalf`, for two reasons specific to this call site:

- `init()` runs before `main()` configures logging, so a `log.Fatalf` here goes out with default settings and then `os.Exit(1)`s without any trace of *which* registration failed. `panic` keeps the stack pointing at the offending `init`. There is also no `log.Fatal*` inside any `init()` in `src/` today; all 15 existing calls are in `main()` after logging is set up.
- The only reachable failure is a duplicate registration introduced by a future edit (the media-type lists are literals), which is a programming error rather than an operational condition.

Either way the instance now refuses to start instead of 500ing every push of that media type at runtime. The `Register` doc comment states the contract.

Worth noting the same log-and-continue pattern exists at all five `processor.Register` sites (`image/index.go:36`, `image/manifest_v2.go:45`, `image/manifest_v1.go:31`, `wasm/wasm.go:53`, `sbom/sbom.go:41`). It is less dangerous there because `processor.Get` falls back to the default processor, so I left them alone to keep this PR small — happy to fix those in a follow-up if wanted.

Vad1mo: addressed thank you

chlins (approved):
lgtm

wy65701436 (approved):
lgtm

PR #23649, opened 2026-08-01
feat: add Hugging Face model import

### LINKED ISSUE #296: Proposal: Hugging Face Model Import
## Summary

Let users import a model from Hugging Face into a Harbor project as an OCI
model-spec (CNAI) artifact, with optional scheduled re-sync. You give a HF repo
slug (plus optional revision, subfolder, and token); Harbor resolves the model,
converts its files into a model-spec artifact, and stores it in a target
repository — so Harbor can be one governed registry for both images and models.

## Background

Harbor already understands OCI model artifacts via the CNAI/model-spec processor
(model card, files, license, distinct artifact type). What's missing is a
supported on-ramp: getting a model from Hugging Face — the de facto hub for open
models — into Harbor today means an out-of-band "download, convert, push" loop.

That friction hurts three groups who want Harbor in this path: teams
standardizing on OCI model artifacts, air-gapped/regulated orgs that must pull
public models into a governed internal registry, and platform teams who'd rather
run one registry than a separate model store. This proposal makes the import a
native Harbor operation.

## Proposal

A project-scoped Model Import capability:

- **Policy** — a new project-scoped resource: HF repo ID, revision (default
  `main`), optional subfolder, target repo/tag, optional token, and a trigger
  that is manual or scheduled (cron).
- **Import job** — a new `MODEL_IMPORT` jobservice job resolves the model via the
  HF API, streams each file, and assembles a model-spec OCI artifact (config plus
  per-file layers typed as weights/config/code/docs), pushes it, and tags it,
  reporting progress via check-ins.
- **Metadata & sync** — captures model-card metadata (source URL, commit SHA,
  license, etc.) into the config; records the last imported commit SHA so
  repeated/scheduled runs are a no-op when upstream is unchanged.
- **Tokens** — user-supplied HF tokens (for private/gated models) are stored
  encrypted at rest, used only by the job, and never returned by the API.
- **REST API** — endpoints under `/projects/{project_name}/model-import/…` for
  policy CRUD, executions/tasks/logs, and a `validate` preview (resolve a
  repo/revision, preview file count and metadata) before importing. Authorized
  against the existing repository RBAC resource/actions.
- **Repository type** (companion change) — a `type` field (`IMAGE`/`MODEL`) on
  the repository API and portal grid, derived from a repo's artifacts (any
  CNAI/model artifact ⇒ `MODEL`).
- **Portal UI** — a guided "Import Model from Hugging Face" flow.

The source path (resolve → stream → convert → push) sits behind interfaces so
other sources can be added later, but Hugging Face is the only source here.

## Non-Goals

- Model serving/inference, and training/fine-tuning — import and governance only.
- Sources other than Hugging Face (S3, MLflow, etc.) — the interfaces anticipate
  them; none are implemented here.
- Exporting models back to Hugging Face — import is one-way.

## Rationale

Why a dedicated subsystem instead of a Hugging Face replication adapter?
Replication moves existing OCI artifacts registry-to-registry; Hugging Face is a
Git-LFS model hub, not a registry. The core work here is *converting* a model's
files into a model-spec OCI artifact on the fly (typed layers + model-card
metadata) — a concept the adapter contract doesn't have and would be distorted
to add. A separate subsystem keeps model semantics explicit while reusing
Harbor's existing pieces (jobservice, scheduler, registry client, encryption,
repository RBAC) rather than adding parallel infrastructure.

Alternatives considered: a HF replication adapter (rejected — protocol mismatch
above); client-side/CLI tooling (rejected — leaves the governance gap that
motivates air-gapped/regulated users). Whether this should eventually converge
with replication is left as an Open Issue.

## Compatibility

Additive only: a new `model_import_policy` table (forward migration), a new
`MODEL_IMPORT` vendor type, new endpoints, and a new optional repository `type`
field (defaults to `IMAGE`). Nothing existing changes. Imported artifacts are
standard model-spec artifacts handled by the existing CNAI processor, so
retention, quota, deletion, and display behave as usual. Requires jobservice
egress to Hugging Face (or a mirror in air-gapped setups), and builds on the
`modelpack`/model-spec libraries already used by CNAI.

## Implementation

Implemented in an open PR against `goharbor/harbor` (backend under
`pkg/modelimport`, `controller/modelimport`, `jobservice/job/impl/modelimport`;
API/handlers; DB migration; portal UI; unit tests). I (@elliott-davis) will own
it through review and maintain it afterward. If smaller PRs are preferred, it can
split into: (1) import backend + API + migration, (2) scheduling/sync, (3) portal
UI, (4) repository `type`.

## Open Issues

- **Scanning/signing.** Scanners like Trivy don't inspect model weights, and
  provenance/signing for model artifacts is unsettled — what should Harbor
  assert beyond the upstream commit SHA and per-file digests?
- **Convergence with replication.** Should model import eventually become a
  replication source, and if so, what would the adapter contract need to grow?

elliott-davis:
Add a project-scoped capability to import models from Hugging Face into Harbor as OCI model-spec (CNAI) artifacts, so Harbor can serve as a single governed registry for both container images and models.

A new MODEL_IMPORT jobservice job resolves a model via the Hugging Face API, streams its files, and converts them into a model-spec OCI artifact (a model config plus per-file layers typed as weights, config, code, or docs), then pushes and tags it in the target repository. Model-card metadata (source URL, commit SHA, license, and other card fields) is captured into the config so imported models are self-describing.

Imports are driven by project-scoped policies with a manual or scheduled (cron) trigger. Each policy records the last imported commit SHA, so repeated or scheduled runs are a no-op when the upstream revision is unchanged. Access tokens for private or gated models are stored encrypted at rest, used only by the import job, and never returned by the API.

The feature is exposed through new REST endpoints under /projects/{project_name}/model-import (policy CRUD, executions, tasks, logs, and a source-validation preview), authorized against the existing repository RBAC resource, plus a guided import flow in the portal.

Closes: goharbor/community#296

https://github.com/user-attachments/assets/df625517-66c0-47f3-9e98-85343c22db13


Vad1mo:
IMO this is the wrong approach; 

The solution should leverage the existing Harbor capabilities of **proxying** or **replication**, not a slap-on thing that only exists for one thing.

elliott-davis:
**TL;DR** Completely valid concern and I expected some pushback when I pushed this through. I always like to lead with a PR to show the art of the possible and get the conversation going. Thank you @Vad1mo for the quick feedback!

---- Longer Discussion -----

Fair pushback, let me address proxying and replication separately since they fail for different reasons.

Replication copies existing OCI artifacts registry-to-registry. HF isn't an OCI registry for the general case, it's a Git-LFS hub. The actual work here is converting a model's files into a typed model-spec artifact (weights/config/code/docs layers + model-card metadata). The replication adapter contract has no notion of that transform, so an "HF adapter" would have to smuggle conversion logic into a copy-oriented interface.

Proxy cache is the closer fit. HF did recently add an OCI-compatible pull surface (docker model pull hf.co/...), so in principle Harbor could proxy that. Two problems: 

1. that surface only covers GGUF/Docker-model-artifact packaging,it doesn't cover arbitrary repos (safetensors, multi-file weights, tokenizer/config/code, adapters), which is most of what people actually want to govern

2. proxy cache is transparent, lazy, read-through caching subject to retention eviction. It intentionally doesn't give you pinned/immutable copies, commit-SHA-based no-op re-sync, per-policy encrypted tokens, or a pre-import validate step, which are the governance primitives that motivate the air-gapped/regulated use case.

On the "only exists for one thing" concern, that's the part I most want to get right. The source path (resolve -> stream -> convert -> push) already sits behind an interface specifically so S3/MLflow/etc can be added without new infrastructure; HF is just the first implementation. It reuses jobservice, scheduler, the registry client, encryption, and repository RBAC rather than adding anything parallel.

I do think convergence with replication is a legitimate long-term question. If HF's OCI surface broadens to cover general repos, a proxy/replication path gets a lot more viable and I'd be happy to move toward it.

Vad1mo:
>The replication adapter contract has no notion of that transform

Yes, but you can make an adapter that does the conversation on the fly. 

IMO, there are two options going forward.

1. Build this capability as an external tool outside Harbor 
2. Adopt the process to existing Harbor building blocks of proxy and replication. 


For example, see:
- https://github.com/container-registry/helm-charts-oci-proxy

elliott-davis:
Sounds good @Vad1mo! I'll rework this to use replication. 

From a process perspective, would you prefer I close this PR and re-open when I'm done or leave it open for the history?

wy65701436:
> Sounds good @Vad1mo! I'll rework this to use replication.
> 
> From a process perspective, would you prefer I close this PR and re-op

thanks @elliott-davis @Vad1mo 

I've also been thinking about how Harbor should manage Hugging Face and AI model artifacts recently, and I'd like to share a perspective on this.

Instead of a single-model import tool or forcing this into our existing OCI replication adapter, what if we shape this as a Model Sync / Adapter Framework parallel to Harbor’s existing Replication architecture?

We could leverage the same proven architectural pattern: Adapter Interface + Policy/Filter/Trigger + Execution/Task Manager

However, we would define a dedicated Model Adapter Contract tailored for model hubs (Hugging Face, MLflow, S3, etc.), where the fetch → convert → push transformation is treated as a first-class stage in the pipeline.

Why not reuse the existing OCI Replication Adapter? We previously ran a PoC attempting to fit Hugging Face into our standard OCI replication adapter interface. Trying to shoehorn non-OCI model hubs into the OCI distribution v2 API created severe impedance mismatches and technical friction.

Defining a clean, parallel model adapter framework keeps the architecture clean, avoids hacking the OCI registry contract, and provides a extensible foundation for future AI model sources.

FYI, the POC we did last year for the KubeCon: https://www.youtube.com/watch?v=0eiXSogHxmQ

And btw, actually, I and @chlins are writing a proposal for it to the community.

elliott-davis:
@wy65701436 I'm happy to defer to y'all on this. I really put this together to show the art of the possible for our internal VMware team. I would also be happy to collaborate on the proposal as I am certainly an outsider when it comes to contributing to Harbor.

wy65701436:
> @wy65701436 I'm happy to defer to y'all on this. I really put this together to show the art of the possible for our internal VMware team. I would also be happy to collaborate on the proposal as I am certainly an outsider when it comes to contributing to Harbor.

Thanks, @elliott-davis. @chlins  is also a [model spec](https://github.com/modelpack/model-spec) maintainer, focused specifically on AI artifact generation and distribution. We're currently working on a proposal and would value your contribution, whether on the conceptual side or the implementation. If you have any real-world user scenarios you could share, it would help us better understand the usage patterns.

elliott-davis:
I do actually as I developed the oci model packaging format currently used in private ai service for VMware. I'd like to see the efforts converge so I can provide feedback on the gaps in our approach and where it is winning.

If you wouldn't mind linking me to where you're working, if it's in the open, I'd be happy to review and provide feedback. 

In the meantime, should I close this PR?

wy65701436:
@elliott-davis 

If you don't mind, I'll close this PR so we can move the discussion over to the proposal.

I'll share the proposal as soon as it's submitted and CC you as a reviewer. If you'd like, you're also welcome to join as a co-author to add your ideas!

chlins:
Thanks, +1 to moving the detailed design into the proposal.

@elliott-davis, we published this CNCF post back in March to lay out the broader enterprise architecture behind this work: https://www.cncf.io/blog/2026/03/27/the-weight-of-ai-models-why-infrastructure-always-arrives-slowly/

The proposal will build on that direction and turn it into an extensible Harbor design.

It goes beyond model import: Hugging Face acts as the source, models are packaged as standard OCI artifacts, Harbor provides enterprise governance such as immutable versioning, RBAC, retention, signing, audit, and replication, Dragonfly handles P2P distribution and preheating, and OCI Volumes or the Model CSI Driver provide declarative Kubernetes deployment.

In that architecture, your work is definitely not throwaway, it is a valuable PoC for the Hugging Face source adapter in the proposed Model Sync / Adapter Framework. Your VMware Private AI experience would be especially useful for validating authentication, metadata and version semantics, governance, and large-scale distribution. I would really value your feedback and would be glad to collaborate with you on the proposal.

PR #23699, opened 2026-08-10
test: eliminate sleep delays in artifact and health tests

Norway-02:
## What does this PR do?

Reduces arbitrary test delays by eliminating over 10 seconds of `time.Sleep` from the unit test suite.

Specifically:
1. `src/pkg/artifact/dao/dao_test.go`: Removed two 1-second sleeps by manipulating `time.Now()` addition directly to generate strictly increasing timestamps.
2. `src/controller/health/checker_test.go`: Replaced an 8-second sequence of `time.Sleep` with fast polling (`assert.Eventually`), dropping the test duration to under 50ms.

## Why is this necessary?

These changes dramatically speed up CI runs and eliminate potential flakiness on slower or overloaded GitHub Action runners. This aligns with recent successful test stability PRs.


Norway-02:

PR #23700, opened 2026-08-10
test: replace arbitrary sleep delays in execution tests

Norway-02:
## What does this PR do?

Reduces arbitrary test delays by replacing hardcoded `time.Sleep()` calls with deterministic mock polling.

Specifically:
1. `src/controller/replication/execution_test.go`: Replaced four 1-second sleeps with `suite.Eventually` blocks checking `AssertExpectations` on the relevant mocks.
2. `src/pkg/task/execution_test.go`: Replaced a 1-second sleep with a similar `suite.Eventually` block.

A `dummyTestingT` struct is used to safely check expectations inside the polling loop without triggering test failures prematurely.

## Why is this necessary?

Hardcoded `time.Sleep()` delays in tests are inherently flaky and slow down CI pipelines. Polling for the expected condition allows the test to complete exactly when the background work finishes.

### DIFF
--- a/src/controller/replication/execution_test.go
+++ b/src/controller/replication/execution_test.go
@@ -35,6 +35,12 @@ import (
 	testingTask "github.com/goharbor/harbor/src/testing/pkg/task"
 )
 
+type dummyTestingT struct{}
+
+func (d *dummyTestingT) Logf(format string, args ...interface{})   {}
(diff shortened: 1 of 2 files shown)


wy65701436 (approved):
lgtm

Norway-02:

PR #23676, opened 2026-08-07
fix(ci): add explicit permissions block to workflows

jUDASmILE:
Fix #23682 

Summary:
Add explicit least-privilege permissions blocks to GitHub Actions workflow files to secure pull_request_target, scheduled, and pull request workflows.

Details:
Configured minimal required token permissions for the following workflows:
- CI.yml (contents: read)
- auto_assign_prs.yml (pull-requests: write, contents: read)
- build-package.yml (contents: read)
- codeql-analysis.yml (security-events: write, contents: read)
- conformance_test.yml (contents: read)
- housekeeping-stale-issues-prs.yaml (issues: write, pull-requests: write)
- label_check.yaml (pull-requests: read, contents: read)
- nightly-trivy-scan.yml (contents: read)
- pass-CI.yml (contents: read)

This adheres to the principle of least privilege for GitHub Actions GITHUB_TOKEN permissions.

Test Plan:
- Verify that workflow syntax remains valid.
- Run CI workflows to ensure GITHUB_TOKEN permissions retain sufficient scope for required actions.

PR #23701, opened 2026-08-10
ci: add zizmor security audit gated on High severity, and fix the High findings

Vad1mo:
Follow-up to #23676 / #23694, which added explicit `permissions:` blocks. That
closed out zizmor's `excessive-permissions` class; this closes out the rest of
the High-severity ones and adds CI so they stay closed.

## What's added

**`.github/workflows/zizmor.yml`** — runs [zizmor](https://zizmor.sh) over
`.github/` on pushes and PRs that touch workflow files. It gates on **High
severity only**. Low and informational findings still surface as PR annotations
but do not block, so the check is actionable rather than noisy.

**`.github/zizmor.yml`** — records the single accepted High finding with its
justification, rather than hiding it with an inline suppression comment.

## What's fixed

**All 49 unpinned action references are now pinned to a commit SHA** with a
matching version comment. Every pin stays inside the major version already in
use, so no action behaviour changes here; Dependabot (already configured on
`main`) will propose major bumps separately.

**`publish_release.yml`** — four High-confidence template-injection sinks.
`${{ github.ref_name }}` and `${{ github.actor }}` expanded directly into `run:`
blocks are replaced with the equivalent runner environment variables
(`$GITHUB_REF_NAME`, `$GITHUB_ACTOR`). Same values, no expression expansion into
the shell.

**`build-package.yml`** — `cache-poisoning`. This workflow publishes release
artifacts, and `actions/setup-go` enables the module cache by default, so a
poisoned cache entry could reach published output. The cache is now disabled for
this workflow only; other workflows keep theirs.

## Accepted, not fixed

`auto_assign_prs.yml` uses `pull_request_target`, which zizmor flags as
`dangerous-triggers` (High). It needs that trigger to get a write token for PRs
from forks, and it never checks out or executes PR head content, so the usual
escalation path does not apply. This is recorded in `.github/zizmor.yml` with
that reasoning.

## Verification

```
$ zizmor --min-severity=high .github
No findings to report. Good job!
```

Remaining below the gate, not addressed here: 4 medium (`dependabot-cooldown`),
38 low (`template-injection` on non-attacker-controlled contexts, `artipacked`),
5 informational.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -53,17 +53,17 @@ jobs:
       - cncf-ubuntu-16-64-x86
     timeout-minutes: 100
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
          path: src/github.com/goharbor/harbor
       - name: Set up Go 1.26
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5.6.0
         with:
            go-version: 1.26.4
            cache-dependency-path: src/github.com/goharbor/harbor/src/go.sum
         id: go
       - name: Cache CI Go tools
-        uses: actions/cache@v4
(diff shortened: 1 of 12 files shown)


bupd (changes requested):
@Vad1mo please rebase and resolve conflicts

PR #23704, opened 2026-08-11
fix: bound cache-invalidation retries so they can't outlive a canceled request

### LINKED ISSUE #23703: DELETE /repository leaves a transaction open for the retry.Timeout duration when the client disconnects mid-request

Expected: if a client disconnects during a repository delete, the server-side transaction closes shortly after (or the delete is aborted).

Actual: `DELETE /api/v2.0/projects/{project}/repositories/{repository}` wraps the whole request in one DB transaction (`transaction.Middleware`, commit/rollback only after the handler returns). Deleting a repository walks its artifacts and calls the redis cache managers' `cleanUp()` for each one, which retries a cache-delete via `retry.Retry` with the library default timeout (1 minute), twice per artifact. `retry.Retry` never checks whether the request context is already canceled before retrying — so if the client gives up mid-request, the retries keep running anyway, and the transaction sits open the whole time.

We hit this in production: a repository delete that the client had already given up on kept a transaction open for close to 2 hours, until it exhausted the DB connection pool and the registry stopped responding entirely.

1. Start a delete on a repository with several artifacts.
2. Disconnect the client before the request finishes (close the connection / client-side timeout).
3. Watch `pg_stat_activity` for the harbor-core connection handling that request: it stays `idle in transaction`, running the same last query (the DB write from the delete), for as long as the cache-invalidation retries keep going — up to `retry.Timeout` (default 1 minute) x 2 per artifact.

We reproduced this outside production against real Postgres and the actual `lib/orm`/`lib/retry` code: same request, client disconnects after ~1.5s, transaction stays open ~10.5s (scales with artifact count and retry timeout in the real deployment). Patching the retry closure to check `ctx.Err()` and bail via `retry.Abort` closes it in ~2.2s instead — same disconnect, same timing, only that one check differs.

**Versions:**

- harbor version: main @ ac8ce7983a6f4eddbd24daf855d9182427b94832

**Additional context:**

Root cause, file by file:

- `src/server/middleware/transaction/transaction.go` — `Middleware` wraps the entire handler in one transaction (`orm.WithTransaction`), commit/rollback only after it fully returns. `DELETE /repository` isn't in `dbTxSkippers` (`src/core/middlewares/middlewares.go`), so it's fully exposed.
- `src/lib/retry/retry.go` — `Retry` (line 104) retries until success or its own `Timeout` elapses. It has an early-exit path (`retry.Abort`, checked via `errors.As` at line 150) but nothing calls it unless the closure explicitly does.
- Nine call sites retry cache deletes with no cancellation check at all, across `src/pkg/cached/{artifact,repository,project,project_metadata,manifest}/redis/manager.go`.
- The fix pattern already exists elsewhere in this codebase and works: `src/controller/quota/controller.go:260-263` checks `ctx.Err()` before retrying and calls `retry.Abort(ctx.Err())`.

We have a PR ready applying that same pattern to the 9 cache-manager call sites (via `BaseManager.DeleteCache`), plus tests covering the cancellation path, which none of the existing tests do today. Will link it here once opened.

Separately worth a maintainer opinion: `retry.Retry` has 22 production call sites total; making the check at the library level (auto-abort when the returned error is `context.Canceled`) would cover this class of bug everywhere instead of per call site, but it's a bigger surface to review, so we're proposing it as a follow-up rather than bundling it into the same PR.

sebidchi:
Title: Bound cache-invalidation retries so they can't hold a request-scoped DB transaction open indefinitely

Fixes #23703

## What's wrong

`DELETE /repository` wraps the whole HTTP request in one DB transaction (`transaction.Middleware`, commit/rollback only after the handler returns). Deleting a repository walks its artifacts and calls each cache manager's `cleanUp()`, which retries a cache-delete via `retry.Retry` with the library default: 1 minute, twice per artifact. `retry.Retry` never checks whether the request context is already canceled before retrying, so if the client disconnects mid-request, the retries keep going anyway and the transaction sits open for however long they take.

We saw this in production: a repository delete the client had already given up on kept a transaction open for close to two hours, until the DB connection pool ran out and the registry stopped responding for everyone.

## What this does

`BaseManager.DeleteCache` (`src/pkg/cached/base_manager.go`) already existed but had no callers and no retry at all. Hardened it instead of touching each call site individually:

```go
func (bm *BaseManager) DeleteCache(ctx context.Context, key string) error {
	return retry.Retry(func() error {
		if ctx.Err() != nil {
			return retry.Abort(ctx.Err())
		}
		return bm.CacheClient(ctx).Delete(ctx, key)
	}, retry.Timeout(cacheDeleteRetryTimeout))
}
```

Same pattern already proven elsewhere in this codebase: `src/controller/quota/controller.go:260-263`.

`cacheDeleteRetryTimeout` is 3s, down from the 1-minute library default. Cache-invalidation failures here are already non-fatal (logged and swallowed by every caller), so there's no reason a best-effort call should hold a request-scoped transaction open for a full minute per attempt.

The 9 call sites across `artifact`, `repository`, `project`, `project_metadata`, and `manifest` cache managers now call `m.DeleteCache(ctx, key)` instead of hand-rolling their own `retry.Retry(...)` closure. No behavior change for the normal case — `ctx.Err()` is nil until something's actually wrong.

## Tests

None of the existing tests exercised the retry/cancellation path — every one stubs the cache mock to succeed on the first call. Added:

- `base_manager_test.go`: cancellation aborts immediately without calling the cache client, a transient failure still retries and succeeds, and the timeout is actually bounded (overridable in tests).
- One cancellation test per manager package, through the public `Delete`/`Update` methods, asserting the cache client's `Delete` is never called once the context is already canceled.

`go test ./pkg/cached/...`: 73 passed. `go vet` and `golint` clean on every touched file (pre-existing lint debt in these packages is untouched, none of it on lines this PR changes).

## Alternative considered

Instead of fixing this per call site, `retry.Retry` itself (`src/lib/retry/retry.go`) could auto-abort whenever the closure's error is `context.Canceled`, which would cover all 22 production call sites instead of just these 9 — at least 6 others (`retention`, `robot`, `replication`, `scan`, `systemartifact`, and the user handler) take a context-bound closure with the same no-cancellation-check shape. Didn't do that here to keep this PR's review surface scoped to the incident; happy to follow up if maintainers would rather fix it at the library level.

## Checklist

- [x] Tests added for the cancellation path
- [x] `go vet` / `golint` clean on touched files
- [ ] DCO sign-off on commits
- [ ] release-note label


sebidchi:
@stonezdj hey there, any news on this pr ? Thanks in advance

sebidchi:
hey @Vad1mo @stonezdj @chlins @MinerYang , can you give us an update here? This is important , thanks in advance

PR #23707, opened 2026-08-11
test: add API test for signature inheritance on OCI index children

Aloui-Ikram:
Covers the inherited_accessories field added in #23540. The test signs an index with cosign and with notation, then checks that a child of that index reports both signatures as inherited with the index digest as their subject, while owning no accessory of its own and returning nothing extra on a request that does not ask for them.

notation cert generate-test refuses to overwrite an existing key, so generate_cert takes a common name and this test uses one of its own instead of colliding with test_notation_sign_artifact.py in the same suite.

# Issue being fixed


wy65701436:
@Aloui-Ikram can you fix the CI failure? I tried to rerun, but cannot have it success.

Aloui-Ikram:
> @Aloui-Ikram can you fix the CI failure? I tried to rerun, but cannot have it success.

@wy65701436  i dont think it comes from this PR , the new case passes in that run : 
image
The failing one is Test Case - Replication From Dockerhub. Harbor gets a 403 from Docker Hub, "pagination offset too large for anonymous requests", because the run passes DOCKER_USER and DOCKER_PWD empty. No replication job succeeds and the test times out, so a rerun will not help , this PR only touches test files, nothing in the replication path.

Aloui-Ikram:
One more thing, it looks like a real adapter bug and not just a CI setup issue. `FetchArtifacts` resolves the name filter to a namespace and then lists every repo in it (`adapter.go:274` calls `getRepos` with an empty name). The `library` namespace has 180 repos, so page 2 is always requested and Docker Hub returns the 403 to anonymous clients.

The listing endpoint already accepts a name and `listReposPath` already builds it, it is just never passed. Asking for the repo directly stays on one page:

```
/v2/repositories/library/?name=alpine&page=1&page_size=100  ->  200, count 1
```

Happy to send that as a separate PR. Passing `DOCKER_HUB_USERNAME` and `DOCKER_HUB_PASSWORD` to `api_run.sh` would also work, but that one needs your side.

Vad1mo (approved):

wy65701436 (approved):
lgtm

bupd (approved):
/lgtm

PR #23722, opened 2026-08-13
test: reduce arbitrary sleep delays in core auth and awsecr tests

Norway-02:
## What does this PR do?

Reduces arbitrary test delays to significantly accelerate test execution time.

Specifically:
1. `src/core/auth/auth_test.go`: Reduced the mock lock expiration time from `2 * time.Second` to `100 * time.Millisecond` and replaced the hardcoded sleep with `assert.Eventually()` polling.
2. `src/pkg/reg/adapter/awsecr/adapter_test.go`: Reduced a 1-second sleep to `150 * time.Millisecond`. The mock explicitly sets the token expiration to 100ms in the future, so 150ms is more than enough to safely test the token refresh logic.

## Why is this necessary?

Arbitrary `time.Sleep()` delays in tests inflate CI pipelines and introduce flakiness. These precise reductions shave off almost 3 full seconds of execution time while preserving the original testing intent.


Norway-02:
@wy65701436 @chlins Could you please review this quick test cleanup when you have a moment? It removes nearly 3 seconds of arbitrary sleep delays, drastically speeding up execution time.

Norway-02:

wy65701436 (approved):
lgtm

chlins (approved):
lgtm

PR #23773, opened 2026-08-22
Fix CPU saturation from full table scans during artifact deletion by indexing sbom_digest

### LINKED ISSUE #23772: Artifact deletion runs an unindexable report::jsonb @> DELETE on sbom_report, causing DB CPU saturation during retention runs
**Expected:** 
Deleting an artifact (e.g., via tag retention) removes its associated sbom_report rows with an indexed lookup, at a cost independent of table size.

**Actual:** 
When an SBOM accessory artifact is deleted, the cleanup handler runs:

`delete from sbom_report where mime_type = $1 and report::jsonb @> $2`

from src/pkg/scan/sbom/dao/dao.go#L122 (https://github.com/goharbor/harbor/blob/v2.15.0/src/pkg/scan/sbom/dao/dao.go#L122) (DeleteByExtraAttr), invoked by the artifact-deletion event handler at src/controller/event/handler/internal/artifact.go#L334 (https://github.com/goharbor/harbor/blob/v2.15.0/src/controller/event/handler/internal/artifact.go#L334).

The report column is plain JSON with no index (0140_2.11.0_schema.up.sql (https://github.com/goharbor/harbor/blob/v2.15.0/make/migrations/postgresql/0140_2.11.0_schema.up.sql)), so the ::jsonb cast + @> containment predicate cannot use any index: every artifact deletion performs a full sequential scan of sbom_report. Tag retention deletes many artifacts in parallel, so retention runs produce N concurrent sequential scans, and the cost grows linearly as sbom_report grows (which it does on every scanned artifact). The same pattern exists for scan_report in src/pkg/scan/dao/scan/report.go#L122 (https://github.com/goharbor/harbor/blob/v2.15.0/src/pkg/scan/dao/scan/report.go#L122).

**Observed impact (production)**

- Harbor v2.15.0, ~170 projects, SBOM generation enabled, nightly per-project tag retention staggered across a 4-hour window. Database: Aurora PostgreSQL 16.6, 8 vCPU.
- `sbom_report` currently holds **101,871 rows / 187 MB** (`scan_report`, which has the identical delete pattern in `src/pkg/scan/dao/scan/report.go`, holds 94,299 rows / 108 MB).
- The only indexes on `sbom_report` are the primary key, the `uuid` unique index, and the `(artifact_id, registration_uuid, mime_type, media_type)` unique constraint;  none can serve the `report::jsonb @>` predicate, so each delete scans and jsonb-casts the full table.
- During retention runs, RDS Performance Insights shows DB load bursts of **90–128 average active sessions per minute (peak 315)** on an instance whose normal load is 1–2. Top SQL by load during the bursts is the `sbom_report` DELETE above. Wait profile is CPU-dominated, consistent with parallel sequential scans.
- Registry p95 latency (via LB metrics) degraded ~6× (180 ms → 1.2 s) during each burst.
- Burst magnitude roughly **tripled over two weeks** (peak 117 → 315) as `sbom_report` grew — the degradation is progressive with adoption.

**Steps to reproduce**

1. Enable SBOM generation and scan a large number of artifacts so sbom_report contains many rows.
2. Run tag retention (or any bulk deletion) on projects whose artifacts have SBOM accessories.
3. EXPLAIN the delete, or watch DB load: each deletion seq-scans sbom_report.

**Suggested fixes**

Any of these removes the O(table-size) cost per deletion:

1. Store sbom_digest as a real column on sbom_report with a btree index, and delete by equality — it's the only attribute this code path queries. Cleanest long-term.
2. Convert report to JSONB and add a GIN index (jsonb_path_ops), as previously proposed for task.extra_attrs in #20505 (closed as stale) that proposal measured 45 s → 2 s query improvements from the same change.
3. Minimal: expression index CREATE INDEX ... ON sbom_report USING gin ((report::jsonb) jsonb_path_ops) without changing the column type.

The same consideration applies to scan_report.DeleteByExtraAttr.

**Versions**

  - harbor: v2.15.0 (code unchanged on main as of 2026-08-20)
  - database: PostgreSQL 16.6 (Aurora)
  - installer: helm

raudhra:
Problem

During artifact deletion, the `sbom_report` and `scan_report` DAOs used a JSONB containment predicate (`report::jsonb @> '{"sbom_digest":"..."}'`) on a plain JSON column without a supporting index. This forced a full sequential scan and JSONB cast on every deletion, causing severe database CPU saturation during parallel tag retention operations.

Fix

Add an indexed `sbom_digest` column to both the `sbom_report` and `scan_report` tables. The DAO `DeleteByExtraAttr` methods were updated to use a simple equality check against this new indexed column instead of the expensive JSONB containment predicate, converting the operation from a full sequential scan to a fast btree index lookup.

# Issue being fixed
Fixes #23772


make/migrations/postgresql/0200_2.16.1_schema.up.sql
```diff
@@ -0,0 +1,38 @@
+/*
+Fix for issue #23772: Add indexed columns for sbom_digest to avoid
```
wy65701436: since we're in the cycle of 2.16.0, the DB updates should in the 2.16.0 schema.

src/pkg/scan/sbom/dao/dao.go
wy65701436, line 118: Since attrName goes straight into fmt.Sprintf to build the SQL query, it'd be great to add a quick check beforehand:

```
if attrName != "sbom_digest" {
	return errors.Errorf("unsupported attribute: %s", attrName)
}
```

make/migrations/postgresql/0200_2.16.1_schema.up.sql
```diff
@@ -0,0 +1,38 @@
...
+Both tables used: DELETE ... WHERE report::jsonb @> '{"sbom_digest":"..."}' 
+which forced full sequential scans + jsonb casts on every deletion.
+The new indexed columns allow simple equality lookups instead.
+*/
+
+/* --- sbom_report --- */
+ALTER TABLE sbom_report ADD COLUMN IF NOT EXISTS sbom_digest VARCHAR(255);
+
+/* Backfill existing rows from JSON report column */
+UPDATE sbom_report
```
wy65701436: we can actually make this even simpler and guarantee zero data drift by using a Postgres generated column.

```
ALTER TABLE sbom_report 
  ADD COLUMN IF NOT EXISTS sbom_digest text 
  GENERATED ALWAYS AS (report::jsonb ->> 'sbom_digest') STORED;
CREATE INDEX IF NOT EXISTS idx_sbom_report_sbom_digest 
  ON sbom_report (mime_type, sbom_digest);
```

make/migrations/postgresql/0200_2.16.1_schema.up.sql
raudhra: Thanks for the review! Renamed the migration file to align with the 2.16.0 cycle.

make/migrations/postgresql/0200_2.16.1_schema.up.sql
raudhra: Aye cap! Swapped to the generated column approach in the SQL file (and removed the manual Go parsing logic since it's no longer needed).

src/pkg/scan/sbom/dao/dao.go
raudhra: Mhm Noted. Added the explicit attrName == "sbom_digest" security checks in both DAOs to prevent SQL injection.

src/pkg/scan/dao/scan/report.go
```diff
@@ -99,6 +100,14 @@ func (d *dao) UpdateReportData(ctx context.Context, uuid string, report string)
 	data := make(orm.Params)
 	data["report"] = report
 
+	// Extract sbom_digest from the JSON report and keep the indexed column in sync.
+	var parsed map[string]any
+	if err := json.Unmarshal([]byte(report), &parsed); err == nil {
+		if dgst, ok := parsed["sbom_digest"].(string); ok && dgst != "" {
+			data["sbom_digest"] = dgst
```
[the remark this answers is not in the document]
raudhra: Fixed. removed the manual JSON parsing logic here and switched to a Postgres stored generated column (GENERATED ALWAYS AS (report::jsonb ->> 'sbom_digest') STORED) on scan_report. The database now automatically derives and updates sbom_digest on every insert or update.

src/pkg/scan/sbom/dao/dao.go
```diff
@@ -99,6 +100,14 @@ func (d *dao) UpdateReportData(ctx context.Context, uuid string, report string)
 	data := make(orm.Params)
 	data["report"] = report
 
+	// Extract sbom_digest from the JSON report and keep the indexed column in sync.
+	var parsed map[string]any
+	if err := json.Unmarshal([]byte(report), &parsed); err == nil {
+		if dgst, ok := parsed["sbom_digest"].(string); ok && dgst != "" {
+			data["sbom_digest"] = dgst
```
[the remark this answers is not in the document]
raudhra: removed this manual Go-level extraction. The sbom_report table now uses a stored generated column to derive sbom_digest directly from the JSON blob at the database level, guaranteeing zero drift across all write/update paths.

make/migrations/postgresql/0200_2.16.0_schema.up.sql
```diff
@@ -0,0 +1,14 @@
+ALTER TABLE sbom_report
```
wy65701436: generally, can you add comments for mention the why and how?

Vad1mo:
Thanks
the direction (index the digest, delete by equality) is right. 

Three findings:

**1. The generated column breaks‼️ report creation**

Postgres rejects any INSERT that explicitly lists a `GENERATED ALWAYS` column, even with `NULL`:

```
ERROR:  cannot insert a non-DEFAULT value into column "sbom_digest"
DETAIL:  Column "sbom_digest" is a generated column.
```

Both models now map the column (`orm:"column(sbom_digest);null"`), and `dao.Create` uses beego's `o.Insert(r)`, which puts every mapped field into the INSERT column list. Once the migration is applied, every `sbom_report`/`scan_report` insert fails at runtime. Verified on Postgres 16. Beego has no read-only-field tag, so the generated column and the ORM mapping can't coexist.

**2. `ADD COLUMN ... GENERATED ... STORED` rewrites the whole table**

The ALTER takes an ACCESS EXCLUSIVE lock and rewrites the table. `scan_report` stores full vulnerability JSON and can be multi-GB in real deployments, so the upgrade migration becomes long and blocking. It also permanently adds a `report::jsonb` cast of the full report on every row write.

**3. `scan_report.DeleteByExtraAttr` has no caller**

On main, only `sbom.Mgr.DeleteByExtraAttr` is invoked (`src/controller/event/handler/internal/artifact.go`); the `scan_report` variant is referenced only by tests. The `scan_report` part of the migration pays the full cost of finding 2 for a dead code path.

**Minor:** main already has `0190_2.16.0_schema.up.sql`
since 2.16.0 is unreleased, the DDL belongs there rather than in a new `0200` file with a duplicate version. 


## Suggested improvement

An expression index achieves the same btree equality lookup with no schema or model change:

```sql
CREATE INDEX idx_sbom_report_sbom_digest
  ON sbom_report (mime_type, ((report::jsonb ->> 'sbom_digest')));
```

DAO:

```sql
delete from sbom_report where mime_type = ? and report::jsonb ->> 'sbom_digest' = ?
```

- Same lookup performance as the generated column.
- No generated-column/ORM conflict —> model files stay untouched.
- Index build instead of a full table rewrite —> no backfill needed.
- Diff shrinks to the migration plus one line per DAO.

raudhra:
@Vad1mo Thanks for the detailed feedback.

This review was super insightful and it got me learning about how Beego explicitly inserts mapped struct fields and how `GENERATED ALWAYS` columns trigger an `ACCESS EXCLUSIVE` table lock on large datasets. It will help me with designing databases at scale.

I worked on the pr to match your guidance:

1. reverted model changes (so no ORM conflicts)
2.  reverted `scan_report` DAO changes (basically dropping the uncalled path)
3.  Added the expression index DDL to `0190_2.16.0_schema.up.sql`
4.  at last updated the `sbom_report` DAO to query expression directly

make/migrations/postgresql/0200_2.16.0_schema.up.sql
raudhra: Thanks for the feedback! Vadim just pointed out that the generated column approach conflicts with the ORM and causes a full table rewrite, so we pivoted to using an expression index instead. because of that, I've completely removed this 0200 file in the latest push and added the new index to the existing 0190 file.

Vad1mo:
**Alternatives considered, for context**

There are really three depths at which this can be fixed:

1. **Expression index** (suggested above) — fixes the query with zero model change. Smallest diff, no backfill, unblocks #23772 immediately.
2. **Real column, populated by application code** at report-persist time (essentially commit 1 of this PR, plus a backfill migration) — fixes the underlying modeling issue: attributes that are queried or deleted by belong in columns; the JSON should be payload only. `scan_report` already follows this pattern for `critical_cnt`/`high_cnt` etc.
3. **Relational link with FK cascade** from the report row to the accessory artifact (`ON DELETE CASCADE`). Removes the cleanup query and the event-handler cleanup path entirely, and with it the class of bugs it carries (orphaned rows when the handler errors, ordering races between accessory deletion and event handling). The cost is cascade coupling on artifact deletion, which Harbor has generally avoided, so this is a larger discussion than this PR.

Vad1mo:
Benchmarked on Postgres 16 with a 100k-row / 198 MB `sbom_report` replica (#23772 reports 101,871 rows / 187 MB):

| Approach | Plan | Execution time |
|---|---|---|
| Current `report::jsonb @>` | Seq Scan (filters 99,999 rows) | 210 ms |
| Expression index + `->> =` | Index Scan | 0.136 ms |
| This PR (generated column + index) | Index Scan | 0.142 ms |

Harshitaakri:
*(edited — I'd missed that Vadim had already posted single-query numbers when I drafted this)*

Vadim's benchmark above covers the fix shape and single-lookup timings at the production-matching size, so the part of my run that still adds information is the **concurrency dimension** — the incident in #23772 is many deletion events at once, so I ran both predicates under 16 concurrent clients (pgbench, 8 s) on a 2M-row copy of the `0140` schema, Postgres 16.15, 14 cores:

| | single lookup | 16 concurrent clients |
|---|---|---|
| `report::jsonb @> …` | 1,886.9 ms (parallel seq scan) | **3.3 tps, 4.8 s avg latency** |
| expression index + `->>` equality (this PR) | 0.047 ms | **35,747 tps, 0.448 ms avg latency** |

Script + full output: https://gist.github.com/Harshitaakri/d6cc24e0ae3153f2798b0e2b23e184a2 — mine is the bare expression index; the PR's composite `(mime_type, expr)` should behave the same or slightly better for this predicate.

One thing I think is still open: plain `CREATE INDEX` takes a write-blocking lock for the build (4.4 s at 2M rows here; scales with table size). If Harbor's migration runner executes files transactionally, `CONCURRENTLY` isn't available and the current form is the pragmatic choice — might be worth a one-line comment in the migration noting the brief lock at upgrade time.

wy65701436 (approved):
lgtm

raudhra:
Sorry for this trouble, hadn't previously realized that this PR was on my main branch, still getting used to the workflow and how everything moves around open source. I have rebased all my old commits back to the main and dropped the force push which was upstream sync.

 Will make sure next PRs are on different branches rather than my main. 
 
 Thank you for understanding.

chlins (approved):
lgtm

beanssoft:
Hello @Vad1mo thanks for taking this forward. As this is going to be available in the next Harbor release; is there any temporal fix I can apply to mitigate this issue on my end? Or do I just need to wait for it?

PR #23787, opened 2026-08-25
fix(session): do not persist an empty session on regenerate

### LINKED ISSUE #23290: OIDC users intermittently logged out: SessionRegenerate stores empty session under concurrent requests (v2.15.1)
**Expected behavior**

An authenticated OIDC user stays logged in while navigating the UI within the session timeout.

**Actual behavior**

OIDC users get logged out intermittently (often within seconds of login). The browser ends up holding a session cookie (`sid`) that points to an **empty session** in Redis, so every subsequent request resolves to an anonymous security context → `401` → the SPA bounces to login, frequently in a loop.

**Root cause analysis**

In `src/core/session/session.go`, `SessionRegenerate` saves an **empty session** when the `oldsid` no longer exists:

```go
if isExist, _ := rp.SessionExist(ctx, oldsid); !isExist {
    err := rp.c.Save(ctx, sid, map[any]any{}, maxlifetime) // empty session persisted
}
```

Under the real Angular UI (many concurrent XHR per page load + background polling), a concurrent request renames `oldsid` before this path runs, so the regenerate stores an **empty** session for the new `sid`, and that new `sid` is sent to the browser via `Set-Cookie`. The user is now anonymous despite a valid-looking cookie.

PR #22882 / #22873 / #22726 only changed the stored empty value (`""` → `map[any]any{}`) and the lifetime; the empty-session-on-race behavior remains in `main`.

**Evidence**

Redis DB0 accumulated **54 empty 23-byte sessions** vs 2 real ~4760-byte ones. The affected user's cookie `sid` existed in Redis with a healthy ~46 min TTL but was 23 bytes (empty). Core logs (`log.level: debug`): the session works ~13 s after callback, then all requests go unauthorized at once; the same endpoint (`/api/v2.0/configurations`) returns `200` then `401` seconds later.

**Environment**

- Harbor v2.15.1 (the relevant code is also present in `main`)
- Auth: OIDC (Keycloak), auto-onboard enabled, external PostgreSQL, embedded `redis-photon`
- Single `harbor-core` replica, behind an external reverse proxy

**Steps to reproduce**

1. Configure OIDC auth mode.
2. Log in and navigate quickly across admin pages that fire many parallel API calls (e.g. project quotas / configuration).
3. Observe intermittent `401`s and empty 23-byte sessions piling up in Redis DB0.

villcabo:
Fixes #23290.

`SessionRegenerate` persists an empty session whenever it thinks `oldsid` is gone:

```go
if isExist, _ := rp.SessionExist(ctx, oldsid); !isExist {
	err := rp.c.Save(ctx, sid, map[any]any{}, maxlifetime)
}
```

beego sends the new sid to the browser through `Set-Cookie` regardless of what this function does (`session.go:322` in beego v2.3.10), so the client is left holding a valid cookie for a session with no `user` in it. Every request after that resolves to an anonymous security context, returns 401, and the UI bounces to the login page. On our registry, Redis DB0 accumulated 54 of those empty 23-byte sessions against 2 real ones.

There are two ways into that branch and only one of them is a race.

`SessionExist` cannot report failure. It wraps `Cache.Contains`, which returns `false` on any transport error, then hardcodes a nil error on top of it, and the caller discards the error anyway. A timeout, an exhausted pool or a reconnect during that single `EXISTS` is therefore indistinguishable from a missing session, and the code answers by wiping the user's session out. No concurrency needed.

The check isn't atomic either. `SessionExist` and `Rename` are two round trips, so a concurrent request can rename `oldsid` in between. `RENAME` then fails with `no such key` and nothing looks at the error, because `Rename` and `Expire` here are the promoted `*redis.Client` methods and their `Cmd` results get dropped.

This change lets RENAME decide on its own. It's atomic and it already reports `no such key`, so it moves the session and detects the miss in a single round trip, and nothing is written when the miss happens. The non-redis branch gets the same treatment through `len(kv) > 0`.

Dropping that write is safe because regenerate never had to persist anything. Its only caller is `BaseController.PopulateUserSession` (`src/core/api/base.go:162`), which sets the user right afterwards, and beego persists the store at the end of the request through the `defer ctx.Input.CruSession.SessionRelease(...)` in its `router.go`. That path uses `Set`, not `SetXX`, so it writes whether or not the key already exists. `SessionReleaseIfPresent` is part of the beego Store interface, but beego never calls it in the HTTP flow.

## Tests

`TestSessionRegenerate` asserted the old behaviour, that regenerating from a missing session creates the new key, so I inverted that assertion. `TestSessionRegenerateConcurrent` is new: 20 goroutines regenerate the same session, and it checks that exactly one survives and still carries its data.

Against `main` both fail:

```
--- FAIL: TestSession/TestSessionRegenerate
    an empty session should not be persisted
--- FAIL: TestSession/TestSessionRegenerateConcurrent
    expected: 1
    actual  : 20
    exactly one regenerated session should be persisted
```

Twenty concurrent regenerations, twenty persisted sessions, nineteen of them empty. With this change, one. The package passes with `-race`.

I can run this on the instance where we hit the bug if that helps.

### DIFF
--- a/src/core/session/session.go
+++ b/src/core/session/session.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"net/http"
+	"strings"
 	"sync"
 	"time"
 
@@ -182,29 +183,38 @@ func (rp *Provider) SessionRegenerate(ctx context.Context, oldsid, sid string) (
 	}
 	maxlifetime := time.Duration(systemSessionTimeout(ctx, rp.maxlifetime))
 	if rdb, ok := rp.c.(*redis.Cache); ok {
-		// RENAME is atomic and reports "no such key" when oldsid is already gone,
-		// so it both moves the session and detects the miss in a single round trip.
+		// RENAME is atomic and answers "no such key" when oldsid is already gone,
+		// so it moves the session and detects the miss in a single round trip.
+		// Any other failure is a transport or server error and is propagated:
+		// beego skips Set-Cookie when this returns an error, which keeps the
+		// client on its current session instead of handing it a sid we were
+		// unable to populate.
 		if err := rdb.Rename(ctx, oldsid, sid).Err(); err != nil {
-			log.Debugf("failed to rename oldsid=%s to sid=%s, error: %s", oldsid, sid, err)
+			if !isNoSuchKey(err) {
+				return nil, err
+			}
+			log.Debugf("oldsid=%s is gone, not persisting an empty session for sid=%s", oldsid, sid)
 		} else if err := rdb.Expire(ctx, sid, maxlifetime).Err(); err != nil {
 			log.Debugf("failed to set the expiration of sid=%s, error: %s", sid, err)
 		}
 	} else {
+		// Not reachable through SessionInit, which always builds a *redis.Cache,
+		// and Store.releaseSession is a no-op for any other implementation, so a
+		// session could not be persisted here anyway.
 		kv := make(map[any]any)
 		err := rp.c.Fetch(ctx, oldsid, &kv)
-		if err != nil && !errors.Is(err, cache.ErrNotFound) {
-			return nil, err
-		}
-
-		if len(kv) > 0 {
-			err = rp.c.Delete(ctx, oldsid)
-			if err != nil {
+		switch {
+		case err == nil:
+			// A session that exists but carries no values is still a session and
+			// has to be moved, so the fetch error is what decides, not len(kv).
+			if err := rp.c.Delete(ctx, oldsid); err != nil {
 				log.Debugf("failed to delete oldsid=%s, error: %s", oldsid, err)
 			}
-			err = rp.c.Save(ctx, sid, kv, maxlifetime)
-			if err != nil {
+			if err := rp.c.Save(ctx, sid, kv, maxlifetime); err != nil {
 				log.Debugf("failed to save sid=%s, error: %s", sid, err)
 			}
+		case !errors.Is(err, cache.ErrNotFound):
+			return nil, err
 		}
 	}
 
@@ -215,6 +225,13 @@ func (rp *Provider) SessionRegenerate(ctx context.Context, oldsid, sid string) (
 	// returns a usable in-memory store, and beego writes it out on SessionRelease
 	// once the caller has populated it.
 	return rp.SessionRead(ctx, sid)
+}
+
+// isNoSuchKey reports whether err is the reply Redis sends when RENAME is called
+// on a key that no longer exists. go-redis surfaces it as a plain server error,
+// so the message is the only thing to match on.
+func isNoSuchKey(err error) bool {
+	return err != nil && strings.Contains(err.Error(), "no such key")
 }
 
 // SessionDestroy delete redis session by id
--- a/src/core/session/session.go
+++ b/src/core/session/session.go
@@ -181,23 +181,22 @@ func (rp *Provider) SessionRegenerate(ctx context.Context, oldsid, sid string) (
 		ctx = context.TODO()
 	}
 	maxlifetime := time.Duration(systemSessionTimeout(ctx, rp.maxlifetime))
-	if isExist, _ := rp.SessionExist(ctx, oldsid); !isExist {
-		err := rp.c.Save(ctx, sid, map[any]any{}, maxlifetime)
-		if err != nil {
-			log.Debugf("failed to save sid=%s, where oldsid=%s, error: %s", sid, oldsid, err)
+	if rdb, ok := rp.c.(*redis.Cache); ok {
+		// RENAME is atomic and reports "no such key" when oldsid is already gone,
+		// so it both moves the session and detects the miss in a single round trip.
+		if err := rdb.Rename(ctx, oldsid, sid).Err(); err != nil {
+			log.Debugf("failed to rename oldsid=%s to sid=%s, error: %s", oldsid, sid, err)
+		} else if err := rdb.Expire(ctx, sid, maxlifetime).Err(); err != nil {
+			log.Debugf("failed to set the expiration of sid=%s, error: %s", sid, err)
 		}
 	} else {
-		if rdb, ok := rp.c.(*redis.Cache); ok {
-			// redis has rename command
-			rdb.Rename(ctx, oldsid, sid)
-			rdb.Expire(ctx, sid, maxlifetime)
-		} else {
-			kv := make(map[any]any)
-			err := rp.c.Fetch(ctx, oldsid, &kv)
-			if err != nil && !errors.Is(err, cache.ErrNotFound) {
-				return nil, err
-			}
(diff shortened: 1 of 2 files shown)


src/core/session/session.go
```diff
@@ -181,23 +181,22 @@ func (rp *Provider) SessionRegenerate(ctx context.Context, oldsid, sid string) (
...
-			if err != nil && !errors.Is(err, cache.ErrNotFound) {
-				return nil, err
-			}
+		kv := make(map[any]any)
+		err := rp.c.Fetch(ctx, oldsid, &kv)
+		if err != nil && !errors.Is(err, cache.ErrNotFound) {
+			return nil, err
+		}
 
+		if len(kv) > 0 {
```
raudhra: This fallback path for non redis caches does fetch, then delete, then save as three separate calls with nothing tying them together atomically.
Another request can sneak in between them, the exact same kind of bug this PR just fixed for Redis.

If it is reachable, something like this using a lock might help:

```
rp.locks.Lock(oldsid)
defer rp.locks.Unlock(oldsid)
kv := make(map[any]any)
err := rp.c.Fetch(ctx, oldsid, &kv)
```

If it's not reachable in practice, even just a comment saying why would be great so the next reader doesn't wonder the same thing.

src/core/session/session_test.go
```diff
@@ -97,9 +99,69 @@ func (s *sessionTestSuite) TestSessionRegenerate() {
...
+	store, err = s.provider.SessionRegenerate(ctx, "session-001", "session-004")
 	s.NoError(err, "session regenerate should not error")
-	s.True(s.provider.SessionExist(ctx, "session-004"))
+	s.NotNil(store, "the caller should still get a usable store")
+
+	exist, _ := s.provider.SessionExist(ctx, "session-004")
+	s.False(exist, "an empty session should not be persisted")
+}
+
+func (s *sessionTestSuite) TestSessionRegenerateConcurrent() {
```
raudhra, line 115: Question on coverage here,

- TestSessionRegenerateConcurrent only runs against the redis backed provider.
- Given the fallback branch mentioned above might have the same kind of race, this test doesn't currently confirm either way for that path.
- Could be that non redis caches never see concurrent regeneration in real usage, in which case this is fine as is.

If it's worth checking, something like a parametrized version or a second test using a non redis provider would confirm whether the same 19 out of 20 failure shows up there too:

```
func (s *sessionTestSuite) TestSessionRegenerateConcurrentFallbackCache() {
    // same body as TestSessionRegenerateConcurrent, but using a
    // non redis provider to check whether the fetch/delete/save
    // fallback has the same race under concurrent regeneration.
}
```

raudhra (commented):
**Thanks for putting this together.** 
Really solid fix overall for what sounds like a genuinely painful bug to track down in production. Let me know if any of this doesn't hold up, could easily be missing something about how the fallback branch actually gets used.

src/core/session/session.go
```diff
@@ -181,23 +181,22 @@ func (rp *Provider) SessionRegenerate(ctx context.Context, oldsid, sid string) (
...
 	maxlifetime := time.Duration(systemSessionTimeout(ctx, rp.maxlifetime))
-	if isExist, _ := rp.SessionExist(ctx, oldsid); !isExist {
-		err := rp.c.Save(ctx, sid, map[any]any{}, maxlifetime)
-		if err != nil {
-			log.Debugf("failed to save sid=%s, where oldsid=%s, error: %s", sid, oldsid, err)
+	if rdb, ok := rp.c.(*redis.Cache); ok {
+		// RENAME is atomic and reports "no such key" when oldsid is already gone,
+		// so it both moves the session and detects the miss in a single round trip.
+		if err := rdb.Rename(ctx, oldsid, sid).Err(); err != nil {
+			log.Debugf("failed to rename oldsid=%s to sid=%s, error: %s", oldsid, sid, err)
```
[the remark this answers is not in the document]
villcabo: Good catch, this one was real. Two things I checked before changing it: Redis answers `ERR no such key` for RENAME on a missing key, so that case can be told apart from the rest; and `Cache.Fetch` wraps *every* error as `cache.ErrNotFound` (`src/lib/cache/redis/redis.go:65`, and the comment there says so outright), so `SessionRead` hands back an empty store with a nil error when the backend is unreachable. That is the path you described.

Now only "no such key" is swallowed and everything else propagates, which stops beego from issuing the new cookie.

`TestSessionRegenerateTransportError` points a provider at a dead port. Without the change it fails with:

```
Error: An error is expected but got nil
Expected nil, but got: &session.Store{... values:map[interface {}]interface {}{} ...}
```

An empty store, returned successfully, with Redis unreachable.

src/core/session/session.go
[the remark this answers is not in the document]
villcabo: Agreed, `len(kv)` conflates a missing session with an empty one. It now decides on the fetch result instead: `err == nil` deletes and saves, `cache.ErrNotFound` skips persistence, and anything else propagates.

villcabo: You are right that fetch/delete/save is not atomic. I went looking for whether that branch is reachable and it is not, in two independent ways.

`SessionInit` always builds the cache through `redis.New(...)` (`session.go:144`), so `rp.c` is always a `*redis.Cache`. And even if it were not, `Store.releaseSession` only writes when the cache is a `*redis.Cache` and does nothing otherwise, so a session could never be persisted through a non-redis cache to begin with.

A lock there would be maintaining code that cannot work, so I took the comment option you offered. The branch did pick up the fetch-result fix from the other review thread.

If you would rather see it deleted outright, or locked anyway, I will do either. Just say which.

src/core/session/session_test.go
villcabo: Same root as the other thread: the fallback cannot be reached through `SessionInit`, and `releaseSession` is a no-op for non-redis caches, so a parametrized run would be covering a path that cannot persist a session in the first place. I would rather not lock the shape of dead code in with a test.

It is doable if you want it anyway. The test lives in `package session`, so it can build `&Provider{c: memCache}` against `src/lib/cache/memory` directly. Say the word and I will add it.

src/core/session/session.go
raudhra: Makes sense. Thanks for checking it out.
Comment explains why this branch is unreachable, a lock would just be protecting code that can never run.
Good to resolve as is.

src/core/session/session_test.go
raudhra: Reasoning holds, same as before, no value testing a path that can't be exercised by real traffic. Not necessary on my end, good to resolve unless a maintainer wants it added for documentation.

villcabo:
All review threads are addressed, and @raudhra marked theirs as good to resolve.

The three workflows (CI, Code scanning, Release Note Label Check) are still waiting for a maintainer to approve the run, since this is my first contribution here. Could someone approve them and add a release-note label? cc @chlins

Happy to rebase or adjust anything once CI has had a pass.

Issue #21075, opened 2024-10-21
Quota update performs multiple UPDATEs with the same value, blocking db operations


ianseyer:
https://github.com/goharbor/harbor/blob/c14c3dbf9e04797da83f94cade863df427a649ab/src/pkg/quota/dao/dao.go#L192

Harbor v2.10.3

This setup results in multiple queries being executed from multiple clients with the same UPDATE on the same resource, often with the same value (would be a problem if multiple clients sent conflicting information). This becomes a blocking query, which eventually get purged by our internal lock killer.

This update call must be performed in serial, but is attempted to be done in parallel when running multiple instances of harbor.

While running a load test against a harbor instance (deployed with helm, 3 replicas per deployment):
```
2024-10-21 19:22:00.753 : $1 = '{""storage"": 154851}', $2 = '2024-10-21 19:22:00', $3 = '381', $4 = '1036', $5 = '380'",,,,,,,,"app - [<redacted>]:54250","client backend",,0
2024-10-21 19:22:00.759 : $1 = '{""storage"": 154851}', $2 = '2024-10-21 19:22:00', $3 = '381', $4 = '1036', $5 = '380'",,,,,,,,"app - [<redacted>]:48234","client backend",,0
2024-10-21 19:22:01.101 : $1 = '{""storage"": 154851}', $2 = '2024-10-21 19:22:01', $3 = '381', $4 = '1036', $5 = '380'",,,,,,,,"app - [<redacted>]:54216","client backend",,0
2024-10-21 19:22:01.835 : $1 = '{""storage"": 154851}', $2 = '2024-10-21 19:22:01', $3 = '381', $4 = '1036', $5 = '380'",,,,,,,,"app - [<redacted>]:41568","client backend",,0
2024-10-21 19:22:02.023 : $1 = '{""storage"": 154851}', $2 = '2024-10-21 19:22:02', $3 = '381', $4 = '1036', $5 = '380'",,,,,,,,"app - [<redacted>]:34674","client backend",,0
2024-10-21 19:22:02.392 : $1 = '{""storage"": 154851}', $2 = '2024-10-21 19:22:02', $3 = '381', $4 = '1036', $5 = '380'",,,,,,,,"app - [<redacted>]:48204","client backend",,0
[...]

PR #18871, opened 2023-06-30
perf: introduce update quota by redis

chlins:
Introduce the quota update provider, improve the performance of pushing artifacts to same project with high concurrency by implementing optimistic lock in redis. By default the function is disabled, open it by set env 'QUOTA_UPDATE_PROVIDER=Redis' for the core container.

Refer to the [wiki](https://github.com/goharbor/perf/wiki/Quota-Update) for more details.

Fixes: #18440

Thank you for contributing to Harbor!

# Comprehensive Summary of your change

# Issue being fixed
Fixes #18440 

Please indicate you've done the following:
- [x] Well Written Title and Summary of the PR
- [x] Label the PR as needed. "release-note/ignore-for-release, release-note/new-feature, release-note/update, release-note/enhancement, release-note/community, release-note/breaking-change, release-note/docs, release-note/infra, release-note/deprecation"
- [x] Accepted the DCO. Commits without the DCO will delay acceptance.
- [x] Made sure tests are passing and test coverage is added if needed.
- [ ] Considered the docs impact and opened a new docs issue or PR with docs changes if needed in [website repository](https://github.com/goharbor/website).

Issue #18440, opened 2023-03-28
when push image concurrently, updating the quota will be stuck for a few seconds


hcwnbs:
**Steps to reproduce the problem:**
push 20 image concurrently

- harbor version: 2.6.0
- docker engine version: 18.09

**Expected behavior and actual behavior:**
A clear and concise description of what you expected to happen and what's the actual behavior. If applicable, add screenshots to help explain your problem.

**set orm.debug=true**
**I found that when update quota concurrently, harbor need to wait another request transaction to commit.**

**this is the orm log:**
[ORM]2023/03/28 02:48:12  -[Queries/default] - [  OK /     db.Exec /     3.2ms] - [UPDATE quota_usage SET used = $1, update_time = $2, version = $3 WHERE id = $4 AND version = $5] - `{"number":43}`, `2023-03-28 02:48:12`, `43`, `3`, `42`
[ORM]2023/03/28 02:48:20  -[Queries/default] - [  OK /     db.Exec /  8030.1ms] - [UPDATE quota_usage SET used = $1, update_time = $2, version = $3 WHERE id = $4 AND version = $5] - `{"number":43}`, `2023-03-28 02:48:12`, `43`, `3`, `42`
[ORM]2023/03/28 02:48:20  -[Queries/default] - [  OK /     db.Exec /     0.3ms] - [UPDATE quota_usage SET used = $1, update_time = $2, version = $3 WHERE id = $4 AND version = $5] - `{"number":44}`, `2023-03-28 02:48:20`, `44`, `3`, `43`
[ORM]2023/03/28 02:48:24  -[Queries/default] - [  OK /     db.Exec / 11855.8ms] - [UPDATE quota_usage SET used = $1, update_time = $2, version = $3 WHERE id = $4 AND version = $5] - `{"number":43}`, `2023-03-28 02:48:12`, `43`, `3`, `42`
[ORM]2023/03/28 02:48:24  -[Queries/default] - [  OK /     db.Exec /     2.8ms] - [UPDATE quota_usage SET used = $1, update_time = $2, version = $3 WHERE id = $4 AND version = $5] - `{"number":45}`, `2023-03-28 02:48:24`, `45`, `3`, `44`
[...]

PR #23789, opened 2026-08-25
fix(quota): enable retry backoff on quota usage updates

Vad1mo:
### What this PR does

Enables the retry package's default exponential backoff (with jitter) for the optimistic-lock retry loops in the quota controller, which currently run with `retry.Backoff(false)` and therefore retry in a zero-delay tight loop for up to 5 minutes per call.

### Why

`updateUsageWithRetry` and `Update` retry `orm.ErrOptimisticLock` conflicts on the per-project `quota_usage` row. With backoff disabled, every losing writer immediately re-reads and re-CASes the row, so N concurrent pushes to one project convert each conflict into a synchronized retry storm against the database.

Measured on a production Harbor (v2.15.x) during a CI burst of ~1000 blob uploads into one project, AWS RDS Performance Insights attributed **13 average active sessions** (on a 2-vCPU instance) to this single statement:

```
UPDATE quota_usage SET used = $1, update_time = $2, version = $3 WHERE id = $4 AND version = $5
```

with `Lock:tuple` as the dominant wait event — i.e. mostly serialization and retry spin, not useful work. The saturation was severe enough that new DB connections failed their TLS handshake, taking `/service/token` down for every client.

### Relation to prior work

This contention has been reported before: #18440 ("when push image concurrently, updating the quota will be stuck for a few seconds") was resolved by #18871, which introduced the optional Redis quota provider (`QUOTA_UPDATE_PROVIDER=redis`); #21075 reported the same symptom on the DB path afterwards. The Redis provider moves the contention off PostgreSQL, but the **default** DB path still retries with zero backoff — and not every deployment can make Redis authoritative for quota state (e.g. single-replica/ephemeral Redis). This PR is complementary to the Redis provider: it fixes the retry behavior of the default path rather than replacing it.

The retry package already defaults to backoff `{Min: 100ms, Max: 1s, Factor: 2, Jitter: true}`; this PR removes the explicit opt-out. The retention job already passes `retry.Backoff(true)` for the same call path (`src/pkg/retention/job.go`), so this aligns the default with existing practice.

### Behavior change

Contended quota updates now sleep between attempts instead of spinning. Total retry budget (5 min timeout) is unchanged. Uncontended updates are unaffected (first attempt still succeeds immediately).


wy65701436 (approved):
lgtm

chlins (approved):
lgtm

PR #23791, opened 2026-08-25
perf(quota): skip usage reservation when hard limits are unlimited

Vad1mo:
### What this PR does

`Request()` skips the quota usage reservation (and its rollback) when every requested resource has an `UNLIMITED` (-1) hard limit, and instead marks the project for a **deferred, coalesced usage refresh** — a background task on the shared `gtask` pool recomputes and stores the usage once per dirty project per interval. On any error determining the limits it falls back to the reservation path.

**The whole mechanism is opt-in.** It activates only when `QUOTA_ASYNC_REFRESH_DURATION` (seconds; also the flush interval, naming and shape following `ARTIFACT_PULL_ASYNC_FLUSH_DURATION`) is set to a positive value. With the env unset — the default — behavior is unchanged: the reservation still runs for unlimited quotas and stored usage stays synchronously visible, so default deployments and the existing API tests (e.g. `test_project_quota.py`, which reads the usage immediately after a push) are unaffected.

### Why

The reservation CAS-updates the single `quota_usage` row per project, retrying optimistic-lock conflicts. When the hard limit is `UNLIMITED`, `quota.IsSafe` can never deny the request — the reservation enforces nothing, yet it still contends on the hot row.

Measured on a production Harbor (v2.15.x) during a ~1000-blob CI push burst into projects with unlimited quotas, the reservation UPDATE accounted for **13 average active sessions** on a 2-vCPU database (`Lock:tuple` wait) while enforcing nothing.

### Why the deferred refresh is part of this change

On the push path the reservation is also the **only writer of the usage figure**: `RefreshForProjectMiddleware` (the full recompute) is mounted only on artifact/repository DELETE routes. Skipping the reservation alone would freeze the usage figure — our end-to-end test caught exactly that. Hence the skip marks the project dirty and the flusher keeps the figure current: one aggregate query (`CalculateUsage` → `SELECT SUM(size) FROM blob JOIN project_blob ...`) per project per interval instead of 2 CAS writes per push.

The flush **recomputes from the database rather than accumulating deltas**, so it is idempotent: a mark lost to a process restart, or a concurrent flush from another core replica, can only delay convergence — never corrupt the stored value. Failed flushes re-mark and retry.

### Consequences

- With the flag enabled, UI, quota API and exporter metric stay current to within one flush interval (10s default); with it unset there is no change at all.
- If an administrator sets a real limit while requests are in flight, enforcement starts with the next request and the deferred refresh reconciles the usage.
- The unlimited check reads the quota row once — a read the reservation path performed anyway.

Tests: with the flag on, unlimited requests bypass the CAS and mark for deferred refresh; with it off, unlimited requests still reserve; limited requests still reserve; over-limit requests are still denied before the wrapped operation runs; flusher coalescing, one-flush-per-reference, empty-set noop and failed-flush re-mark. An e2e suite exercising a concurrent shared-base-layer push burst against a live instance accompanies the stack.

### DIFF
--- a/src/controller/quota/controller.go
+++ b/src/controller/quota/controller.go
@@ -389,6 +389,31 @@ func (c *controller) Request(ctx context.Context, reference, referenceID string,
 		return f()
 	}
 
+	// When every requested resource is unlimited, the reservation cannot
+	// deny anything: IsSafe always passes for UNLIMITED hard limits, so the
+	// reserve/rollback pair degenerates into contended writes on the single
+	// quota_usage row per project with no enforcement effect. When the
+	// deferred refresh is enabled (opt-in via QUOTA_ASYNC_REFRESH_DURATION),
+	// skip the reservation and let the coalesced flush keep the usage
+	// figure current. Without the env var, behavior is unchanged: the
+	// reservation still runs and the usage stays synchronously visible.
+	// If a real limit is set concurrently, enforcement starts with the
+	// next request and the refresh reconciles the usage.
+	if AsyncRefreshEnabled() {
+		if unlimited, err := c.isUnlimited(ctx, reference, referenceID, resources); err == nil && unlimited {
+			err := f()
+			if err == nil {
+				// the skipped reservation was also the only usage writer
+				// on this path - keep the usage figure current via the
+				// deferred coalesced refresh
+				MarkRefresh(reference, referenceID)
+			}
+			return err
+		} else if err != nil {
+			log.G(ctx).Warningf("failed to check hard limits for %s %s, falling back to reservation, error: %v", reference, referenceID, err)
+		}
+	}
+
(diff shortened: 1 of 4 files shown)


Vad1mo:
The `APITEST_DB` failure on the previous push was real: `test_project_quota.py` reads the project usage immediately after a push, and with the fast path active by default on unlimited-quota projects (the CI default) that read raced the 10s deferred flush.

Rather than weakening the test, the latest push makes the whole mechanism **opt-in**: the unlimited fast path and the deferred flush activate only when `QUOTA_ASYNC_REFRESH_DURATION` is set. With the env unset, quota behavior is byte-for-byte unchanged, which the new `TestRequestUnlimitedReservesWhenAsyncDisabled` pins down.

PR #23798, opened 2026-08-27
Return 404 instead of 500 for a missing retention policy

swasabs:
`GetPolicy()` in `src/pkg/retention/dao/retention.go` returns the raw beego `orm.ErrNoRows` straight through when a retention policy id doesn't exist, instead of wrapping it with `orm.WrapNotFoundError` like the other DAOs do (project, quota, robot, blob, etc. all do this on their `Get`). Because the raw ORM error isn't recognized by the generic API error handler, it falls back to a plain `500 internal server error` instead of a 404.

Ran into this after a project got deleted and recreated - `GET /api/v2.0/retentions/{id}` for the policy that used to belong to the old project came back as a 500 instead of a 404. It also confused tooling built on top of the API (Terraform/OpenTofu's harbor provider treats a 500 here as a hard failure instead of ordinary "this is gone, recreate it" drift).

Small fix: wrap the read the same way every other DAO already does. Updated the existing DAO test's not-found check to assert `errors.IsNotFoundErr(err)` instead of matching the raw ORM error string, matching how the project DAO test checks the same thing.

Test plan:
- `go build ./pkg/retention/...` passes
- `go vet ./pkg/retention/dao/...` passes, gofmt clean
- Didn't run `TestPolicy` itself since it needs a live Postgres instance (`dao.PrepareTestForPostgresSQL()`) I don't have set up locally - happy to if someone points me at the harness, or if CI covers it

### DIFF
--- a/src/pkg/retention/dao/retention.go
+++ b/src/pkg/retention/dao/retention.go
@@ -64,7 +64,7 @@ func GetPolicy(ctx context.Context, id int64) (*models.RetentionPolicy, error) {
 		ID: id,
 	}
 	if err := o.Read(p); err != nil {
-		return nil, err
+		return nil, orm.WrapNotFoundError(err, "retention policy %d not found", id)
 	}
 	return p, nil
 }
--- a/src/pkg/retention/dao/retention_test.go
+++ b/src/pkg/retention/dao/retention_test.go
@@ -3,13 +3,13 @@ package dao
 import (
 	"encoding/json"
 	"os"
-	"strings"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
 
 	"github.com/goharbor/harbor/src/common/dao"
+	"github.com/goharbor/harbor/src/lib/errors"
 	"github.com/goharbor/harbor/src/lib/orm"
 	"github.com/goharbor/harbor/src/lib/q"
 	"github.com/goharbor/harbor/src/pkg/retention/dao/models"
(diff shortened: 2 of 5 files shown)


src/pkg/retention/dao/retention.go
```diff
@@ -64,7 +64,7 @@ func GetPolicy(ctx context.Context, id int64) (*models.RetentionPolicy, error) {
 		ID: id,
 	}
 	if err := o.Read(p); err != nil {
-		return nil, err
+		return nil, orm.WrapNotFoundError(err, "retention policy %d not found", id)
```
raudhra, line 67: Nice fix.
Wanted to ask does `WrapNotFoundError` check that the underlying error is actually `orm.ErrNoRows` before converting it or does it wrap any error unconditionally?
If the latter then a DB/connection failure in `o.Read` would get misreported as `404` instead of `500`.

If it does check then this looks good to me pending the CI passing _TestPolicy_.

swasabs: WrapNotFoundError only converts orm.ErrNoRows; other DB errors pass through unchanged

src/pkg/retention/dao/retention_test.go
```diff
@@ -99,5 +99,5 @@ func TestPolicy(t *testing.T) {
 
 	p1, err = GetPolicy(ctx, id)
 	assert.NotNil(t, err)
-	assert.True(t, strings.Contains(err.Error(), "no row found"))
+	assert.True(t, errors.IsNotFoundErr(err))
```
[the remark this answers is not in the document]
swasabs: Fixed in cd3bf9e9c

PR #22815, opened 2026-02-04
Origin/18124 custom role feature

maxgraustenzel-create:
This change adds custom roles to harbor. 

**Directly Solved Issues:**

#18124 RBAC - be able to create a custom role 
#18143 Create custom project roles 
#21222 Add role for security auditor
#15711 Add a readonly admin role in Harbor 
#12062 Make permission system more fine-grained 
#21306 - Read-only auditor role (December 2024 - very recent!)
#8632 - Limited guest role
#1486 - Guest shouldn't see all members (from 2016!)
...

**Partially Solved Issues:** 

#10159 - Repository-level permissions 
...

**Design changes :**
* Store permissions in role_permission and permission_policy table (from rbac_role.go, migration with init_permissions.sql)
* Permission management functionality for the system administrator (swagger + ui)
* Dynamic permission selection when adding members
* Caching of role permissions for the user and session in rbac_user.go (contains now the roles object as part of session with the associated permissions)

Issue #18124, opened 2023-01-20
RBAC - be able to create a custom role


flbla:
**Is your feature request related to a problem? Please describe.**
We can't create custom role : https://goharbor.io/docs/2.1.0/administration/managing-users/user-permissions-by-role/

**Describe the solution you'd like**
As an admin I would like to be able to create a custom role for user

zyyw:
Unfortunately, Harbor simply does not support creating a custom role so far. Could you please justify why you would like to create a custom role?

flbla:
For eg : I would like to allow some users to be able to only delete tags.
Or some other to be able to only scan images.

sizowie:
+1

dgutierr27:
+1

Custom role use case: You are using Harbor to host images for external customers as a service.  You want them to be able to only download images, not view security or vulnerability details, configurations, etc.

Also, is there a way to disable the web interface so clients only connect and download via CLI?

uspiotrowska:
+1

mr-Fernz:
+1

maxgraustenzel-create:
Hi all, I would like to contribute to the custom role functionality. In our use case we use Harbor registry for a secured Kubernetes platform : 

* Our users need to create robot accounts, web hooks, for their projects... but they are not allowed not modify security and scanning settings. 

* We also have some other roles like security administrators and auditors that require quite specific roles

msartor92:
other example: i need some person in the company to see the interrogation service overview but without having admin role

PR #23804, opened 2026-08-28
feat: custom project roles (hybrid — built-ins in code, custom in DB)

maxgraustenzel-create:
## What / why

Custom project roles, implemented as a **hybrid** that keeps built-in roles exactly as they are today. This is an **alternative to #22815**, following the direction suggested in review (thanks @stonezdj): rather than seeding built-in role permissions into the database, keep resolving built-in roles from the compile-time `rolePoliciesMap` and store **only custom roles** in the DB.

Tracking: #18124
Alternative to: #22815

## Design

- **Built-in roles are unchanged.** `common/rbac/project` still resolves the five built-in project roles from `rolePoliciesMap` — no database access, no cache. The RBAC diff is a small, additive overlay: `projectRBACRole` gains an optional `custom *role.Role`, and the role controller is consulted **only for non-built-in role IDs**. So authorization for built-in roles takes the identical code path as `main`.
- **Custom roles live in the database.** `pkg/role` (model/dao/manager) + `controller/role` provide CRUD and persistence; the existing `role_permission` / `permission_policy` tables store custom permissions. Custom roles are loaded lazily during permission evaluation (once per project per request), so a permission edit takes effect on the next request.
- **No permission cache.** Built-ins are free (code path); custom roles are rare and read per-request. There is no L1/L2 cache to keep coherent across nodes.
- **REST API** (`/roles`, sysadmin-gated) with name validation and an orphan-delete guard; an effective-permissions endpoint; anti-escalation checks when assigning roles to members/robots; and audit events for role create/update/delete.
- **Migration `0191`** adds `is_builtin`/`description`/audit columns to `role`, widens the name, enforces name uniqueness, and marks the seeded roles built-in. **Built-in permissions are not seeded into the DB** — custom role IDs come from the existing `SERIAL` sequence (always ≥ 6), so they never collide with the built-in IDs 1–5.

## Scope

Backend + API only. The portal UI ships separately so this can be reviewed on its own.

## Performance

Built-in-role authorization has **no meaningful cost vs `main`**. Measured on bare metal (500 VUs × 5000 iters/endpoint, snapshot-restored DB, `ANALYZE`, per-endpoint warm-up, order-balanced) as a **non-admin project-admin member** (admin bypasses the RBAC evaluator):

- **Feature vs `main` HEAD:** comparable-endpoint mean **+3.5%**, individual endpoints −5% … +8%, 100% success.
- **A/A control** (same binary as both cells): **−0.1%** — the ±per-endpoint swing above is the harness noise floor, not the feature.

i.e. parity — the feature tracks `main` for built-in roles, as the code predicts.

## Testing

- `go build ./...` and unit tests pass (`common/rbac/project`, `controller/role`, `server/v2.0/handler`).
- Deployed and exercised end-to-end: built-in roles serve with **zero** `role_permission` rows; creating a custom role persists its permissions and authorizes correctly.

Signed-off-by present (DCO).


tests/integration/.gitignore
```diff
@@ -0,0 +1,6 @@
+# local Python virtualenvs used to run the integration tests
```
stonezdj: Please remove the .gitignore file in this directory, you should update it in the top level git ignore file

src/controller/event/handler/init.go
```diff
@@ -67,6 +67,9 @@ func init() {
 	_ = notifier.Subscribe(event.TopicDeleteTag, &auditlog.Handler{})
 	_ = notifier.Subscribe(event.TopicCreateRobot, &auditlog.Handler{})
 	_ = notifier.Subscribe(event.TopicDeleteRobot, &auditlog.Handler{})
+	_ = notifier.Subscribe(event.TopicCreateRole, &auditlog.Handler{})
```
stonezdj, line 70: Because current PR contains too many changes,  for easy review, can you please separate the new audit log event part from the current PR?

src/server/v2.0/handler/role.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+}
+
+type roleAPI struct {
+	BaseAPI
+	roleCtl role.Controller
+}
+
+func (rAPI *roleAPI) CreateRole(ctx context.Context, params operation.CreateRoleParams) middleware.Responder {
+	if err := rAPI.checkSysAdmin(ctx); err != nil {
```
stonezdj: Please define the role as a resource in src/common/rbac/const.go,  like the user resource, it should be a system level resource, and check the current permission by API similar to src/server/2.0/handler/user.go
```
	if err := u.RequireSystemAccess(ctx, rbac.ActionList, rbac.ResourceUser); err != nil {
		return u.SendError(ctx, err)
	}
```

src/server/v2.0/handler/role.go
```diff
@@ -0,0 +1,336 @@
...
+	if err != nil {
+		return rAPI.SendError(ctx, err)
+	}
+
+	return operation.NewGetRoleByIDOK().WithPayload(model.NewRole(r).ToSwagger())
+}
+
+// checkSysAdmin returns UnauthorizedError for unauthenticated callers and
+// ForbiddenError for callers who are not sysadmins.
+func (rAPI *roleAPI) checkSysAdmin(ctx context.Context) error {
```
stonezdj: Please remove this function, replace it with  action+resource checking API. 
```
	if err := u.RequireSystemAccess(ctx, rbac.ActionList, rbac.ResourceUser); err != nil {
		return u.SendError(ctx, err)
	}
```
it will make the permission is managable, athough we want only the system can do that.

api/v2.0/swagger.yaml
```diff
@@ -3578,6 +3679,82 @@ paths:
           $ref: '#/responses/404'
         '500':
           $ref: '#/responses/500'
+  /roles/{role_id}:
```
stonezdj: Please put the /roles/{role_id} api just behind the /roles API, they server to same feature, should put them together.

src/controller/member/controller.go
```diff
@@ -239,7 +239,8 @@ func isValidRole(role int) bool {
 		common.RoleLimitedGuest:
 		return true
 	default:
-		return false
+		// Custom role IDs are assigned sequentially above the built-in range
```
stonezdj: isValidRole didn't have any validation function, for any integer > 0 it will return true. please consider how to verify the role with ID. it should exist in the database.

src/common/rbac/const.go
```diff
@@ -311,5 +316,102 @@ var (
 
 			{Resource: ResourceQuota, Action: ActionRead},
 		},
+		ScopeRole: {
+			{Resource: ResourceSelf, Action: ActionRead},
```
stonezdj: I don't think the ResourceSelf should be include.

src/controller/member/controller.go
stonezdj: The error message should be updated, role id is not only limited to 1,2,3

stonezdj, line 113: The role should be validate before Update too.

api/v2.0/swagger.yaml
```diff
@@ -526,6 +526,34 @@ paths:
           $ref: '#/responses/404'
         '500':
           $ref: '#/responses/500'
+  '/projects/{project_name_or_id}/permissions/effective':
```
stonezdj: This API is duplicated with the /api/v2.0/users/current/permissions, if there is any difference, you can extend the parameter of old API to meet your requirement

src/server/v2.0/handler/role.go
maxgraustenzel-create: I removed the function,

But finally I did not put the role:* in the ScopeSystem catalog (NolimitProvider.GetPermissions in common/rbac/const.go). I think it opens the door to much for privilege escalation and only the Sysadmin shall manage the permissions.

- For non admin humans it would add confusion, assign the a role property does not have effect.
- For system robots it is a good permission. Restricting is not trivial as the robot may have different rights in different projects.

Open for discussion if you see this different.

src/common/rbac/const.go
maxgraustenzel-create: I have removed it from the options. But you require the self - read permission in order to access a project. I think the best place is in common/rbac/project/rbac_user.go to have this assigned by default to all users havin some permissions on a project .

src/server/v2.0/handler/role.go
```diff
@@ -0,0 +1,336 @@
...
+func (rAPI *roleAPI) validate(permissions []*models.RolePermission) error {
+	if len(permissions) == 0 {
+		return errors.New(nil).WithMessage("bad request empty permission").WithCode(errors.BadRequestCode)
+	}
+
+	for _, perm := range permissions {
+		if len(perm.Access) == 0 {
+			return errors.New(nil).WithMessage("bad request empty access").WithCode(errors.BadRequestCode)
+		}
+	}
```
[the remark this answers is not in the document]
stonezdj: @maxgraustenzel-create please fix this comment

make/migrations/postgresql/0191_2.16.0_schema.up.sql
```diff
@@ -0,0 +1,24 @@
+/*
```
stonezdj: Please update it in the 0190_2.16.0_schema.up.sql, because 2.16.0 is not released yet, just keep all the   SQL in one sql file.

src/server/v2.0/handler/robot.go
```diff
@@ -457,6 +470,37 @@ func containsAccess(policies []*types.Policy, item *models.Access) bool {
...
+var robotToHumanResourceMap = map[rbac.Resource]rbac.Resource{
+	rbac.ResourceProject: rbac.ResourceSelf,
+}
+
+func mapRobotToHumanResource(r rbac.Resource) rbac.Resource {
+	if mapped, ok := robotToHumanResourceMap[r]; ok {
+		return mapped
+	}
+	return r
+}
```
stonezdj: Suggest to use if to simplify it
```
// mapRobotToHumanResource maps ScopeProject resource names to their ScopeRole equivalents.
// Needed because robots use "project" for the project entity while human roles use "" (ResourceSelf).
func mapRobotToHumanResource(r rbac.Resource) rbac.Resource {
	if r == rbac.ResourceProject {
		return rbac.ResourceSelf
	}
	return r
}
```

src/common/rbac/const.go
stonezdj: Agree, the ResourceSelf is ensential to each user.

src/server/v2.0/handler/role.go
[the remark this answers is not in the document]
maxgraustenzel-create: Thanks — already fixed in a0e7de9. `validate()` rejects a nil permission, empty `Access`, and a nil access entry before dereferencing, so `permissions: [null]` / `access: [null]` return 400 instead of panicking. If you meant a stale code comment instead, point me to the line and I'll fix it.

src/common/rbac/const.go
maxgraustenzel-create: Thanks! Final shape (b6128ce6): `self:*` stays out of the catalog (not selectable), and `self:read` is granted to every project member in `rbacUser.GetPolicies` — scoped per-project, no DB seeding — so built-in and custom roles get visibility the same way.

src/controller/role/controller.go
```diff
@@ -0,0 +1,324 @@
...
+		return nil
+	}
+	rolePermissions, err := d.rbacMgr.GetPermissionsByRole(ctx, ROLETYPE, r.ID)
+
+	if err != nil {
+		log.Errorf("failed to get permissions of role %d: %v", r.ID, err)
+		return err
+	}
+	if len(rolePermissions) == 0 {
+		return nil
```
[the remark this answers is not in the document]
stonezdj: @maxgraustenzel-create it seems to be a valid issue

src/server/v2.0/handler/role.go
stonezdj: yes, please limit the permission to sysadmin can create/update roles

src/pkg/role/model/model.go
```diff
@@ -0,0 +1,67 @@
...
+	RoleMask    int64     `orm:"column(role_mask)" json:"role_mask"`
+	RoleCode    string    `orm:"column(role_code)" json:"role_code"`
+	IsBuiltin   bool      `orm:"column(is_builtin)" json:"is_builtin"`
+	Description string    `orm:"column(description)" json:"description"`
+	Modified    bool      `orm:"column(modified)" json:"modified"`
+	CreatedBy   string    `orm:"column(created_by)" json:"created_by"`
+	CreatedAt   time.Time `orm:"column(created_at);auto_now_add;type(datetime)" json:"created_at"`
+	ModifiedBy  string    `orm:"column(modified_by)" json:"modified_by"`
+	ModifiedAt  time.Time `orm:"column(modified_at);auto_now;type(datetime)" json:"modified_at"`
+}
```
bupd, line 43: No `project_id`. This single omission is what forces role management to be sysadmin-only, makes `uq_role_name` global, and makes the delete guard count across every project. Adding `project_id int64 \`orm:"column(project_id)"\`` here with `UNIQUE (project_id, name)` is what makes the rest of the design work as a project-level feature.

make/migrations/postgresql/0190_2.16.0_schema.up.sql
```diff
@@ -19,3 +19,28 @@ ALTER SEQUENCE robot_id_seq AS bigint MAXVALUE 9007199254740991;
...
+ALTER TABLE role ADD COLUMN IF NOT EXISTS modified     BOOLEAN      NOT NULL DEFAULT FALSE;
+ALTER TABLE role ADD COLUMN IF NOT EXISTS created_by   VARCHAR(255);
+ALTER TABLE role ADD COLUMN IF NOT EXISTS created_at   TIMESTAMP WITH TIME ZONE;
+ALTER TABLE role ADD COLUMN IF NOT EXISTS modified_by  VARCHAR(255);
+ALTER TABLE role ADD COLUMN IF NOT EXISTS modified_at  TIMESTAMP WITH TIME ZONE;
+
+-- Widen the role name to match the API/UI contract (was varchar(20)) and enforce
+-- name uniqueness so custom roles cannot collide.
+ALTER TABLE role ALTER COLUMN name TYPE varchar(255);
+CREATE UNIQUE INDEX IF NOT EXISTS uq_role_name ON role (name);
```
bupd: Global unique index on `name` means two projects can never both have a role called "auditor". If roles become project-scoped this wants to be `UNIQUE (project_id, name)`.

Separately, this is case-sensitive, so "Auditor" and "auditor" coexist. #22815 review asked for a decision on that and it is still open. `lower(name)` in the index if you want them to collide.

src/common/rbac/project/evaluator.go
```diff
@@ -41,12 +42,29 @@ func NewBuilderForUser(user *models.User, ctl project.Controller) RBACUserBuilde
...
+			if isBuiltinProjectRole(roleID) {
+				roles = append(roles, &projectRBACRole{projectID: p.ProjectID, roleID: roleID})
+				continue
+			}
+			// Custom roles load their permissions from the database.
+			r, err := ctlR.Get(ctx, int64(roleID), &role.Option{WithPermission: true})
+			if err != nil {
+				log.Errorf("failed to get role %d: %v", roleID, err)
+				return nil
+			}
```
bupd, line 69: Returning `nil` here drops the whole `rbacUser`, not just the role that failed to load. The namespace evaluator then has no evaluator for the project, so the member loses **every** role they hold there, built-ins included, and gets a 403 with only a core log line to explain it.

Two consequences: a transient DB error becomes a total denial, and a `project_member.role` pointing at a deleted role is a permanent lockout. Either `continue` past the failing role or propagate a real error so the caller can distinguish "no access" from "lookup broke".

src/server/v2.0/handler/robot.go
```diff
@@ -457,6 +470,33 @@ func containsAccess(policies []*types.Policy, item *models.Access) bool {
...
+
+// validateNoEscalation ensures that a human user cannot grant a robot more permissions than they hold.
+func (rAPI *robotAPI) validateNoEscalation(ctx context.Context, permissions []*models.RobotPermission) error {
+	for _, perm := range permissions {
+		for _, acc := range perm.Access {
+			resource := mapRobotToHumanResource(rbac.Resource(acc.Resource))
+			has, err := rAPI.HasProjectPermission(ctx, perm.Namespace, rbac.Action(acc.Action), resource)
```
bupd, line 487: This compares requested robot access against the creator's **effective** project permissions, but the robot UI populates its picker from the `ScopeProject` catalog, which is a strict superset. The one element of difference is `metadata:list`, so a project admin can no longer create a robot with a permission the UI offered them. Repro is in the summary comment.

Fix either way: add `{Resource: ResourceMetadata, Action: ActionList}` to `rolePoliciesMap["projectAdmin"]`, or validate against the catalog instead of effective permissions.

src/server/v2.0/handler/role.go
```diff
@@ -0,0 +1,325 @@
...
+	// to validate the access scope
+	for _, perm := range permissions {
+		if perm.Kind == role.LEVELROLE {
+			polices := provider.GetPermissions(rbac.ScopeRole)
+
+			for _, acc := range perm.Access {
+				if !containsAccess(polices, acc) {
+					return errors.New(nil).WithMessagef("bad request permission: %s:%s", acc.Resource, acc.Action).WithCode(errors.BadRequestCode)
+				}
+			}
```
bupd, line 268: `containsAccess` matches on Resource and Action only, so `Effect` is never validated. The client string lands in `permission_policy.effect` and goes verbatim to casbin, whose model is `e = some(where (p.eft == allow)) && !some(where (p.eft == deny))`.

So `"effect": "Allow"` (capitalised) or `"effect": "foo"` matches neither arm: 201 Created, `GET /roles/{id}` echoes the permission back, and it silently grants nothing forever. A literal `deny` is also accepted with no documented semantics and no UI. Reject anything but `""` and `allow`, or support `deny` on purpose.

src/controller/role/controller.go
```diff
@@ -0,0 +1,324 @@
...
+	}
```
bupd, line 295: Permissions are populated from `role_permission` only, and built-in roles have no rows there by design, so `GET /roles` and `GET /roles/{id}` report all five built-ins as having zero permissions. `BuiltinRolePolicies()` was added in this PR for exactly this and is wired into the member escalation check but not into the read path. Any UI built on `GET /roles` will render projectAdmin as permissionless.

src/server/v2.0/handler/permissions.go
```diff
@@ -33,16 +31,19 @@ import (
...
-		mc: member.NewController(),
 	}
 }
 
+// GetPermissions returns the available permission catalog.
+// Accessible to all authenticated users — the catalog is read-only metadata
+// (action names such as repository:pull) used by project-level UI components
+// (robot accounts, webhooks, role management). System-level permissions are
+// only included in the response for system admins.
 func (p *permissionsAPI) GetPermissions(ctx context.Context, _ permissions.GetPermissionsParams) middleware.Responder {
```
bupd, line 53: This drops the previous `isSystemAdmin || isProjectAdmin` gate: before this PR a plain member got a 403, now any authenticated user gets the full project and role catalog. Read-only metadata, granted, but it is an authorization change unrelated to custom roles and it is not mentioned in the PR body or the thread. Worth calling out explicitly, or keeping the project-admin gate and adding a narrower endpoint for what the role editor needs.

src/common/rbac/const.go
```diff
@@ -311,5 +317,98 @@ var (
 
 			{Resource: ResourceQuota, Action: ActionRead},
 		},
+		ScopeRole: {
```
bupd, line 320: Hand-writing this catalog creates a third place every future project resource has to be registered, and forgetting it fails silently: the resource simply becomes ungrantable to custom roles, with no compile error and no test failure.

Measured against the built-in maps, this list is exactly `projectAdmin` plus `metadata:list` minus `self:{read,update,delete}` (73 entries vs 75). Deriving it from `rolePoliciesMap` at init keeps it correct by construction and makes the ceiling of a custom role self-documenting.

src/server/v2.0/handler/role.go
```diff
@@ -0,0 +1,325 @@
...
+	if err := rAPI.RequireAuthenticated(ctx); err != nil {
+		return rAPI.SendError(ctx, err)
+	}
+
+	query, err := rAPI.BuildQuery(ctx, params.Q, params.Sort, params.Page, params.PageSize)
+	if err != nil {
+		return rAPI.SendError(ctx, err)
+	}
+
+	query.Keywords["Visible"] = true
```
bupd: `model.Role` has no `Visible` field, and `setFilters` in `lib/orm/query.go` silently skips keywords it cannot filter on, so this line is a no-op left over from #22815. Consequence is that list returns built-ins too, which compounds the empty-permissions problem on the read path.

src/controller/role/controller.go
```diff
@@ -0,0 +1,324 @@
...
+	// Apply the field update and permission replacement atomically: deleting the
+	// old permissions and inserting the new set must never be observable as a
+	// partial state, which would silently change the role's authorization.
+	if err := orm.WithTransaction(func(ctx context.Context) error {
+		if err := d.roleMgr.Update(ctx, &model.Role{
+			ID:          r.ID,
+			Description: r.Description,
+			Modified:    true,
+			ModifiedBy:  modifiedBy,
+		}, "description", "modified", "modified_by", "modified_at"); err != nil {
```
bupd: `Update` persists only `description, modified, modified_by, modified_at`, but the swagger body for `PUT /roles/{id}` is a full `Role` including `name`. A rename is silently dropped and still returns 200. Either persist `name` (with the uniqueness check) or reject a changed name with a 400.

src/controller/role/controller.go
```diff
@@ -0,0 +1,324 @@
...
+		if err != nil {
+			return err
+		}
+		if assigned > 0 {
+			return errors.PreconditionFailedError(nil).WithMessagef("cannot delete role %d: it is still assigned to %d project member(s)", id, assigned)
+		}
+		if err := d.roleMgr.Delete(ctx, id); err != nil {
+			return err
+		}
+		return d.rbacMgr.DeletePermissionsByRole(ctx, ROLETYPE, id)
```
bupd: `DeletePermissionsByRole` returns `NotFoundError` when it deletes zero rows, and that error propagates unchanged out of `orm.WithTransaction`, rolling the delete back. So a custom role with no permission rows cannot be deleted. `Update` already guards this exact case with `&& !errors.IsNotFoundErr(err)` on line 195; `Delete` should match.

src/controller/role/controller.go
```diff
@@ -0,0 +1,324 @@
...
+		roleID = id
+		return d.createPermission(ctx, r)
+	})(ctx); err != nil {
+		return 0, err
+	}
+	return roleID, nil
+}
+
+// Delete ...
+func (d *controller) Delete(ctx context.Context, id int64, option ...*Option) error {
```
bupd: `option` is never read in this function, which revive's `unused-parameter` will flag once the build jobs run. Drop the variadic or use `_ ...*Option` if the interface shape has to stay.

src/pkg/member/dao/dao.go
```diff
@@ -138,6 +140,18 @@ func (d *dao) GetTotalOfProjectMembers(ctx context.Context, projectID int64, _ *
 	return count, err
 }
 
+func (d *dao) GetTotalOfProjectMembersByRole(ctx context.Context, roleID int) (int, error) {
+	var count int
+	o, err := orm.FromContext(ctx)
+	if err != nil {
+		return 0, err
+	}
+	if err := o.Raw("SELECT COUNT(1) FROM project_member WHERE role = ?", roleID).QueryRow(&count); err != nil {
```
bupd, line 149: No project filter, so this counts assignments of the role in every project. With roles global that is the only correct query, which is itself the argument for scoping them; with `project_id` on the role it becomes `AND project_id = ?` and the guard stops being cross-tenant.

src/server/v2.0/handler/role.go
```diff
@@ -0,0 +1,325 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+// isValidPermissionScope checks if permission slice A is a subset of permission slice B
+func isValidRolePermissionScope(creating []*models.RolePermission, creator []*role.Permission) bool {
+	creatorMap := make(map[string]*role.Permission)
+	for _, creatorPerm := range creator {
```
bupd, line 336: `isValidRolePermissionScope` and `containsRoleAccess` are never called outside `role_test.go`. They will not trip the `unused` linter because golangci loads test files, so they will sit here looking covered while doing nothing in production.

Worth keeping rather than deleting, though: a subset check against the creator's own permissions is exactly the guard needed to make role authoring delegable to project admins. It looks like it was written for that and then stranded when management went sysadmin-only.

src/server/v2.0/handler/role.go
```diff
@@ -0,0 +1,325 @@
...
+		if !r.Editable && errors.IsNotFoundErr(err) {
+			return operation.NewDeleteRoleOK()
+		}
+		return rAPI.SendError(ctx, err)
+	}
+	return operation.NewDeleteRoleOK()
+}
+
+func (rAPI *roleAPI) ListRole(ctx context.Context, params operation.ListRoleParams) middleware.Responder {
+	if err := rAPI.RequireAuthenticated(ctx); err != nil {
```
bupd, line 128: `ListRole` and `GetRoleByID` gate on `RequireAuthenticated` only, while writes are sysadmin-only, so any logged-in user can enumerate every custom role and its complete permission set. Probably deliberate for member pickers, but it is worth stating in the swagger description since it is asymmetric with the write path.

bupd (changes requested):
Thanks for splitting this out of #22815. Keeping built-in roles on the compile-time `rolePoliciesMap` and persisting only custom roles is the right call, and the built-in path really is untouched: no DB lookup, no cache, one optional field on `projectRBACRole`. Rollback is a dropped column set.

Two things before the line-by-line.

### DCO

The last four commits are unsigned: `fef2d243`, `b0108521`, `da386e73`, `4966fb36`. `git rebase --signoff` over those will clear the check.

### Roles are global objects, and I think that is the core design issue

The `role` table has no `project_id`. Everything awkward downstream follows from that:

* `uq_role_name` is global, so two teams cannot both define an "auditor" role.
* The delete guard counts members across all projects (`SELECT COUNT(1) FROM project_member WHERE role = ?`), so usage in project A blocks cleanup in project B.
* `createPermission` hardcodes `policy.Scope = "/project/*"`, so the stored scope carries no project information at all.
* Management has to be sysadmin-only, because a global object edited by one project admin silently changes authorization in every other project.

That last point is what makes role management undelegable in this shape, and it is also the honest answer to the concern about a robot needing different rights in different projects. @stonezdj raised project scoping earlier in the thread, and I would settle it before this API freezes.

Concretely:

```sql
role: role_id, project_id NOT NULL, name, ...
UNIQUE (project_id, name)
```

with role management as an ordinary project permission rather than a system one:

* `role:read` for every project member, so member pickers keep working.
* `role:create`, `role:update`, `role:delete` grantable, so a project admin can delegate role authoring.
* Built-ins protected by a single condition: reject any write where `role_id < 6`.

This is only ever safe in the downward direction, since a custom role can carve out of what its author already holds and nothing more. That is exactly the property `isValidRolePermissionScope` in this PR enforces, and it is currently unreachable outside tests, which suggests it was written for this model.

### The robot regression

`metadata:list` is a live regression against 2.15:

```bash
curl -u projadmin:Harbor12345 -X POST https://harbor/api/v2.0/robots \
  -H 'Content-Type: application/json' -d '{
    "name":"repro","level":"project","duration":30,
    "permissions":[{"kind":"project","namespace":"library",
      "access":[{"resource":"metadata","action":"list"}]}]}'

# 2.15:     201 Created
# this PR:  403 permission escalation not allowed: you do not have metadata:list
```

The permission sits in the robot `ScopeProject` catalog and the UI offers it, but it is absent from `rolePoliciesMap["projectAdmin"]`, so the new effective-permission check rejects a project admin's own robot. Sysadmins never hit it because they bypass the evaluator. Set difference between the robot catalog and projectAdmin, after `mapRobotToHumanResource` maps `project` to `self`, is exactly `{metadata:list}`.

The rest is in line comments below.

maxgraustenzel-create:
@bupd thanks — this is a genuinely good analysis, and you're right that `project_id` is the hinge for a project-scoped, delegable model.

**On the design.** This PR is deliberately scoped to **global, sysadmin-managed roles**, because that's what the feature proposal targets: a curated, org-wide set of custom roles (an `auditor` that means the same thing in every project), administered centrally rather than authored per-project. For that use case a global object owned by a sysadmin is the right shape — and the parts you flag as awkward (global `uq_role_name`, the cross-project delete count) are the *correct* behavior for a global role, not accidents. @stonezdj also confirmed role management should stay sysadmin-only.

Project-scoped roles are a real, complementary feature and we don't want to foreclose the design you describe. The clean way to add it later without a breaking change would be a **scope discriminator** on `role` — a `scope` column, `global` today and `project` later:
- **global** (this PR): `project_id` NULL, `UNIQUE(name)`, sysadmin-managed.
- **project** (future): `scope='project'`, `project_id` set, `UNIQUE(project_id, name)`, management as a delegable project permission (`role:create/update/delete`), built-ins protected by `role_id < 6`, and the downward-only subset check — which is exactly what `isValidRolePermissionScope` in this PR already implements (kept deliberately for that path).

So this PR ships the global tier and leaves a clean additive seam for the project tier to drop in beside it.

**On the bugs** — all valid, thank you. Fixed in `bd9248824`:
- `evaluator.go` nil-drop → skip only the failing role (no more total lockout on a transient error / dangling `project_member.role`).
- `metadata:list` regression → added to `rolePoliciesMap["projectAdmin"]` (restores 2.15; project admin can create the robot again).
- built-ins reported permissionless on `GET /roles(/{id})` → now resolved from `rolePoliciesMap` via `BuiltinRolePolicies`; dead `Visible` filter removed.
- `/permissions` gate → restored `IsSysAdmin || IsProjectAdmin`; role-level catalog now sysadmin-only.
- `Effect` never validated → reject anything but `""`/`allow`.
- `Delete` on a zero-permission role → NotFound no longer rolls back; unused `option` param dropped.
- `Update` silently dropped a rename → now persisted (unique index rejects collisions).
- `const.go` hand-written `ScopeRole` catalog drift → added a test asserting it equals `projectAdmin` minus `self:{read,update,delete}` (couldn't derive it directly — `common/rbac` can't import `common/rbac/project` without a cycle — so the test guards it instead).
- read/write authz asymmetry → documented in swagger.
- DCO sign-off in progress.

`isValidRolePermissionScope` is kept as you suggested — it's the guard the future project tier needs. On case-sensitivity of role names: keeping it case-sensitive for now; happy to switch to `lower(name)` if you feel strongly.

src/server/v2.0/handler/role.go
maxgraustenzel-create: Confirmed — role create/update/delete are sysadmin-only. They use `RequireSystemAccess(ctx, action, rbac.ResourceRole)`, and `ResourceRole` is deliberately kept out of the grantable system catalog, so only a sysadmin (via the admin evaluator) can pass — it can't be delegated to a robot or a project admin.

bupd:
case-insensitive is the way to go for roles otherwise there could be misunderstandings

Maintainer / maintainer would yield different results

maxgraustenzel-create:
@bupd agreed — switched to case-insensitive uniqueness in `b17acacd2`: `uq_role_name` is now on `lower(name)`, so `Maintainer` and `maintainer` collide. Thanks.

chlins (commented):
Went through the latest revision, a few things:

**1. Robot secret refresh can be used to escalate.** Custom roles can now be granted `robot:update` on its own (`common/rbac/const.go`), but `RefreshSec` in `handler/robot.go` only checks `robot:update` and then hands out a new secret. So a member whose role has `robot:update` but no `repository:push` can reset the secret of an existing robot that does have push, and use it. Create/Update run the escalation check on the robot's permissions; refresh should do the same, since taking over the credential is effectively the same as granting those permissions.

**2. Deny entries are treated as required grants in the member escalation check.** `handler/member.go` walks every access of the target role and requires the caller to hold it, without looking at `Effect`. A role with `allow member:create` + `deny repository:delete` therefore can't be assigned by someone holding that same role, because the check demands `repository:delete`. Only allow entries should count here.

**3. Role delete vs. member assign race.** The delete transaction counts members then deletes, but member Create/UpdateRole validate the role in a separate request with no shared lock or FK on `project_member.role`. Validate → count(0) → delete → insert member still lands a dangling role id. The follow-up comment in the code acknowledges this; just flagging that the orphan-delete guard isn't complete as-is. A FK or a shared lock on the role row would close it.

Also small test note: the role CRUD transaction tests only cover the happy path, so dropping the transaction wouldn't fail anything. A failure-injection case for Create/Update/Delete would be good.

stonezdj:
@bupd the custom role is suppose to be defined in system level role, not in project level role.

src/controller/role/controller.go
```diff
@@ -0,0 +1,331 @@
...
+}
+
+func (d *controller) createPermission(ctx context.Context, r *Role) error {
+	if r == nil {
+		return nil
+	}
+
+	for _, per := range r.Permissions {
+		policy := &rbac_model.PermissionPolicy{}
+		policy.Scope = "/project/*"
```
stonezdj, line 240: Please do not ust the `/project/*` to be the scope of the current permission, in the context of casbin, it could match the resource `/project/1/robot` or any other resource under the catagory of project. it should be `/project/:pid`, where the pid is the concret project id.  see: https://github.com/goharbor/harbor/blob/main/src/pkg/permission/evaluator/rbac/casbin_match.go#L91

src/controller/role/model.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"github.com/goharbor/harbor/src/pkg/permission/types"
+	"github.com/goharbor/harbor/src/pkg/role/model"
+)
+
+const (
+	LEVELROLE = "project-role"
+
+	// SCOPEALLPROJECT ...
+	SCOPEALLPROJECT = "/project/*"
```
stonezdj, line 26: same issue as the controller.go

PR #23806, opened 2026-08-29
fix(exporter): use consistent time units for cache expiration

### LINKED ISSUE #23805: Exporter cache cleaner compares Unix seconds with Unix nanoseconds

The exporter cache cleaner should remove entries only after their configured cache duration expires.

Currently, cached entries are stored with an expiration timestamp in Unix seconds:

```go
Expiration: time.Now().Unix() + c.CacheDuration
```

However, `StartCacheCleaner` compares this value with a Unix nanosecond timestamp:

```go
now := time.Now().UnixNano()
if v.Expiration < now {
    delete(c.store, k)
}
```

Because Unix seconds and Unix nanoseconds have different scales, nearly all cache entries are treated as expired when the cleaner runs for the first time, even if their TTL has not elapsed.

1. Enable the Harbor exporter cache with a positive cache duration.
2. Add an entry to the exporter cache.
3. Wait for the cache cleaner interval.
4. Observe that the entry is removed before its configured expiration time.

**Versions:**

- harbor version: `main` at commit `5d19f2625`
- docker engine version: [28.1.1]
- docker-compose version: [v2.35.1-desktop.1]

**Additional context:**

The cache expiration value and all expiration checks should use the same time unit. The proposed fix is to use `time.Now().Unix()` in the cleaner and add a deterministic regression test.

cuishuang:

Fix the Harbor exporter cache cleaner so that it compares cache expiration timestamps using the same time unit used when entries are stored.

`CachePut` stores expiration timestamps in Unix seconds, while `StartCacheCleaner` previously used Unix nanoseconds for the current time. This caused valid cache entries to be deleted during the first cleaner run.

This change:

- Uses Unix seconds in `StartCacheCleaner`.
- Extracts the expiration removal logic into `cleanExpired(now int64)`.
- Adds regression coverage for valid and expired cache entries.

# Issue being fixed

Fixes  https://github.com/goharbor/harbor/issues/23805

- [ ] Label the PR as needed.
- [x] Considered the docs impact; no documentation changes are required for this bug fix.

### DIFF
--- a/src/pkg/exporter/cache.go
+++ b/src/pkg/exporter/cache.go
@@ -70,7 +70,10 @@ func CacheDelete(key string) {
 
 // StartCacheCleaner start a cache clean job
 func StartCacheCleaner() {
-	now := time.Now().UnixNano()
+	cleanExpired(time.Now().Unix())
+}
+
+func cleanExpired(now int64) {
 	c.Lock()
 	defer c.Unlock()
 	for k, v := range c.store {
--- a/src/pkg/exporter/cache_test.go
+++ b/src/pkg/exporter/cache_test.go
@@ -68,3 +68,28 @@ func TestCleanExpired(t *testing.T) {
 		t.Fatal("cleanExpired retained an expired cache entry")
 	}
 }
+
+func TestStartCacheCleaner(t *testing.T) {
+	CacheInit(&Opt{CacheDuration: 60})
+	CachePut("fresh", "v")
+
+	c.Lock()
+	c.store["stale"] = cachedValue{
+		Value:      "v",
+		Expiration: time.Now().Unix() - 1,
+	}
+	c.Unlock()
+
+	StartCacheCleaner()
+
+	if _, ok := CacheGet("fresh"); !ok {
+		t.Fatal("StartCacheCleaner removed a fresh entry")
+	}
+
+	c.RLock()
+	_, ok := c.store["stale"]
+	c.RUnlock()
+	if ok {
+		t.Fatal("StartCacheCleaner retained an expired entry")
+	}
+}
--- a/src/pkg/exporter/cache_test.go
+++ b/src/pkg/exporter/cache_test.go
@@ -1,6 +1,7 @@
 package exporter
 
 import (
+	"testing"
 	"time"
 
 	"github.com/stretchr/testify/suite"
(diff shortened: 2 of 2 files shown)


chlins:
Thanks for the fix, the analysis is correct. `CachePut`/`CacheGet` both use Unix seconds while the cleaner compared against `UnixNano()`, so every cleaner tick wiped the entire cache. This has been there since the cache was introduced (d95f22448, 2020). Since `CacheGet` checks expiration with consistent units, it was a cache-effectiveness bug rather than a correctness one, which is probably why it went unnoticed for so long.

One gap in the regression test: `TestCleanExpired` calls `cleanExpired(now)` directly with a seconds value, so it verifies the comparison logic but never exercises `StartCacheCleaner`, where the actual bug lived (the unit chosen at the call site). I verified this by reintroducing the original bug on top of your branch:

```go
func StartCacheCleaner() {
	cleanExpired(time.Now().UnixNano())
}
```

`TestCleanExpired` still passes. The `codecov/patch` failure is flagging the same uncovered line.

Suggest routing the assertions through the public entry point, which stays deterministic (60s TTL vs an immediate call):

```go
func TestStartCacheCleaner(t *testing.T) {
	CacheInit(&Opt{CacheDuration: 60})
	CachePut("fresh", "v")
	c.Lock()
	c.store["stale"] = cachedValue{Value: "v", Expiration: time.Now().Unix() - 1}
	c.Unlock()

	StartCacheCleaner()

	if _, ok := CacheGet("fresh"); !ok {
		t.Fatal("StartCacheCleaner removed a fresh entry")
	}
	if _, ok := c.store["stale"]; ok {
		t.Fatal("StartCacheCleaner retained an expired entry")
	}
}
```

With that covered, LGTM.

cuishuang:
> Thanks for the fix, the analysis is correct. `CachePut`/`CacheGet` both use Unix seconds while the cleaner compared 
> With that covered, LGTM.

@chlins Thanks for the review. I added `TestStartCacheCleaner` to cover the actual cleaner entry point and pushed. Please review it again.

chlins (approved):
lgtm

cuishuang:
> lgtm

@chlins Thank you for your approval.  If there's anything need I to do, please feel free to let me know.

chlins:
@cuishuang Please squash your commits to resolve the DCO action.

wy65701436 (approved):
lgtm

PR #23824, opened 2026-09-02
fix(orm): close transaction on panic and honor request context in WithTransaction

### LINKED ISSUE #23813: Manifest push can leave a connection idle in transaction, blocking later pushes

Expected: when a client disconnects during a push, the database transaction that harbor-core opened for that request gets closed.

Actual: it stays open. It holds onto its database connection and its row locks until harbor-core is restarted or the connection is killed by hand.

The effect is that one stuck push blocks every later push to the same project. In my case the transaction was holding a row lock on `quota_usage`, nine more requests queued up behind it, and the project accepted no new artifacts until I restarted harbor-core.

This has happened to me twice and neither time did it recover on its own. I killed the first one with `pg_terminate_backend`; this one needed the restart. Evidence and what I found in the code are in Additional context below.

I could not get the original push to fail again on demand, so I went at it from the other end. The test below calls `WithTransaction` the way the middleware does, then looks at Postgres over a separate connection to see whether the transaction is still open and still blocking other writers.

I wrote this test with [Claude Code](https://claude.com/claude-code). I have run it and checked the output below myself, but flagging it so you know where it came from.

1. Start a throwaway Postgres:

```
docker run -d --name pg-test -e POSTGRES_PASSWORD=root123 -e POSTGRES_DB=registry -p 55432:5432 postgres:15
```

2. Save the file below as `src/lib/orm/test/txleak_repro_test.go`.

3. Run it:

```
cd src
export POSTGRESQL_HOST=127.0.0.1 POSTGRESQL_PORT=55432 POSTGRESQL_USR=postgres \
       POSTGRESQL_PWD=root123 POSTGRESQL_DATABASE=registry
go test ./lib/orm/test/ -run 'TestLeak|TestControl' -v -count=1
```

All three pass on v2.15.2:

```
--- PASS: TestLeak_HandlerBlocks    LEAK CONFIRMED
--- PASS: TestLeak_HandlerPanics    LEAK CONFIRMED
--- PASS: TestControl_NormalReturn  CONTROL PASSED
```

`TestLeak_HandlerBlocks` takes a row lock and then blocks while the request context gets cancelled, which stands in for the client disconnecting. The transaction stays open and another writer is stuck behind it. `TestLeak_HandlerPanics` takes the lock and panics, with the same result. `TestControl_NormalReturn` runs the same code but returns normally, and that one commits and releases the lock, so the test is not manufacturing the problem by itself.

<details>
<summary>src/lib/orm/test/txleak_repro_test.go</summary>

```go
package test

import (
	"context"
	"database/sql"
	"fmt"
	"os"
	"sync"
	"testing"
	"time"

	_ "github.com/lib/pq"

	beegoorm "github.com/beego/beego/v2/client/orm"
	. "github.com/goharbor/harbor/src/lib/orm"
)

func inspectDSN() string {
	return fmt.Sprintf("host=%s port=%s user=%s password=%s dbname=%s sslmode=disable",
		os.Getenv("POSTGRESQL_HOST"), os.Getenv("POSTGRESQL_PORT"), os.Getenv("POSTGRESQL_USR"),
		os.Getenv("POSTGRESQL_PWD"), os.Getenv("POSTGRESQL_DATABASE"))
}

var setupOnce sync.Once

func setup(t *testing.T) {
	t.Helper()
	setupOnce.Do(func() {
		_ = beegoorm.RegisterDriver("postgres", beegoorm.DRPostgres)
		if err := beegoorm.RegisterDataBase("default", "postgres", inspectDSN()); err != nil {
			t.Fatalf("register beego db: %v", err)
		}
		db, err := sql.Open("postgres", inspectDSN())
		if err != nil {
			t.Fatalf("open: %v", err)
		}
		defer db.Close()
		if _, err := db.Exec(`CREATE TABLE IF NOT EXISTS foo (
			id SERIAL PRIMARY KEY NOT NULL,
			name VARCHAR(30),
			UNIQUE (name))`); err != nil {
			t.Fatalf("create table: %v", err)
		}
	})
}

func openInspector(t *testing.T) *sql.DB {
	t.Helper()
	setup(t)
	db, err := sql.Open("postgres", inspectDSN())
	if err != nil {
		t.Fatalf("inspector open: %v", err)
	}
	return db
}

func idleInTxHoldingFoo(t *testing.T, db *sql.DB) int {
	t.Helper()
	var n int
	err := db.QueryRow(`
		SELECT count(DISTINCT a.pid)
		FROM pg_stat_activity a
		JOIN pg_locks l ON l.pid = a.pid
		WHERE a.state = 'idle in transaction'
		  AND l.relation = 'foo'::regclass
		  AND l.mode = 'RowExclusiveLock'`).Scan(&n)
	if err != nil {
		t.Fatalf("inspect: %v", err)
	}
	return n
}

func writeBlocked(t *testing.T, db *sql.DB, name string, timeout time.Duration) bool {
	t.Helper()
	ctx, cancel := context.WithTimeout(context.Background(), timeout)
	defer cancel()
	_, err := db.ExecContext(ctx, `UPDATE foo SET name = name || '-x' WHERE name = $1`, name)
	return err != nil
}

func seedRow(t *testing.T, db *sql.DB, name string) {
	t.Helper()
	if _, err := db.Exec(`INSERT INTO foo (name) VALUES ($1) ON CONFLICT DO NOTHING`, name); err != nil {
		t.Fatalf("seed: %v", err)
	}
}

// ctxFor builds the context WithTransaction expects, exactly like
// server/middleware/orm.Middleware does per request.
func ctxFor(parent context.Context) context.Context {
	return NewContext(parent, beegoorm.NewOrm())
}

// TestLeak_HandlerBlocks: the wrapped function never returns while the client
// goes away and its request context is cancelled.
func TestLeak_HandlerBlocks(t *testing.T) {
	inspector := openInspector(t)
	defer inspector.Close()

	const row = "leak-blocks"
	seedRow(t, inspector, row)

	release := make(chan struct{})
	entered := make(chan struct{})
	reqCtx, cancelReq := context.WithCancel(context.Background())

	go func() {
		_ = WithTransaction(func(ctx context.Context) error {
			o, err := FromContext(ctx)
			if err != nil {
				return err
			}
			// Take the row lock, like the quota_usage UPDATE on manifest push.
			if _, err := o.Raw(`UPDATE foo SET name = name WHERE name = ?`, row).Exec(); err != nil {
				return err
			}
			close(entered)
			<-release // the downstream call that never comes back
			return nil
		})(ctxFor(reqCtx))
	}()

	<-entered
	cancelReq() // client disconnects
	time.Sleep(2 * time.Second)

	if got := idleInTxHoldingFoo(t, inspector); got != 1 {
		t.Fatalf("expected 1 abandoned idle-in-transaction backend holding the row lock, got %d", got)
	}
	if !writeBlocked(t, inspector, row, 3*time.Second) {
		t.Fatal("expected an independent writer to be blocked by the abandoned transaction")
	}
	t.Log("LEAK CONFIRMED: request context cancelled, transaction still open, row lock still held, writers blocked")

	close(release)
}

// TestLeak_HandlerPanics: a panic unwinds past Commit and Rollback.
func TestLeak_HandlerPanics(t *testing.T) {
	inspector := openInspector(t)
	defer inspector.Close()

	const row = "leak-panics"
	seedRow(t, inspector, row)

	done := make(chan struct{})
	go func() {
		defer close(done)
		defer func() { _ = recover() }()
		_ = WithTransaction(func(ctx context.Context) error {
			o, err := FromContext(ctx)
			if err != nil {
				return err
			}
			if _, err := o.Raw(`UPDATE foo SET name = name WHERE name = ?`, row).Exec(); err != nil {
				return err
			}
			panic("boom")
		})(ctxFor(context.Background()))
	}()

	<-done
	time.Sleep(2 * time.Second)

	if got := idleInTxHoldingFoo(t, inspector); got < 1 {
		t.Fatalf("expected an abandoned idle-in-transaction backend holding the row lock, got %d", got)
	}
	if !writeBlocked(t, inspector, row, 3*time.Second) {
		t.Fatal("expected an independent writer to be blocked by the abandoned transaction")
	}
	t.Log("LEAK CONFIRMED: panic unwound past Commit/Rollback, transaction still open, row lock still held")
}

// TestControl_NormalReturn: the same code returning normally commits and releases.
func TestControl_NormalReturn(t *testing.T) {
	inspector := openInspector(t)
	defer inspector.Close()

	const row = "control-ok"
	seedRow(t, inspector, row)

	err := WithTransaction(func(ctx context.Context) error {
		o, err := FromContext(ctx)
		if err != nil {
			return err
		}
		_, err = o.Raw(`UPDATE foo SET name = name WHERE name = ?`, row).Exec()
		return err
	})(ctxFor(context.Background()))
	if err != nil {
		t.Fatalf("control transaction failed: %v", err)
	}

	time.Sleep(1 * time.Second)
	if writeBlocked(t, inspector, row, 3*time.Second) {
		t.Fatal("control: writer should not be blocked after a normal commit")
	}
	t.Log("CONTROL PASSED: normal return commits and releases the lock")
}
```

</details>

**Versions:**

- harbor version: v2.15.2 (`goharbor/harbor-core:v2.15.2`), Helm chart 1.19.2 on Kubernetes
- beego v2.3.10, PostgreSQL 18.6 (CloudNativePG, external to the chart)
- `src/lib/orm/orm.go` and `src/lib/orm/tx.go` are identical on `main` as of writing, so I do not think this is already fixed there
- my full Harbor deployment is public if it helps: https://git.xcd.dev/gabrielcosi/home-ops/src/branch/main/kubernetes/apps/tools/harbor/app

**Additional context:**

### What happened

A `PUT /v2/<repo>/manifests/<digest>` came in and the manifest stored fine, registry returned 201. Two milliseconds later harbor-core ran one more query and went quiet. Postgres has shown that connection as `idle in transaction` ever since.

How long each of my two incidents ran says more about when I noticed than about the bug. A CloudNativePG alert caught the first within the hour. The second started while I was away. The first one was holding a read lock only, so nothing queued behind it and the damage stopped at the one push that hung, which was luck rather than different behaviour.

### The stuck connection

Nothing was blocking it. Every lock it held was granted. It was not waiting on another query; it was waiting for harbor-core to send it something, and harbor-core never did.

```
pid          | 286726
state        | idle in transaction
wait_event   | ClientRead
xact_start   | 2026-08-29 11:03:38.317+02
query_start  | 2026-08-29 11:03:38.479+02   <- its last query; nothing sent since
query        | SELECT ... FROM "artifact" T0 WHERE T0."repository_name" = $1
               AND T0."digest" = $2 ORDER BY ...

holds (all granted): quota_usage RowExclusiveLock, transactionid 264461 ExclusiveLock
pg_blocking_pids(286726) = {}          <- nothing is blocking it
```

Behind it, all of them running `UPDATE quota_usage SET used=$1, update_time=$2, version=$3 WHERE id=$4 AND version=$5`:

```
  pid   |          blocked_by           |     age
--------+-------------------------------+-------------
 287321 | {286726}                      | 2 days 00:25
 287320 | {287321}                      | 2 days 00:24
 287322 | {287321,287320}               | 2 days 00:24
 ...    | (6 more, each behind the last)|
 389213 | {287321,...,325198}           |      09:42
```

The client gave up and retried twice, and each retry added more waiting connections. These are registry log timestamps in UTC, so 09:03:38 here is the same moment as 11:03:38+02 above:

```
09:03:38.477  PUT 201 /v2/<repo>/manifests/sha256:64fc7c21...   <- the request that got stuck
09:03:38.479  the SELECT above runs, then nothing
09:16:50      client retry
09:18:19      client retry
```

### What I found in the code

I had a read through the code. Everything below this point is guesswork.

`transaction.Middleware` wraps the whole request in one transaction, and a manifest PUT is not in `dbTxSkippers` (`src/core/middlewares/middlewares.go`), so the entire push runs inside it, including the `quota_usage` update.

Two things then seem to stop it ever closing:

1. `ormerTx.Begin()` (`src/lib/orm/tx.go:66`) calls beego's `Ormer.Begin()`, and beego starts the transaction with `context.Background()` instead of the request's context. My understanding is that `database/sql` rolls a transaction back when its context is cancelled, but `context.Background()` never gets cancelled, so a client disconnecting cannot trigger it.
2. `WithTransaction` (`src/lib/orm/orm.go:129`) calls `Commit` or `Rollback` only on the normal return paths. There is no `defer`, so if the wrapped function panics instead of returning, neither one runs.

What I could not work out is which call made harbor-core go quiet in the first place. The next thing after that SELECT in `pkg/cached/artifact/redis.Manager.GetByDigest` is a Redis write, and I do run with `CACHE_ENABLED=true`, but my Redis client uses the default 3 second timeout, so that on its own should not park a connection for days. I could not get a goroutine dump either. harbor-core catches SIGQUIT for graceful shutdown (`signal.Notify` at `src/core/main.go:116`), so the Go runtime never printed any stacks. Next time I will have `PPROF_ENABLED=true` set and grab `/debug/pprof/goroutine?debug=2` instead.

### Related issues

#23703 looks like the same kind of problem on `DELETE /repository`, where `retry.Retry` keeps the handler busy. That one stops once `retry.Timeout` elapses. Mine never stopped, and it was a push rather than a delete, so I am filing it separately, though the two may share a cause. #21075 and #18440 are older reports about `quota_usage` contention during pushes.

### Getting out of it

If anyone else hits this, what worked for me was finding the connection with:

```
SELECT pid, client_addr, state, now()-xact_start, left(query,120)
FROM pg_stat_activity WHERE datname='harbor' AND state='idle in transaction';
```

then `SELECT pg_terminate_backend(<pid>);`, or just restarting harbor-core. The queue clears immediately afterwards.

pujitha24:

`orm.WithTransaction` (src/lib/orm/orm.go) only called `Commit`/`Rollback` on
its two normal return paths. If the wrapped function panicked, neither ran,
so the transaction — and the Postgres backend connection and row locks it
held — was never closed. Separately, `ormerTx.Begin()` (src/lib/orm/tx.go)
started the transaction via beego's `Ormer.Begin()`, which internally uses
`context.Background()` instead of the caller's context, so canceling the
request context (e.g. a client disconnect) gave `database/sql` no way to
roll the transaction back on its own. Together these two defects can leave a
transaction open indefinitely, blocking later writers behind its row locks
until the connection is killed by hand or harbor-core is restarted.

This PR fixes both:
- `ormerTx.Begin` now takes a `context.Context` and starts the transaction
  with beego's `BeginWithCtx(ctx)` instead of the context-less `Begin()`, so
  the underlying `sql.Tx` is tied to the caller's context.
- `WithTransaction` now defers a rollback that only fires if neither the
  existing success-path `Commit` nor the existing error-path `Rollback` ran
  — i.e. exactly the panic case. Normal commit/rollback/error-return
  behavior is unchanged.

This fixes two proven code-level defects: a leaked transaction on panic, and
`Begin()` not honoring the request context. It does **not** claim to fully
explain the specific hang described in the issue (a stall after a `SELECT`,
with no panic or confirmed context cancellation observed in that incident) —
that mechanism remains unconfirmed, though these are exactly the kind of bug
that produces the symptom described (an abandoned "idle in transaction"
backend holding `quota_usage` locks).

## Behavior change

Tying the transaction to the request context is a behavior change, not just
a leak fix: previously, if a client disconnected mid-request, the handler
still ran to completion and committed; now `database/sql` rolls the
transaction back as soon as the context is canceled. For something like a
manifest PUT, that can leave the blob already written to storage while the
DB row is rolled back — but that's the same state the code already reaches
today whenever an error occurs after the storage write, and it's healed the
same way, by the client retrying the push.

## Validation

- `go build ./lib/orm/...`
- `go vet ./lib/orm/`
- `golangci-lint run ./lib/orm/...` — no new issues (one pre-existing,
  unrelated `govet` finding in `query.go`)
- `go test ./lib/orm/... -run TestWithTransaction -v` — added
  `TestWithTransaction_PanicRollsBackInsteadOfLeaking`,
  `TestWithTransaction_ErrorRollsBackOnce`, and
  `TestWithTransaction_BeginUsesRequestContext`, all using in-memory fakes
  of beego's `orm.Ormer`/`orm.TxOrmer` (no live database required).
  Confirmed the original two tests fail against the pre-fix code (one
  assertion failure, one nil-pointer panic from the old context-less
  `Begin` signature) and pass after the fix.
- Updated the shared `testing/lib/orm.FakeOrmer.BeginWithCtx` fake to return
  a usable `FakeTxOrmer` instead of `(nil, nil)`, since `WithTransaction` now
  calls it instead of `Begin()`; confirmed `go test ./controller/project/...`
  (flagged by Copilot as a caller that would panic on the nil fake) passes.
- The sibling `lib/orm/test` package requires a live Postgres via
  `POSTGRESQL_HOST` and is a separate, pre-existing integration suite,
  unaffected by this change; it was not run here.
- Base branch (`main`) CI was green at the commit this PR is built on.

# Issue being fixed
Fixes #23813


src/lib/orm/orm.go
```diff
@@ -145,17 +145,32 @@ func WithTransaction(f func(ctx context.Context) error) func(ctx context.Context
...
+		// closed guards the deferred rollback below: it is set right before the
+		// explicit Rollback/Commit calls on every normal return path, so the
+		// deferred call is a no-op then. If f panics, neither of those lines run,
+		// closed stays false, and the deferred call rolls back the transaction
+		// instead of leaking it while the panic unwinds the stack.
+		closed := false
+		defer func() {
+			if !closed {
+				if e := tx.Rollback(); e != nil {
+					log.Errorf("rollback transaction failed: %v", e)
```
chlins: Nit: after this change, when the request context is canceled `database/sql` already rolls the tx back on its own, so both the explicit `Rollback()` on the error path and this deferred one will return `sql.ErrTxDone` and log it at Error level. Client disconnects mid-push are fairly common, so this could get noisy. Consider skipping the log (or dropping it to Debug) when `errors.Is(e, sql.ErrTxDone)`.

src/lib/orm/tx_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	wrapped := WithTransaction(func(context.Context) error {
+		return nil
+	})
+	require.NoError(t, wrapped(ctx))
+
+	require.NotNil(t, o.beginCtx)
+	assert.Equal(t, "request-ctx", o.beginCtx.Value(markerKey{}), "transaction must begin with the request context so cancellation can roll it back")
+	assert.Equal(t, 1, tx.commits)
+}
```
chlins, line 121: Suggest one more case: `f` returns an error, and assert `rollbacks == 1` (not 2) and `commits == 0`. That pins the `closed = true` guard on the error path too, so a future refactor that moves the assignment doesn't silently double-rollback.

chlins (commented):
Could you add a short note to the description (and it's worth a release note) about the behavior change: previously a request whose client disconnected mid-way would still run to completion and commit; now the tx is rolled back on cancellation. For manifest PUT this can mean the blob is already in storage while the DB row is rolled back, but that is the same state we already reach today whenever an error happens after the storage write, and a client retry heals it. Just want it stated explicitly since it's a behavior change, not only a leak fix.

src/lib/orm/tx.go
```diff
@@ -63,11 +63,13 @@ func (o *ormerTx) rollbackToSavepoint() error {
...
 
-func (o *ormerTx) Begin() error {
+// Begin starts the transaction with ctx so that database/sql rolls it back
+// automatically when ctx is canceled, e.g. when the client disconnects.
+func (o *ormerTx) Begin(ctx context.Context) error {
 	if o.TxOrmer != nil {
 		return o.createSavepoint()
 	}
-	txOrmer, err := o.Ormer.Begin()
+	txOrmer, err := o.Ormer.BeginWithCtx(ctx)
```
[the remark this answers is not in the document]
wy65701436: Agree with Copilot here — testing/lib/orm.FakeOrmer.BeginWithCtx returns nil, nil, so every existing test that goes through WithTransaction with the shared fake will now panic on a nil TxOrmer.

chlins (approved):
lgtm

wy65701436 (approved):
lgtm

pujitha24:
This is approved with green CI and no outstanding comments as far as I can tell — ready whenever you have a moment. Happy to rebase first if you'd like it freshened.

PR #23838, opened 2026-09-04
ci: Make go mod tidy work without generated swagger code so Dependabot can update Go modules

Vad1mo:

Every Dependabot `gomod` job on `main` and on `release-2.15.0` fails with "Dependabot can't resolve your Go dependency files", so no Go module can be bumped at all.

Dependabot runs `go mod tidy` on a plain checkout. On a plain checkout the packages under `src/server/v2.0/{models,restapi}` do not exist yet: they are go-swagger output and are git-ignored (`.gitignore` lines 57-58). `go mod tidy` then reports, for `models`, `restapi` and each of the 38 `restapi/operations/<tag>` packages that `src/server/v2.0/handler` imports:

```
go: github.com/goharbor/harbor/src/server/v2.0/handler imports
    github.com/goharbor/harbor/src/server/v2.0/restapi: module
    github.com/goharbor/harbor@latest found (v2.15.2+incompatible,
    replaced by ../), but does not contain package ...
```

Full log in Dependabot job `1544163229`. Dependabot matches that text and turns it into a fatal, unrecoverable resolution error.

The tidy run cannot be skipped. dependabot-core always runs `go mod tidy` for gomod updates and no longer passes the `-e` tolerance it used to, and the `replace github.com/goharbor/harbor => ../` directive is not stubbed out because the stubber only replaces relative paths that resolve outside the repository. There is no `dependabot.yml` option to disable it.

This is why #23093, #23079 and #22927 had to be closed. The updates themselves were fine; Dependabot simply could not rebase or recreate the branches.

## The change

* One `zz_generated_placeholder.go` per generated package, carrying only the package clause and a comment. go-swagger never emits a file of that name, so code generation leaves them alone.
* The placeholder in `restapi` additionally blank-imports the nine external packages the generated code uses. Without that anchor, tidy on a fresh checkout demotes `go-openapi/{loads,spec,validate}` to `// indirect` and prunes `go.sum`, because only generated code imports them directly.
* `.gitignore` still ignores everything generated, written level by level because git cannot re-include a file whose parent directory is excluded.
* The `swagger_generate_server` macro no longer wipes the two output directories wholesale. It now fails with an explicit message in two cases: code generation produced a package that has no committed placeholder, or a placeholder was left behind for a package go-swagger no longer generates. Adding or removing a swagger tag therefore cannot silently drift.
* `make go_check` gains a `tidy_check` step that runs `go mod tidy -diff` in `src/` before `gen_apis`, on a tree that still has no generated code. A future change to the import set of the generated code therefore cannot silently stale the blank imports in the anchor file. The flag is read-only, so it needs no git and writes nothing.
* `.golangci.yaml` stops flagging the underscores in the `scan_all`, `scan_data_export`, `system_cve_allowlist` and `project_metadata` package names. Those names come from the swagger tags. The generated files are skipped as generated code; the committed placeholders are not, so revive lints them.

## Verification

On a fresh checkout without generated code, `go mod tidy` and `go list ./...` both succeed and leave `go.mod` and `go.sum` untouched. A simulated Dependabot bump (`go get golang.org/x/text@latest && go mod tidy`) produces a correct minimal diff with no `go.sum` pruning.

With generated code present, `go mod tidy` is a no-op, `go build ./...` passes, `go vet ./...` is byte-identical to `main`, and `golangci-lint run ./...` reports 0 issues. After code generation `git status` is clean, so `mocks_check` does not fire.

Both new guards were confirmed to fail as intended: with a placeholder missing, with an orphaned placeholder, and with one blank import removed from the anchor file.

## release-2.15.0

`release-2.15.0` has the identical `.gitignore` entries, the identical `rm -rf` in the Makefile, and no tracked files under the two output directories, so it needs the same change. I verified that this commit cherry-picks onto `release-2.15.0` with no conflicts, that it generates the identical set of 41 packages and 38 operation tags, and that `tidy_check` and `go list ./...` then pass there too. Happy to open that PR once this one is agreed.

# Issue being fixed
No issue filed. The evidence is the Dependabot job logs on this repository and the closed pull requests linked above.

- [x] Label the PR as needed. `release-note/infra`


wy65701436 (approved):
lgtm

Vad1mo:
TBO I am not very happ with that soluton!

However the **only two things that work are the placeholder approach here or committing the generated code.**

@wy65701436 @stonezdj  @chlins @bupd 

The proper long-term fix is committing the generated code with a drift check.

Vad1mo:
Follow-up for the long-term direction: #23857 proposes committing the go-swagger output with a CI drift check (as cilium, rekor and others do), which would make the placeholder files and the blank-import anchor in this PR unnecessary. This PR is the immediate fix so Dependabot works again.

Vad1mo:
Amended: CI's golangci-lint 2.9.0 (newer revive than my local 2.8.0) flagged the `scanner` and `user` placeholders for colliding with standard library package names; the existing var-naming exclusion is now extended to that message for the placeholder files. Reproduced and re-verified locally with 2.9.0: 2 issues before, 0 after.

stonezdj (approved):

PR #23867, opened 2026-09-07
Fix unnecessary Docker Hub authentication during proxy-cache pulls

### LINKED ISSUE #23750: Proxy cache serves 404 when the Docker Hub login endpoint is rate limited
**Expected behavior**
Pulling through a Docker Hub proxy-cache project authenticated with a Docker Hub account on the Team tier, without any pull rate limits, succeeds and does not produce a rate-limit error.

**Actual behavior**
When multiple not-yet-cached images are requested (when renovate bumps our dependencies in the git repo, and CI tries to build it), `harbor-core` logs produce

```
[ERROR] [/server/middleware/repoproxy/proxy.go:162]: failed to proxy manifest, fallback to local,
request uri: /v2/cache/library/node/manifests/sha256:d32cdf619f63fe0471182d08996dd516c6275bb5fd31ae06e55a570bd9e1ad43,
error: login to dockerhub error: login to dockerhub error: {"detail": "Rate limit exceeded", "error": false}
```

and clients receive **`404 not found`** for images that exist upstream but are not yet in the proxy cache.

It seems to me that Harbor is calling Docker Hub's /login endpoint on every pull instead of caching credentials and refreshing them only when their TTL is nearing expiration.

1. Add a Docker Hub registry endpoint with credentials (password or PAT).
2. Create a proxy-cache project using that endpoint.
3. Concurrently pull a number of **not-yet-cached** multi-arch tags through the project — a few dozen parallel pulls, as a CI burst produces.
4. `harbor-core` logs `login to dockerhub error: {"detail": "Rate limit exceeded", "error": false}`, and clients get 404 for the uncached tags. Cached tags keep working.

**Versions:**

- harbor version: **2.15.1**
- docker engine version: 29.3.0 / containerd via GitLab Kubernetes executor
- helm chart version: 1.19.1

**Additional context:**

This issue began appearing after I upgraded Harbor from version 2.13.0 to 2.15.1.

tjohnson-bfunds:

Docker Hub adapter construction eagerly authenticated against `POST /v2/auth/token`. Because proxy-cache requests create short-lived adapters, concurrent image pulls could generate unnecessary Hub API login requests, trigger rate limiting, and cause uncached images to return not-found errors when Harbor fell back to local storage.

This change makes Docker Hub API authentication lazy. Proxy-cache pulls continue to authenticate through the native registry client, without requiring a separate Docker Hub API login.

This bug was introduced in https://github.com/goharbor/harbor/pull/23096
Reference issue - https://github.com/goharbor/harbor/issues/23750

## Implementation

- Defer Hub API authentication until an API operation requires it.
- Retain per-client token caching and the existing nine-minute refresh window.
- Return the cached token while holding the client mutex to avoid concurrent
  reads racing with token refresh.
- Preserve the configured registry HTTP client timeout.
- Reject successful authentication responses that omit an access token.
- Invalidate a token rejected with HTTP 401 so the next request can refresh it,
  without automatically replaying the failed request.
- Avoid invalidating a newer token when an older in-flight request returns 401.

The process-wide cache from the original revision has been removed.
This PR does not introduce cross-client token sharing or a general
authentication retry/backoff policy.

## Regression coverage

Tests cover:

- Concurrent adapter construction without Hub API login requests.
- Authenticated native manifest access while Hub API login is unavailable.
- Rejection of invalid registry credentials.
- Lazy authentication, per-client token reuse, expiry, and concurrent refresh.
- Anonymous access and incomplete credentials.
- Authentication errors, malformed responses, and empty tokens.
- HTTP 401 invalidation without replay, including late responses for older tokens.
- Authentication timeout and subsequent recovery.

## Validation

Passed locally:

```bash
go test -race -count=1 ./pkg/reg/adapter/dockerhub ./pkg/reg/adapter/native ./pkg/registry/...
go test -race -count=20 ./pkg/reg/adapter/dockerhub
go vet ./pkg/reg/adapter/dockerhub
git diff --check
```

The constructor regression tests fail against unpatched main. The concurrent
refresh test also fails when the refresh lock is removed.

This PR targets `main`; a subsequent backport to `release-2.15.0` is intended.

# Issue being fixed

Fixes #23750

- [ ] Label the PR as needed: `release-note/update`
- [x] Accepted the DCO. The commit includes a `Signed-off-by` trailer.
- [x] Made sure tests are passing locally and added regression coverage.
- [x] Considered the docs impact. No configuration or public API changes require documentation updates.

### DIFF
--- a/src/pkg/reg/adapter/dockerhub/adapter_test.go
+++ b/src/pkg/reg/adapter/dockerhub/adapter_test.go
@@ -2,13 +2,19 @@ package dockerhub
 
 import (
 	"fmt"
+	"io"
 	"net/http"
+	"net/http/httptest"
+	"sync"
+	"sync/atomic"
 	"testing"
 
+	"github.com/opencontainers/go-digest"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"gopkg.in/h2non/gock.v1"
 
+	"github.com/goharbor/harbor/src/pkg/reg/adapter/native"
 	"github.com/goharbor/harbor/src/pkg/reg/model"
 )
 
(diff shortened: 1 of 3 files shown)


tjohnson-bfunds:
@Vad1mo - Could we get a review here please?

strigazi:
Overall LGTM, fixes the issue.

Issue #15649, opened 2021-09-23
core hang when a lot request coming at the same time


Jarvis-Zhou:
If you are reporting a problem, please make sure the following information are provided:

**Expected behavior and actual behavior:**
call delete or other api 500+ at once
expected: success
actual: core hang

- harbor version: 2.1.0

**Additional context:**

we find core create a transcaction in middleware (transaction.Middleware(dbTxSkippers...)). but in dao, we always use globalOrm to create / delete from database. this can case transaction in transaction. when db connect is not enough and a lot request comming at the same time, the core will hang as all the request go pass transaction.Middleware(dbTxSkippers...) and then not enough db connect to deal with the next dao request.

wy65701436:
We have fix several DB connection & transtion issues, you can try with the latest build with the scenario.

Jarvis-Zhou:
but in the code, i also can find some where that use WithTransaction after  transaction.Middleware(dbTxSkippers...) , will this still case transtion issue ?

Jarvis-Zhou:
why we need transaction.Middleware(dbTxSkippers...)  every requst ?

bitsf:
please wait and try release 2.3.3 or 2.4.0

heww:
Closed as this is fixed in harbor 2.3.3+

PR #23880, opened 2026-09-09
fix: reuse request context in audit-log user name resolver

### LINKED ISSUE #23879: core deadlocks on its own DB pool: audit-log resolver takes a second connection inside transaction middleware (bulk user delete wedges core)

Expected: a burst of concurrent write requests (here `DELETE /api/v2.0/users/{id}` fired in parallel by the portal's bulk user delete) completes, or fails with an error, and core keeps serving.

Actual: harbor-core deadlocks on its own database connection pool and stays wedged until the pods are restarted. Every request that touches the DB hangs forever; `/api/v2.0/ping` still answers 200, `/api/v2.0/health` times out (504 at the ingress), jobservice hook calls to core time out, harbor-exporter crash-loops on its startup probe. Postgres shows exactly `POSTGRESQL_MAX_OPEN_CONNS` sessions per core pod in state `idle in transaction`, last statement `begin`, no locks, no active queries. The database itself answers `select 1` in 40 ms.

Root cause: every non-GET request runs inside `transaction.Middleware` (`src/core/middlewares/middlewares.go:97`), which takes one pool connection with `begin` and stores the tx ormer in the request context. Inside that transaction, before the handler runs, `log.Middleware` calls the audit-log resolver's `PreCheck`, and `src/pkg/auditext/event/user/user.go:60-61` does:

```go
// use different context to so that the user is visible before the transaction is committed
user, err := pkgUser.Mgr.Get(orm.Context(), int(id))
```

`orm.Context()` builds a fresh non-transactional ormer, so this query needs a **second** connection from the same `database/sql` pool. With N = `POSTGRESQL_MAX_OPEN_CONNS` concurrent such requests on one pod, all N connections are held by `begin` and all N goroutines wait in `database/sql.(*DB).conn` for a connection that will never be released. Beego v2.3.10 hardcodes `context.Background()` into `Raw`/`Read`/`Insert`/`Begin`, so no request cancellation or deadline reaches the pool wait, and Harbor sets no acquisition timeout anywhere.

The comment's justification is backwards: `PreCheck` runs before the handler deletes the row, and a transaction sees its own state plus everything committed, so the request ctx would return the same user with zero extra connections.

Cascade: one config loader (`security.Middleware` → `config.AuthMode` → `FetchOrSave` under `lib/cache.keyMutex`) also ended up waiting for a pool connection while holding the key mutex, so 1300+ further goroutines, including GETs and the health checker, parked on `keyMutex.Lock`. That is why even read-only requests hung.

Goroutine dump excerpt (25 identical per pod):

```
goroutine 8080354 [select, 164 minutes]:
database/sql.(*DB).conn
github.com/beego/beego/v2/client/orm.(*DB).QueryContext
github.com/beego/beego/v2/client/orm.(*querySet).All
github.com/goharbor/harbor/src/pkg/user/dao.(*dao).List
github.com/goharbor/harbor/src/pkg/user.(*manager).Get
github.com/goharbor/harbor/src/pkg/auditext/event/user.userIDToName
github.com/goharbor/harbor/src/pkg/auditext/event.(*Resolver).PreCheck
github.com/goharbor/harbor/src/controller/event/metadata/commonevent.(*Metadata).PreCheckMetadata
github.com/goharbor/harbor/src/server/middleware/log.Middleware...
github.com/goharbor/harbor/src/server/middleware/security.Middleware...
github.com/goharbor/harbor/src/lib/orm.WithTransaction.func1
github.com/goharbor/harbor/src/server/middleware/transaction.Middleware...
```

`pg_stat_activity` at the time (2 core pods, `POSTGRESQL_MAX_OPEN_CONNS=25`):

```
 client_addr   |        state        | count | max_xact_age | last_query
 10.10.18.216  | idle in transaction |    25 | 02:42:00     | begin
 10.10.179.252 | idle in transaction |    25 | 02:42:00     | begin
```

**Same pattern elsewhere (verified in v2.15.2 and current `main`):**

| Site | Route that reaches it | Notes |
|---|---|---|
| `src/pkg/oidc/helper.go` `populateGroupsDB` (`orm.Context()` twice) | every non-GET API call with an OIDC token, OIDC-CLI basic auth on `/v2/**` writes (docker push), `POST /c/oidc/onboard` | reached from `security.Middleware` inside the tx; `usergroup.Mgr.Populate` → `dao.ReadOrCreate` → `orm.WithTransaction` opens a real second transaction on the second connection |
| `src/pkg/authproxy/http.go:116` | non-GET `/v2/**` under `http_auth`, `POST /c/login` | same shape as OIDC |
| `src/core/api/internal.go:81-86` | `POST /api/internal/syncquota` | `cfgMgr.Save` on a non-tx ormer while the request tx is open |
| `src/pkg/auditext/event/member/member.go:222` `ensureORMContext` (main only, from #23586) | `DELETE /api/v2.0/projects/{p}/members/{id}` | explicitly detects the tx ormer and swaps in a fresh `beegoorm.NewOrm()` |

Every other `orm.Context()` / `orm.Copy()` call site runs in a goroutine, at startup, or in jobservice, and does not hold a request transaction. Nested `orm.WithTransaction` uses SAVEPOINT on the same connection and is fine.

Prior art: #15649 (2021) describes this exact mechanism at a different call site ("core hang when a lot request coming at the same time ... not enough db connect to deal with the next dao request") and was closed as fixed in 2.3.3. #21455 reintroduced it in the audit path. #21062 / #23114 / #23703 / #23813 are related pool-exhaustion reports but each involves a single connection held too long, not one request needing two connections.

1. Harbor v2.15.2, one core replica, `POSTGRESQL_MAX_OPEN_CONNS=25` (default in the helm chart is 100; lower values make it trivial, higher values just need more concurrency).
2. Create 30 users.
3. Log in as admin, select all users on the Users page, delete. The portal issues one `DELETE /api/v2.0/users/{id}` per user in parallel. Equivalent: `seq ... | xargs -P 30 -I{} curl -u admin:... -X DELETE https://harbor/api/v2.0/users/{}`.
4. Core stops responding to anything except `/api/v2.0/ping`. `select state, query from pg_stat_activity where datname='registry'` shows 25 × `idle in transaction` / `begin`. It never recovers without a restart.

**Versions:**

- harbor version: v2.15.2 (code identical on `main` as of 2026-09-09)
- deployment: helm on EKS, 2 core replicas, external Aurora PostgreSQL
- docker engine version: n/a (API-level trigger)

**Additional context:**

Suggested fix direction, not implemented:

1. Give `IDToNameFunc` a `ctx` parameter and pass the request context through `Resolver.PreCheck`; same for `ensureORMContext`, `populateGroupsDB` and `authproxy`. A request must never acquire more than one pool connection.
2. Bound pool acquisition (a pool-level timeout, since beego's non-`WithCtx` methods swallow the request ctx) so a future regression of this class turns into 5xx errors instead of a permanent wedge.
3. Optionally a lint rule forbidding `orm.Context()` under `src/server`, `src/pkg/auditext`, `src/pkg/oidc`, `src/pkg/authproxy`.

KR-Ravindra:

## Problem

A burst of concurrent write requests, for example the portal's bulk user delete firing one `DELETE /api/v2.0/users/{id}` per user in parallel, wedges harbor-core. Every request that touches the database hangs, `/api/v2.0/health` times out, and Postgres shows exactly `POSTGRESQL_MAX_OPEN_CONNS` sessions per core pod in `idle in transaction` with `begin` as the last statement. Core never recovers without a restart.

## Root cause

Every non-GET request runs inside `transaction.Middleware` (`src/core/middlewares/middlewares.go`), which takes one connection from the `database/sql` pool with `begin` and stores the transaction ormer in the request context. Before the handler runs, `log.Middleware` calls the audit-log resolver's `PreCheck`, and for `DELETE` on a user that reaches `userIDToName` in `src/pkg/auditext/event/user/user.go:60-61`:

```go
// use different context to so that the user is visible before the transaction is committed
user, err := pkgUser.Mgr.Get(orm.Context(), int(id))
```

`orm.Context()` builds a fresh, non-transactional ormer, so this lookup needs a second connection from the same pool while the first one is held. With N = `POSTGRESQL_MAX_OPEN_CONNS` concurrent requests all N connections are held by `begin` and all N goroutines wait in `database/sql.(*DB).conn` for a connection that will never be released. beego hardcodes `context.Background()` into the query path, so no request deadline reaches the wait.

The comment's reasoning does not hold: `PreCheck` runs before the handler deletes the row, and a transaction sees its own state plus everything committed, so the request context returns the same user with no extra connection.

## Fix

- `src/pkg/auditext/event/basic.go`: `ResolveIDToNameFunc` now takes a `context.Context`. `PreCheck` passes its own (request) context through. `Resolve` keeps passing a fresh `orm.Context()`: it is executed by `notification.Middleware`, which wraps `transaction.Middleware`, in a detached goroutine after the request transaction has committed or rolled back, so the ormer in `ce.Ctx` cannot be reused there. Behaviour on the create/update path is unchanged.
- `src/pkg/auditext/event/user/user.go`: `userIDToName` uses the context it is given and the misleading comment is removed.

A request now never needs more than one pool connection on this path.

The issue lists further sites with the same shape outside the audit-log package (`src/pkg/oidc/helper.go`, `src/pkg/authproxy/http.go`, `src/core/api/internal.go`) and the `ensureORMContext` swap in the member resolver on `main`. Those are left out of this change to keep it reviewable; happy to follow up on them separately.

## How tested

Unit tests only, no database needed.

`TestPreCheckDeleteReusesRequestContext` (`src/pkg/auditext/event/user/user_resolve_test.go`) installs a mock `pkgUser.Mgr`, calls `PreCheck` with a context carrying a known ormer, and asserts the lookup receives that same ormer. Before the fix it fails because `userIDToName` builds its own ormer instead (the beego panic is what a fresh ormer looks like in a unit test with no registered database):

```
--- FAIL: TestPreCheckDeleteReusesRequestContext (0.00s)
panic: <Ormer.Using> unknown db alias name `default` [recovered, repanicked]
...
github.com/goharbor/harbor/src/lib/orm.Context()
	src/lib/orm/orm.go:92 +0x25
github.com/goharbor/harbor/src/pkg/auditext/event/user.userIDToName({0xd957d3, 0x1})
	src/pkg/auditext/event/user/user.go:61 +0x45
github.com/goharbor/harbor/src/pkg/auditext/event.(*Resolver).PreCheck(...)
	src/pkg/auditext/event/basic.go:67 +0x1be
FAIL	github.com/goharbor/harbor/src/pkg/auditext/event/user	0.016s
```

After the fix:

```
ok  	github.com/goharbor/harbor/src/pkg/auditext/event	0.014s
ok  	github.com/goharbor/harbor/src/pkg/auditext/event/config	0.015s
ok  	github.com/goharbor/harbor/src/pkg/auditext/event/login	0.016s
ok  	github.com/goharbor/harbor/src/pkg/auditext/event/member	0.019s
ok  	github.com/goharbor/harbor/src/pkg/auditext/event/project	0.017s
ok  	github.com/goharbor/harbor/src/pkg/auditext/event/user	0.014s
```

`TestEventResolver_PreCheckPassesContext` (`basic_test.go`) additionally checks that the base resolver forwards the context it is given to `IDToNameFunc`. `gofmt`, `go vet` and `golangci-lint run ./pkg/auditext/...` are clean; `go build` of `./pkg/... ./controller/... ./server/middleware/... ./lib/... ./common/...` passes.

## Links

- https://github.com/goharbor/harbor/issues/23879
- https://github.com/goharbor/harbor/issues/15649 (same mechanism at a different call site, 2021)

# Issue being fixed
Fixes #23879

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.

### DIFF
--- a/src/pkg/auditext/event/basic.go
+++ b/src/pkg/auditext/event/basic.go
@@ -27,6 +27,7 @@ import (
 	"github.com/goharbor/harbor/src/controller/event/metadata/commonevent"
 	"github.com/goharbor/harbor/src/controller/event/model"
 	"github.com/goharbor/harbor/src/lib/config"
+	"github.com/goharbor/harbor/src/lib/orm"
 	"github.com/goharbor/harbor/src/pkg/notifier/event"
 )
 
@@ -37,7 +38,7 @@ const (
 )
 
 // ResolveIDToNameFunc is the function to resolve the resource name from resource id
-type ResolveIDToNameFunc func(string) string
+type ResolveIDToNameFunc func(ctx context.Context, id string) string
 
 type Resolver struct {
 	ResourceType string
@@ -64,7 +65,7 @@ func (e *Resolver) PreCheck(ctx context.Context, url string, method string) (cap
 		re := regexp.MustCompile(e.ResourceIDPattern)
 		m := re.FindStringSubmatch(url)
 		if len(m) >= 2 && e.IDToNameFunc != nil {
-			resName = e.IDToNameFunc(m[1])
+			resName = e.IDToNameFunc(ctx, m[1])
 		}
 	}
 	return true, resName
@@ -99,7 +100,9 @@ func (e *Resolver) Resolve(ce *commonevent.Metadata, event *event.Event) error {
 			}
 			evt.ResourceName = m[1]
 			if e.IDToNameFunc != nil {
-				resourceName = e.IDToNameFunc(m[1])
+				// Resolve runs asynchronously after the request transaction has completed, so the
+				// ormer in ce.Ctx cannot be reused here; only PreCheck runs inside the request
+				resourceName = e.IDToNameFunc(orm.Context(), m[1])
 			}
 		}
 		if e.ShouldResolveName && resourceName != "" {
@@ -125,7 +128,7 @@ func (e *Resolver) Resolve(ce *commonevent.Metadata, event *event.Event) error {
 		}
 		evt.ResourceName = m[1]
 		if e.IDToNameFunc != nil {
-			resourceName = e.IDToNameFunc(m[1])
+			resourceName = e.IDToNameFunc(orm.Context(), m[1])
 		}
 		if e.ShouldResolveName && resourceName != "" {
 			evt.ResourceName = resourceName
--- a/src/pkg/auditext/event/basic_test.go
+++ b/src/pkg/auditext/event/basic_test.go
@@ -40,8 +40,8 @@ func TestEventResolver_PreCheck(t *testing.T) {
 		wantCapture      bool
 		wantResourceName string
 	}{
-		{"test normal", fields{ResourceIDPattern: `/api/v2.0/tests/(\d+)`, ResourceType: "test", SucceedCodes: []int{200}, ShouldResolveName: true, IDToNameFunc: func(string) string { return "test" }}, args{context.Background(), "/api/v2.0/tests/123", "DELETE"}, true, "test"},
-		{"test resource name", fields{ResourceIDPattern: `/api/v2.0/tests/(\d+)`, ResourceType: "test", SucceedCodes: []int{200}, ShouldResolveName: true, IDToNameFunc: func(string) string { return "test_resource_name" }}, args{context.Background(), "/api/v2.0/tests/234", "DELETE"}, true, "test_resource_name"},
+		{"test normal", fields{ResourceIDPattern: `/api/v2.0/tests/(\d+)`, ResourceType: "test", SucceedCodes: []int{200}, ShouldResolveName: true, IDToNameFunc: func(context.Context, string) string { return "test" }}, args{context.Background(), "/api/v2.0/tests/123", "DELETE"}, true, "test"},
+		{"test resource name", fields{ResourceIDPattern: `/api/v2.0/tests/(\d+)`, ResourceType: "test", SucceedCodes: []int{200}, ShouldResolveName: true, IDToNameFunc: func(context.Context, string) string { return "test_resource_name" }}, args{context.Background(), "/api/v2.0/tests/234", "DELETE"}, true, "test_resource_name"},
 		{"test no resource name", fields{ResourceIDPattern: `/api/v2.0/tests/(\d+)`, ResourceType: "test", SucceedCodes: []int{200}, ShouldResolveName: true}, args{context.Background(), "/api/v2.0/tests/234", "GET"}, true, ""},
 	}
 	for _, tt := range tests {
@@ -62,5 +62,26 @@ func TestEventResolver_PreCheck(t *testing.T) {
 				t.Errorf("EventResolver.PreCheck() gotResourceName = %v, want %v", gotResourceName, tt.wantResourceName)
 			}
 		})
+	}
+}
+
+func TestEventResolver_PreCheckPassesContext(t *testing.T) {
+	type ctxKey struct{}
+	ctx := context.WithValue(context.Background(), ctxKey{}, "request")
+	var got context.Context
+	e := &Resolver{
+		ResourceIDPattern: `/api/v2.0/tests/(\d+)`,
+		ResourceType:      "test",
+		ShouldResolveName: true,
+		IDToNameFunc: func(c context.Context, _ string) string {
+			got = c
+			return "test"
+		},
+	}
+	if _, gotResourceName := e.PreCheck(ctx, "/api/v2.0/tests/123", "DELETE"); gotResourceName != "test" {
+		t.Errorf("EventResolver.PreCheck() gotResourceName = %v, want %v", gotResourceName, "test")
+	}
+	if got == nil || got.Value(ctxKey{}) != "request" {
+		t.Errorf("EventResolver.PreCheck() did not pass the request context to IDToNameFunc")
 	}
 }
--- a/src/pkg/auditext/event/user/user.go
+++ b/src/pkg/auditext/event/user/user.go
@@ -15,6 +15,7 @@
 package user // nolint:revive
 
 import (
+	"context"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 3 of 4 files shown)


KR-Ravindra:
Round 1 self-review.

1. `IDToNameFunc` is only set by the user resolver (`src/pkg/auditext/event/user/user.go`), so the signature change has no other callers; the member resolver already goes through `ensureORMContext`.
2. The ordering claim holds: `notification.Middleware` is registered ahead of `transaction.Middleware` in `src/core/middlewares/middlewares.go`, and `event.BuildAndPublish` runs `Resolve` in a detached goroutine after the handler returns, so keeping `orm.Context()` on the async path and using the request context only in `PreCheck` is the right split.
3. `go test ./pkg/auditext/event/...` passes at 6fb57ac (all six packages), `go vet` clean.
4. `Check release-note label set` needs a `release-note/*` label from a maintainer. UTTEST and the API tests are still running; I will mark this ready once they are green.

KR-Ravindra:
Round 2 self-review.

1. UTTEST, UI_UT, APITEST_DB / _LDAP / _PROXY_CACHE, OFFLINE, CodeQL and DCO are green on 6fb57ac.
2. The two remaining reds are gates I cannot clear from the branch: `Check release-note label set` needs a `release-note/*` label, and codecov/patch flags the two `Resolve()` call sites, which keep the existing `orm.Context()` behaviour and are only touched for the signature change.
3. No other PR references #23879.

Marking ready for review.

KR-Ravindra:
Rebased onto current main (`19103c1`) — the branch was 28 commits behind.

On the two red checks, both need a maintainer rather than a push:

- **Check release-note label set** — needs a `release-note/...` label applied, which I cannot set.
- **codecov/patch** — the uncovered lines are the two `orm.Context()` calls in `Resolve`. I tried to cover them with a unit test and could not: `orm.Context()` calls `orm.NewOrm()`, which panics with `unknown db alias name `default`` unless a database is registered, so exercising that path needs the DB-backed suite rather than a plain unit test. `PreCheck` is covered by `TestEventResolver_PreCheckPassesContext`, which is the half that can be tested without a database.

The change itself is unchanged: `PreCheck` runs inside the request and keeps the request context, while `Resolve` runs after the transaction has completed and so takes a fresh ormer.

Drafted with AI assistance and checked against the code before posting.

Issue #19230, opened 2023-08-22
Install Harbor with Podman only (not docker) manually


dioguinhocst:
I'm in the middle of graduate work and trying to install Harbor. I first installed it only with Docker, but the goal is to install it with Podman and only. The installation scripts check if Docker is present, so it's not possible (I tried by putting an alias). I also tried to manually re-execute the docker-compose that the script generated for me on my Podman machine. Please tell me what to do because I'm working on my degree please.

dioguinhocst:
I'm on the 5th attempt, I've already managed to adapt the prepare script a few times and it's currently returning this error, even if I create it manually

```bash
admin@machine-production:~/harbor$ ./prepare 
prepare base dir is set to /home/admin/harbor
Trying to pull docker.io/goharbor/prepare:v2.8.3...
Getting image source signatures
Copying blob 9e054aa5e0c4 done  
Copying blob 64766fbe86f3 done  
Copying blob 5e8573822658 done  
Copying blob b325f953ccaa done  
Copying blob 4a61406f052d done  
Copying blob 35af786c7219 done  
Copying blob ac0b5dc3429a done  
Copying blob c1fcbc663df3 done  
Copying blob 35ef2e38ddee done  
Copying blob 2a411a6b181f done  
Copying config 4e3cddf54a done  
Writing manifest to image destination
Storing signatures
Error: statfs /home/admin/harbor/common/config: no such file or directory
admin@machine-production:~/harbor$ cat prepare 
#!/bin/bash
set -e

# If compiling source code this dir is harbor's make dir.
# If installing harbor via package, this dir is harbor's root dir.
if [[ -n "$HARBOR_BUNDLE_DIR" ]]; then
    harbor_prepare_path=$HARBOR_BUNDLE_DIR
else
    harbor_prepare_path="$( cd "$(dirname "$0")" ; pwd -P )"
fi
echo "prepare base dir is set to ${harbor_prepare_path}"

# Clean up input dir
rm -rf ${harbor_prepare_path}/input
# Create a input dirs
mkdir -p ${harbor_prepare_path}/input
input_dir=${harbor_prepare_path}/input

# Copy harbor.yml to input dir
if [[ ! "$1" =~ ^\-\- ]] && [ -f "$1" ]
then
    cp $1 $input_dir/harbor.yml
    shift
else
    if [ -f "${harbor_prepare_path}/harbor.yml" ];then
[...]

Issue #28, opened 2016-03-21
Send email failed when certificate does not match with smtp server.


tuxknight:
```
# cat harbor.cfg|grep email_server
email_server = smtp.xxxxxx.com
email_server_port = 25
```

``Mar 21 11:12:36 registry.idc docker/ui[15559]: 2016/03/21 03:12:36 [password.go:169][E] send email failed: x509: certificate is valid for *.icoremail.net, icoremail.net, not smtp.xxxxxx.com ``

It is a common situation that serve name does not match with the domain name in certificate when using an email vendor. So, it is better to make smtp TLS configurable.

ywk253100:
Maybe the email vendor should fix the problem of mismatching the domain name in certificate and the server name, or how to guarantee the security of communications?

hainingzhang:
@tuxknight, what is your suggestion for the TLS configuration under this scenario?

tuxknight:
@ywk253100  Good point. It will be better to assign one certificate each domain name. 
But in this scenario, a smtp server such as  *smtp.xxxxxxx.com* has a cname like *smtp.icoremail.net* which is the real server of the vendor. And xxxxxxx.com could be  a customer's domain name.

```
dig smtp.xxxxxxx.com
;; ANSWER SECTION:
smtp.xxxxxxx.com. 600	IN	CNAME	smtp.icoremail.net.
smtp.icoremail.net.	600	IN	CNAME	mail1.icoremail.net.
mail1.icoremail.net.	600	IN	A	223.252.214.64
mail1.icoremail.net.	600	IN	A	223.252.214.65
```

@hainingzhang  Perhaps It's good to make smtp TLS configurable. Or you could just add another configuration item called "domain" which used to match the certificate. In my scenario, it would be "icoremail.net".

Thanks.

reasonerjt:
@tuxknight , thanks for elaborating.
So given it's a cname, I suppose you can simply configure smtp.icoremail.net as email_server in harbor.cfg?

ywk253100:
@tuxknight  Does setting cname work? If it works, I will close this issue.

xiaods:
@ywk253100 please add label to category the issue's type. it's will fine.

tuxknight:
It works while using smtp.icoremail.net.

Harbor is online and I will keep migrating projects to harbor :)

hainingzhang:
@tuxknight Did you run Harbor in a private environment? We'd like to learn more about your use case if possible. Thanks.

PR #15986, opened 2021-11-10
Initial support for podman + docker-compose

eb4x:
Some minor cleanups and support for podman/docker-compose combination,

https://www.redhat.com/sysadmin/podman-docker-compose

PR #23881, opened 2026-09-09
feat: support Podman and Podman Compose in the installer

eugene-harold-krabs:
## Summary

Add native Podman support to Harbor's existing installer, preserving the Docker path. This is a concrete implementation for the installation problem in #19230; it does not change the Kubernetes deployment path or require Docker compatibility wrappers.

- Detect Docker or Podman, with explicit `CONTAINER_RUNTIME=podman` selection.
- Select Podman Compose for Podman and use the selected runtime for image loading and preparation.
- Prepare the configuration directory before starting the preparation container.
- Give generated long-form bind mounts shared SELinux labels, matching the existing short-form `:z` mounts.
- Add Docker/Podman installer regression checks to the existing installer CI entry point and document usage and boundaries.

## Why SELinux `z` is needed

On an SELinux-enforcing host, ordinary host file permissions are not sufficient for container access: bind-mounted files also need an appropriate SELinux context. Harbor's Compose template already uses `:z` on its short-form mounts, but its long-form bind mounts did not request relabeling. `bind.selinux: z` supplies the equivalent behavior for those mounts.

The lowercase **`z` is shared**, whereas uppercase **`Z` is private**. Multiple Harbor services mount the same paths, notably `common/config/shared/trust-certificates`. If containers have different SELinux process labels, privately relabeling one shared source for a later container can deny the earlier container access. Shared labels avoid that conflict. Files still need the normal mount and Unix permissions; shared labeling does not provide per-container SELinux isolation for the relabeled contents, so the installation should use dedicated Harbor directories.

The short-lived **prepare container is a separate case**. It is already privileged and bind-mounts the host root at `/hostfs` to resolve configured certificate and data paths. Recursively applying `z` or `Z` to that mount could relabel the host filesystem. The Podman invocation therefore uses `--security-opt label=disable` for that container only, without disabling host SELinux or adding a host-root relabel option.

Reference: [Podman volume labeling documentation](https://docs.podman.io/en/latest/markdown/podman-run.1.html#labeling-volume-mounts).

## Validation

- `bash tests/ci/installer_runtime_test.sh` passes after rebasing onto current `main`: mocked Docker and Podman install/prepare/start flows, preparation cleanup, and relabeling checks for all 36 long-form bind mounts.
- Bash syntax and `git diff --check` pass.
- The prior live EL10/Podman validation passed in [infrastructure pipeline 2823882005](https://gitlab.com/xrow-public/infrastructure/-/pipelines/2823882005), including the fork prepare-image build, installation, Harbor integration/client checks, and review cleanup. The validation harness is in [merged infrastructure MR !37](https://gitlab.com/xrow-public/infrastructure/-/merge_requests/37).
- That live run used fork source `a6a778da4723cc2fb5f320018fb5f9796da8c213` with the prepare image/version pinned by the harness. The installer, prepare source/templates, and installer regression files are byte-identical after this rebase; the additional change is documentation. The full live run has not been repeated on the rebased repository SHA.
- Rootless Podman installation is not validated. Upstream CI results will be reported by this PR; the local mocked checks are not a claim that the complete upstream suite has passed.

## Tracking and checklist

Related to #19230. Owner-approved contribution tracked by [infrastructure work item #28](https://gitlab.com/xrow-public/infrastructure/-/work_items/28).

- [x] Clear title and summary
- [x] DCO sign-off on each commit
- [x] Focused regression coverage and existing live validation recorded above
- [x] Documentation impact addressed in the README
- [ ] Suggested release-note label: `release-note/new-feature` (requires a maintainer)

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Harbor is hosted by the [Cloud Native Computing Foundation](https://cncf.io) (CN
 * **Graphical user portal**: User can easily browse, search repositories and manage projects.
 * **Auditing**: All the operations to the repositories are tracked through logs.
 * **RESTful API**: RESTful APIs are provided to facilitate administrative operations, and are easy to use for integration with external systems. An embedded Swagger UI is available for exploring and testing the API.
-* **Easy deployment**: Harbor can be deployed via Docker compose as well Helm Chart, and a Harbor Operator was added recently as well.
+* **Easy deployment**: Harbor can be deployed via Docker Compose or Podman Compose, as well as with the Helm chart or Harbor Operator.
 
 ## Architecture
 
@@ -53,7 +53,7 @@ For learning the architecture design of Harbor, check the document [Architecture
 ## Install & Run
 **System requirements:**
 
-**On a Linux host:** docker 20.10.10-ce+ and docker-compose 1.18.0+.
+**On a Linux host:** Docker 20.10.10-ce+ with Docker Compose 1.18.0+, or Podman with Podman Compose.
 
 Download binaries of **[Harbor release ](https://github.com/goharbor/harbor/releases)** and follow **[Installation & Configuration Guide](https://goharbor.io/docs/latest/install-config/)** to install Harbor.
 
--- a/README.md
+++ b/README.md
@@ -60,6 +60,27 @@ Download binaries of **[Harbor release ](https://github.com/goharbor/harbor/rele
 If you want to deploy Harbor on Kubernetes, please use the **[Harbor chart](https://github.com/goharbor/harbor-helm)**.
 
 Refer to the **[documentation](https://goharbor.io/docs/)** for more details on how to use Harbor.
+
+### Installing with Podman
+
+The installer detects Docker first, then Podman. To select Podman explicitly, run
+`sudo CONTAINER_RUNTIME=podman ./install.sh` (add `--with-trivy` to enable scanning).
+Install `podman-compose` first; the installer also accepts a configured `podman compose` provider.
+This uses Podman directly, without a Docker compatibility wrapper. Rootless installation is not validated.
+
+On SELinux hosts, generated bind mounts use the shared `z` label, including the
+long-form Compose equivalent `bind.selinux: z`. Harbor services share files such as
+`common/config/shared/trust-certificates`; a private `Z` label can prevent another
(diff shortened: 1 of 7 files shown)


xrow:
Hi, I am the owner of @eugene-harold-krabs. I already did a code review. Please let us know.

xrow:
Hi @stonezdj, @wy65701436 , @OrlinVasilev ,

I just wanted to poke you to move on. What do you think?

xrow:
@eugene-harold-krabs Did yo follow up on all the things REF https://github.com/goharbor/harbor/blob/main/CONTRIBUTING.md#design-new-features?

StripedMonkey:
It's unclear how this PR fixes one of the biggest (still) outstanding problems with a previous PR #15986 where podman does not support the `syslog` log driver. This is still an open request at https://github.com/podman-container-tools/podman/issues/28779 so there's no support for it yet so far as I can tell, though it may be right around the corner if the right people get interested.

Not to be the peanut gallery but this is a great example of why AI generated PRs generate unnecessary noise. I'm personally wanting podman support but this doesn't add that. Validating these changes locally should reveal it still won't work even with the unrelated selinux support changes.

xrow:
@StripedMonkey Thanks for you valuable feedback. I need to think about it more. From your expertise do you think journald support can help harbor?

As a side note I need to run it on any Red Hat Stack but OCP. I do what ever it takes. If the podman way is not a proper way to go, I will find out.

Also I will also try to ship real proof for review with this PR.

StripedMonkey:
I don't know enough about the harbor stack in order to know if journald support is the answer. I tried getting harbor working with podman and discovered the syslog issue first. After realizing that the logs were being aggregated into a single container for forwarding (there's a spot in the web interface for the location to forward logs) I determined that it would be more than a simple config change to resolve the syslog dependency and maintain feature parity.

If you're legitimately motivated I believe the solution might be to push for syslog log driver support in podman. Not only would that resolve the issue here, but it would fix the issue for other projects as well. Based on my cursory research it appears to be possible to plumb up, though it may require changes in projects upstream to podman.

xrow:
@StripedMonkey Short off... The lag of syslog support in podman might be journald being the only desired option in RHEL for certain reasons. I return in the next weeks.

PR #23913, opened 2026-09-12
fix(cache): avoid double prefix when removing expired entries

Hanabi9248:
### Summary

With `cache.Prefix("prefix:")`, reading an expired `key` calls `Delete` with `prefix:key`. `Delete` adds the prefix again, leaving the expired entry in storage and deleting the unrelated live key `prefix:key`, if present.

Pass the original key to `Delete` in both `Contains` and `Fetch`. The regression covers both paths, checking that the expired entry is removed and the other value survives. Both cases fail before the fix.

### Validation

- `go test ./lib/cache/memory ./lib/cache -count=1`
- `go test -race ./lib/cache/memory ./lib/cache -count=1`
- `go vet ./lib/cache/memory ./lib/cache`

All pass on Windows with Go 1.27.1.

- [x] Title and summary describe the change.
- [x] Commit is signed off under the DCO.
- [x] Regression coverage added and affected package tests pass.
- [x] Docs impact considered; no public API change.
- [ ] Apply the appropriate release-note label (maintainer access required).

### DIFF
--- a/src/lib/cache/memory/memory.go
+++ b/src/lib/cache/memory/memory.go
@@ -51,7 +51,7 @@ func (c *Cache) Contains(ctx context.Context, key string) bool {
 	}
 
 	if e.(*entry).isExpirated() {
-		err := c.Delete(ctx, c.opts.Key(key))
+		err := c.Delete(ctx, key)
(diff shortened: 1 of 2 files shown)


chlins (approved):
lgtm

wy65701436 (approved):
lgtm

Issue #22010, opened 2025-05-19
Harbor 2.12.2 and 2.12.3 HTTP Login Issue


JCObject:
I am deploying Harbor locally, and the host IP address is 192.168.31.116.

In version 2.12.2, I can comment out the HTTPS-related configuration in the configuration file, and then successfully log in via HTTP using the default credentials (admin / Harbor12345) at http://192.168.31.116.

However, in version 2.12.3, even though I comment out the HTTPS section and the deployment completes successfully, I consistently receive the error message “用户名或者密码不正确” ("Username or password is incorrect") when attempting to log in via http://192.168.31.116.

Is this behavior a change introduced in the new version? Will every future version of Harbor require HTTPS configuration for access?

Sheng08:
I’m facing the same problem.

zer010gic:
Same here as well

pzc163:
Same error，for the locally deployed Harbor, after commenting out the HTTPS section in harbor.yml, when logging in with the default password harbor_admin_password: Harbor12345, an error is reported that the username or password is incorrect.

MinerYang:
Hi @JCObject ,
Is it a upgrade from harbor v2.12.2 to v2.12.3? Have you changed the admin password before upgrade?

JCObject:
> Hi [@JCObject](https://github.com/JCObject) , Is it a upgrade from harbor v2.12.2 to v2.12.3? Have you changed the admin password before upgrade?

No, it is not an upgrade. Both Harbor v2.12.2 and v2.12.3 were deployed from scratch, and no admin password changes were made prior to or during the deployment.

JCObject:
> Hi [@JCObject](https://github.com/JCObject) ,
> 
> Could you try to clear the browser cookie and try again? Or try using docker login to verify.

After adding "insecure-registries": ["192.168.31.116"] to /etc/docker/daemon.json, I am able to successfully log in using docker login.

However, even after clearing the browser cookies in Harbor v2.12.3, I still cannot log in to the Web UI at http://192.168.31.116.

MinerYang:
Hi all,

There's a regression changes upon our upstream library https://github.com/gorilla/csrf/compare/v1.7.2...v1.7.3, that a current http connection or a https with none 443 port will cause 403 error with invalid origin message when we login harbor via Web browser., We will fix it as soon as possible.

Best,
Miner

JCObject:
> Hi all,
> 
[...]

PR #23915, opened 2026-09-14
fix(csrf): replace gorilla/csrf with net/http.CrossOriginProtection

bupd:
`govulncheck` reports Harbor's own code as affected by CVE-2025-24358:

```
Vulnerability #1: GO-2025-3607
  Module: github.com/gorilla/csrf
    Found in: github.com/gorilla/csrf@v1.7.2
    Fixed in: github.com/gorilla/csrf@v1.7.3
    Example traces found:
      #1: server/middleware/csrf/csrf.go:89:34: csrf.Middleware calls csrf.csrf.ServeHTTP
```

That is symbol-level reachability, not a manifest match — `csrf.Middleware` is on
the request path for every session-authenticated request.

The fix is v1.7.3, and `src/go.mod` pins below it on purpose:

```go
// pin the github.com/gorilla/csrf to v1.7.2 because of issue https://github.com/goharbor/harbor/issues/22010
github.com/gorilla/csrf v1.7.2
```

#22010 → #23759 reverted the bump because v1.7.3 broke HTTP-only logins. So the
vulnerability cannot be resolved by upgrading, which is what this PR is about.

## Re-taking v1.7.3 does not work

I tested this rather than assuming it. A plain-HTTP Harbor was deployed from
v2.15.2 images behind the bundled nginx, with core rebuilt from source against
v1.7.3.

v1.7.3 presumes TLS unless the caller sets the plaintext context key, so the
first attempt wired `csrf.PlaintextHTTPRequest` to the external endpoint's
scheme. Login still failed, for a second and independent reason: `nginx.conf`
uses `proxy_set_header Host $host`, which **drops the port**, while `sameOrigin`
compares host *including* port. Same core image, only the `Host` header varied:

```
Origin is always http://192.168.0.5:8099
  Host: 192.168.0.5:8099  (what the browser sent) -> HTTP 200
  Host: 192.168.0.5       (what nginx forwards)   -> HTTP 403
```

And the plaintext signal is a single global boolean, so it cannot describe a
deployment reachable on several ingresses. With an `http` external endpoint and
users arriving over https:

```
gorilla v1.7.3 + plaintext signal, HTTPS ingresses:
  us.harbor.dev  403    eu.harbor.dev  403
  harbor.dev     403    harbor.com     403
```

Setting the endpoint to `https` simply inverts which half breaks.

## This change

`net/http.CrossOriginProtection`, added in Go 1.25 (`go.mod` is on 1.26), decides
on `Sec-Fetch-Site`, which browsers have sent since 2023. No configured endpoint,
`Host`, or scheme guess takes part in the decision:

```go
var protect = http.NewCrossOriginProtection()
```

Same deployment, same four hostnames, nothing configured:

```
Modern browser, HTTPS ingresses:  us./eu./harbor.dev/harbor.com  all 200
Modern browser, HTTP ingress:     us.harbor.dev                      200
Pre-2023 browser, Host preserved: us.harbor.dev, harbor.com          200
Non-browser client (docker/CLI):                                     200
Attacker: cross-site / same-site subdomain / old-browser foreign    403
```

It also refuses the request the token scheme allowed. A sibling subdomain is
same-site, so `SameSite=Strict` still sent `_gorilla_csrf`, and a subdomain can
write cookies on the parent domain — leaving both halves of the double-submit
attacker-controlled. Only the Origin check stood in the way, and CVE-2025-24358
is precisely that the check never ran. Against a live logged-in session:

```
cross-site form POST (classic CSRF)  -> 403
same-site subdomain POST             -> 403
cross-site + forged token header     -> 403
```

### One gap closed beyond the library default

`CrossOriginProtection` admits a request carrying neither `Sec-Fetch-Site` nor
`Origin`, reading it as non-browser traffic. `csrfSkipper` has already excused
the `/v2/`, `/api/` and `/service/` routes that non-browser clients use, so an
unsafe request arriving without either header is refused instead:

```go
if slices.Contains(safeMethods, req.Method) {
	return nil
}

if req.Header.Get("Sec-Fetch-Site") == "" && req.Header.Get("Origin") == "" {
	return errors.New("request carries neither Sec-Fetch-Site nor Origin, cannot confirm its origin")
}
```

This preserves the previous behaviour, which rejected any write presenting no
token. Verified not to catch legitimate traffic — safe methods, `docker`/CLI
clients on basic auth, and the registry v2 API are all unaffected.

## Result

`gorilla/csrf` and `gorilla/securecookie` leave `go.mod`. `govulncheck` goes from
one affected vulnerability to none, and stops reporting CVE-2025-47909 as well,
which affects every released version of `gorilla/csrf` with **no fix available**
and so could never have been resolved by upgrading.

Worth noting the Go vulnerability database entry for CVE-2025-47909 recommends
this exact migration.

## Follow-ups, deliberately not in this PR

`CSRF_KEY` is now unused, and the portal's `__csrf` / `X-Harbor-CSRF-Token`
interceptor is dead code. Both degrade quietly — the interceptor omits the header
when localStorage holds no token — so removing them is left for separate changes
rather than folded into a security fix.

## Testing

`csrf_test.go` is rewritten and no longer carries the `db` build tag, since the
middleware no longer reads configuration or the database. It covers the allow
set (including a second ingress no configuration knows about), the attack set
(including the same-site subdomain case), the skipper matrix, and the
neither-header rule.

- `go build ./...` clean
- `go test ./server/middleware/csrf/...` green
- `golangci-lint run ./server/middleware/csrf/...` — 0 issues
- `go mod tidy` clean
- End-to-end on a real plain-HTTP deployment: login, authenticated read, and a
  session-carrying `POST /api/v2.0/projects` returning 201


wy65701436:
cc @MinerYang please help to take a review, thanks.

stonezdj:
Nightly e2e pipeline(ldap_offline) passed with this PR.

stonezdj (approved):

Issue #16398, opened 2022-02-22
Make API Calls with Token/Robot Account to Harbor


AlexBarth13:
Dear Harbor-Community,

We are using the Harbor registry in our company and we noticed, that API calls (```GET```, ```PUT``` and ```POST```) are not working with ```unauthorized``` error message.
Our version is v2.4 and Harbor is secured by the company's OIDC and MFA solution.

During a test, were we wanted to create a robot account automatically we noticed this issue. To create the robot account, we have tried with the following credentials:
1. username:cli_secret
2. robot_account:secret
3. Solution 1 and Solution 2 from the FAQ: https://github.com/goharbor/harbor/wiki/Harbor-FAQs#api

Here is a example how we tried to call the API:

```
 curl -i -k -X 'POST' \
  'https://<FQDN>/api/v2.0/robots' \
  -v \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -H 'Authorization:  Bearer <TOKEN>' \
  -d '{
  "secret": "string",
  "disable": false,
  "name": "testing",
  "level": "system",
  "duration": -1,
  "description": "Created with API",
  "permissions": [
    {
      "access": [
        {
          "action": "create",
          "resource": "scan"
        }
      ],
      "kind": "project",
      "namespace": "*"
    }
  ]
}'
```

When we are using the credentials from 1. and 2. - either in plain text with ```curl -u``` or base64 encoded and as ```-H 'Authorization``` - we are facing this issue:
```
{"errors":[{"code":"UNAUTHORIZED","message":"unauthorized"}]}
* Connection #0 to host <host> left intact
```
We also get this message when we are creating the bearer token like it is mentioned in the FAQ solution 1.
When we tested the solution 2 in the FAQ, we are not seeing the needed ```_xsrf``` under ```Set-Cookies```.

Furthermore, we tested the creation of robot accounts from the Swagger UI and this is working fine.

Currently we were not able to find the correct documentation, how we have to authenticate against the API to use it.

How are we able to authenticate correctly and make API calls to Harbor? Is there a solution to this?

Thanks a lot!

Alexander Barth (alexander.barth@mercedes-benz.com) on behalf of Daimler TSS, [Provider Information](https://github.com/mercedes-benz/daimler-foss/blob/master/PROVIDER_INFORMATION.md)

YangJiao0817:
The robot account does not have permission to create a robot account, you should use the correct account.

Here is an example of calling the API:
 ```
[...]

Issue #16047, opened 2021-11-23
First login to harbor with oidc-auth and api/curl


dm0610:
Hello everyone.
Tell me, is there a real opportunity to make the first login to Harbor via oidс not through GUI, but through api? 
I easily get the id_token from the keycloak, but I can't log in. 

We are trying to automate getting of cli-tokens for technical accounts.
here i get token
`    private static String auth(script, String credentialsId){
        script.withCredentials([script.usernamePassword(credentialsId: credentialsId,
            usernameVariable: 'USER', passwordVariable: 'PASS'),
            script.usernamePassword(credentialsId: 'admin_user_id', usernameVariable: 'USER2', passwordVariable: 'PASS2')]) {
            def out = script.sh script: "curl -s --location --request POST 'https://sso.mycompany.com/auth/realms/my_company/protocol/openid-connect/token' \
            --header 'Content-Type: application/x-www-form-urlencoded' \
            --data-urlencode 'grant_type=password' \
            --data-urlencode 'client_id=$script.USER2' \
            --data-urlencode 'client_secret=$script.PASS2' \
            --data-urlencode 'scope=openid' \
            --data-urlencode 'response_type=id_token' \
            --data-urlencode 'username=$script.USER' \
            --data-urlencode 'password=$script.PASS'", returnStdout: true
            return script.readJSON(text: out).id_token
        }
    }`
After that I try to connect via api:
curl -v -H "Authorization:  Bearer $token" https://harbor.mycompany.com/api/v2.0/users/current
`{ [5 bytes data]
* Mark bundle as not supporting multiuse
< HTTP/1.1 401 Unauthorized
< Server: nginx
< Date: Tue, 23 Nov 2021 13:39:42 GMT
< Content-Type: application/json; charset=utf-8
< Content-Length: 62
< Connection: keep-alive
< Set-Cookie: sid=55abaabc7eaef26vvqc84f02423e5ca45e; Path=/; Secure; HttpOnly
< X-Request-Id: 7f72qa2ab-4422-45bf-a455-70dfs26f3e9732
< 
{ [62 bytes data]
{"errors":[{"code":"UNAUTHORIZED","message":"unauthorized"}]}`

Is there a real solution to this problem?

Vad1mo:
- A user can get the CLI Token with that he can do also API Calls and push/pull images with docker.
- Another approach is to use robot accounts, push pull operation. 
 
 There is no other way. 
 
[...]

Issue #14958, opened 2021-05-24
API access when OIDC is configured


mdconner:
**Is your feature request related to a problem? Please describe.**
No way for an account (except admin) to use Harbor API when OIDC is configured.  We expected a User's "CLI Secret" to grant this access; but, found it is limited to docker CLI commands.

**Describe the solution you'd like**
Ability to login (e.g., token, basic auth) as a specific user when OIDC is configured and issue Harbor API calls for things that user has access.

**Describe the main design/architecture of your solution**
A means to login (CLI Secret, or other) to create a time-limited token and/or ability to pass that authentication into every Harbor API call

**Describe the development plan you've considered**
None - not familiar with Harbor's implementation or architecture

**Additional context**
This may be related to #14236 ; but started to go in other directions once tied to Robot accounts

phin1x:
@mdconner whats your use-case for a user to access the harbor api directly? for all the scenarios I can think of, a robot account would be better.

mdconner:
This would be for external applications to interact with Harbor (automate project creation, get list of projects/repositories/images,  identify image digest by tag, etc.).  We have not been able to get robot accounts to login via API due to OIDC.

phin1x:
we are using oidc and robot accounts without problems like this with harbor 2.2. just use basic auth for robot accounts.
what errors do you get?

mdconner:
We get a 401 error using the Robot's CLI Secret as either the Bearer Token or Basic Auth.  How are credentials for a Robot to be passed in?

reasonerjt:
@mdconner  You can use ID token of the user as Bearer Token to access the API

mdconner:
Is there a way to get the ID Token from Harbor?  We have not been able to get it from our OIDC system.

reasonerjt:
@mdconner You should not try to get it from Harbor.  
But if you set the log level to `DEBUG` you can see the id token when you login via OIDC SSO flow.

mdconner:
We have been able to get an OAuth token and can now login via API

Issue #13683, opened 2020-12-03
How to call Harbor api V2 when OIDC configured programatically (without browser login first)


pingcrosby:
What can we help you?

How do we programatically call the  Harbor api V2 when OIDC is configured?.  The only way we can invoke the api seems to be after we login into Harbor (via the UI) ( the api has no /login endpoint ).  Once we have logged in via browser then curling the API works as expected.

However I need to invoke the api calls from within code; so logging in via the UI is not practical. Using either a robot account or personal account always results in unauthorised.

Using OIDC i need to pass the id_token; i am only able to get this if i have previously logged in ( again with a browser )

To Reproduce - 
  Harbor version v2.1.1-5f52168e
  Configured OIDC 
  Create an personal_account with admin access to a project
  Create a robot_account for project

Calling the v2 api with  basic auth using personal_account and using cli secret [ from user profile ] results in 404.. This means we cannot do an initial login; without this initial login we can not then move on to request and "id_token".

```
curl -v -u personal_account:cli_secret "https://hostapi/v2.0/configurations"

auth=$(echo "personal_account:cli_secret" | base64)
curl -v  -H "authorization: Basic $auth" "https://hostapi/v2.0/configurations"
```

Calling the v2 api with  basic auth using robot account and using cli secret results in 404
```
curl -v -u robot_account:cli_secret "https://hostapi/v2.0/configurations"

auth=$(echo "robot_account:cli_secret" | base64)
curl -v  -H "authorization: Basic $auth" "https://hostapi/v2.0/configurations"
```

^^^^ both fail with unauth

flbla:
I'm facing the same issue, 
I saw this comment in another issue : https://github.com/goharbor/harbor/issues/10597#issuecomment-603159112 
and in the FAQ : https://github.com/goharbor/harbor/wiki/Harbor-FAQs#api
but it didn't work for me.

pingcrosby:
Yes the key line in the FAQ is..
_Go to Harbor URL and login with OICD_, here use the username and password for the user you have just created.

Basically you have to log in first via the browser then you can use id_token etc... Obviously this is no good for machine2machine; however I do believe the Harbor team has a fix which will be released in v2.2

reasonerjt:
Strictly speaking it's not a fix to this issue. And the comment in #13093 is quite inaccurate because not all permissions are going to be enabled in robot account -- you can't create a robot account and use it to call ANY API.

[...]

PR #23936, opened 2026-09-15
Auto-onboard OIDC users from a verified ID token on the API

### LINKED ISSUE #23935: OIDC: honour Automatic onboarding on the API ID-token path (users must log into the UI once before bearer calls work)

With `auth_mode: oidc_auth` and **Automatic onboarding** enabled, an API request that carries a valid ID token from the configured provider (`Authorization: Bearer <id_token>`) is still answered with **401** for any user who has not yet logged into Harbor's own UI at least once.

Cause (v2.15.2, unchanged on `main`): `src/server/middleware/security/idtoken.go` → `idToken.Generate` verifies the token, then `user.Ctl.GetBySubIss(sub, iss)`; on `NotFound` it returns `nil` and the request falls through to `unauthorized`. Onboarding exists only in the browser callback (`src/core/controllers/oidc.go` → `Callback` → `userOnboard`), so `oidc_auto_onboard` never applies to the API path. There is no other way to create an OIDC user either: `POST /api/v2.0/users` is refused outside `db_auth`.

Impact: every system that shares the IdP with Harbor and forwards the user's token (a VM/image portal in our case, using Dex cross-client audience) has to tell each user to "open Harbor once first". Earlier reports of the same wall: #13683, #14958, #16047, #16398; #17520 asks for a broader OIDC-based flow.

When `oidc_auto_onboard` is enabled and the subject is unknown, the ID-token path onboards the user exactly like the callback does (username from `oidc_user_claim` with spaces replaced by `_`, e-mail, groups, a random CLI secret) and continues. With `oidc_auto_onboard` disabled nothing changes.

Two details need care, both handled in the PR:

1. *CLI secrets.* `pkg/oidc/secret.go` → `VerifySecret` re-validates the stored OAuth2 token and refreshes it when `oauth2.Token.Valid()` is false. The ID-token path has no refresh token, so the record stores the ID token as the access token with the ID token's own `exp`; on later bearer requests the stored token is renewed when it holds no refresh token and is older than the presented one. Records created by the browser flow (which hold a refresh token) are never modified.
2. *Insert conflicts.* A conflict on `OnboardOIDCUser` is either a concurrent first request from the same user (re-read by sub+iss and continue) or a stale row with the same username / e-mail under a different subject (for example after an IdP connector change); the latter is reported and left to an administrator, nothing half-created remains.

Cost: one extra `oidc_user` read per bearer request for the renew check; a write only when the stored token is older than the presented one, i.e. once per IdP login.

Done: a one-file change to `idtoken.go` plus `idtoken_onboard_test.go` (error-code matching through `errors.Wrap`, stored token shape, DB onboarding path incl. race and stale-username cases, runs under the package's existing Postgres-backed `TestMain`). Verified on a v2.15.2 deployment behind Dex (LDAP connector): a deleted user was recreated on the first API call from a peer client's ID token; an existing browser-onboarded record stayed unchanged. PR follows this issue. A short docs note for the OIDC page (website repo) can accompany it.

**Additional context**

Harbor v2.15.2, Dex 2.x as OIDC provider, `oidc_user_claim: preferred_username`, `oidc_auto_onboard: true`.

aivanov-wq:

With `auth_mode: oidc_auth`, Harbor creates (onboards) an OIDC user only inside the browser callback (`/c/oidc/callback`). The API bearer path (`idToken` security context generator in `src/server/middleware/security/idtoken.go`) verifies the ID token, looks the user up by `sub`+`iss` and returns unauthenticated when nothing is found, so a valid ID token from a trusted peer client is rejected with 401 until the user has visited Harbor's own UI once, even with **Automatic onboarding** enabled.

This change makes the ID-token path honour `oidc_auto_onboard`: when the lookup misses and the setting is on, the user is onboarded from the ID-token claims with the same primitives the callback uses (username from the configured user claim with spaces replaced by `_`, e-mail, groups via `InjectGroupsToUser`, a random CLI secret, `user.Ctl.OnboardOIDCUser`). With `oidc_auto_onboard` disabled the behaviour is unchanged.

Two details:

* **CLI secrets.** `VerifySecret` (`pkg/oidc/secret.go`) re-validates the stored OAuth2 token and refreshes it when expired. This path has no refresh token, so the stored token is the ID token presented as the access token with the ID token's own expiry; on later bearer requests it is renewed only for records that hold no refresh token and are older than the presented token. Records created by the browser flow are never modified.
* **Insert conflicts.** Either a concurrent first request from the same user (resolved by re-reading by `sub`+`iss`) or a stale row with the same username / e-mail under a different subject (reported with an explicit error, nothing half-created). Error codes are matched by walking the wrap chain because `errors.IsErr` only inspects the outermost `*errors.Error`, which is code-less after `errors.Wrap` as returned by `OnboardOIDCUser`.

Cost: one extra `oidc_user` read per bearer request for the renew check; a write only when the stored token is older than the presented one, i.e. once per IdP login. The onboarding comment is `Onboarded via OIDC (ID token)` (29 chars; `harbor_user.comment` is varchar(30)) so such users are distinguishable from callback-onboarded ones.

**Tests** (`idtoken_onboard_test.go`, runs under the package's existing Postgres-backed `TestMain`): error-code matching through `errors.Wrap`; stored token shape and validity; DB onboarding path — user created with the expected fields and OIDC metadata, repeated first contact resolves to the same user, a different subject reusing the username is refused without creating a row, an empty username claim is refused.

**Verified** on a v2.15.2 deployment behind Dex (LDAP connector, cross-client audience): a deleted user was recreated on the first API call carrying a peer client's ID token; an existing browser-onboarded user's record stayed unchanged.

# Issue being fixed
Fixes #23935 (related: #13683, #14958, #16047, #16398)

- [ ] Label the PR as needed. "release-note/ignore-for-release, release-note/new-feature, release-note/update, release-note/enhancement, release-note/community, release-note/breaking-change, release-note/docs, release-note/infra, release-note/deprecation" — suggest `release-note/enhancement` (I cannot set labels)
- [x] Considered the docs impact and opened a new docs issue or PR with docs changes if needed in [website repository](https://github.com/goharbor/website). — `docs/administration/configure-authentication/oidc-auth.md` says the CLI can be used "after you have … logged into the Harbor interface for the first time"; with this change an API client presenting the ID token onboards the user too when Automatic onboarding is on. I will open the docs PR once this is accepted.


aivanov-wq:
Pushed adfbb9edf to address the codecov patch result: the first commit tested the helpers but never drove `idToken.Generate` itself, so the new branches and `renewStoredToken` counted as unexecuted.

`idtoken_generate_test.go` now exercises `Generate` end to end against the package's Postgres-backed `TestMain`, standing in for the provider through two package variables (`verifyIDToken`, `userInfoFromIDToken`, same pattern as `uctl` in `oidc_cli.go`):

- auto onboard on: first contact creates the user from the claims, a second call resolves to the same user;
- auto onboard off: unchanged (unauthenticated);
- renewal of a record created from an ID token when a fresher token arrives, and not when an older one does;
- a browser-created record holding a refresh token is left untouched;
- a record with an unusable stored token is renewed;
- a stale username under a different subject is refused without creating a row;
- rejected token, unparsable claims, missing username claim, non-API path.

Per-function coverage of `idtoken.go` with the package tests: `Generate` 89.2%, `onboardFromIDToken` 81.0%, `renewStoredToken` 80.0%, `encryptIDTokenAsToken` 75.0%, `hasCode` 100%. What remains uncovered are the error branches that need a missing secret key or a failing DB write.

PR #23941, opened 2026-09-15
Allow GC worker counts above 10 

meik99:

GC worker count was capped at 10. The limit was enforced in three places: the API validation in `validateWorkers`, a hardcoded `WORKER_OPTIONS` dropdown in the portal, and the validation error message.

The cap doesn't reflect what the job actually does. Blob removal can be mostly IO-bound, for example against object storage like S3, so a worker count well above the core count is reasonable there. On a local filesystem you may want to stay low. Either way it's a deployment decision, not something that needs to be hardcoded.

This removes the upper bound. Workers must still be greater than 0. The portal dropdown is replaced with a number input plus helper text explaining the tradeoff, and the error message no longer mentions a limit that isn't enforced.

# Issue being fixed

There was no open issue but the one in my heart and soul


Vad1mo:
I am kind og agrees and kind of disagree here. 

if the harbor operator and harbor admin are the same the limit is for the ops people do decice. 
if hover the admin is not the operator, them we have a problem. 

To make it safer, I would probaly add an env var that adds a cap that the operator can set.

meik99:
> I am kind og agrees and kind of disagree here.
> 
> if the harbor operator and harbor admin are the same the limit is for the ops people do decice. if hover the admin is not the operator, them we have a problem.
> 
> To make it safer, I would probaly add an env var that adds a cap that the operator can set.

Makes sense, just one question: how should I handle that in the UI? 
An env var on the server cannot be read from the angular frontend, so a user side form validation is not as easily possible. I mean, would be but requires some nginx config tomfoolery iirc or some additional API endpoint i guess. Should I add a warning that a limit could be set to the helper text and show the error as an alert if the user entered a number above the operator limit?

meik99:
I have implemented the env var and some checks on the server, the jobservice and the ui

meik99:
@Vad1mo could I have another review?

PR #23942, opened 2026-09-15
fix(retention): drop retention_id metadata on project delete

### LINKED ISSUE #17255: when project which has a retention policy deleted, the policy is still in db

A clear and concise description of what you expected to happen and what's the actual behavior. If applicable, add screenshots to help explain your problem.

when project which has a retention policy deleted, the policy is still in db.
does this need clean up?

Please provide the steps to reproduce this problem.

1. create a project
2. create a retention policy with given project
3. delete the project
4. get retention by /retentions/{id} api

**Versions:**

- harbor version: [2.3.1]
- docker engine version: [20.10.7]
- docker-compose version: [1.24.1]

**Additional context:**

warisshaikh1:
# Comprehensive Summary of change

Deleting a project removes its tag retention policies, but it never removed the
`project_metadata` row that points at them. Every deleted project that had a retention
policy therefore leaks a `retention_id` row referencing a policy that no longer exists.
A user on the issue reported 3189 such rows on a single instance:

```sql
select count(*) from project as p
  join project_metadata as pm on p.project_id = pm.project_id
  where p.deleted = 't' and pm.name = 'retention_id';
 count
-------
  3189
```

The cleanup already exists on the other path: `DELETE /retentions/{id}` drops the policy
*and* the `retention_id` metadata, in `retentionAPI.DeleteRetention`. The project-delete
path goes through `retentionCtl.DeleteRetentionByProject` (via the `DELETE_PROJECT` event
handler) instead, and that function stops after deleting the policy rows.

This adds the matching metadata cleanup to `DeleteRetentionByProject`, so both paths leave
the same state behind. It runs only after the policies are successfully deleted, so a
failed policy deletion still leaves the metadata consistent with what is in the DB. It also
cleans up already-orphaned rows on the next project deletion, since `Delete` is a no-op
when no row matches.

The controller now takes its metadata manager from the shared `pkg.ProjectMetaMgr`, the
same instance the rest of the codebase uses. That matters because when caching is enabled
it is the Redis-backed wrapper, whose `Delete` also invalidates the cached entry — a plain
`metadata.New()` would have left a stale value in the cache. This mirrors how the
constructor already sources `pkg.ProjectMgr` and `pkg.RepositoryMgr`.

No schema change and no API change; the scope is one function.

## Testing

Added `TestDeleteRetentionByProject`, which creates a project-scoped retention policy,
deletes it by project, and asserts both that the policy row is gone and that the
`retention_id` metadata is cleared. Verified it fails without the one-line fix
(`Should have called with given arguments`) and passes with it.

Ran against a local PostgreSQL 16 with the repo's migrations applied:

```
ok  github.com/goharbor/harbor/src/controller/retention
ok  github.com/goharbor/harbor/src/controller/event/handler/internal
ok  github.com/goharbor/harbor/src/pkg/retention/...
```

# Issue being fixed
Fixes #17255


warisshaikh1:
@OrlinVasilev @MinerYang @bupd please review the PR

bupd (approved):
/lgtm

chlins (approved):
lgtm

PR #23971, opened 2026-09-19
fix: return the responder built by SendError instead of discarding it

arpitjain099:
`BaseAPI.SendError` is a constructor, `return NewErrResponder(err)` and nothing else, so calling it without returning the value does nothing. `retention.go` gets this right with `return r.SendError(...)`, but five call sites drop it.

Three are the `Prepare` methods of the artifact, repository and scan handlers. When `unescapePathParams` fails they build a responder, throw it away, and return nil, which the generated wrapper reads as "carry on", so the handler runs with the raw parameter it could not unescape. The other two are `GetConfigurations`, which answers 200 with a null payload after `toResponseModel` fails, and the purge history listing, which answers 200 with a non-zero total and an empty list after the execution list fails.

The fix is the missing `return` in each. I added a test that calls all three `Prepare` methods with a repository name that `url.PathUnescape` rejects and asserts the responder is not nil; it fails on main and the handler package is green with the change.


wy65701436 (approved):
lgtm

chlins (approved):
lgtm

Issue #23178, opened 2026-04-27
GC silently fails to delete blobs on S3-compatible storage (NetApp ONTAP S3), reports success with fake freed space


velmoga:
**Expected behavior and actual behavior:**

**Expected:** Garbage Collection deletes unreferenced blobs from S3 storage and reports the actual freed space.

**Actual:** GC completes with `"job_status":"Success"` and reports hundreds of GBs freed per run, but **no blobs are actually deleted from S3 storage**. Storage usage grows indefinitely. In our environment Harbor reports 696 GB usage while actual S3 storage is 3.98 TB.

The root cause is two interacting issues:

1. The Distribution S3 driver's `Delete()` fails with `PathNotFoundError` on S3-compatible stores that don't normalize leading slashes in object keys. The `storagePathRoot` in `registry/storage/paths.go` produces paths like `/docker/registry/v2/blobs/sha256/...` (with leading `/`). AWS S3 silently normalizes this, but NetApp ONTAP S3 (and likely other S3-compatible implementations) treats `/docker/...` as a distinct key from `docker/...`. Objects are written and read successfully (those code paths normalize correctly), but the delete code path constructs a key that doesn't match any existing object. Related upstream issue: https://github.com/distribution/distribution/issues/1948 (open since 2016).

2. Harbor GC's `ignoreNotFound` wrapper in `src/jobservice/job/impl/gc/util.go` swallows the resulting `PathNotFoundError`, treats it as success, increments the freed-space counter, and the blob record is deleted from the database. The GC job reports overall "Success" with inflated freed-space numbers.

**Steps to reproduce the problem:**

1. Deploy Harbor with S3 storage backed by NetApp ONTAP S3 (or any S3-compatible store that doesn't normalize leading-slash object keys)
2. Push several images
3. Delete some tags/repositories to create GC candidates
4. Run Garbage Collection
5. Observe GC reports "Success" with freed space
6. Check actual S3 storage — blobs are still present

**registryctl logs show every blob delete fails (but GC ignores this):**
```
time="2026-04-26T00:00:44Z" level=info msg="Deleting blob: /docker/registry/v2/blobs/sha256/16/16c5846df37d117cb9f11f980f5a71ee9a84f31b09de5d1358d22287386d81f6"
[...]

PR #21896, opened 2025-04-17
Upgrade to distribution (registry) v3

AaronDewes:
If you're looking for a project with Distribution v3 already integrated (independently of this PR), please check out [Harbor-Next](https://github.com/container-registry/harbor-next).

---

# Comprehensive Summary of your change

Previously submitted as #19784, this uses the stable distribution v3.

This includes all the benefits of the v3 distribution, but also all breaking changes.

Most notably, Image Manifest v2 Schema v1 support has been dropped, as well as the `oss` and `swift` storage drivers.

Currently, this still relies on v2's github.com/docker/distribution/registry/client/auth/challenge, because that code has been removed from the public API in v3. The corresponding code has been copied to Harbor.

There's still a change to the redis config needed, as distribution changed the format of the options, so this is a draft currently.

# Issue being fixed
Fixes #(issue)

Please indicate you've done the following:
- [x] Well Written Title and Summary of the PR
- [x] Label the PR as needed. "release-note/ignore-for-release, release-note/new-feature, release-note/update, release-note/enhancement, release-note/community, release-note/breaking-change, release-note/docs, release-note/infra, release-note/deprecation"
- [x] Accepted the DCO. Commits without the DCO will delay acceptance.
- [ ] Made sure tests are passing and test coverage is added if needed.
- [ ] Considered the docs impact and opened a new docs issue or PR with docs changes if needed in [website repository](https://github.com/goharbor/website).

PR #23972, opened 2026-09-19
fix(gc): do not count blobs missing from storage as freed space

velmoga:
### Summary

`sweep()` wraps `registryCtlClient.DeleteBlob()` in `ignoreNotFound()`, so a blob that registryctl reports as not found does not fail the job. That is intended and keeps the GC idempotent across re-runs. But the blob's size was still added to `sweepSize`, so the job reported it as freed space although nothing was removed from storage in this run.

On a backend where every delete ends in `PathNotFoundError` (#23178, S3-compatible NetApp ONTAP S3) this produced GC jobs with `job_status: Success` and hundreds of GB of `freed_space` per run while the bucket kept growing. The number that operators use to judge whether GC works was the one thing hiding that it did not.

This PR:

- remembers whether the last `DeleteBlob` attempt returned not-found and skips that blob's size when computing `freed_space`;
- logs a warning per such blob (`blob not found in storage, its size is not counted as freed space: <digest>`), so the mismatch shows up in the job log;
- adds `TestSweepBlobNotFound`, which fails on `main` (`freed_space: 1234`) and passes with this change (`freed_space: 0`).

Intentionally unchanged: the blob record is still deleted from the database and still counted in `purged_blobs`, as before. `purged_manifests` is not affected, manifest deletion never contributed to `freed_space`.

### Which issue(s) this PR fixes

Refs #23178. This covers the reporting half only. On the reporting environment the not-found answers were genuine by the time Harbor saw them: the S3 driver's `Delete()` in distribution 2.8.x turns any `ListObjects` error into `PathNotFoundError`, and on that backend a bucket-policy gap (`ListBucket` missing for the registry user, by elimination) made a fraction of deletes fail that way for a few weeks. Harbor dropped the DB rows, counted the sizes as freed and left the objects behind. Returning the listing error instead of masking it is a driver change and belongs to #21896; this PR only makes the job report what actually happened.

A concrete example from that environment: on 2026-07-07 two consecutive GC runs each reported ~9,480 GiB freed and ~377,500 purged blobs against a 4.9 TiB bucket with ~93,000 blob rows, while the real drop in bucket usage was zero. With this change those runs would have reported a small `freed_space` and a log full of `blob not found in storage` warnings.

### Special notes for reviewer

- The `notFound` flag is assigned on every retry attempt, so it reflects the final outcome: a not-found ends the retry loop, a read-only abort exits through the error branch before the size is counted.
- Net change is 10 lines in `garbage_collection.go`; the rest is the test.

### Validation

- `go test ./jobservice/job/impl/gc/ -run TestGCTestSuite -count=1` against a local PostgreSQL: all subtests pass except `TestRun`, which needs Redis and fails identically on `main`.
- `TestSweepBlobNotFound` verified to fail without the change.
- `go vet` and `gofmt` clean.

### Does this PR introduce a user-facing change?

```release-note
GC no longer counts blobs that are missing from the storage as freed space and logs a warning for each of them.
```

- [x] Title and summary describe the change.
- [x] Commit is signed off under the DCO.
- [x] Regression test added, affected package tests pass.
- [x] Docs impact considered: none.
- [ ] Release-note label to be applied by a maintainer (`release-note/update`).


chlins:
@velmoga Please fix the unit tests.

velmoga:
@chlins Fixed in 56ed4ce — the new test was missing the UpdateBlobStatus mock, the GC package is green locally with -race. The CI run for the new commit is waiting for workflow approval. Would you prefer the fix squashed into the original commit, or is it fine to keep it separate and squash on merge?

chlins (approved):
lgtm

wy65701436 (approved):
lgtm

wy65701436:
@velmoga thanks for the fixing, can you help to back port this into release-2.15?

velmoga:
@wy65701436 Sure, the backport to release-2.15.0 is in #23980.

PR #23973, opened 2026-09-21
test: Add a reproduction for the DB pool deadlock in #23879 (not a fix)

bupd:
> **This is a reproduction, not a fix.**
>
> It changes no production code. The defect it demonstrates is in `src/` and is still present on `main`, and it should be fixed — see [What needs fixing](#what-needs-fixing) below for what that takes. This PR exists so that #23879 has something runnable attached to it, and so a future fix has a way to be shown to work.

## What this is

`tests/dbpool-exhaustion/` is a self-contained, fully containerised reproduction of the pool deadlock in #23879. It builds core from the checkout and drives the real HTTP API; nothing is mocked and no production code is touched.

```bash
make gen_apis            # if src/server/v2.0/restapi is not generated yet
cd tests/dbpool-exhaustion && ./run.sh
```

Everything runs under the compose project `harbor-dbpool-repro` with no published host ports, so it cannot collide with a Harbor install on the same machine. The only setting moved away from a shipped default is `POSTGRESQL_MAX_OPEN_CONNS`, lowered from 100 to 5 so the arithmetic is small enough to watch. PostgreSQL is started with `max_connections=200`, so the server itself can never be what runs out. `CACHE_ENABLED` is `false`, so the config-cache singleflight is not involved.

## The mechanism

`transaction.Middleware` opens a transaction — and takes a pooled connection — for every non-GET request, and holds it for the whole request. Inside that span, `log.Middleware` calls `commonevent.Metadata.PreCheck` *before* invoking the next handler, and for a DELETE on a resource with `ShouldResolveName` that calls `IDToNameFunc`:

```go
// src/pkg/auditext/event/user/user.go
// use different context to so that the user is visible before the transaction is committed
user, err := pkgUser.Mgr.Get(orm.Context(), int(id))
```

`orm.Context()` builds a fresh context with a *pooled* Ormer, so that lookup takes a **second** connection from the same pool while the request's first one is still held. `POSTGRESQL_MAX_OPEN_CONNS` such requests in flight at once is a closed cycle.

Nothing bounds the wait: beego's `Ormer.Begin` is `BeginWithCtx(context.Background())`, and its non-`WithCtx` query methods pass `context.Background()` as well, so no request deadline, client disconnect or shutdown signal reaches the pool.

`notification.Middleware` sits outside `transaction.Middleware` by design, so audit events publish after the commit. That protects `Resolve`. It does not protect `PreCheck`, which is the one that runs inside the transaction.

## What a run shows

```
PHASE 0   ping/health/projects/systeminfo all 200
PHASE 1    4 concurrent deletes -> 4/4 HTTP 200, all inside 0.09s
PHASE 2   10 concurrent deletes -> 0 complete; 5 sessions idle in transaction
PHASE 3   unrelated GETs hang too; GET /api/v2.0/ping still 200 in 0.001s
PHASE 4   every client gives up at 45s; all 5 transactions still open
PHASE 5   nothing was deleted
```

`pg_stat_activity` during the wedge:

```
        state        | wait_event | sessions |  max_xact_age
---------------------+------------+----------+----------------
 idle in transaction | ClientRead |        5 | 00:00:08.00831
```

All sessions `idle in transaction` on `ClientRead`, no lock waits — PostgreSQL is idle and waiting on Harbor, not the other way round.

Two details are worth pulling out.

**Phase 3 is why this is an outage rather than a broken endpoint.** A `GET` skips the transaction middleware and touches none of the contended rows, but it still needs one pooled connection, so it hangs with the rest. Meanwhile `GET /api/v2.0/ping` answers in a millisecond, because `pingSkipper` keeps it clear of every DB-touching middleware — Kubernetes liveness and readiness stay green on a core that is serving nothing.

**Phase 4 is why it needs an operator.** After every client has hung up, all the transactions are still open. There is no deadline left anywhere in the stack to cancel them. `run.sh` restarts core afterwards to show that this is what clears it.

Core logs no error at all through any of this — no 500, no timeout, no rollback. Every wedged request gets as far as the audit-log config read in `PreCheck` and stops.

## The threshold

`./sweep.sh` restarts core at each pool size and fires real concurrent deletes, with no barrier and no synthetic handler:

```
pool=5     4 concurrent ->  4/4  completed, 0 idle-in-tx  |   5 -> 0/5  completed,  5 idle-in-tx
pool=10    9 concurrent ->  9/9  completed, 0 idle-in-tx  |  10 -> 0/10 completed, 10 idle-in-tx
pool=25   24 concurrent -> 24/24 completed, 0 idle-in-tx  |  25 -> 0/25 completed, 25 idle-in-tx
```

The boundary is structural rather than statistical, and it does not need the requests lined up artificially: a request holds its transaction connection for its entire duration, so requests starting milliseconds apart are all still holding connection #1 when the last reaches for #2.

Pool 25 is the configuration the instance in #23879 was running. Twenty-four concurrent bulk deletes are fine; the twenty-fifth takes the instance down.

## Other call sites with the same shape

Not exercised by this reproduction, but present and reachable inside the request transaction:

- `src/pkg/oidc/helper.go` — `populateGroupsDB`, on every OIDC-authenticated write
- `src/pkg/authproxy/http.go` — http_auth writes, `POST /c/login`
- `src/core/api/internal.go` — `POST /api/internal/syncquota`
- `src/pkg/auditext/event/member/member.go` — `ensureORMContext`

## What needs fixing

To be explicit about what this PR deliberately leaves undone. Two changes are needed, and they are independent.

**1. A request must never hold two pooled connections.** Thread the request context through the call sites that build their own: `IDToNameFunc` (`src/pkg/auditext/event/user/user.go`), `ensureORMContext` (`src/pkg/auditext/event/member/member.go`), `populateGroupsDB` (`src/pkg/oidc/helper.go`), `src/pkg/authproxy/http.go`, `src/core/api/internal.go`. Where a value genuinely must be read outside the transaction — `userIDToName`'s own comment says the user has to be visible before the commit — the read wants deferring until after the commit rather than running on a second connection during it. `commonevent.Metadata` is already carried through to `notification.Middleware`, which runs outside the transaction by design, so the name could be resolved there instead of in `PreCheck`.

**2. The wait for a connection must be bounded.** `ormerTx.Begin` (`src/lib/orm/tx.go`) calls beego's `Ormer.Begin()`, which is `BeginWithCtx(context.Background())` (`beego/v2@v2.3.10/client/orm/orm.go:530`). Routing it through `BeginWithCtx(ctx)` would let a request deadline reach `BeginTx`. Beego's non-`WithCtx` query methods pass `context.Background()` too, at fourteen call sites in that one file, so the same applies to reads taken inside a request.

Change 2 fixes no deadlock by itself, and that is the point: it converts every one of them — the cycle above, the other call sites listed above, and the ones not found yet — from a total outage into slow requests or errors on one endpoint. Change 1 without change 2 leaves the next such call site free to take the whole instance down.

A patched tree should make `./run.sh` fail: phase 2 should complete instead of wedging, or return an error instead of hanging forever.

Worth considering alongside them: `GET /api/v2.0/ping` deliberately skips every DB-touching middleware, which is right for what that endpoint is for, but it means a readiness probe cannot tell a healthy core from a wedged one.

## Scope

Tests only. No production code, no schema, no API changes, and nothing that runs in CI — the reproduction is opt-in and deadlocks by design, so it must not be wired into a test job.

Happy to reshape this if maintainers would rather have it as a build-tagged Go test under `src/` than as a compose rig under `tests/`, or to fold it into a PR that carries the fix as well.


wy65701436 (approved):
lgtm

